PR #130, opened 2017-02-02
prometheus: put endpoint in separate label

brancz:
Before the job label and the endpoint were concatenated into the job
label, however, this same effect of uniqueness can be achieved by having
granular labels that are separate.

@alexsomesan @fabxc


fabxc:
👍

alexsomesan:
👍

PR #131, opened 2017-02-02
Properly exit on error

fabxc:
@alexsomesan @brancz 

This uses the `errgroup` package and contexts. On the first error returned by either operator's `Run()`, the context gets canceled and ensures the other one exits as well and with it the program as a whole.

I disabled closing of the server for now and just let it run. It's not graceful and just returns confusing errors.

We can do it properly once https://tip.golang.org/pkg/net/http/#Server.Shutdown lands in go1.8

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -15,14 +15,16 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


brancz:
lgtm :+1:

alexsomesan:
That's pretty cool. 👍

PR #138, opened 2017-02-09
test: Add e2e alertmanager & prometheus exposure tests

mxinden:
Ensuring functionality of the "Exposing prometheus and alertmanaer"
tutorial. Testing exposure of alertmanager and prometheus via Nodeport,
Kubernetes API and Ingress controller.

https://github.com/coreos/prometheus-operator/blob/master/Documentation/exposing-prometheus-and-alertmanager.md

### DIFF
--- a/test/e2e/alertmanager_e2e_test.go
+++ b/test/e2e/alertmanager_e2e_test.go
@@ -15,7 +15,11 @@
 package e2e
 
 import (
+	"fmt"
+	"k8s.io/client-go/pkg/api/v1"
+	"net/http"
 	"testing"
+	"time"
 )
 
 func TestAlertmanagerCreateDeleteCluster(t *testing.T) {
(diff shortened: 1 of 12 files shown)


mxinden:
@brancz Could you run the tests on your local machine. They all run green on mine.

fabxc:
My cluster still has test-related pods in it after successful cleanup and exit: 

```
prometheus-operator-e2e-tests-gyf9wmk7ezk5b   po/default-http-backend-1bg0m       1/1       Running   0          6h
prometheus-operator-e2e-tests-gyf9wmk7ezk5b   po/nginx-ingress-controller-rn6w5   1/1       Running   0          6h
```

test/e2e/framework/alertmanager.go
```diff
@@ -53,29 +53,38 @@ func (f *Framework) MakeBasicAlertmanager(name string, replicas int32) *v1alpha1
 	}
 }
 
-func (f *Framework) MakeAlertmanagerService(name, group string) *v1.Service {
-	return &v1.Service{
+func (f *Framework) MakeAlertmanagerNodePortService(name, group string) *v1.Service {
+	return f.MakeAlertmanagerService(name, group, v1.ServiceTypeNodePort)
+}
```
fabxc: How about making this accept the actual node port as an argument, calling `MakeAlertmanagerService` and setting the specified node port on the returned service.

This would make it clearer in the test why we are actually checking for `30903`.

test/e2e/framework/prometheus.go
```diff
@@ -86,29 +86,36 @@ func (f *Framework) MakeBasicServiceMonitor(name string) *v1alpha1.ServiceMonito
...
 				},
 			},
 			Selector: map[string]string{
 				"prometheus": name,
 			},
 		},
 	}
+	if serviceType == v1.ServiceTypeNodePort {
+		service.Spec.Ports[0].NodePort = 30900
+	}
```
fabxc: Same argument as for the Alertmanager node port.

test/e2e/framework/replication-controller.go
```diff
@@ -0,0 +1,60 @@
...
+	"os"
+)
+
+func createReplicationControllerViaYml(filepath string, f *Framework) error {
+	manifest, err := os.Open(filepath)
+	if err != nil {
+		return err
+	}
+
+	replicationController := v1.ReplicationController{}
```
fabxc: `var replicationController v1.ReplicationController` is the typical way to instantiate a zeroed object in Go.
With such a small function scope, i.e. not a lot of identifies to keep track of, one would probably go for a shorter name like `rc`.

fabxc:
A few comments but looks great in general!

brancz:
The ingress tests are failing on my machine using minikube and the kvm driver. I'll try to get my virtualbox and virtualbox driver working and give it a spin on that as well.

fabxc:
Any data on what is failing specifically?

brancz:
Not really

```
=== RUN   TestExposingAlertmanagerWithIngress
2017/02/10 17:03:05 Creating Alertmanager (prometheus-operator-e2e-tests-nfpjxusgpj0zd/main)
2017/02/10 17:03:40 Deleting Alertmanager (prometheus-operator-e2e-tests-nfpjxusgpj0zd/main)
--- FAIL: TestExposingAlertmanagerWithIngress (38.43s)
        alertmanager_e2e_test.go:191: Waiting for url http://192.168.42.191:80/metrics timed out
```

fabxc:
That seems to be a networking issue between your host and the VM.

brancz:
I don't think so as the other tests performing HTTP requests against it are passing.

brancz:
Strange! Using the virtualbox driver the tests pass. We should figure out why, but it shouldn't block this.

test/e2e/alertmanager_e2e_test.go
```diff
@@ -79,3 +83,111 @@ func TestAlertmanagerVersionMigration(t *testing.T) {
...
+
+	if err := framework.CreateAlertmanagerAndWaitUntilReady(alertmanager); err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := framework.KubeClient.CoreV1().Services(framework.Namespace.Name).Create(alertmanagerService); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := framework.WaitForServiceReady(time.Minute, alertmanagerService.Name); err != nil {
```
brancz: maybe we can merge these with the above function to keep the theme of `CreateXAndWaitUntilReady`

test/e2e/alertmanager_e2e_test.go
```diff
@@ -79,3 +83,111 @@ func TestAlertmanagerVersionMigration(t *testing.T) {
...
+
+	if _, err := framework.KubeClient.CoreV1().Services(framework.Namespace.Name).Create(alertmanagerService); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := framework.CreateIngress(ingress); err != nil {
+		t.Fatal(err)
+	}
+
+	err := framework.WaitForHTTPSuccessStatusCode(time.Second*30, fmt.Sprintf("http://%s:80/metrics", framework.ClusterIP))
```
brancz: Why specify port 80 explicitly here? If it's not required I would skip it as it makes it seem like it is.

test/e2e/framework/alertmanager.go
brancz: Agreed

test/e2e/framework/framework.go
```diff
@@ -208,6 +208,45 @@ func waitForPodsReady(client v1client.CoreV1Interface, timeout time.Duration, ex
 	}
 }
 
+func (f *Framework) WaitForServiceReady(timeout time.Duration, serviceName string) error {
```
brancz: looks like we're duplicating these ticker functions, if you think we're going to keep adding these we should maybe extract the common ticker code

test/e2e/framework/ingress.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (f *Framework) CreateIngress(i *v1beta1.Ingress) error {
+	_, err := f.KubeClient.Extensions().Ingresses(f.Namespace.Name).Create(i)
+	return err
+}
+
+func (f *Framework) SetupNginxIngressControllerIncDefaultBackend() error {
+	// Create Nginx Ingress Replication Controller
+	if err := createReplicationControllerViaYml("./framework/ressources/nxginx-ingress-controller.yml", f); err != nil {
```
brancz: This is great! Once we get this merged we should extract the struct we instantiate in go code right now into yaml files.

fabxc: I had some thoughts on this but left them out before.
I advocated for not using plain YAML files before in other places – e.g. in cases where operators create objects. But as we see in the case of the ingress that gets really noisy and nearly impossible to read.

Should we use YAMLs everywhere then? What about those where we insert dynamic values. We could use `text/template` there to populate them.

brancz: Either that or parse and then populate, doesn't seem extremely clean but realistically the exact same result.

test/e2e/prometheus_e2e_test.go
```diff
@@ -299,6 +299,110 @@ func TestPrometheusAlertmanagerDiscovery(t *testing.T) {
...
+			t.Fatal(err)
+		}
+	}()
+
+	if err := framework.CreatePrometheusAndWaitUntilReady(basicPrometheus); err != nil {
+		t.Fatal("Creating prometheus failed: %s", err)
+	}
+
+	if err := framework.CreatePrometheusServiceAndWaitUntilReady(service); err != nil {
+		t.Fatal("Creating prometheus service failed: ", err)
```
brancz: "%s" is missing here

test/e2e/framework/framework.go
mxinden: @brancz You mean something like a function "ticker" that takes a function with the specific logic?

mxinden: @brancz What do you think about a centralized standardized timeout value. Might come in handy when we have to increase it in case of slow infrastructure.

test/e2e/framework/ingress.go
mxinden: @brancz @fabxc Should I replace all instantiated structs into yml files?

I would stick with instantiating structs for small objects and create yml files for big ones, but I am fine with extracting them all to yml files as well.

test/e2e/prometheus_e2e_test.go
mxinden: @brancz `t.Fatal` is only concatenating strings.

test/e2e/framework/framework.go
mxinden: @brancz I moved every occurrence of ticker  into a `framework.Poll` function.

test/e2e/alertmanager_e2e_test.go
```diff
@@ -79,3 +83,111 @@ func TestAlertmanagerVersionMigration(t *testing.T) {
...
+
+	if _, err := framework.KubeClient.CoreV1().Services(framework.Namespace.Name).Create(alertmanagerService); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := framework.WaitForServiceReady(time.Minute, alertmanagerService.Name); err != nil {
+		t.Fatal(err)
+	}
+
+	ProxyGet := framework.KubeClient.CoreV1().Services(framework.Namespace.Name).ProxyGet
```
fabxc: No variables starting with uppercase in Go :)

test/e2e/prometheus_e2e_test.go
brancz: oops, totally right, then the %s is not required in the fatal above this

test/e2e/framework/ingress.go
brancz: I think we can continually improve this, no need to move everything right away, but whenever we touch them, we move them

fabxc:
Comment on `t.Fatal(err)`, and by extension `if err := f(); err != nil { return err}`. This is an extremely common pattern in Go and something I'm just trying to get is away from in various places.

There's a nice package: https://godoc.org/github.com/pkg/errors

It allows you to wrap errors into appropriate context so you can understand where they are coming from when errors are propagated up a long call chain until finally being shown to the user.

I'd recommend using it extensively. Let's get in this PR first and worry about fixing this up in various places later.

Similarly, in various Prometheus projects I started to introduce https://godoc.org/github.com/stretchr/testify/require.
I haven't gotten any resistance yet, so it seems like people are okay with that and we can use it consistently.
Essentially just a bunch of boilerplate avoiding all the `if err != nil { ...}` occurrences in tests when we know that we just want to fail.

mxinden:
@fabxc @brancz I included all of your comments. Thank you very much for the feedback! Any other concerns?

I will look into the error handling improvement.

brancz:
lgtm :+1:

PR #141, opened 2017-02-14
prometheus: syncronize nodes to enpoints object

brancz:
As the kubelet will not be self-hosted for the time being, we need to be able to handle it anyways. This is an implementation of a feature synchronizing nodes into an Endpoints object for discovery by Prometheus.

@fabxc

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -45,6 +45,7 @@ func init() {
 	flagset.StringVar(&cfg.TLSConfig.CertFile, "cert-file", "", " - NOT RECOMMENDED FOR PRODUCTION - Path to public TLS certificate file.")
 	flagset.StringVar(&cfg.TLSConfig.KeyFile, "key-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to private TLS certificate file.")
 	flagset.StringVar(&cfg.TLSConfig.CAFile, "ca-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to TLS CA file.")
+	flagset.StringVar(&cfg.KubeletObject, "kubelet-object", "", "Endpoints object to write kubelets into in format \"namespace/name\"")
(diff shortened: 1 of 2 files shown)


cmd/operator/main.go
```diff
@@ -45,6 +45,8 @@ func init() {
 	flagset.StringVar(&cfg.TLSConfig.CertFile, "cert-file", "", " - NOT RECOMMENDED FOR PRODUCTION - Path to public TLS certificate file.")
 	flagset.StringVar(&cfg.TLSConfig.KeyFile, "key-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to private TLS certificate file.")
 	flagset.StringVar(&cfg.TLSConfig.CAFile, "ca-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to TLS CA file.")
+	flagset.StringVar(&cfg.KubeletObjectName, "kubelet-object-name", "kubelet-prometheus-discovery", "Endpoints object name to write kubelets into")
+	flagset.StringVar(&cfg.KubeletObjectNamespace, "kubelet-object-namespace", "kube-system", "Endpoints object namespace to write kubelets into")
```
fabxc: That can probably boil down to `<ns>/<name>` and split later.

pkg/prometheus/operator.go
```diff
@@ -222,6 +239,93 @@ func (c *Operator) handleUpdatePrometheus(old, cur interface{}) {
...
+						Port: 10250,
+					},
+				},
+			},
+		},
+	}
+
+	cache.ListAll(c.nodeInf.GetStore(), labels.Everything(), func(obj interface{}) {
+		n, ok := obj.(*v1.Node)
+		if ok {
```
fabxc: Skip the `ok`. If it doesn't work its a bug and should panic.

brancz:
@fabxc applied changes and tested on minikube.

Issue #149, opened 2017-02-16
Missing headless service


aerokite:
When creating "Prometheus" instance, operator is creating a headless service "prometheus-operated". 

* What is the purpose of this "headless"  service?

And for each Prometheus, a StatefulSet is created. But no headless service named "prometheus" which is used in  StatefulSet is created. 

* Why not using "prometheus-operated" as governing Service in StatefulSet?
* Or why not creating a headless service for each statefulset by operator?

brancz:
Actually the "prometheus-operated" is meant to be the governing `Service`. This is a bug introduced by some recent renamings. We decided to have one of these `Service`s per namespaces where at least one `StatefulSet` is created in, as the `Pod`s are all sufficiently differently named there is no need for further separation.

TamalSaha:
@brancz totally makes sense. Any timeline for the fix?

brancz:
As this is a bug we will be addressing this very shortly and do a release as soon as we've verified that the fix is working.

brancz:
#150 is merged so this can be closed. Thanks for reporting and fixing!

PR #150, opened 2017-02-17
Use governingServiceName as ServiceName in StatefulSet for Prometheus

aerokite:
This is supposed to fix missing headless service problem(#149).


brancz:
I was actually just testing this exact thing locally :slightly_smiling_face:

brancz:
Unfortunately I'm at the airport and internet speed is limited, so it will take me until the afternoon to properly test this.

brancz:
e2e tests passing and tested against kube-prometheus

lgtm :+1: thanks!

PR #155, opened 2017-02-21
prometheus: allow components not to relabel namespace label

### LINKED ISSUE #154: Add possibility to set whether target should be namespaced
In very specific Kubernetes components the metrics contain the `namespace` label which is relabelled onto the target, which causes label collisions. As this is special to Kubernetes components it's somewhat reasonable to have special code for this in the Prometheus Operator, however, it's a very loose concept so it should rather be kept within an annotation and outside of the TPR.

Proposed annotation name on discovered services is: `alpha.monitoring.coreos.com/namespaced`

The resulting behavior is kept as is when the annotation is set to `true` or _unset_, but when set to `false`, the `namespace` label will not be populated by discovery, therefore the exposed metrics of the target are free to use it.

@fabxc

brancz:
For components that want to expose metrics including a namespace label, like Kubernetes components do, we need to allow those components not to get the namespace label relabelled onto them when doing discovery.

Components this is required for:

* kubelets (nodes resource is not namespaced in kubernetes either)
* kube-state-metrics (only one kube-state-metrics instance per cluster should exist)

@fabxc 

Closes #154

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -315,6 +315,9 @@ func (c *Operator) syncNodeEndpoints() {
 			Labels: map[string]string{
 				"k8s-app": "kubelet",
 			},
+			Annotations: map[string]string{
+				"alpha.monitoring.coreos.com/non-namespaced": "true",
+			},
 		},
 		Spec: v1.ServiceSpec{
 			Type:      v1.ServiceTypeClusterIP,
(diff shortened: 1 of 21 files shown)


fabxc:
Just wondering, setting something to "false" and it being "true" by default goes against most people's intuition when it comes to booleans.

Can you think of any straightforward name inverting the condition? 

"drop_namespace"?

brancz:
Maybe "non-namespaced"?

fabxc:
SGTM

brancz:
Adapted. PTAL @fabxc .

fabxc:
👍 

Wondering how to test the configuration as it gets more complex. Only useful way to do it would be an e2e test. Or does this not meet the threshold for adding one?

brancz:
An interesting one. As we are declaring this as an explicit feature I think an e2e test would actually be worth it.

I'm also playing with the idea of having one huge e2e (maybe acceptance?) test that sets up the entire monitoring pipeline (essentially the content of kube-prometheus) and checks that various scenarios work properly on a higher level as in "do all Grafana charts display something", "did Prometheus discover all Kubernetes components", etc. As that is basically the final check I perform myself to verify nothing broke.

However, what this means now for this PR is that I'll add an e2e test.

brancz:
Actually once I looked into the requirements for an e2e test I think it's not worth the effort as we are getting very close to the e2e/acceptance test I described.

The test would look something like this:

* Setup application exposing metrics with `namespace` label
* Create `Service` for the deployed application
* Create `ServiceMonitor` to capture the application
* Create `Prometheus` discovering the `ServiceMonitor`
* Ensure target does not contain `namespace` label
* Ensure query does not return metric with `exported_namespace` label

Instead I will add a unit test to ensure the config is generated correctly.

fabxc:
Okay, tested configs should at least be parsed once with Prometheus config package to ensure its validation passes.

The particular Prometheus package is not free of issues, which is why we just generate `map[string]interface{}`s rather than the types in that package.
Ideally, we'd get to prometheus/prometheus#2347 at some point – probably not before 2.0.

brancz:
@fabxc unit test added.

fabxc:
LGTM

Issue #98, opened 2017-01-13
Documentation of user specific findings/questions


galexrt:
A list for every user of `prometheus-operator` that misses something from the documentation that should be documented :wink: (If a finding/question is missing let me know and I'll add).

- [x] Preferred way to run exporters on the cluster for deployed applications (which don't have prometheus metrics support) (see #114)
- [x] Persistence for prometheus collected data (using Kubernetes PVs) (see #170)
- [x] Exposing prometheus instances to the outside/"public"
- [ ] How to secure/authenticate against exposed Prometheus/Alertmanager
- [x] Add Grafana to the Prometheus "stack" (see #115)
- [x] Getting started with Prometheus-Operator
- [ ] Meta Monitoring
- [x] High availability scheme (#117)
- [x] RBAC requirements (#153)
- [x] Network policies (#156)
- [ ] Upgrading cluster monitoring components (+ Effects on the running "system")
- [ ] Checking if prometheus-operator is running correctly
- [x] alerting

### Grafana Watcher
- [ ] What is the use case of it? (Taken from #245)
- [ ] Example usage? (Taken from #245)
- [ ] How grafana dashboard must be modified? (Taken from #245)

## User Guides
- [x] Cluster Monitoring
- [x] Application Monitoring
- [x] Example Exporter Configs/Manifests (this is [kube-prometheus](https://github.com/coreos/kube-prometheus), which is planned to be merged with this repository)
- [ ] Pushing (cronjob) metrics to Pushgateway
- [x] Use Kubernetes Ingress to expose Prometheus

brancz:
I opened #114 . It should cover:

* Preferred way to run exporters on the cluster for deployed applications (which don't have prometheus metrics support)
* node_exporters volume mounts for cronjob pushing 

Could you have a look at it @galexrt and tell me whether it makes sense? Thanks! 🙂

galexrt:
@brancz If you have some time, I'd like to hear some feedback from you if the new points on the list are partly not documented or should be handled in their separate own issues (most likely as feature requests).

brancz:
Sure thing!

> Persistence for prometheus collected data (using Kubernetes PVs)
> Exposing prometheus instances to the outside/"public"

These are definitely on my list, however, for the second one I'm unsure what you meant the difference to be compared to this one:

> How to persist Prometheus data

And I'm also unsure what you meant by this one:

> node_exporters volume mounts for cronjob pushing

Lastly 

[...]

PR #156, opened 2017-02-21
add network policies examples

gianrubio:


brancz:
Definitely makes sense to have these, thanks for you effort!

As a side note: @fabxc I feel we should work towards merging kube-prometheus and this repository as things like this are slightly unclear as to where they belong. As we discussed and agreed that we do want to merge kube-prometheus into this repository merging examples like these in here seems fine and in line with that. Once we do that we may want to restructure these files, but for the time being I think this is good to merge as is, wdyt @fabxc?

fabxc:
Yes, kube-prometheus is just a larger user guide in the end. Let's move the contents into `Documentation/user-guides/{cluster-monitoring,application-monitoring}` or similar.

Can we move those files into `Documentation/user-guides/network-policies/assets`? We should have at least a bit of text for this one as well. If you want to write something up @gianrubio that would certainly be great. If you don't have the time, we can add that ourselves later.

gianrubio:
Sure, I can write some documents, I'll try to do until next week.

👍 for merging kube-prometheus + prometheus-operator

brancz:
That would be extra dope, thanks!

brancz:
Mentioned this in #98 to track progress on user guides and docs.

brancz:
Ping @gianrubio. Are you still at this or would you like us to take over?

gianrubio:
Sorry @brancz I complete forgot this issue, I'll do the docs during this weekend.

gianrubio:
@brancz sorry for the delay. I wrote the doc explaining how to apply the network policies, I'm not a good english writer so If you have changes let me know.

Documentation/network-policies.md
```diff
@@ -0,0 +1,30 @@
+
+# Network policies
+
+[Network policies](https://kubernetes.io/docs/user-guide/networkpolicies/) allow you to restrict the ingress traffic between pods. 
+To keep your cluster safer, it's strongly recomended to enable network policies into prometheus namespace.
```
brancz: recomended -> recommended

Documentation/network-policies.md
```diff
@@ -0,0 +1,30 @@
...
+
+# Example
+
+In this example we are closing all the inbound communication on the namespace monitoring and just allowing the necessary traffic.
+
+
+Follow the steps [here](http://docs.projectcalico.org/v1.5/getting-started/kubernetes/installation/) to install calico, also dont' forget to [enable network policy](https:/kubernetes.io/docs/user-guide/networkpolicies) in you k8s cluster.
```
brancz: Are there other providers than calico for network policies? If so this would be the point to state that the solutions would need to be evaluated against each other and say that these examples were tested with calico.

Documentation/network-policies.md
```diff
@@ -0,0 +1,30 @@
...
+Once you've done that, you should annotate your namespace to deny all the ingress (inboud) traffic.
+
+``` 
+$ kubectl annotate ns monitoring "net.beta.kubernetes.io/network-policy={\"ingress\":{\"isolation\":\"DefaultDeny\"}}"
+namespace "monitoring" annotated
+```
+
+In this step you can't reach any port in your pod, so let's apply this network policies examples to allow the necessary traffic.
+
+```
```
brancz: As this will be rendered onto coreos.com at some point, we'll need to render the markdown files into this document. It would be enough if you just do a header for each of them like this:

```markdown
#### Alertmanager
[embedmd]:# (../example/networkpolicies/alertmanager.yaml)
...
```

Note what I used above:
```markdown
[embedmd]:# (../example/networkpolicies/alertmanager.yaml)
```

That is used to embed the yaml files into the document. Once you add those lines and run `make docs` they will automatically be up to date with what is in the files. (we make sure the documentation is up to date by checking with the `make docs` in our CI pipeline)

brancz:
Pretty good! Just a few comments, then it should be good :tada:

Documentation/network-policies.md
gianrubio: done

Documentation/network-policies.md
gianrubio: done

Documentation/network-policies.md
gianrubio: Cool feature, I liked it! ❤️

gianrubio:
@brancz please review again

brancz:
Awesome! Thanks a lot! lgtm :+1:

Issue #157, opened 2017-02-21
Prometheus service has no labels


galexrt:
The `prometheus-$INSTANCE_NAME` and `prometheus-operated` don't have a label. Meaning that there is no way to create a ServiceMonitor to use for scraping the Prometheus instance(s). The only way is, as @brancz, mentioned to create a separate Service.

The questions for is what label(s) should the Prometheus instance service(s) should have that "Meta Monitoring" is simply possible.

brancz:
I would suggest the `prometheus-operated` and `alertmanager-operated` named `Service`s to have the labels `app=prometheus` for prometheus and `app=alertmanager` for alertmanager. And an additional one to distinguish whether this is the `operated` `Service` which would be `operated=true`. Wdyt @fabxc ?

galexrt:
What about when single Prometheus instance deployments should be scraped?
Let's say I have my `prometheus-main` and `prometheus-testing` and I just want to scrape `prometheus-testing`. How about adding label `instance=testing` or something to the Service for that case?

brancz:
That's really just usage of the Prometheus Operator, as those services are not created by it, they're always created by the user - meaning the labeling is entirely up to you :slightly_smiling_face:. It wouldn't hurt though to label the `Service`s in `example/` respectively, just to be a role model.

galexrt:
Well, I completely oversaw that the services are created by the user (me) and not automatically by prometheus-operator. I create a PR adding labels to the example service manifests.

fabxc:
@brancz as mentioned on #98 I think there's some danger to adding `app=prometheus` to the governing service. People label their own services the same and then pull in every Prometheus pod twice unless they set a negative match on `operated=true`. Generally, they probably won't even know about that governing service.

brancz:
I see your concern now. Maybe we need to be more expressive then with just one label, maybe `operated-prometheus=true` ?

galexrt:
@brancz That sounds better. From my perspective this should cause less confusion/mismatching for users, exactly what @fabxc means.

PR #158, opened 2017-02-21
Add labels to example services

galexrt:
See #157.

***

Should an example ServiceMonitor for Prometheus be added too?

### DIFF
--- a/example/alertmanager-service.yaml
+++ b/example/alertmanager-service.yaml
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: Service
 metadata:
   name: alertmanager-main
+  labels:
+    app: alertmanager
+    alertmanager: main
 spec:
   type: NodePort
   ports:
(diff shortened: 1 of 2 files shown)


example/alertmanager-service.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: Service
 metadata:
   name: alertmanager-main
+  labels:
+    app: alertmanager
+    instance: main
```
fabxc: I think in our stateful sets etc. we went with `app: alertmanager` and `alertmanager: main`.
`instance` is only really meaningful in combination of both labels whereas the other one can serve as a meaningful selector in itself.

galexrt: The same for the Promtetheus? Like this `app: prometheus` and `prometheus: main`?

brancz: Yep

galexrt:
Renamed the label names.

brancz:
lgtm

PR #159, opened 2017-02-22
test: Setup jenkins ci

mxinden:
Automate prometheus-operator e2e tests via Jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,60 @@
+job('e2e-tests') {
+    concurrentBuild()
+
+    scm {
+        git {
+            remote {
+                github('coreos/prometheus-operator')
+                refspec('+refs/pull/*:refs/remotes/origin/pr/*')
+            }
+            branch('${sha1}')
+        }
+    }
+
+    wrappers {
+        credentialsBinding {
+            amazonWebServicesCredentialsBinding{
+                accessKeyVariable('AWS_ACCESS_KEY_ID')
+                secretKeyVariable('AWS_SECRET_ACCESS_KEY')
+                credentialsId('Jenkins-Monitoring-AWS-User')
+            }
+            usernamePassword('QUAY_ROBOT_USERNAME', 'QUAY_ROBOT_SECRET', 'quay_robot')
+        }
+    }
+
+    triggers {
+        githubPullRequest {
+            useGitHubHooks()
+            orgWhitelist(['coreos-inc'])
+        }
+    }
+
+    steps {
+        shell('docker build -t cluster-setup-env scripts/jenkins/.')
+    }
+
+    steps {
+        shell('docker run --rm -v /var/jenkins/workspace/e2e-playground:/var/jenkins/workspace/e2e-playground -v /var/run/docker.sock:/var/run/docker.sock cluster-setup-env /bin/bash -c "cd /var/jenkins/workspace/e2e-playground && make crossbuild"')
+    }
+
+    steps {
+        shell('docker build -t quay.io/coreos/prometheus-operator-dev:$BUILD_ID .')
+        shell('docker login -u="$QUAY_ROBOT_USERNAME" -p="$QUAY_ROBOT_SECRET" quay.io')
+        shell('docker push quay.io/coreos/prometheus-operator-dev:$BUILD_ID')
+    }
+
+    steps {
+        shell('docker run --rm -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -v $PWD:/go/src/github.com/coreos/prometheus-operator cluster-setup-env /bin/bash -c "cd /go/src/github.com/coreos/prometheus-operator/scripts/jenkins && REPO=quay.io/coreos/prometheus-operator-dev TAG=$BUILD_ID make"')
+    }
+
+    publishers {
+        postBuildScripts {
+            steps {
+                shell('docker run --rm -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -v $PWD:/go/src/github.com/coreos/prometheus-operator cluster-setup-env /bin/bash -c "cd /go/src/github.com/coreos/prometheus-operator/scripts/jenkins && make clean"')
+                shell('docker rmi quay.io/coreos/prometheus-operator-dev:$BUILD_ID')
+            }
+            onlyIfBuildSucceeds(false)
+            onlyIfBuildFails(false)
+        }
+    }
+}
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ NAMESPACE?=prometheus-operator-e2e-tests-$(shell LC_CTYPE=C tr -dc a-z0-9 < /dev
 PROMU := $(GOPATH)/bin/promu
 PREFIX ?= $(shell pwd)
 
-CLUSTER_IP?=$(shell minikube ip)
+CLUSTER_IP?=$(shell kubectl config view --minify | grep server: | cut -f 3 -d ":" | tr -d "//")
 
 pkgs = $(shell go list ./... | grep -v /vendor/ | grep -v /test/)
 
@@ -30,7 +30,7 @@ container:
 	docker build -t $(REPO):$(TAG) .
 
 e2e-test:
-	go test -timeout 20m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig "$(HOME)/.kube/config" --operator-image=quay.io/coreos/prometheus-operator:$(TAG) --namespace=$(NAMESPACE) --cluster-ip=$(CLUSTER_IP)
+	go test -timeout 20m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig "$(HOME)/.kube/config" --operator-image=$(REPO):$(TAG) --namespace=$(NAMESPACE) --cluster-ip=$(CLUSTER_IP)
 
 e2e-status:
 	kubectl get prometheus,alertmanager,servicemonitor,statefulsets,deploy,svc,endpoints,pods,cm,replicationcontrollers --all-namespaces
--- a/scripts/jenkins/Dockerfile
+++ b/scripts/jenkins/Dockerfile
@@ -1,0 +1,38 @@
+FROM golang:1.7.5-wheezy
+
+ENV TERRAFORM_VERSION 0.8.7
+ENV KOPS_VERSION 1.5.1
+ENV DOCKER_VERSION 1.13.1
+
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ NAMESPACE?=prometheus-operator-e2e-tests-$(shell LC_CTYPE=C tr -dc a-z0-9 < /dev
 PROMU := $(GOPATH)/bin/promu
 PREFIX ?= $(shell pwd)
 
-CLUSTER_IP?=$(shell minikube ip)
+CLUSTER_IP?=$(shell kubectl config view --minify | grep server: | cut -f 3 -d ":" | tr -d "//")
```
brancz: how can we make sure that we locally don't accidentally run the tests against a production cluster as that happens to be the context that is currently selected?

scripts/jenkins/ed.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
brancz: maybe explain why this script is required, so we in the future or someone new can understand this hack

scripts/jenkins/wait-for-cluster.sh
```diff
@@ -0,0 +1,4 @@
+until kubectl cluster-info
```
brancz: missing she-bang and it's always good to have a quick description of what this script does ... generally following https://google.github.io/styleguide/shell.xml

Jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+                credentialsId('Jenkins-Monitoring-AWS-User')
+            }
+            usernamePassword('QUAY_ROBOT_USERNAME', 'QUAY_ROBOT_SECRET', 'quay_robot')
+        }
+    }
+
+    triggers {
+        githubPullRequest {
+           useGitHubHooks()
+            admins(['mxinden'])
```
brancz: this looks like it's a leftover from testing, is this still required?

Jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+            orgWhitelist(['coreos-inc'])
+        }
+    }
+
+    steps {
+        shell('docker build -t cluster-setup-env scripts/jenkins/.')
+    }
+
+    steps {
+        shell('docker run --privileged --rm -v $PWD:/go/src/github.com/coreos/prometheus-operator/ cluster-setup-env /bin/bash -c "cd /go/src/github.com/coreos/prometheus-operator && CGO_ENABLED=0 GOOS=linux go build --ldflags \'-extldflags \\"-static\\"\' github.com/coreos/prometheus-operator/cmd/operator"')
```
brancz: we already discussed this offline and makes sense to only address this in a follow up PR, but we should work towards replacing this with `promu`

scripts/jenkins/clusters/prom-test-indenml/spec.yaml
```diff
@@ -0,0 +1,9 @@
+awsRegion: eu-west-1
+prometheus:
+  machineType: d2.medium
+  scrapeInterval: 4s
+  instances:
```
brancz: Is this intended to be blank?

scripts/jenkins/clusters/prom-test-indenml/spec.yaml
```diff
@@ -0,0 +1,9 @@
+awsRegion: eu-west-1
```
brancz: Is this yaml still required? If yes the folder should probably be renamed.

scripts/jenkins/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM golang:1.7.5-wheezy
+
+ENV TERRAFORM_VERSION 0.8.4
```
brancz: Hoping this does not make a difference but `0.8.7` is latest.

scripts/jenkins/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip" && \
+    unzip awscli-bundle.zip && \
+    ./awscli-bundle/install -i /usr/local/aws -b /bin/aws
+
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && \
+    chmod +x ./kubectl && \
+    mv ./kubectl /bin/kubectl
+
+RUN curl -L https://get.docker.com/builds/Linux/x86_64/docker-1.6.0 > /usr/local/bin/docker && \
```
brancz: Why this version? Also should be moved to the env vars above.

scripts/jenkins/wait-for-cluster.sh
brancz: You'll also want a `set -e` here to make the script exit with a non 0 exit code if say `kubectl cluster-info` errors.

scripts/jenkins/ed.sh
brancz: Additionally we probably want to validate the input at least a bit, to protect against the Makefile accidentally writing somewhere crazy here and to make sure the source actually exists.

brancz: And if it's more than just `cat $1 > $2` we probably want to name the variables to express their content.

Makefile
mxinden: Resolved for now

brancz: For the reference: may want to run locally against other clusters than minikube therefore a simple "if CI" won't work, so skipping this for now.

mxinden:
test this please

scripts/jenkins/ed.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Kops requires user input through an editor. This shell script simply returns
+# the file input '$1' to STDOUT which enables us to script editing nodes through
```
brancz: I'm not sure this is 100% correct. Maybe @alexsomesan can comment. IIRC how this hack works is that kops calls `$EDITOR current-state-of-file after-edit-done-file`. And as we can pass the first one directly to kops and we are sure this is what we want as it has been checked in to git we can just write it to the "destination" file.

Either way naming $1 and $2 would definitely help.

brancz: Looking at how this is used it's actually that we produce a fake editor for every "edit" with kops that always just writes the same file (`$1`) to the "to-edit" file (`$2`), which is probably some tmp file kops controls.

mxinden:
test this please

scripts/jenkins/wait-for-cluster.sh
mxinden: resolved offline.

mxinden:
test this please

scripts/jenkins/wait-for-cluster.sh
brancz: Always good to documented anyways what the result of the offline discussion was for everyone else to understand and for us to understand when we go back.

For the reference: there was only the `until` statement and the `sleep` function so there was nothing to error.

However, as this script is now more than that, I propose that `set -e` is added. Plus, even if we don't see a possible error condition doesn't mean there isn't one :wink:

Jenkinsfile
```diff
@@ -0,0 +1,60 @@
...
+        shell('docker build -t cluster-setup-env scripts/jenkins/.')
+    }
+
+    steps {
+        shell('docker run --rm -v /var/jenkins/workspace/e2e-playground:/var/jenkins/workspace/e2e-playground -v /var/run/docker.sock:/var/run/docker.sock cluster-setup-env /bin/bash -c "cd /var/jenkins/workspace/e2e-playground && make crossbuild"')
+    }
+
+    steps {
+        shell('docker build -t quay.io/coreos/prometheus-operator-dev:$BUILD_ID .')
+        shell('docker login -u="$QUAY_ROBOT_USERNAME" -p="$QUAY_ROBOT_SECRET" quay.io')
```
glerchundi: as this is going to be expanded, the credentials will be leaked in the build log, right?

fabxc: Very good point, thanks for bringing that up! @mxinden can you check this?

mxinden: @glerchundi @fabxc As we are using the Jenkins credentials plugin it automatically masks them in the log output: `docker login -u=**** -p=**** quay.io`. Thank you for paying attention though! Good feedback.

mxinden:
test this please

mxinden:
@brancz Could you re-review this PR?

Jenkinsfile
```diff
@@ -0,0 +1,60 @@
...
+                secretKeyVariable('AWS_SECRET_ACCESS_KEY')
+                credentialsId('Jenkins-Monitoring-AWS-User')
+            }
+            usernamePassword('QUAY_ROBOT_USERNAME', 'QUAY_ROBOT_SECRET', 'quay_robot')
+        }
+    }
+
+    triggers {
+        githubPullRequest {
+           useGitHubHooks()
```
brancz: everything else is indented with 4 spaces, any reason this is not?

scripts/jenkins/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+ENV TERRAFORM_VERSION 0.8.7
+ENV KOPS_VERSION 1.5.1
+ENV DOCKER_VERSION 1.13.1
+
+RUN echo "deb http://ftp.debian.org/debian wheezy-backports main" >> /etc/apt/sources.list
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+		ca-certificates \
+		wget \
```
brancz: tabs -> spaces

Jenkinsfile
brancz: phew, quite a lot of trust we're putting in that plugin, but I guess this is ok

Jenkinsfile
```diff
@@ -0,0 +1,60 @@
...
+                credentialsId('Jenkins-Monitoring-AWS-User')
+            }
+            usernamePassword('QUAY_ROBOT_USERNAME', 'QUAY_ROBOT_SECRET', 'quay_robot')
+        }
+    }
+
+    triggers {
+        githubPullRequest {
+            useGitHubHooks()
+                orgWhitelist(['coreos-inc'])
```
brancz: is this indentation correct? it looks like two function calls one after another to me, so they should be aligned, no?

brancz:
lgtm :+1: nice job!

Jenkinsfile
glerchundi: thanks to you guys for such a good project 👍

PR #160, opened 2017-02-22
Governing service labeling

### LINKED ISSUE #157: Prometheus service has no labels

brancz:
This PR adds unique labeling to the governing services of the Alertmanager and Prometheus. Additionally it instead of just trying to create the service reconciles them, in case an old `Service` exists and is thus updated, which may lead to using informers for this eventually but right now by requesting the API. Finally this PR vendors and introduces `pkg/errors`, and implements it on all occurrences that were touched.

In addition with #158 this closes #157.

@fabxc 

/cc @galexrt

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -25,6 +25,7 @@ import (
 	"github.com/coreos/prometheus-operator/pkg/prometheus"
 
 	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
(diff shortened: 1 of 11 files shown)


fabxc:
:+1:

PR #161, opened 2017-02-22
rbac: add get action for services and endpoints

brancz:
As Service and Endpoints objects are now reconciled, the current object
needs to be retrieved from the API, therefore the get action must also
be allowed.

@fabxc


fabxc:
👍

PR #167, opened 2017-02-24
*: convert replicas field to pointer

### LINKED ISSUE #163: default replicas to 1 / pointer replicas
While we default to 1 replica through the `StatefulSet`, this is not the case when reconciling. This is the reason why upstream controllers/resources have the replicas as a pointer, to be able to distinguish from set/unset to default to 1 on unset, but allow to scale to 0.

Fortunately this does not mean a breaking change for users, as parsing manifests will work the same way, but it does require a change in the struct itself. This applies to the `Prometheus` and `Alertmanger` TPR.

@fabxc

brancz:
Closes #163 

@fabxc


fabxc:
👍

PR #168, opened 2017-02-24
test: Fix flaky TestPrometheusReloadConfig

mxinden:
Instead of listening to logs for reload we change the prometheus.yml and
check if configuration changes.

### DIFF
--- a/test/e2e/framework/prometheus.go
+++ b/test/e2e/framework/prometheus.go
@@ -15,8 +15,10 @@
 package framework
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


test/e2e/framework/prometheus.go
```diff
@@ -168,3 +170,53 @@ func (f *Framework) DeletePrometheusAndWaitUntilGone(name string) error {
...
+			return false, err
+		}
+
+		if len(targets) == amount {
+			return true, nil
+		}
+
+		return false, nil
+	}); err != nil {
+		return fmt.Errorf("Waiting for targets timed out. %v of %v targets found. %v", len(targets), amount, err)
```
brancz: > Error strings should not start with a capital letter because they'll often be prefixed before printing

https://github.com/golang/go/wiki/Errors#errors

test/e2e/prometheus_e2e_test.go
```diff
@@ -95,49 +99,63 @@ func TestPrometheusVersionMigration(t *testing.T) {
...
+		},
+	}
 
 	cfg := &v1.ConfigMap{
 		ObjectMeta: apimetav1.ObjectMeta{
 			Name: fmt.Sprintf("prometheus-%s", name),
 		},
 		Data: map[string]string{
-			"prometheus.yaml": "",
+			"prometheus.yaml": "global:\n  scrape_interval: 1m\nscrape_configs:\n  - job_name: testReloadConfig\n    metrics_path: /metrics\n    static_configs:\n      - targets:\n        - 111.111.111.111:9090",
```
brancz: make this a multi-line string rather than having "\n" as line breaks

test/e2e/prometheus_e2e_test.go
```diff
@@ -95,49 +99,63 @@ func TestPrometheusVersionMigration(t *testing.T) {
...
-			return false, err
-		}
+	if err := framework.WaitForTargets(1); err != nil {
+		t.Fatal(err)
+	}
 
-		if strings.Contains(logs, "config map updated") && strings.Contains(logs, "successfully triggered reload") {
-			return true, nil
-		}
+	cfg.Data["prometheus.yaml"] = "global:\n  scrape_interval: 1m\nscrape_configs:\n  - job_name: testReloadConfig\n    metrics_path: /metrics\n    static_configs:\n      - targets:\n        - 111.111.111.111:9090\n        - 111.111.111.112:9090"
```
brancz: make this a multi-line string rather than having "\n" as line breaks

test/e2e/framework/prometheus.go
```diff
@@ -168,3 +170,53 @@ func (f *Framework) DeletePrometheusAndWaitUntilGone(name string) error {
...
+	}
+
+	return nil
+}
+
+func (f *Framework) GetActiveTargets() ([]*Target, error) {
+	resp, err := http.Get(fmt.Sprintf("http://%s:30900/api/v1/targets", f.ClusterIP))
```
brancz: this is ok for now but we can keep in the back of our heads that we should rather proxy these requests when possible rather than exposing on node port

mxinden:
@brancz Could you re-review.
The new multi line strings have to be indented by spaces. Otherwise Prometheus crashes.
http://www.yaml.org/faq.html

brancz:
Indeed, common pitfall, I've done the same mistake countless times!

lgtmog

PR #176, opened 2017-02-27
Documentation for running exporters on the cluster

galexrt:
This adds:
* ~~`kube-state-metrics` deployment, service and servicemonitor as an example~~
* ~~ServiceMonitor for the `default/kubernetes kube-apiserver` service~~
* ~~ServiceMonitor for the `kubelet` service (created by the prometheu-operator)~~
* Documentation for running exporters on the cluster

### DIFF
--- a/Documentation/user-guides/running-exporters.md
+++ b/Documentation/user-guides/running-exporters.md
@@ -1,0 +1,68 @@
+# Running Exporters
+Running exporters and scraping them with Prometheus configured by the prometheus-operator.
+
+## The goal of ServiceMonitor(s)
+The goal for one ServiceMonitor should be to cover a large number of `Service`s.
+This can be achieved by creating a generic `ServiceMonitor`.
+
+## Create a service that exposes Pod(s)
+For scraping an exporter or separate metrics port you need a service that targets the pod(s) of the exporter or application.
+An example for the `kube-state-metrics` is the below `Service` and a generic `ServiceMonitor` that covers more than just the `kube-state-metrics` `Service`.
+After you have created the `Service`, you then continue on by creating the `ServiceMonitor` for it.
+The order in which the `Service` and `ServiceMonitor` is created is not important.
+
+### `kube-state-metrics` Service example
+```
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: kube-state-metrics
+    k8s-app: kube-state-metrics
+  annotations:
+    alpha.monitoring.coreos.com/non-namespaced: "true"
+  name: kube-state-metrics
+spec:
+  ports:
+  - name: http-metrics
+    port: 8080
+    targetPort: metrics
+    protocol: TCP
+  selector:
+    app: kube-state-metrics
+```
+This Service targets all pods with the label `k8s-app: kube-state-metrics`.
+
+## Create a matching ServiceMonitor
+### Generic `ServiceMonitor` example
+```
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: ServiceMonitor
+metadata:
+  name: k8s-apps-http
+  labels:
+    k8s-apps: http
+spec:
+  jobLabel: k8s-app
+  selector:
+    matchExpressions:
+    - {key: k8s-app, operator: Exists}
+  namespaceSelector:
+    matchNames:
+    - kube-system
+    - monitoring
+  endpoints:
+  - port: http-metrics
+    interval: 15s
+```
+(A better example for monitoring Kubernetes cluster components can be found [User Guide "Cluster Monitoring"](user-guides/cluster-monitoring.md))
+This ServiceMonitor targets **all** Services with the label `k8s-app` (`spec.selector`) any value, in the namespaces `kube-system` and `monitoring` (`spec.namespaceSelector`).
+
+## Troubleshooting
+### Namespace "limits"/things to keep in mind
+See the ServiceMonitor Documentation:
+> While `ServiceMonitor`s must live in the same namespace as the `Prometheus`
+TPR, discovered targets may come from any namespace. This is important to allow
+cross-namespace monitoring use cases, e.g. for meta-monitoring. Using the
+`namespaceSelector` of the `ServiceMonitorSpec`, one can restrict the
+namespaces the `Endpoints` objects are allowed to be discovered from.


example/exporter/kube-state-metrics/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    prometheus.io/scrape: 'true'
```
brancz: this is not actually required

example/exporter/kube-state-metrics/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: kube-state-metrics
+  labels:
+    app: kube-state-metrics
```
brancz: kube-state-metrics is regarded as an add-on to kubernetes so it's fine to do `k8s-app: kube-state-metrics`

example/exporter/kube-state-metrics/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  ports:
+  - name: kube-state-metrics
```
brancz: This can actually be chosen to be reusable similar to [here](https://github.com/brancz/kube-prometheus/blob/f3973972d95abcc8cd5411dcd17cf5645f55460f/manifests/exporters/kube-state-metrics-svc.yaml) in combination with a `ServiceMonitor` like [this](https://github.com/coreos/kube-prometheus/pull/33/files#diff-88a7a3ebbd1605371aae973804dff0c4). Then we just need 2-3 `ServiceMonitor`s to monitor the entire Kubernetes stack.

example/exporter/kube-state-metrics/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    prometheus.io/scrape: 'true'
+  name: kube-state-metrics
+  labels:
```
brancz: I made this same mistake. I missed that `kube-state-metrics` exposes metrics with a `namespace` label, while the config that the Prometheus Operator generates also automatically adds the `namespace` label, which causes colliding `namespace` labels, which Prometheus itself handles somewhat gracefully but causes confusion because one of the labels get renamed to `exported_namespace`. The solution was a lately introduced feature to disable the Prometheus Operator from relabelling that `namespace` onto the target, using the an annotation like this:

```
alpha.monitoring.coreos.com/non-namespaced: "true"
```

I used it [here](https://github.com/coreos/kube-prometheus/pull/33/files#diff-11b5648b58a69c037faec93a547d6f17R8). I'm hoping to merge https://github.com/coreos/kube-prometheus/pull/33 soon, which will be soon after the `v0.6.0` release of the Prometheus Operator, which should be some time this week.

galexrt:
@brancz On it. I'm writting a doc for "Running exporters" (and scraping them) using prometheus-operator.

example/exporter/kube-state-metrics/deployment.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: kube-state-metrics
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: kube-state-metrics
```
brancz: this should then also be adapted to be `k8s-app: kube-state-metrics`

example/exporter/kube-state-metrics/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+    alpha.monitoring.coreos.com/non-namespaced: "true"
+  name: kube-state-metrics
+spec:
+  ports:
+  - name: http-metrics
+    port: 8080
+    targetPort: metrics
+    protocol: TCP
+  selector:
+    app: kube-state-metrics
```
brancz: this should then also be adapted to be `k8s-app: kube-state-metrics`

example/exporter/kube-state-metrics/service.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: kube-state-metrics
```
brancz: this label should be removed

example/exporter/kube-state-metrics/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+    k8s-app: kube-state-metrics
+  annotations:
+    alpha.monitoring.coreos.com/non-namespaced: "true"
+  name: kube-state-metrics
+spec:
+  ports:
+  - name: http-metrics
+    port: 8080
+    targetPort: metrics
```
brancz: The target port of the pod is unnamed, so this probably creates some confusion as it doesn't do anything. Probably best to just remove this line.

example/exporter/kube-state-metrics/servicemonitor.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "monitoring.coreos.com/v1alpha1"
+kind: "ServiceMonitor"
+metadata:
+  name: "kube-state-metrics"
+  labels:
+    app: kube-state-metrics
+spec:
+  selector:
+    matchLabels:
+      app: kube-state-metrics
```
brancz: this should then also be adapted to be `k8s-app: kube-state-metrics`

example/exporter/kube-state-metrics/servicemonitor.yaml
```diff
@@ -0,0 +1,13 @@
...
+  endpoints:
+  - port: "kube-state-metrics"
```
brancz: this needs to be adapted to `port: "http-metrics"`

example/exporter/kube-state-metrics/servicemonitor.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "monitoring.coreos.com/v1alpha1"
+kind: "ServiceMonitor"
+metadata:
+  name: "kube-state-metrics"
```
brancz: this can be generalized to be something like `k8s-apps-http`, so signal to the user reading this you don't actually need to create a `ServiceMonitor` per `Service`, but rather create a convention for `Service`s, so all `Service`s can be discovered by one `ServiceMonitor`. In reality, probably more than one `ServiceMonitor` will be necessary, but the idea is to keep the number of `ServiceMonitor`s very small compared to the number of `Service`s.

example/exporter/kube-state-metrics/servicemonitor.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    app: kube-state-metrics
```
brancz: going with the above comment, labels are used to uniquely identify an object or a number of objects, but as a single `ServiceMonitor` is intended to be used for many `Service`s this should more describe the class of `Service`s discovered. Maybe something in the direction of `k8s-apps: http`.

example/exporter/kube-apiserver-servicemonitor.yaml
```diff
@@ -0,0 +1,21 @@
...
+    matchNames:
+      - default
+  selector:
+    matchLabels:
+      component: apiserver
+      provider: kubernetes
+  endpoints:
+  - scheme: "https"
+    tlsConfig:
+      insecureSkipVerify: true
```
brancz: if you specify the `caFile` and `serverName` here then there is no need to skip the cert verification here

mxinden:
test this please

galexrt:
@brancz I've had some time now and added two documentation pages that cover:
* Running Exporters
* Monitoring Kubernetes

Documentation/monitoring-kubernetes.md
```diff
@@ -0,0 +1,27 @@
+# Monitoring Kubernetes
```
brancz: Maybe we should concentrate on how to run exporters in this PR as an elaborate guide for cluster monitoring is also here https://github.com/coreos/prometheus-operator/pull/188. I will make sure to include things from here in that cluster monitoring guide that are missing.

Documentation/running-exporters.md
```diff
@@ -0,0 +1,66 @@
+# Running Exporters
+Running exporters and scraping them with prometheus-operator.
```
brancz: As the Prometheus Operator doesn't actually scrape anything this sentence should be rewritten to something like: "Running exporters and configuring Prometheus to scrape them using means provided by the Prometheus Operator."

Documentation/running-exporters.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Two approaches to utilizing ServiceMonitors
```
brancz: As the documentation in the repository will end up on coreos.com I would leave this section out.

Documentation/running-exporters.md
```diff
@@ -0,0 +1,66 @@
...
+What approach is the "best" is currently mostly a "personal" preference.
+
+### One ServiceMonitor for multiple Services
+You create a ServiceMonitor to target a label that is shared among the Services you want to monitor.
+This approach enables you to use less ServiceMonitors, but may require as mentioned above a shared label among all services to be monitored.
+
+### One ServiceMonitor per one Service
```
brancz: Let's combine these two sections. And it's actually not recommended to have one `ServiceMonitor` per `Service`, only if there is no other way it is the last resort. One should always aim to have as many `Service`s underneath one `ServiceMonitor` as that will allow re-configuring to be a rare thing otherwise the chance of the Prometheus server to keep being re-configured is very high.

Documentation/running-exporters.md
```diff
@@ -0,0 +1,66 @@
...
+You create a ServiceMonitor to target only the labels of the **one** specific Service you want to target.
+This approach enables you to see the targeted pods separated in these groups, but requires more ServiceMonitors to be created.
+
+## Create a service that exposes Pod(s)
+For scraping an exporter or metrics "interface" you need a service that targets the pod(s) of the exporter or application.
+An example for the kube-state-metrics is the below Service.
+After you have created the Service, you then continue on by creating the prometheus-operator ServiceMonitor object.
```
brancz: Kubernetes object kinds should be in backward ticks, to highlight them.

Documentation/running-exporters.md
```diff
@@ -0,0 +1,66 @@
...
+### One ServiceMonitor for multiple Services
+You create a ServiceMonitor to target a label that is shared among the Services you want to monitor.
+This approach enables you to use less ServiceMonitors, but may require as mentioned above a shared label among all services to be monitored.
+
+### One ServiceMonitor per one Service
+You create a ServiceMonitor to target only the labels of the **one** specific Service you want to target.
+This approach enables you to see the targeted pods separated in these groups, but requires more ServiceMonitors to be created.
+
+## Create a service that exposes Pod(s)
+For scraping an exporter or metrics "interface" you need a service that targets the pod(s) of the exporter or application.
```
brancz: > `metrics "interface"`

I've never heard of them being referred as that. Do you have a reference? If you intended to describe it as that then maybe restructure the sentence a bit to not confuse people and they start searching on google for that.

Documentation/running-exporters.md
```diff
@@ -0,0 +1,66 @@
...
+      k8s-apps: http
+  endpoints:
+  - port: http-metrics
+    interval: 30s
+```
+This ServiceMonitor targets **all** Services with the label `k8s-app: http`.
+
+## Troubleshooting
+### Namespace "limits"/things to keep in mind
+See the ServiceMonitor Documentation:
```
brancz: Very nice pointer!

Documentation/running-exporters.md
galexrt: This section?
> # Running Exporters
> Running exporters and scraping them with Prometheus configured by the prometheus-operator.

Documentation/running-exporters.md
galexrt: Added backticks.

Documentation/running-exporters.md
brancz: No the beginning is fine, I was talking about this

```markdown
## Two approaches to utilizing ServiceMonitors
What approach is the "best" is currently mostly a "personal" preference.
```

We need to give users guidance, things like that make users uneasy. It's better to go one way and if necessary rethink the way we are showing.

brancz: Which is why I commented on the lower point of how to use `ServiceMonitor`s. It's better to tell users the recommended way and people are more likely to use it that way and once they fully understand what's happening and consciously decide for the other way it's fine.

example/exporter/kubelet-servicemonitor.yaml
```diff
@@ -0,0 +1,20 @@
+# You need to have "--kubelet-object=kube-system/kubelet" as a prometheus-operator arg
```
ese: If you have defined `--kubelet-object=kube-system/kubelet` you don't need this `ServiceMonitor` . isn't it?

brancz: The `ServiceMonitor` is still required as that feature only synchronizes the `InternalIP`s of your `Node`s into a `Service`/`Endpoints` object. How you organize the `ServiceMonitor`s is still up to you. More elaborate cluster monitoring guide coming in #188 .

galexrt: Should I change the comment to reflect that a `ServiceMonitor` is required too?

example/exporter/kube-state-metrics/deployment.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: kube-state-metrics
```
galexrt: Added.

example/exporter/kubelet-servicemonitor.yaml
brancz: That would be great @galexrt !

Documentation/running-exporters.md
```diff
@@ -0,0 +1,66 @@
+# Running Exporters
+Running exporters and scraping them with Prometheus configured by the prometheus-operator.
+
+## Two approaches to utilizing ServiceMonitors
+What approach is the "best" is currently mostly a "personal" preference.
```
brancz: Remove line 4 and 5 and combine the other sections into one. And describe, that the goal is always to have the least amount of `ServiceMonitor`s discovering a large number of `Service`s. In certain cases it won't be avoidable to have a 1-1 relationship, but it's not really a preference thing as every `ServiceMonitor` results in Prometheus actually reloading the configuration, which means sample ingestion stops for that time. So it's best to have the least amount of `ServiceMonitor`s, or at least a set that does not change (add, remove, modify) very often.

example/exporter/kubelet-servicemonitor.yaml
ese: @brancz thanks. I am wondering if would be better use a more generic definition that match all `k8s-app` with `https-metrics` endpoints

```yaml
kind: ServiceMonitor
metadata:
  name: k8s-apps-https
  labels:
    k8s-apps: https
spec:
  jobLabel: k8s-app
  selector:
    matchExpressions:
    - {key: k8s-app, operator: Exists}
  namespaceSelector:
    matchNames:
    - kube-system
  endpoints:
  - port: https-metrics
    interval: 15s
    scheme: https
    tlsConfig:
      caFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
      insecureSkipVerify: true
    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```

brancz: @ese I agree 100%! It's exactly what we do in [kube-prometheus](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus) (our cluster-monitoring manifest collection):

https://github.com/coreos/prometheus-operator/blob/882dab40a3c0c7727ac08b0635be1b51453c2097/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-servicemonitors.yaml

galexrt:
Note: I'm still working on this, but haven't gotten time for it yet. Will definitely work on this this weekend.

brancz:
No worries if you can't find time on it! You already put a lot of good work into this, if you want to finish it perfect, if not we're happy to continue it. :slightly_smiling_face:

example/exporter/kubelet-servicemonitor.yaml
galexrt: @brancz I'm removing the ServiceMonitor manifests as they are already added in the user guide "Cluster Monitoring".

galexrt:
I added your feedback. For now I removed the exporters and will add my exporters manifests in another PR (at least the `ceph-exporter`, `mysql-exporter` and `nginx-vts-exporter`).

Documentation/user-guides/running-exporters.md
```diff
@@ -0,0 +1,68 @@
+# Running Exporters
+Running exporters and scraping them with Prometheus configured by the prometheus-operator.
+
+## The goal of ServiceMonitor(s)
+The goal for one ServiceMonitor should be to cover a large number of `Service`s.
+This can be achieved by creating a generic `ServiceMonitor` that for example would .
```
brancz: Incomplete sentence, it would make sense if we just cut ~~that for example would~~.

galexrt: Removed incomplete part.

galexrt:
I added your last feedback and squashed everything together.

brancz:
Awesome! Nice job! Thanks for baring with me :slightly_smiling_face:

Issue #172, opened 2017-02-25
Deploying whole stacks with exporters: operator, operator+bash or helm?


whereisaaron:
I just finished [simple automated deployment/update, display, delete scripts](https://github.com/whereisaaron/kubernetes-prometheus-operator-deployment) for a stack consisting of:
- prometheus-operator with RBAC
- 'main' Prometheus with ConfigMap, RBAC, TLS Ingress, DNS CNAME, and Basic Auth
- 'main' Alertmanger with ConfigMap, RBAC (no access), TLS Ingress, DNS CNAME, and Basic Auth
- kube-state metrics Deployment with RBAC
- node-extractor DaemonSet with RBACS (no access)

I just did this with simple `bash` scripts, `kubectl`, `envsubst`, and `cli53`, based on many of the manifest examples here. I am deploying this to clusters from CI/CD and using it for alerts and as a datasource for Grafana.

I am decently happy with it, I see the main limitations currently being security:
- Prometheus and Alertmanager Pods have no TLS or auth within the cluster - could be fixed with an haproxy sidecar
- Alarmmonitor receivers have passwords in the plain in the ConfigMap - could be changed to use a Secret
- Metrics end-points for exporters are have no authentication - not sure if this can be easily fixed without sidecars on everything?

The stable ['prometheus' chart](https://github.com/kubernetes/charts/tree/master/stable/prometheus) for helm is not dissimilar from what prometheus-operator and my script is doing. Except prometheus-operator does more to simplify updating the Prometheus scraping configuration with ServiceMonitors.

Do you see prometheus-operator eventually doing all of the above for me (everything my script does), or do you see prometheus-operator maybe be part of a [helm chart](https://github.com/kubernetes/helm) or similar. E.g. where prometheus-operator focuses on active management of the configuration files, and the helm chart provides the supplemental resources like exporters and Services/Ingresses?

brancz:
Such an effort exists in [kube-prometheus](https://github.com/coreos/kube-prometheus) and there is work moving it to a helm chart by one of the helm chart maintainers: https://github.com/mgoodness/kube-prometheus-charts

As this is alpha software, so is the helm chart, everything is subject to change, therefore we will not be pushing for getting it into stable helm charts just yet.

whereisaaron:
[...]

PR #177, opened 2017-02-27
README: make the relationship with kube-prometheus clearer

brancz:
As mentioned in #172, we want to make more clear that kube-prometheus is not a contender, but rather makes use of the Prometheus Operator.

@fabxc 

/cc @whereisaaron


fabxc:
👍

PR #181, opened 2017-03-01
prometheus: use chunk encoding v2

### LINKED ISSUE #179: use prometheus v2 chunk encoding
We should be using the v2 chunk encoding in the Prometheus deployments, it uses more CPU but significantly less memory, which is the main concern for most deployments.

@fabxc

brancz:
Closes #179 

@fabxc


brancz:
test this please

fabxc:
👍

PR #182, opened 2017-03-01
test: Fix working dir spec and add string param for manual build

mxinden:
1. The current working dir was not properly specified when building the prometheus operator binary. It is now retrieved via the `$PWD` environment variable. 

2. The sha1 of the commit to be tested was previously only specified via the *GitHub Pull Request Builder* hook. This prevents the ability to trigger the job manually in Jenkins as Jenkins doesn't know which commit to test. This commit adds the sha1 string parameter to the `Jenkinsfil` job dsl for users to specify the sha1 when triggering jobs manually.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,10 @@ job('prometheus-operator-unit-tests') {
 job('prometheus-operator-e2e-tests') {
     concurrentBuild()
 
+    parameters {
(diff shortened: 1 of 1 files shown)


mxinden:
Retest this please

brancz:
lgtm :+1:

PR #183, opened 2017-03-01
contrib: migrate grafana-watcher from kube-prometheus

brancz:
As we want to migrate the kube-prometheus repository content into this repository the first step is to find a new home for the grafana-watcher. As discussed we'll put it in `contrib/` of this repository.

I looked at the history of the grafana-watcher and figured it was not worth transfering into this repository, so I dropped the history for it and just adapted the import paths to work with its new home.

@fabxc


fabxc:
👍

PR #188, opened 2017-03-03
Cluster monitoring

brancz:
Opening to get some preliminary feedback as this turned out to become a larger PR than I intended.

@fabxc 

Caveat: we will have to make sure to migrate the kube-prometheus repository into `contrib/kube-prometheus` before merging this, but the documentation is largely ready to be reviewed.

PR #192, opened 2017-03-06
migrating kube-prometheus

brancz:
Importing kube-prometheus and its history to unify the two repositories.

We will need to merge this before #188 as it depends on the content of this PR to be in `contrib/kube-prometheus`.

@fabxc


fabxc:
👍

PR #197, opened 2017-03-07
test: Check whether alertmanager mesh network initializes correctly

mxinden:
Spin up three alertmanagers via prometheus operator and check with the
/status API endpoint if they discover each other correctly.

❗️ ❗️ ❗️  Depends on alertmanager [PR 644](https://github.com/prometheus/alertmanager/pull/644)

### DIFF
--- a/test/e2e/alertmanager_e2e_test.go
+++ b/test/e2e/alertmanager_e2e_test.go
@@ -15,12 +15,12 @@
 package e2e
 
 import (
-	"encoding/json"
 	"fmt"
 	"github.com/coreos/prometheus-operator/pkg/client/monitoring/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/pkg/api/v1"
 	"net/http"
+	"strconv"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


brancz:
Great! There is actually a container image built from master for the Alertmanager with the `master` tag.

https://quay.io/repository/prometheus/alertmanager?tag=latest&tab=tags

Since the PR is merged and it was already built you can just set the test to use the `master` version, for the time being.

mxinden:
@brancz Updated PR to use alertmanager:master. Let's see what the CI says. Runs green locally.

Kept changes in separate commit so we can revert it as soon as meshStatus information is present in  alertmanager api v1.

test/e2e/alertmanager_e2e_test.go
```diff
@@ -188,3 +191,71 @@ func TestExposingAlertmanagerWithIngress(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if err := framework.CreateServiceAndWaitUntilReady(alertmanagerService); err != nil {
+		t.Fatal(err)
+	}
+
+	proxyGet := framework.KubeClient.CoreV1().Services(framework.Namespace.Name).ProxyGet
+	err := framework.Poll(time.Second*20, time.Second, func() (bool, error) {
+		request := proxyGet("", alertmanagerService.Name, "web", "/api/v1/status", make(map[string]string))
```
brancz: requesting a random instance load balanced through a service may not be the best thing to do here, I think we should rather test that each alertmanager instance has discovered all other peers, so instead of testing this for one instance we should test it for all

mxinden:
@brancz Could you re-review?

brancz:
Good stuff! :+1: lgtm

PR #199, opened 2017-03-08
Generate API docs

### LINKED ISSUE #110: Auto generate docs for API
As the code is always the source of truth and the types published as part of the API should be well commented, we should be able to auto generate at least the documentation for those objects.

Not a priority, but opening this for tracking.

@alexsomesan @fabxc

brancz:
Closes #110.

It has happened multiple times, that our docs were off, were not documented or differently documented in code and docs. So now we can generate the API docs from code comments.

Consolidated `prometheus.md`, `alertmanager.md` and `servicemonitor.md` into `design.md` and fixed references to them in the other docs.

@fabxc


Documentation/api.md
```diff
@@ -0,0 +1,193 @@
...
+| namespaceSelector | Selector to select which namespaces the Endpoints objects are discovered from. | `NamespaceSelector` | false |
+
+## StorageSpec
+
+StorageSpec defines the configured storage for a group Prometheus servers.
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| class | Name of the StorageClass to use when requesting storage provisioning. More info: https://kubernetes.io/docs/user-guide/persistent-volumes/#storageclasses | `string` | true |
+| selector | A label query over volumes to consider for binding. | `*metav1.LabelSelector` | true |
```
fabxc: I suppose there's no convenient way to have those link out to the main docs, right?

Makefile
```diff
@@ -53,7 +53,14 @@ embedmd:
 	GOARCH=$(subst x86_64,amd64,$(patsubst i%86,386,$(shell uname -m))) \
 	go get github.com/campoy/embedmd
 
-docs: embedmd
+apidocgen:
+	@GOOS=$(shell uname -s | tr A-Z a-z) \
+	GOARCH=$(subst x86_64,amd64,$(patsubst i%86,386,$(shell uname -m))) \
```
fabxc: Why do we need this?

Documentation/api.md
brancz: I have been thinking of this and most of the time we can actually get a reliable link, but I haven't implemented that, as I thought we could do that as a second step. Let me know if you want me to add that now.

Makefile
brancz: I reused this from the `promu` target in the Prometheus `Makefile`, not sure if it is actually required. In my tests it worked fine without, but I thought it was there for a reason in the Prometheus `Makefile`.

brancz:
Added links and trimmed the `Makefile`. PTAL @fabxc

fabxc:
nice 👍

PR #200, opened 2017-03-09
Default prom am version

mxinden:
### Refactor WaitForPods function

* Use framework internal Poll function instead of reimplementing it.
* Don't return pods list - only needed by one function call.
* Have one specification of how long to wait for a pod.
 	
### Introduce WaitForPodsRunImage

Instead of checking for pods running correct image in WaitForPodsReady
this functionallity is extracted into WaitForPodsRunImage function.  By
not requiring an image name when calling WaitForPodsReady we can remove
the image specification in MakeBasicPrometheus and
MakeBasicAlertmanager. Thereby basic tests are not fixed to specific am
and prom versions.


mxinden:
@brancz Could you review again?

brancz:
lgtm :+1:

PR #201, opened 2017-03-09
Use secrets for configuration files

### LINKED ISSUE #166: Alertmanager Configmap requires storing passwords in the plain on disk
With prometheus-operator, Alertmanager is configured with a file stored in the configmap. This means passwords and security tokens for alerters get stored in the plain and Kubernetes writes them to disk on any relevant nodes.

Normally I would map secrets like that from k8s secret resource into environment variables in the container. But it appears alertmanager requires secrets get stored in plain text in the config yaml file? If so, perhaps the whole config should be a secret instead? Then k8s will keep it off disk even when mounted in the pod. For this to work the config-reloader would need to be able to watch the secret instead.

(I think both secrets and configmaps have the same size limit of 1MB - imposed by etcd record size limit).

brancz:
Fixes #166 

This introduces using `Secret`s instead of `ConfigMap`s for the Prometheus and Alertmanager configuration files. I also adapted the Documentation referring to the old behavior.

While this is ready to be reviewed, we should hold off on merging as it would be confusing for someone browsing the documentation and reading about secrets when configmaps are actually still used with the currently latest release.

@fabxc 

@mxinden we should talk about a config reload test for the Alertmanager part.

/cc @whereisaaron


mxinden:
@brancz 
Testing: Sounds good. I will look into that.

fabxc:
👍

PR #205, opened 2017-03-09
Allow for multiple kube-state-metrics series

mikebryant:
This can happen if you run multiple replicas, or if you redeploy kube-state-metrics.
In either case, prometheus records multiple metrics with the instance ip in, and the dashboard fails.

This uses aggregation functions to get sensible output in either case

I've switched between max and min to try and be pessimistic where appropriate (e.g. max for generations and total replicas, min for available replicas etc.)
Shouldn't really matter given multiple copies should report the same data anyway

### DIFF
--- a/contrib/kube-prometheus/assets/grafana/deployment-dashboard.json
+++ b/contrib/kube-prometheus/assets/grafana/deployment-dashboard.json
@@ -302,7 +302,7 @@
           "targets": [
             {
               "refId": "A",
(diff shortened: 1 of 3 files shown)


brancz:
Great stuff! Do you think it's worth bumping the kube-state-metrics deployment to at least have two replicas then? As in HA by default.

mikebryant:
@brancz Yeh, sounds good, done :)

brancz:
Awesome! lgtm :+1:

PR #206, opened 2017-03-10
test: Add jenkins job building the docs to check if they changed

mxinden:
The prometheus operator api docs are generated automatically. To ensure
the docs are not out of sync with the code we generate the docs on every
pr and check if they changed.


brancz:
lgtmog

mxinden:
@brancz Rebased. Should be green in 30 minutes.

PR #207, opened 2017-03-10
Rebuild api documentation with nodeSelector

mxinden:
Docs can be automatically build with `make docs`. This PR keeps the current api doc in sync with current master code.

❗️  Please wait for [206](https://github.com/coreos/prometheus-operator/pull/206) to fail.


brancz:
Never thought I'd say this, but the test failed so lgtm :+1:

PR #208, opened 2017-03-10
Rule selection

### LINKED ISSUE #194: rule file selector
We will eventually hit the size limit of a single `ConfigMap` with alerting rules/recording rules therefore we need a mechanism to select multiple. I propose to instead of trying to mount a single `ConfigMap` to select `ConfigMap`s via a label selector.

This would be a breaking change to the current way we handle rule files.

Also cross referencing to #166, should we keep using `ConfigMap`s or use `Secret`s here as well. I would say Alertmanager and Prometheus configs should be `Secret`s but the rule files can remain to be `ConfigMap`s, but I'm happy to be convinced otherwise.

@fabxc 

/cc @whereisaaron

brancz:
Closes #194 

We should be ready to cut `v0.7.0` after these changes.

@fabxc


contrib/kube-prometheus/hack/cluster-monitoring/deploy
```diff
@@ -32,6 +32,12 @@ kctl apply -f manifests/prometheus/prometheus-k8s-service.yaml
 kctl apply -f manifests/alertmanager/alertmanager-config.yaml
 kctl apply -f manifests/alertmanager/alertmanager-service.yaml
 
+# unfortunately statefulsets cannot be changed except for their replica count
+# so we need to make sure that the rule files are created before we create the
+# prometheus resource so it can properly discover the rule files when creating
+# the statefulset
+sleep 5
```
fabxc: That's gonna be a real bummer for all deployments not backed by disk.

contrib/kube-prometheus/hack/scripts/generate-alertmanager-config-secret.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cat <<-EOF
+apiVersion: v1
+kind: Secret
+metadata:
+  name: alertmanager-main
+data:
+  alertmanager.yaml: $(cat assets/alertmanager/alertmanager.yaml | base64 --wrap=0)
```
fabxc: There's no kubectl command that can be ran against a file doing the same?

pkg/client/monitoring/v1alpha1/types.go
```diff
@@ -73,6 +73,8 @@ type PrometheusSpec struct {
 	RoutePrefix string `json:"routePrefix,omitempty"`
 	// Storage spec to specify how storage shall be used.
 	Storage *StorageSpec `json:"storage,omitempty"`
+	// A selector to select which ConfigMaps to mount for loading rule files from.
+	RuleSelector *metav1.LabelSelector `json:"ruleSelector,omitempty"`
```
fabxc: Should there be any default or a documented "best practice" around labeling those rule configmaps?

pkg/prometheus/statefulset.go
```diff
@@ -104,25 +107,51 @@ func makeStatefulSet(p v1alpha1.Prometheus, old *v1beta1.StatefulSet, config *Co
...
+}
+
 func makeEmptyConfig(name string) *v1.Secret {
 	return &v1.Secret{
 		ObjectMeta: apimetav1.ObjectMeta{
-			Name: configSecretName(name),
+			Name:   configSecretName(name),
+			Labels: managedByOperatorLabels,
+			Annotations: map[string]string{
+				"empty": "true",
```
fabxc: Is this used anywhere?

contrib/kube-prometheus/hack/cluster-monitoring/deploy
fabxc: Certainly okay for now. But wondering whether we could/should have a sidecar polling configmaps directly and writing them to disk instead of mounting them into pods. Would that get us into trouble with needing more RBACs?

pkg/prometheus/statefulset.go
brancz: This is meant for debugging purposes (hence an annotation) to see whether the Prometheus Operator overrode the config thinking it needs an empty one, or whether it is a generated one. I'm happy to remove it just think it may be useful to be able to debug a users report if we know what the Prometheus Operator was doing.

contrib/kube-prometheus/hack/scripts/generate-alertmanager-config-secret.sh
brancz: There is indeed an equivalent kubectl command, however, this is the [officially documented way](https://kubernetes.io/docs/user-guide/secrets/#creating-a-secret-manually) to do it manually and a `kubectl --dry-run` actually needs a running cluster, which would stop us from running this generate script in our CI pipeline as it's silly to setup a cluster even though we're not actually using it as we're just running `--dry-run` commands. Plus using `kubectl create secret generic --from-file` does not allow setting labels, which we are currently not doing for this secret but for the other `ConfigMap`s.

fabxc: Makes sense.

pkg/prometheus/statefulset.go
fabxc: Yea, that's probably fine to keep for that. I assume most people `kubectl replace` the config rather than editing them. So  the annotation should stick around accidentally too often,

contrib/kube-prometheus/hack/cluster-monitoring/deploy
brancz: Actually a sidecar won't be enough, as what happens is that the `StatefulSet` gets created without the rule file `ConfigMap`s mounted and then the Prometheus Operator wants to modify the `PodTemplate`, which is an invalid action on `StatefulSet`s (only replicas and containers may be changed).

What we could do: if `ruleSelector` is specified, we could not create the `StatefulSet` at all until `ConfigMap`s selected by the `ruleSelector` are found. That would not require any new RBAC requirements. If we agree on this I'm happy to add it to this PR.

pkg/prometheus/statefulset.go
brancz: I see you point, but the same would be true for the `managed-by: prometheus-operator` label.

contrib/kube-prometheus/hack/cluster-monitoring/deploy
fabxc: Mh, that's the point of the sidecar though. Removing the dynamic configmap mounts from the pod template and making it all a runtime internal of the pod.

Currently extending selected configmaps would literally mean deleting the entire stateful set first, which is pretty disruptive.

pkg/client/monitoring/v1alpha1/types.go
brancz: Definitely! I came up with:

```yaml
role: rulefiles
prometheus: main
```

If that sounds plausible then I'll add it to the documentation.

contrib/kube-prometheus/hack/cluster-monitoring/deploy
brancz: Makes sense now. Will develop the sidecar in a follow-up PR.

brancz:
Can you re-review @fabxc? I also cleaned up some of the dashboard stuff as we recently merged a new one and there was a problem with it.

fabxc:
👍

cemo:
@brancz when will it be released? 
Thanks

fabxc:
@cemo per my comment above, it would be better to have a sidecar pull ConfigMaps rather than adjusting the pod template every time, which cannot be done without major disruption. 
This will likely be added before a release happens. As a nice addition, this will mean that changes will become more instantaneous as we don't have to wait on the ConfigMap sync of the volume mount.

PR #209, opened 2017-03-10
Add resource requests dashboard

mikebryant:
This presents the resource requests vs the allocatable capacity in the cluster.

screenshot from 2017-03-10 20-05-56

### DIFF
--- a/contrib/kube-prometheus/assets/grafana/resource-requests-dashboard.json
+++ b/contrib/kube-prometheus/assets/grafana/resource-requests-dashboard.json
@@ -1,0 +1,424 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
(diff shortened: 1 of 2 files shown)


mikebryant:
Suggestions welcome on how this dashboard could be improved :)

brancz:
Awesome! Do you think it would make sense to add this to the charts in the all-nodes dashboard or would that be overloaded with the information already present?

On a side note, I'm not sure why the bot is going crazy, we'll look in to that (/cc @mxinden).

mikebryant:
I'm not sure. It is quite big already.
My biggest reason for making it separate is the all-nodes dashboard is focused on actual resource usage, it reflects what's really there.
This one is focused on the scheduling side of things, it's about the usage of the cluster being maxed out not because the resources have actually been used, but because they've been requested.

I don't have that strong an opinion, would you like me to move these new panels to the bottom of the all-nodes dashboard? I'm happy to do so

brancz:
Fair enough! I think it's valid to have a "scheduling" dashboard where we could add things like that pods were unable to be scheduled and things like that.

lgtm :+1:

PR #210, opened 2017-03-10
Add warning about Grafana HA

mikebryant:
We've just hit an issue where:
- if you have multiple grafana pods
- then you update the configmap to change dashboards
- then some of the grafana pods get restarted (but not all)

then the prometheus datasource ID is 0 on the new ones and N on the others, so URLs generated by one grafana instance don't work on the others.

We've fixed this by adding this to our Ingress:
```
  annotations:
    ingress.kubernetes.io/affinity: cookie
```


brancz:
Good point. :+1: Thanks! lgtm

PR #214, opened 2017-03-13
test: Check whether alertmanager reloads its config

mxinden:
This test first creates a config inside a kubernetes secret, checks if
alertmanager picks the config up correctly. Then it changes the config
and checks again if it is loaded by alertmanager correctly.

This is related to PR #201

### DIFF
--- a/test/e2e/alertmanager_e2e_test.go
+++ b/test/e2e/alertmanager_e2e_test.go
@@ -233,7 +233,6 @@ func TestMeshInitialization(t *testing.T) {
 
 func TestAlertmanagerReloadConfig(t *testing.T) {
 	alertmanager := framework.MakeBasicAlertmanager("reload-config", 1)
-	alertmanagerService := framework.MakeAlertmanagerService(alertmanager.Name, "alertmanager-service", v1.ServiceTypeClusterIP)
 
 	firstConfig := `
 global:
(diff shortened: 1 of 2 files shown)


test/e2e/framework/alertmanager.go
```diff
@@ -174,22 +174,46 @@ func amImage(version string) string {
...
+func (f *Framework) WaitForAlertmanagerResolveTimeoutConfig(amName string, resolveTimeout int) error {
+	return f.Poll(time.Minute*5, time.Second, func() (bool, error) {
+		config, err := f.GetAlertmanagerConfig("alertmanager-" + amName + "-0")
+		if err != nil {
 			return false, err
 		}
 
-		if len(amStatus.Data.MeshStatus.Peers) == amountPeers {
-			return true, nil
+		if config.Data.ConfigJSON.Global.ResolveTimeout != 6*60*1000000000 {
```
brancz: It's very hard to understand what's happening here. It would be better if we just compare the actually loaded [config](https://github.com/prometheus/alertmanager/blob/fced9e126b334fe0b9eb6733c4138638459a4bda/api/api.go#L171). We can just compare the loaded config string with the expected one, instead of some arbitrary value that's a subset of the config.

mxinden:
@brancz Could you review my newest changes?

test/e2e/alertmanager_e2e_test.go
```diff
@@ -230,3 +230,81 @@ func TestMeshInitialization(t *testing.T) {
 		}
 	}
 }
+
+func TestAlertmanagerReloadConfig(t *testing.T) {
+	alertmanager := framework.MakeBasicAlertmanager("reload-config", 1)
+	alertmanagerService := framework.MakeAlertmanagerService(alertmanager.Name, "alertmanager-service", v1.ServiceTypeClusterIP)
```
brancz: This `Service` is unused as we request the pod directly through the proxy, so we can remove it. :tada:

brancz:
I don't think the e2e test job failed because of a test, it seems like it's an issue with the test setup.

So lgtmog

mxinden:
Test this please

brancz:
Nice! :+1: lgtm

mxinden:
@brancz Wow, that was fast - thanks.

PR #216, opened 2017-03-15
network policy is a duplicate path

gianrubio:


brancz:
Oops missed that. Thanks for noticing! lgtm :+1:

PR #217, opened 2017-03-15
prometheus watcher sidecar

brancz:
This is a followup of #208 as it introduced a [timing issue](https://github.com/coreos/prometheus-operator/pull/208#discussion_r105612896) with the config map reloading. This PR solves that timing issue.

The "prometheus-watcher" sidecar is an extension of the previously used configmap reloader, which is still used for Alertmanager clusters. The sidecar watches a file called `configmaps.json` which contains a list of configmap keys and a hash of each configmap that belongs to that key. When the config volume is remounted, which can be the cause of either a config change or a rule configmap change then the sidecar retrieves all rule files from all selected configmaps and reloads prometheus.

(for the sake of the e2e tests I already pushed a preliminary version of the prometheus-watcher image)

@fabxc @mxinden


cmd/operator/main.go
```diff
@@ -48,7 +48,9 @@ func init() {
 	flagset.StringVar(&cfg.KubeletObject, "kubelet-object", "", "Endpoints object to write kubelets into in format \"namespace/name\"")
 	flagset.BoolVar(&cfg.TLSInsecure, "tls-insecure", false, "- NOT RECOMMENDED FOR PRODUCTION - Don't verify API server's CA certificate.")
 	flagset.BoolVar(&analyticsEnabled, "analytics", true, "Send analytical event (Cluster Created/Deleted etc.) to Google Analytics")
-	flagset.StringVar(&cfg.ConfigReloaderImage, "config-reloader-image", "jimmidyson/configmap-reload:latest", "Reload Image ")
+	flagset.StringVar(&cfg.PrometheusWatcherImage, "prometheus-watcher-image", "quay.io/coreos/prometheus-watcher:v0.0.1", "Config and rule reload image")
+	flagset.StringVar(&cfg.ConfigReloaderImage, "config-reloader-image", "jimmidyson/configmap-reload:latest", "Reload Image")
```
fabxc: We have a coreos hosted version of this now, no?

cmd/operator/main.go
```diff
@@ -48,7 +48,9 @@ func init() {
 	flagset.StringVar(&cfg.KubeletObject, "kubelet-object", "", "Endpoints object to write kubelets into in format \"namespace/name\"")
 	flagset.BoolVar(&cfg.TLSInsecure, "tls-insecure", false, "- NOT RECOMMENDED FOR PRODUCTION - Don't verify API server's CA certificate.")
 	flagset.BoolVar(&analyticsEnabled, "analytics", true, "Send analytical event (Cluster Created/Deleted etc.) to Google Analytics")
-	flagset.StringVar(&cfg.ConfigReloaderImage, "config-reloader-image", "jimmidyson/configmap-reload:latest", "Reload Image ")
+	flagset.StringVar(&cfg.PrometheusWatcherImage, "prometheus-watcher-image", "quay.io/coreos/prometheus-watcher:v0.0.1", "Config and rule reload image")
```
fabxc: Not sure the name is the best fit. Just from the name I'd think the Prometheus server is watched in some way. Maybe `prometheus-rules-watcher`? A bit long but utility components don't need a catchy name.

contrib/prometheus-watcher/examples/prometheus-watcher.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - '-config-volume-dir=/etc/prometheus/config'
+        - '-rule-volume-dir=/etc/prometheus/rules'
+        - '-reload-url=http://localhost:9090/-/reload'
+        resources:
+          requests:
+            memory: "16Mi"
+            cpu: "50m"
+          limits:
+            memory: "32Mi"
+            cpu: "100m"
```
fabxc: 0.1 of a core for a stand-by process seems a bit much.

contrib/prometheus-watcher/main.go
```diff
@@ -0,0 +1,291 @@
...
+
+	w.logger.Log("msg", "Starting...")
+	w.Refresh()
+	err = watcher.Add(w.cfg.configVolumeDir)
+	if err != nil {
+		w.logger.Log("msg", "Adding config volume to be watched failed.", "err", err)
+		os.Exit(1)
+	}
+
+	<-done
```
fabxc: The done channel is never closed anywhere so there doesn't seem to be actual termination handling.
In this case we can just move the for loop down out of the go routine and down here, right?

pkg/prometheus/statefulset.go
```diff
@@ -107,49 +109,71 @@ func makeStatefulSet(p v1alpha1.Prometheus, old *v1beta1.StatefulSet, config *Co
...
-			},
-		}
-		volumeMounts[i] = v1.VolumeMount{
-			Name:      volumeName,
-			ReadOnly:  true,
-			MountPath: ruleFilePath,
+	return s, nil
+}
+
+type ConfigMap struct {
```
fabxc: Probably better to call it `ConfigMapReference` – especially if it's exported.

fabxc:
Looks good in general.

cmd/operator/main.go
brancz: As it's not specific to rules but also reloads Prometheus on main config changes I would propose something rather in the direction of `prometheus-config-reloader`, wdyt?

cmd/operator/main.go
brancz: Yes, should we default to that one?

contrib/prometheus-watcher/examples/prometheus-watcher.yaml
brancz: True, I'll run it in the kube-prometheus environment and replace with something that I'll see from metrics.

pkg/prometheus/statefulset.go
brancz: Good point.

cmd/operator/main.go
fabxc: Yea, that works.

cmd/operator/main.go
fabxc: Yea, I think so.

contrib/prometheus-watcher/examples/prometheus-watcher.yaml
brancz: Actually this was just the example manifest, we're giving the container `5m` in reality

brancz:
test this please

brancz:
All comments applied. PTAL @fabxc .

fabxc:
👍

PR #219, opened 2017-03-17
cut 0.7.0

brancz:
Only a few lines in the changelog but a pretty large release overall.

I have intentionally left out mentioning kube-prometheus, grafana-watcher, and prometheus-config-reloader so far as I would like to figure out a versioning scheme for those, for the time being that means they're "released" with the main Prometheus Operator releases.

@fabxc @mxinden


fabxc:
👍

brancz:
@mxinden seems like the ingress tests are flaking, I noticed this a couple times now, we should look into that. Merging here as flaky test.

mxinden:
@brancz Thanks for the hint. I will look into the ingress tests.

PR #223, opened 2017-03-21
test: Fix flaky ingress tests

mxinden:
If the ingress controller is not yet, polling will result in a global
error. Before this would end the polling. Instead now we still keep on
polling to give the ingress controller more time to start.

Issue noticed in PR #219


mxinden:
test this please

brancz:
lgtm :+1:

PR #227, opened 2017-03-23
Removed the erroneous dash beside apiVersion

eugene-chow:
Previous version had a dash on the first line which prevent the script from being imported into k8s


brancz:
lgtm :+1: Thanks!

PR #229, opened 2017-03-23
kube-prometheus: add RBAC resources

brancz:
With Kubernetes 1.6 RBAC is becoming a default. Therefore we should make sure that everything necessary to run end to end monitoring is RBAC aware.

Should be good to merge if you can successfully deploy this on a 1.6.x cluster.

@fabxc @mxinden

PR #231, opened 2017-03-24
grafana-watcher: use official documented API

### LINKED ISSUE #224: Grafana roadmap
In reference to [#115](https://github.com/coreos/prometheus-operator/issues/115)
The current behavior of [grafana-watcher](https://github.com/coreos/prometheus-operator/blob/master/contrib/grafana-watcher) relies on a ConfigMap as the master dashboard repository. 
When you design a new dashboard in Grafana you need to manually export it and merge it into the ConfigMap manifest, which is quite tedious to do. Is there a plan to automate this process ?

In addition, while going through this process, I noticed that we were [using](https://github.com/coreos/prometheus-operator/blob/8dfc9d768ef9f7b31494e1dbb47e00dfcbfd6353/contrib/grafana-watcher/grafana/dashboard.go#L87) the Grafana `/api/dashboards/import` API, which seems to be undocumented in the[ Grafana API reference](http://docs.grafana.org/http_api/dashboard/). Why aren't we using the documented API?

brancz:
Fixes #224 

@fabxc @mxinden 

Will push the new version (`v0.0.1`) to quay once this we merge this change.

### DIFF
--- a/contrib/grafana-watcher/Makefile
+++ b/contrib/grafana-watcher/Makefile
@@ -3,18 +3,19 @@ all: build
 FLAGS =
 ENVVAR = GOOS=linux GOARCH=amd64 CGO_ENABLED=0
 REGISTRY = quay.io/coreos
-TAG = latest
+TAG = v0.0.1
+NAME = grafana-watcher
 
 build:
-	$(ENVVAR) go build -o grafana-watcher main.go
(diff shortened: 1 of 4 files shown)


brancz:
The failed docs job is fixed with #229

fabxc:
👍

PR #234, opened 2017-03-27
grafana-watcher: revert to import endpoint

brancz:
The dashboard create endpoint behaves slightly different when templating
is used. The import API works exactly the same as the front-end import
functionality.

@fabxc @mxinden 

/cc @nirdothan


brancz:
@mxinden we have an ingress failure, did we figure out what that was? Other than that all tests passed. Seem like a timeout issue.

mxinden:
@brancz I thought we fixed the flaky ingress test with 82638226528db6835516eb576d3838bc3cc1bc7f

I will look into it.

brancz:
@mxinden we should add some wrapping with `pkg/errors` so we know where the timeout error is actually coming from.

fabxc:
👍 on this change

PR #236, opened 2017-03-27
kube-prometheus: fix rolebinding

philips:
This entire thing is built in the monitoring namespace but the
rolebinding was on default. This caused the operator to never launch.


fabxc:
Thanks 👍

PR #237, opened 2017-03-27
kube-prometheus: put in kube-system

philips:
These manifests should live in kube-system; otherwise the label selector
won't work.


brancz:
So far this was solved by setting the namespace when calling kubectl, but as these manifests always have to be in kube-system namespace this makes sense. :+1: lgtm

brancz:
Can you just run `make docs`? It embeds the yaml files into our docs, we make sure our docs are up to date with the manifests this way.

jonboulle:
> BIN  contrib/.DS_Store

remove this

fabxc:
I'll remove it along with a later commit.

PR #240, opened 2017-03-27
*: generate RBAC aware bundle

### LINKED ISSUE #235: rbac: make rbac default
With Kubernetes v1.6 RBAC will be on by default. We should start telling people how to use that directly from the readme instead of just recommending `kubectl create -f deployment.yaml`.

brancz:
As Kubernetes 1.6 will be RBAC by default we should ship an RBAC aware bundle that is easily deployed. Tested with minikube 0.17.1, kubernetes 1.6 beta.3

@philips can you test whether this works in your environment?

@mxinden can we change the jenkins job to be a more general "generate" job?

@fabxc 

Closes #235


mxinden:
@brancz Can you be more specific on "generate" job? Sorry.

brancz:
@mxinden we currently have the "docs" job which just checks whether something changes within the repository when the `make docs` make target is executed. Can we change the job and the make target to `make generate` instead as we'll be doing more generation of content therefore we should check for more than the job.

brancz:
@mxinden we have another ingress failure, but now with more elaborate error message :slightly_smiling_face: maybe we can now figure out where we need to adapt/fix the timeout issue.

I just tested the result of this PR (the resulting `bundle.yaml` in the root dir) and it works.

Can you review @mxinden @fabxc ?

Documentation/rbac.md
```diff
@@ -8,7 +8,7 @@ In order for the Prometheus Operator to work in an RBAC based authorization envi
 
 Here is a ready to use manifest of a `ClusterRole` that can be used to start the Prometheus Operator:
 
-[embedmd]:# (../example/rbac/prometheus-operator-cluster-role.yaml)
+[embedmd]:# (../example/rbac/prometheus-operator/prometheus-operator-cluster-role.yaml)
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1alpha1
 kind: ClusterRole
```
brancz: done

Documentation/rbac.md
```diff
@@ -82,7 +82,7 @@ As Prometheus does not modify any Objects in the Kubernetes API, but just reads
```
brancz: done

brancz:
Re-review @mxinden ? :slightly_smiling_face:

PR #243, opened 2017-03-28
test: Replace own framework.Poll with client-go wait.Poll

mxinden:
@brancz As discussed replacing our *Poll* implementation with the client-go implementation.


brancz:
Code we don't have to maintain is my fav code! Nice! lgtm :+1:

mxinden:
@brancz Thanks for the fast review!

PR #248, opened 2017-03-29
Update prometheus-operator chart to v0.7.0

mgoodness:
Manually published to my personal charts repo until we have a CoreOS/Quay repo.


brancz:
lgtm, we should figure out a place for the "true" roles so we don't have to maintain them manually and keep them in sync and I'll talk to @ant31 about getting this onto quay apps

PR #254, opened 2017-03-31
Fix ingress repo

gianrubio:
Ingress resources was moved from kubernetes/contrib to kubernetes/ingress


mxinden:
@gianrubio Thanks for the fix!

PR #255, opened 2017-03-31
Support both alpha & beta RBAC API

### LINKED ISSUE #253: helm package templates hard coded rbac v1alpha1. Broken in k8s 1.6
Started a clean minikube with k8s version 1.6. kubectl version 1.6. helm version 2.2.0.
Installed tiller via helm init
Tried to install prometeus-operator with this command:
```helm install --name my-release opsgoodness/prometheus-operator```

Expected to see a successful installation of the operator, but got this error.
```
Error: release my-release failed: the API version in the data (rbac.authorization.k8s.io/v1alpha1) does not match the expected API version (rbac.authorization.k8s.io/v1beta1)
```
Running ```helm install --name my-release opsgoodness/prometheus-operator --set rbacEnable=false``` worked, but of course it didn't configure the rbac.

I saw in the templates that the version is hard coded. I think the best solution is to move the api version to a variable in the values.yaml file.

mgoodness:
Use `rbac.authorization.k8s.io/v1beta1` if the cluster supports it, otherwise fall back to `rbac.authorization.k8s.io/v1alpha1`.

Published as v0.7.1 and confirmed still working in K8s v1.5, but I don't yet have a K8s v1.6 cluster to test against. Recommend we wait for @peterrosell (or other) to deploy & confirm before merging.

### DIFF
--- a/helm/prometheus-operator/Chart.yaml
+++ b/helm/prometheus-operator/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 name: prometheus-operator
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.7.0
+version: 0.7.1
--- a/helm/prometheus-operator/Chart.yaml
+++ b/helm/prometheus-operator/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 name: prometheus-operator
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.7.1
+version: 0.0.1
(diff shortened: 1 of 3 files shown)


mgoodness:
Fixes #253

brancz:
Minikube supports 1.6, does it work on that?

brancz:
I'm wondering how we should handle the versioning of this. It should be along with the version of the Prometheus operator I feel, that would mean it shouldn't be v0.7.1 ... maybe it should be entirely independent then to not confuse people. Wdyt @fabxc @mgoodness @mxinden ?

mgoodness:
Totally forget about minikube. Tested & working!

mgoodness:
Re: versioning, current best practice is to de-couple the chart from the upstream application. Only requirement is that it follow SemVer. I've jumped around in my own repo without causing any complaints so far, but we should stabilize before we publish to an "official" CoreOS repo.

brancz:
Then I would like to start versioning with a preview release version like v0.0.1 for now and bump to v0.1.0 when we have everything in place we think we want for a start and then bump minor versions when we bump minor versions of the Prometheus Operator, makes sense?

fabxc:
SGTM

brancz:
generally lgtm

Issue #251, opened 2017-03-30
 Can't attach bearer token file to Prometheus pods


slavakl:
Endpoint specification allows to provide bearerTokenFile parameter which,I assume, can be used to pass a token different from a service account token.

As Prometheus nodes are created by Operator based on Prometheus resource, how a file with a token can be attached to Prometheus pods? Prometheus resource spec has nothing related to volume mount.

If this is not implemented yet, would adding a BearerToken parameter to Endpoint spec be an acceptable solution? Will be happy to submit a PR

galexrt:
Do you mean a `bearerToken` field for the "plain" token or a path to the `bearerTokenFile`?
See https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#endpoint

brancz:
Do you want to specify the bearer token for connecting to a kubernetes apiserver, or to be sent to your targets when scraping @slavakl ?

slavakl:
@galexrt I could, but I'm not sure how I can mount Secret to prometheus pod. It's not in Prometheus spec.

@brancz I want a token that will be send to my targets (my applications)

slavakl:
@brancz It seems that I misread prometheus configuration docs. token is only used to access Kubernetes API. If so, this issue can be closed.

brancz:
Actually both is possible in the Prometheus configuration. Currently we only support setting the bearer token file, which is intended as we don't want it to be specified in plain text in our resource, that would defeat the purpose.

There is some work on kubernetes/kubernetes which automatically mounts secrets in a pod based on the serviceaccount used. I think that would make the most sense here as well.

brancz:
This is actually related to #250, and we are aiming to fix this in the next release.

brancz:
As #264 is merged it is now possible to mount custom secrets for bearer tokens for example. Closing here.

slavakl:
@brancz, perfect! Thanks a lot!

PR #264, opened 2017-04-05
Mount custom secrets

### LINKED ISSUE #250: kube-prometheus: monitoring authenticated kubelets
**What did you do?**

Deployed kube-prometheus

kubelet being scraped

kubelet not being scraped

**Environment**

Tectonic v1.5.5 via Terraform AWS

The problem is that Tectonic has certificate protection on these endpoints; need to figure out how to configure this.

screen shot 2017-03-30 at 1 25 15 am

brancz:
This enables the ability to mount custom secrets into a Prometheus to be able to use those for the bearer token or TLS config in the Endpoint object of the ServiceMonitor.

@fabxc @mxinden 

Closes #250  and #251


fabxc:
Looks good in general. Still having somewhat of an itch about the documented path we provide... but all auto-resolve solutions I can think of right now seem prone to unexpected behavior too – at least if we assume that people may provide their own prometheus image with embedded secrets.

brancz:
So far I assumed the `baseImage` is purely there to be able to specify another location of but otherwise equivalent image repository to the canonical repository on quay.io. The use case being that the image has to come from a custom registry because of audit/security reasons.

brancz:
We would have the same concerns with the prometheus-config-reloader writing the rules to a specific location.

fabxc:
Yea, I think it is generally reasonable to assume people use secrets. But they could generally host their own image which is our image as a base + directly integrated certs. Those could have a random location basically – not sure whether this has to be supported though. For rules I think this is less likely.

If we assume people don't do anything like this, we could say that all cert and bearer token files should be simply specified as `<secretname>/<path_in_secret>`.

brancz:
As we've identified inconsistencies with where the serviceaccount secret is mounted and the inability to validate TPRs we will go with the way this is currently implemented.

brancz:
While mounting the secrets is now working I still have to figure out how to configure it properly so protected kubelets such as in #250 can be scraped.

brancz:
I validated, that this works for secure kubelets. Merging.

t-trout:
Could someone please provide an example of how to do this using Promtheus-Operator? I have been unsuccessful getting access to the ETCD nodes (the endpoints are exposed, but since it requires HTTPS, I need to use the etcd-client.pem and etcd-client-key.pem files generated by kube-AWS when the k8s cluster was generated).

I've tried specifying the tlsConfig for the ETCD service monitor with the caFile, certFile, and keyFile values, but since these are located on the underlying node and not in the Prometheus pod, it can't find them. I tried to create a volume with a hostPath to the location of these files within my Prometheus.yaml file, but since Prometheus is spun up by Prometheus-Operator, the volume isn't showing up (the volume does not show up when I look at the pods)

I'm thinking that maybe the best approach would be to create a secret with the 3 tlsConfig files (etcd-client.pem, etcd-client-key.pem, ca.pem) but an example showing how to mount the secret in the Prometheus.yaml file and then specifying the secret in my ETCD serviceMonitor would be so wonderfully helpful. Thanks -

PR #265, opened 2017-04-06
Documentation: Prometheus requires RBAC role for apiserver metrics

brancz:
When RBAC is enabled the Prometheus instance requires an RBAC role to be able to scrape the `/metrics` endpoint of the Kubernetes apiserver.

@fabxc @mxinden


fabxc:
👍

PR #267, opened 2017-04-06
prometheus: add pod and service label

### LINKED ISSUE #242: Customize relabel configs?
We would like some additional labels to come over from our targets. The prometheus-operator generated `relabel_configs` only bring over certain labels, is there any way to customize?

I couldn't find anything about it in the docs. If it's not currently a feature, is there any plan to add this?

Thanks!

brancz:
If available relabel the pod and service name onto pods. Using these labels can be joined onto them from metrics exposed by kube-state-metrics.

Closes #242

@fabxc @mxinden


fabxc:
👍

PR #270, opened 2017-04-07
cut 0.8.0

brancz:
@fabxc @mxinden 

@mgoodness as we're versioning the helm chart individually I have not upgraded the version there. Can you take care of that and bump the version of the chart?


fabxc:
👍

mgoodness:
Yup. I plan to work on the charts extensively this weekend. Will include the version bump.

PR #277, opened 2017-04-09
Added kube-prometheus chart with necessary for end-to-end exporters

goruha:
scripts: added kube-prometheus chart with basic exporters

Use this chart to install prometheus end-to-end on k8s with helm chart.
Exporters of metrics move to separated charts allow to disable some of them
Derivative metrics and alert rules added to right exporter chart
Here is example how to extend prometheus with alerts and derivative metrics https://github.com/cloudposse/charts/tree/master/incubator/prometheus-alerts


goruha:
@mgoodness could you look at this charts? 
Are they good enough to be accepted?
Please ping me or @osterman for any questions 

Thanks.

brancz:
Not a bad start, I think if @mgoodness is fine with this we can merge this.

One thing I want to mention though is, that we should try to stay as close as possible to kube-prometheus itself, so that we can ultimately generate the charts from there. I'm thinking of something like [this](https://github.com/heptio/kube.libsonnet), to add the additional labels/templating to the "base" manifests from kube-prometheus. That way kube-prometheus stays unopinionated and could also be packaged with other package managers if necessary.

goruha:
@brancz for that purpose I think we can just make helm chart as base and generate manifests from it.
Helm chart is a list of templates and it can generate result to stdout. We can forward that stdout to yaml files and generate simple yaml manifests. 
I'm not a fun of kube.libsonnet because this is just a template dsl. Does not add any value in compare with helm templates. But it will add dependency and have developers to learn jsonnet. 
Looks like overhead from my point of view.
@mgoodness what do you think?

brancz:
My concern is that if we build everything around helm then it is fundamentally impossible to generate let's say a kpm package out of it.

> [...] because this is just a template dsl. Does not add any value in compare with helm templates

I disagree with this. While jsonnet is a touring complete language, there is only so much you can do with helm templating.

For now having separate charts from the manifest collection is fine by me, but ultimately there should only be one source of truth for everything. I think the correct comparison here is that when I release software I don't create a debian package and expect everyone to create packages out of that, instead I release a binary in a release tarball, that everyone can build opinionated packages from.

goruha:
@brancz  Is there any other package managers except kpm you want to support?

brancz:
It's not about supporting a specific package manager, it's about having a set of unopinionated manifests.

brancz:
@mgoodness ping?

mgoodness:
Sorry about the ridiculously long delay. I'm trying to clear my slate at work so I can give this the attention it deserves. **Really** hoping to review this week. Sorry again.

brancz:
No worries, that happens!

unguiculus:
This looks great for a start. I'd suggest to make the exporters optional. They may not work in every environment. E. g., most of them won't work on GKE.

brancz:
@unguiculus certainly. It should be possible to opt-out of any component you don't want installed.

helm/kube-prometheus/charts/exporter-kube-api/templates/service.yaml
```diff
@@ -0,0 +1,21 @@
...
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+  name: {{ template "fullname" . }}
+  namespace: kube-system
+spec:
+  clusterIP: None
+  ports:
+    - name: https-metrics
+      port: 443
+      protocol: TCP
+      targetPort: 443
```
bradbeam: I think this should be exposed as a configurable option

goruha: @bradbeam hm... the case is that api by default exposed on 443. 
Do any one use other ports for api?

brancz: While this is a sensible default it can happen, that the apiserver is bound to a different port. For a first attempt at the helm charts this would be fine for me as it can be optimized in further PRs. @mgoodness' call here, I believe he said that he is preparing to release some charts some time soon.

bradbeam: I think that's fair to hold off, but an example is kargo. We default the kube apiserver port to 6443 so 443 will be available for people wanting to run ingress controllers on 443. 

https://github.com/kubernetes-incubator/kargo/blob/master/roles/kargo-defaults/defaults/main.yaml#L91

brancz: That sounds reasonable.

goruha:
@mgoodness Does "Verified" means approved?
Are we good to merge this PR ?

mgoodness:
@goruha no, "Verified" only means that I GPG-signed the commit when I rebased. I'm actively working (like, _right now_) to get this merged, though.

PR #280, opened 2017-04-10
Increase memory limits to avoid OOMKilled

gianrubio:
Related to https://github.com/kubernetes/kube-state-metrics/issues/112#issuecomment-292965833

### DIFF
--- a/contrib/kube-prometheus/manifests/exporters/kube-state-metrics-deployment.yaml
+++ b/contrib/kube-prometheus/manifests/exporters/kube-state-metrics-deployment.yaml
@@ -18,9 +18,9 @@ spec:
           containerPort: 8080
         resources:
           requests:
-            memory: 30Mi
+            memory: 100Mi
             cpu: 100m
           limits:
-            memory: 50Mi
+            memory: 200Mi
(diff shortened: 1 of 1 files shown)


fabxc:
If we get OOM killed at 50 the requested memory should probably be adjusted already to 100Mi and the limit to 200Mi.

gianrubio:
Should I do in this PR now?

fabxc:
Yea, that would be best.

gianrubio:
Done!

brancz:
Awesome thanks! We should definitely benchmark kube-state-metrics so we can do some vertical autoscaling as its memory consumption grows linearly with the cluster size.

brancz:
Just for the record, do you have an estimate count of objects you have in your Kubernetes cluster @gianrubio ?

gianrubio:
@brancz I'm not sure if this it you mean

``` 
# etcdctl ls --recursive /registry|wc -l
1060 
```

## v0.3.0
screen shot 2017-04-10 at 17 03 30

## v0.4.1 (looks lighter)

screen shot 2017-04-10 at 17 22 48

PR #281, opened 2017-04-10
Add network-policies for kube-state-metrics

gianrubio:


brancz:
Is there a need for a policy about kube-state-metrics accessing the Kubernetes API server? For example in a self-hosted environment the apiserver is a pod itself.

eugene-chow:
Can you elaborate? The NetworkPolicy is to allow `prometheus` to scrape from `kube-state-metrics`. Without this definition, the scraping won't happen at all. I've tested it in my environment.

brancz:
That makes sense. But `kube-state-metrics` needs to access the apiserver, so I'm wondering why it doesn't require a NetworkPolicy. I'm not super familiar with NetworkPolicies, so I might be missing something or it might be an explicit non-goal.

eugene-chow:
That's because `kube-apiserver` binds to the hostNetwork which isn't affected by NetworkPolicy. Recall [this conversation](https://github.com/coreos/prometheus-operator/issues/228)? :)

brancz:
Thanks for clarification @eugene-chow! I didn't make that connection, I kept thinking it uses Pod networking, but this way it all makes sense.

Then lgtm :+1:

PR #282, opened 2017-04-11
Add security recommendations for ingress

gianrubio:
This PR add grafana as ingress service. It also add some advice on how to protect this resources over internet.

### DIFF
--- a/Documentation/user-guides/exposing-prometheus-and-alertmanager.md
+++ b/Documentation/user-guides/exposing-prometheus-and-alertmanager.md
@@ -176,6 +176,8 @@ kubectl create -f https://raw.githubusercontent.com/kubernetes/ingress/master/ex
 
 > It is highly recommended to compare the available ingress controllers for a production environment. The nginx ingress controller may or may not be what is suitable for your production environment. Also have a look at HA Proxy, Træfɪk, GCE, AWS, and more.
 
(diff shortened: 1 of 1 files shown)


galexrt:
@gianrubio Do you know for what ingresses the `whitelist-source-range` annotation works for?

brancz:
I think it's very confusing to have a manifest mentioning grafana when the document is called "Exposing Prometheus and Alertmanager". I'm all for the security aspect of this PR, but I think we should leave out grafana in this document as it would be very confusing to users, because nothing else in the Documentation ever mentions grafana. We will have dedicated docs for grafana once we do implement the grafana TPR, until then I would like to keep everything Prometheus and Alertmanager scoped in the docs for the operator. Main reason being these docs are now published on coreos.com. https://coreos.com/operators/prometheus/docs/latest/

gianrubio:
> @gianrubio Do you know for what ingresses the whitelist-source-range annotation works for? (Maybe worth noting if it will only work for certain ingress providers)

@galexrt not yet, https://github.com/kubernetes/ingress/issues/566

@brancz got it, shall I keep only the security recommendation and remove the grafana expose?

brancz:
> shall I keep only the security recommendation and remove the grafana expose?

That sounds perfect. :+1:

gianrubio:
@brancz done!

brancz:
Awesome! Thanks a lot! :tada:

PR #285, opened 2017-04-13
Added kube-state-metrics.yaml

eugene-chow:


eugene-chow:
Fyi, I copied the NetworkPolicy scripts from the [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) project and change its namespace to `monitoring` so that it works with prometheus-operator ootb.

brancz:
The RBAC manifests already exist in https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus/manifests/exporters and there is an open PR (#281), where I also commented about the same question I have now: 

> Is there a need for a policy about kube-state-metrics accessing the Kubernetes API server? For example in a self-hosted environment the apiserver is a pod itself.

eugene-chow:
Ah! The original copy I used was a month old. I have nothing to add in that case.

brancz:
Thanks again for clearing things up and your contributions, very helpful!

PR #286, opened 2017-04-13
prometheus: add insecure port to kubelet endpoints object

brancz:
As the insecure port is going to be necessary for the time being regarding metrics, we should have it on the Endpoints object we synchronize the kubelets into.

@fabxc @mxinden


mxinden:
Test this please

brancz:
Test this please

PR #289, opened 2017-04-15
grafana-watcher login fixes

sjuxax:
There is a potential race condition when the grafana manifests are applied where the credentials may not be applied prior to the deployment, which depends on them. I believe this was the cause of some issues I had with `grafana-watcher`.

There is now a helpful error message when `grafana-watcher` cannot log in, and `grafana-credentials.yaml` are applied prior to other grafana manifests to prevent weird login conditions like the one I experienced.


fabxc:
Thanks! LGTM.

PR #295, opened 2017-04-18
Use new Prometheus 1.6 storage flag

fabxc:
@mxinden

Prometheus 1.6 should be significantly better at staying within certain memory bounds. Make use of it by using the new flag.
I did not flip the default version yet as I ran into https://github.com/prometheus/prometheus/issues/2635 along the way and we should wait for v1.6.1.


fabxc:
Yes, everything is passing.

PR #299, opened 2017-04-20
Bump default Prometheus version to 1.6.1

fabxc:
@mxinden


sdurrheimer:
Personally I would remove all `version` in examples and let the prometheus-operator use his default version.

fabxc:
I thought about that too. OTOH, I think that one should generally be explicit about the version and rather should set organisation-wide defaults in config management and other tooling.

I'd keep it in mind for the future but not make a change yet.

fabxc:
Version doesn't really matter for the migration we have now. Changed it regardless.

PR #301, opened 2017-04-20
Add useful alerts

gianrubio:
- Add pods alerts for status waiting or timeout
- Alerts for docker hang/timeout

Desc: I'm using this alarm and was quite useful, specially when new deploys comes and the pods are in restart/waiting status.

# How to test?

`$ kubectl run fake-pod-test --image nginx --command fail-me
`

ps. Please share any thoughts on this alerts

### DIFF
--- a/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
+++ b/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
@@ -516,4 +516,52 @@ data:
         summary = "Kubelet is close to pod limit",
         description = "Kubelet {{$labels.instance}} is running {{$value}} pods, close to the limit of 110",
       }
-    
+
+    ALERT K8SPodWaiting
+      IF sum(sum_over_time(kube_pod_container_status_waiting[5m]) > 0 ) BY (job, exported_pod, instance, namespace)
+      FOR 5m
+      LABELS {
+        service = "k8s",
+        severity = "critical",
+      }
+      ANNOTATIONS {
+        summary = "Pod is waiting too long",
+        description = "Pod {{ $labels.namespace }}/{{ $labels.exported_pod }} is in waiting state",
+      }
+
+    ALERT K8sPodRestartingTooMuch
+      IF sum (rate(kube_pod_container_status_restarts[5m]) > 0) BY(job, exported_pod, instance, namespace)
+      FOR 5m
+      LABELS {
+        service = "k8s",
+        severity = "critical",
+      }
+      ANNOTATIONS {
+        summary = "Pod is in a restart loop",
+        description = "Pod {{ $labels.namespace }}/{{ $labels.exported_pod }} is restarting constantly",
+      }
+
+  docker.rules: |+
+    ALERT DockerHasErrors
+      IF irate(kubelet_docker_operations_errors[10m])  > 0
+      FOR 1m
+      LABELS {
+        service = "docker",
+        severity = "critical",
+      }
+      ANNOTATIONS {
+        summary = "Docker has errors!",
+        description = "Docker on instance {{ $labels.instance }} has some errors",
+      }
+
+    ALERT DockerHasTimeout
+      IF irate(kubelet_docker_operations_timeout[5m]) > 0
+      FOR 1m
+      LABELS {
+        service = "docker",
+        severity = "critical",
+      }
+      ANNOTATIONS {
+        summary = "Docker timeout!",
+        description = "Docker on {{ $labels.instance }} has timeouts",
+      }


contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -516,4 +516,52 @@ data:
...
+        severity = "critical",
+      }
+      ANNOTATIONS {
+        summary = "Pod is waiting too long",
+        description = "Pod {{ $labels.namespace }}/{{ $labels.exported_pod }} is in waiting state",
+      }
+
+    ALERT K8sPodRestartingTooMuch
+      IF sum (rate(kube_pod_container_status_restarts[5m]) > 0) BY(job, exported_pod, instance, namespace)
+      FOR 5m
```
brancz: This would fire if there is a single restart no? I think that might be a bit too sensitive for a critical warning. How about a `severity="warning"` for this one over `10m` time and `> 1` and `> 3` can be critical. Wdyt @fabxc ?

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -516,4 +516,52 @@ data:
...
+        description = "Pod {{ $labels.namespace }}/{{ $labels.exported_pod }} is restarting constantly",
+      }
+
+  docker.rules: |+
+    ALERT DockerHasErrors
+      IF irate(kubelet_docker_operations_errors[10m])  > 0
+      FOR 1m
+      LABELS {
+        service = "docker",
+        severity = "critical",
```
brancz: I would do a warning here and a critical second one with a high rate of errors

brancz:
Thanks a lot for these @gianrubio ! This is definitely something we need. @fabxc should have a glimpse over these as he has the most experience with these kinds of thresholds and timings.

gianrubio:
@brancz I agree with the changes, let's wait @fabxc opinion so I'll change this rules

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
fabxc: Yes, both alerts are on the pod level and could get pretty noisy. Critical/pages should be dedicated to service level objectives, i.e. "service is responding slowly". Pods not starting etc. are potential causes and should be a warning at max.

I'm wondering here how accurate the expression is in general. Rate will only return to 0 after the last restart is 5 minutes in the past, meaning the alert will resolve rather late.  
But it might also stop firing because the restart backoff got quite long, which can cause the alert to flap on and off between restart attempts.

Does it maybe make sense to catch all non-ready pods more generally? I think ultimately we want to know what ports are not running and the exact failure mode is something we'd look up with  `kubectl` anyway.

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -516,4 +516,52 @@ data:
         summary = "Kubelet is close to pod limit",
         description = "Kubelet {{$labels.instance}} is running {{$value}} pods, close to the limit of 110",
       }
-    
+
+    ALERT K8SPodWaiting
+      IF sum(sum_over_time(kube_pod_container_status_waiting[5m]) > 0 ) BY (job, exported_pod, instance, namespace)
```
fabxc: `exported_pod` is the metric's `pod` label prefixed by `exported_` as it would conflict with the exporters target `pod` label otherwise. In this case here, the exporter target should not have a `pod` label to begin with.

This is fine for this PR. We have to find a generalised approach to address this problem.

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
fabxc: Yes, same here. Docker daemon erroring does not necessarily impact the service. In case Kubernetes is a service provided by ops to developers, there are likely higher-level metrics indicating if devs cannot deploy for example.

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
fabxc: Why do we sum over time here before summing over containers? This would trigger even if *right now* there's no waiting container at all.

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -516,4 +516,52 @@ data:
...
+        severity = "critical",
+      }
+      ANNOTATIONS {
+        summary = "Pod is in a restart loop",
+        description = "Pod {{ $labels.namespace }}/{{ $labels.exported_pod }} is restarting constantly",
+      }
+
+  docker.rules: |+
+    ALERT DockerHasErrors
+      IF irate(kubelet_docker_operations_errors[10m])  > 0
```
fabxc: `irate()` is not recommended to be used in alerts as it is very spiky. It actually only considers the last two samples regardless of the 10m rate window.

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -516,4 +516,52 @@ data:
...
+      FOR 1m
```
fabxc: Probably 5 minutes is a more reasonable default.

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -516,4 +516,52 @@ data:
...
+        service = "docker",
+        severity = "critical",
+      }
+      ANNOTATIONS {
+        summary = "Docker has errors!",
+        description = "Docker on instance {{ $labels.instance }} has some errors",
+      }
+
+    ALERT DockerHasTimeout
+      IF irate(kubelet_docker_operations_timeout[5m]) > 0
```
fabxc: Same concern about `irate()` and `FOR`

contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
fabxc: While different aggregation syntaxes exist, we converged towards `sum by(...) (...)` upstream as it seems more readable overall. Staying consistent with that probably makes sense.

gytisgreitai: Sorry to bump into your review, another option might possibly be:  `rate(kube_pod_container_status_restarts[1m]) > 1/(5*60)` ? : `More than 1 restart in 5 minute period` ?

brancz:
The rules have changed rather significantly at this point, I'll close this here.

Issue #290, opened 2017-04-16
ServiceMonitor error: the server does not allow access to the requested resource


klinakuf:
**What did you do?**

- I deployed the bundle.yaml in the default namespace. 
- I deployed the Prometheus TPR:

```
apiVersion: monitoring.coreos.com/v1alpha1
kind: Prometheus
metadata:
  name: prometheus-mqs
  labels:
    prometheus: mqs
spec:
  # Define that all ServiceMonitor TPRs with the label `tier = frontend` should be included
  # into the server's configuration.
  serviceMonitorSelector:
      matchLabels:
        queue: rabbit
```
- When I deploy the **ServiceMonitor** then inside the logs of Prometheus pod I see the following error: 
```
2017-04-16T21:55:14.749575321Z time="2017-04-16T21:55:14Z" level=info msg="Using pod service account via in-cluster config" source="kubernetes.go:94" 
2017-04-16T21:55:14.793099473Z time="2017-04-16T21:55:14Z" level=error msg="github.com/prometheus/prometheus/vendor/k8s.io/client-go/1.5/tools/cache/reflector.go:109: Failed to list *v1.Endpoints: the server does not allow access to the requested resource (get endpoints)" component="kube_client_runtime" source="kubernetes.go:73" 
```

Is there something extra to be configured regarding the ClusterRoles and ServiceAccounts?

* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.1", GitCommit:"b0b7a323cc5a4a2019b2e9520c21c7830b7f708e", GitTreeState:"clean", BuildDate:"2017-04-03T20:44:38Z", GoVersion:"go1.7.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"7+", GitVersion:"v1.7.0-alpha.1", GitCommit:"176eb0e509a255dd520de135cda31096f8796be5", GitTreeState:"clean", BuildDate:"2017-04-06T20:00:18Z", GoVersion:"go1.7.5", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: 

kubeadm

klinakuf:
Here is the solution: https://github.com/coreos/prometheus-operator/blob/master/Documentation/rbac.md 

**However, this is still an issue because this material https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md is not up to date with RBAC**

[...]

PR #310, opened 2017-04-25
docs: Make getting-started guide RBAC aware

mxinden:
Currently the getting started guide only creates a `ClusterRole` and a `ClusterRoleBinding` for the Prometheus operator but not for the Prometheus pods. This PR adds the steps to do so for the Prometheus pods as well.

Tested with tectonic-installer and minikube cluster.

Related to #290

### DIFF
--- a/Documentation/rbac.md
+++ b/Documentation/rbac.md
@@ -10,7 +10,7 @@ Here is a ready to use manifest of a `ClusterRole` that can be used to start the
 
 [embedmd]:# (../example/rbac/prometheus-operator/prometheus-operator-cluster-role.yaml)
 ```yaml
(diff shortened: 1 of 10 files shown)


mxinden:
@brancz I have updated RBAC related apiVersions to `apiVersion: rbac.authorization.k8s.io/v1beta1`, otherwise `kubectl` complains. What do you think?

brancz:
@mxinden I think we should be good with upgrading the apiVersion for the RBAC resources, master should represent latest, which is 1.6 clusters for us now.

brancz:
lgtmog

mxinden:
Test this please.

PR #328, opened 2017-05-02
basic auth

### LINKED ISSUE #308: Enable basic auth on ServiceMonitor
The current ServiceMonitor [Endpoint](https://github.com/coreos/prometheus-operator/blob/master/pkg/client/monitoring/v1alpha1/types.go#L176) doesn't support basic auth. On the previous version I had a basic_auth for one of my jobs.

```
- job_name: graylog
  kubernetes_sd_configs:
  static_configs:
    - targets: ['graylog-metrics.logging:9100']

  basic_auth:
    username: prometheus
    password: <hidden>

  metrics_path: '/api/plugins/org.graylog.plugins.metrics.prometheus/metrics'
```

Should I do a PR to enable basic_auth on `ServiceMonitor`?

gianrubio:
Fix #308

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -858,8 +858,10 @@ func (c *Operator) createConfig(p *v1alpha1.Prometheus, ruleFileConfigMaps []*v1
 	if err != nil {
 		return errors.Wrap(err, "selecting ServiceMonitors failed")
 	}
+	sClient := c.kclient.CoreV1().Secrets(p.Namespace)
+
 	// Update secret based on the most recent configuration.
-	conf, err := generateConfig(p, smons, len(ruleFileConfigMaps))
+	conf, err := generateConfig(p, smons, len(ruleFileConfigMaps), sClient)
 	if err != nil {
 		return errors.Wrap(err, "generating config failed")
 	}
@@ -872,7 +874,6 @@ func (c *Operator) createConfig(p *v1alpha1.Prometheus, ruleFileConfigMaps []*v1
 		"generated": "true",
 	}
 	s.Data["prometheus.yaml"] = []byte(conf)
-	sClient := c.kclient.CoreV1().Secrets(p.Namespace)
 
 	_, err = sClient.Get(s.Name)
 	if apierrors.IsNotFound(err) {
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -169,7 +169,9 @@ func New(conf Config, logger log.Logger) (*Operator, error) {
 		&v1.Secret{}, resyncPeriod, cache.Indexers{},
 	)
 	c.secrInf.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.handleSecretAdd,
 		DeleteFunc: c.handleSecretDelete,
+		UpdateFunc: c.handleSecretUpdate,
 	})
 
 	c.ssetInf = cache.NewSharedIndexInformer(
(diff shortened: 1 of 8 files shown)


brancz:
As you already mentioned in https://github.com/coreos/prometheus-operator/issues/308#issuecomment-298661198 it would be better if we can pre-load the secrets before we generate the config, that part should be purely about generating, not performing any access, that way we can also avoid erroring in there.

I'd suggest creating a map that has unique identifiers (just like we used [here](https://github.com/gianrubio/prometheus-operator/blob/0aa12a4358ecbfac296c1953f1f9cfce707fbcd9/pkg/prometheus/promcfg.go#L78)) to the endpoint and values being a BasicAuth struct that has username/password fields.

brancz:
There are also some conflicts now, could you rebase?

gianrubio:
@brancz could you verify if this structure is acceptable?

pkg/prometheus/operator.go
```diff
@@ -853,13 +853,77 @@ func (c *Operator) destroyPrometheus(key string) error {
...
+							password = string(p)
+						} else {
+							return nil, fmt.Errorf("Secret username of servicemonitor %s not found.",
+								mon.Name)
+						}
+
+					}
+				}
+
+				secrets[fmt.Sprintf("%v/%v", mon, ep)] = map[string]interface{}{
```
brancz: Since we already have a [format of identifying a single endpoints config section](https://github.com/gianrubio/prometheus-operator/blob/88a25963c2a61e2103970a35ae35574039c4f3e3/pkg/prometheus/promcfg.go#L77) I think it would be best to reuse that here.

pkg/prometheus/operator.go
```diff
@@ -853,13 +853,77 @@ func (c *Operator) destroyPrometheus(key string) error {
 	return nil
 }
 
+func (c *Operator) loadBasicAuthSecrets(mons map[string]*v1alpha1.ServiceMonitor, s *v1.SecretList) (map[string]interface{}, error) {
+
+	secrets := map[string]interface{}{}
```
brancz: We're sort of mixing the config generation and secret loading part here. As long as we can I think it would be best to stay type-safe here and leave the `interface{}{}` doings to completely up to the config generator. I'd suggest to introduce a new struct here, like

```
type basicAuthCredentials struct {
  username string
  password string
}
```

And use that instead of `map[string]interface{}{}`, and then in the config generation step convert that to `map[string]interface{}{}`. That way we can rely on the fields existing when we pass it to the config generator.

gianrubio:
@brancz could you review again?

Documentation/user-guides/cluster-monitoring.md
```diff
@@ -480,6 +480,49 @@ spec:
 
 Read more in the [alerting guide](alerting.md) on how to configure the Alertmanager as it will not spin up unless it has a valid configuration mounted through a `Secret`. Note that the `Secret` has to be in the same namespace as the `Alertmanager` resource as well as have the name `alertmanager-<name-of-alertmanager-object` and the key of the configuration is `alertmanager.yaml`.
 
+## Basic auth
```
brancz: I think this should be a separate user-guide documentation. Called something like "Basic Auth for targets", and the way the `embedmd` things work is that the file content is synced into the documentation with the `make generate` make target. So just make sure those files exist and then run that make target, then the content will be synced (we test that the documentation is consistent in one of the jenkins jobs).

gianrubio: Ok! I splited the Basic Auth onto a new file and sorry for the `embedmd` I just forgot to add the files to git.

brancz:
Great that all looks great. @fabxc reminded me to watch out for the informers and indeed we need to make sure to register further handlers now here: 

https://github.com/gianrubio/prometheus-operator/blob/30120d32a7996c515eba740e0e424d980188a47d/pkg/prometheus/operator.go#L172

To basically have the same handlers as for the ServiceMonitors.

https://github.com/gianrubio/prometheus-operator/blob/30120d32a7996c515eba740e0e424d980188a47d/pkg/prometheus/operator.go#L392-L411

The current `handleSecretDelete` handler can be completely replaced with the logic to enqueue for that namespace.

Let me know if this makes sense, if not I can also quickly do this.

gianrubio:
Make sense, otherwise when the user change the secrets, operator will not generate new configuration. I did the changes!

brancz:
test this please

brancz:
whops, looks like there is something not working with the e2e test job @mxinden ?

brancz:
Sorry @mxinden actually informed me of this, we are undergoing some changes in the jenkins infrastructure. Testing should be back shortly.

brancz:
Ran e2e tests locally and they were green so merging. :tada: Thanks a lot @gianrubio ! Nice work!

PR #329, opened 2017-05-02
client: return errors directly from Kubernetes dynamic client

brancz:
Wrapping the errors the way we did before this does not allow us to
check the errors for equality, and did not add any valuable semantics.

@fabxc @mxinden

### DIFF
--- a/pkg/client/monitoring/v1alpha1/alertmanager.go
+++ b/pkg/client/monitoring/v1alpha1/alertmanager.go
@@ -16,7 +16,6 @@ package v1alpha1
 
 import (
 	"encoding/json"
-	"fmt"
 
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/pkg/api"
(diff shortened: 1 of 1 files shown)


mxinden:
@brancz Out of curiosity where do we need to compare error messages?

brancz:
@mxinden when using this client in the tectonic-prometheus-operator :slightly_smiling_face:

pkg/client/monitoring/v1alpha1/alertmanager.go
```diff
@@ -83,20 +82,20 @@ func (a *alertmanagers) Create(o *Alertmanager) (*Alertmanager, error) {
...
 	ua, err := UnstructuredFromAlertmanager(o)
 	if err != nil {
-		return nil, fmt.Errorf("unstructuring alertmanager failed: %s", err)
+		return nil, err
 	}
 
 	ua, err = a.client.Update(ua)
 	if err != nil {
-		return nil, fmt.Errorf("updating alertmanager failed: %s", err)
+		return nil, err
```
fabxc: Using `error.Wrap` does not work instead?

brancz: I didn't feel it would add a lot more information to it here, and if we did wrap the error we would be requiring the user of this client to use `errors.Cause`, and therefore forcing a dependency on a user. I'd like to avoid that.

fabxc:
👍

PR #330, opened 2017-05-02
prometheus: Fall back to externalIP addresses

mikebryant:
Why:
* Some clusters (e.g. openstack) may only have externalIP addresses

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -282,6 +282,34 @@ func (c *Operator) handleAddNode(obj interface{})         { c.syncNodeEndpoints(
 func (c *Operator) handleDeleteNode(obj interface{})      { c.syncNodeEndpoints() }
 func (c *Operator) handleUpdateNode(old, cur interface{}) { c.syncNodeEndpoints() }
 
+// nodeAddresses returns the provided node's address, based on the priority:
+// 1. NodeInternalIP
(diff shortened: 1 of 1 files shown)


fabxc:
Thanks, fine with adding this in general. Actually, Prometheus's own node discovery does something similar: https://github.com/prometheus/prometheus/blob/master/discovery/kubernetes/node.go#L172-L198

Can we just copy this function here and use it as well. That's the "official" fallback order and we should probably stick to that.

mikebryant:
@fabxc How's that?

pkg/prometheus/operator.go
```diff
@@ -308,19 +336,18 @@ func (c *Operator) syncNodeEndpoints() {
...
-					TargetRef: &v1.ObjectReference{
-						Kind:       "Node",
-						Name:       n.Name,
-						UID:        n.UID,
-						APIVersion: n.APIVersion,
-					},
-				})
-			}
+		address, _, err := nodeAddress(n)
+		if err == nil {
```
fabxc: We should probably log that error and flip the condition to `err != nil` to move the larger block below one level to the left.

mikebryant: Done

brancz:
lgtm

fabxc:
Thanks! LGTM

PR #331, opened 2017-05-03
rollout new Prometheus on flag and resource changes

fabxc:
@mxinden @brancz note the bunch of changes in unrelated tests I made because we are not cleaning up properly or are creating namespace-less objects, which then cannot be used downstream.

In general, there are a bunch of objects potentially created and cleaning them all up in a  `defer` in each test is not feasible, which is why it's not being done. This means all subsequent tests may fail or, even worse, only succeed because of remains of other tests.

Funnily enough, in contrast to what I saw in the TPO, namespace deletion works fast and reliable here and could solve this.

In general though, I think we should move away from the framework global and towards instantiating a new namespace per test, which is asynchronously deleted afterwards. The operator can be created once before any tests start and deleted after all tests ran.
Non-namespaced objects and such can just be prefixed with a random string for each test.

Something along those lines should happen for PO and TPO alike. My confidence in the e2e tests is somewhat diminished by the range of potential side effects.

We should also define a stricter set of naming patterns for helper methods. A lot of them chain together two other small methods and the naming patterns do not consistently reflect semantics across them.  
Also a lot of them are super short and just renamings, e.g.:

```go
func DeleteService(kubeClient kubernetes.Interface, namespace string, serviceName string) error {
	if err := kubeClient.CoreV1().Services(namespace).Delete(serviceName, nil); err != nil {
		return errors.Wrap(err, fmt.Sprintf("deleting service %v failed", serviceName))
	}
	return nil
}
```


fabxc:
This PR doesn't seem to trigger an e2e test run.

mxinden:
Test this please.

mxinden:
Test this please.

mxinden:
@fabxc We are getting `java.lang.NullPointerException` from our GitHub Pull request builder plugin. I am investigating.

mxinden:
Test this please.

mxinden:
Test this please.

mxinden:
Test this please.

brancz:
@fabxc a unit test is failing as we forgot to remove it in an earlier PR, where the jenkins job didn't run

mxinden:
Test this please.

mxinden:
Re-triggering tests due to Ingress flake.

brancz:
Test this please.

brancz:
The failing test is definitely the ingress flake. Other than that lgtm

mxinden:
Yes, just the ingress flake :/

`ReloadConfig` is failing on my local minikube setup. Would like to invest
a little into that before merging.

brancz:
@mxinden fyi I've seen that one failing locally before as well, but working on jenkins.

PR #334, opened 2017-05-03
Use kubectl apply in custom service example deploy

Tapppi:
https://github.com/kubernetes/kubernetes/issues/29542 has been fixed,
use apply instead of create to create third party resources in custom service
monitoring example.


brancz:
lgtm :+1: thanks!

PR #336, opened 2017-05-04
prometheus: do not delete secret when not managed by Operator

### LINKED ISSUE #333: Prometheus deletion deletes a Secret not managed by Prometheus
**What did you do?**
1. Create a `Secret` named `prometheus-foo` with `prometheus.yaml` data key.
2. Create a `Prometheus` resource named `foo` omitting `serviceMonitorSelector` in the spec.
3. Verify prometheus is launched with the config file specified in step 1.
4. Delete the Prometheus object created in step 2. (`kubectl delete prometheus foo`)
5. Check if Secret from step 1 is still there (`kubectl get secret prometheus-foo`). It is not.

5. Secret is not deleted when `Prometheus` resource is deleted.

5. Secret is deleted when `Prometheus` resource is deleted.

**Environment**
prometheus-operator: v0.8.2
k8s: 1.6

Context: #324

brancz:
Fixes #333 

@fabxc @mxinden

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -831,11 +831,23 @@ func (c *Operator) destroyPrometheus(key string) error {
 	}
 
 	// Delete the auto-generate configuration.
-	// TODO(fabxc): add an ownerRef at creation so we don't delete config maps
+	// TODO(fabxc): add an ownerRef at creation so we don't delete Secrets
 	// manually created for Prometheus servers with no ServiceMonitor selectors.
 	s := c.kclient.Core().Secrets(sset.Namespace)
-	if err := s.Delete(sset.Name, nil); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/prometheus/operator.go
```diff
@@ -831,11 +831,23 @@ func (c *Operator) destroyPrometheus(key string) error {
 	}
 
 	// Delete the auto-generate configuration.
-	// TODO(fabxc): add an ownerRef at creation so we don't delete config maps
+	// TODO(fabxc): add an ownerRef at creation so we don't delete Secrets
```
mxinden: @brancz Is this still a *ToDo*?

fabxc: Yes

fabxc:
👍

PR #338, opened 2017-05-04
test: Introduce test context per test

mxinden:
* Introduce test context and separate namespace per test
* Remove unneeded delete steps
* Improve error feedback
* Cleanup main namespace properly after tests
* Wait for NodePort services to be deleted

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -28,85 +28,101 @@ import (
 )
 
 func TestAlertmanagerCreateDeleteCluster(t *testing.T) {
+	ctx := testFramework.NewTestCtx(t)
+	defer ctx.CleanUp(t)
+	ctx.BasicSetup(t, framework.KubeClient)
+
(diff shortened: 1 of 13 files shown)


test/e2e/alertmanager_test.go
```diff
@@ -28,86 +28,83 @@ import (
 )
 
 func TestAlertmanagerCreateDeleteCluster(t *testing.T) {
+	ctx := testFramework.NewTestCtx(t)
+	defer ctx.CleanUp(t)
+	ctx.BasicSetup(t, framework.KubeClient)
+
 	name := "test"
 
 	defer func() {
```
brancz: If we're testing the actual deletion here it makes more sense to actually put the delete after the create instead of deferring it.

mxinden: 👍 Fixing in #340

mxinden:
Closing here. Continuing in #340

PR #340, opened 2017-05-04
test: Introduce a context and a namespace per test and run save tests in parallel

mxinden:
test: Introduce test context per test

test: Clean up e2e test code
* Remove unneeded delete steps
* Improve error feedback
* Cleanup main namespace properly after tests
* Wait for NodePort services to be deleted

test: Run save tests in parallel

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -34,13 +34,11 @@ func TestAlertmanagerCreateDeleteCluster(t *testing.T) {
 
 	name := "test"
 
-	defer func() {
-		if err := framework.DeleteAlertmanagerAndWaitUntilGone(ctx.Id, name); err != nil {
-			t.Fatal(err)
-		}
-	}()
-
 	if err := framework.CreateAlertmanagerAndWaitUntilReady(ctx.Id, framework.MakeBasicAlertmanager(name, 3)); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := framework.DeleteAlertmanagerAndWaitUntilGone(ctx.Id, name); err != nil {
 		t.Fatal(err)
 	}
 }
@@ -51,12 +49,6 @@ func TestAlertmanagerScaling(t *testing.T) {
 	ctx.BasicSetup(t, framework.KubeClient)
 
 	name := "test"
-
-	defer func() {
-		if err := framework.DeleteAlertmanagerAndWaitUntilGone(ctx.Id, name); err != nil {
-			t.Fatal(err)
-		}
-	}()
 
 	if err := framework.CreateAlertmanagerAndWaitUntilReady(ctx.Id, framework.MakeBasicAlertmanager(name, 3)); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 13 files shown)


test/e2e/alertmanager_test.go
```diff
@@ -28,86 +28,89 @@ import (
 )
 
 func TestAlertmanagerCreateDeleteCluster(t *testing.T) {
+	t.Parallel()
+
+	ctx := testFramework.NewTestCtx(t)
+	defer ctx.CleanUp(t)
```
fabxc: `CleanUp` -> `Cleanup`. At least we aren't calling it `SetUp` anywhere either.

test/e2e/alertmanager_test.go
```diff
@@ -28,86 +28,89 @@ import (
...
+	ctx.BasicSetup(t, framework.KubeClient)
```
fabxc: This just creates a namespace. In favor of explicitness, I'd just rename this to `ctx.CreateNamespace` and actually have it return the namespace name.

Below we are calling functions expecting a namespace name with `ctx.Id`, which seems somewhat cryptic unless you know that `Id` is always equal to a namespace created during the basic setup.

```go
ns := ctx.CreateNamespace(test) // creates namespace `<ctx.Id>-test`
// ...
framework.DeleteAlertmanagerAndWaitUntilGone(ns, name)
```

This seems a bit clearer overall. What I find important is that it translates seamlessly for tests that need to create multiple namespaces, which would not have an obvious solution from looking at tests that just need `BasicSetup`.

```go
ns1 := ctx.CreateNamespace("test1")
ns2 := ctx.CreateNamespace("test2")
// .. do whatever needed with either namespace
```

test/e2e/alertmanager_test.go
```diff
@@ -28,86 +28,89 @@ import (
 )
 
 func TestAlertmanagerCreateDeleteCluster(t *testing.T) {
+	t.Parallel()
+
+	ctx := testFramework.NewTestCtx(t)
```
fabxc: I think `NewTestCtx` should be a method to the `framework` we are using elsewhere. 
This way you don't have to pass in `framework.KubeClient` if you want this context to create a non-namespaced object.

test/e2e/alertmanager_test.go
fabxc: Ah, the Go convention is to always keep abbreviations uppercase, e.g. `ServeHTTP`. So `ctx.Id` should be `ctx.ID`.

mxinden:
Test this please.

test/e2e/alertmanager_test.go
brancz: Agreed.

brancz:
Nice work! If it makes things easier maybe close #338 and we'll do everything in here. Your choice @mxinden.

mxinden:
Test this please. CoreOS bot.

mxinden:
test this please.

mxinden:
please test this coreos bot

mxinden:
please test this coreosbot

mxinden:
test this please

mxinden:
please test this

test/e2e/alertmanager_test.go
mxinden: @fabxc Ok, I have changed this. Do you want to save the kubeClient in each ctx on creation as well? Would seem strange to me.

mxinden:
@fabxc @brancz Thanks for the feedback. Addressed your comments from #340 and #338 
Any further improvements?

test/e2e/alertmanager_test.go
```diff
@@ -28,86 +28,87 @@ import (
 )
 
 func TestAlertmanagerCreateDeleteCluster(t *testing.T) {
+	t.Parallel()
+
+	ctx := framework.NewTestCtx(t)
+	defer ctx.Cleanup(t)
+	ns := ctx.CreateNamespaceInCtx(t, framework.KubeClient)
```
brancz: The `InCtx` suffix is unnecessary here I think, as the `CreateNamespace` method would be bound to the ctx struct anyways, that's "context" enough I think.

fabxc: Yep, this.

mxinden:
test this please

PR #345, opened 2017-05-05
prometheus: remove automatically relabelling kubernetes labels

brancz:
This may relabel unwanted labels onto the targets and if any labels are
necessary for further querying they can be joined onto the results using
the label metrics available in kube-state-metrics. If it turns out we
need explicit support in the Prometheus Operator then we can reintroduce
it with an explicit whitelisting approach.

@fabxc @mxinden


fabxc:
👍

PR #316, opened 2017-04-27
Honor labels

brancz:
Fixes #309 

This PR does 3 things that are somewhat connected to each other:

* add ability to specify the `honor_labels` configuration
* remove non-namespaced alpha annotation
* update kube-prometheus ServiceMonitors to reflect the previous two changes

@fabxc @mxinden

PR #349, opened 2017-05-06
User proper metric property for namespace

sheerun:
Related: https://github.com/kubernetes/kube-state-metrics/issues/138

### DIFF
--- a/contrib/kube-prometheus/manifests/grafana/grafana-dashboards.yaml
+++ b/contrib/kube-prometheus/manifests/grafana/grafana-dashboards.yaml
@@ -1218,7 +1218,7 @@ data:
                         },
                         "targets": [
                             {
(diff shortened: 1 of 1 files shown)


sheerun:
I've also fixed `exported_pod` metric name

brancz:
As I mentioned in kubernetes/kube-state-metrics#138 it is actually the `honor_labels` configuration that needs to be set correctly for Prometheus not to resolve label collisions like that.

Setting the `honor_labels` configuration is not supported in any currently available release of the Prometheus Operator, but has been fixed by #316, we will soon be releasing v0.9.0 of the Prometheus Operator which will include it. The state of kube-prometheus on master reflects the state of it with [using `honor_labels`](https://github.com/coreos/prometheus-operator/blob/ccbaf1dbc9bf0327a8657cf0ac2d1e7c2375fbc9/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-service-monitor-kube-state-metrics.yaml#L19).

PR #351, opened 2017-05-08
test: Move e2e run logic into scripts

mxinden:
This moves the e2e execution logic to shell scripts. 
In the long run these scripts will be reused to run tests on master.

### DIFF
--- a/scripts/jenkins/post-e2e-tests.sh
+++ b/scripts/jenkins/post-e2e-tests.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# This is a cleanup script, if one command fails we still want all others to run
+# set -e
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+# error on unset variables
+set -u
+# print each command before executing it
+set -x
+
+PO_GOPATH=/go/src/github.com/coreos/prometheus-operator
+
+docker run \
+       --rm \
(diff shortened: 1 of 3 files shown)


mxinden:
test this please

scripts/jenkins/post-e2e-tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
brancz: safer to use `/usr/bin/env bash`, as some distros (like ubuntu) use odd things like `dash` in this case which can cause this script to fail .. better to describe the environment explicitly here that you want to use

mxinden: @brancz Good hint. Thanks

gianrubio:
Sorry for introdude on it :)

Please use `set -exo pipefail` after /bin/bash so it can fail fast on errors and gives more output.

I also recommend to add the same command on all e2e shell scripts

brancz:
Agreed with @gianrubio! Suggesting best practices is never intruding :slightly_smiling_face:

brancz:
lgtmog

mxinden:
test this please

mxinden:
test this please

mxinden:
test this please

mxinden:
@gianrubio Thanks for the feedback, always very appreciated!

PR #354, opened 2017-05-09
Chart v0.0.2

### LINKED ISSUE #353: Latest HELM chart does not include kubeletObject options
**What did you do?**
Using HELM chart to install Prometheus-Operator and wanting to have the kubeletObject values set

KubeletObject value set to defaults and the kubelet-object argument sent to Prometheus-operator on startup

No kubelet-object argument gets passed to Prometheus-operator upon startup.

The code changes for including the kubeletObject values are shown as having been reviewed and checked in on Apr. 7, and these are included in the latest README documentation, but the latest chart at http://charts.opsgoodness.com/prometheus-operator-0.0.1.tgz does not include these updates.  

Would really like to have these updates included as they are needed to get the kubelet information.

mgoodness:
Fixes #353 

* Renamed `kubeletObject` -> `kubeletService` to make the effect clearer (IMHO)
* Renamed `kubeletObject.object` -> `kubeletService.name` for same reason
* Updated default prometheus-operator image to `v0.8.2`

I'll publish v0.0.2 to my `opsgoodness` repo once this is merged.

### DIFF
--- a/helm/prometheus-operator/values.yaml
+++ b/helm/prometheus-operator/values.yaml
@@ -10,7 +10,7 @@ global:
 ##
 image:
   repository: quay.io/coreos/prometheus-operator
-  tag: v0.7.0
+  tag: v0.8.2
   pullPolicy: IfNotPresent
 
 ## Node labels for prometheus-operator pod assignment
@@ -37,9 +37,8 @@ resources: {}
   #   cpu: 100m
   #   memory: 50Mi
 
-## The prometheus operator will create an endpoint for the kubelet.
-## Format: namespace/name
+## If enabled, prometheus-operator will create a service for scraping kubelets
 ##
-kubeletObject:
+kubeletService:
   enable: true
-  object: kube-system/kubelet
+  name: kube-system/kubelet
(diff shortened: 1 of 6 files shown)


helm/prometheus-operator/values.yaml
```diff
@@ -37,9 +37,8 @@ resources: {}
...
 
-## The prometheus operator will create an endpoint for the kubelet.
-## Format: namespace/name
+## If enabled, prometheus-operator will create a service for scraping kubelets
 ##
-kubeletObject:
+kubeletService:
   enable: true
-  object: kube-system/kubelet
+  name: kube-system/kubelet
```
brancz: since we have this abstraction layer here it would make sense to divide this into two fields: name and namespace. Then we can do the concatenation in the actual template.

brancz:
Also we're about to release 0.9.0 so maybe we can hold off on this until that release is done, wdyt?

mgoodness:
I was just going to comment to that effect. Same-day chart release ftw!

brancz:
I'm guessing you tested this, in that case lgtm :+1:

PR #366, opened 2017-05-12
prometheus: add experimental 2.0 support

brancz:
Tested this again current state of the Prometheus `dev-2.0` branch and it works nicely.

Let me know if you want to wait to merge this until we actually release `v2.0.0-alpha.1`, but this code is functional against current state of 2.0.

@fabxc @mxinden


mxinden:
@brancz Like the idea! Hopefully lowers the barrier for testing out Prometheus 2.x

fabxc:
Mh, one could argue for guarding against the full version as flag changes might crash later ones if people run an old operator. OTOH... the result is the same, Prometheus won't run. So this is probably fine.

fabxc:
👍

PR #367, opened 2017-05-14
prometheus: generate config deterministically

brancz:
Because maps have a non-deterministic order when newly instantiated, we need to use means to make sure that the prometheus yaml configuration we generate is deterministic.

This changes any use of a map in the config to use the `yaml.MapSlice`, which ensures deterministic map behavior, because the `yaml.MapItem`s are sorted by their keys before marshaling to yaml.

Note that this does rely on Kubernetes not changing the order or arrays in objects, as the `Alertmanagers` and `Endpoints` lists are reflected 1 to 1 in the generated config. If this does turn out to be a problem we'll have to find a way to sort those items as well, but that will likely require individual sorting for each of the lists.

Let me know if you want to keep the test that I wrote to make sure that it is deterministic or not.

@fabxc @mxinden

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -17,6 +17,7 @@ package prometheus
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
 	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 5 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -37,133 +38,175 @@ func configMapRuleFileFolder(configMapNumber int) string {
...
 	for _, am := range p.Spec.Alerting.Alertmanagers {
 		alertmanagerConfigs = append(alertmanagerConfigs, generateAlertmanagerConfig(am))
 	}
 
-	cfg["scrape_configs"] = scrapeConfigs
-	cfg["alerting"] = map[string]interface{}{
-		"alertmanagers": alertmanagerConfigs,
-	}
+	cfg = append(cfg,
+		yaml.MapItem{
```
fabxc: I think typical formatting would have this in the same line as `append`.

fabxc:
LGTM in general.

In hindsight, one could have probably written a recursive function transforming our previous `map[string]interface{}` nestings to the equivalent `yaml.MapSlice` :) but now that it's all converted, that's fine too.

mxinden:
👍 For keeping the tests! Especially to ensure this when upgrading k8s versions.

PR #369, opened 2017-05-16
test: Add slack integration to master job

mxinden:
From now on, if a job on master fails, it will be reported in our #team-monitoring slack channel.


brancz:
lgtm

PR #371, opened 2017-05-17
Run kube-state-metrics as single replica

fabxc:


brancz:
that'll require a `make generate` then lgtm

Issue #218, opened 2017-03-15
Work with externalName of Kubernetes service


fabxc:
Kubernetes services have an `externalName` field  as part of their spec (https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_servicespec), which maps to a DNS record instead of having an associated endpoints object.
Prometheus's K8S discovery only works with endpoint objects and probably should keep doing so.

Instead the operator could generate a DNS discovery config for the Prometheus instance if a ServiceMonitor references a service with an `externalName`.

tonglil:
Are there any updates on this? It would be a huge blessing for using Prometheus operator more generally for easy operations to support the collection of any metrics, not just ones from in the cluster.

max-lobur:
What's the current status? I've been trying to get this working on latest prometheus-operator, using serviceMonitor which selector points to the externalName type svc - no luck :(

brancz:
External type services are not possible to be monitored in a general whitebox style. If this is what you want to do, I recommend using a separate service discovery mechanism offered by Prometheus to discover the whitebox targets of the external service.

max-lobur:
@brancz thanks for pointing.
Yeah I realized that ELB won't show up the internals so you don't know the targets.
So I end up with a EC2 discovery based scraper:
```
additionalScrapeConfigs:
      - job_name: 'mypp'
        metrics_path: '/metrics'
        ec2_sd_configs:
          - region: us-east-2
            access_key: 
            secret_key: 
            port: 80
        relabel_configs:
          - source_labels: [__meta_ec2_tag_Name]
            regex: MyApp.*
            action: keep
          - source_labels: [__meta_ec2_private_ip]
            target_label: instance
```

brancz:
Using EC2 discovery sounds good. Have a look at our docs regarding [how to use the additionalScapeConfigs feature](https://github.com/coreos/prometheus-operator/blob/master/Documentation/additional-scrape-config.md).

brancz:
I think we can close this. Given that these services are "external", there is no general purpose way for us to monitor them. Using DNS+SRV discovery is also very much generalizing as that would assume that those types of records actually exist. I think the only real way to monitor external services is to actually use a separate discovery mechanism that is available for that service, kubernetes discovery is simply always about in-cluster workloads.

PR #372, opened 2017-05-17
Add support for externalName services

Chronojam:
Hi,

This is an attempt to add support for externalName services, using the following configuration:
( in reference to #218 )
```
apiVersion: v1
kind: Service
metadata:
  labels:
    local: local-k8s
  name: local-k8s
spec:
  externalName: localhost
  ports:
  - name: web
    port: 9090
    protocol: TCP
    targetPort: 9090
  sessionAffinity: None
  type: ExternalName
status:
  loadBalancer: {}
```

```
kind: ServiceMonitor
metadata:
  labels:
    local: k8s
  name: local-k8s
spec:
  externalName:
    dnsType: AAAA
    port: web
  selector:
    matchLabels:
      local: local-k8s
```

I was able to generate the following configuration
```
- job_name: default/local-k8s/0
  dns_sd_configs:
  - names:
    - localhost
    type: AAAA
    refresh_interval: 30s
    port: 9090
  relabel_configs:
  - replacement: default
    target_label: namespace
  - replacement: local-k8s
    target_label: service
```

Its still a WIP ( still need to test SRV records, and support matchExpressions ) but im hoping I can get your thoughts on the current approach?

Thanks


pkg/prometheus/promcfg.go
```diff
@@ -83,8 +87,25 @@ func generateConfig(p *v1alpha1.Prometheus, mons map[string]*v1alpha1.ServiceMon
 
 	var scrapeConfigs []yaml.MapSlice
 	for _, identifier := range identifiers {
-		for i, ep := range mons[identifier].Spec.Endpoints {
-			scrapeConfigs = append(scrapeConfigs, generateServiceMonitorConfig(mons[identifier], ep, i, basicAuthSecrets))
+		if len(mons[identifier].Spec.Endpoints) == 0 {
+			listOpts := metav1.ListOptions{
+				LabelSelector: metav1.FormatLabelSelector(&mons[identifier].Spec.Selector),
+			}
+			svcList, err := kclient.Core().Services(mons[identifier].Namespace).List(listOpts)
```
brancz: So far all config generation does not do any network calls or retrieval but is a simple function of input -> output, it would be good if we can keep it that way. Basically just move the retrieval part out of here and just pass in the list of services, you need.

Additionally actually calling out to the Kubernetes API is not necessary here, you can just read all Services from the informer cache of the Operator struct.

pkg/prometheus/promcfg.go
```diff
@@ -109,7 +130,131 @@ func generateConfig(p *v1alpha1.Prometheus, mons map[string]*v1alpha1.ServiceMon
...
+			// Fallback to the first port number on the service
+			portConf.Value = svc.Spec.Ports[0].Port
+		}
 
+		dnsConf = append(dnsConf, portConf)
+	}
+
+	cfg = append(cfg, yaml.MapItem{Key: "dns_sd_configs", Value: []yaml.MapSlice{dnsConf}})
+
+	if m.Spec.ExternalName.Interval != "" {
```
brancz: From here on onwards is essentially duplicate logic of what is generated for the kubernetes endpoints discovery. You can just expect an Endpoints struct to be passed, then you just pass in `m.Spec.ExternalName.Endpoint` to get the right type, or just not embed the type but make it a field of the `ExternalName` struct.

Chronojam:
Heya, Thanks for the feedback. 

Ive attempted to use the informer cache on the operator like you mentioned. Although it is the first time ive had a go at it, so I hope this is the correct way to go about it.

There's currently a problem where the selector im creating causes all services to be selected, rather than the ones specific to that service monitor, (likely something to do with only checking for match expressions i expect) so ill probably look at that in the next couple of days.

I think having external name services handled by Prometheus itself sounds like an excellent idea.

Chronojam:
This is now ready for a review.
Ive fixed up the previous issue and tried to add some documentation,

fabxc:
So we are generally keen to have this supported. I haven't worked with ExternalName services myself, so I'm wondering about the DNS record type:

Is there no specification/expectation at all what should this CNAME should point to? In my ideal view, this would one day be directly supported in the K8S discovery of Prometheus, without the indirection of DNS discoveries.
This would mean though that the K8S SD has to do this internally. If there's no constraint on what record type we have to look for, this will obviously never work. 

Seems kind of a bummer that ExternalName is supposed to make the externality transparent in K8S but ultimately might not actually be.

Chronojam:
I believe it can CNAME to anything, I imagine the way the discoverer could handle it is by doing a lookup against all types, then prioritise on a set order?

It would probably be part of the endpoints role discoverer

```
## Prom config
 kubernetes_sd_configs:
  - role: endpoints

## Kube service config
External service:
 blah...
 type: ExternalName
 ExternalName: foo.bar.com

## Do a lookup on all DNS types, discover
## Available DNS types for given CNAME
foo.bar.com AAAA 
foo.bar.com SRV

Endpoint:
  # Created because AAAA type has higher priority than SRV
  <Resolved Address of foo.bar.com:AAAA>
```
Then the k8s discoverer creates an endpoint scrape based on priority?

Or it could generate an endpoint for each type available, then attach some labels so the user can relabel to whatever they want?

__meta_kubernetes_service_dnstype or something? 

```
Endpoints:
 <Resolved Address of foo.bar.com:AAAA>
 <Resolved Address of foo.bar.com:SRV:1>
 <foo.bar.com:SRV:2>

  .. etc
```

the second solution is probably my favourite

brancz:
@Chronojam that sounds reasonable, do we want to take this upstream?

Chronojam:
Yup, that sounds good - ive opened an issue there to track it. (https://github.com/prometheus/prometheus/issues/2791)
Ill close this now

Thanks!

PR #377, opened 2017-05-18
cut 0.9.1

brancz:
Preliminary tests show that this works but we'll want to hold off on merging this until the Prometheus v1.6.3 release is through. However, I wanted to get this out so we can get started on reviewing this so it's just a merge away once the Prometheus release is done and we have a final check that this works.

@fabxc @mxinden


brancz:
Test this please

brancz:
checked and it works like a charm, releasing :tada:

PR #379, opened 2017-05-18
A bit of configuration instructions for AlertManager discovery from Prometheus

ds0nt:


brancz:
Yeah I think functionality wise this is the easiest for users to understand, technically it is the `Endpoints` object, but I think in this document it is fine.

ds0nt:
nice. mission accomplished!

PR #383, opened 2017-05-22
Tolerations 180

mmerrill3:
Implement Tolerations to fix issue 180.  Updated API documentation to reflect new optional parameter


brancz:
We will probably not introduce more admission control features into the Prometheus Operator as this is essentially an endless game of syncing with the upstream PodTemplate. Instead we were playing with the idea of using `PodPreset`s to set any of these features for Prometheus Pods, this way any admission control feature is also immediately available when it is introduced in Kubernetes and won't require additional implementation here.

I hope this wasn't too much hassle for you, but I can't promise we will merge this. I hope you understand. We should have definitely communicated this on the issue.

mmerrill3:
Thanks for the response, it was no issue for us, since we had an immediate need to implement tolerations.  I figured sharing the work with the community only helps to further the project.  Thanks for the guidance on PodPreset.

onzo-operry:
@brancz Hi any rough time scale on implementing the PodPreset? as upgrading to 1.6 and using taints has a negative effect on cluster monitoring :)

brancz:
`PodPreset`s have only landed as alpha in 1.6, so we're hesitating to to rely on yet another alpha feature. We were also debating whether it might make sense to use a `PodTemplate`, that is used as a base to build the `StatefulSet`, but have also found a number of problems with this approach. For the time being we'll have to deal with just using the node-selector for adjusting scheduling.

We realize this is definitely something that is required, but don't want to introduce fields that we will deprecate in the next version again.

Suggestions are welcome :slightly_smiling_face: .

brancz:
superseded by #600

PR #385, opened 2017-05-24
Don't actively delete pods on resource changes

fabxc:
Changes of the requested resources will be reflected in pods and still trigger an update. But it can no longer cause flaps by resource overwrites by admission control.

@brancz


brancz:
lgtm

PR #388, opened 2017-05-26
Support relabeling __address__ and __metrics_path__ for endpoints

### LINKED ISSUE #387: How to use with out-of-pod exporters?
We are trying to use Prometheus operator to monitor our production workloads. We have a number of 
use-cases where we want to run exporter as a separate pod from workload. Examples are:

- We are running PostgreSQL database using StatefulSet.
- We run HAProxy based ingress controller using deployment.

The reasons we want to run exporter as a separate pod from workload instead of a side-car are:

- Exporter can be added later without restarting production workload.
- Exporter can be updated to add more meaningful metrics without restarting workload.
- The same exporter can be used to expose metrics for workloads of same kind (multiple postgres DB in same/diff namespace, etc.). This is particularly useful for smaller cluster.

If we were writing Prometheus config by hand, this can be done by rewriting the `__address__` and `__metrics_path__` labels, as shown [here](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml#L79).

The way Prometheus operator generates Prometheus config, there is no way to add these labels. Can this model of out-of-pod exporters be supported in operator? Is there way to do this already in operator? We are willing to contribute code, if that helps to add this feature.

cc: @aerokite

TamalSaha:
- An `Address` field added to `Endpoint` struct. If set, this will be used to replace `__address__`

- Endpoint paths are parsed to identify if this contains one or more `__meta` labels and necessary replacement config is generated. This also handles query string.

Fixes #387

### DIFF
--- a/pkg/client/monitoring/v1alpha1/types.go
+++ b/pkg/client/monitoring/v1alpha1/types.go
@@ -181,6 +181,8 @@ type ServiceMonitorSpec struct {
 
 // Endpoint defines a scrapeable endpoint serving Prometheus metrics.
 type Endpoint struct {
+	// Address of http server scraped for this endpoint
+	Address string `json:"address,omitempty"`
 	// Name of the service port this endpoint refers to. Mutually exclusive with targetPort.
 	Port string `json:"port,omitempty"`
 	// Name or number of the target port of the endpoint. Mutually exclusive with port.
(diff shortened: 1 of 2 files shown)


brancz:
As described in #387, this is an anti-pattern and is not something we want to encourage doing. If you absolutely need this, I recommend using a custom config.

PR #392, opened 2017-05-31
small changes in k8s/minikube and example-app

eedugon:
I found some minor issues that I have fixed. Please review and let me know if you agree:

1) in example-app, the selector of prometheus-frontend-svc was incorrect.
2) in kube-scheduler and kube-controller discovery services for minikube, the namespace has to be kube-system to be compatible with the current deployment of k8s-app ServiceMonitor, otherwise the service won't be selected by the ServiceMonitor.
(note: the self-hosted resources are okay)


brancz:
@coreosbot please test this

mxinden:
test this please

brancz:
makes sense, thanks! lgtm

PR #393, opened 2017-05-31
Clarify Alertmanager configuration setup on the user guide

### LINKED ISSUE #389: Clarify alertmanager configuration and templates documentation
Hi!

I was configuring the altermanager on the operator and I've reach the point that the documentation wasn't very clear.

I started with the alertmanager configuration, I read the [spec](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#alertmanagerspec) and didn't saw where is the alertmanager configuration set, then I went to the [user guides](https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/alerting.md) and I could imagine how to do it, but it's not clear at all, so I started [reading the code](https://github.com/coreos/prometheus-operator/tree/e31a9dc1f309ae7ef1aa479802303e33d69a333e/pkg/alertmanager/statefulset.go#L273-L279) and I saw that the secret needs to be called `alertmanager-{alertmanager_name}`. 

Then it came the templating stuff, my configuration needs templates, so I searched for a place to set a ConfigMap or Secret and I didn't saw anything on the [AlertmanagerSpec](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#alertmanagerspec) neither, I ended up setting it on the same secret resource as the configuration, like this:
```
apiVersion: v1
kind: Secret
metadata:
  name: alertmanager-main
data:
  alertmanager.yaml: ............
  hipchat.tmpl: ............
```
and then using this path for the templates on the alertmanager.conf:
```
templates:
- '*.tmpl'
```

slok:
This PR tries to clarify (or at least explain the magic behind the operator) the setup of the configuration on the alertmanager along with alertmanager's templating files.

Closes #389

### DIFF
--- a/Documentation/user-guides/alerting.md
+++ b/Documentation/user-guides/alerting.md
@@ -44,8 +44,35 @@ receivers:
 
 Save the above alertmanager config in a file called `alertmanager.yaml` and create a secret from it using `kubectl`.
 
+The Alertmanager instance needs an special naming on the secret resource in order to be able to grab it. It looks for
+`alertmanager-{ALERTMANAGER_NAME}` pattern, in this example the name of the Alertmanager is `example` so the secret name
+should be `alertmanager-example` and the name of the config file `alertmanager.yaml` 
+
 ```bash
 $ kubectl create secret generic alertmanager-example --from-file=alertmanager.yaml
+```
+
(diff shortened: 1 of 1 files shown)


brancz:
Awesome! Thanks a lot! Lgtm

PR #400, opened 2017-06-08
*: default -route-prefix to root

brancz:
Whenever serving Prometheus or Alertmanager through non root
ExternalURLs we require the reverse proxy to trim the URL to be to the
root Prometheus/Alertmanager.

@fabxc @mxinden @Gouthamve


fabxc:
:+1:

PR #409, opened 2017-06-12
prometheus: always add route prefix flag

### LINKED ISSUE #408: Prometheus route-prefix flag not set to root
Although defaulting in other regards to the route prefix being `/`, the actual flag is only applied when it is != `/`. The flag should always be added, like in the Alertmanager Pods.

brancz:
Fixes #408 

@fabxc @mxinden


fabxc:
:+1:

PR #410, opened 2017-06-12
Fix basicAuth, swapping key and name

### LINKED ISSUE #402: basicAuth secret configuration is backwards
The [BasicAuth section of the API docs](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#basicauth) says that [v1.SecretKeySelector scheme](https://kubernetes.io/docs/api-reference/v1.6/#secretkeyselector-v1-core) from Kubernetes is used. Based on that spec, the `name` is the name of the secret to use, and `key` is the key inside that secret. I specified it accordingly, and my metrics collection kept silently failing with no errors in any of the logs:

```
spec:
  endpoints:
  - basicAuth:
      password:
        name: prometheus-basic-auth
        key: password
      username:
        name: prometheus-basic-auth
        key: username
# ...
```

Then I came across some of the user guide examples, the `name` and `key` fields are actually reversed in prometheus-operator:

https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/basic-auth.md#basic-auth-for-targets

Swapping the key names, everything started working:

```
spec:
  endpoints:
  - basicAuth:
      password:
        key: prometheus-basic-auth
        name: password
      username:
        key: prometheus-basic-auth
        name: username
# ...
```

gianrubio:
**Warning: This is a breaking change!!!**

The `key` and `name` were in a inverse (wrong) order.  Before applying this patch edit your `ServiceMonitor` and swap your `key` and `name` values of `basicAuth`.

fix #402

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -880,19 +880,19 @@ func (c *Operator) loadBasicAuthSecrets(mons map[string]*v1alpha1.ServiceMonitor
 
 				for _, secret := range s.Items {
 
-					if secret.Name == ep.BasicAuth.Username.Key {
+					if secret.Name == ep.BasicAuth.Username.Name {
 
-						if u, ok := secret.Data[ep.BasicAuth.Username.Name]; ok {
+						if u, ok := secret.Data[ep.BasicAuth.Username.Key]; ok {
 							username = string(u)
 						} else {
-							return nil, fmt.Errorf("Secret password of servicemonitor %s not found.")
+							return nil, fmt.Errorf("Secret password of servicemonitor %s not found.", mon.Name)
 						}
 
 					}
 
-					if secret.Name == ep.BasicAuth.Password.Key {
+					if secret.Name == ep.BasicAuth.Password.Name {
 
-						if p, ok := secret.Data[ep.BasicAuth.Password.Name]; ok {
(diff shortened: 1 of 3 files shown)


brancz:
@coreosbot test this please

brancz:
@archSeer can you confirm that this works? I don't think this is a breaking change, as it does not work as intended, but it does need to be mentioned in release notes.

pkg/prometheus/operator.go
```diff
@@ -902,11 +902,16 @@ func (c *Operator) loadBasicAuthSecrets(mons map[string]*v1alpha1.ServiceMonitor
 					}
 				}
 
-				secrets[fmt.Sprintf("%s/%s/%d", mon.Namespace, mon.Name, i)] =
-					BasicAuthCredentials{
-						username: username,
-						password: password,
-					}
+				if username == "" || password == "" {
```
brancz: actually I think basic auth can just be a username or password

gianrubio: Should I change from "or" to "and"?

brancz: Yes I think that makes more sense. As that sounds like a configuration issue from the administrator using the Prometheus Operator.

gianrubio: done!

brancz:
@coreosbot test this please

brancz:
lgtm, thanks a lot!

PR #414, opened 2017-06-12
*: update default versions and example usages

brancz:
This updates the Prometheus and Alertmanager versions we default to and is used in example manifests to the latest releases.

@fabxc @mxinden


fabxc:
👍

mxinden:
Looks gold to me! Thanks.

PR #417, opened 2017-06-13
fix alert rule bug

WIZARD-CXY:
happend to see this bug, see ref https://github.com/coreos/etcd/pull/8081


brancz:
Thanks a lot! lgtm

Yes they are the same rules right now, we are working on solutions to properly sync/vendor rule files.

PR #418, opened 2017-06-13
cut v0.10.1

brancz:
@fabxc @mxinden @Gouthamve


mxinden:
Looks good to me!

gianrubio:
@brancz suggestion for the next release, link the PR with the release notes.

brancz:
I never felt like users cared much about that (we don't do it in Prometheus either). We usually just write things in the release notes that is relevant to a user, and they tend to not care what PR that involved, but I'm happy to add it if wanted.

PR #419, opened 2017-06-14
Allow overriding configmap images

crandles:
Allow overriding all images the operator uses and adds additional "PullPolicy" flags for configmap images.

### DIFF
--- a/helm/prometheus-operator/README.md
+++ b/helm/prometheus-operator/README.md
@@ -77,6 +77,8 @@ The following tables lists the configurable parameters of the prometheus-operato
 
 Parameter | Description | Default
 --- | --- | ---
+`configmapReload.repository` | configmap-reload image | `quay.io/coreos/configmap-reload`
+`configmapReload.tag` | configmap-reload tag | `v0.0.1`
 `global.hyperkube.repository` | Hyperkube image | `quay.io/coreos/hyperkube`
 `global.hyperkube.tag` | Hyperkube image tag | `v1.6.2_coreos.0`
 `global.hyperkube.pullPolicy` | Hyperkube image pull policy | `IfNotPresent`
@@ -87,6 +89,8 @@ Parameter | Description | Default
 `kubeletService.namespace` | The namespace in which the kubelet service should be created | `kube-system`
 `kubeletService.name` | The name of the kubelet service to be created | `kubelet`
 `nodeSelector` | Node labels for pod assignment | `{}`
+`prometheusConfigReloader.repository` | prometheus-config-reloader image | `quay.io/coreos/prometheus-config-reloader`
+`prometheusConfigReloader.tag` | prometheus-config-reloader tag | `v0.0.1`
 `rbacEnable` | If true, create & use RBAC resources | `true`
 `resources` | Pod resource requests & limits | `{}`
 `sendAnalytics` | Collect & send anonymous usage statistics | `true`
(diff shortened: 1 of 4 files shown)


crandles:
I'm not sure its necessary to add the pullPolicy flags, but I wanted to allow the same fields to be modified for all images in the values.yaml

mgoodness:
I have a few comments on the Helm chart-related changes, but will hold off until @brancz has a chance to weigh in.

brancz:
I'm not sure I understand why this would be useful aside from debugging. The tag is bumped whenever anything changes in the respective images, which are all controlled by us, so I don't see the necessity to adapt this configuration.

Generally the point of the Prometheus Operator is so one does not have to think about details like this, but I'm happy to be convinced with a good use case :slightly_smiling_face: .

crandles:
I can drop the PullPolicy, I don't have a specific need for that. I wasn't sure about adding it in the first place.

What I need from the PR is the option to set the configmap images, so I can use my own internal docker hub hosted images.

crandles:
Alternatively, those images could be configured via the AlertmanagerSpec/PrometheusSpec, instead of the current prometheus-operator's CLI options. Would that be a better solution for this?

brancz:
The idea is that those images are merely utility, so we assume that they will be the same for all Pods, we didn't feel it was just as strong of an argument for the Prometheus/Alertmanager images.

In that sense as setting all images to custom repos is already possible, I'd suggest to revert the changes to the operator and just add it to the helm chart.

brancz:
Now it's up to @mgoodness :slightly_smiling_face:

helm/prometheus-operator/README.md
```diff
@@ -80,6 +80,10 @@ Parameter | Description | Default
 `global.hyperkube.repository` | Hyperkube image | `quay.io/coreos/hyperkube`
 `global.hyperkube.tag` | Hyperkube image tag | `v1.6.2_coreos.0`
 `global.hyperkube.pullPolicy` | Hyperkube image pull policy | `IfNotPresent`
+`global.prometheus-config-reloader.repository` | prometheus-config-reloader image | `quay.io/coreos/prometheus-config-reloader`
+`global.prometheus-config-reloader.tag` | prometheus-config-reloader tag | `v0.0.1`
+`global.configmap-reload.repository` | configmap-reload image | `quay.io/coreos/configmap-reload`
+`global.configmap-reload.tag` | configmap-reload tag | `v0.0.1`
 `image.repository` | Image | `quay.io/coreos/prometheus-operator`
```
mgoodness: Let's "de-globalize" these. I'll be doing the same for the hyperkube values.

Also, please stick to camelCase.

helm/prometheus-operator/templates/deployment.yaml
```diff
@@ -25,6 +25,8 @@ spec:
           {{- if .Values.kubeletService.enable }}
             - --kubelet-service={{ .Values.kubeletService.namespace }}/{{ .Values.kubeletService.name }}
           {{- end }}
+            - --prometheus-config-reloader={{ .Values.global.prometheus-config-reloader.repository }}:{{ .Values.global.prometheus-config-reloader.tag }}
+            - --config-reloader-image={{ .Values.global.configmap-reload.repository }}:{{ .Values.global.configmap-reload.tag }}
```
mgoodness: Drop `.global` and make camelCase.

helm/prometheus-operator/values.yaml
```diff
@@ -6,6 +6,16 @@ global:
...
+  ## Prometheus-config-reloader image to use for config and rule reloading
+  prometheus-config-reloader:
+    repository: quay.io/coreos/prometheus-config-reloader
+    tag: v0.0.1
+
+  ## Configmap-reload image to use for reloading configmaps
+  configmap-reload:
+    repository: quay.io/coreos/configmap-reload
+    tag: v0.0.1
+
```
mgoodness: Move out of `global` and make camelCase.

mgoodness:
Pushed to opsgoodness repo.

PR #423, opened 2017-06-15
Revert "alerting rules: replace severity with action"

brancz:
Reverts coreos/prometheus-operator#422

@fabxc @mxinden @gouthamve

### DIFF
--- a/contrib/kube-prometheus/assets/prometheus/rules/alertmanager.rules
+++ b/contrib/kube-prometheus/assets/prometheus/rules/alertmanager.rules
@@ -4,7 +4,7 @@ ALERT AlertmanagerConfigInconsistent
        label_replace(prometheus_operator_alertmanager_spec_replicas, "service", "alertmanager-$1", "alertmanager", "(.*)") != 1
   FOR 5m
   LABELS {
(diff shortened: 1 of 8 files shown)


mikebryant:
Just going through pulling in the latest changes of kube-prometheus - I'm curious, why did this get reverted? Seems like a nice change

brancz:
There was a discussion about severity handling in the Prometheus upstream team, and while we could agree to some degree that an action is more appropriate than a meaningless level, things like `ticket` are not easily implementable as there are no native integrations for systems to handle these things nor are there plans to implement them into Alertmanager (for maintenance sake). Therefore the decision was that for examples we should stick with the severity levels to at least document an indication, but this is absolutely something you can adapt to fit your process.

mikebryant:
I see, thanks, that's helpful :)

brancz:
The takeaway is, that users add their opinion on what to do on a severity level, which is often hard to decide, as warning or critical could both mean a page to a user. Therefore I do recommend thinking your personal strategy through and changing the labels and/or add this opintion/meaning through the Alertmanager configuration.

PR #424, opened 2017-06-16
Update AM on relevant changes.

Gouthamve:
This is similar to how Prometheus is updated.

@brancz @fabxc @mxinden

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -16,6 +16,7 @@ package alertmanager
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


brancz:
@coreosbot please test this

pkg/alertmanager/operator.go
```diff
@@ -473,6 +482,22 @@ func AlertmanagerStatus(kclient *kubernetes.Clientset, a *v1alpha1.Alertmanager)
...
 
+func needsUpdate(pod *v1.Pod, tmpl v1.PodTemplateSpec) bool {
+	c1 := pod.Spec.Containers[0]
+	c2 := tmpl.Spec.Containers[0]
+
+	if c1.Image != c2.Image {
+		return true
+	}
+
+	// TODO: Prometheus TPR checks on Args which seems to be the sane thing to do.
```
fabxc: Yea, we should probably do that right away.

brancz: just fyi this will cause a redeploy on Prometheus Operator upgrades, but I'm fine with switching form command to args .. not even sure why we didn't do that in the first place

brancz:
@coreosbot test this please

Gouthamve:
Hmm, this failed for `TestExposingPrometheusWithIngress` and `TestExposingAlertmanagerWithIngress` which I don't think are relevant to this change. Is this a flake?

brancz:
Yes unfortunately Ingress tests flake every now and then. We identified the problem, but won't be able to fix it until we switch to using the tectonic-installer to bring up the e2e test cluster. As all other tests passed (most notably the version upgrade test) this lgtm.

PR #425, opened 2017-06-16
prometheus: configure namespaces to actually necessary namespaces

### LINKED ISSUE #39: Namespace constraints
The Prometheus TPR includes ServiceMonitors via label selection. It only selects ServiceMonitors that are in the same namespace as the Prometheus TPR. That is aligned with semantics of other Kubernetes resources and there seems little reason to diverge from that.

ServiceMonitors select Services by label selection. These label selection work across all namespaces. There are very valid use cases for this, such as meta-monitoring. In the general case however, namespaces are expected to provide isolation and I'd expect my setup to not be negatively affected by actions of other people in their own namespaces.

It seems like we have three use cases:
1. Consider services from same namespace as ServiceMonitor (and transitively Prometheus)
2. Consider services from a subset of namespaces, e.g. a prod and dev namespace
3. Consider services from all namespaces, e.g. meta-monitoring

Currently we only have 3.

How do we want to expose the other options to users? Do we select multiple namespaces by simply providing a list of them or do we also allow selecting them by label, e.g. so that I can select all namespaces belonging to my team?
Label selectors are always more dynamic, but they come with a cost of complexity and should be compared to their likely real-world benefit. For example, Alertmanager services are added to Prometheus TPRs by simple namespace/name references rather than label selection.

I think selecting all namespaces by default, as it is now is potentially dangerous and should only be done explicitly. My current idea looks as follows:

```yaml 
kind: ServiceMonitor
metadata:
  name: frontend
  labels:
    tier: frontend
spec:
  # Simple list of namespaces to select services from.
  namespaces: <string array>
  # Explicit enabling of all namespaces.
  allNamespace: <bool>
  # Dynamic selection could be added later relatively seamlessly
  # if there's an actual need?
  namespaceSelector: <label selector>
  # The above are mutually exclusive. If none is provided, default to just selecting within
  # ServiceMonitors namespace?
  
  # Should this be `serviceSelector` then for explicitness?
  # Other Kubernetes resources stick with just `selector` where reasonable.
  selector:
    # label selector of services
  endpoints:
    # ...
```

This is merely one idea. Would be great if anyone can think of something smarter and can verify what I said against their real-world experience.

@matthiasr @grobie @brancz @brian-brazil

brancz:
Through the ServiceMonitors we're able to infer the namespaces the Prometheus instance will need.

This allows us to run Prometheus instances with the RBAC roles set for the namespaces the instance will actually need.

@fabxc @mxinden @Gouthamve

Closes #39

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -111,6 +111,17 @@ func generateConfig(p *v1alpha1.Prometheus, mons map[string]*v1alpha1.ServiceMon
 }
 
 func generateServiceMonitorConfig(m *v1alpha1.ServiceMonitor, ep v1alpha1.Endpoint, i int, basicAuthSecrets map[string]BasicAuthCredentials) yaml.MapSlice {
+	nsel := m.Spec.NamespaceSelector
+	namespaces := []string{}
+	if !nsel.Any && len(nsel.MatchNames) == 0 {
+		namespaces = append(namespaces, m.Namespace)
+	}
+	if !nsel.Any && len(nsel.MatchNames) > 0 {
+		for i := range nsel.MatchNames {
+			namespaces = append(namespaces, nsel.MatchNames[i])
+		}
+	}
+
 	cfg := yaml.MapSlice{
 		{
 			Key:   "job_name",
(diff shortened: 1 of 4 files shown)


fabxc:
👍

brancz:
I realized this breaks compatibility with Prometheus versions pre 1.7. I think this is ok as with the current release (`v0.10.1`), Prometheus can be updated to 1.7 and then upgrade the operator, to have this functionality. That way there is no interruption necessary. We'll have to write this in the release notes though.

wdyt @fabxc ?

PR #427, opened 2017-06-16
Fix relative links in kube-prometheus readme

cjauvin:
This fixes two relative links, probably made invalid by the repo move.


brancz:
Thanks!

PR #428, opened 2017-06-17
kube-prometheus: update serviceMonitorSelector

jordanjennings:
In commit 9cd05a8c39c6ed03688859de92372d93af5cc6e2 the labels on the ServiceMonitors were changed from k8s-apps to k8s-app. This commit makes the Prometheus serviceMonitorSelector match that change.


mxinden:
Looks good to me. Leaving the last decision to @brancz as he recently looked at label standardization across the project.

brancz:
While I think this change is correct, it's odd, because it actually should not work off of current master - but it does.

brancz:
I discovered the cause of this. Thanks for the contribution! Will merge and open an additional PR for the other fixes.

PR #429, opened 2017-06-19
kube-prometheus: fix correct selector

brancz:
When reviewing #428, it seemed odd that before and after #428 the Prometheus instance had the correct configuration.

It turns out that we never had the correct `serviceMonitorSelector`, however, if a `LabelSelector` is empty it selects everything, which is why it still worked. Because we used a wrong key name this previous selector was parsed to be an empty and thus an "everything" selector.

Changing the selector to the correct one caused the Prometheus and Alertmanagers not to be configured correctly, but the labels just had to be adapted.

@mxinden @fabxc @gouthamve


brancz:
Grafana is not a scrape target, I don't even think it exposes any metrics, at least not in Prometheus format.

`example/node-exporter-monitor.yaml` doesn't seem to be used anywhere, I'll remove it.

PR #432, opened 2017-06-20
Add recomendation to install operator before alertmanager/prometheus

gianrubio:
Because the `thirdypartyresource` is not yet installed, you got an error when you are trying to install  prometheus/alermanager.

```
$ helm install alertmanager --name alertmanager --debug
[debug] Created tunnel using local port: '57662'

[debug] SERVER: "localhost:57662"

[debug] CHART PATH: /Users/grubio/work/src/github.com/coreos/prometheus-operator/helm/alertmanager

Error: apiVersion "monitoring.coreos.com/v1alpha1" in alertmanager/templates/alertmanager.yaml is not available
```

### DIFF
--- a/helm/alertmanager/README.md
+++ b/helm/alertmanager/README.md
@@ -14,6 +14,7 @@ This chart bootstraps an [Alertmanager](https://github.com/prometheus/alertmanag
 
 ## Prerequisites
   - Kubernetes 1.4+ with Beta APIs & ThirdPartyResources enabled
+  - [Prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
 
 ## Installing the Chart
 
--- a/helm/alertmanager/README.md
+++ b/helm/alertmanager/README.md
@@ -14,7 +14,7 @@ This chart bootstraps an [Alertmanager](https://github.com/prometheus/alertmanag
 
 ## Prerequisites
   - Kubernetes 1.4+ with Beta APIs & ThirdPartyResources enabled
-  - [Prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
+  - [prometheus-operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
 
 ## Installing the Chart
 
--- a/helm/prometheus/README.md
+++ b/helm/prometheus/README.md
@@ -14,6 +14,7 @@ This chart bootstraps a [Prometheus](https://github.com/prometheus/prometheus) d
 
 ## Prerequisites
   - Kubernetes 1.4+ with Beta APIs & ThirdPartyResources enabled
+  - [Prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
 
 ## Installing the Chart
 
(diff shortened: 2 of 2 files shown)


brancz:
lgtm, I'll leave the final call up to @mgoodness .

Wouldn't it be great if helm had the ability to specify dependencies?

helm/alertmanager/README.md
```diff
@@ -4,6 +4,8 @@ Installs a [Prometheus](https://prometheus.io) Alertmanager instance using the C
 
 ## TL;DR;
 
+Before install alertmanager you should install [prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
```
brancz: Before install -> Before installing

helm/prometheus/README.md
```diff
@@ -4,6 +4,8 @@ Installs a [Prometheus](https://prometheus.io) instance using the CoreOS [promet
 
 ## TL;DR;
 
+Before install prometheus you should install [prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
```
brancz: Before install -> Before installing

gianrubio:
> Wouldn't it be great if helm had the ability to specify dependencies?

I don't know what the goals of this charts, I would like to see this charts on [kubernetes/charts](http://github.com/kubernetes/charts). In that case makes no sense to specify prometheus and alert manager as a dependency of operator.

gianrubio:
I just realise that these charts are not deploying alert manager and prometheus. So it's nice to have this charts as a dependencies.

gianrubio:
Is prometheus-operator a dependencie of alertmanager/prometheus or the opposite? 

Make sense to have prometheus operator without alertmanager/prometheus?

mgoodness:
>Wouldn't it be great if helm had the ability to specify dependencies?

Helm supports dependencies, and my original `prometheus` and `alertmanager` charts included the `prometheus-operator` one. Unfortunately, Helm (currently) has no way to dictate the order of installation, so deployments were unreliable. So now `prometheus-operator` has to be installed first. Just failed to document that fact.

>I would like to see this charts on kubernetes/charts

As one of the maintainers of that repo, I can say we're in favor of upstream projects (like this one) creating & maintaining their own charts. It's easier to keep a chart updated with the latest application functionality when it lives alongside the code. There's also an existing `prometheus` chart that doesn't use the operator.

>Is prometheus-operator a dependencie of alertmanager/prometheus or the opposite?

prometheus-operator must be installed (by the chart or some other method) before the `alertmanager` or `prometheus` charts can be deployed.

>Make sense to have prometheus operator without alertmanager/prometheus?

Yes. We at Ticketmaster will use this `prometheus-operator` chart, but maintain our own `alertmanager` and `prometheus` charts because of customization.

helm/alertmanager/README.md
```diff
@@ -4,6 +4,8 @@ Installs a [Prometheus](https://prometheus.io) Alertmanager instance using the C
 
 ## TL;DR;
 
+Before installing alertmanager you should install [prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
+
```
mgoodness: There is a `Prerequisites` section in the README. Just add the `prometheus-operator` chart link as a second bullet.

helm/prometheus/README.md
```diff
@@ -4,6 +4,8 @@ Installs a [Prometheus](https://prometheus.io) instance using the CoreOS [promet
 
 ## TL;DR;
 
+Before installing prometheus you should install [prometheus operator](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/README.md).
+
```
mgoodness: See above comment.

helm/alertmanager/README.md
gianrubio: done

PR #433, opened 2017-06-20
Cleanup thirdypartyresource before deleting the chart

gianrubio:
When you 

1 - install the chart
2- delete the chart 
3- install the chart again

 you got the error below. This job will delete the tpr before deleting the chart.

```
$ helm install prometheus-operator --name prometheus-operator
Error: jobs.batch "prometheus-operator-prometheus-operator-cleanup-tpr" already exists
```

### DIFF
--- a/helm/prometheus-operator/templates/clusterrole.yaml
+++ b/helm/prometheus-operator/templates/clusterrole.yaml
@@ -13,51 +13,35 @@ metadata:
     release: {{ .Release.Name }}
   name: {{ template "fullname" . }}
 rules:
-- apiGroups:
-  - extensions
-  resources:
-  - thirdpartyresources
-  verbs:
-  - "*"
-- apiGroups:
-  - apiextensions.k8s.io
-  resources:
-  - customresourcedefinitions
-  verbs:
-  - "*"
-- apiGroups:
-  - monitoring.coreos.com
-  resources:
-  - alertmanagers
-  - prometheuses
-  - servicemonitors
-  verbs:
-  - "*"
-- apiGroups:
-  - apps
-  resources:
-  - statefulsets
-  verbs: ["*"]
-- apiGroups: [""]
-  resources:
-  - configmaps
-  - secrets
-  verbs: ["*"]
-- apiGroups: [""]
-  resources:
-  - pods
-  verbs: ["list", "delete"]
-- apiGroups: [""]
-  resources:
-  - services
-  - endpoints
-  verbs: ["get", "create", "update"]
-- apiGroups: [""]
-  resources:
-  - nodes
-  verbs: ["list", "watch"]
-- apiGroups: [""]
-  resources:
-  - namespaces
-  verbs: ["list"]
+  - apiGroups: [""]
+    resources: ["configmaps", "secrets"]
+    verbs: ["*"]
+
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["delete", "list"]
+
+  - apiGroups: [""]
+    resources: ["services", "endpoints"]
+    verbs: ["create", "get", "update"]
+
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["list", "watch"]
+
+  - apiGroups: ["batch"]
+    resources: ["jobs"]
+    verbs: ["get", "update"]
+
+  - apiGroups: ["apps"]
+    resources: ["statefulsets"]
+    verbs: ["*"]
+
+  - apiGroups: ["extensions"]
+    resources: ["thirdpartyresources"]
+    verbs: ["create", "get"]
+
+  - apiGroups: ["monitoring.coreos.com"]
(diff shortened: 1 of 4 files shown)


mgoodness:
I wrestled with this in a previous iteration of the chart. Do we really want an external process like this to delete TPRs created by the operator? For example, what if there are multiple prometheus-operators running in the cluster? Deleting one would render the other(s) inoperable. Are there any side effects if a user instantiates a `prometheus`, `alertmanager`, or `servicemonitor` TPR, then deletes the TPRs?

Not opposed, mind you. Just want to make sure this wouldn't cause more problems that it would solve.

Note: if we do go ahead, the job's name needs to be randomized to avoid the `already exists` error. Also need to do that for the existing `get-tprs` job.

brancz:
> I wrestled with this in a previous iteration of the chart. Do we really want an external process to delete TPR types created by the operator? For example, what if there are multiple prometheus-operators running in the cluster? Deleting one would render the other(s) inoperable. Are there any side effects if a user instantiates a prometheus, alertmanager, or servicemonitor TPR, then deletes the TPR types?

Note that this is currently not advisable, because the Prometheus Operator does not perform any leader election or similar, however, because of how the Prometheus Operator works, it might still work, but there is no guarantee.

gianrubio:
> For example, what if there are multiple prometheus-operators running in the cluster

When I built this job I didn't realise this scenario, what about an argument to cleanup this `tpr`?

> Note: if we do go ahead, the job's name needs to be randomized to avoid the already exists error. Also need to do that for the existing get-tprs job.

I don't like this idea to randomise names, looks like there's a fix https://github.com/kubernetes/helm/issues/1769#issuecomment-309981012 coming on helm, so we can wait.

mgoodness:
>>For example, what if there are multiple prometheus-operators running in the cluster

>When I built this job I didn't realise this scenario, what about an argument to cleanup this tpr?

Yeah, let's do a bool `cleanupOnDelete`.

>>Note: if we do go ahead, the job's name needs to be randomized to avoid the already exists error. Also need to do that for the existing get-tprs job.

>I don't like this idea to randomise names, looks like there's a fix kubernetes/helm#1769 (comment) coming on helm, so we can wait.

Fair enough. In the meantime, I'll review what you have here so far.

helm/prometheus-operator/templates/cleanup.yaml
```diff
@@ -0,0 +1,34 @@
...
+kind: Job
+metadata:
+  annotations:
+    helm.sh/hook: pre-delete
+  labels:
+     app: {{ template "name" . }}
+     chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+     heritage: {{ .Release.Service }}
+     release: {{ .Release.Name }}
+  name: {{ template "fullname" . }}-cleanup-tpr
```
mgoodness: Let's name this (and the pod template below) `{{ template "fullname" . }}-delete-tprs` to be consistent with the existing `{{ template "fullname" . }}-get-tprs` Job.

Also, rename the file to `delete-tprs-job.yaml`.

helm/prometheus-operator/templates/cleanup.yaml
```diff
@@ -0,0 +1,34 @@
...
+        release: {{ .Release.Name }}
+      name: {{ template "fullname" . }}-cleanup-tpr
+    spec:
+      containers:
+      - name: hyperkube
+        image: "{{ .Values.global.hyperkube.repository }}:{{ .Values.global.hyperkube.tag }}"
+        imagePullPolicy: "{{ .Values.global.hyperkube.pullPolicy }}"
+        command:
+          - ./kubectl
+          - delete
```
mgoodness: Need `- thirdpartyresources` here.

helm/prometheus-operator/templates/cleanup.yaml
```diff
@@ -0,0 +1,34 @@
...
+          - --ignore-not-found=true
+          - alertmanager.monitoring.coreos.com
+          - prometheus.monitoring.coreos.com
+          - service-monitor.monitoring.coreos.com
+      restartPolicy: OnFailure
+    {{- if .Values.rbacEnable }}
+      serviceAccountName: {{ template "fullname" . }}
+    {{- end }}
```
mgoodness: Wrap the entire file with a `{{- if .Values.cleanupOnDelete -}}` statement.

brancz:
@mgoodness does this now look good to you?

brancz:
@gianrubio could you update this to delete the CRDs instead of the TPRs?

gianrubio:
Sure, I'll do during the weekend!

gianrubio:
@brancz I did the changes, unfortunatelly I don't have a cluster with `crd` enabled so I can't test the deletion.

brancz:
Thanks! I should be able to test this soon, probably not today, but soon. Again, thanks a lot for taking care of this!

PR #434, opened 2017-06-20
create rbac rules for prometheus + bump version

gianrubio:
bump alertmanager, prometheus and operator version

### DIFF
--- a/helm/prometheus/templates/clusterrolebinding.yaml
+++ b/helm/prometheus/templates/clusterrolebinding.yaml
@@ -1,0 +1,24 @@
+{{- if .Values.rbacEnable }}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1beta1" }}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+{{- else if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1alpha1" }}
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+{{- end }}
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app: {{ template "name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: {{ template "fullname" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "fullname" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "fullname" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end }}
+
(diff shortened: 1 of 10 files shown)


helm/prometheus/values.yaml
```diff
@@ -23,7 +23,7 @@ externalUrl: ""
 ##
 image:
   repository: quay.io/prometheus/prometheus
-  tag: v1.5.2
+  tag: v1.7.0
```
brancz: v1.7.1 has been released

brancz:
as always with helm changes, I'm leaving the last call up to @mgoodness

mgoodness:
Are there any functionality changes we can/should include? Image tags/versions can be specified on the command line (`--set image.tag=...`) or via values file, without bumping chart versions.

gianrubio:
@mgoodness I also add the rbac rules for prometheus. 

Do you think it's a good ideia to add rbac rules for alertmanager? For now it's running with the default `serviceAccount` and most of the case it's a root account.

I thought to add an empty `clusterRole` just to isolate alertmanager in the cluster.

helm/prometheus/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: {{ template "fullname" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "fullname" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "fullname" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end }}
```
xiang90: new line?

example/rbac/alertmanager/alertmanager-cluster-role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: alertmanager
+rules:
+- apiGroups: [""]
+  resources:
+  - configmaps
```
brancz: Actually the Alertmanager pods are not accessing the Kubernetes API at all, so there is no need for a ClusterRole at all.Actually the Alertmanager pods are not accessing the Kubernetes API at all, so there is no need for a ClusterRole at all.

gianrubio: Yes, I realised this and removed the rbac for alertmanager.

helm/prometheus/templates/clusterrolebinding.yaml
gianrubio: Done

gianrubio:
> edit2: alertmanager has a config-reloader pod, so I'll add the right rbac rule.

@mgoodness alertmanager doesn't need any rbac rules, because it didn't talk to the k8s api. 

The PR are ready to merge!

mgoodness:
LGTM! Thanks, and sorry it took so long to merge. Published to the opsgoodness repo.

PR #441, opened 2017-06-22
Add alpha3 support

fabxc:


pkg/prometheus/statefulset.go
```diff
@@ -331,6 +331,12 @@ func makeStatefulSetSpec(p v1alpha1.Prometheus, c *Config, ruleConfigMaps []*v1.
 	}
 	promArgs = append(promArgs, "-web.route-prefix="+webRoutePrefix)
 
+	if version.Major == 2 {
+		for i, a := range promArgs {
+			promArgs[i] = "-" + a
```
brancz: this is pretty sketchy, but should work :smile:

brancz:
as 2.0 support is experimental, this should be good, we'll clean up the 2.0 support properly as we go along

brancz:
lgtmog

PR #448, opened 2017-06-27
add serviceMonitor to alertmanager, prometheus-operator and prometheus

gianrubio:
@mgoodness I didn't add the `serviceMonitor`for prometheus-operator because there's a circular dependency.
What do you think it's the right way to do:

1 - Run a job to apply the serviceMonitor after the creation of `tpr` or
2- Add a new chart on kube-prometheus to create the serviceMonitor for prometheus-operator


mgoodness:
I think a post-install hook in the prometheus-operator chart is the cleaner, more consistent approach.

gianrubio:
According to the changes requested from @mgoodness, the prometheus-operator is created by a job.

screen shot 2017-10-02 at 14 25 33

brancz:
lgtm, thanks!

PR #449, opened 2017-06-27
test: Improve error feedback on e2e framework setup

mxinden:
/CC @Gouthamve


pkg/k8sutil/k8sutil.go
```diff
@@ -54,6 +54,11 @@ func WaitForTPRReady(restClient rest.Interface, tprGroup, tprVersion, tprName st
 
 		return true, nil
 	})
+
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("timed out waiting for TPR %s", tprName))
```
brancz: errors.Wrap returns nil if err is nil, so the if is not needed

mxinden: Thanks. Good to know!

brancz:
lgtmog

PR #365, opened 2017-05-11
prometheus: add externalLabels to PrometheusSpec API

crandles:
This allows the prometheus config `external_labels` to be configured via the PrometheusSpec TPR.  This is useful for adding unique identifiers for any federation or alerting.

PR #450, opened 2017-06-27
helm/prometheus: add externalLabels spec to be configured

crandles:
Adds the externalLabel spec option from #365 to the Prometheus helm chart.


brancz:
lgtm, but @mgoodness as always has the last call on helm charts

mgoodness:
Version bumped and pushed to the opsgoodness repo. Thanks for the contribution!

PR #453, opened 2017-06-28
kube-prometheus: ensure triggering alerts on down targets

brancz:
Previously we did not trigger when there is no cluster label and a target is discovered, but not down.

@fabxc @mxinden @Gouthamve


contrib/kube-prometheus/assets/prometheus/rules/kube-apiserver.rules
```diff
@@ -1,5 +1,5 @@
 ALERT K8SApiserverDown
-  IF absent({job="apiserver"}) or (count by(cluster) (up{job="apiserver"} == 1) < count by(cluster) (up{job="apiserver"}))
+  IF absent({job="apiserver"} == 1)
```
fabxc: Is the `up` missing here?

PR #454, opened 2017-06-28
vendor: add backoff

brancz:
@mxinden @fabxc @Gouthamve 

Without this our unit tests will fail.


mxinden:
@brancz Should we add a vendor job to the CI that checks whether everything in the vendor folder matches `vendor.json`?

fabxc:
👍

brancz:
@mxinden yes we should at some point, but it is more easily said than done :slightly_smiling_face:

PR #456, opened 2017-06-28
test: Bring up k8s cluster via tectonic-installer

mxinden:
This PR replaces the previous Kops code with the Tectonic installer. 
* We are using current master of the Tectonic installer. If this breaks too much we can switch to stable releases. But this way we make sure we are compatible with the upcoming release.
* From now on we archive artefacts like the `tfstate` files for better debugging and to be able to clean up in case it is not properly done by Jenkins.
* We access everything inside the cluster via `kubeproxy`. So all Ingress and Nodeport tests are removed. @brancz Let me know if you want to add something here in the long run.
* We test the Prometheus operator inside a k8s cluster with RBAC enabled.


mxinden:
@brancz Do you have time to look at this? Otherwise I can ask Carlos as well. I am a little worried having `master` failing for such a long time. That would be fixed with this PR. Sorry for the Ping.

brancz:
Thanks everyone!

mxinden:
@cpanato Thanks!

PR #457, opened 2017-06-28
grafana watcher example: Remove imagePullPolicy

discordianfish:
Maybe I miss a reason for that being there but probably just cruft?


brancz:
Yeah I think this was just in place for development purposes. Thanks @discordianfish !

PR #458, opened 2017-06-28
test/e2e: add test for zero downtime alertmanager deploys

brancz:
The way we perform rolling updates of the Alertmanager right now can't 100% ensure that there is no data loss, so we rely on the Alertmanager mesh to synchronize correctly on deploy. While this mechanism can and should be improved further, it shows that our rolling update does usually work without data loss.

This test ensures that there are no duplicate notifications are sent, while 3 Prometheus instances are firing the same alert every 100ms. The count is retrieved from the number of calls the webhook server the Alertmanagers fire against.

Before having written this test I felt like we may be in a worse position than we are, but having written this and having run it a lot of times I'm a lot more confident.

The example webhook image is currently under `quay.io/brancz/alertmanager-webhook-example`, let me know if you think this is ok, or whether you want me to push it to some coreos org repository.

@fabxc @mxinden @Gouthamve

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -17,12 +17,16 @@ package e2e
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"testing"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
 
-	"github.com/coreos/prometheus-operator/pkg/client/monitoring/v1alpha1"
 	testFramework "github.com/coreos/prometheus-operator/test/e2e/framework"
 )
 
@@ -130,17 +134,8 @@ func TestMeshInitialization(t *testing.T) {
 	ns := ctx.CreateNamespace(t, framework.KubeClient)
 	ctx.SetupPrometheusRBAC(t, ns, framework.KubeClient)
 
-	var amountAlertmanagers int32 = 3
-	alertmanager := &v1alpha1.Alertmanager{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "test",
-		},
-		Spec: v1alpha1.AlertmanagerSpec{
-			Replicas: &amountAlertmanagers,
-			Version:  "v0.7.1",
-		},
-	}
-
+	amClusterSize := 3
+	alertmanager := framework.MakeBasicAlertmanager("test", int32(amClusterSize))
 	alertmanagerService := framework.MakeAlertmanagerService(alertmanager.Name, "alertmanager-service", v1.ServiceTypeClusterIP)
 
 	if err := framework.CreateAlertmanagerAndWaitUntilReady(ns, alertmanager); err != nil {
@@ -151,9 +146,9 @@ func TestMeshInitialization(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	for i := 0; i < int(amountAlertmanagers); i++ {
+	for i := 0; i < amClusterSize; i++ {
 		name := "alertmanager-" + alertmanager.Name + "-" + strconv.Itoa(i)
-		if err := framework.WaitForAlertmanagerInitializedMesh(ns, name, int(amountAlertmanagers)); err != nil {
+		if err := framework.WaitForAlertmanagerInitializedMesh(ns, name, amClusterSize); err != nil {
 			t.Fatal(err)
 		}
 	}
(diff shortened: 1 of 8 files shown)


test/e2e/alertmanager_test.go
```diff
@@ -291,3 +285,186 @@ receivers:
...
+					Port:       5001,
+					TargetPort: intstr.FromString("web"),
+				},
+			},
+			Selector: map[string]string{
+				"app": "alertmanager-webhook",
+			},
+		},
+	}
+	if _, err := framework.KubeClient.CoreV1().Services(ns).Create(whsvc); err != nil {
```
mxinden: Can you use `CreateServiceAndWaitUntilReady` here?

test/e2e/alertmanager_test.go
```diff
@@ -291,3 +285,186 @@ receivers:
...
+		t.Fatal(err)
+	}
+	if _, err := framework.KubeClient.ExtensionsV1beta1().Deployments(ns).Create(whdpl); err != nil {
```
mxinden: Same here, can we use `CreateDeployment`. It might not add a lot of convenience at the moment, but maybe we want to add more *clean up* functionality in the long run, which would be easier when we use these functions everywhere. What do you think?

test/e2e/alertmanager_test.go
```diff
@@ -291,3 +285,186 @@ receivers:
...
+		t.Fatal(err)
+	}
+	err := testFramework.WaitForPodsReady(framework.KubeClient, ns, 5*time.Minute, 1,
```
mxinden: Is `5*time.Minute` just a guess or evolved by best experience. If it is the former, can we use `framework.DefaultTimeout` here?

test/e2e/alertmanager_test.go
```diff
@@ -291,3 +285,186 @@ receivers:
...
+	}
+	if _, err := framework.MonClient.Alertmanagers(ns).Create(alertmanager); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := testFramework.CreateServiceAndWaitUntilReady(framework.KubeClient, ns, amsvc); err != nil {
+		t.Fatal(err)
+	}
+
+	p := framework.MakeBasicPrometheus(ns, "test", "test", 3)
+	framework.AddAlertingToPrometheus(p, ns, alertmanager.Name)
```
mxinden: Instead of `alertmanager.Name`, shouldn't this be `amsvc.Name`? I know in the end it is the same string, but just in case we change the Alertmanager name, but not the Alertmanager service.

test/e2e/alertmanager_test.go
```diff
@@ -291,3 +285,186 @@ receivers:
...
+		t.Fatal(err)
+	}
+	if _, err := framework.MonClient.Alertmanagers(ns).Create(alertmanager); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := testFramework.CreateServiceAndWaitUntilReady(framework.KubeClient, ns, amsvc); err != nil {
+		t.Fatal(err)
+	}
+
+	p := framework.MakeBasicPrometheus(ns, "test", "test", 3)
```
mxinden: Shouldn't we set the new `evaluationInterval` parameter here to `100ms`?

brancz: done! nicely spotted, let's hope it still works :)

test/e2e/alertmanager_test.go
brancz: turns out default wasn't enough, had to bump it again

mxinden:
@brancz I am sorry for delaying my review for so long. I would like to get #456 in first, so we don't have to tune timeouts twice.

mxinden:
@brancz Whenever you have time, can you rebase to current master to get changes introduced via #456?

brancz:
@mxinden rebased. Thanks for the ping! I forgot about this.

test/e2e/alertmanager_test.go
```diff
@@ -231,3 +226,187 @@ receivers:
 		t.Fatal(err)
 	}
 }
+
+func TestAlertmanagerZeroDowntimeRollingDeployment(t *testing.T) {
+	t.Parallel()
+
+	ctx := framework.NewTestCtx(t)
+	defer ctx.Cleanup(t)
+	ns := ctx.CreateNamespace(t, framework.KubeClient)
```
mxinden: @brancz Can you add the PrometheusRBAC setup just like I did [here](https://github.com/coreos/prometheus-operator/blob/master/test/e2e/prometheus_test.go#L46)? Switching to the tectonic-installer forces us to care about RBAC initialization.

mxinden: This is the reason the tests are failing right now. The Prometheus instance can not be started, because the Prometheus service account is not present in the test namespace.

brancz: Got it, thanks for the heads up @mxinden !

brancz:
@mxinden can you give this another go?

test/e2e/alertmanager_test.go
```diff
@@ -231,3 +226,188 @@ receivers:
...
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app": "alertmanager-webhook",
+					},
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:  "webhook-server",
+							Image: "quay.io/brancz/prometheus-alertmanager-test-webhook",
```
mxinden: @brancz Can you move this to a CoreOS repo as suggested in your PR description. I think that would be easier in terms of managing permissions in case I need to push a new image.

PR #459, opened 2017-06-28
Remove scrape: true annotation

discordianfish:
This is probably cruft, grafana has no /metrics endpoint.


brancz:
Absolutely right! Thanks again! :)

PR #460, opened 2017-06-28
Move to CRDs

Gouthamve:
This works when you launch a new PO in a 1.7 cluster.

TODO:
- [x] Add compat to older k8s versions by creating TPRs when required.
- [x] Add a migration helper and look at different strategies.

cc @brancz @fabxc @mxinden @nikhita


mxinden:
@coreosbot  test this please

pkg/k8sutil/k8sutil.go
```diff
@@ -156,3 +158,14 @@ func CreateOrUpdateEndpoints(eclient clientv1.EndpointsInterface, eps *v1.Endpoi
 
 	return nil
 }
+
+// GetMinorVersion returns the minor version as an integer
+func GetMinorVersion(kclient *kubernetes.Clientset) (int, error) {
```
brancz: let's just pass the discovery client here, as we don't need the rest

brancz:
@coreosbot test this please

pkg/prometheus/operator.go
```diff
@@ -1037,9 +1058,59 @@ func (c *Operator) createTPRs() error {
 	}
 
 	// We have to wait for the TPRs to be ready. Otherwise the initial watch may fail.
```
brancz: /TPR/CRD/

pkg/client/monitoring/v1alpha1/alertmanager.go
```diff
@@ -26,8 +28,8 @@ import (
 )
 
 const (
-	TPRAlertmanagersKind = "Alertmanager"
-	TPRAlertmanagerName  = "alertmanagers"
+	CRDAlertmanagersKind = "Alertmanager"
```
brancz: We should rethink the naming here as we're using it in CRD and TPR cases, maybe just "AlertmanagerKind". Same goes for all other functions, vars, etc we should get rid of the TPR/CRD in the naming if it is not actually specific to one or the other.

For example "CreateTPRs" and "CreateCRDs" are both valid namings, but where we use things in both contexts, they should be more generic.

pkg/prometheus/operator.go
Gouthamve: This is the createTPRs function, the createCRDs function has its appropriate name.

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,368 @@
...
+	m.restClient = restClient
+	m.tprClient = tprClient
+	m.cmClient = cmClient
+	m.nsClient = kclient.CoreV1().Namespaces()
+
+	return m, nil
+}
+
+func (m *Migrator) migrateTPR(tprKind, tprName, tprOldReg, configmap string, shortNames []string) error {
+	m.logger.Log("msg", fmt.Sprintf("migrating TPR \"%s\"", tprKind))
```
fabxc: `%q` is the equivalent for `\"%s\"`.

cmd/operator/main.go
```diff
@@ -99,6 +105,30 @@ func Main() int {
...
+		fmt.Fprint(os.Stderr, err)
+		return 1
+	}
+	if mv >= 7 {
+		// See if the TPRs managed by this operator exist.
+		tprClient := client.Extensions().ThirdPartyResources()
+		_, err = tprClient.Get("prometheus."+v1alpha1.Group, metav1.GetOptions{})
+		if err == nil {
+			migrator.MigrateTPR2CRD(conf, logger.With("component", "migrator"))
+		}
```
fabxc: This probably needs to become a bit more reliable before merging.

pkg/client/monitoring/v1alpha1/client.go
```diff
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	TPRGroup   = "monitoring.coreos.com"
-	TPRVersion = "v1alpha1"
+	Group   = "monitoring.coreos.com"
+	Version = "v1alpha1"
```
fabxc: Ah, it's in your TODO.

brancz:
@Gouthamve can you update me on the status of this?

Gouthamve:
@coreosbot test this please

brancz:
@coreosbot test this please

Gouthamve:
@brancz This took way longer than I expected but now the tests pass. Can you take a look now?

I will first document what happens and how rollback happens and incase of a failed rollback how to restore state. I want to tackle `v1alpha1` --> `v1` in another PR if possible.

brancz:
@Gouthamve no worries. Almost everything always takes longer than one expects. I'm fine with separating in multiple PRs. I'll give this a review as soon as I can.

test/e2e/prometheus_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 The prometheus-operator Authors
+/// Copyright 2016 The prometheus-operator Authors
```
brancz: extra "/"

test/migration/migration_test.go
```diff
@@ -0,0 +1,201 @@
...
+	// Delete and launch new operator.
+	if err := framework.
+		KubeClient.
+		Extensions().
+		Deployments(framework.Namespace.Name).
+		Delete("prometheus-operator", nil); err != nil {
+		t.Fatal(err)
+	}
+
+	// TODO: Wait until terminated.
```
brancz: This one would be good if we can just put a `watch.Poll` in here

pkg/migrator/migrator.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	for _, ns := range nsList.Items {
+		resClient := m.dynClient.Resource(
+			&metav1.APIResource{
+				Kind:       tprKind,
+				Name:       tprName,
+				Namespaced: true,
+			},
+			ns.ObjectMeta.Name,
```
Gouthamve: Yes, for `DELETE`, it fails if namespace is `""`.

pkg/k8sutil/k8sutil.go
```diff
@@ -18,22 +18,25 @@ import (
...
 	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/rest"
 )
 
-// WaitForTPRReady waits for a third party resource to be available
+// WaitForCRDReady waits for a third party resource to be available
 // for use.
-func WaitForTPRReady(restClient rest.Interface, tprGroup, tprVersion, tprName string) error {
+// TODO(gouthamve): Move to clientset.Get()
```
Gouthamve: clientset.Get() is a CRD specific implementation. See here: https://github.com/Gouthamve/prometheus-operator/blob/crd/pkg/migrator/migrator.go#L219-L237

This should ideally be done once we deprecate 1.6 support.

brancz: Sounds good. Deprecating 1.6 is a long way to go :smile:, I'm guessing we'll have to keep 1.6 support until 1.9 is released.

pkg/migrator/migrator.go
brancz: Got it. Makes sense. I was just hoping to not require the additional RBAC role for namespaces, but sounds reasonable.

pkg/k8sutil/k8sutil.go
```diff
@@ -157,3 +160,18 @@ func CreateOrUpdateEndpoints(eclient clientv1.EndpointsInterface, eps *v1.Endpoi
 
 	return nil
 }
+
+// GetMinorVersion returns the minor version as an integer
+func GetMinorVersion(dclient discovery.DiscoveryInterface) (int, error) {
+	v, err := dclient.ServerVersion()
+	if err != nil {
+		return 0, nil
```
brancz: Why not propagate the error? Maybe we should even crash if we can't find this out.

Gouthamve: True, I don't remember why I ignored it before. Fixed it. And we crash if we fail to get the `MinorVersion` now.

Also, hashicorp's semver is the only one that can parse but it failed on `1.7.0-alpha.2` on minikube (the version returned was garbage anyways like: 0.0.0-master-$Format$ or something). Should we be strict about that too, because compiling k8s from source might produce garbage

brancz: I think it's fine to rely on properly compiled Kubernetes, and crash if it's not. From your comment I read that on latest minikube releases it works?

Gouthamve: Yes. It works from rc and up.

brancz:
@coreosbot test this please.

Once this is green it lgtm. Nice work @Gouthamve !

test/migration/migration_test.go
```diff
@@ -0,0 +1,219 @@
...
+		KubeClient.
+		Extensions().
+		Deployments(framework.Namespace.Name).
+		Delete("prometheus-operator", &metav1.DeleteOptions{
+			PropagationPolicy: &deletePolicy,
+		}); err != nil {
+		t.Fatal(err)
+	}
+
+	wait.Poll(time.Second, 3*time.Minute, func() (bool, error) {
```
mxinden: 👍 for the apierrors

Can you extract this into a framework function `WaitUntilDeploymentGone` in https://github.com/coreos/prometheus-operator/blob/master/test/e2e/framework/deployment.go

mxinden: If `3*time.Minute` is not specifically tuned, can you use `framework.DefaultTimeOut`? This helps us in the future to be able to adjust the timeouts in one place for slow environments.

test/framework/framework.go
```diff
@@ -136,17 +137,17 @@ func (f *Framework) setupPrometheusOperator(opImage string) error {
 	}
 	f.OperatorPod = &pl.Items[0]
 
-	err = k8sutil.WaitForTPRReady(f.KubeClient.Core().RESTClient(), v1alpha1.TPRGroup, v1alpha1.TPRVersion, v1alpha1.TPRPrometheusName)
+	err = k8sutil.WaitForCRDReady(f.KubeClient.Core().RESTClient(), v1alpha1.Group, v1alpha1.Version, v1alpha1.PrometheusName)
```
mxinden: In the migration tests we are spinning up an old prometheus-operator, right? So we should either wait for TPRs if it is the old operator, or wait for CRDs if it is the new operator?

Gouthamve: Well I just renamed the function and the function does basically the same as the old function.

test/migration/migration_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	// Compare old and new objects.
+	obj, err = monClient.Prometheuses(ns2).List(metav1.ListOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	promsNew := obj.(*v1alpha1.PrometheusList)
+	if len(promsNew.Items) != len(proms.Items) {
+		t.Fatalf("expected %d proms, got %d", len(proms.Items), len(promsNew.Items))
```
Gouthamve: As in write the proms object? That would be a huge struct.

test/migration/migration_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	for i, prom := range proms.Items {
+		if !reflect.DeepEqual(prom.Spec, promsNew.Items[i].Spec) {
+			t.Fatalf("The prometheus object changed %d", i)
```
Gouthamve: You are right. What would be the best approach here?

test/migration/migration_test.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func TestMigration(t *testing.T) {
+	// Create all the clients required.
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	kclient, err := kubernetes.NewForConfig(config)
```
Gouthamve: I am doing it as I need to check the version to skip the test. Should I launch the operator and then check? I moved them monClient though.

mxinden: @Gouthamve Ah, OK. Can you add a comment above the line so I remember in the future?

test/migration/migration_test.go
mxinden: @Gouthamve Does the Prometheus operator log something when he is done with the transition? If so, we can just watch the logs for that?
https://github.com/coreos/prometheus-operator/blob/master/test/e2e/prometheus_test.go#L291

test/framework/framework.go
mxinden: 👍

test/migration/migration_test.go
mxinden: Let me rephrase that: If I understand correctly `proms` is a Prometheus. If so, in the error string, can you write "expected %s Prometheus, got %s"? Sorry for the confusion.

mxinden: ^^ @Gouthamve

brancz:
Awesome @Gouthamve ! Fantastic work!

mxinden:
I am not sure, that we can trust the test suite (see [comment](https://github.com/coreos/prometheus-operator/pull/460#discussion_r129548730)). @Gouthamve Do you have some spare time to fix this?

test/migration/migration_test.go
Gouthamve: I think the comments are lacking. Will send in a PR with more comments and maybe breaking it into smaller ones today.

mxinden: @Gouthamve I see, thanks for clarifying!

PR #462, opened 2017-06-29
kube-prometheus: Update RBAC for kube-state-metrics

jordanjennings:
After installing kube-prometheus from hack/cluster-monitoring/self-hosted-deploy I noticed that the API server logs were getting flooded with RBAC DENY messages for kube-state-metrics:

```
I0629 16:36:49.945856       5 rbac.go:87] RBAC DENY: user "system:serviceaccount:monitoring:kube-state-metrics" groups [system:serviceaccounts system:serviceaccounts:monitoring system:authenticated] cannot "list" resource "replicationcontrollers" cluster-wide
```

Looks like kube-state-metrics needs additional permissions for replicationcontrollers, limitranges, and services. Once I added those permissions the messages went away and kube-state-metrics seems to be happy.

### DIFF
--- a/contrib/kube-prometheus/manifests/kube-state-metrics/kube-state-metrics-cluster-role.yaml
+++ b/contrib/kube-prometheus/manifests/kube-state-metrics/kube-state-metrics-cluster-role.yaml
@@ -7,7 +7,10 @@ rules:
   resources:
   - nodes
   - pods
+  - services
   - resourcequotas
+  - replicationcontrollers
+  - limitranges
   verbs: ["list", "watch"]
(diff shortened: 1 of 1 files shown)


brancz:
Out of ease of maintenance, can you syncronize the order of the resources with the [upstream cluster role example](https://github.com/kubernetes/kube-state-metrics/blob/master/kubernetes/kube-state-metrics-cluster-role.yaml)?

brancz:
Other than that lgtm

jordanjennings:
@brancz Synced with upstream

brancz:
Perfect. Thanks a lot!

PR #463, opened 2017-06-30
Check createDatasourceFromFile error

exarkun:
Previously the error result of createDatasourceFromFile in the loop in createDatasourcesFromFile was dropped.  This meant any errors from datasource creation would be ignored.  This change makes the loop check and propagate any errors.


brancz:
Thanks, good catch! Could you update all the usages of the image in the repo and bump the patch version tag by 1? Then I can just build and push the image and then merge the PR.

exarkun:
@brancz like that?

brancz:
Perfect! I'll get it up and then merge.

PR #477, opened 2017-07-12
Scrapes cAdvisor port for metrics in Kubernetes 1.7

alindeman:
Kubernetes 1.7 [inadvertently stopped reporting cAdvisor metrics on the main /metrics endpoint](https://github.com/kubernetes/kubernetes/issues/48483). They are now only reported on the cAdvisor /metrics endpoint. As a temporary solution, I've added the cAdvisor port to the kubelet endpoint and servicemonitor so they get scraped as well.

I'm not totally sure it's a good idea to merge this, since [the original behavior might be restored soon](https://github.com/kubernetes/kubernetes/issues/48483#issuecomment-314828142), but I did want to bring it up in this repository in an attempt to save other folks some headdesking trying to figure out why pod-level metrics are not being reported in Kubernetes 1.7.

As a stopgap for my use case, I've built my own container with this fix and pushed it to a registry. Hopefully the original behavior will be restored in an upcoming point release of Kubernetes.

### DIFF
--- a/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-service-monitor-kubelet.yaml
+++ b/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-service-monitor-kubelet.yaml
@@ -9,6 +9,8 @@ spec:
   endpoints:
   - port: http-metrics
     interval: 30s
+  - port: cadvisor
+    interval: 30s
     honorLabels: true
   selector:
     matchLabels:
--- a/Documentation/user-guides/cluster-monitoring.md
+++ b/Documentation/user-guides/cluster-monitoring.md
@@ -326,6 +326,8 @@ spec:
   endpoints:
   - port: http-metrics
     interval: 30s
+  - port: cadvisor
+    interval: 30s
     honorLabels: true
   selector:
     matchLabels:
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -336,15 +336,19 @@ func (c *Operator) syncNodeEndpoints() {
 			},
 		},
 		Subsets: []v1.EndpointSubset{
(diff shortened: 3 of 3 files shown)


contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-service-monitor-kubelet.yaml
```diff
@@ -10,6 +10,9 @@ spec:
   - port: http-metrics
     interval: 30s
     honorLabels: true
```
fabxc: This can be removed now as it only was needed for cAdvisor metrics.
Those being exposed on separate endpoints now is actually good from a monitoring perspective.

fabxc:
ok to test

fabxc:
Can you run `make docs` to carry over the YAML changes into the guide?
The CI output is currently not publicly reachable. Sorry about that.

alindeman:
@fabxc Thanks for the review. I've adjusted the diff based on your feedback.

mxinden:
@alindeman @fabxc The CI failure is a false negative. Linting, unit and e2e tests succeeded. Sorry for the trouble.

danp:
Heads up that https://github.com/kubernetes/kubernetes/issues/48483 has been fixed by https://github.com/kubernetes/kubernetes/pull/49079.

brancz:
lgtm, thanks a lot @alindeman ! It's a good idea to separate out cAdvisor anyways.

brancz:
@danp it seems like this is a solution that will continue working the same way as the separate path, but this is available now, so right now the better option.

alindeman:
> lgtm, thanks a lot @alindeman ! It's a good idea to separate out cAdvisor anyways.

Perfect, thanks.

PR #478, opened 2017-07-12
Added a few extra node rules

zyonash:
Alerts for a node out of disk, and a node under either memory or disk pressure were not configured by default.  We added them into our implementation, and figured these would be nice to have by default.

### DIFF
--- a/contrib/kube-prometheus/assets/prometheus/rules/node.rules
+++ b/contrib/kube-prometheus/assets/prometheus/rules/node.rules
@@ -8,3 +8,35 @@ ALERT NodeExporterDown
     summary = "node-exporter cannot be scraped",
     description = "Prometheus could not scrape a node-exporter for more than 10m, or node-exporters have disappeared from discovery.",
   }
+ALERT K8SNodeOutOfDisk
+  IF kube_node_status_out_of_disk{condition="true"} == 1
+  LABELS {
+    service = "k8s",
+    severity = "critical"
+  }
+  ANNOTATIONS {
+    summary = "Node ran out of disk space.",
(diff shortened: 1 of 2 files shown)


WIZARD-CXY:
good point

mxinden:
@zyonash Thank you very much for the contribution. This looks good to me, but I would like to wait for @brancz to look over this. He developed the existing rule set.

brancz:
Thanks a lot for your contribution! These look great! Could you just add the rules [here](https://github.com/zyonash/prometheus-operator/blob/ae861894ba8765a1798cb1c9d4b910fd70506626/contrib/kube-prometheus/assets/prometheus/rules/node.rules). The manifest is actually generated through [this script](https://github.com/zyonash/prometheus-operator/blob/ae861894ba8765a1798cb1c9d4b910fd70506626/contrib/kube-prometheus/hack/scripts/generate-manifests.sh), just run that script from the `contrib/kube-prometheus/` directory.

brancz:
ping @zyonash ? No worries if you're busy, we can take over, just let us know. Thanks for your effort so far!

zyonash:
@brancz Apologies, got side tracked.  Will update this today.  Thanks!

zyonash:
@brancz Rules have been added to the appropriate file.  Thanks.

brancz:
Thanks! lgtm!

galexrt:
Will these new rules be added to the rules deployed by default for a Tectonic cluster?

brancz:
They will undergo additional review, but yes typically we vendor into Tectonic what we collaborate on here with additional rules that only apply to Tectonic clusters.

PR #482, opened 2017-07-17
Add link to PersistentVolumeClaim reference

aknuds1:
In the API docs, link to the PersistentVolumeClaim reference.

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -6,7 +6,7 @@ This Document documents the types introduced by the Prometheus Operator to be co
 
 ## AlertingSpec
 
-AlertingSpec defines paramters for alerting configuration of Prometheus servers.
+AlertingSpec defines parameters for alerting configuration of Prometheus servers.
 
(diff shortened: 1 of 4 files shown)


brancz:
Those links are actually auto generated through the `cmd/apidocgen` tool. Could you just add the link [here](https://github.com/coreos/prometheus-operator/blob/master/cmd/apidocgen/main.go#L38-L45) and run `make generate` to verify it works? That would be of huge help! :slightly_smiling_face:

aknuds1:
@brancz Done. I also modified the Makefile to use `embedmd` and `apidocgen` from `$GOPATH/bin/` in case the user doesn't have `$GOPATH/bin` in $PATH or they have other versions in there.

brancz:
@coreosbot test this please

brancz:
lgtm thanks a lot @aknuds1 !

aknuds1:
Thanks @brancz!

PR #489, opened 2017-07-20
Update charts to accept fullname up to 63 characters

cofyc:
Hi, 

In testing, I found some of objects created with truncated name, like `abcd-monitor-exporter-no`, `abcd-monitor-exporter-ku`, which is exactly 24 characters long.

But Kubernetes now accept service names up to 63 characters (https://github.com/kubernetes/kubernetes/pull/29523), and helm had updated their charts too (https://github.com/kubernetes/helm/pull/1740).


brancz:
thanks! lgtm :+1:

PR #492, opened 2017-07-20
Cut 0.11.0

brancz:
@mxinden @Gouthamve


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 0.11.0 / 2017-07-20
+
+Warning: this release deprecates the previously used storage definition in favor of upstream PersistentVolumeClaim templates. While this should not have an immediate effect on a running cluster, Prometheus object definitions that have storage configured need to be adapted. The previously existing fields are still there, but have no effect anymore.
```
Gouthamve: *This (captial T)

mxinden: Can we point people to an example how this storage definition changes are done best?

brancz: The thing is that the previously used storage definition is basically identical with the PVC template. So I think we should be good without an explicit example here, as the storage docs have been adapted.

mxinden: 👍

PR #495, opened 2017-07-21
[WIP] tests: Make sure tectonic-builder image is up to date

mxinden:
@Gouthamve Noticed, that we are not actually spinning up a k8s 1.7 cluster, even though current tectonic-installer:master has it.

My guess is, that the tectonic-installer image on our Jenkins workers is not up to date.


brancz:
lgtmog

Gouthamve:
LGTM

PR #500, opened 2017-07-23
Configure Prometheus logger format and level

julianvmodesto:
https://github.com/coreos/prometheus-operator/issues/474

### DIFF
--- a/pkg/client/monitoring/v1alpha1/types.go
+++ b/pkg/client/monitoring/v1alpha1/types.go
@@ -68,6 +68,8 @@ type PrometheusSpec struct {
 	Retention string `json:"retention,omitempty"`
 	// Interval between consecutive evaluations.
 	EvaluationInterval string `json:"evaluationInterval,omitempty"`
+	// The Prometheus logger format and level configuration.
+	Log *LogSpec `json:"log,omitempty"`
 	// The labels to add to any time series or alerts when communicating with
 	// external systems (federation, remote storage, Alertmanager).
 	ExternalLabels map[string]string `json:"externalLabels,omitempty"`
@@ -125,6 +127,16 @@ type PrometheusStatus struct {
 	AvailableReplicas int32 `json:"availableReplicas"`
 	// Total number of unavailable pods targeted by this Prometheus deployment.
 	UnavailableReplicas int32 `json:"unavailableReplicas"`
+}
+
+// LogSpec defines the configured log messages from the Prometheus logger.
+type LogSpec struct {
+	// The log target and format. Example:
+	// "logger:syslog?appname=bob&local=7" or "logger:stdout?json=true"
+	Format string `json:"format"`
+	// Only log messages with the given severity or above. Valid labels:
+	// [debug, info, warn, error, fatal]
+	Level string `json:"level"`
 }
 
 // AlertingSpec defines parameters for alerting configuration of Prometheus servers.
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -303,6 +303,19 @@ func makeStatefulSetSpec(p v1alpha1.Prometheus, c *Config, ruleConfigMaps []*v1.
 				"-storage.local.target-heap-size="+fmt.Sprintf("%d", reqMem.Value()/3*2),
 			)
 		}
+
+		if p.Spec.Log != nil {
+			if p.Spec.Log.Format != "" {
+				promArgs = append(promArgs,
+					"-log.format="+p.Spec.Log.Format,
+				)
+			}
+			if p.Spec.Log.Level != "" {
+				promArgs = append(promArgs,
+					"-log.level="+p.Spec.Log.Level,
+				)
+			}
+		}
 	case 2:
 
 		// Prometheus 2.0 is in alpha and is highly experimental, and therefore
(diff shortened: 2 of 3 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -319,6 +333,20 @@ func makeStatefulSetSpec(p v1alpha1.Prometheus, c *Config, ruleConfigMaps []*v1.
 			"-storage.tsdb.path=/var/prometheus/data",
 			"-storage.tsdb.retention="+p.Spec.Retention,
 		)
+
+		log := p.Spec.Log
```
brancz: Why duplicate these arguments in this case? You should be able to just add all common flags below.

pkg/client/monitoring/v1alpha1/types.go
```diff
@@ -125,6 +127,16 @@ type PrometheusStatus struct {
 	UnavailableReplicas int32 `json:"unavailableReplicas"`
 }
 
+// LogSpec defines the configured log messages from the Prometheus logger.
+type LogSpec struct {
```
brancz: I'd like to wait for @fabxc to return from vacation, as he has a much better idea on how the logging story is going to continue with Prometheus 2.0.

pkg/prometheus/statefulset.go
julianvmodesto: Done

pkg/client/monitoring/v1alpha1/types.go
julianvmodesto: Sure thing for confirmation. Looks like Prometheus 2.0 also uses logrus in prometheus/common/log.
```
$ docker run prom/prometheus:v2.0.0-beta.0 -h
usage: prometheus [<flags>]

The Prometheus monitoring server

Flags:
  -h, --help                     Show context-sensitive help (also try
                                 --help-long and --help-man).
      --version                  Show application version.
      --log.level="info"         Only log messages with the given severity or
                                 above. One of: [debug, info, warn, error,
                                 fatal]
      --log.format="logger:stderr"
                                 Set the log target and format. Example:
                                 "logger:syslog?appname=bob&local=7" or
                                 "logger:stdout?json=true"
```

brancz: That makes sense, I just know that there is some fuzz going on around logging and as 2.0 is not final yet there may still be changes to it.

brancz: Turns out this was a good call :slightly_smiling_face: seems like the semantics are going to change quite a bit, I'd suggest we hold off until we fully support 2.0 (and 2.0 has a stable release) https://github.com/prometheus/prometheus/pull/3061

brancz:
The logging story for 2.0 is now much more clear, and basically what ended up happening, is that the `--log-format` flag has been removed. So to ensure compatibility between both 1.x and 2.x let's just introduce the `log_level` field for the Prometheus spec.

This also needs a rebase :slightly_smiling_face: .

julianvmodesto:
Ooh thanks! Will take a look.

TatyanaBol:
Hello
We use prometheus-operator.
It fills our logs with a lot of info entries.
How can we configure log level to be "error"?
Thanks

brancz:
@TatyanaBol if you take this PR and finish it, I'm happy with merging it :slightly_smiling_face:

TatyanaBol:
I have no clue how to do it!
Is it impossible now?

brancz:
It's not possible yet, but it's pretty easy to implement, so I'll probably go ahead tomorrow and implement it :slightly_smiling_face: .

TatyanaBol:
Great!
Please let me know when and how I can use it
Thanks!

brancz:
Closing this in favor of #837

TatyanaBol:
How can I configure log level?

brancz:
We'll need to release this first, but then it will be the `logLevel` field in the `PrometheusSpec` as specified in the documentation: https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#prometheusspec

TatyanaBol:
Can you please explain, where should I specify logLevel?
Which file? What is format?
When it will be released?
Thanks!

brancz:
> Can you please explain, where should I specify logLevel?
> Which file? What is format?

In your Prometheus manifest. As all Kubernetes manifests, this is yaml.

> When it will be released?

We have collected quite a few features, so it shouldn't be long.

TatyanaBol:
Is it this file? https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s.yaml
Or this? https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/prometheus-operator/prometheus-operator.yaml

What and where should I add?

Thank you for your help!

brancz:
The first :) and on the same level as `version`.

TatyanaBol:
in this file https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s.yaml  should be
spec:
   replicas: 2
   version: v2.0.0
    **logLevel: Error**
   serviceMonitorSelector:

Correct?
Is it case sensitive?

I appreciate your help a lot

brancz:
Correct, and yes it’s case sensitive

TatyanaBol:
Thank you so much!
please let me know when I can use it

TatyanaBol:
Was this fix released?

brancz:
It hasn't been released yet, sorry for the inconvenience. We should be able to get it soon though.

pkg/client/monitoring/v1alpha1/types.go
sneko: Hi, sorry to post here but just one question: it seems there is no way to separate TRACE/DEBUG/INFO/WARN on the stdout and ERROR on the stderr, am I right?

And if yes, could you explain why please :) ? Thanks

squat: It’s not possible as it is currently written. Typically, it’s good practice to have all logs for a service be ingested by your logging solution (journald, or some custom logging and processing stack) and then filter/process those as you wish. It’s not really feasible for the operator to encode all possible fancy processing capabilities and preferable for the user to create a custom pipeline.

brancz: @sneko I believe what you are talking about is the logs from the Prometheus Operator, unfortunately those come from kubernetes/client-go and they use a global logger, that we have little control over. You can however still control the log level of that one (see the [glog documentation](https://github.com/golang/glog)).

Issue #483, opened 2017-07-19
Memory-Leak in prometheus-operator


steigr:
**What did you do?**

Fresh k8s/tectonic 1.6.7-tectonic.1 install with tectonic-installer on 4 nodes ( 3 masters + 1 worker).

**What did you expect to see?**

prometheus-operator should run continuously.

**What did you see instead? Under which circumstances?**

the operator memory usage grows and get eventually kill because of OOM. k8s reschedules the pod.
<img width="531" alt="memory-leak" src="https://user-images.githubusercontent.com/186793/28345213-12810d5e-6c29-11e7-83d9-646e458c1b0b.png">

**Environment**

k8s/tectonic 1.6.7-tectonic.1, 3 masters with 2 threads and 2gb ram, 1 worker with 8 threads and 48 gb ram. The nodes are connected over a VPN (wireguard). Latency between masters is low (~1ms). The worker is reachable within 15-20ms.

* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.4", GitCommit:"d6f433224538d4f9ca2f7ae19b252e6fcb66a3ae", GitTreeState:"clean", BuildDate:"2017-05-19T20:41:24Z", GoVersion:"go1.8.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.7+coreos.0", GitCommit:"c8c505ee26ac3ab4d1dff506c46bc5538bc66733", GitTreeState:"clean", BuildDate:"2017-07-06T17:38:33Z", GoVersion:"go1.7.6", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: 

via tectonic-installer 1.6.7

* Manifests:

[tectonic-installer/.../prometheus-k8s.yaml](https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml)

* Prometheus Operator Logs:

```
...
ts=2017-07-19T00:26:55Z caller=operator.go:972 component=prometheusoperator msg="updating config skipped, no configuration change"
ts=2017-07-19T00:26:55Z caller=operator.go:584 component=prometheusoperator msg="update handler" old=324451 cur=324458
ts=2017-07-19T00:26:55Z caller=operator.go:321 component=alertmanageroperator msg="update handler" old=324451 cur=324458
[...]

PR #504, opened 2017-07-24
*: add profiling endpoints

brancz:
#483 suggests there might be a memory leak in the Prometheus Operator. To be able to investigate properly, we'll need to perform some profiling.

@Gouthamve @fabxc @mxinden

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


brancz:
@mxinden I was thinking about that as well, but we expose similar data in the metrics and the Prometheus Operator doesn't make sense to be public to begin with. I started implementing this with a flag in place to be able to disable it, but decided to drop it and wait for the necessity and a real "use case".

PR #508, opened 2017-07-25
generate Prometheus Operator deployments with jsonnet

### LINKED ISSUE #464: Which is the canonical representation of the operator, the bundle or the kube-prometheus manifests?
I'm trying to figure out how to install the Prometheus operator, but there is the issue that this repository says either to install it from [bundle.yaml](https://github.com/coreos/prometheus-operator/blob/master/bundle.yaml) or from the [kube-prometheus manifests](https://github.com/coreos/prometheus-operator/tree/v0.10.1/contrib/kube-prometheus/manifests/prometheus-operator), and the two differ from each other. Can you please either synchronize them so they are equivalent or make one of them the One True Way of installing the operator? It would be extremely helpful.

brancz:
#464 described that the Prometheus Operator deployment manifests can easily get out of sync because they are manually maintained.

This PR introduces generating these manifests from a single source of truth using jsonnet, in fact as described in #464, there is no reason for the manifest in the `bundle.yaml` to be different from the one in `kube-prometheus`. Another great thing this introduces is that there is only 1 version that actually has to be changed when releasing, the `VERSION` file.

I decided not to put this in the `generate` make target, as a lot of tooling is necessary for it (jsonnet, lib-ksonnet, json2yaml converter). We could create a docker image that contains all this tooling.

Closes #464

@Gouthamve @mxinden @fabxc


mxinden:
@brancz Awesome! I would suggest to add some Jenkins automation code with this PR, so it always stays in sync. What do you think?

scripts/jenkins/check-make-generate.sh
```diff
@@ -8,6 +8,8 @@ set -u
 # print each command before executing it
 set -x
 
+make generate
+
 docker run \
```
mxinden: This docker run call has to be removed @brancz

brancz:
@coreosbot test this please

brancz:
@coreosbot test this please

mxinden:
@brancz Same issue that we had before:
```
--- FAIL: TestPrometheusResourceUpdate (13.65s)
	prometheus_test.go:161: resources don't match. Has {map[] map["memory":{{%!q(int64=104857600) '\x00'} {%!q(*inf.Dec=<nil>)} "100Mi" "BinarySI"}]}, want {map[] map["memory":{{%!q(int64=209715200) '\x00'} {%!q(*inf.Dec=<nil>)} "" "BinarySI"}]}
2017/07/27 08:27:30 retrieving stateful set failed: statefulsets.apps "prometheus-test" not found
```

brancz:
@mxinden yes looks like it. Now I wonder whether it might be a bug instead of a flake. I wonder how we could reproduce reliably. Actually looking at it more closely, it seems like we're doing something wrong in the expectation that we're setting.

Anyhow for the purpose of this PR I will re-run. @coreosbot test this please.

brancz:
@mxinden could you have another final look?

mxinden:
No objections. Will merge here.

aknuds1:
Good stuff!

PR #514, opened 2017-07-27
grafana helm chart: add server root url configuration

tsloughter:
This change is needed to allow configuring the `root_url` in cases the default doesn't work. For example to be able to load properly through simply running `kubectl proxy` you can do:

```
helm install opsgoodness/grafana --name my-release --set server_root_url=http://127.0.0.1:8001/api/v1/proxy/namespaces/default/services/my-release-grafana:80/
```

I realize now it isn't perfect since doing that requires not using a generated release name, I don't think there is a way to have a variable in a variable for the helm template? but yea.

### DIFF
--- a/helm/grafana/templates/grafana-deployment.yaml
+++ b/helm/grafana/templates/grafana-deployment.yaml
@@ -23,6 +23,8 @@ spec:
       - name: grafana
         image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
         env:
+        - name: GF_SERVER_ROOT_URL
+          value: "{{ .Values.server_root_url }}"
         - name: GF_AUTH_BASIC_ENABLED
           value: "true"
         - name: GF_AUTH_ANONYMOUS_ENABLED
--- a/helm/grafana/values.yaml
+++ b/helm/grafana/values.yaml
@@ -5,6 +5,8 @@ annotations: {}
 adminUser: "admin"
 adminPassword: "admin"
 
+server_root_url: "%(protocol)s://%(domain)s:%(http_port)s/"
+
 service:
 
   ## Annotations to be added to the Service


tsloughter:
And related to this, the `README` for it the grafana chart says `helm install opsgoodness/grafana` will work but it appears to not have been published there yet.

brancz:
Yes I don't think @mgoodness has not had a chance to publish them yet.

brancz:
@weiwei04 could you have a look at this?

helm/grafana/templates/grafana-deployment.yaml
```diff
@@ -23,6 +23,8 @@ spec:
       - name: grafana
         image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
         env:
+        - name: GF_SERVER_ROOT_URL
+          value: "{{ .Values.server_root_url }}"
         - name: GF_AUTH_BASIC_ENABLED
```
weiwei04: to avoid the ingress fqdn confict with server_root_url, it's better check if .Values.ingress.enabled, if the .Values.ingress.enabled then we should ignore the server_root_url, if .Values.ingress.enabled is false, we can set .Values.server_root_url.

brancz:
I personally think the ingress resource should be left up to the user entirely, this setting should simply be surfaced in the helm chart, otherwise we're forced to implement pretty much all features of ingress/services in helm, which is a never ending maintenance story. My preference would be to just allow setting this value in helm and then leave it up to the user to expose it with an ingress object.

tsloughter:
Ok, so just wrapping it in a `if` is needed then?

brancz:
That would be my idea.

weiwei04:
@brancz I agree
>otherwise we're forced to implement pretty much all features of ingress/services in helm, which is a never ending maintenance story.

reconsider the design, I should make ingress in values.yaml field like requests: {} just expose the raw k8s config, but as the chart already published, some user may have already deploy some instances, change the ingress filed may have compatibility issues when upgrade, I'm still trying to find a better way(may be add a new field and prefer the new field over current ingress field).

for now, I think we can stick to the ** Name based virtual hosting ** type ingress, and make sure for grafana, the server_root_url work nicely with current ingress setting by add a `if` check?

brancz:
I understand the compatibility concern, but given that the chart is in `v0.0.3` and the grafana-watcher is pre-alpha, it's perfectly safe to do all sorts of breaking changes. The Prometheus Operator has gone through numerous breaking changes in the early days, and that's perfectly fine and normal in software development. Until we figure out something we're actually happy with we should be breaking. I don't think any form of exposing services is the job of any helm chart, it's a template that people can build their own opinion around. Otherwise we'll have to start including authentication proxies, databases, etc. all duplicated in all charts out there that have any dependency - and that, is real maintenance madness.

gianrubio:
Closing as outdated

PR #517, opened 2017-07-27
alertmanager: adapt storage mounted to prometheus changes

brancz:
Fixes https://github.com/coreos/prometheus-operator/issues/516

As we deprecated the storage fields we should have adapted the Alertmanager accordingly, so here the fix for that.

@fabxc @Gouthamve @mxinden


fabxc:
👍

PR #525, opened 2017-07-31
Add docs about TPR->CRD migration

Gouthamve:


Documentation/migration.md
```diff
@@ -0,0 +1,81 @@
+# Migration from TPR to CRDs
```
brancz: Maybe drop an introductory paragraph in here why we're doing this and that the Prometheus Operator will automatically perform this migration, and that typically a user will not have to interact.

Gouthamve: Done, PTAL.

brancz:
lgtm :+1: very nice!

PR #529, opened 2017-08-01
Documentation: add compatibility doc

### LINKED ISSUE #524: Stable version for prometheus-operator and prometheus to work with
Guys, which versions of both prometheus-operator and prometheus are stable atm?
Because I'm using both latest versions and some function not working properly
e.g Targets page does not show my own service when using ServiceMonitor even though I follow strictly the user guideline, or The config-reloader not auto-reload configuration file whenever I update it.

brancz:
As we already have the compatibility "matrix", we should have this information in docs file. I extended our tool to generate the API docs with also being able to generate this doc.

Closes #524 

@fabxc @Gouthamve @mxinden


fabxc:
👍

PR #531, opened 2017-08-01
Update prometheus-k8s-roles.yaml

xvzup:
Resource pods added to role prometheus-k8s for namespace default. This is required to monitor kube-apiserver.


brancz:
Perfect! Thanks a lot for reporting and fixing! lgtm :+1:

PR #537, opened 2017-08-02
Adds scrape timeout property to ServiceMonitor endpoint

tao-lu:
In reference to #536 .

PR #539, opened 2017-08-03
Adding a configurable scrapeTimeout for prometheus operator.

gmcquillan:
Hi CoreOS team,

I couldn't find a comprehensive developers guide for ensuring this works. I figured I'd make a local build locally and test it out in our QA cluster. If there's something faster and easier than just running ``go test`` in the pkg/prometheus subdir, I'd love to hear about it. 

Cheers!

### DIFF
--- a/pkg/client/monitoring/v1alpha1/types.go
+++ b/pkg/client/monitoring/v1alpha1/types.go
@@ -66,6 +66,8 @@ type PrometheusSpec struct {
 	Replicas *int32 `json:"replicas,omitempty"`
 	// Time duration Prometheus shall retain data for.
 	Retention string `json:"retention,omitempty"`
+	// Number of seconds to wait for target to respond before erroring.
+	ScrapeTimeout string `json:"scrape_timeout,omitempty"`
 	// Interval between consecutive evaluations.
 	EvaluationInterval string `json:"evaluationInterval,omitempty"`
 	// The labels to add to any time series or alerts when communicating with
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -76,8 +76,7 @@ func generateConfig(p *v1alpha1.Prometheus, mons map[string]*v1alpha1.ServiceMon
 		Key: "global",
 		Value: yaml.MapSlice{
 			{Key: "evaluation_interval", Value: evaluationInterval},
-			{Key: "scrape_timeout", Value: "30s"},
-			{Key: "scrape_interval", Value: scrapeTimeout},
+			{Key: "scrape_timeout", Value: scrapeTimeout},
 			{Key: "external_labels", Value: stringMapToMapSlice(p.Spec.ExternalLabels)},
 		},
 	})
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -67,12 +67,17 @@ func generateConfig(p *v1alpha1.Prometheus, mons map[string]*v1alpha1.ServiceMon
 	if p.Spec.EvaluationInterval != "" {
 		evaluationInterval = p.Spec.EvaluationInterval
 	}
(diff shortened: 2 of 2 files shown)


gmcquillan:
It also looks like making this change might require a corresponding change to the Kubernetes API; is that accurate?

If so, I'd wonder if it makes sense to use labels for configuration key/values?

``prometheus-operator/scrapetime: "30s"``, say.

brancz:
`make test` executes all unit tests, and you can execute the e2e tests on a local minikube by compiling the static binary (which is what is used for the container images) with `make crossbuild` and then build the container image with the docker host from within minikube by running `eval $(minikube docker-env)`, then you can build the container using `make container` and then finally run the e2e tests using `make e2e-tests`.

Note that generally we will also run all of this on CI, however, we're having some trouble with our jenkins instances right now, should be working soon again though.

As a side note a section in the readme on developing might be helpful for future reference. Do you want to create a PR to add such a section :slightly_smiling_face: ?

gmcquillan:
Happy to. I'll see if I can get my mini kube working today or tomorrow.

Cheers,
Gavin

On Aug 7, 2017 8:49 AM, "Frederic Branczyk" <notifications@github.com>
wrote:

make test executes all unit tests, and you can execute the e2e tests on a
local minikube by compiling the static binary (which is what is used for
the container images) with make crossbuild and then build the container
image with the docker host from within minikube by running eval $(minikube
docker-env), then you can build the container using make container and then
finally run the e2e tests using make e2e-tests.

Note that generally we will also run all of this on CI, however, we're
having some trouble with our jenkins instances right now, should be working
soon again though.

As a side note a section in the readme on developing might be helpful for
future reference. Do you want to create a PR to add such a section 🙂 ?

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/coreos/prometheus-operator/pull/539#issuecomment-320654391>,
or mute the thread

fabxc:
My opinion so far has been that the scrape timeout being as long as the scrape interval should generally be sane for all setups.
Can you elaborate on the exact use case that makes you want to set it explicitly?

gmcquillan:
Yes -- in this case we're querying an appliance (vsphere) which is
responsible for 10s - 100s of machines. We're querying its SOAP API and it
takes about 20 seconds to pull down the complete metrics set.

Also, I've found similar issues when polling SNMP traps on some embedded
devices where the response simply takes 10s of seconds. A universal
scrapTimeout precludes any such metrics.

Given these circumstances, I think it's pretty reasonable to give the
ServiceMonitor an opportunity to override the default for the managed
Prometheus service.

brancz:
In that case I think the functionality from this PR #537 might already all we need. I'd personally also prefer not to bloat the Prometheus object itself.

gmcquillan:
Sure. That PR didn't exist when I began work on this branch. I'd be happy with either one getting merged.

I'd argue that making the API Spec attribute name match the configuration attribute in Prometheus is a good idea, however (i.e. `s/timeout/scrape_timeout/`).

gmcquillan:
Closing in favor of #537

carlosedp:
Any chance of getting this functionality into the operator? It would help setting the global timeout for the serviceMonitors that doesn't specify it.

A use case would be the vast amounts of RaspberryPi clusters where some scraping times-out due to low performance.

I can send a new PR addressing this.

brancz:
I'm not entirely opposed to it, but I do have to wonder how much that will actually help you, as ServiceMonitors/PodMonitors can specify scrape timeouts themselves which will always take precedence.

carlosedp:
Many targets don't have scrape_timeout definitions like the core ones (kube-apiserver, kube-scheduler, kube-controller) so having the hability to change the default value is desired.

Opened https://github.com/coreos/prometheus-operator/pull/3250 to address this.

PR #546, opened 2017-08-07
tests: Set thresholds how long to keep past builds

mxinden:
There were a bunch of old builds laying around on the worker machine. This should fix the problem for the future.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,9 @@
 job('po-tests-pr') {
     concurrentBuild()
 
+    // logRotator(daysToKeep, numberToKeep)
+    logRotator(10, 10)
+
     parameters {
         stringParam('sha1')
     }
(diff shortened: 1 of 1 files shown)


brancz:
approach sgtm, I'm actually questioning whether we even need 50 master builds, but feel free to merge as you see fit

mxinden:
@brancz Changed to 30. Let's merge once the CI is running again.

brancz:
lgtm and sgtm

brancz:
@mxinden this still sounds like it makes sense, can we merge this or do you want to rebase and run tests?

mxinden:
@brancz I have re-based. The Jenkinsfile is parsed correctly. I have stopped the execution, in my opinion there is no need to start a k8s cluster for this. Feel free to merge.

PR #551, opened 2017-08-08
WIP: Move to v1 from v1alpha1

Gouthamve:
A super early version, migration is buggy hence, tests will fail.

So we cannot update the CRD version as that field is immutable. Which means the only option is to run a migration.

1. Do we release a `0.12.0` with the TPR --> CRD changes and then have another migration which moves from `v1alpha1` to `v1`.
2. Do we make `v1alpha1` to `v1` part of the TPR --> CRD change and delay `0.12.0` to tomorrow (hopefully)?

I favour a `0.12.0` now and then a migration because we have the setup for a migration in place and adding a new kind of migration is fairly straight forward. This PR in its current state is trying to migrate to `v1` as part of TPR --> CRD change.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -25,6 +25,9 @@ import (
 	"os/signal"
 	"syscall"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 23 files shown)


fabxc:
One consideration here is that TPR->CRD migration is considerate of cluster versions. So the next operator version will run seamlessly.

This change is breaking as it will query a different resource path for Prometheus. Obviously quite tedious to have all the code handle v1alpha1 and v1. Since our schema didn't change, we could just do this transparently by tweaking the request path but use `monitoringv1` in the code.

Or we just make it a breaking change and people will have to upgrade to k8s 1.7 if they want to use the latest version.

brancz:
@fabxc given the [compatibility of client-go 4.0-beta.0](https://github.com/kubernetes/client-go#compatibility-matrix) we may just want to do the latter option.

Gouthamve:
So I am kinda blocked here because I am seeing weird behaviour on my local: https://github.com/kubernetes/apiextensions-apiserver/issues/6 https://github.com/kubernetes/apiextensions-apiserver/issues/5

Can you trigger the tests once? @fabxc @brancz @mxinden

brancz:
@Gouthamve we're currently having some difficulties with CI, so we'll probably only be able to trigger tests tomorrow. Releasing is not super urgent right now, so we're not much in a hurry.

brancz:
Oops, that conflict was not intentional, this will need a rebase @Gouthamve .

brancz:
@coreosbot test this please.

brancz:
Closing here in favor of #555

Issue #503, opened 2017-07-24
tests: Expose logs of the Prometheus Operator after CI test run


mxinden:
The Jenkins-CI should expose the logs of the Prometheus Operator for better debugging. One way of solving this is to write the logs of the Prometheus Operator into a file into `./build` before killing it. Whatever is in that folder will be archived by Jenkins and thereby visible after the test execution.

brancz:
Out of necessity I added this to #555. As that's merged now, I'm closing here.

PR #555, opened 2017-08-09
Move to v1 from v1alpha1

brancz:
The main change that we had to do here is that we needed to not do one TPR after another but all TPRs step by step, but each step for all TPRs, then we could successfully register and reach the api group.

@fabxc

### DIFF
--- a/Documentation/design.md
+++ b/Documentation/design.md
@@ -1,8 +1,8 @@
 # Design
 
-This document describes the design and interaction between the third party resources that the Prometheus Operator introduces.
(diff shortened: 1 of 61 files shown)


brancz:
@mxinden out of my own necessity I added the code for #503 as my latest commit.

brancz:
@coreosbot test this please

brancz:
@coreosbot test this please

fabxc:
👍

brancz:
@coreosbot test this please

PR #556, opened 2017-08-10
Allow Grafana watcher to monitor multiple directories

unguiculus:


contrib/grafana-watcher/main.go
```diff
@@ -16,31 +16,67 @@ package main
...
+type watchDirSet map[string]struct{}
+
+func (w *watchDirSet) String() string {
+	s := *w
+	return strings.Join(s.asSlice(), ",")
+}
+
+func (w *watchDirSet) Set(value string) error {
+	s := *w
+	cols := strings.Split(value, ",")
```
brancz: I'm not a huge fan of introducing both `--flag=a --flag=b,c`, let's just stick with a repeatable flag `--flag=a --flag=b --flag=c`.

brancz:
Thanks a lot for your contribution!

Could you also bump the usage of the v0.0.6 tag everywhere in this repo and make sure it still works as expected? Specifically the manifest in `contrib/kube-prometheus/`. I just want to make sure manifests are adapted if needed especially regarding the pflags switch.

contrib/grafana-watcher/main.go
unguiculus: I just copied this from kube-state-metrics.

brancz: Interesting, I guess then it's better to be consistent. Thanks for pointing it out.

brancz:
Already pushed the image to the quay repo. Nice work! lgtm

PR #557, opened 2017-08-10
Use nodeSelector value in grafana helm chart

ryanwalls:
Currently the `nodeSelector` value is ignored.


brancz:
lgtm, but last call for Grafana charts go to @weiwei04

weiwei04:
lgtm, thanks @ryanwalls

Issue #520, opened 2017-07-28
Adding to Grafana dashboards


tsloughter:
First, is there a chat where operators are discussed? Questions like this I would first bring there if there was a gitter/slack/irc channel for discussing operators. I try to bring the helm specific questions to the helm-users channel on slack first, but often can't find answers.

So I'm trying to figure out a good way of adding dashboards to a grafana deployed through the prometheus-operator's grafana helm chart. When I had been deploying prometheus-operator through `kube-prometheus` scripts I would have my application's helm chart write over the grafana configmap with a new one. However, this is not possible when grafana was deployed with a separate helm chart.

I wanted to know if anyone had solved this issue. Not necessarily a full solution for dynamically adding dashboards for resources, but at least being able to recreate/rewrite what grafana's dashboard volume is made of.

brancz:
We don't have a channel dedicated to the Prometheus Operator, however, lots of people use the `#prometheus` (on freenode) irc channel or `sig-instrumentation` on the Kubernetes slack.

As I don't use the grafana helm chart I cannot comment reliably on the functionality. As far as I understand, the grafana chart should be independent of the dashboards it is deployed with and only when deploying the `kube-prometheus` chart the grafana chart should be used to be deployed with a set of dashboards.

@weiwei04 being the maintainer of this chart, he can probably more reliably answer your questions.

tsloughter:
@brancz thanks!

Sure, the issue is, at least for me, wanting to have an unmodified kube-prometheus and grafana charts that get installed and then when I install another app through a chart that needs specific dashboards they have to manually added each time. Or grafana has to be a dep of each chart, and then include its dashboards, and then the user switches grafanas to see the different dashboards.

A separate issue is it seems `values.yaml` can't be templated... So I haven't found a way to populate the grafana `serverDashboardFiles` variable from a set of json files.

weiwei04:
@tsloughter 

1. The grafana chart under github.com/coreos/prometheus-operator/helm is mainly for kube-prometheus, when helm install kube-prometheus, prometheus(with predefined k8s monitor targets, alertrulles), alertmanager and grafana(with pre-installed dashboard for k8s) are installed together with one cmd.

[...]

PR #558, opened 2017-08-11
remove pre-installed grafana dashboards

weiwei04:
According to #520, the grafana chart shound't pre-install kube-prometheus dashboards, the dashboards should mv into kube-prometheus value.yaml.


brancz:
lgtm

PR #559, opened 2017-08-11
Require prometheus service url to setup grafana

gianrubio:
The current prometheus url configuration on grafana is pointing to the grafana release name and not to  prometheus svc. With this patch the user will be forced to fill the right prometheus svc.

### DIFF
--- a/helm/grafana/README.md
+++ b/helm/grafana/README.md
@@ -43,6 +43,7 @@ Parameter | Description | Default
 `ingress.fqdn` | Grafana Ingress fully-qualified domain name | `""`
 `ingress.tls` | TLS configuration for Grafana Ingress | `[]`
 `nodeSelector` | Node labels for pod assignment | `{}`
+`prometheus.serviceURL`| Prometheus service  url. Ex. prometheus-v01.namespace.svc.cluster.local| `{}`
 `resources` | Pod resource requests & limits | `{}`
 `service.annotations` | Annotations to be added to the Grafana Service | `{}`
 `service.clusterIP` | Cluster-internal IP address for Grafana Service | `""`
--- a/helm/grafana/templates/dashboards-configmap.yaml
+++ b/helm/grafana/templates/dashboards-configmap.yaml
@@ -20,6 +20,6 @@ data:
       "basicAuth": false,
       "name": "prometheus",
       "type": "prometheus",
-      "url": "http://{{ printf "%s-%s" .Release.Name "prometheus" }}:9090"
+      "url": "http://{{ required "A valid .Values.prometheus.serviceURL entry required!" .Values.prometheus.serviceURL}}:9090"
     }
   {{- end }}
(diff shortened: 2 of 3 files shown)


brancz:
I would expect the grafana helm chart to accept datasource definitions, just like it does for dashboards. Then the helm chart can be used universally and everyone can configure them to their need.

/cc @weiwei04

weiwei04:
@gianrubio, there's a dataSource{} filed in value.yaml, If the dataSource{} is unspecified, the default dataSource is a prometheus data source, and url is like `ReleaseName-prometheus:9090`(same trick as alertmanager when deployed with kube-prometheus, the grafana by default to connect to prometheus deployed by kube-prometheus), If you want to set a customerized dataSource, just provide the value.yaml and set the dataSource filed, for example:

```yaml
# for  prometheus data source:
dataSource:
  prometheus-datasource.json: |+
    {
      "access": "proxy",
      "basicAuth": false,
      "name": "prometheus",
      "type": "prometheus",
      "url": "http://prom-svc:prom-port"
    }
```

```yaml
# for graphite data source:
dataSource:
   graphite-datasource.json: |+
     {
       "name":"test_datasource",
       "type":"graphite",
       "url":"http://mydatasource.com",
       "access":"proxy",
       "basicAuth":false
     }
```

I'm not sure if this satisfy your need? I should admit there's still some traps, like the json file must named like xxx-datasource.json. 

I can open a pr to improve this, change the value file to
```yaml
dataSources:
   - name: "dataSource0"
     type: ""
     url: ""
     access: ""
     basicAuth: ""
   - name: "dataSource1"
      ...
```

and change the dashboards-configmap.yaml like
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app: {{ template "grafana.fullname" . }}
    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
    heritage: "{{ .Release.Service }}"
    release: "{{ .Release.Name }}"
  name: {{ template "grafana.server.fullname" . }}
data:
  # ignore the grama:)
  for datasource in .Values.dataSources
  {{ .Values.dataSource.name }}-datasource.json: |+
    {
      "access": {{ .Values.dataSource.access }},
      "basicAuth": {{ .Values.dataSource.basicAuth }},
      "name": {{ .Values.dataSource.name }},
      "type": {{ .Values.dataSource.type }},
      "url": {{ .Values.dataSource.url }}
    }
```

you can set any grafana supported datasources(non't only prometheus), what do you think? @gianrubio @brancz

brancz:
Yep, everything that @weiwei04 in detail said is what I was going for. Does that work for you @gianrubio? Apart from that, regarding the suffixes, as it seems to give people difficulties, maybe we should instead of specifying directories we should specify globs directly in the grafana-watcher?

gianrubio:
@weiwei04 the second suggestion it's much more elegant. Please remember to specify examples and don't fill the default value for the prometheus service.  The prometheus service name it's based on the release name - provided by the users -  so it's always a different endpoint. 
I'm closing this PR for now

PR #561, opened 2017-08-12
Update to latest kube-state-metrics

ryanwalls:
I updated to this release on our QA cluster.  I visually checked the kube-prometheus grafana dashboards and prometheus alerts and didn't see anything broken.  So I think it is safe to default to v1.0.0.

### DIFF
--- a/contrib/kube-prometheus/assets/prometheus/rules/kubelet.rules
+++ b/contrib/kube-prometheus/assets/prometheus/rules/kubelet.rules
@@ -1,5 +1,5 @@
 ALERT K8SNodeNotReady
-  IF kube_node_status_ready{condition="true"} == 0
+  IF kube_node_status_condition{condition="Ready", status="true"} == 0
   FOR 1h
   LABELS {
     severity = "warning",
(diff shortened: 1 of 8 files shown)


brancz:
I would expect there need to be changes for RBAC enabled clusters. Additionally we recently got around adding the addon-resizer to the kube-state-metrics upstream project which I think is worth integrating here (although maybe with a separate PR).

brancz:
Good catch with the alerting rules! Could you just update the Cluster role associated with kube-state-metrics? https://github.com/kubernetes/kube-state-metrics/blob/master/kubernetes/kube-state-metrics-cluster-role.yaml

brancz:
Note that the rules etc. all need to be modified in the `contrib/kube-prometheus` directory first or simultaneously.

ryanwalls:
@brancz What all do we need to add?  `jobs`, `deployments`?  Anything else?

brancz:
I typically just copied the entire object, that way the diff should be pretty readable :slightly_smiling_face:

ryanwalls:
@brancz Copy from which object?  Is there an example object somewhere that has all the right permissions?

brancz:
Yes, we maintain the correct roles in the upstream project: https://github.com/kubernetes/kube-state-metrics/blob/master/kubernetes/kube-state-metrics-cluster-role.yaml

ryanwalls:
@brancz Ah, got it.  That's why you linked it earlier.  Thanks!  Will try to get to it today.

ryanwalls:
@brancz How does it look now?

brancz:
yet lgtm thanks!

PR #562, opened 2017-08-12
allow params to be set for metrics endpoints

rubbish:
On [the prometheus configuration page](https://prometheus.io/docs/operating/configuration/), there is a config for setting HTTP GET params. This is required if one wants to federate prometheuses _(promethei?)_. I had tried just making the ServiceMonitor's path like `/federate?match[]={__name__=~".+"}`, but it looked as if it was getting URL encoded, unless I'm doing this wrong somehow.

With this PR, then it'd be just setting up a `ServiceMonitor` to federate another prometheus and grab all metrics like so

```yaml
kind: ServiceMonitor
metadata:
  name: k8s-prometheus
  labels:
    team: infra
spec:
  selector:
    matchLabels:
      prometheus: k8s
  namespaceSelector:
    matchNames:
    - kube-public
  endpoints:
  - port: web
    interval: 30s
    path: '/federate'
    params:
      'match[]':
      - '{__name__=~".+"}'      
    honorLabels: true
```
Not that I'm a fan of prom's `params` syntax, but that's how it could be done.

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -91,7 +91,7 @@ Endpoint defines a scrapeable endpoint serving Prometheus metrics.
 | targetPort | Name or number of the target port of the endpoint. Mutually exclusive with port. | intstr.IntOrString | false |
 | path | HTTP path to scrape for metrics. | string | false |
 | scheme | HTTP scheme to use for scraping. | string | false |
-| params | Optional HTTP URL parameters | map[string]\[]string | false |
+| params | Optional HTTP URL parameters | map[string][]string | false |
 | interval | Interval at which metrics should be scraped | string | false |
 | scrapeTimeout | Timeout after which the scrape is ended | string | false |
 | tlsConfig | TLS configuration to use when scraping the endpoint | *[TLSConfig](#tlsconfig) | false |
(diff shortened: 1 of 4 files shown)


brancz:
Sorry @rubbish ! I completely missed this PR, I apologize for that. We recently moved all objects from v1alpha1 to v1, so this PR needs to be updated, to instead include the field in the v1 objects.

rubbish:
@brancz All rebased and updated. I tested this out locally in trying to federate another prometheus and it worked great!

brancz:
@coreosbot test this please

Documentation/api.md
```diff
@@ -91,6 +91,7 @@ Endpoint defines a scrapeable endpoint serving Prometheus metrics.
 | targetPort | Name or number of the target port of the endpoint. Mutually exclusive with port. | intstr.IntOrString | false |
 | path | HTTP path to scrape for metrics. | string | false |
 | scheme | HTTP scheme to use for scraping. | string | false |
+| params | Optional HTTP URL parameters | map[string]\[]string | false |
```
brancz: Unfortunately the doc generation tests failed here because they expected `map[string][]string` instead of `map[string]\[]string`, but I'm guessing you tested this out which is why the `\` is in there. If that's the case, we'll need to adapt our code for doc generation.

rubbish: @brancz I didn't realize that the docs were generated from the types. I had thought a backslash was necessary due to markdown reference links. Turns out that's not the case and it's ok without the backslash. 

Now this should be good to go.

brancz:
lgtm :+1: thanks a lot! And sorry for the detours and taking so long!

PR #565, opened 2017-08-14
add rbac policy kube-state, sidecar container addon-resizer into kube-prometheus helm chart

gianrubio:


brancz:
Out of curiosity @mgoodness, I'd expect kube-state-metrics to be in it's own helm chart and then only included in the kube-prometheus chart. Is that planned and we just haven't gotten around to doing this?

brancz:
Could you update this to the RBAC roles required for v1.0.0?

gianrubio:
@brancz rbac review done! I also add the sidecar container addon-resizer

PR #567, opened 2017-08-15
Feature/grafana configmap generator

eedugon:
For @brancz :
I have included the grafana configmap generator tool and I have integrated it with a very simple modification in generate-manifests.sh file.

generate-manifests.sh file will now call the new tool instead of generate-dashboards-configmap.sh script.

Please take a look and let me know your thoughts.

I have made changes in the tool so it can be easily handled by "make generate" or be used as an standalone tool to create multiple configmaps based on the dashboards of a given directory.

I would propose to test this to generate multiple configmaps (the size limit can be reduced easily with -s option) and check if the new grafana watcher supports the multiple configmaps.


eedugon:
@brancz : I think that maybe you prefer all  the code in the same file (directly into hack/scripts).
If so, let me know and I will adapt it.
But for the moment you can test this, for example raising a 'make generate'.

Regards!

contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/bin/grafana_dashboards_generate.sh
```diff
@@ -0,0 +1,317 @@
+#!/bin/bash
```
gianrubio: please add this lines on the header of these shell scripts so it always fail fast

```
# exit immediately when a command fails
set -e
# only exit with zero if all commands of the pipeline exit successfully
set -o pipefail
# error on unset variables
set -u
```

brancz:
Great work! Thanks a lot for your effort!

I don't have a lot of time this week as I'm traveling to a conference and doing talks, so I don't know when I'll get around to reviewing this.

Just to make sure that we're all aligned with the workflow, @unguiculus as you submitted the patches for the grafana watcher to work with multiple directories, could you validate that this works, and or whether you think that this fits into the existing workflow?

unguiculus:
I'll try and test it but I'll be at PromCon as well.

eedugon:
@unguiculus : If you like I could test it also next week, just let me know how to verify that grafana-watcher is fetching the dashboards from multiple configmaps.
For my tool to generate multiple configmaps we just need to add a few dashboards to the assets/grafana-dashboards directory before running `make generate'.

contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/bin/grafana_dashboards_generate.sh
eedugon: @gianrubio : Thanks for your comment, I will do the changes next week and update the PR, this week I will be away on vacations.

Also i'm thinking on doing more small changes to have the same philosophy as the rest of your hack/scripts scripts... for example now i'm logging some lines with messages about the processed dashboards and the amount of configmaps generated due to the size limit.

Do you want me to remove the output? (if you have a minute just run `make generate` with my branch to take a look).

The reason of not using `set -e` was also that my script is full of preliminary verifications before the payload.

unguiculus:
@eedugon It does support multiple directories and thus multiple configmaps.

eedugon:
quick update: I will test this early next week, as I'm still out on vacations ;-)

By the way, @unguiculus, do I need to define every configmap in the grafana or grafana watcher deployment definition?

Because I think the plan here (@brancz, correct me if I'm wrong) is to define in grafana yaml a configmap called grafana-dashboards, but internally create multiple dashboards labeled grafana-dashboards-0, grafana-dashboards-1, etc.

I'm sorry for asking too many things (or the same too many times :) ), just want to check if we are aligned on the expectations before testing.

brancz:
Yes I would expect a higher level abstraction to take care of the configuration of multiple configmaps being mounted and configured in the grafana-watcher, this could be a or the grafana operator, which at this point will have to exist inevitably due to the high demand around grafana. I'll start some discussions internally whether we have any bandwidth to maintain it, but given the high interest I can also see the community taking some higher ownership in it as well. I have a lot of ideas in that direction, some I've already documented in [this](https://github.com/coreos/prometheus-operator/issues/115) issue, and I'll update with my latest thoughts when I find some time.

eedugon:
Hi @unguiculus , @brancz : As mentioned in this [ticket](https://github.com/coreos/prometheus-operator/issues/535) , I have verified the tool with today's master repository, which contains the multiple configmap support developed by @unguiculus .

They work fine, but after creating multiple configmaps we have to update grafana watcher configuration and volumes to be able to fetch them, and that wasn't the original objective.

Anyway I do the requested changes by @gianrubio into `hack/grafana-dashboards-configmap-generator` script and if you agree we could merge the tool for future usage but not integrating it into the current `generate-manifests` process until we decide what to do with `grafana-watcher`.

Any thought? :)

brancz:
Sorry I was out sick monday and tuesday, I basically already answered on the other issue, I think we can just generate the Grafana deployment yaml as well with the scripts that way we can make sure the manifest is in sync with the generated dashboard manifests.

brancz:
Do you think that would be something for a follow up PR? These changes generally lgtm, so I'm happy to merge here and handle the integration with the other concerns in a follow up PR, what do you think?

eedugon:
@brancz : We cannot merge exactly this, because if for any reason `make generate` generates multiple configmaps (with my code), then none of them will be visible as grafana wouldn't be configured with them.

Tomorrow I will update this PR, including the necessary changes in grafana manifest to have all aligned, then at least we could merge without problems.

contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/bin/grafana_dashboards_generate.sh
eedugon: @gianrubio : changes done and script tested. One quick comment... after adding set -e I had to add `|| true` in this type of commands:
```
    (( total_files_processed++ )) || true
```
please take a look to the code now and let me know your thoughts.

eedugon:
@brancz , @unguiculus : here you have... please take a look to the difference between the original grafana-deployment.yaml and the grafana-deployment.yaml that we will generate with for example 2 configmaps. Do you think we need more changes?
```
$ diff manifests/grafana/grafana-deployment.yaml grafana_test22.yaml
46c46,47
<           - '--watch-dir=/var/grafana-dashboards'
---
>           - '--watch-dir=/var/grafana-dashboards-0'
>           - '--watch-dir=/var/grafana-dashboards-1'
67,68c68,71
<         - name: grafana-dashboards
<           mountPath: /var/grafana-dashboards
---
>         - name: grafana-dashboards-0
>           mountPath: /var/grafana-dashboards-0
>         - name: grafana-dashboards-1
>           mountPath: /var/grafana-dashboards-1
72c75
<       - name: grafana-dashboards
---
>       - name: grafana-dashboards-0
74c77,80
<           name: grafana-dashboards
---
>           name: grafana-dashboards-0
>       - name: grafana-dashboards-1
>         configMap:
>           name: grafana-dashboards-1
```
@brancz : for more consistency, even if we generate only 1 configmap I prefer to name it `grafana-dashboards-0` instead of `grafana-dashboards`. Is it okay for you?

So, I have updated the default `unique` configmap to be called grafana-dashboards-0.

If we really want to have `grafana-dashboards` label when only one configmap is created then we need to add more logic.

brancz:
I think having `grafana-dashboards-0` is perfectly fine. I'm hoping to getting to review this today, but can't promise. Fantastic work though!

brancz:
Flying over it, generally looks good to me, I'll make sure to try it out tomorrow. Also feel free to delete the "original" grafana deployment, after this we have no use for it anymore :)

eedugon:
@brancz : About the "original" `grafana-deployment.yaml`, this PR is already replacing it with content generated with the new code, including 1 configmap called `grafana-dashboards-0`.

contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/templates/grafana-deployment-original.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: grafana
```
brancz: I'm talking about this file regarding the "original" deployment, I think this might be here by mistake as I can't see it being used anywhere. I might be wrong though and just can't see where it's used :slightly_smiling_face: .

eedugon: you are right :-) removed!
one important thing to agree: If we follow up this approach, now the main grafana manifest to maintain and keep updated will be `grafana-deployment-template.yaml`.

If anyone has a better way to adapt the grafana-deployment manifest to multiple configmaps  than the `sed` commands I'm using please let me know...

brancz: I'm perfectly fine with the `sed` approach you took, the only sort of precedence where we're doing "templating" [here](https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/hack/scripts/generate-grafana-credentials-secret.sh) close to this, but was far simpler. So I think the way you did is just fine.

brancz:
Once more, fantastic job, thanks a lot! lgtm :+1:

Issue #553, opened 2017-08-08
Use SIGHUP instead of http reload


brancz:
The prometheus-config-reloader currently reloads Prometheus by doing a `POST` request to the reload endpoint. As noted by a user, this endpoint is by default disabled in Prometheus 2.0, however, as the SIGHUP handler is available in 1.x and 2.x either way it seems more sane to reload via SIGHUP then expose these endpoints.

brancz:
It turns out there are currently some technical limitations on this, primarily that the currently supported version of Docker in Kubernetes is not able to share PID namespace, so closing this for the time being.

Issue #480, opened 2017-07-13
Unable to deploy `v2.0.0-beta.0`


george-angel:
We are currently running `1.7.1`.

Having updated the `version` of the `Prometheus` type to `v2.0.0-beta.0`, I have then deleted the `StatefulSet`. New one has the correct image: `quay.io/prometheus/prometheus:v2.0.0-beta.0` and following args:

```
      - args:
        - -config.file=/etc/prometheus/config/prometheus.yaml
        - -storage.local.path=/var/prometheus/data
        - -storage.tsdb.no-lockfile
        - -storage.tsdb.retention=72h
        - -web.external-url=<<address>>
```

But is unable to start:

`prometheus: error: unknown short flag '-c'`

fabxc:
We changed our flags to use double dashes (`--`) but forgot to adjust those in the default entry point of our docker container. This is already fixed in HEAD of `dev-2.0`.

In the meantime, overriding the entry point when starting your container should work.
Additonally, `-storage.local.path` is `-storage.tsdb.path` now.

george-angel:
@fabxc thanks for replying, when you speak of `entrypoint` what do you refer to, I only have access to: https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#prometheusspec for modification using the operator.

Which creates the `StatefulSet`. Are you suggesting using our own version of `baseImage` with overridden `entrypoint`?

Thanks

fabxc:
Oh, my bad – I thought this issue was filed in prometheus/prometheus. For the operator it's our job to make this transition of course.

george-angel:
Shouldn't https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L318 have two dashes?

vnandha:
@fabxc Is the flag `storage.local.target-heap-size` is supported in 2.x ?  (looks like it is not from the --help). 

What is the alternate way/flags to control memory?

brancz:
@vnandha the new Prometheus storage is `mmap`ed which means that the operating system will take care of evicting memory and therefore additional resource control by Prometheus should not be necessary anymore! :slightly_smiling_face: 

Regarding what is happening in this reported issue, I'll have to investigate and report back. (I just got back from vacation and have a good size backlog :slightly_smiling_face:)

vnandha:
[...]

PR #568, opened 2017-08-16
Enable Lifecycle API on Prometheus 2.0.0-beta.0 and up

Capitrium:
The Lifecycle API is disabled by default starting with Prometheus v2.0.0-beta.0, causing the config reloader to error out:

```
ts=2017-08-15T19:12:35Z caller=main.go:214 component=volume-watcher msg="Reloading Prometheus temporarily failed." err="Received response code %!s(int=403), expected 200" next-retry=25.169469033s
```

This PR adds a flag to Prometheus to enable the Lifecycle API on v2.0.0-beta.0 and higher, allowing the config reloader to work correctly:

```
ts=2017-08-16T15:59:20Z caller=main.go:219 component=volume-watcher msg="Prometheus successfully reloaded."
```

### DIFF
--- a/contrib/prometheus-config-reloader/main.go
+++ b/contrib/prometheus-config-reloader/main.go
@@ -20,12 +20,10 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
-	"strconv"
-	"syscall"
 	"time"
 
 	fsnotify "gopkg.in/fsnotify.v1"
@@ -187,21 +185,17 @@ func (w *volumeWatcher) writeConfigMapFile(filename, content string) error {
 }
 
 func (w *volumeWatcher) ReloadPrometheus() error {
-	cmd := exec.Command("/bin/sh", "-c", "ps cax | grep '/bin/prometheus ' | grep -v grep | grep -o '^[ ]*[0-9]*'")
-	res, err := cmd.Output()
+	req, err := http.NewRequest("POST", w.cfg.reloadUrl, nil)
 	if err != nil {
 		return err
 	}
-
-	pid, err := strconv.Atoi(strings.TrimSpace(string(res)))
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return err
 	}
-
-	err = syscall.Kill(pid, syscall.SIGHUP)
-	if err != nil {
-		w.logger.Log("msg", "Error sending SIGHUP to process", "err", err, "pid", pid)
-		return err
+	resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("Received response code %s, expected 200", resp.StatusCode)
 	}
 	return nil
 }
--- a/contrib/prometheus-config-reloader/main.go
+++ b/contrib/prometheus-config-reloader/main.go
@@ -20,10 +20,12 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
+	"strconv"
+	"syscall"
 	"time"
 
 	fsnotify "gopkg.in/fsnotify.v1"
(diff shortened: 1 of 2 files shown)


brancz:
Actually our intention was to not expose the lifecycle APIs at all, as they have a security risk, for obvious reasons. Instead the idea was to change how the prometheus-config-reloader performs reloading in the first place, as in rather than calling the reload HTTP endpoint the sidecar can just send the SIGHUP signal to the process. We're tracking this in #553. If you want to I can guide you through implementing that if you need guidance. Feel free to do that in this PR if you are interested in taking this over. Wdyt? :slightly_smiling_face:

Capitrium:
Ah, fair enough. I ran into this while testing the latest Prometheus beta and #480 mentioned enabling the lifecycle API, but a SIGHUP works just as well. I may take a shot at that if I can find some time for it in the next couple of days.

brancz:
No worries. Let us know if you want us to take over, no problem! :slightly_smiling_face:

Capitrium:
@brancz I've modified the reloader to send a SIGHUP instead of an HTTP POST, seems to be working fine; not sure if there's a better way to enable signalling between containers other than using `hostPID`, although I'm sure the code to get the prometheus pid could be cleaner :stuck_out_tongue:

mikebryant:
Isn't using `hostPID` a big escalation in how much access we're giving prometheus?

We're currently only allowing admin level things that level of access in our clusters, not user level monitoring tools.

contrib/prometheus-config-reloader/main.go
```diff
@@ -185,17 +187,21 @@ func (w *volumeWatcher) writeConfigMapFile(filename, content string) error {
 }
 
 func (w *volumeWatcher) ReloadPrometheus() error {
-	req, err := http.NewRequest("POST", w.cfg.reloadUrl, nil)
+	cmd := exec.Command("/bin/sh", "-c", "ps cax | grep '/bin/prometheus ' | grep -v grep | grep -o '^[ ]*[0-9]*'")
```
brancz: `ps` is not really magic, we can just read the procfs and find the process ourselves without relying on `ps` and `grep`.

mikebryant: Also, what happens if there's more than one prometheus instance on the same host?

brancz: The PID namespace is shared only within containers of the same Pod, so selecting the first process running the `/bin/prometheus` binary should be safe.

brancz: Ah I just saw the hostPID part! Thanks for pointing it out. Yes that's not something we want to do, it is definitely an escalation risk.

brancz:
Turns out this space is not as easy as it looks at first sight. I talked to the container linux team and they told me that this is a fragile piece, as for example docker and rkt handle these situation differently. Only starting Docker 1.13.1 (which is not supported by Kubernetes yet) will enable us to share PID namespace between containers of the same Pod (upstream PR regarding this: https://github.com/kubernetes/kubernetes/pull/45236).

Sorry for all the forth and back, it turns out that enabling the lifecycle API will stay being our best option for the time being. Let's go with that instead then.

Capitrium:
No worries! I kind of figured this would end up being the case; I've reverted back to simply enabling the lifecycle api.

pkg/prometheus/statefulset.go
```diff
@@ -327,6 +327,13 @@ func makeStatefulSetSpec(p v1alpha1.Prometheus, c *Config, ruleConfigMaps []*v1.
 			"-storage.tsdb.path=/var/prometheus/data",
 			"-storage.tsdb.retention="+p.Spec.Retention,
 		)
+
+		if len(version.Pre) > 0 && version.Pre[0].VersionStr == "alpha" {
```
brancz: we only need to support the latest beta, if I understand correctly, we can drop this as we don't support the alphas anymore, feel free to remove them from the compatibility matrix and only keep the latest beta

Capitrium: Done and done (figured this would be the case as well :simple_smile:)

brancz:
Thanks a lot! And sorry for the forth and back!

PR #572, opened 2017-08-17
helm/prometheus: allow for configuration of PrometheusSpec secrets

crandles:
Adds `secrets` configuration to the prometheus chart


brancz:
lgtm

PR #575, opened 2017-08-22
*: ensure using the apps v1beta1 api

brancz:
When vendoring different versions of client-go the `Apps` function can behave differently here as the version is not fixed to a specific version.

@fabxc


brancz:
ran all tests aside from the migration tests against a cluster and they passed, so moving forward here

PR #576, opened 2017-08-22
Remove port name of Grafana service for compatibility with kubectl proxy

aknuds1:
Remove the port name for the Grafana service so that you can access Kibana via `kubectl proxy` at http://localhost:8001/api/v1/proxy/namespaces/monitoring/services/grafana, without the `:web` suffix as is required if the port is named.

### DIFF
--- a/contrib/kube-prometheus/manifests/grafana/grafana-deployment.yaml
+++ b/contrib/kube-prometheus/manifests/grafana/grafana-deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
   name: grafana
--- a/contrib/kube-prometheus/manifests/grafana/grafana-service.yaml
+++ b/contrib/kube-prometheus/manifests/grafana/grafana-service.yaml
@@ -7,9 +7,9 @@ metadata:
 spec:
   type: NodePort
   ports:
-  - name: web
-    port: 3000
+  - port: 3000
     protocol: TCP
     nodePort: 30902
+    targetPort: web
   selector:
     app: grafana


contrib/kube-prometheus/manifests/grafana/grafana-deployment.yaml
```diff
@@ -27,6 +27,8 @@ spec:
             secretKeyRef:
               name: grafana-credentials
               key: password
+        - name: GF_SERVER_ROOT_URL
```
brancz: I understand your need for this, but I think this might be too opinionated, what if people prefer accessing grafana via ingress, or node port?

aknuds1: Hey @brancz! It's a valid question you're bringing up and difficult to answer. I guess it comes down to what you consider the basic way of accessing the Grafana UI. My argument is that it's analogous to Kibana in [the official Elasticsearch/Fluentd/Kibana add-on](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/fluentd-elasticsearch) to k8s, and here one has chosen to support `kubectl proxy` as the basic way of accessing it. To me it makes sense, as it works right out of the box, and it's better than the other out of the box method that I'm aware of, i.e. `kubectl port-forward` as `kubectl proxy`, unlike the former, connects to the service rather than individual pods and provides a more stable connection (`kubectl port-forward` often doesn't last long before it terminates).

Indeed I am using kube-prometheus in tandem with the official EFK add-on, so to me it makes total sense that Grafana (from the former) is configured in the same way wrt. access as Kibana (from the latter). If it's a good solution for Kibana it should also be so for Grafana.

If you don't agree with `kubectl proxy` being the default way of accessing Grafana, would you say the same for Kibana in the official EFK add-on? Should the latter have done it differently?

aknuds1: If people prefer accessing Grafana as you say via ingress or node port, it should be a matter of editing the manifests should it not, same as it is for Kibana (in the EFK add-on)?

brancz: I totally understand your argument so far we've tried to be as un-opinionated as possible (although not totally unavoidable) about the namespace things are deployed in, and Prometheus/Alertmanager currently also use node port services as examples. Something to note about the kube-prometheus manifest collection is that it's just that: a collection of manifests to start out with, it's totally meant to take and adapt :slightly_smiling_face: . The helm charts (which do admittedly need a lot of work) on the other hand are totally meant to be configurable in terms of nodeport/proxy/port-forward/ingress as the method of access. Does that make sense?

aknuds1: @brancz OK I see what you mean about other services in kube-prometheus also being node ports by default. However, are you sure it doesn't make sense to align with the official Elasticsearch/Fluentd/Kibana add-on in this regard? They want to be unopinionated as well, but as you have to take a certain stand with the default configuration, they went with supporting `kubectl proxy` as the default. 

Regardless though, does it make sense to name the service port ("web")? It was extremely tricky to figure out that the port name was the reason that the typical `kubectl proxy` URL to the Grafana service didn't work (requiring a `:web` suffix).

brancz: The thing about proxying through the apiserver is that it exactly requires this kind of odd URL prefixing and the application to know it's external accessible URL, which in turn means that we need to put this information in all of the manifests we provide which in turn hardcodes which namespace they run in, and the name of the service they are exposed by. By a user it's not obvious that if one of these parameters are changed that this opaque string needs to be changed. Aside from that it's easier to comprehend for users to add something to a manifests, than to remove something. I realize that proxying is not a bad default, but certainly not a production solution either. Once the required components are more matured (mainly the bitly oauth2 proxy) I'd like to provide samples for ingress instead of node port as I believe that's the closest to what people actually want.

Regarding your other question: I think the service port naming makes sense the way you proposed.

aknuds1:
You make some fair points @brancz, I've removed the GF_SERVER_ROOT_URL variable.

brancz:
lgtm, thanks! Let's work on the helm charts, so we can make these things properly configurable, because your case is absolutely plausible.

aknuds1:
Agreed @brancz, except that I've yet to try Helm. I've held off on it, since I'm not sure how mature/good of a solution it is. What's your take on this packaging tech?

brancz:
tbh, and this is purely my own opinion, helm is not a particularly great packaging tool, especially in the sense of a package, because what we're actually looking for (again my own opinion) is not packaging, but a tool that allows high customization and handling deployment of our manifests in a sane way.

There are two things have have worked out for us particularly well. The first is part of the value we ship with our product: Tectonic. We have operators, that make sure that everything in your cluster is in the state it should be in, and if not, it reconciles to get into that state, this includes, migrations on breaking changes between versions and zero down-time deployments for everything. This is all actual go code, and therefore allows us to customize the deployment/migration logic to any extend. This takes a certain amount of work, but the result is a very consistent and great experience.

The second thing, and this is much more in the direction of helm, is if you just have a collection of manifests of primarily stateless applications (not that it's only meant for those, but it has proven useful for those for us so far). We've found that customizing manifests using ksonnet/jsonnet has been immensely helpful. The way we use it in the Prometheus Operator is that we have multiple copies of very similar manifests in the repository, that only have a different namespace, or serviceaccount or something like that. Using ksonnet/jsonnet we can simply build a base manifest and then extend it with little snippets (see [here](https://github.com/coreos/prometheus-operator/tree/master/hack/generate)). jsonnet is turing complete, so in contrast to helm it actually allows us to do anything (although the language is a bit quirky to me, but that is probably because I'm a newcomer to it). That said, jsonnet and especially ksonnet are very early projects and there are also tools around it in the works to make the process smother and better.

My opinion, if you are starting from scratch, build a solid CI pipeline, generate consistent manifests with ksonnet and apply them to your cluster. My biggest concern around helm is it's RBAC and extended authorization/authentication story which to this day is unresolved, as tiller (a component of helm that runs in your cluster) essentially requires cluster admin RBAC access to perform its tasks, however since you are hand crafting your packages anyways in helm, you precisely know the RBAC requirements you have.

aknuds1:
Thanks for the detailed reply @brancz, it's very useful! I am very into the idea of Tectonic myself, I'm in the process of trying to roll out my own implementation of Tectonic Installer for DigitalOcean. It's not quite working yet, but hopefully I'll get there soon and gain some experience from using it in production.

brancz:
Nice! Always happy to chat! :)

PR #582, opened 2017-08-25
Fix typo for service monitoring

punitag:
The `ServiceMonitor` must have a label with key `k8s-app` (and not `k8s-apps`)


brancz:
You're absolutely right, thanks! :slightly_smiling_face:

Issue #583, opened 2017-08-27
[ kube-prometheus ] Kubernetes scheduler and controller manager pods have no service which match the related kube-prometheus servicemonitor 


csahli:
I deployed a kubernetes cluster (v1.6.7 multi master cluster) using kubespray and then deployed the kube-prometheus (v0.11). Everything was fine but I get some alerts visible on promDash UI. Alerts were related to k8s scheduler and controller unavailability.

To solve the problem I just deployed kubernetes service to target kube-scheduler and kube-controller pods with labels matching related servicemonitor label selectors.

This cleared prometheus alerts and make k8s scheduler and controller metrics available.

Here the k8s manifests I have to add:

- kube-controller-manager-service.yml

```
apiVersion: v1
kind: Service
metadata:
  labels:
    k8s-app: kube-controller-manager
  name: kube-controller-manager
  namespace: kube-system
spec:
  ports:
  - name: http-metrics
    port: 10252
    targetPort: 10252
  selector:
    k8s-app: kube-controller
```

- kube-scheduler-service.yml

```
apiVersion: v1
kind: Service
metadata:
  labels:
    k8s-app: kube-scheduler
  name: kube-scheduler
  namespace: kube-system
spec:
  ports:
  - name: http-metrics
    port: 10251
    targetPort: 10251
  selector:
    k8s-app: kube-scheduler
```

brancz:
As I mentioned in the PRs, these manifests already exist, but are highly dependent on the tool used to provision your cluster, therefore we unfortunately cannot cover all cases generically.

csahli:
yes it makes sense ... I didn't figure it out :(

brancz:
No worries that happens :slightly_smiling_face: if there is anything for your cluster that doesn't work we'd still like to hear from it, so we can make this project as compatible as possible :slightly_smiling_face: .

PR #585, opened 2017-08-27
Create prometheus-k8s-scheduler-service.yaml

csahli:
[ kube-prometheus ] Kubernetes scheduler and controller manager pods have no service which match the related kube-prometheus servicemonitor #583


brancz:
This already exists in `manifests/k8s/self-hosted/`.

https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/k8s/self-hosted/kube-scheduler.yaml

I'll close this here, but feel free to re-open if you feel it's necessary.

PR #586, opened 2017-08-27
Update deploy

csahli:
Update deploy to include k8s scheduler and k8s controller services manifests creation in kube-system namespace.

### DIFF
--- a/contrib/kube-prometheus/hack/cluster-monitoring/deploy
+++ b/contrib/kube-prometheus/hack/cluster-monitoring/deploy
@@ -27,7 +27,15 @@ kctl apply -f manifests/node-exporter
 kctl apply -f manifests/kube-state-metrics
 kctl apply -f manifests/grafana/grafana-credentials.yaml
 kctl apply -f manifests/grafana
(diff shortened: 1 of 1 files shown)


brancz:
The `hack/` scripts are not meant for anything more than it's name says, a hack :slightly_smiling_face: . The reason we explicitly left out the kube-controller-manager, and kube-scheduler is that

a) we don't know whether these components will be self-hosted in your cluster (GKE for example doesn't)
b) if it is self-hosted, then labeling is not standardized, so every tool chooses different labelings for these pods (for example kubeadm and kubespray use different labeling schemes), although we did attempt to standardize them.

The purpose of the kube-prometheus manifests is for you to take them and adapt them to your needs. They are highly opinionated, and for example node-port services to access Prometheus and Alertmanager are less than optimal in a production environment, but they are just meant to see a quick result.

Does that make sense?

brancz:
Aside that the `hack/cluster-monitoring/self-hosted-deploy` and `hack/cluster-monitoring/self-hosted-teardown` should cover your case. Could you make sure that that is true?

PS: next time it would be easier if you just create a single PR for all of this :slightly_smiling_face: .

PR #589, opened 2017-08-28
kube-prometheus: bump kube-state-metrics to v1.0.1 release

brancz:
This bumps kube-state-metrics to the latest release as well as introduced the addon-resizer for it.

@fabxc


brancz:
Verified that discovery and metric retrieval still works properly and neither kube-state-metrics nor addon-resizer has any errors in logs so RBAC roles are ok.

Issue #592, opened 2017-08-29
rework helm-charts


brancz:
I just had a chat with @mgoodness regarding the state and future of the kube-prometheus helm charts, and we wanted to share what our vision is regarding them and we invite anyone who wants to help on this to participate.

We imagine the general structure of the charts to be like this:

The kube-prometheus chart (this is mostly a "meta" chart that includes multiple other charts and adds the Prometheus alerting/recording rules, grafana dashboards as well as service-monitors.

It's dependencies are:

  * Prometheus chart (with opinionated configuration tailored for cluster monitoring, and to work with the alertmanager instance shipped)
  * Alertmanager chart
  * Prometheus Operator chart (because it's a dependency of the prometheus and alertmanager chart)
  * grafana chart (without dashboards, but preconfigured with the k8s prometheus as datasource) (this one is already actively maintained by @weiwei04)
  * kube-state-metrics chart
  * node-exporter chart

Something that we had previously done in the kube-prometheus charts is also consider the choice of how the web UIs of the above components are exposed, we'd like to move away from this and leave this completely up to the user.

If anyone is willing to work on any of these charts, contributions are highly appreciated and very welcome. The individual charts that are a dependency of the kube-prometheus chart should be relatively low maintenance as the upstream projects are not released that often.

MattiasGees:
This will be a really good way forward. 
The place I am working at for the moment is a heavy kubernetes, prometheus and helm user. I am willing to help with development of the HELM chart.
I think some of my colleagues are also willing to help with pushing the helm-charts forward and helping out the community.

Kind regards,
Mattias

brancz:
That sounds fantastic! Let me know if I help you in any way to get started :slightly_smiling_face: .

ringods:
And @MattiasGees, don't be shy to say we are working for [Skyscrapers](https://skyscrapers.eu/). 😄

kevinjqiu:
@brancz At my $DAYJOB, we're currently using the kube-prometheus manifests to deploy our prometheuses, alerts and dashboards. It's been great but we're looking into rewriting the manifests to helm charts just like you've described. We'd like to contribute to this too if you could point us to a place to get started.

Our background:
- using Kubernetes 1.6 in prod for a couple of months
- using prometheus, alert manager and grafana in multiple environments

Let me know :)

Thanks!

punitag:
similar use case to @kevinjqiu . would definitely contribute back once our infra is set up.

[...]

PR #608, opened 2017-09-08
Remove problematic line

punitag:
Seems like a typo. The line `{{ toYaml .Values.serviceMonitorsSelector | indent 4 }}` does not belong here


brancz:
I believe so too, thanks! As noted in #592 there is a lot of work to be done around the helm charts, if you are interested in using them it would be great if we could collectively work on them to make sure they're solid. :slightly_smiling_face:

PR #609, opened 2017-09-08
test: fix resource flake

brancz:
Previously we had a race where we updated the Prometheus object and didn't actually wait for the deployment to finish, but rather check whether enough pods exist, which can be the case if the deployment hasn't actually started, so the value is still the old value. Now we check periodically whether the resources are already in the state that we expect and timeout after 2 minutes.

@fabxc @mxinden


fabxc:
👍

PR #612, opened 2017-09-08
[kube-prometheus/exporter-kube-state] Use fullname template so it's consistent with other exporters

kevinjqiu:
Part of: https://github.com/coreos/prometheus-operator/issues/592


brancz:
lgtm, thanks!

PR #621, opened 2017-09-15
Convert Grafana dashboards to grafanalib scripts

aknuds1:
Convert Grafana dashboards from raw JSON files to [grafanalib](https://github.com/weaveworks/grafanalib) Python scripts for far easier reading and modification.

Everything works well as far as I can see, with the exception that the API Server Request Error Rate panel of the Kubernetes Control Plane Status dashboard gives a "Multiple Series" error. I haven't been able to figure out why this happens, there are only negligible differences between the JSON I generate and the original JSON for that dashboard. Maybe someone can help?

I also plan to refactor and DRY up the implementation in a second PR, but I don't want to fiddle with this before this branch is integrated into master :)

**Note:** It's necessary to use my [fork](https://github.com/aknuds1/grafanalib) of grafanalib, since I've had to make changes in order to generate correct Grafana dashboards.


mikebryant:
Instead of being required to use a fork - can the required changes be merged into the upstream grafanalib?

aknuds1:
@mikebryant Fair question! I have [a PR](https://github.com/weaveworks/grafanalib/pull/74) to grafanalib for the changes I'm making in order to support kube-prometheus, but it's much more efficient if I don't have to solve two problems at once. I've made some radical changes to grafanalib to be able to generate functioning graphs for Prometheus Operator, and I'm not sure if I'm in the end will be able to retain compatibility with the grafanalib interface.

As such, in my opinion it's better to retain the grafanalib fork for kube-prometheus' purposes, and concentrate on getting that right. Then we can think of consolidating our grafanalib fork back into upstream. This is a potentially lengthy process, so better concentrate on the kube-prometheus part first IMO.

aknuds1:
@mikebryant I solved it by depending on tag v0.4.0 of my fork of grafanalib, in [requirements.txt](https://github.com/aknuds1/prometheus-operator/blob/grafanalib/requirements.txt#L1). Do you agree with this solution? It should be fine from what I can tell.

brancz:
Finally got around testing this, and it works like a charm! Thanks a lot for taking care of this, it is a huge help for us @aknuds1 !

lgtm :+1:

aknuds1:
Thank you so much @brancz - I'm super happy you guys integrated my work! 😀

aknuds1:
I can look to get my fork merged back into grafanalib as well - I have a use case for reference, and time on my hands again :)

PR #629, opened 2017-09-21
prometheus: make node sync periodic instead of watching

brancz:
@fabxc


redbaron:
Just curious , what was wrong with node watchers?

brancz:
We still have to investigate the core of the issue, but the leak seemed to come from the StreamWatcher of the Node Informer. This "solution" merely avoids that code path, but we still have to investigate why this is happening, we have planned to do so over the coming weeks, but fix the leak with a pragmatic approach for now.

PR #630, opened 2017-09-21
prometheus: make node sync periodic instead of watching

brancz:
@fabxc


brancz:
Manually ran tests due to difficulties in the test infra, this passes all tests. Moving forward as we already merged #629 has already been merged and released as well.

PR #637, opened 2017-09-22
v1 in API in helm charts

redbaron:
What is the status of helm charts provided in this repo? They seem to see no love. If they are not supported maybe removing them is an option?

### DIFF
--- a/helm/alertmanager/templates/alertmanager.yaml
+++ b/helm/alertmanager/templates/alertmanager.yaml
@@ -1,4 +1,4 @@
-apiVersion: monitoring.coreos.com/v1alpha1
+apiVersion: monitoring.coreos.com/v1
 kind: Alertmanager
 metadata:
   labels:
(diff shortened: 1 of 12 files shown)


brancz:
They currently purely rely on community contributions. #592 describes an effort where we want to get them into shape, if you are interested in working on that I'm more than happy to help out. In fact I've already started contributing to the stable upstream charts for the kube-state-metrics chart, as it is unnecessarily duplicated here. I will soon start on creating a chart for the node-exporter as well, however, this is purely in my free time, as we're not using helm as the mean of deploying these in CoreOS Tectonic. I'd like to see the helm charts get into shape, but it will have to be something we need to work on together. If you are interested in improving them to a point where you're happy with them I'm more than happy to review and merge PRs.

I have also thought of removing them from this repository, but am still hopeful, that we may be able to get them into shape and once they are, ensure that they're up to date going forward.

brancz:
In terms of this PR, it lgtm :+1:

redbaron:
Seems that examples are kept very much up to date, is it what you use to install prometheus-operator & co in Tectonic? I'll sync helm charts with examples folder and make a separate PR

brancz:
We have the same operator pattern for the Tectonic Monitoring stack, as in we reconcile everything to be in the state we expect it to be in based on an object/configuration. The content of `contrib/kube-prometheus/` is effectively our place to try these out as plain manifests and then we sync them into Tectonic and add proprietary parts and more opinion as we can make more assumptions about Tectonic clusters.

You can also feel free to change the architecture of the helm charts where you see fit. Currently the leaf charts contain the alerting rules and service monitors, whereas, they should really live in the meta kube-prometheus chart as they're a result of the combination of the lower level charts.

PR #646, opened 2017-09-26
tests: Reduce amount of kept builds

mxinden:
This PR reduces the amount of builds again. In addition it keeps all PR
builds for at least one month, so long running community PRs don't loose
their test logs.

@brancz Let me know if these numbers are still ok for you?


brancz:
yay, green build :green_apple:

PR #652, opened 2017-09-28
kube-prometheus: add tolerations to node-exporter chart

atopuzov:
* add tolerations to allow node-exporter running on all nodes
* Sync from e87acbd88b57fe2cf8dfcb7e403a68e3eb3444d5


brancz:
lgtm, thanks!

PR #653, opened 2017-09-29
tests: Add cleanup job to remove old docker images periodically

mxinden:
As our Jenkins worker is running out of disk space quickly, we should remove all unneeded docker images periodically.

@bison Thanks for the hint!


brancz:
lgtm :+1: thanks!

PR #654, opened 2017-09-29
charts: Allow custom labels on prometheus and alertmanager

atopuzov:
* Allows the user to specify labels on the following resources: service, ingress, alertmanager and prometheus

For me personally came from the need to add `dns: route53` to the service of both alertmanager and prometheus.


brancz:
lgtm, thanks for all your contributions! Highly appreciated!

PR #655, opened 2017-09-29
kube-prometheus: Move rules to values.yaml

atopuzov:
* **WARNING**: Requires `helm` >= 2.5.0
* Moves alerting rules to `values.yaml`

Moving the rules to `values.yaml` allows users to change the rules (greater flexibility as otherwise a modification to chart is necessary). Because rules themselves use helm templating (for job names, etc.) `tpl` function is needed which got added in `helm` 2.5.0 [1].

To verify that the same output is generated I used helm template plugin [2] (only difference in whitespace [3]) and to verify it's working I installed to minikube.

[1] https://github.com/kubernetes/helm/commit/d01f7978d224a9da96a3fcede3d2d181c3b6cbc9
[2] https://github.com/technosophos/helm-template
[3] https://gist.github.com/atopuzov/ff9bc797984a97d040fc56ef43bff381


gianrubio:
@atopuzov not really related to this PR, but do you think you could keep in sync the [examples rules](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus/assets/prometheus/rules) with these helm charts?

atopuzov:
@gianrubio As a first step I can verify if that the [rules](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus/assets/prometheus/rules) are in sync with the ones in the chart and then later perhaps write some automation that gets the rules and puts them into charts.

brancz:
So sorry, I just merged #565, and didn't realize it would cause conflicts. Could you resolve them? Then this lgtm.

atopuzov:
@brancz Resolved. But please wait with the merge I'll add a README that says helm >= 2.5.0 is required.

brancz:
I think I would like to eventually move all the alerting rules and servicemonitors out of the leaf charts, and have them in the kube-prometheus meta chart, as some of the alerting rules really only make sense when used in combination with the node-exporter and or kube-state-metrics chart. That way we can keep the leaf charts on topic and just deploy exactly the workload they represent, maybe with a cluster IP service, but not more. Then those charts should actually even go into the upstream stable charts (at least the node-exporter, kube-state-metrics, prometheus-operator charts). The kube-prometheus chart(s) probably are too broad to be an upstream chart, on the other hand there are similar stable charts.

atopuzov:
Good on my side. Thanks!

brancz:
Great, thanks a lot for all your efforts! They are highly appreciated!

PR #662, opened 2017-10-03
kube-prometheus: remove app label from servicemonitors

atopuzov:
* Removes `app=prometheus` from servicemonitor objects
* Also removes it from the serviceMonitor slector for prometheus
* This allows alertmanager servicemonitor to be selected
note: app on alertmanager is set to alertmanager


brancz:
lgtm, thanks!

PR #665, opened 2017-10-04
Add parallel pod management

### LINKED ISSUE #491: Use `podManagementPolicy` for the StatefulSets
In 1.7, a new field `podManagementPolicy` is available.
https://kubernetes.io/docs/api-reference/v1.7/#statefulsetspec-v1beta1-apps

For Prometheus and AlertManager, we could use `podManagementPolicy: Parallel` to get faster scale up.
This would also help out in a situation where the AZ where the prometheus-0 volume is located is unavailable. (This currently means the whole prometheus cluster is down, even though we set replicas to 2, as it waits for -0 to start up before it starts -1)

In `Parallel` mode they'd all start up, which would be much better for us. And as the daemons don't interact, it shouldn't cause any problems.

mikebryant:
Part of #491 (Prometheus only)

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -459,8 +459,9 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMaps []
 	podLabels["app"] = "prometheus"
 	podLabels["prometheus"] = p.Name
 	return &v1beta1.StatefulSetSpec{
-		ServiceName: governingServiceName,
-		Replicas:    p.Spec.Replicas,
+		ServiceName:         governingServiceName,
+		Replicas:            p.Spec.Replicas,
+		PodManagementPolicy: v1beta1.ParallelPodManagement,
 		UpdateStrategy: v1beta1.StatefulSetUpdateStrategy{
 			Type: v1beta1.RollingUpdateStatefulSetStrategyType,
 		},


mikebryant:
cc @brancz @julianvmodesto

mikebryant:
I haven't yet verified this works for us. (Though since all the tests pass, and this is a strongly typed api, I don't see why it wouldn't..)

I'm hoping the existing integration tests will validate this for me :p

brancz:
@coreosbot test this please.

Thanks for taking care of this!

I think we're actually ok with just doing this by default and removing all the version compatibility checks, as starting with v0.12.0 we only supported v1.7.0+, and the PodManagementPolicy feature was introduced with 1.7.0.

mikebryant:
@brancz Ah, ok. Shall I take care of removing the `StatefulSetUpdatesAvailable` flag too then? :)

brancz:
> Shall I take care of removing the StatefulSetUpdatesAvailable flag too then? :)

That would be great if you're already at it :) 

> Didn't realise 1.6 support had been dropped

We mentioned it in the release notes, but it wasn't visible enough and multiple people didn't realize it, so we also added a header to all docs pointing this out (unfortunately this header has not landed on coreos.com yet, but is on all docs in the repo).

mikebryant:
Done in #666 

This PR is now very simple :)

brancz:
As mentioned in my last comment on #666, this will need a rebase so we can run the tests successfully again. Thanks!

mikebryant:
@brancz done :)

brancz:
Wow, thanks for the quick action!

@coreosbot test this please.

brancz:
It seems, that with the Alertmanager we can't quite do the parallel strategy yet.

Excerpt from the logs:

```
--- FAIL: TestMeshInitialization (27.36s)
	alertmanager_test.go:152: timed out waiting for the condition
--- PASS: TestAlertmanagerVersionMigration (23.62s)
--- PASS: TestExposingPrometheusWithKubernetesAPI (6.31s)
2017/10/05 11:49:24 Expected Targets: []string{"http://10.2.3.18:9090/metrics"}
--- PASS: TestPrometheusDiscoverTargetPort (12.98s)
2017/10/05 11:49:32 retrieving stateful set failed: statefulsets.apps "prometheus-test" not found
2017/10/05 11:49:34 retrieving stateful set failed: statefulsets.apps "prometheus-test" not found
--- PASS: TestAlertmanagerScaling (41.81s)
2017/10/05 11:49:36 retrieving stateful set failed: statefulsets.apps "prometheus-test" not found
--- FAIL: TestAlertmanagerZeroDowntimeRollingDeployment (86.47s)
	alertmanager_test.go:394: One notification expected, but 2 received.
		
		Alertmanager Notification Payload Received
		Alertmanager Notification Payload Received
```

Let's revert to the default strategy for Alertmanager clusters for now and just use it for the Prometheus clusters.

mikebryant:
@brancz Sure, done :)

brancz:
@coreosbot test this please

brancz:
Thanks a lot for the contributions!

PR #670, opened 2017-10-05
Fix merging of -datasource.json files

valdisrigdon:
The merge was missing a new line at the end of the merge, so only
the first datasource was being put into the ConfigMap.

### DIFF
--- a/contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/bin/grafana_dashboards_generate.sh
+++ b/contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/bin/grafana_dashboards_generate.sh
@@ -199,6 +199,7 @@ addArrayToConfigMap() {
 
     # Dashboard foot
     test "$type" = "dashboard" && cat $DASHBOARD_FOOT_FILE
+    [ "$(tail -c 1 "$file")" ] && echo
   done
   echo "---"
 


brancz:
@eedugon can you have a look at this?

This generally lgtm, just checking to make sure I'm not missing anything :slightly_smiling_face:

eedugon:
Hi!

The change looks perfectly fine to me, but could anyone show me the original problem? @valdisrigdon, would you?
Becasue I haven't seen any problem with the previous code.
Maybe the extra `datasource` file that you were using was missing the endline character at the end of the file?
I didn't test the script before with multiple datasources, so maybe there was really a problem, but I haven't seen it.

valdisrigdon:
This is what a snippet looked like before the change:

```
...

es-logs-datasource.json: |+
    {
      "orgId": 1,
      "name": "elasticsearch-logs",
      "type": "elasticsearch",
      "access": "proxy",
      "url": "https://foo",
      "password": "",
      "user": "",
      "database": "[logstash-]YYYY.MM.DD",
      "basicAuth": false,
      "isDefault": false,
      "jsonData": {
        "esVersion": 5,
        "interval": "Daily",
        "timeField": "@timestamp"
      }
    }  prometheus-datasource.json: |+
    {
        "access": "proxy",
        "basicAuth": false,
        "name": "prometheus",
        "type": "prometheus",
        "url": "http://prometheus-k8s.monitoring.svc:9090"
    }
```

See where `prometheus-datasource.json:` isn't on a newline, causing that datasource to not be imported.

eedugon:
Got it... And don't you think the reason of the problem could be a missing newline character in the last line of the file 'es-logs-datasource.json' ?

With your proposed change, and if I'm right, then in case the original file is 'correct' (proper newline at the end of the file) we would be adding an extra empty line between both datasources, and I'm not sure if that will work, but in any case it won't look nice.

Let me know your thoughts, and please check your datasource file.

I will do some quick tests, i think we can check if the file is not ended properly and add the newline only when it's needed.

Regards!

brancz:
> I will do some quick tests, i think we can check if the file is not ended properly and add the newline only when it's needed.

This feels like it's the most appropriate and smoothest solution for the majority of users.

eedugon:
@valdisrigdon : Could you check if this solves also the issue you faced? (adding at the beginning of the loop a verification to see if final newline character is missing).
```
  for file in ${to_process[@]}; do
    # check that file exists
    test -f "$file" || { echo "# INTERNAL ERROR IN ARRAY: File not found: $file"; continue; }
    # If file is not ended properly we add newline character
    [ "$(tail -c 1 "$file")" ] && echo "" >> "$file"   
```
Only the last 2 lines above are the added lines.
If this doesn't solve the problem then check if your file was `DOS` formatted instead of `Unix`, but I hope this will also fix the issue and will be more generic.

Very good finding anyway, I didn't consider this scenario.

valdisrigdon:
That seems to fix the problem as well. My file has always used `LF`, so it's not `DOS` formatted.

To be clear, I replaced my oneliner with:
```[ "$(tail -c 1 "$file")" ] && echo```

**Comment by brancz** [comment]
That solution sgtm. Could you update the PR @valdisrigdon ? Then @eedugon gives a final review, and then we can probably merge. Thanks for bearing with us :slightly_smiling_face: .

**Comment by eedugon** [comment]
Yeah, @valdisrigdon : please update the PR, and add the new code just after the main `cat $file` , because we want to fix potential `dashboards` wrongly ended, not only `datasources` (which was this case)
Again, thanks for finding this out. I think this is a good improvement.

If for any reason you don't want to change this PR then feel free to cancel it and I could add the fix in a new PR.

**Comment by valdisrigdon** [comment]
@eedugon Sorry for the delay. I've updated and tested again.

**Comment by brancz** [comment]
Thanks a lot! lgtm :+1:

**Comment by eedugon** [comment]
HI @brancz , i'm sorry for my late response,

The change looks good, however the added command should be added one row before, because of this:
With the current code:
```
    # File content: Indent 4
    cat $file | sed "s/^/    /"

    # Dashboard foot
    test "$type" = "dashboard" && cat $DASHBOARD_FOOT_FILE
    [ "$(tail -c 1 "$file")" ] && echo
```
If file was a `datasource` then we are fine and we have fixed the discovered issue.
But if the file was a `dashboard` we could potentially have the same problem and in this case we wouldn't be fixing it, because there's a new `cat` of the footer before our fix.

So, just moving the added line up we would be covering both potential issues:
```
    # File content: Indent 4
    cat $file | sed "s/^/    /"

    [ "$(tail -c 1 "$file")" ] && echo

    # Dashboard foot
    test "$type" = "dashboard" && cat $DASHBOARD_FOOT_FILE
```
Let me know if you agree and I will raise a small PR with this change.
Of course if the original files are well ended we will never face this issue, but technically we should move the line up to cover any potential file wrongly ended, not only `datasources`.

What do you think?

brancz:
You're absolutely right, feel free to open a PR.

Issue #631, opened 2017-09-21
contrib/kube-prometheus manifests generator sets newline incorrectly 


mad01:
**What did you do?**
in `contrib/kube-prometheus` rerun of `make generate` the k8s deployments newlines is set incorrectly see diff  

**What did you expect to see?**
expected to get a working k8s deployment. 

**What did you see instead? Under which circumstances?**
```
--- a/contrib/kube-prometheus/manifests/grafana/grafana-deployment.yaml
+++ b/contrib/kube-prometheus/manifests/grafana/grafana-deployment.yaml
@@ -64,11 +64,8 @@ spec:
             memory: "32Mi"
             cpu: "100m"
         volumeMounts:
-        - name: grafana-dashboards-0
-          mountPath: /var/grafana-dashboards-0
+        - name: grafana-dashboards-0n          mountPath: /var/grafana-dashboards-0
       volumes:
       - name: grafana-storage
         emptyDir: {}
-      - name: grafana-dashboards-0
-        configMap:
-          name: grafana-dashboards-0
+      - name: grafana-dashboards-0n        configMap:n          name: grafana-dashboards-0
```

brancz:
Interesting. What OS are you running on and executing this? When I run this I don't get a diff, I'm assuming this is due to incompatible tooling. Maybe we should just use a Dockerfile that packages all the required tools anyways, just like we do for the `generate` target in the `Makefile` in the root of this repo.

@eedugon as you initially developed these scripts, your opinion is also welcome :slightly_smiling_face: .

mad01:
id did this in macOS sierra

eedugon:
Hi @mad01 , @brancz,
Sorry for my late response about this. Interesting issue.

I believe the reason is how the `sed` command in macOS is handling the newline characters.
@mad01 : Have you tried installing in your system `GNU sed / gsed` instead of the default version?

@brancz : There are some options to try to mitigate this, for example:
- using `perl` instead of `sed`
- change the code to direcly include escaped newlines instead of `\n`-s, or trying with `\\n` or something...

If you think it's worthy let me know. But of course if the tools are built via a Dockerfile that would be a better option I guess.

For your info, the problem is here:
- We use some variables with this kind of content as an example: `- name: $configmap\n  mountPath: /var/$configmap`
- Then we have to indent them, so I add the neccesary spaces with this:
`echo "$string" | sed -e "s#^#$indent_string#" -e "s#\\\n#\\\n$indent_string#g"`
[...]

PR #672, opened 2017-10-06
change make generate to use docker run to be more protable

mad01:
change manifest generate to use docker to improve probability issues seen in #631.


brancz:
lgtm :+1: thanks! Easier than I thought :smile:

PR #681, opened 2017-10-10
Added prometheusArgs param

### LINKED ISSUE #426: prometheus_local_storage_memory_series leak over time on an empty cluster
On a basic cluster setup  3 nodes . 
K8s 1.5 
Prometheus 1.7 
No pods except for  K8s system pods and monitoring pods are running.
prometheus_local_storage_memory_series is growing with time , exploding when it reaches max amount of RAM available .
Number of series is growing without any reason for that.
Probably some ID that changes frequently is being part of a label , but cannot identify where it comes from .
Started a cluster with 40K series ..ended up with 150K after 12 hours  before Prometheus died .

robin-opsguru:
Allows users to selectively add or override cli params for a given prometheus
instance.

We've run into a few use cases where hand-tuning of certain cli params (particularly related to memory, queries, and storage) have been essential to our ability to use the operator. 

Fixes #426

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -163,6 +163,7 @@ Specification of the desired behavior of the Prometheus cluster. More info: http
 | secrets | Secrets is a list of Secrets in the same namespace as the Prometheus object, which shall be mounted into the Prometheus Pods. The Secrets are mounted into /etc/prometheus/secrets/<secret-name>. Secrets changes after initial creation of a Prometheus object are not reflected in the running Pods. To change the secrets mounted into the Prometheus Pods, the object must be deleted and recreated with the new list of secrets. | []string | false |
 | affinity | If specified, the pod's scheduling constraints. | *v1.Affinity | false |
 | tolerations | If specified, the pod's tolerations. | []v1.Toleration | false |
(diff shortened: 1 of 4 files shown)


brancz:
This is a non-goal of the Prometheus Operator. If there is a knob missing (particularly around memory and storage) then that's an explicit feature rather than specifying raw command line parameters. The Prometheus Operator allows you to perform upgrades from one version to another and if a flag changes we can handle that gracefully, if raw arguments are passed, we have no such guarantee.

FWIW Prometheus 2.0 is in release candidate, and especially the memory and storage things have changed heavily, removing and changing most of the configurable parameters (in a good way). The target heap memory etc. are now entirely handled by the operating system instead of estimating it from within Prometheus. The operating system can evict memory as needed, generally making the Prometheus experience much much easier to operate from storage and memory perspective, which has been a pain point for many people in Prometheus 1.x. I highly recommend that you give it a try! :)

(the Prometheus Operator of course already supports Prometheus 2.0, with no changes to the Prometheus object, except setting the version :wink:, the rest the Prometheus Operator handles as it knows how to translate these things to the breaking changes)

robin-opsguru:
Thanks for the feedback @brancz. Unfortunately Prometheus 2.0 wasn't an option on one of our projects due to its pre-release state - but we'll certainly be testing it in the near future.

In our case, the operator was extremely useful in deploying prometheus but was an obstacle in applying our configs. We have people that are quite good at tuning prometheus and were hoping this patch would be useful for others in a similar situation.

PR #682, opened 2017-10-10
Documentation/api.md: fix broken links

flyer103:


brancz:
Everything in the API documentation is generated with the `cmd/po-docgen` tool and the external links are here:

https://github.com/coreos/prometheus-operator/blob/c6c96796e1e1a0397db96d2c36231379ee055bf9/cmd/po-docgen/api.go#L43-L49

And in the comments here:

https://github.com/coreos/prometheus-operator/blob/c6c96796e1e1a0397db96d2c36231379ee055bf9/pkg/client/monitoring/v1/types.go

Would you like to give it a go and fix them there?

flyer103:
got, I will submit a new commit to fix it.

flyer103:
Some tools generating docs just extract the comments but `po-docgen` goes further to manage the external links of some table fields. This operation is interesting and I think there're some improvements to do:

* Add the usage doc of `po-docgen` (it needs a few minutes to know how to use it)
* Manage all of the external links in the comments. IMHO it just manages links of some table fields

Anyway, it's happy to understand the design of `po-docgen` :D

brancz:
> Add the usage doc of po-docgen (it needs a few minutes to know how to use it)

Feel free to add a section in the readme under development on how to use the tool to generate docs. As a first time user you're probably more aware of the things to watch out for than us :slightly_smiling_face: .

> Manage all of the external links in the comments. IMHO it just manages links of some table fields

The difference between the two types of links is that one links to general documentation, whereas the others link to versioned documentation, which we update when we update the respective upstream types.

In the scope of this PR, everything looks good to me, but feel free to open follow up PRs :slightly_smiling_face:. Thanks for your contribution!

flyer103:
Thanks, I'll submit other PRs to solve problems I have found :D

Recently I have written an operator and am learning the design of `prometheus-operator` and `etcd-operator`. I have summarized the good points of the design and potential not good parts.
For `prometheus-operator`, I don't find design docs and code related to automatic fault recovery and there's no ChaosMonkey design, etc.

This project is great for users and developers.

brancz:
> For prometheus-operator, I don't find design docs and code related to automatic fault recovery and there's no ChaosMonkey design, etc.

As Prometheus doesn't have any type of clustering all a chaos monkey would test is that Kubernetes correctly attaches storage to a Pod, which we didn't feel like it was then worth it. For Alertmanager clusters this could be worth it, however, we're building more sophisticated fault injection and load tests for the Alertmanager, so that the chaos monkey approach will be tested, just not inside of this repo.

Regarding our design docs, what were you looking for or felt is missing? We have a lot of docs, and also a lot of room for improvement for our docs :slightly_smiling_face:.

flyer103:
Thanks for your reply and that solves my questions :D

PR #684, opened 2017-10-10
tests: Don't archive Tectonic installer build folder

mxinden:
Archiving the /build folder served the purpose of better cleaning up
clusters in case the cleanup steps fail. On the downside it takes a lot
of space and it is difficult to manage.

This patch removes the archive steps.


brancz:
lgtm :+1: thanks!

PR #685, opened 2017-10-11
Customize the Crd Group and propagate optional labels

### LINKED ISSUE #679: Deploy prometheus-operator scoped in one namespace
I would like to be able to deploy multiple prometheus-operator in one cluster and limit each of them to gather metric from a single namespace. 
I propose the following changes:

- [x] Add an option to change the crd api-group name: 
 ```
 -crd-apigroup-name string
    	Set the Prometheus operator CRD ApiGroup Name
```
- [x] Add an option to select a namespace and filter k8s api calls to this namespace
```
  -namespace string
    	Scope Prometheus to single kubernetes namespace
```

### PR DESCRIPTION
**Pull request by ant31**
fixes #679 #680

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -80,7 +80,21 @@ type Labels struct {
 
 // Implement the flag.Value interface
 func (labels *Labels) String() string {
-	return fmt.Sprintf(labels.LabelsString)
+	return labels.LabelsString
+}
+
+// Merge labels create a new map with labels merged.
+func (labels *Labels) Merge(otherLabels map[string]string) map[string]string {
+	mergedLabels := map[string]string{}
+
+	for key, value := range otherLabels {
+		mergedLabels[key] = value
+	}
+
+	for key, value := range labels.LabelsMap {
+		mergedLabels[key] = value
+	}
+	return mergedLabels
 }
 
 // Implement the flag.Set interface
@@ -358,9 +372,9 @@ func (c *Operator) syncNodeEndpoints() error {
 	eps := &v1.Endpoints{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: c.kubeletObjectName,
-			Labels: map[string]string{
+			Labels: c.config.Labels.Merge(map[string]string{
 				"k8s-app": "kubelet",
-			},
+			}),
 		},
 		Subsets: []v1.EndpointSubset{
 			{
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by brancz** [comment]
This PR needs to be rebased for tests to be run

File: pkg/prometheus/operator.go [hunk 1]
```diff
@@ -86,6 +86,7 @@ type Config struct {
 	PrometheusConfigReloader     string
 	AlertmanagerDefaultBaseImage string
 	PrometheusDefaultBaseImage   string
+	K8sNamespace                 string
```
brancz: I was thinking we should just pass in the entire group, instead of "namespace". We will need the namespace as well, but that's for just watching the objects in a single namespace.

brancz:
Looks pretty good so far. Now we just also need to make sure, that the labels are applied to all objects created off of this.

pkg/prometheus/operator.go
```diff
@@ -83,6 +83,14 @@ func (labels *Labels) String() string {
 	return fmt.Sprintf(labels.LabelsString)
 }
 
+// Merge labels
+func (labels *Labels) Merge(dict map[string]string) map[string]string {
```
brancz: `dict` -> `otherLabels`, dict is not a thing in go

pkg/prometheus/operator.go
```diff
@@ -83,6 +83,14 @@ func (labels *Labels) String() string {
...
+	for key, value := range labels.LabelsMap {
```
brancz: I would prefer if we create an entirely new map here and transfer both sets of key/value pairs into the new map and not modify any of the inputs.

ant31:
@brancz ptal

pkg/prometheus/operator.go
```diff
@@ -73,6 +73,45 @@ type Operator struct {
...
 }
 
+type Labels struct {
+	LabelsString string
+	LabelsMap    map[string]string
+}
+
+// Implement the flag.Value interface
+func (labels *Labels) String() string {
+	return fmt.Sprintf(labels.LabelsString)
```
brancz: `return labels.LabelsString` should be equivalent, and since we're not using `.LabelsString` outside of this file, we can leave it unexported.

ant31: yes, that's a mistake from previous code

brancz:
Nice, thanks :+1: lgtm

PR #687, opened 2017-10-11
*: scope list watching to configurable namespace

brancz:
An empty value means the Prometheus Operator will continue to list watch
against all namsepaces, therefore this is a non-breaking, backward
compatible change.

@fabxc @ant31

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -37,7 +37,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/pkg/api"
 	"k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/pkg/apis/apps/v1beta1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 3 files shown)


pkg/alertmanager/operator.go
```diff
@@ -98,18 +98,23 @@ func New(c prometheusoperator.Config, logger log.Logger) (*Operator, error) {
...
+			Namespace:                    c.Namespace,
+		},
 	}
 
 	o.alrtInf = cache.NewSharedIndexInformer(
 		&cache.ListWatch{
-			ListFunc:  o.mclient.MonitoringV1().Alertmanagers(api.NamespaceAll).List,
-			WatchFunc: o.mclient.MonitoringV1().Alertmanagers(api.NamespaceAll).Watch,
+			ListFunc:  o.mclient.MonitoringV1().Alertmanagers(o.config.Namespace).List,
+			WatchFunc: o.mclient.MonitoringV1().Alertmanagers(o.config.Namespace).Watch,
```
ant31: How do you still do `api.NamespaceAll` ?

brancz: I made sure to use the `api.NamespaceAll` variable as the default value for the flag.

brancz:
@coreosbot test this please

PR #688, opened 2017-10-11
[kube-prometheus] Update to Grafana 4.5.2 in template

jordanjennings:
In the generated manifest for the grafana deployment at /contrib/kube-prometheus/manifests/grafana/grafana-deployment.yaml the version was bumped to 4.5.2 but seems it was overlooked in this templated version of the file.


brancz:
Thanks!

PR #691, opened 2017-10-12
[kube-prometheus] Make grafana script portable

jordanjennings:
`stat -c%s` doesn't work on macOS, this change makes the script work on both Linux and macOS

Tested on Amazon Linux, Debian 8, and macOS 10.12.6


brancz:
As there will be many incompatibilities between gnu, bsd and other versions of the used tools, we have decided to unify the generation by using a docker debian image for this.

Please have a look at the `Makefile` in the `contrib/kube-prometheus/` directory, specifically the `generate` target, it runs the generation step in a docker container. This is meant to be the solution to run these scripts in a unified way.

jordanjennings:
@brancz Got it, totally missed that. I'll adjust my scripts to call the makefile instead of the script directly. Thanks!

dimm0:
The makefile is working fine if I put the dashboards together with original ones (assets/grafana), but script in hack/grafana-dashboards-configmap-generator/bin still fails on mac. Is it supposed to be called during the main makefile execution?

brancz:
The scripts are not meant to be compatible with anything other than the debian image running in docker, which is what the makefile utilizes.

PR #702, opened 2017-10-19
tests: Make sure to delete folders created by root

mxinden:
In our Jenkinsfile we include the `wsCleanup()` step. This does not fail
the build even though it was not able to successfully clean up the
workspace.

In addition, with this patch we make sure ourself that the workspace is
cleaned up after the test execution.


mxinden:
Needs more investigation. Cleaned up Jenkins for now.

PR #712, opened 2017-10-26
Grafana helm Chart: Expose grafana-watcher image in values.yml

pmontanari:
Expose grfana watcher image so it can be overrinded and pulled from another repository

### DIFF
--- a/helm/grafana/templates/dashboards-configmap.yaml
+++ b/helm/grafana/templates/dashboards-configmap.yaml
@@ -20,6 +20,6 @@ data:
       "basicAuth": false,
       "name": "prometheus",
       "type": "prometheus",
-      "url": "http://prometheus-prometheus:9090"
+      "url": "http://{{ printf "%s-%s" .Release.Name "prometheus" }}:9090"
     }
   {{- end }}
--- a/helm/grafana/templates/dashboards-configmap.yaml
+++ b/helm/grafana/templates/dashboards-configmap.yaml
@@ -20,6 +20,6 @@ data:
       "basicAuth": false,
       "name": "prometheus",
       "type": "prometheus",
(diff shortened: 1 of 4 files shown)


helm/grafana/templates/dashboards-configmap.yaml
```diff
@@ -20,6 +20,6 @@ data:
       "basicAuth": false,
       "name": "prometheus",
       "type": "prometheus",
-      "url": "http://{{ printf "%s-%s" .Release.Name "prometheus" }}:9090"
+      "url": "http://prometheus-prometheus:9090"
```
brancz: In my limited understanding, I understand, that multiple helm charts can be created as one release, therefore the templating from before was applicable.

weiwei04: agreeed with @brancz.

pmontanari: Ok.. But the datasource will need to be updated manually after installation because, Prometheus's service name is wrong...

brancz: The idea is that it's not, when they're created through the same helm release.

pmontanari: Ok. Got it... Will revert the commit

weiwei04:
lgtm, please squash the commit :-)

pmontanari:
done

helm/grafana/values.yaml
```diff
@@ -44,6 +44,10 @@ image:
   repository: grafana/grafana
   tag: 4.4.1
 
+grafana_watcher:
```
unguiculus: Use camel case instead of underscore as suggested by Helm chart best practices and to be consistent.

brancz:
lgtm. @weiwei04 PTAL as well

brancz:
let's move forward with this

PR #714, opened 2017-10-27
fixing a small typo on the OutOfDisk condition

fmoctezuma:
Faced some errors generating the rules for the k8s configmap, found an small typo on the conditional
```
curl -X POST http://ingress.mycluster/metrics/-/reload
failed to reload config: one or more errors occurred while applying the new configuration (-config.file=/etc/prometheus/config/prometheus.yaml)
```
After change this, rules are loaded correctly

### DIFF
--- a/contrib/kube-prometheus/assets/prometheus/rules/node.rules
+++ b/contrib/kube-prometheus/assets/prometheus/rules/node.rules
@@ -9,7 +9,7 @@ ALERT NodeExporterDown
     description = "Prometheus could not scrape a node-exporter for more than 10m, or node-exporters have disappeared from discovery.",
   }
 ALERT K8SNodeOutOfDisk
-  IF kube_node_status_condition{condition"OutOfDisk", status="true"} == 1
+  IF kube_node_status_condition{condition="OutOfDisk", status="true"} == 1
   LABELS {
     service = "k8s",
     severity = "critical"


brancz:
Can you open this PR against the master branch, and re-generate the rule manifest using `make generate` in the `contrib/kube-prometheus/` directory.

fmoctezuma:
Sure, just to double check, to re-generate the rule manifest `make generate` should be executed in master or in release-0.12 branch ?

brancz:
All of that from master branch.

fmoctezuma:
This typo only affects  the branch `release-0.12` , this is not an issue in master. That's why I created the PR against  `release-0.12` and I can not make the change in master as requested.

brancz:
I'm confused, as to why you want it to be in the release branch? We're not going to cut a new release for a change in the kube-prometheus manifests. The release branches are meant to version the Prometheus Operator code. The remaining code is versioned on best effort.

fmoctezuma:
Sorry for my delay response,
So in that case feel free to close this PR,
I was using tectonic installer with prometheus-operator v0.12.0 and prometheus v1.7.1,
Generated manifest [prometheus-k8s-rules.yaml](https://github.com/coreos/prometheus-operator/blob/release-0.12/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml)  it's in a good state, but if you want to generate new rules, [`contrib/kube-prometheus/assets/prometheus/rules/node.rules` ](https://github.com/coreos/prometheus-operator/blob/release-0.12/contrib/kube-prometheus/assets/prometheus/rules/node.rules#L12)
contains the typo I was trying to fix, and rules are not loaded at all.

brancz:
In that case I recommend to just use more up to date releases. But thanks for all your efforts!

PR #719, opened 2017-10-30
Omit static compute resource requests and limits from deployment

svend:
Do not specify static resource settings in the kubernetes-deployment to avoid
unnecessary replacement of pods when the manifest is reapplied (`kubectl
apply`).

The addon-resizer will dynamically set the pod compute resource values. If the
values are also set statically in the deployment configuration, reapplying the
configuration will result in the pods getting replaced. Without the static
resource, the deployment configuration can be reapplied, and the pods will not
be replaced.

This change was also made in the upstream kube-state-metrics example manifests.

https://github.com/kubernetes/kube-state-metrics/pull/285


brancz:
Thanks again! :slightly_smiling_face:

PR #721, opened 2017-10-31
Update KOPSonAWS.md

dmfenton:
The AWS CLI SG commands are not necessary, kops creates an SG with these ports open to the correct SG by default.

You may want to consider deleting this file altogether


brancz:
Thanks for updating! :tada:

PR #728, opened 2017-11-06
Add running tests on Travis CI

brancz:
@ant31


.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+go:
+- 1.9
+services:
+- docker
+jobs:
+  include:
+  - stage: Check generated contents are up to date and code is formatted.
+    script: ./scripts/jenkins/check-make-generate.sh
+  - stage: Compile binary
+    script: make crossbuild
```
mxinden: To save execution time either this or https://github.com/coreos/prometheus-operator/blob/master/scripts/jenkins/run-e2e-tests.sh#L23 can be skipped.

scripts/travis-e2e.sh
```diff
@@ -0,0 +1,32 @@
...
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+# error on unset variables
+set -u
+# print each command before executing it
+set -x
+
+export KUBERNETES_VERSION=v1.7.5
+
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/$KUBERNETES_VERSION/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
```
mxinden: Would you mind breaking these long lines up a bit?

scripts/travis-e2e.sh
```diff
@@ -0,0 +1,32 @@
...
+
+export KUBECONFIG=$HOME/.kube/config
+minikube version
+sudo minikube start --vm-driver=none --kubernetes-version=$KUBERNETES_VERSION
+
+minikube update-context
+
+JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'; until kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do sleep 1; done
+
+make crossbuild
```
mxinden: Once again this is also done in https://github.com/coreos/prometheus-operator/blob/master/scripts/jenkins/run-e2e-tests.sh#L23

brancz: My idea was that if necessary all the stages could be parallelized. And I'm not sure the result of previous stages are available, but I'm fine with removing the other stage :+1:

brancz:
@mxinden PTAL. Also can you take care of removing the jenkins hooks once this is merged?

Makefile
```diff
@@ -75,6 +75,6 @@ jsonnet:
 	jsonnet -J /ksonnet-lib hack/generate/prometheus-operator-rbac.jsonnet | json2yaml > contrib/kube-prometheus/manifests/prometheus-operator/prometheus-operator.yaml
 
 jsonnet-docker:
-	docker build -f scripts/jenkins/jsonnet/Dockerfile -t po-jsonnet .
+	docker build -f scripts/Dockerfile -t po-jsonnet .
```
mxinden: Might be easier to understand the purpose of this dockerfile if it is inside a folder like `jsonnet`?

mxinden:
@brancz If this works fine in the long run I will ask for the deletion of the jenkins-monitoring, ok?

mxinden:
Deleted all Jenkins related webhooks.

PR #734, opened 2017-11-07
Add flag to customize CRD Kinds

ant31:


ant31:
cc @brancz

brancz:
this lgtm, once we validate that it works for this purpose we can merge

PR #737, opened 2017-11-08
*: upgrade to Prometheus 2.0

### LINKED ISSUE #735: Update to use v2.0.0 final
**What did you do?**
started up an instance of prometheus using the operator

to have version v2.0.0 running

v2.0.0-rc.1 was running and no way to set a newer version

brancz:
Closes #735

@ant31


stuart-warren:
could we get a release with this fix in please?

brancz:
kube-prometheus is not versioned along with the prometheus-oprator, in fact it's not versioned at all, so you can go ahead today and use the content of the kube-prometheus manifest files :slightly_smiling_face:

PR #740, opened 2017-11-09
Update KOPSonAWS.md

harryleesan:
Instead of using `deploy` we can use `self-hosted-deploy`, which achieves this same thing.


brancz:
lgtm :+1: thanks!

PR #748, opened 2017-11-12
Generate ExternalLabels deterministically

jfoy:
ExternalLabels now have a stable ordering in the generated ConfigMap prometheus.yaml file.


brancz:
Good catch! Thanks a lot! lgtm :+1:

jfoy:
Thank you! How close do you think we are to a v0.14.2 release that might include this?

brancz:
A new release should be out very soon, as we would like to get a few features out, that were recently developed.

PR #753, opened 2017-11-13
Allow templating for alertmanager

atopuzov:
User can now specify custom templates (as described in [1]) in  `templateFiles` in `values.yaml`.
Eg.

```
templateFiles:
  template_1.tmpl: |-
    .... template file
```
Also has to add:
```
    templates:
    - '*.tmpl'
```
in `config` (the config that populates `alertmanager.yaml`).

[1] https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/alerting.md

### DIFF
--- a/helm/alertmanager/values.yaml
+++ b/helm/alertmanager/values.yaml
@@ -17,6 +17,10 @@ config:
     receiver: webhook
     repeat_interval: 12h
 
+## Alertmanager template files to include
+#
+templateFiles: {}
+
 ## External URL at which Alertmanager will be reachable
 ##
 externalUrl: ""
--- a/helm/alertmanager/values.yaml
+++ b/helm/alertmanager/values.yaml
@@ -20,6 +20,22 @@ config:
 ## Alertmanager template files to include
 #
 templateFiles: {}
+#
+# An example template:
+#   template_1.tmpl: |-
+#       {{ define "cluster" }}{{ .ExternalURL | reReplaceAll ".*alertmanager\\.(.*)" "$1" }}{{ end }}
+#
+#       {{ define "slack.myorg.text" }}
+#       {{- $root := . -}}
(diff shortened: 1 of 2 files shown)


helm/alertmanager/values.yaml
```diff
@@ -17,6 +17,10 @@ config:
     receiver: webhook
     repeat_interval: 12h
 
+## Alertmanager template files to include
+#
+templateFiles: {}
+
```
gianrubio: Could you provide an example keeping the content commented?

atopuzov: I use something like this, shamelessly stolen from [1]. I'll add a more simpler example to the code.
```
  templateFiles:
    template_1.tmpl: |-
      {{ define "cluster" }}{{ .ExternalURL | reReplaceAll ".*alertmanager\\.(.*)" "$1" }}{{ end }}

      {{ define "slack.myorg.text" }}
      {{- $root := . -}}
      {{ range .Alerts }}
        *Alert:* {{ .Annotations.summary }} - `{{ .Labels.severity }}`
        *Cluster:*  {{ template "cluster" $root }}
        *Description:* {{ .Annotations.description }}
        *Graph:* <{{ .GeneratorURL }}|:chart_with_upwards_trend:>
        *Runbook:* <{{ .Annotations.runbook }}|:spiral_note_pad:>
        *Details:*
          {{ range .Labels.SortedPairs }} • *{{ .Name }}:* `{{ .Value }}`
          {{ end }}
      {{ end }}
      {{ end }}
```
[1] https://engineering.infinityworks.com/slack-prometheus-alertmanager/

atopuzov: This si what it looks like:
template

gianrubio:
lgtm

brancz:
thanks!

PR #754, opened 2017-11-14
Helm Chart - Add support to node-exporter >= 0.15.0

pmontanari:
Some flags have been renamed in 0.15.0
Keeping back-compatibility with if


helm/kube-prometheus/charts/exporter-node/templates/deamonset.yaml
```diff
@@ -21,8 +21,13 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: "{{ .Values.image.pullPolicy }}"
           args:
+          {{ if lt .Values.image.tag "1.15.0" }}
```
brancz: This should be v0.15.0

pmontanari: Oups... Right... Fixing that

brancz:
lgtm, thanks a lot!

PR #755, opened 2017-11-14
Daemonset rules

atopuzov:
Alerts for making sure that all daemonsets are running properly.

I first tried with 5 min but a rolling cluster update with kops would trigger them, so I bumped up to 10m.


brancz:
Awesome! Thanks! lgtm :+1:

PR #757, opened 2017-11-14
Configurable node exporter

atopuzov:
Allows you to:
* omit `node-exporter` deployment by setting `deployExporterNode` to `False`.
* listen on a different port (so 2 versions of node exporter can coexist, eg testing out a new version)

### DIFF
--- a/helm/kube-prometheus/values.yaml
+++ b/helm/kube-prometheus/values.yaml
@@ -1,3 +1,6 @@
+# exporter-node configuration
+deployExporterNode: True
+
 alertmanager:
   ## Alertmanager configuration directives
   ## Ref: https://prometheus.io/docs/alerting/configuration/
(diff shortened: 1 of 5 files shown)


helm/kube-prometheus/values.yaml
```diff
@@ -1,3 +1,6 @@
+# exporter-node configuration
+deployExporterNode: True
```
brancz: the exporter is called "node-exporter" so I think we should call this "deployNodeExporter"

atopuzov: I used the sub-chart name which is `exporter-node`  since you are controlling if a sub-chart gets deployed.
I'm fine with changing the name if you believe it's better.

brancz: hmm I see .. let's fix naming later and keep this constructive, I'm gonna go ahead and merge this

brancz:
Thanks :+1: lgtm

PR #766, opened 2017-11-17
Fix docu concerning initial deployment

stroebitzer:
After cloning the repo I wanted to deploy kube-prometheus into my running minikube. I did the steps mentioned in the docu which brought me to the following exception:

namespace "monitoring" created
error: the path "manifests/prometheus-operator" does not exist
Waiting for Operator to register custom resource definitions.................................................................................................................^C

### DIFF
--- a/contrib/kube-prometheus/README.md
+++ b/contrib/kube-prometheus/README.md
@@ -35,7 +35,7 @@ Simply run:
 
 ```bash
 export KUBECONFIG=<path> # defaults to "~/.kube/config"
-hack/cluster-monitoring/deploy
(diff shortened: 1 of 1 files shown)


brancz:
I think this is more confusing then before, because a lot of the scripts in `hack/` have to be executed from that directory. I would prefer to add a 

```bash
cd $GOPATH/src/github.com/coreos/prometheus-operator/contrib/kube-prometheus
```

As the first command. What do you think?

brancz:
```
cd contrib/kube-prometheus/
```
works for me as well

stroebitzer:
Sounds great ;)

brancz:
lgtm, thanks!

PR #781, opened 2017-11-23
Adding v to version image tag

rawkode:
Just to be sure this worked, I created a Playground

https://play.golang.org/p/E3sjIbjRd4


brancz:
lgtm, thanks for fixing!

rawkode:
@brancz When does the Charts repo get updated?

brancz:
@rawkode currently manually, we have yet to figure out when to do this, as the kube-prometheus stack as well as the helm charts are not versioned along with the prometheus operator.

For now @ant31 can you update the repo with the current state off of master?

PR #783, opened 2017-11-24
Adding Grafana as an optional dependency

rawkode:


rawkode:
I just noticed there is a Grafana helm chart in this repository, but it doesn't seem to offer anything more than the official chart. Let me know your thoughts - I think it's best to reuse official charts whenever possible.

brancz:
The grafana chart from here is actually very different, as we configure grafana entirely from config files, and in a declarative manner, where we don't need any persistent storage for it.

rawkode:
OK. Using the customised Grafana

helm/kube-prometheus/values.yaml
```diff
@@ -1,6 +1,9 @@
 # exporter-node configuration
 deployExporterNode: True
 
+# Grafana
+deployGrafana: False
```
gianrubio: I would like to have this the default values as `True`, in that case `kube-prometheus` will provide monitoring, alerting and dashboarding.

rawkode:
@gianrubio Updated and squashed

gianrubio:
I forgot to ask you to bump the version of the [kube-prometheus](https://github.com/rawkode/prometheus-operator/blob/44a94fe1aa9d40ed1f8f56d4ed8da10a12b6d469/helm/kube-prometheus/Chart.yaml#L10), charts versions should be immutable.

PR #784, opened 2017-11-24
Helm review

gianrubio:
Automatically package helm charts after a PR
Add prometheus-operator and grafana as requirement for kube-prometheus
Avoid chart overriding, comparing the generated hash and the hash in the server

Before merge this PR it's required to configure the aws credentials on travis ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,14 @@ jobs:
     script: ./scripts/travis-e2e.sh
   - stage: E2e helm
     script: ./scripts/travis-e2e-helm.sh
-  - stage: helm update
-    script: 'if [ "$TRAVIS_BRANCH" == "master" ]; then make helm-packages; fi'
+
+deploy:
+  provider: script
+  script: helm/hack/helm-package.sh "alertmanager grafana prometheus prometheus-operator exporter-kube-api \
+                exporter-kube-dns exporter-kube-scheduler exporter-kubelets exporter-node \
+                exporter-kube-controller-manager exporter-kube-etcd exporter-kube-state exporter-kubernetes" && \
+        helm/hack/sync-repo.sh && \
+        helm/hack/helm-package.sh kube-prometheus && \
+        helm/hack/sync-repo.sh
+  on:
+    branch: master
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ go:
 - 1.9
 services:
 - docker
+before_install:
+  - pip install --user awscli
+  - export PATH=$PATH:$HOME/.local/bin
 jobs:
   include:
   - stage: Check generated contents are up to date and code is formatted.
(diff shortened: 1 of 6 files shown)


brancz:
Is it safe to just always sync? Would that override old content, or potentially create problems when things are accidentally merged?

ant31:
Can we make sure that already pushed chart (same version) aren't overwritten. 
A version must be pinned to a single  digest.

gianrubio:
> Is it safe to just always sync? Would that override old content, or potentially create problems when things are accidentally merged?

@brancz I'm thinking to enable e2e tests on the helm charts, it's the best way to have good charts and avoid problems like that.  WDYT?

gianrubio:
> Can we make sure that already pushed chart (same version) aren't overwritten.
> A version must be pinned to a single digest.

Agree! I'm going to add the same validations from [kubernetes/charts](https://github.com/kubernetes/charts/blob/ec99fad17ec5f0764b6712237d3e6b9474581b52/test/circle/lint.sh#L93) when I work on the e2e tests. This will avoid changes without bumping the chart version.

For now, can we ask the users to bump the version when they change the helm charts?

ant31:
> For now, can we ask the users to bump the version when they change the helm charts?

Yes, but CI must fail if they don't.

gianrubio:
@ant31 CI was failing because awscli was not installed, I already fixed that but probably we're going to fail again because there's no AWS secrets and credentials configured on travis. Could you configure this credentials on travis-ci?

unguiculus:
FWIW, over at https://github.com/kubernetes/charts we are planning to package up the checks we run on CircleCI in a Docker image so everyone can run those on their own. Ping @mattfarina before you start with this.

mattfarina:
My first pass will be to package up the static analysis we run on charts (not the e2e ones). I've already started it but not published it yet. Coming soon.

gianrubio:
> FWIW, over at https://github.com/kubernetes/charts we are planning to package up the checks we run on CircleCI in a Docker image so everyone can run those on their own. Ping @mattfarina before you start with this.

Good news!! I'm willing to integrate this container here even if it's just static analysis. Is there an issue where I can follow the progress?

gianrubio:
@ant31 In case someone submit changes on the charts without bumping a version the [merge will fail](https://github.com/gianrubio/prometheus-operator/blob/45dea583f8e68a3727b03bfb4c7dc5a3375e21af/helm/hack/sync-repo.sh#L15-L23) 

According to [travis docs](https://docs.travis-ci.com/user/environment-variables/#Encrypting-environment-variables), encrypted variables are not visible to PR. It will be visible only after the merge, thats why the aws s3 sync command was not working.

Can you review the changes?

brancz:
this lgtm, but I'll give @ant31 the final word

gianrubio:
@ant31 can you review?

ant31:
thanks

PR #788, opened 2017-11-27
Per issue 593, append .yml to each filename used in the generated con…

nuala33:
…figmap. I wasn't able to see my alerts in the Prometheus UI until I did this

### DIFF
--- a/contrib/kube-prometheus/hack/scripts/generate-rules-configmap.sh
+++ b/contrib/kube-prometheus/hack/scripts/generate-rules-configmap.sh
@@ -13,6 +13,6 @@ EOF
 
 for f in assets/prometheus/rules/*.rules.y*ml
 do
-  echo "  $(basename $f): |+"
+  echo "  $(basename $f | sed s/\.yml//): |+"
   cat $f | sed "s/^/    /g"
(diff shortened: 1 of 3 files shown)


contrib/kube-prometheus/hack/scripts/generate-rules-configmap.sh
```diff
@@ -13,6 +13,6 @@ EOF
 
 for f in assets/prometheus/rules/*.rules.y*ml
 do
-  echo "  $(basename $f): |+"
+  echo "  $(basename $f | sed s/\.yml//): |+"
```
brancz: We'll need the same regex in this case as [here](https://github.com/coreos/prometheus-operator/pull/780/files#diff-2005db90391a36dfb67a1e855f6c770fR14), as we should support both `yml` as well as `yaml`.

nuala33: Fixed using quotes in the most recent commit. Additionally, please let me know if the values.yml file looks correct now :)

brancz:
all of this lgtm, thanks a lot!

PR #789, opened 2017-11-28
Fix cluster:container_cpu_usage:ratio rule

slok:
Hi! :)

This small PR fixes `cluster:container_cpu_usage:ratio` rule for K8s metrics rules.


brancz:
thanks!

PR #794, opened 2017-11-29
kube-prometheus: Add Prometheus 2.0 rules

brancz:
@ant31

### DIFF
--- a/contrib/kube-prometheus/assets/prometheus/rules/prometheus.rules.yaml
+++ b/contrib/kube-prometheus/assets/prometheus/rules/prometheus.rules.yaml
@@ -42,3 +42,30 @@ groups:
     annotations:
       description: Prometheus {{ $labels.namespace }}/{{ $labels.pod}} is not connected
         to any Alertmanagers
+  - alert: PrometheusTSDBReloadsFailing
+    expr: increase(prometheus_tsdb_reloads_failures_total[2h]) > 0
+    for: 12h
+    labels:
+      severity: warning
+    annotations:
+      description: '{{$labels.job}} at {{$labels.instance}} had {{$value | humanize}}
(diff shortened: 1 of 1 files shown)


contrib/kube-prometheus/assets/prometheus/rules/prometheus.rules.yaml
```diff
@@ -42,3 +42,30 @@ groups:
...
+    for: 12h
+    labels:
+      severity: warning
+    annotations:
+      description: '{{$labels.job}} at {{$labels.instance}} had {{$value | humanize}}
+        reload failures over the last four hours.'
+      summary: Prometheus has issues reloading data blocks from disk
+  - alert: PrometheusTSDBCompactionsFailing
+    expr: increase(prometheus_tsdb_compactions_failed_total[2h]) > 0
+    for: 12h
```
ant31: ?

brancz: this means that there have to be compaction fails within two hour periods for the past 12 hours

contrib/kube-prometheus/assets/prometheus/rules/prometheus.rules.yaml
```diff
@@ -42,3 +42,30 @@ groups:
...
+    labels:
+      severity: warning
+    annotations:
+      description: '{{$labels.job}} at {{$labels.instance}} had {{$value | humanize}}
+        compaction failures over the last four hours.'
```
Quentin-M: Where does the `four hours` come from?

brancz: whoops, you're right, seems like that was mistaken with the alerting rule below

Issue #765, opened 2017-11-17
deployment breaks in namespaces other than `monitoring`


hawksight:
* What I was doing:
I was attempting to deploy using the kube-prometheus script method to the `default` namespace

* What I seen
Everything deployed (or seemed to) but prometheus was not picking up most metrics therefore Grafana was graphing almost nothing as it couldn't get most of the data it needed.
Notably grafana could not detect nodes, pods or deployments, so those options on dashboard were `None`

* Environment
minikube version 1.8
kubectl version 1.8.2

* Kubernetes version information:
```
Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.2", GitCommit:"bdaeafa71f6c7c04636251031f93464384d54963", GitTreeState:"clean", BuildDate:"2017-10-24T19:48:57Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.0", GitCommit:"0b9efaeb34a2fc51ff8e4d34ad9bc6375459c4a4", GitTreeState:"dirty", BuildDate:"2017-10-17T15:09:55Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: minikube

* Manifests:
There are a few places where `monitoring` namespace is hard coded and therefore I think the issue I faced was that the operator could not find the relevant rbac permission to get the data it needed

For example, [prometheus-k8s-role-bindings.yaml](https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-role-bindings.yaml)
```
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: RoleBinding
metadata:
  name: prometheus-k8s
  namespace: monitoring
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: monitoring
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: RoleBinding
metadata:
  name: prometheus-k8s
  namespace: kube-system
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: monitoring
---
[...]

PR #795, opened 2017-11-29
contrib: namespace warning to kube-prometheus

hawksight:
NAMESPACE feature in scripts implies it works. Warning ensures
that users are aware that its use requires further changes.

Alleviates #765


brancz:
lgtm, thanks!

PR #799, opened 2017-12-01
Fix operator endpoint

PhilippeDupont:
Correct me if I'm wrong but shouldn't it be `http` to match deployment manifest? 
https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/templates/deployment.yaml#L32

### DIFF
--- a/helm/prometheus-operator/templates/servicemonitor-configmap.yaml
+++ b/helm/prometheus-operator/templates/servicemonitor-configmap.yaml
@@ -25,5 +25,5 @@ data:
           matchNames:
             - {{ .Release.Namespace | quote }}
         endpoints:
-        - port: web
+        - port: http
           interval: 30s


helm/prometheus-operator/templates/servicemonitor-configmap.yaml
```diff
@@ -25,5 +25,5 @@ data:
           matchNames:
             - {{ .Release.Namespace | quote }}
         endpoints:
-        - port: web
-          interval: 30s
\ No newline at end of file
+        - port: html
```
brancz: that would be `http` not `html`

PhilippeDupont: Of course sorry, this is what I meant ...

brancz:
lgtm

PR #804, opened 2017-12-07
Adding requested and limit values to CPU and limit value to memory

BradErz:
CPU
image

Memory
image

Pods dashboard

- CPU now shows the requested value and the limit
- Memory now shows the limit


brancz:
Awesome! Thanks, lgtm (sorry for taking so long)

PR #813, opened 2017-12-12
ServiceMonitor and Prometheus Namespaces Shouldn't Have to Match

midnightconman:
To match functionality with PR (https://github.com/coreos/prometheus-operator/pull/687), making it so that ServiceMonitor namespaces don't have to match Prometheus cluster namespace.

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1000,9 +1000,8 @@ func (c *Operator) selectServiceMonitors(p *monitoringv1.Prometheus) (map[string
 		return nil, err
 	}
 
-	// Only service monitors within the same namespace as the Prometheus
-	// object can belong to it.
-	cache.ListAllByNamespace(c.smonInf.GetIndexer(), p.Namespace, selector, func(obj interface{}) {
+	// Use specified namespace for ServiceMonitor scope (default v1.NamespaceAll)
+	cache.ListAllByNamespace(c.smonInf.GetIndexer(), c.config.Namespace, selector, func(obj interface{}) {
 		k, ok := c.keyFunc(obj)
 		if ok {
 			res[k] = obj.(*monitoringv1.ServiceMonitor)


midnightconman:
TravisCI failed on [clone timeout](https://travis-ci.org/coreos/prometheus-operator/jobs/315146542), how do we re-run it?

brancz:
It's intended for the `ServiceMonitor` objects to be in the same namespace as the `Prometheus` object. This is because the `ServiceMonitor` makes up how the Prometheus configuration is generated, and it shouldn't be possible to be able to inject configuration into the Prometheus server from any namespace.

midnightconman:
PR https://github.com/coreos/prometheus-operator/pull/687 makes it possible to restrict which namespaces the prometheus-operator watches and the `ServiceMonitor` [selector](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#servicemonitorspec) makes it possible to restrict which labels a `ServiceMonitor` must include to be added to a prometheus cluster. This makes it so that users can have multiple prometheus clusters in a single kubernetes cluster and direct which `ServiceMonitor`'s should be applied to which prometheus clusters by labels and selectors.

I wasn't able to find this feature (`ServiceMonitors` and prometheus clusters have to be in the same namespace) anywhere in the documentation, I only saw this in the code.

Does it makes sense to apply the namespace parameter that is used for the watcher to this logic (`ServiceMonitor`'s and Prometheus cluster namespace match)?

midnightconman:
I just remembered that there is also a `namespaceSelector` attached to prometheus clusters, why can endpoints be in a different namespace, but `ServiceMonitor`'s have to be in the same namespace? Why can't `ServiceMonitor`'s be in the same namespace as the service (or endpoints) they point to?

brancz:
There are two RBAC entities involved, and we are keeping privileges to a minimum.

1) The Prometheus Operator only associates ServiceMonitor objects in the same namespace as Prometheus objects, because the Prometheus configuration has to be written into that same namespace, and ServiceMonitor objects are what the Prometheus configuration is generated from. This means it requires read and write access to the objects/resources involved. This allows the Prometheus Operator to be fully functional even in a namespaced/restricted access environment where it can only access things in a single namespace.
2) Prometheus itself is a separate RBAC (ServiceAccount) entity, it can be configured in it's configuration to discover objects from other namespaces, requiring only read, never any write access.

mrIncompetent:
@brancz why shouldn't any namespace inject configuration into a prom cluster?
It would be the same behavior as a ingress. 

IMO Its the responsibility of an application to define how it should be monitored.

brancz:
> Its the responsibility of an application to define how it should be monitored.

This is absolutely true, and the way Prometheus and the Prometheus Operator is intended to be used is that teams create their own Prometheus server that they can configure to their need.

> why shouldn't any namespace inject configuration into a prom cluster?
> It would be the same behavior as a ingress.

I disagree with this for rule files, as rule files are the raw format and that would users the possibility to crash an entire setup, the parallel with the Ingress is actually the ConfigMap that has extra configuration for the Ingress controller, which is only accessible for admins.

mrIncompetent:
@brancz 
Then we maybe need an abstraction around the Rules which can be validated and people are not able to break the Prom?

But in he meantime, we should already allow ServiceMonitors to be in different Namespaces?

midnightconman:
> This is absolutely true, and the way Prometheus and the Prometheus Operator is intended to be used is that teams create their own Prometheus server that they can configure to their need.

I think this is the true crux of this conversation. We are using the prometheus operator to help run a centralized prometheus offering, where a centralized monitoring and analytics team owns the prometheus, alertmanager, and grafana clusters and the individual teams own the servicemonitor and custom rules configmaps for their services. This means that we don't have to give all teams permissions to create / delete servicemonitor and configmap objects in the same namespace as the prometheus cluster, but we do have to give them permissions to create servicemonitors and configmaps in their own namespace. It is true that malicious users could potentially bring down a prometheus cluster by creating a servicemonitor object in their namespace that would affect the behavior of the prometheus cluster, but couldn't they still do that if they had permission to create servicemonitor objects in the same namespace as the prometheus cluster?

Another question I would like to bring up is medium scale kubernetes clusters... if you have a medium sized cluster with about 30 namespaces, each namespace owned by a different scrum team, would you recommend that company run 30 prometheus-operators, 30 prometheus clusters, 30 alertmanager clusters, and 30 grafana clusters? In my local fork, we have 1 prometheus-operator, 2 prometheus clusters, 2 alertmanager clusters, and 2 grafana clusters for 30 namespaces (5 scrum teams).

> I disagree with this for rule files, as rule files are the raw format and that would users the possibility to crash an entire setup, the parallel with the Ingress is actually the ConfigMap that has extra configuration for the Ingress controller, which is only accessible for admins.

This is somewhat true, but we have the prometheus config reloader and alerts around failed prometheus config loads, so it should be easy to see that a configmap with broken rules was created and triage should be straight forward for cluster admins.

brancz:
> Another question I would like to bring up is medium scale kubernetes clusters... if you have a medium sized cluster with about 30 namespaces, each namespace owned by a different scrum team, would you recommend that company run 30 prometheus-operators, 30 prometheus clusters, 30 alertmanager clusters, and 30 grafana clusters? In my local fork, we have 1 prometheus-operator, 2 prometheus clusters, 2 alertmanager clusters, and 2 grafana clusters for 30 namespaces (5 scrum teams).

This definitely sounds like lack of guidance from our side, I apologize for that. The way it's intended to work is:

* 1 Prometheus Operator to be used by your entire organization.
* 1 central Alertmanager cluster to be used by your entire organization.
* 1 set of Prometheus servers to monitor the Kubernetes cluster
* n number of Prometheus servers to monitor your own applications, a typical approach is that a team that is responsible for an application collects metrics for its entire stack. Often also applications that are similar can be monitored by the same set of Prometheus servers.

Note that the `ServiceMonitor` already has the ability to discover `Service`s across namespaces. The thought here is that you should have one (or a few) common approach in your organization for monitoring applications created by your organization. This is in line with how Prometheus configuration works today.

> This is somewhat true, but we have the prometheus config reloader and alerts around failed prometheus config loads, so it should be easy to see that a configmap with broken rules was created and triage should be straight forward for cluster admins.

This one I'm still uneasy about as collisions of alerts, recording rules can cause a lot of problems and worst case including arbitrary code (recording rules are turing complete) can cause an outage of the monitoring stack. The goal here is to ensure that the monitoring stack is the most reliable component in your infrastructure.

optiomas:
> This definitely sounds like lack of guidance from our side, I apologize for that. The way it's intended to work is:
> 1 Prometheus Operator to be used by your entire organization.
> 1 central Alertmanager cluster to be used by your entire organization.
> 1 set of Prometheus servers to monitor the Kubernetes cluster
> n number of Prometheus servers to monitor your own applications, a typical approach is that a team that is responsible for an application collects metrics for its entire stack. Often also applications that are similar can be monitored by the same set of Prometheus servers.

What about Grafana? It seems like a good idea to centralize your dashboards.

brancz:
Yes that's the intention as well, sadly the tooling for grafana is not as good as one would like to, but we're working on that: https://github.com/brancz/kubernetes-grafana

midnightconman:
> This definitely sounds like lack of guidance from our side, I apologize for that. The way it's intended to work is:
> 
> 1 Prometheus Operator to be used by your entire organization.
> 1 central Alertmanager cluster to be used by your entire organization.
> 1 set of Prometheus servers to monitor the Kubernetes cluster
> n number of Prometheus servers to monitor your own applications, a typical approach is that a team that is responsible for an application collects metrics for its entire stack. Often also applications that are similar can be monitored by the same set of Prometheus servers.

This makes sense and I agree with this approach (mostly)... I do think there are some gaps that could make this better for larger organizations that would like to have a centralized prometheus, alertmanager, and grafana offering:
  - In having 30+ namespaces, wouldn't you start to feel pain around grafana integration? As I think you would need to track datasources manually.
  - Larger organizations normally require aggregated / consolidated analytics for planning / vision / retrospectives / etc. , this aggregate doesn't normally have to be at the same granularity as the original source, but one query point (sometimes with a longer retention) for all business verticals. I think there are some gaps here:
    - Remote writing configuration wasn't supported until recently by the prometheus-operator and I don't think there is good documentation for it.
    - I don't think that federation is supported by the prometheus-operator, so that would have to be done manually.
    - This aggregate might be possible at grafana, but I am not sure grafana would support aggregates of 30+ datasources.
  - Helm? Helm charts are not meant to include manifests in multiple namespaces, how would we include the serviceMonitor manifest in the same release as an application in a different namespace than the application?

Quentin-M:
IMHO one other very reasonable reason for having one Prometheus cluster to which other namespaces should be able to push ServiceMonitors into is for enabling the Kubernetes Custom Metrics. 

1/ Teams should not have to maintain their own Prometheus clusters just for a few metrics that are used for HPA.
2/ The [k8s-prometheus-adapter](https://github.com/DirectXMan12/k8s-prometheus-adapter) only support 1:1 mapping.
3/ Even if it would, it should not be the responsibility of the Kubernetes cluster's maintainers, nor a good practice for them to maintain the list/mapping of Prometheus clusters to get data for HPA.

I also believe that @midnightconman's use case and POV are pretty valid, both for resource utilization (the additional 2-8GB of RAM is significant when numerous namespaces are in place, what would an op say when asked to run an Elastic Search cluster per team?), separation of concerns / administration overhead and security reasons. At the end of the day, this should not even matter much to the operator whether it should watch ServiceMonitors in other namespaces. I am of the opinion that teams should even be able to configure records/alerts/how the alerts should be dealt with. Just my two cents on this matter, but the Custom Metrics issue is however a blocker.

carlpett:
I think there are some good points above, would also like to add a scenario which I see quite often: I help quite a few, well... less cloud-native... organizations who want to introduce Kubernetes and/or Prometheus. Frequently, these organizations' teams are not keen to take ownership of their own Prometheus installation, at least from the beginning. Initially, this is typically handled by some infrastructure team, or perhaps a group of champion-type users, gradually helping teams get more comfortable, instrumenting their code, etc. 
If the dev teams' choice is to manage their own Prometheus, or go without, almost all of them would rather not take the perceived burden of another service. Down the road, some of these teams might want the additional flexibility that having their own instance would give them, but this is probably 12-18 months away in most cases.

I fully understand how this is not the ideal setup, but it seems the one that a lot, if not the majority, of new users want. Is there some way to cater to such "non-perfect" use-cases without making unacceptable trade-offs for the ideal case? 
Turning these users away until they are "more mature" seems like a shame to me. For a lot of them, even if this exposes them to a higher risk of monitoring outage, it is still a _vastly_ better situation than they would have without Prometheus.

brancz:
Note that we do understand the need for easier monitoring across namespaces, we just need to make sure we have well thought through approaches, as once we add these things to the API, it's hard to ever remove them.

### What already exists?

Firstly I want to make sure everyone is aware, that there is already a mechanism to monitor Services from another namespace, the [`namespaceSelector`](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#namespaceselector) of a [`ServiceMonitor`](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#servicemonitorspec) allows selecting Services from a different namespace than the ServiceMonitor. It also supports specifying multiple namespaces.

I ask you to look at your concrete use cases and evaluate if a simple labeling scheme on Service objects with label "prometheus: true", or any other convention you want to use in your organization, would be feasible. With the namespace selector, having an Ops team managing the Prometheus servers for other teams can very well be done in another namespace and just have the namespace selector select the namespace(s) for the respective teams. Maybe all it takes is clearer documentation and examples around this. If you really want to maintain a single Prometheus object, then having a single place for its configuration is actually a good thing.

### The discussion so far

In terms of memory consumption, having multiple Prometheus server actually only comes with little overhead. Note that Prometheus 2.0 works by having the Kernel decide when to evict memory, this in turn means Prometheus will take all the memory that is available, but it also automatically evicts the memory when any other process wants to allocate it, and Prometheus "degrades" in performance, which in most cases doesn't actually mean it degrades in performance, just that it doesn't have all the storage in memory. Realistically you can ingest a reasonable amount of metrics per second with little memory dedicated to Prometheus. On Kubernetes the resource limit is what the Kernel will end up using as the threshold when it will stop allocating memory.

A less nice thing from the technical side of enabling Prometheus servers to select ServiceMonitors from any namespace is that currently only Prometheus objects of a namespace get reconciled, when there is a change to any ServiceMonitor/ConfigMap/Secret in the same namespace, as any of those can influence the state the Prometheus Pod(s) should be in. Changing this to be global can have a significant impact on the usability of the Prometheus Operator as any change in the cluster to any of those object kinds reconciles all Prometheus objects. So in fact even if that's what we want to do, this PR is not the complete solution.

Some thoughts on the concrete points discussed:

* Have a central place to read metrics from for autoscaling.
  * This is a valid point, and the Federation answer is not great either, as it requires additional manual input whenever one wants to setup autoscaling, in addition to just getting the metric/query right. Problem is that eventually a single Prometheus server doesn't cut it, and then we're in front of the same problem all over again.
* Having a central place to read metrics from for dashboarding.
  * Grafana actually allows specifying multiple datasources, so this may primarily be a tooling problem.
* Enabling a Prometheus Pod to run alerting rules globally.
  * Monitoring must be the most reliable system in the entire infrastructure. Some companies actually have scheduling priorities such as `normal < critical < monitoring`, simply because monitoring is that important. Enabling users to influence a Prometheus server on a global level is not acceptable, as it can make the whole system crash and in turn leaves the whole cluster not being monitored.

### What can we do?

We believe some of the above issues can be solved if there was a system that allowed us to write metrics a Prometheus server gathers locally into a global metrics sink. Fortunately such a system exists: [Thanos](https://github.com/improbable-eng/thanos). Thanos is a Prometheus long term storage.

The way Thanos works is by having an unmodified, upstream Prometheus with a sidecar, that exposes a storage API. The storage API can be used by the "querier" component, which allows fanning out Prometheus queries to multiple Prometheus servers. Additionally once Prometheus finishes a block of data (typically 2h blocks), it can be written to long term storage, which is any S3 compatible object storage. The querier transparently queries the "live" Prometheus servers as well as data from object storage and returns a single result from all data. The storage part here is not strictly necessary, one can also just use the querier without object storage if all one wants is the global view.

Alerting in Thanos is still local to the Prometheus servers, as this does not come with any additional indirections, and we have the same reliability guarantees that we appreciate about Prometheus. This also means that if a single Prometheus server is bricked due to faulty rules, then it only affects that one server, not the entire infrastructure.

Implementing support for Thanos in the Prometheus Operator would still keep the same idioms of configuring a single Prometheus server as it is done today with the Prometheus Operator, but it gives the advantage of having the global view on metrics which is important for autoscaling as well as makes dashboarding a lot easier.

cc @fabxc @ant31 

I apologize that this comment ended up rather in blog post length than a comment, I just wanted to make sure I include and address as much as possible.

Let me know what you think!

midnightconman:
It seems like there is a use for this kind of namespace selection logic, would you guys be welcome to a PR that would add a feature flag (maybe something like `--strict-namespace-selection`) to control this behavior? The default could be `true` for backward compatibility?

cornelius-keller:
Hi all, I am working pretty much in the same situation as @midnightconman and others in this thr ead described, and I also needed to have self service monitoring in a setup where I also need a central prometheus setup both for acceptance and maturity of devops in the teams and company policies and compliance requirements.

I think I found a valid solution now for my setup based on the suggestions of @brancz and wanted to share it:

I used the helm packaging for the setup, and I put the values file I used in a git project. In this values there is a set of predefined ServiceMonitors selecting services via labels that match our most common use cases. For example there is one that is compatible for the prometheus exporter that we agreed to use in all springboot services, another one to monitor all elasticsearch clusters that we might have in the cluster (currently one for log aggregation) etc.

So the most common usecases are self service out of the box just by labeling your service accordingly.
Whenever somebody needs a custom service monitor configuration, for example because of a different metrics path or other authentication etc. He should create a pull request to the values file for the helm chart containing the service monitor definitions.
By this it can be tested and reviewed and should not bring down the prometheus cluster.

Same applies for custom grafana dashboards, alerting rules etc.

I also think that it should be as easy as possible to hook up your service with monitoring and therefore I support the intention of @midnightconman .
On the other hand I think in combination with the possibility for developers to influence the prometheus setup via pull requests you get a pretty complete solution.

Just my 2 cents / Hope that Helps

gmauleon:
Just adding my 2 cents to the discussion.

I must say I agree with @midnightconman on having the capabilities to select ServiceMonitor across namespaces. It would really ease the standardization and adoption of the prometheus operator to have that as a valid choice. What about prometheus rules files across namespaces btw?

Our ideal case here is to deliver a "technology" in the form of an helm chart with servicemonitor and prometheus rules included (and ideally grafana dashboards or any other pieces that we think should be part of the "technology"). Think every chart out there could come with monitoring/metrics integrated with prometheus out of the box.

Right now this force me to have a prometheus per namespace which is not practical at all because we're using namespaces to isolate different piece of technology or services we provide, so it just become a proliferation of Prometheus with all the overhead it provides.

IMHO, security shouldn't be a concern if people have the choice to constraint the prometheus "servicemonitor selector" to a single namespace if they want, very much like the servicemonitor do for endpoints.

I don't know the technical side at all here, I would just love to have the choice :)

midnightconman:
This functionality was merged and should be in the next release, thank you everyone for all the discussions and work. :D

dmfenton:
@midnightconman Can you summarize the new functionality that is being added?

mxinden:
@dmfenton You find details to the new feature here: https://github.com/coreos/prometheus-operator/pull/1227

This feature will become available with `v0.19.0`: https://github.com/coreos/prometheus-operator/pull/1262

dmfenton:
Fantastic, thank you very much!

PR #822, opened 2017-12-14
Add cert expiration rule

ant31:
if there's more than 1 certs that's going to expire on a bucket, sum is > 0 and fires an alert. 
Send a warning for 7 days, 1 day, and critical if expired.
(We can remove the 1 day if it's redundant)

### DIFF
--- a/contrib/kube-prometheus/assets/prometheus/rules/kubernetes.rules.yaml
+++ b/contrib/kube-prometheus/assets/prometheus/rules/kubernetes.rules.yaml
@@ -84,3 +84,17 @@ groups:
     annotations:
       description: No API servers are reachable or all have disappeared from service
         discovery
+
+  - alert: K8sCertificateExpirationNotice
+    labels:
+      severity: warning
+    annotations:
+      description: Kubernetes API Certificate is expiring soon (less than 7 days)
+    expr: sum(apiserver_client_certificate_expiration_seconds_bucket{le="604800"}) > 0
+
+  - alert: K8sCertificateExpirationNotice
+    labels:
+      severity: critical
+    annotations:
+      description: Kubernetes API Certificate is expiring in less than 1 day
+    expr: sum(apiserver_client_certificate_expiration_seconds_bucket{le="86400"}) > 0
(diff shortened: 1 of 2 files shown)


contrib/kube-prometheus/assets/prometheus/rules/kubernetes.rules.yaml
```diff
@@ -84,3 +84,24 @@ groups:
...
         discovery
+
+  - alert: K8sCertificateExpirationNotice
+    labels:
+      severity: warning
+    annotations:
+      description: Kubernetes API Certificate is expiring soon (less than 7 days)
+    expr: sum(apiserver_client_certificate_expiration_seconds_bucket{le="604800"}) > 0
+
+  - alert: K8sCertificateExpirationLastDayNotice
```
brancz: Typically alert names are chosen the same, with different severity, that way we can inhibit the less severe alerts when an equally named one is firing, so this would also be `K8sCertificateExpirationNotice`.

contrib/kube-prometheus/assets/prometheus/rules/kubernetes.rules.yaml
```diff
@@ -84,3 +84,24 @@ groups:
...
+    labels:
+      severity: warning
```
brancz: I would also make this `severity: critical`, although not yet interrupting service, this will definitely be causing outages within 24 hours.

mbenabda:
unfortunately, this won't work if merged as is since the kube-api-exporter chart has been removed from the kube-prometheus dependencies: `apiserver_client_certificate_expiration_seconds_bucket` is not available anymore

brancz:
@mbenabda I'm not sure I understand. This PR is about the bare kube-prometheus manifests as opposed to the charts.

mbenabda:
@brancz oh then it's me who didn't understand, sorry. Can i ask you why there is this duplication ? can't the bare manifests be generated off of the chart ?

brancz:
@mbenabda first and foremost because the bare manifests were there before the charts :slightly_smiling_face:. We may generate them one day from the charts, but right now the charts are not the kind of quality that we imagine, until then we'll be working the other way around. Other than that CoreOS doesn't use these charts.

mbenabda:
@brancz ok, thanks for the info :+1:

PR #797, opened 2017-12-01
Bump prometheus chart to v2.0

gianrubio:
This is a big PR to update all helm charts, I Bumped all helm componentes to the latest version. 
- prometheus to 2.0.0
- prometheus-operator to 0.15.0
- alertmanager to 0.12.0
- node-exporter to 0.15.1
- kube-state 1.1.0

I also wrote a [python script](https://github.com/gianrubio/prometheus-operator/blob/59421f7018bbeaabbec27631d882b6a7ab521a1d/helm/hack/sync_kube_prometheus.py) to keep kube-prometheus, alerts and dashboards, in sync with with kube-prometheus charts.

### How to test?
```
git clone https://github.com/gianrubio/prometheus-operator/
cd prometheus-operator
git checkout helm-prometheus-2.0
sed -ie 's/    repository/#    repository/g' $(pwd)/helm/*/requirements.yaml
sed -ie 's/#e2e-repository/repository/g' $(pwd)/helm/*/requirements.yaml
helm dep update helm/kube-prometheus 
helm install  --name prometheus-operator helm/prometheus-operator --namespace monitoring  
helm install  --name kube-prometheus helm/kube-prometheus --namespace monitoring  
```

PR #829, opened 2017-12-15
support prometheus-operator below v0.12.0 (eg: for k8s 1.6)

mbenabda:
This PR solves 2 issues:

- prometheus-operator has started using CRDs instead of TPRs from v0.12.0. While allowing to specify the operator's image tag to use, the charts in the helm/ directory of this repository were forcing the use of CRDs, thus preventing users of kubernetes 1.6.x from using it. 

- When using kube-prometheus with ```rbacEnable``` set to ```false```, the value was not propagated to the dependencies. This was preventing us from installing the chart in clusters without rbac (the dependencies would try to install rbac resources, and fail doing so).

Please note that this PR breaks backward compatibility in the sense that ```rbacEnable``` has been moved to the global scope in all the charts having such a value

### DIFF
--- a/helm/exporter-kube-controller-manager/templates/kube-controller-manager.rules.yaml
+++ b/helm/exporter-kube-controller-manager/templates/kube-controller-manager.rules.yaml
@@ -3,7 +3,7 @@ groups:
 - name: kube-controller-manager.rules
   rules:
   - alert: K8SControllerManagerDown
-    expr: absent(up{job="{{ template "fullname" .  }}"} == 1)
+    expr: absent(up{job="{{ template "exporter-kube-controller-manager.fullname" .  }}"} == 1)
     for: 5m
     labels:
       severity: critical
--- a/helm/prometheus-operator/templates/clusterrole.yaml
+++ b/helm/prometheus-operator/templates/clusterrole.yaml
@@ -7,11 +7,11 @@ apiVersion: rbac.authorization.k8s.io/v1alpha1
 kind: ClusterRole
 metadata:
   labels:
-    app: {{ template "name" . }}
+    app: {{ template "prometheus-operator.name" . }}
     chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-  name: {{ template "fullname" . }}
+  name: {{ template "prometheus-operator.fullname" . }}
 rules:
 - apiGroups:
   - extensions
@@ -28,8 +28,11 @@ rules:
 - apiGroups:
   - monitoring.coreos.com
   resources:
+  - alertmanager
   - alertmanagers
+  - prometheus
   - prometheuses
+  - service-monitor
   - servicemonitors
   verbs:
   - "*"
--- a/helm/prometheus/templates/configmap.yaml
+++ b/helm/prometheus/templates/configmap.yaml
@@ -7,7 +7,7 @@ metadata:
     heritage: {{ .Release.Service }}
     prometheus: {{ .Release.Name }}
     release: {{ .Release.Name }}
-  name: {{ template "fullname" . }}
+  name: {{ template "prometheus.fullname" . }}
 data:
 {{- if .Values.prometheusRules }}
 {{- $root := . }}
--- a/helm/prometheus/templates/configmap.yaml
+++ b/helm/prometheus/templates/configmap.yaml
@@ -7,6 +7,7 @@ metadata:
     heritage: {{ .Release.Service }}
     prometheus: {{ .Release.Name }}
     release: {{ .Release.Name }}
+    role: alert-rules
   name: {{ template "prometheus.fullname" . }}
 data:
 {{- if .Values.prometheusRules }}
--- a/helm/alertmanager/Chart.yaml
+++ b/helm/alertmanager/Chart.yaml
@@ -9,4 +9,4 @@ maintainers:
 name: alertmanager
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.6
+version: 0.0.7
--- a/helm/alertmanager/templates/_helpers.tpl
+++ b/helm/alertmanager/templates/_helpers.tpl
@@ -2,7 +2,7 @@
 {{/*
 Expand the name of the chart.
 */}}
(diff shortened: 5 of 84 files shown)


brancz:
I only flew over this, but it generally looks ok to me. @gianrubio PTAL.

mbenabda:
thanks @brancz for looking at this PR !

gianrubio:
@mbenabda thanks for your PR. Please wait until #797 is merged, after that I'll review your PR

mbenabda:
@gianrubio you're welcome !

gianrubio:
@mbenabda all the PR got merged, could you please rebase this PR?

mbenabda:
@gianrubio done ! i added a commit to make template names chart-specific in order to prevent overlapping of template names (see https://github.com/kubernetes/helm/blob/master/docs/chart_template_guide/named_templates.md#declaring-and-using-templates-with-define-and-template)

helm/exporter-kube-controller-manager/templates/kube-controller-manager.rules.yaml
```diff
@@ -3,7 +3,7 @@ groups:
 - name: kube-controller-manager.rules
   rules:
   - alert: K8SControllerManagerDown
-    expr: absent(up{job="{{ template "fullname" .  }}"} == 1)
+    expr: absent(up{job="{{ template "exporter-kube-controller-manager.fullname" .  }}"} == 1)
```
gianrubio: All the helm/*/rules.yaml are generated by [sync_kube_prometheus](https://github.com/coreos/prometheus-operator/blob/master/helm/hack/sync_kube_prometheus.py) script. Rather than edit manually this files, I would recommend to make the changes on the script and generate all the rules.

helm/prometheus-operator/templates/clusterrole.yaml
```diff
@@ -28,8 +24,11 @@ rules:
 - apiGroups:
   - monitoring.coreos.com
   resources:
+  - alertmanager
```
gianrubio: Is this resource required?

mbenabda: ```kubectl get thirdpartyresource``` yields:
```
NAME                                    DESCRIPTION                           VERSION(S)
alertmanager.monitoring.coreos.com      Managed Alertmanager cluster          v1alpha1
prometheus.monitoring.coreos.com        Managed Prometheus server             v1alpha1
service-monitor.monitoring.coreos.com   Prometheus monitoring for a service   v1alpha1
```

Prometheus-operator below 0.12.0 creates the ```alertmanager```, ```prometheus``` and ```service-monitor``` thirdpartyresources in group ```monitoring.coreos.com```.

Not being well versed in rbac, I assumed the ClusterRole's responsibility was to describe the set of accesses on resources that prometheus-operator's ServiceAccount requires. Hence my additions to the clusterrole.yaml. please let me know if these need to be removed.

helm/exporter-kube-controller-manager/templates/kube-controller-manager.rules.yaml
mbenabda: ok, will do. thank's for the info

mbenabda:
@gianrubio i've updated the sync_kube_prometheus script, regenerated the rules and dashboards, and bumped the versions. thanks for your feedbacks

mbenabda:
@gianrubio Are there any other changes i should apply before this PR can get merged ? thanks

gianrubio:
@mbenabda your PR looks good on the changes for the backward compability of crd and tpr. 

I'm a bit affraid to introduce the global values for rbac because it can break the way that users install the charts. Rather than introduce this global variable, what about to have all the variables from the dependents charts in the `kube-prometheus/values.yaml` and document to the users how to use the set flag? Could be something simple like the comand line below. 

## Installing with RBAC
`helm install coreos/kube-prometheus --name kube-prometheus --set rbacEnable=true --namespace monitoring` 

what are you thoughts?

mbenabda:
@gianrubio Hello, following up on your last preview coment:
I understand your concerns: I'm going to remove the rbac changes from this PR and make another one, so we can move forward with the crd/tpr issue.

helm/alertmanager/README.md
```diff
@@ -43,6 +43,7 @@ The following tables lists the configurable parameters of the alertmanager chart
 
 Parameter | Description | Default
 --- | --- | ---
+`global.rbacEnable` | If true, create & use RBAC resources | `true`
```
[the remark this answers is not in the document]
brancz: Maybe a personal preference, but I like to have ServiceAccounts for all applications I deploy, even those that don't get any roles. I do this so I don't accidentally bind anything to a `default` ServiceAccount and grant an application access it shouldn't have.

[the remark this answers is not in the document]
gianrubio: Good point @brancz I liked your suggestion

gianrubio:
@mbenabda yes, go ahead

helm/prometheus/templates/configmap.yaml
```diff
@@ -7,7 +7,8 @@ metadata:
     heritage: {{ .Release.Service }}
     prometheus: {{ .Release.Name }}
     release: {{ .Release.Name }}
-  name: {{ template "fullname" . }}
+    role: alert-rules
```
gianrubio: Could you explain why did you add this label role?

mbenabda: when i added chart specific prefixes to the template names to prevent overlapping, it broke the rules discovery powered by https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus/templates/prometheus.yaml#L72
```yaml
ruleSelector:
    matchLabels:
      app: {{ template "name" . }}
      prometheus: {{ .Release.Name }}
```
because in order for this selector to discover rules declared in downstream charts, they are required to rely on the "name" template name to resolve the same as in the parent chart.

One solution would have been to remove the app label alltogether, but the selector would've then matched all the configmaps declared in the chart, amongst which some are not used to define alerting rules.

another one would have been to allow for custom labels to be defined on alerts-defining configmaps to provide a way for the user to make them match with the selector

gianrubio:
thanks @mbenabda

PR #831, opened 2017-12-18
Fix helm sync

gianrubio:
@ant31 this will fix the failure before the s3 sync and the hash comparison


brancz:
let's give it a spin :)

PR #832, opened 2017-12-19
Regenerate prometheus-k8s-rules.yaml

mrtyler:
after change to prometheus.rules.yaml in c4be29f65.

IOW, this is the result of running `make generate` with current master.


brancz:
Thanks! We should really put that in our CI jobs to make sure the generated content is always up to date.

PR #836, opened 2017-12-20
remote storage config

sclasen:
add remote_read and remote_write to prometheus config generation for version 2+.

- [remote_write](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<remote_write>)
- [remote_read](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<remote_read>)

Opening this PR now to get feedback on the likelihood of acceptance before finishing it off. Please let me know if you would like me to continue.

Still needs

- [x] basic auth and bearer tokens stored in secrets


brancz:
I'm happy with merging this, once finished. Only thing is since remote read/write are still young and experimental in Prometheus we need to declare them as such here as well.

sclasen:
thanks @brancz, by marking as experimental, do  you just mean adding some language to the docs or is there something more formal?

Will finish the rest up, thanks for the feedback

brancz:
Yeah just the language in the docs, just write "This is an experimental feature, it may change in any upcoming release in a breaking way." to both of the comments that end up in the generated API docs.

Thanks a lot for taking care of this. Highly highly appreciated!

sclasen:
@brancz this is all set from a feature-completeness perspective. anything else you'd like to see?

brancz:
This looks good to me, I would like to find an abstraction for the relabel configs, but I don't think in this case there is anything we can do.

Thanks a lot!

sclasen:
Cheers!

PR #843, opened 2017-12-22
Support dropping metrics at scrape time

### LINKED ISSUE #826: Feature Request: support dropping metrics at scrape time
We recently ran a batch of about 2 million k8s jobs in under 24 hours, which led to some serious time series churn since the kubelets and kube-state-metrics expose metrics for every pod, and Prometheus was inevitably OOM-killed even with the memory requests/limits set to over 50Gi.

Our temporary solution was to simply stop scraping those sources entirely to prevent Prometheus from getting overloaded. This works, but is a bit excessive since we still want to collect those metrics for longer-running pods.

Ideally, we'd be able to configure the prometheus operator to ignore those metrics when scraping those sources. Are there any thoughts on adding this functionality to the operator? Maybe adding a `dropTargets` field to the Endpoint spec, something like this:

```yaml
dropTargets:
- match: my-job-pod-.+
  source:
  - pod_name
```

The above would add the following to the `relabel_configs` in the Prometheus configuration generated for that ServiceMonitor:

```yaml
- source_labels: [__meta_kubernetes_pod_name]
  separator: ;
  regex: (my-job-pod-.+)
  action: drop
```

Currently, the only long-term solution is to use a custom Prometheus configuration, and I'd rather not have to give up all the other functionality in the prometheus operator associated with doing that.

Capitrium:
We've found that heavy use of the k8s batch api leads to some serious time series churn, since the kubelets and kube-state-metrics expose metrics for every pod, which eventually overloads Prometheus unless we stop scraping those targets altogether. This PR adds functionality to configure dropping selected metrics when scraped by Prometheus, rather than having to drop the entire target.

A new optional `skipMetrics` field is added to the `Endpoint` type which allows defining a search pattern and a set of source labels:

```yaml
skipMetrics:
- match: my-job-pod-.+
  source:
  - pod_name
```

This generates the following Prometheus configuration for the associated endpoint:

```yaml
metric_relabel_configs:
- source_labels: [pod_name]
  separator: ;
  regex: my-job-pod-.+
  replacement: $1
  action: drop
```

Closes #826


brancz:
For the better or worse we actually added the relabel object in #836, technically there can be any case people may want to do, so it's probably better to do so here as well.

brancz:
lgtm, thanks!

PR #844, opened 2017-12-22
merge helm index file while packaging

gianrubio:
I realize to support versioning on the helm server it's required to download the current index file from the remote server. It also avoid the error below because the kube-prometheus is missing from the index.yaml 

```
$ helm install --name kube-prometheus coreos/kube-prometheus --namespace=monitoring
Error: file "coreos/kube-prometheus" not found
```


gianrubio:
e2e tests are flake https://travis-ci.org/coreos/prometheus-operator/builds/320600382

PR #849, opened 2018-01-03
for glog, re-using flag.Commandline instead of a new flagset

mmerrill3:
Pull request for https://github.com/coreos/prometheus-operator/issues/823


brancz:
lgtm, thanks!

PR #855, opened 2018-01-04
prometheus: Add --storage.tsdb.no-lockfile flag for Prometheus 2.0

### LINKED ISSUE #854: Prometheus pod failed to start with db "Locked by other process"
**What did you do?**
Prometheus was happily collecting metrics in the k8s cluster. There was a shutdown of a machine due to power outage. When the machine was brought back up, the prometheus pod failed to start. 

v0.14.1 of the prometheus operator was running.

Expected prometheus to start up successfully again after the node restart.

The `prometheus-k8s-0` pod failed to start with a "Locked by other process" message. 

The following was found in its log.
```
level=info ts=2018-01-03T21:56:07.173825499Z caller=main.go:215 msg="Starting Prometheus" version="(version=2.0.0, branch=HEAD, revision=0a74f98628a0463dddc90528220c94de5032d1a0)"
level=info ts=2018-01-03T21:56:07.173871294Z caller=main.go:216 build_context="(go=go1.9.2, user=root@615b82cb36b6, date=20171108-07:11:59)"
level=info ts=2018-01-03T21:56:07.173888097Z caller=main.go:217 host_details="(Linux 4.13.16-coreos-r2 #1 SMP Wed Dec 6 04:27:34 UTC 2017 x86_64 prometheus-k8s-0 (none))"
level=info ts=2018-01-03T21:56:07.175870142Z caller=main.go:314 msg="Starting TSDB"
level=info ts=2018-01-03T21:56:07.175877793Z caller=web.go:380 component=web msg="Start listening for connections" address=0.0.0.0:9090
level=info ts=2018-01-03T21:56:07.175890242Z caller=targetmanager.go:71 component="target manager" msg="Starting target manager..."
level=error ts=2018-01-03T21:56:07.176126854Z caller=main.go:323 msg="Opening storage failed" err="open DB in /var/prometheus/data: Locked by other process"
```

This looks like the same root cause as this [prometheus issue](https://github.com/prometheus/prometheus/issues/2689). The only workarounds mentioned were to delete the `lock` file or else start the prometheus container with the parameter `--storage.tsdb.no-lockfile`

How can we recover from this issue? The prometheus pod needs to be able to start even though the lock file still exists from an unclean shutdown. Is there a better solution than passing the `no-lockfile` arg? Or how do we tell the prometheus operator to use that flag?

I was able to workaround this issue by adding the `no-lockfile` arg to the container args:
```
kubectl  delete deploy prometheus-operator
kubectl patch sts prometheus-k8s -p '{"spec":{"replicas":0}}'
kubectl edit sts prometheus-k8s
# Add the following argument to the prometheus container args
  - --storage.tsdb.no-lockfile
kubectl patch sts prometheus-k8s -p '{"spec":{"replicas":1}}'
```

Then I was able to connect to the container and delete the lockfile in /var/prometheus/data and restart the prometheus operator.

**Environment**

```
Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.2", GitCommit:"bdaeafa71f6c7c04636251031f93464384d54963", GitTreeState:"clean", BuildDate:"2017-10-24T19:48:57Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.11", GitCommit:"b13f2fd682d56eab7a6a2b5a1cab1a3d2c8bdd55", GitTreeState:"clean", BuildDate:"2017-11-25T17:51:39Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
```
* Kubernetes cluster kind: 
kubeadm

brancz:
Fixes #854

@ant31

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -335,21 +335,12 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMaps []
 
 		securityContext = &v1.PodSecurityContext{}
 	case 2:
-
-		// Prometheus 2.0 is in alpha and is highly experimental, and therefore
-		// flags and other things may change for the final release of 2.0. This
-		// section is also regarded as experimental until a Prometheus 2.0 stable
-		// has been released. These flags will be updated to work with every new
-		// 2.0 release until a stable release. These flags are taregeted at version
-		// v2.0.0-alpha.3, there is no guarantee that these flags will continue to
-		// work for any further version, this feature is experimental and developed
-		// on a best effort basis.
-
 		promArgs = append(promArgs,
 			fmt.Sprintf("-config.file=%s", prometheusConfFile),
 			fmt.Sprintf("-storage.tsdb.path=%s", prometheusStorageDir),
 			"-storage.tsdb.retention="+p.Spec.Retention,
 			"-web.enable-lifecycle",
+			"-storage.tsdb.no-lockfile",
 		)
 
 		gid := int64(2000)


mxinden:
@brancz what if a user accidentally mounts the same persistent volume into two Prometheus containers?

brancz:
We won't be able to prevent users from shooting themselves in the foot entirely. We explicitly create multiple PVCs, one for each Pod, if the PVC are assigned incorrectly then there is nothing that we can do (we do also [enforce to ReadWriteOnce mode](https://github.com/coreos/prometheus-operator/blob/e9c25cc89f7422d7570d8a74a48f2e80c671b434/pkg/prometheus/statefulset.go#L201)).

PR #858, opened 2018-01-05
Add support for prometheus, alertmanager to be hostNetworked

fatsheep9146:
Prometheus Operator is great.

In our company, the prometheus and alertmanager belongs to system components which are very important and must be hostnetworked , so we add prometheus-operator to support deploying hostnetworked prometheus and alertmanager.

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -98,6 +98,7 @@ Specification of the desired behavior of the Alertmanager cluster. More info: ht
 | resources | Define resources requests and limits for single Pods. | [v1.ResourceRequirements](https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#resourcerequirements-v1-core) | false |
 | affinity | If specified, the pod's scheduling constraints. | *v1.Affinity | false |
 | tolerations | If specified, the pod's tolerations. | []v1.Toleration | false |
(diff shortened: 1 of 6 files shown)


brancz:
Can you elaborate why you need it to be host networking?

cofyc:
@brancz 

Hi, we developed a neutron-based SDN network plugin for multi-tenant purpose. In this network model, pod network in each namespace are isolated, and host-network can access pod-network but not vice versa. This prohibits user pods access our internal components. 

Prometheus is one of our internal components, it needs to access all user pod network to scrape metrics, but are not allowed to be accessed from pod network. So it needs to be in host network.

This PR enables promethues-opetator to deploy `Prometheus` and `Alertmanager` in host network, makes it possible to use in this special case.

brancz:
It seems questionable to me how applicable this is to all users. Have you tried achieving this with PodPresets instead?

cofyc:
@brancz 
I didn't try `PodPresets`, but from its [docs](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/service-catalog/pod-preset.md#abstract) and [spec](https://kubernetes.io/docs/api-reference/v1.9/#podpreset-v1alpha1-settings), it only supports mounting volumes and setting environment variables.

PR #862, opened 2018-01-05
Fix link to RBAC manifest

richarddli:
The current link only links to one of the manifests you need, so I thought changing it to the folder made more sense.

### DIFF
--- a/Documentation/user-guides/getting-started.md
+++ b/Documentation/user-guides/getting-started.md
@@ -146,7 +146,7 @@ spec:
         app: example-app
     spec:
       containers:
-      - name: example-app 
+      - name: example-app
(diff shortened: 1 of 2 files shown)


Documentation/user-guides/getting-started.md
```diff
@@ -146,7 +146,7 @@ spec:
         app: example-app
     spec:
       containers:
-      - name: example-app 
+      - name: example-app
```
brancz: This is an embedded yaml located in `../../example/user-guides/getting-started/example-app-deployment.yaml`, could you make this whitespace change there as well? Then CI should be green.

brancz:
lgtm, thanks!

Issue #861, opened 2018-01-05
[kube-prometheus] Helm chart won't deploy without TPR/CRDs in place


whereisaaron:
**What did you do?**

```
helm repo add coreos https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/
helm install coreos/kube-prometheus
```

**What did you expect to see?**

Prometheus deployed

**What did you see instead? Under which circumstances?**

```
Error: apiVersion "monitoring.coreos.com/v1" in kube-prometheus/charts/exporter-kubernetes/templates/servicemonitor.yaml is not available
```

**Environment**

Kubernetes 1.8 with RBAC

* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.5", GitCommit:"cce11c6a185279d037023e02ac5249e14daa22bf", GitTreeState:"clean", BuildDate:"2017-12-07T16:16:03Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"8+", GitVersion:"v1.8.4+coreos.0", GitCommit:"4292f9682595afddbb4f8b1483673449c74f9619", GitTreeState:"clean", BuildDate:"2017-11-21T17:22:25Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: 

[`kube-aws`](https://github.com/kubernetes-incubator/kube-aws)

* Manifests:

Used the [Helm chart](https://github.com/coreos/prometheus-operator/tree/master/helm).

* Discussion:

Looks like a chicken and egg problem with the chart and CRDs. Could we have a `helm/README.md` that gives us the steps to deploy with Helm? Maybe a Helm chart just to create the CRDs, then a second chart to actually deploy?

whereisaaron:
Found a [comment in another issue](https://github.com/coreos/prometheus-operator/issues/592#issuecomment-355142097) by @gianrubio. Looks like the missing bootstrap info for `helm/README.md`!


> Charts finally got synced with s3, everyone can enjoy prometheus 2.0 using helm :)
```
$ helm repo add coreos https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/ …
$ helm install coreos/prometheus-operator --name prometheus-operator
$ helm install coreos/kube-prometheus --name kube-prometheus --set rbacEnable=true
```

[...]

PR #863, opened 2018-01-05
A stub of a README for the Helm chart with vital bootstrap info

whereisaaron:
As discussed in #861. It's not detailed, but it is 100% better that current non-existent README :-)

It provide the vital info that you need to install `prometheus-operator` first, and then `kube-prometheus` second.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -1,0 +1,7 @@
+# TL;DR
+
+```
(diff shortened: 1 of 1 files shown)


gianrubio:
Thanks @whereisaaron

PR #864, opened 2018-01-05
storagespec: add EmptyDirVolumeSource

### LINKED ISSUE #851: Support tmpfs as a storage option
Instead of defaulting to emptydir volumes, I'd like to use a tmpfs volume. Is this a change that would be accepted upstream?

This section would need to allow for the EmptyDir.Medium to be set to "Memory"
https://github.com/coreos/prometheus-operator/blob/281d3e7d5c26e72ad66871cee503bc183c1bc6a1/pkg/prometheus/statefulset.go#L137-L142

To support this, I think an additional configuration would need to be added to the PrometheusSpec. Open to suggestions or alternate solutions.

crandles:
Adds EmptyDirVolumeSource to the `StorageSpec` and adds support for it to AlertManager & Prometheus. Resolves #851


brancz:
I believe the CI failure was a flake, I restarted it.

crandles:
Thanks. I couldn't tell why that was erroring

Documentation/api.md
```diff
@@ -320,6 +320,7 @@ StorageSpec defines the configured storage for a group Prometheus servers.
 | Field | Description | Scheme | Required |
 | ----- | ----------- | ------ | -------- |
 | class | Name of the StorageClass to use when requesting storage provisioning. More info: https://kubernetes.io/docs/user-guide/persistent-volumes/#storageclasses DEPRECATED | string | true |
+| emptyDir | EmptyDirVolumeSource to be used by the Prometheus StatefulSets. If specified, used in place of any volumeClaimTemplate. More info: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir | *v1.EmptyDirVolumeSource | false |
```
crandles: Just noticed this. How do I get the doc generator to link this `*v1.EmptyDirVolumeSource ` to `https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#emptydirvolumesource-v1-core` ?

brancz: you need to add the link to the mapping at the beginning of the file

https://github.com/coreos/prometheus-operator/blob/6bfeae2e69e8ce438963345d63a290082ecdd9b5/cmd/po-docgen/api.go#L42-L50

it's a bit tribal, but it's gotten us far :slightly_smiling_face: , maybe one day I'll find some time to automate that table away

brancz:
lgtm, thanks!

PR #869, opened 2018-01-10
kube-prometheus: Add support for off-cluster etcd

jjungnickel:
This will allow users to specify a list of ip-addresses for their etcd installation and will allow the chart to be used as-is without manual manifests like https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus/manifests/etcd


brancz:
Nice addition! I'll leave it to @gianrubio to review in detail.

gianrubio:
Lgtm, thanks @jjungnickel

PR #874, opened 2018-01-10
helm/prometheus: support not having a serviceMonitorSelector

crandles:
The prometheus helm chart supports passing in a static config, but because the serviceMonitorsSelector is defined by default, it is not usable. If a config is defined, the serviceMonitorsSelector shouldn't be set.


gianrubio:
To fix the ci errror you need to bump the prometheus version in the requirements of kube-prometheus

crandles:
Thanks, CI looks good now

gianrubio:
Thanks

PR #886, opened 2018-01-13
kube-prometheus: Add missing RBAC rules to kube-state-metrics

zerda:


helm/exporter-kube-state/templates/clusterrole.yaml
```diff
@@ -23,6 +23,8 @@ rules:
   - replicationcontrollers
   - limitranges
   - persistentvolumeclaims
+  - persistentvolumes
```
champtar: Might be on purpose, but you are missing `- namespaces`

zerda: `namespaces` already exists in this file.

brancz:
Thanks!

PR #895, opened 2018-01-16
Added example on how to create custom Service Monitor

alexanderdavidsen:


helm/prometheus/README.md
```diff
@@ -87,6 +87,34 @@ $ helm install coreos/prometheus --name my-release -f values.yaml
...
+
+### Service Monitors
+
+Custom service monitors can be added in values.yaml in the `serviceMonitors` section. Please refere to `values.yaml` for all available parameters.
+
+
+#### Example service monitor
+
+
+This example Service Monitor will monitor applications matching `app: nginx-ingress`. The port `metrics` will be scraped with the path `/merics`. The endpoint will be scraped every 30 seconds.
```
gianrubio: Could you add a warning telling the users to check if the labels in the serviceMonitor match the labels defined in the crd from prometheus?

alexanderdavidsen:
Sure. Updated now.

gianrubio:
Thanks @alexanderdavidsen

helm/prometheus/README.md
javajon: Correct "path `/merics" with "path `/metrics" above.

gianrubio: @javajon would you like to make a PR to fix that?

Hashfyre:
@alexanderdavidsen this PR didn't propagate to https://github.com/helm/charts/tree/master/stable/prometheus-operator when the migration was done.

Also, other chart-devs may not be aware. Currently the `stable/prometheus-operator` chart enforces the following label for `serviceMonitorSelector`
```
> kubectl get prometheus -o yaml --all-namespaces -o \ jsonpath='{.items[*].spec.serviceMonitorSelector.matchLabels}'

map[release:k8s]
```

However https://github.com/helm/charts/tree/master/stable/nginx-ingress gets created with `release: nginx-ingress` and hence the ServiceMonitor never gets picked up. 

The solution is buried in this post migration PR and not in the current documentation. I can make a PR on https://github.com/helm/charts/tree/master/stable/prometheus-operator to fix this.

PR #897, opened 2018-01-16
Use whole numbers in replica graphs

zmlka:
No need to have decimals on the axes showing the number of replicas

### DIFF
--- a/contrib/kube-prometheus/manifests/grafana/grafana-dashboards.yaml
+++ b/contrib/kube-prometheus/manifests/grafana/grafana-dashboards.yaml
@@ -639,6 +639,7 @@ data:
                   "format": "none",
                   "label": "",
                   "logBase": 1,
+                  "decimals":0,
                   "show": true
                 },
                 {
@@ -7379,6 +7380,7 @@ data:
                   "format": "none",
                   "label": "",
                   "logBase": 1,
+                  "decimals":0,
                   "show": true
                 },
                 {


brancz:
Thanks! You're absolutely right. The Grafana dashboards are generated by the python scripts in `contrib/kube-prometheus/assets/grafana/`. Could you adapt this there as well and run `make generate` either in the root of the repo or in `contrib/kube-prometheus`?

zmlka:
Aha. That would explained the checks failing :sweat_smile: 
Currently behind a proxy and having trouble with `make generate`. I'll get to it tonight.

brancz:
sounds good to me, thanks a lot!

zmlka:
Seems to me that your grafanalib does not support `decimals`, although [upstream does](https://github.com/weaveworks/grafanalib/blob/v0.4.0/grafanalib/core.py#L299).

For example. this generates a dashboard just fine:
```
yAxes=YAxes
    YAxis(min=None, label='bla bla'),
    YAxis(format='short', min=None, show=False),
```
while this fails:
```
yAxes=YAxes
    YAxis(min=None, decimals=0),
    YAxis(format='short', min=None, show=False),
```

I tried adding `decimals` to the function definition in `_grafanalib.py`, but that didn't seem to be the necessary quick fix. Could you check it out?

zmlka:
(I mis-clicked "close and comment")

brancz:
That's unfortunate. I've been meaning to convert the python scripts to jsonnet anyways, then we could add whatever we wanted. Maybe we should just let this be until we do that and keep this open as a reminder. Wdyt?

zmlka:
Sure, I'll gladly revisit this in the future.

Your call about keeping it open - it's your project. 
I won't be hurt if you close this not-so-significant cosmetic change :).

metalmatze:
Sadly kube-prometheus has moved on to a jsonnet based stack which also includes the dashboards. Hence, I will close this PR as outdated.
Feel free to open a newer PR if the problem still exists today.

PR #901, opened 2018-01-17
Add role rules permissions to get metrics from kubelet

luisyonaldo:
Fixes https://github.com/coreos/prometheus-operator/issues/867

### DIFF
--- a/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-roles.yaml
+++ b/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-roles.yaml
@@ -47,5 +47,9 @@ kind: ClusterRole
 metadata:
   name: prometheus-k8s
 rules:
+- apiGroups: [""]
+  resources:
+  - nodes/metrics
+  verbs: ["get"]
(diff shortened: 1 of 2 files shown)


brancz:
Thanks @luisyonaldo! Could you add this [here](https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-roles.yaml#L49-L51) as well and run `make generate` in the root of the repo? Thanks!

luisyonaldo:
nothing change after `make generate`, is that the expected behaviour?

luisyonaldo:
also, on which doc should be stated that `kubelet` must be running with `--authentication-token-webhook` and `--authorization-mode=Webhook` options?

brancz:
I felt like the file is used to generate some content, but CI will tell us whether that's true or not :slightly_smiling_face: if you say so then CI should be green and then this would be good to merge for me.

luisyonaldo:
looks like the file is not being used to generate something

brancz:
Perfect, thanks!

PR #791, opened 2017-11-28
[WIP] tests: Run scripts in parallel on travis

mxinden:
See further details here:
https://docs.travis-ci.com/user/build-stages/#How-to-define-Build-Stages%3F

PR #906, opened 2018-01-19
Run e2e tests in parallel

gianrubio:
Speed up travis tests running simultaneously e2e tests 

Similar to #791

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,9 @@ jobs:
     script: ./scripts/govendor-ensure.sh
   - stage: Unit tests
     script: make test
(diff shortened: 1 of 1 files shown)


gianrubio:
Took the same time to run in parallel or sequential :(

brancz:
right, but they were started at the same time, no?

gianrubio:
Yes, they run at the same time. I did one more change and now all the sanity check and unit tests are grouped. The total time decreased from 40m to 28m, any thoughts?

brancz:
Other than that this is awesome? No :) thanks! Nice work as always!

PR #911, opened 2018-01-22
*: Use non-root

brancz:
@ant31

### DIFF
--- a/contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/templates/grafana-deployment-template.yaml
+++ b/contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/templates/grafana-deployment-template.yaml
@@ -9,9 +9,12 @@ spec:
       labels:
         app: grafana
     spec:
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
       containers:
       - name: grafana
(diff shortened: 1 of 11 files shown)


contrib/kube-prometheus/hack/grafana-dashboards-configmap-generator/templates/grafana-deployment-template.yaml
```diff
@@ -9,6 +9,9 @@ spec:
       labels:
         app: grafana
     spec:
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
```
ant31: what's that id?

brancz: > The UID 65534 is commonly reserved for nobody, a user with no system privileges, as opposed to an ordinary (i.e., non-privileged) user. This UID is often used for individuals accessing the system remotely via FTP (file transfer protocol) or HTTP (hypertext transfer protocol).

brancz:
Do not merge this yet, I discovered there are some filesystem issues with the grafana pods.

brancz:
@ant31 this is now ready for review

PR #917, opened 2018-01-25
Make RBAC global

djsd123:
At present, you can't deploy this chart to a cluster that does __Not__ have `RBAC` enabled without knowing which sub-charts have `RBAC` resources included.

The docs state that one can toggle `RBAC` when deploying the `kube-prometheus` Parent-chart by providing the `--set rbacEnable` flag.  Correct me if I am wrong but that flag is a __no-op__ (It essentially does nothing).

Both the `prometheus` and `exporter-kube-state` charts include `RBAC` resources.  Therefore, one must namespace the argument like `--set exporter-kube-state.rbacEnable <BOOLEAN>` to override the default.

I'm assuming these charts will typically be deployed together, so modifying `RBAC` to be set globally makes sense, while still allowing the sub-charts to be configured on a standalone basis. 

For example running  `helm install coreos/kube-prometheus --name kube-prometheus --set global.rbacEnable=false --namespace monitoring` will toggle `rbacEnable` to false for the `prometheus` and `exporter-kube-state` charts.

I also removed the `rbacEnable` value from the `alertmanager` chart.  At present, I cannot see where it's  being used.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -3,7 +3,7 @@
 ```
 helm repo add coreos https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/
 helm install coreos/prometheus-operator --name prometheus-operator --namespace monitoring
-helm install coreos/kube-prometheus --name kube-prometheus --set rbacEnable=true --namespace monitoring
+helm install coreos/kube-prometheus --name kube-prometheus --set global.rbacEnable=true --namespace monitoring
 ````
 
 # How to contribue?
--- a/helm/exporter-kube-state/Chart.yaml
+++ b/helm/exporter-kube-state/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart singleton for kube-state-metrics
 name: exporter-kube-state
-version: 0.1.7
+version: 0.1.8
 maintainers:
   - name: Michael Goodness
     email: mgoodness@gmail.com
--- a/helm/exporter-kube-state/templates/clusterrole.yaml
+++ b/helm/exporter-kube-state/templates/clusterrole.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.rbacEnable }}
+{{- if .Values.global.rbacEnable }}
 {{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1beta1" }}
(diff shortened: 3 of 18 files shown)


gianrubio:
@djsd123 I'll need few days to review your PR

gianrubio:
OMG I never realized about this helm behaviour. The PR #829 added the same change but I reject it :( my bad @mbenabda. 
I'm wondering how can we merge this PR without breaking the current arguments, any thougts?

helm/alertmanager/values.yaml
```diff
@@ -142,10 +142,6 @@ service:
   ##
   type: ClusterIP
 
-## If true, create & use RBAC resources
-##
-rbacEnable: true
```
gianrubio: this was already discussed on https://github.com/coreos/prometheus-operator/pull/829#discussion_r160369963

> Maybe a personal preference, but I like to have ServiceAccounts for all applications I deploy, even those that don't get any roles. I do this so I don't accidentally bind anything to a default ServiceAccount and grant an application access it shouldn't have.

I agreed this approach so I'd like to keep the serviceAccount on alertmanager.

djsd123: looking at the master branch I still cannot see where it is being used??  You would need a service account resource and then bind that account to the alertmanager resource?

I can raise a new PR to this and pull my last commit?

djsd123:
Hi @gianrubio,

Not sure I understand "__I'm wondering how can we merge this PR without breaking the current arguments, any thougts?__" .

I did not realise my PR would break something.  I'll gladly amend it if you explain what it breaks?

brancz:
The charts are still pre-alpha, I think even if the changes are breaking but bring us forward, we should break.

gianrubio:
This will not break the charts once the default value for rbac is true and rbacEnable=false is not working.

@djsd123 could you follow the [guide to contribute](https://github.com/coreos/prometheus-operator/tree/master/helm#how-to-contribue), we're missing steps 3 to 7.

djsd123:
Thanks, @gianrubio, all done.  I removed my last commit.  I'll get another PR in to cover the alertmanager chart.

gianrubio:
thanks @djsd123

Issue #919, opened 2018-01-25
Helm grafana with ingress have many 404 on resources


sebglon:
**What did you do?**

I want to acces on my grafana with an ingress.
**What did you expect to see?**
Grafana work well

**What did you see instead? Under which circumstances?**
When i go on my fqdn, index page retunr HTTP 200 but all CSS, JS are on HTTP 404.

**Environment**
kubectl version
Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.6", GitCommit:"6260bb08c46c31eea6cb538b34a9ceb3e406689c", GitTreeState:"clean", BuildDate:"2017-12-21T06:34:11Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"8+", GitVersion:"v1.8.6-gke.0", GitCommit:"ee9a97661f14ee0b1ca31d6edd30480c89347c79", GitTreeState:"clean", BuildDate:"2018-01-05T03:36:42Z", GoVersion:"go1.8.3b4", Compiler:"gc", Platform:"linux/amd64"}

* Chart values.yaml
```
    adminUser: "admin"
    adminPassword: "admin"
#    auth:
#        anonymous:
#            enable: false
    service:
        type: NodePort
    ingress:
        enabled: true
        annotations:
            kubernetes.io/ingress.global-static-ip-name: "grafana-customer"
            kubernetes.io/ingress.class: "gce"
            kubernetes.io/ingress.allow-http: "true"
        fqdn: "dashboard.cloud.akeneo.com"
```

sebglon:
If on ingress manifest i remove `path: /`, it work.
I create a PR to fix this.

PR #920, opened 2018-01-25
remove path on helm grafan ingress

sebglon:
#919


sebglon:
@gianrubio can you check that?

sebglon:
@gianrubio version bunped

gianrubio:
thanks @sebglon

PR #933, opened 2018-01-29
Add targetHeapSizeRate options.

whypro:
**what changed**
Add targetHeapSizeRate options for prometheus CRD.

It would be defined in prometheus CRD
```
targetHeapSizeRate: "0.5"
```
That means:
- Set `-storage.local.target-heap-size` = requests.memory * targetHeapSizeRate
- The valid value is in range (0, 1)
- If omit or empty, the default value is 2/3

**why to do this**
prevent prometheus instances being OOM killed even though the default rate is 2/3 (1.x version)


brancz:
I'd like to avoid an abstraction like this as it is only the case for Prometheus 1.x, in Prometheus 2.x these values are automatically tuned by the kernel, which is a lot more powerful. I highly recommend to just switch to Prometheus 2.0.

PR #938, opened 2018-01-30
Alert Notifications updater for grafana-watcher

Stelminator:
Currently getting built on docker hub as [`stelminator/grafana-watcher:alertnotifications`](https://hub.docker.com/r/stelminator/grafana-watcher/)

### DIFF
--- a/contrib/grafana-watcher/.dockerignore
+++ b/contrib/grafana-watcher/.dockerignore
@@ -1,0 +1,3 @@
+.dockerignore
+Dockerfile
+grafana-watcher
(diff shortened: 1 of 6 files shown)


brancz:
I'm not sure this is a good idea. Alert de-duplication in Grafana is based on writing whether a notification was successful into the database, as the grafana-watcher implies that all Grafana instances are using their own local SQLite database, no such synchronization can happen.

Have you looked at using Prometheus + Alertmanager for alerting? The HA model in general is a lot more thought through than in Grafana.

Stelminator:
@brancz Those are great points.  I agree that Alertmanager is the right long-term HA solution.  I wanted some way to bootstrap some simple triggers in a single grafana instance from scratch, and this did the job.

brancz:
In that case I recommend carrying your patch yourself for the time being. As Grafana is implementing datasources + dashboards from disk in v5.0 (already available in nightlies) we are soon to deprecate and remove the grafana-watcher.

brancz:
Thanks for your efforts though! I just feel this would give users the wrong idea of how Grafana is operated and kept consistent.

PR #952, opened 2018-02-01
Add the ability to set sessionAffinity for prometheus service

jaglade:


helm/prometheus/values.yaml
```diff
@@ -121,6 +121,9 @@ rules:
 secrets: {}
 
 service:
+  ## Maintains session affinity.  Should be set to ClientIP for HA setup
+  ## Only options are ClientIP and None.  Do not leave blank.
+  sessionAffinity: ClientIP
```
gianrubio: I'm afraid to change the default behaviour for ClientIP because we already have other users using None.  WDYT to rollback this value to None?

jaglade:
Changed sessionAffinity default to None.

gianrubio:
Thanks @jaglade

PR #975, opened 2018-02-09
prometheus-operator ClusterRole should specify finalizer subresource

derekwaynecarr:
OwnerReferencesPermissionEnforcement admission plugin is on by default
in OpenShift, and without this change, the operator cannot deploy.

### DIFF
--- a/Documentation/rbac.md
+++ b/Documentation/rbac.md
@@ -37,6 +37,7 @@ rules:
   resources:
   - alertmanagers
   - prometheuses
+  - prometheuses/finalizers
   - servicemonitors
   verbs:
   - "*"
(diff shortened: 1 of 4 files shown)


derekwaynecarr:
this should probably be in all the examples, will look to update shortly.

brancz:
You need to change the file here: https://github.com/coreos/prometheus-operator/blob/master/example/rbac/prometheus-operator/prometheus-operator-cluster-role.yaml. From there the bundle is generated.

derekwaynecarr:
@brancz - thanks for the pointer.  this should be good to go now.

brancz:
lgtm, thanks! :+1:

PR #981, opened 2018-02-12
helm/prometheus: add expected label to templated ServiceMonitors so they are discovered by default

crandles:
Adds the `prometheus` label to templated ServiceMonitors so the default serviceMonitorsSelector (`prometheus: {{ .Release.Name }}`) matches against any defined ServiceMonitors. Expected this to match by default -- I think it used to.

### DIFF
--- a/helm/kube-prometheus/requirements.yaml
+++ b/helm/kube-prometheus/requirements.yaml
@@ -5,7 +5,7 @@ dependencies:
     repository: https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/
 
   - name: prometheus
-    version: 0.0.14
+    version: 0.0.15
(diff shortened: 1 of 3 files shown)


gianrubio:
@crandles I didn't understand why you add this labels, could you explain?

crandles:
The default servicemontorSelector is looking for this label, but the helm chart isn't adding it to the servicemonitors it creates.

https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus/templates/prometheus.yaml#L63-L65

With this added to the template the default selector will apply to the servicemonitors.

gianrubio:
you're absolute right, lgtm! Thanks @crandles

PR #986, opened 2018-02-13
Add missing resource to operator role

ironcladlou:
Add a missing 'prometheuses/finalizers' resource to the operator cluster role.
Without this role, the operator will fail to create a default secret for a
Prometheus instance in the absence of any service monitors.

The fix seems to already be present in the `examples` version of the role; this
patch brings the contrib version in line with the example.


brancz:
Yes I was thinking of this as well yesterday evening, thanks for adapting!

PR #987, opened 2018-02-14
helm: add ability to exclude alertmanager from kube-prometheus

cspargo:
Some of us use an external alertmanager. Add an option to exclude alertmanager. It still includes it by default.


gianrubio:
thanks @cspargo

PR #991, opened 2018-02-15
Cross building images

dmathieu:
This is a quite basic way of building docker images for other platform, enabling ARM as well.


Dockerfile.arm
```diff
@@ -0,0 +1,5 @@
+FROM arm32v6/alpine
```
dmathieu: It doesn't look like there's arm busyboxes.

brancz:
As we're pretty much using all the tooling of the upstream Prometheus project, I would feel most comfortable with having crossplatform images in Prometheus first. Maybe we can tackle it there and then come back to this?

dmathieu:
Sure, makes sense.
I've opened prometheus/prometheus/pull/3847. Happy to take reviews there and then come back here :)

dmathieu:
Let's wait until prometheus supports this

devinrsmith:
It looks like prometheus supports arm now - I think this issues should be re-opened! https://github.com/prometheus/prometheus/pull/5031

thejefflarson:
@brancz Is there anything holding this issue back? I can probably figure it out, but from where I sit @dmathieu's solution looked good.

brancz:
We don't use promu anymore, so this solution won't work anymore.

maxromanovsky:
What can be done to allow building arm images?

brancz:
We just need to get the appropriate automation in place.

maxromanovsky:
@brancz I'll try to look into it during upcoming days. I was able to build it locally "the dirty way", i.e. without proper `Makefile` automation alongside other binaries/containers for `arm/v7` Docker arch to run `prometheus-operator` chart on k3s Raspberry PI cluster. That went well, although it's really painful for RPIs to run these workloads :)

Now I'll do my best to make it work out of the box for official releases of these components

PR #993, opened 2018-02-15
Give operator permission to manager alertmanagers

ironcladlou:
Add a missing resource to the operator's cluster role allowing it to correctly
manager alertmanagers.

### DIFF
--- a/Documentation/rbac.md
+++ b/Documentation/rbac.md
@@ -38,6 +38,7 @@ rules:
   - alertmanagers
   - prometheuses
   - prometheuses/finalizers
+  - alertmanagers/finalizers
   - servicemonitors
   verbs:
   - "*"
--- a/Documentation/user-guides/getting-started.md
+++ b/Documentation/user-guides/getting-started.md
@@ -52,6 +52,7 @@ rules:
   - alertmanagers
   - prometheuses
   - prometheuses/finalizers
+  - alertmanagers/finalizers
   - servicemonitors
   verbs:
   - "*"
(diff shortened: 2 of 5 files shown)


brancz:
Sounds like we need this for `ServiceMonitor` objects as well then? Could you elaborate what the finalizer subresource is necessary for in OpenShift? Then we can make sure all permissions are adapted as they need to.

ironcladlou:
This isn't specific to origin; the k8s admission controller chain contains [garbage collection permissions enforcement](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/gc/gc_admission.go#L87) to ensure that if you try to create an OwnerReference with BlockOwnerDeletion=true that you have permission to delete the owner *and* to mutate finalizers on the owner (to avoid clients from arbitrarily preventing deletion of other resources to which they may not actually have such authorization).

The operator sets BlockOwnerDeletion for owner references to StatefulSets created for Prometheus and AlertManager instances, which makes sense, but implies adherence to the permissions setup I described above.

ServiceMonitor isn't necessary because the operator isn't creating any child resources owned by the ServiceMonitor.

Does this help clarify?

brancz:
Got it, makes sense. Could you adapt it [here](https://github.com/coreos/prometheus-operator/blob/master/example/rbac/prometheus-operator/prometheus-operator-cluster-role.yaml) as well then? And then run `make generate` in the root of the repo. Thanks!

ironcladlou:
@brancz I updated the example, ran the generator, and decided to go ahead and squash it all down. Thanks!

brancz:
lgtm thanks!

PR #1003, opened 2018-02-20
prometheus-config-reloader: propagate errors for refreshing config files

### LINKED ISSUE #1002: prometheus-config-reloader errors may prevent rules from loading
**What did you do?**
  Installed helm chart for Prometheus instance.

  Prometheus instance should have loaded the included rules.

  Prometheus had no data for the associated rules.

* prometheus-config-reloader Logs

```
$ kubectl logs -n *** -f prometheus-***-***-0 prometheus-config-reloader
ts=2018-02-15T16:49:59Z caller=main.go:231 component=volume-watcher msg=Starting...
ts=2018-02-15T16:49:59Z caller=main.go:204 component=volume-watcher msg="Updating rule files..."
ts=2018-02-15T16:50:21Z caller=main.go:207 component=volume-watcher msg="Updating rule files failed." err="Get https://192.168.128.1:443/api/v1/namespaces/***/configmaps/***-***-***-***: unexpected EOF"
ts=2018-02-15T16:50:21Z caller=main.go:212 component=volume-watcher msg="Reloading Prometheus..."
ts=2018-02-15T16:50:21Z caller=main.go:219 component=volume-watcher msg="Prometheus successfully reloaded."
```

It looks like the code [here](https://github.com/coreos/prometheus-operator/blob/master/contrib/prometheus-config-reloader/main.go#L204-L210)
```
	w.logger.Log("msg", "Updating rule files...")
	err := w.UpdateRuleFiles()
	if err != nil {
		w.logger.Log("msg", "Updating rule files failed.", "err", err)
	} else {
		w.logger.Log("msg", "Rule files updated.")
	}
```
could be propagating the errors instead of continuing to trigger a reload of prometheus. After this initial error, the reloader will not attempt to load the files again, so Prometheus will remain without rules until this container is restarted.

crandles:
Resolves #1002 

Errors from `func (w *volumeWatcher) Refresh()` should be propagated. Right now, errors for refreshing the rules are ignored and may result in a Prometheus instance without rules, or without the latest rules loaded.

### DIFF
--- a/contrib/prometheus-config-reloader/main.go
+++ b/contrib/prometheus-config-reloader/main.go
@@ -200,11 +200,14 @@ func (w *volumeWatcher) ReloadPrometheus() error {
 	return nil
 }
 
-func (w *volumeWatcher) Refresh() {
+func (w *volumeWatcher) Refresh() error {
 	w.logger.Log("msg", "Updating rule files...")
-	err := w.UpdateRuleFiles()
+	err := backoff.RetryNotify(w.UpdateRuleFiles, backoff.NewExponentialBackOff(), func(err error, next time.Duration) {
+		w.logger.Log("msg", "Updating rule files temporarily failed.", "err", err, "next-retry", next)
+	})
 	if err != nil {
 		w.logger.Log("msg", "Updating rule files failed.", "err", err)
+		return err
 	} else {
 		w.logger.Log("msg", "Rule files updated.")
 	}
@@ -215,9 +218,11 @@ func (w *volumeWatcher) Refresh() {
 	})
 	if err != nil {
 		w.logger.Log("msg", "Reloading Prometheus failed.", "err", err)
+		return err
 	} else {
 		w.logger.Log("msg", "Prometheus successfully reloaded.")
 	}
+	return nil
 }
 
 func (w *volumeWatcher) Run() {
@@ -229,7 +234,11 @@ func (w *volumeWatcher) Run() {
 	defer watcher.Close()
 
 	w.logger.Log("msg", "Starting...")
-	w.Refresh()
+	err = w.Refresh()
+	if err != nil {
+		w.logger.Log("msg", "Initial loading of config volume failed.", "err", err)
+		os.Exit(1)
+	}
 	err = watcher.Add(w.cfg.configVolumeDir)
 	if err != nil {
 		w.logger.Log("msg", "Adding config volume to be watched failed.", "err", err)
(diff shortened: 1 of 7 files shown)


brancz:
Good catch! Thanks a lot! I'm pushing the image and will then re-trigger the build.

crandles:
Thanks. I don't feel great about the second os.Exit, but didn't see a better way to resolve.

contrib/prometheus-config-reloader/main.go
```diff
@@ -229,7 +234,10 @@ func (w *volumeWatcher) Run() {
 	defer watcher.Close()
 
 	w.logger.Log("msg", "Starting...")
-	w.Refresh()
+	err = w.Refresh()
+	if err != nil {
+		os.Exit(1)
```
brancz: Could you add another log line here that logs the error?

crandles:
Forgot to bump the operator chart version, did that as well.

brancz:
Not sure why the first e2e job is failing so consistently. Could you try running those locally on minikube?

crandles:
I'm getting a PASS from that (and other) tests running locally

brancz:
In that case, could you just try to add a longer timeout for the travis-e2e tests? https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

Maybe `travis_wait 30 ./scripts/travis-e2e.sh` would be enough.

Sorry for the inconvenience.

crandles:
That didn't go well

brancz:
This is frustrating .. could you try rebasing this PR?

crandles:
ok. build in progress again. If this fails, I might try removing the `travis_wait`

crandles:
I am not sure how to get this to pass

brancz:
Me neither, but let me merge this and make it my problem. Thanks for your effort! :+1: :tada:

crandles:
Originally you had to push the image. Did that happen again?

mxey:
This change introduced a new issue. If a ConfigMap has has invalid rules, Prometheus will fail to reload. The reload will be retried due to the new backoff. Since prometheus-config-reload is stuck there, it will not react to a modified ConfigMap anymore. 

I think it would be better to have the backoff around the whole Refresh() method.

crandles:
I would expect the container to exit, and it should load a newer ConfigMap if it is updated. I'm not sure an additional backoff would help here.

 Maybe we should distinguish between the possible errors reloading prometheus could trigger (an invalid config file vs unable to reach prometheus)? If it is a "valid" error, (and we are in the [watcher.Events loop](https://github.com/coreos/prometheus-operator/blob/master/contrib/prometheus-config-reloader/main.go#L254)), log the error but don't exit.

mxey:
The container does exit eventually, yes. But it takes longer than it should for a quick config fix.

brancz:
I'd suggest that we only back off on errors other than when Prometheus errors on reload. If the reload fails, the process should exit.

Issue #1014, opened 2018-02-23
Documentation: Clarify _RelabelConfig_ usage in API documentation


opusmagnum:
**What did you do?**
Tried to relabel some metrics, adding `RelabelConfig` spec to the ServiceMonitor manifest.
**What did you expect to see?**
I have expected, that this spec will be transformed into `<relabel_configs>` section. 
**What did you see instead? Under which circumstances?**
A `<metric_relabel_configs>` section was generated by Prometheus Operator instead.
**Environment**

* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.3", GitCommit:"d2835416544f298c919e2ead3be3d0864b52323b", GitTreeState:"clean", BuildDate:"2018-02-09T21:51:54Z", GoVersion:"go1.9.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.3+coreos.0", GitCommit:"f588569ed1bd4a6c986205dd0d7b04da4ab1a3b6", GitTreeState:"clean", BuildDate:"2018-02-10T01:42:55Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: 

Kubespray

* Manifests:

```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: appds
  namespace: ns1
  labels:
    k8s-app: myapp
spec:
  endpoints:
    - port: http-metrics
      path: /metrics/prometheus
      # honorLabels: true
      interval: 30s
      # targetPort: 0
      metricRelabelings:
        - sourceLabels:	['__meta_kubernetes_pod_node_name']
          separator: ';'
          targetLabel: 'node_name'
          regex: '(.*)'
          replacement: '$1'
          action: replace
  jobLabel: my-app
  namespaceSelector:
    matchNames:
      - test
  selector:
    any: true
```

* Prometheus Operator Logs:

```n/a```

I have created a pull request with a suggested clarification.

opusmagnum:
Do I undestand correctly, that `__meta_` labels are only available for [relabeling](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) but not for [metrics_relabeling](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs)?

[...]

PR #1015, opened 2018-02-23
documentation: clarify _RelabelConfig_ usage

opusmagnum:
The _RelabelConfig_ is confusing and connotes, that it is about `<relabel_config>`-section of the configuration, but in reality it is about `<metric_relabel_configs>`-section.

#1014

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -234,7 +234,7 @@ Most recent observed status of the Prometheus cluster. Read-only. Not included w
 
 ## RelabelConfig
 
-RelabelConfig allows dynamic rewriting of the label set.
+RelabelConfig allows dynamic rewriting of the label set, being applied to samples before ingestion. It defines `<metric_relabel_configs>`-section of Prometheus configuration. More info: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs
 
 | Field | Description | Scheme | Required |
 | ----- | ----------- | ------ | -------- |
--- a/example/prometheus-operator-crd/prometheus.crd.yaml
+++ b/example/prometheus-operator-crd/prometheus.crd.yaml
@@ -1036,8 +1036,10 @@ spec:
                   writeRelabelConfigs:
                     description: The list of remote write relabel configurations.
                     items:
-                      description: RelabelConfig allows dynamic rewriting of the label
-                        set.
+                      description: 'RelabelConfig allows dynamic rewriting of the
+                        label set, being applied to samples before ingestion. It defines
+                        `<metric_relabel_configs>`-section of Prometheus configuration.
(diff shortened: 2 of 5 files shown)


brancz:
Sorry for the confusion! And thanks for taking the time to contribute!

The relabelconfig struct is actually relevant for any relabeling configuration, I would prefer, if we add a comment like this into the actual `metricsRelabelConfig` field, here: https://github.com/coreos/prometheus-operator/blob/master/pkg/client/monitoring/v1/types.go#L312-L313

Also note that the API docs are generated, so we need to change it in the actual type definition as seen above, then you can just regenerate the docs by running `make generate` in the root of the repo. Thanks! :slightly_smiling_face:

opusmagnum:
Thank you @brancz to look into it,
understood with the API-documentation source.

Your comment in the code is terse and clear for everybody, who is deep in Prometheus Configuration/relabeling. But believe me, newbies will be glad to have a link to the appropriate detailed documentation section of Prometheus configuration as well. It is about quicker and easier adoption of Prometheus-operator.

brancz:
Understood and agreed. In that case, could you just adapt the text here: https://github.com/coreos/prometheus-operator/blob/master/pkg/client/monitoring/v1/types.go#L221

And then run the `make generate` target? Thanks!

opusmagnum:
I have added code comment in `types.go` and run `make generate` afterwards. Travis CI build is failing on this feature branch? The reason is unclear to me (Golang beginner ;). 
Thanks in advance.

brancz:
To be honest I don't know where this happens, but I get the same result as CI when I just run `./scripts/check-make-generate.sh`. Could you just run that and satisfy CI? :slightly_smiling_face:

opusmagnum:
There was a space at the end of the line. Travis CI run `./scripts/check-make-generate.sh` script and found it out, because it makes `git diff --exit-code` at the end. But this space in question was not visible with "normal" diff. I found it out with -- `git diff --color --ws-error-highlight=new,old`. 
Fixed and commited.

PS: Travis CI will always fail if `go fmt` changes something relative to already commited code. I see few other pull requests failed Travis CI, may be it is the reason for it, that a particular commit has some deviations from Golang standard formatting?

brancz:
Thanks for adapting! This lgtm now.

> PS: Travis CI will always fail if go fmt changes something relative to already commited code. I see few other pull requests failed Travis CI, may be it is the reason for it, that a particular commit has some deviations from Golang standard formatting?

Can you point me at which PRs you mean? If it's a problem then we need to fix it, but generally we want the build to fail if `go fmt` changes anything.

opusmagnum:
I meant e.g. something like -- https://github.com/coreos/prometheus-operator/pull/950, https://travis-ci.org/coreos/prometheus-operator/jobs/335850230 but see now that the diff is obvious there (it is definitely not an invisible space) and you have already recommended how to fix it. Never mind.

Thank you for support and explanations!

PR #1031, opened 2018-02-28
Added information on webhook token authentication

carlosedp:
Added information on webhook token authentication and information on how to update the manifests if Kubernetes is already deployed with kubeadm.


brancz:
Perfect, thanks a lot!

PR #774, opened 2017-11-21
Update rules

brancz:
Closes #686

@ant31

PR #1032, opened 2018-03-01
Update alert rule for kubelet

atoato88:
Update alert rule check kubelet down ratio from 1% to 10%.
In #774 , it is changed to 1%, so returns to 10%.


brancz:
Perfect, thanks a lot!

PR #1035, opened 2018-03-02
Add summary to Alertmanager rules where missing

alexanderdavidsen:
Added summary to Alertmanager rules where missing. The summary is based off the description label, but contains less information - ideal for headers when sending notifications to Slack or HipChat.


brancz:
Could you add these to the alert definitions in `contrib/kube-prometheus/assets/prometheus/rules` and the run `make generate` from the `contrib/kube-prometheus/` directory? Otherwise lgtm!

gianrubio:
To update the helm charts you should run [this script](https://github.com/coreos/prometheus-operator/blob/master/helm/hack/sync_kube_prometheus.py)  and follow [this steps](https://github.com/coreos/prometheus-operator/tree/master/helm#how-to-contribue)

helm/grafana/templates/grafana-dashboards.yaml
```diff
@@ -1,7486 +1,9 @@
 {{ define "grafana-dashboards.yaml.tpl" }}
-deployment-dashboard.json: |
```
brancz: I think this got deleted by accident.

alexanderdavidsen: I guess one of the scripts removed this line. I executed "make generate" in contrib/kube-prometheus and ran helm/hack/sync_kube_prometheus.yaml. 

Should "make generate" create grafana-dashboard.json?

brancz: I'm assuming something there went wrong, it's probably best to just revert the changes in this file, but ultimately we should fix the scripts. (cc @gianrubio )

alexanderdavidsen: I have reverted grafana-dashboards.yaml to it's previous state and pushed the new commit. Should be building now.

brancz:
lgtm on green

PR #1047, opened 2018-03-06
Add the possibility to mount grafana config

abelgana:
This change enable Grafna config customization.

### DIFF
--- a/helm/grafana/Chart.yaml
+++ b/helm/grafana/Chart.yaml
@@ -8,4 +8,4 @@ maintainers:
 name: grafana
 sources:
 - https://github.com/coreos/prometheus-operator
-version: 0.0.22
+version: 0.0.23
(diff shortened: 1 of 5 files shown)


brancz:
I would expect that there would be a field to specify the "custom" Grafana config.

abelgana:
@brancz Thanks for the feedback. I think it’s easier to just provide a way of mounting the config folder. By default grafana will look at /etc/grafana for the config file (grafana.ini). The idea is to use the same folder for other configuration files like ldap. I am open to work on improving it.

abelgana:
Hi,

The pipeline is green. Is there something I can in order to merge the PR?

gianrubio:
@abelgana could you rebase it?

abelgana:
@gianrubio 
I will do it later today. Thanks

abelgana:
Should be ok now to merge. The pipeline is green.

PR #1051, opened 2018-03-07
*: Allow listening locally and specifying additional containers

brancz:
This will allow one to specify a proxy that is actually the entry point for a Prometheus/Alertmanager Pods UI.

@ant31 @fabxc

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -173,9 +173,14 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*v1beta1.
 
 	amArgs := []string{
 		fmt.Sprintf("-config.file=%s", alertmanagerConfFile),
-		fmt.Sprintf("-web.listen-address=:%d", 9093),
 		fmt.Sprintf("-mesh.listen-address=:%d", 6783),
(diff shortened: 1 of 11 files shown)


brancz:
@gianrubio any idea why the helm hashes are different?

gianrubio:
Hum... it's because your branch it's not rebased to master and the script it's not smart enough :) Do you mind to rebase it or to temporary remove the e2e-helm step?
As soon I have time I'll definitely fix that helm e2e script

brancz:
@gianrubio hmm .. I rebased but it doesn't seem to have fixed the issue.

pkg/alertmanager/statefulset.go
```diff
@@ -200,6 +205,24 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*v1beta1.
 		},
 	}
 
+	var livenessProbe *v1.Probe
+	var readinessProbe *v1.Probe
+	if !a.Spec.ListenLocal {
```
ant31: why removing the probe?

brancz: probes on loopback don't seem to work, so I decided it's the responsibility of the "entry point" container

brancz:
Note there are some problems with tests on master, I am fixing those before I move forward with this PR. I'll comment here once the issues on master are resolved.

PR #1054, opened 2018-03-08
[helm] Fixed persistency for Grafana

vglafirov:
/var/grafana-storage wasn't used and persistence for plugins wasn't working. Fixing it now.


lostick:
the path `/var/grafana-storage` is not necessary wrong, depending on how the helm chart is configured

vglafirov:
@lostick That's true, you could tweak environment variables to point grafana to store stuff in `/var/grafana-storage` folder. But the question here, why? why do you need that if you could mount default directory at very beginning? Another thing is volume mounted anyway even if you don't specify storageSpec it's mounted as emptyDir, I think it's worth to use it for that purpose by default, otherwise it's storing the state within container, which is not good idea in general.

vglafirov:
Abandon it, as repository link got lost for some reason :(

PR #1057, opened 2018-03-08
skip helm version check when chart is kube-prometheus

gianrubio:
@brancz

### DIFF
--- a/helm/hack/sync-repo.sh
+++ b/helm/hack/sync-repo.sh
@@ -26,7 +26,7 @@ do
     else
       echo "File ${tgz} does not exist "
     fi
-  done
+  fi
 done
 
 # sync charts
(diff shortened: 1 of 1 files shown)


helm/hack/sync-repo.sh
```diff
@@ -11,24 +11,28 @@ AWS_REGION=${3:-"eu-west-1"}
...
+        echo "ERROR: Current hash should be the same as the remote hash. Please bump the version of chart {$tgz}."
+        exit 1
+      fi
+    else
+      echo "File ${tgz} does not exist "
     fi
-  else
-    echo "File ${tgz} does not exist "
-  fi
+  done
```
crandles: i think this might need to be `fi`

gianrubio: Indeed

PR #1060, opened 2018-03-09
skip helm version check when chart is kube-prometheus #1057

gianrubio:


brancz:
the normal e2e tests failing is my fault, I'm working on it

PR #1065, opened 2018-03-09
Use wildcard serviceMonitorSelector in example

dobesv:
Specifying a specific `serviceMonitorSelector` is a more advanced topic, it's not needed for "getting started".

See also https://github.com/coreos/prometheus-operator/issues/1053

### DIFF
--- a/Documentation/user-guides/getting-started.md
+++ b/Documentation/user-guides/getting-started.md
@@ -189,8 +189,6 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: example-app
-  labels:
-    team: frontend
 spec:
   selector:
     matchLabels:
@@ -255,8 +253,6 @@ For more information, see the [Prometheus Operator RBAC guide][prom-rbac].
 
 ## Include ServiceMonitors
 
-Finally, a Prometheus object defines the `serviceMonitorSelector` to specify which ServiceMonitors should be included. Above the label `team: frontend` was specified, so that's what the Prometheus object selects by.
-
 [embedmd]:# (../../example/user-guides/getting-started/prometheus.yaml)
 ```yaml
 apiVersion: monitoring.coreos.com/v1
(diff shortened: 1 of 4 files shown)


helm/prometheus/templates/prometheus.yaml
```diff
@@ -60,9 +60,7 @@ spec:
   serviceMonitorSelector:
 {{ toYaml .Values.serviceMonitorsSelector | indent 4 }}
 {{- else }}
-  serviceMonitorSelector:
-    matchLabels:
-      prometheus: {{ .Release.Name }}
+  serviceMonitorSelector: {}
```
brancz: The old labeling is still the better choice here I think, as we don't want these to possibly collide. cc @gianrubio

Documentation/user-guides/getting-started.md
```diff
@@ -276,6 +270,9 @@ spec:
 
 This enables the frontend team to create new ServiceMonitors and Services which allow Prometheus to be dynamically reconfigured.
 
+Note: `serviceMonitorSelector: {}` selects all service monitors.  If you omit this key altogether, it won't auto-detect any
```
brancz: Here the other methods should be mentioned as well (as well as their effect):

* Omitting the `serviceMonitorSelector` or setting it to `null` enabled one to use a [custom configuration](https://github.com/coreos/prometheus-operator/blob/master/Documentation/custom-configuration.md).
* `serviceMonitorSelector: {}` selects all ServiceMonitors in the same namespace as the Prometheus object.
* Specifying a selector selects exactly those ServiceMonitors (in the same namespace as the Prometheus object) with the matching selectors. Find the documentation for selectors
 [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#labelselector-v1-meta) and usage guide [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements).

Something along those lines.

Also all the text goes on a single line, as a paragraph, otherwise rendering these on a website gets weird.

brancz:
Thanks a lot for taking your time for this! Just a few changes then I think this will be good to merge!

PR #1066, opened 2018-03-09
prometheus: fix remote read & write basicAuth support

crandles:
The remote_read & remote_write basicAuth configuration wasn't looking for the associated secrets, and was instead embedding the raw BasicAuth contents into Prometheus' config. 

Refactored the basicAuth secret function & added support to resolve the remote secrets.


brancz:
Thanks a lot! This lgtm :+1:

PR #1069, opened 2018-03-11
docs: Add missing argument in `kubeadm` workaround script.

jessestuart:
### What does this PR do?
One of the `sed` commands for updating the `kubeadm` systemd file (`/etc/systemd/system/kubelet.service.d/10-kubeadm.conf`) was missing the reference to the file itself, which would cause it to hang indefinitely if ran as declared.

I also wrapped this second `sed` in an `if grep ...` check, in order to make the operation idempotent.


brancz:
Thanks! lgtm :+1:

PR #1075, opened 2018-03-12
prometheus: remote_read config - support read_recent and required_matchers; and rename proxy_url in CRD

crandles:
- Adds two new configurations for the RemoteReadSpec (`readRecent` & `requiredMatchers`)
- and renames `proxy_url` to `proxyURL` in the spec to be consistent with the field naming convention
  - I can revert this if it isn't desired, it seemed to be the only instance where the underscore naming was used over camel casing.


brancz:
lgtm, thanks! :+1:

PR #1082, opened 2018-03-14
*: Use data directory defined as volume in Dockerfile

brancz:
This changes Prometheus and Alertmanager Pods to mount the data volume under the same directory as specified as a volume in the respective `Dockerfile`.

@ant31 @fabxc @derekwaynecarr @ironcladlou

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -36,7 +36,7 @@ const (
 	defaultVersion         = "v0.14.0"
 	alertmanagerConfDir    = "/etc/alertmanager/config"
 	alertmanagerConfFile   = alertmanagerConfDir + "/alertmanager.yaml"
-	alertmanagerStorageDir = "/var/alertmanager/data"
+	alertmanagerStorageDir = "/alertmanager"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


ironcladlou:
Thanks... I have an identical patch I was starting to test before my PTO but I got hung up executing the e2e tests against Origin. Did you verify these manually with origin, automatically against kube, or something else? Or are the unit tests high enough confidence for the change?

ant31:
Which Dockerfile,  what's the issue ?

ironcladlou:
@ant31 
https://github.com/prometheus/prometheus/blob/master/Dockerfile#L16

The `VOLUME` directive here will typically cause Docker to try and provision host storage to support the container volume mount, a behavior which is explicitly disallowed in OpenShift (https://github.com/projectatomic/docker-novolume-plugin) for security reasons. So, containers using the image will fail to be created because the host directory can't be created for the `VOLUME`.

Easy fix is for the pod spec using such an image to shadow the `VOLUME` mount path with an explicit Kube volume mount pointing to a directory backed by some other supported volume type, which will override the Docker daemon behavior for the path and allow the volume to work.

Did that explanation make sense?

ant31:
@ironcladlou  thanks for the details

PR #1085, opened 2018-03-14
Improve and simplify code in promcfg

metalmatze:
Just a bunch of random things I found while looking into that file.

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -44,7 +44,7 @@ func stringMapToMapSlice(m map[string]string) yaml.MapSlice {
 	res := yaml.MapSlice{}
 	ks := make([]string, 0)
 
-	for k, _ := range m {
+	for k := range m {
 		ks = append(ks, k)
 	}
 	sort.Strings(ks)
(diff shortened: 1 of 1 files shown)


brancz:
this needs a `make generate`, otherwise lgtm

metalmatze:
Runing `make generate` doesn't change anything for me locally. :confused:

brancz:
huh .. that's curious .. could you try `./scripts/check-make-generate.sh`?

pkg/prometheus/promcfg.go
```diff
@@ -610,7 +608,7 @@ func generateRemoteWriteConfig(version semver.Version, specs []v1.RemoteWriteSpe
 		}
 
 		if spec.WriteRelabelConfigs != nil {
-			relabelings := []yaml.MapSlice{}
+			var relabelings []yaml.MapSlice
```
brancz: Actually looking at this again, I remember why we did this, consistently everywhere. This way we make sure we actually have a slice whereas with `var ...`, we may have nil, which can cause curious situations. I prefer actually initializing the empty slice.

PR #1095, opened 2018-03-16
add ability to set TLSConfig for alertmanager connections

### LINKED ISSUE #1094: no way to set insecureSkipVerify for alertmanager connections
I have the problem that I need to use an external alert manager. I've created an endpoint object and service as mentioned in https://github.com/coreos/prometheus-operator/issues/871#issuecomment-356598118

The problem I'm having here is that my company doesn't sign certs for IP addresses, so I can't push via https to our external alertmanager unless I could can enable `insecureSkipVerify` for alertmanager pushes

I also tried using port 80 (the alertmanager is returning an HTTP 302 redirect to) but prometheus doesn't appear to follow the redirect (and silently fails!)

It would be great if we could surface an option to set `insecure_skip_verify` on the alertmanager's `tls_config`

> level=error ts=2018-03-16T04:04:57.777880847Z caller=notifier.go:473 component=notifier alertmanager=https://X.X.X.X:443/api/v1/alerts count=0 msg="Error sending alert" err="Post https://X.X.X.X:443/api/v1/alerts: x509: cannot validate certificate for X.X.X.X because it doesn't contain any IP SANs"

jpiper:
Fixes #1094

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -65,6 +65,7 @@ AlertmanagerEndpoints defines a selection of a single Endpoints object containin
 | port | Port the Alertmanager API is exposed on. | intstr.IntOrString | true |
 | scheme | Scheme to use when firing alerts. | string | false |
 | pathPrefix | Prefix for the HTTP path alerts are pushed to. | string | false |
+| insecureSkipVerify | Disable target certificate validation. | bool | false |
 
 [Back to TOC](#table-of-contents)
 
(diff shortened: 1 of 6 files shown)


brancz:
As suggested in #1094 I prefer exposing the whole of `TLSConfig`. Thanks for taking care of this!

jpiper:
Done. Let me know if there's anything (code style, etc.) you want changed

brancz:
lgtm, thanks!

PR #1096, opened 2018-03-16
helm/prometheus: prometheus template

### LINKED ISSUE #914: Cannot use matchExpressions syntax with Rules Selector in Prometheus Helm Chart
**What did you do?**
I deployed Prometheus using Helm.

As with `serviceMonitorsSelector` I expected that the values from the Prometheus chart for `rulesSelector` can be written like this:
```
rulesSelector: {
  matchExpressions: [{key: prometheus, operator: In, values: [alertmanager, prometheus-inf]}]
}
```
This syntax only works for `serviceMonitorsSelector`.

The bug is [here](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus/templates/prometheus.yaml), line 68. Comparing it to line 59, the difference is clear. 

I need to specify that matchExpressions syntax works for both. Right now I edit the Prometheus resource in Kubernetes by hand in order to achieve the same results as with `serviceMonitorSelector`. 

If you need more details, please tell me.

MihaiAnei:
allow Rules Selector to use other Selectors besides matchLabels

Fixes #914

### DIFF
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -9,4 +9,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.31
+version: 0.0.32
--- a/helm/kube-prometheus/requirements.yaml
+++ b/helm/kube-prometheus/requirements.yaml
@@ -6,7 +6,7 @@ dependencies:
     condition: deployAlertManager
 
   - name: prometheus
-    version: 0.0.21
+    version: 0.0.22
     #e2e-repository: file://../prometheus
     repository: https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/
 
--- a/helm/prometheus/Chart.yaml
+++ b/helm/prometheus/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 name: prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.21
+version: 0.0.22
--- a/helm/prometheus/templates/prometheus.yaml
+++ b/helm/prometheus/templates/prometheus.yaml
@@ -74,8 +74,7 @@ spec:
 {{- end }}
 {{- if .Values.rulesSelector }}
   ruleSelector:
-    matchLabels:
-{{ toYaml .Values.rulesSelector | indent 6 }}
+{{ toYaml .Values.rulesSelector | indent 4 }}
 {{- else }}
   ruleSelector:
     matchLabels:


MihaiAnei:
Maybe I should've increased the Chart version as well?

gianrubio:
Yes, please follow this [steps](https://github.com/coreos/prometheus-operator/tree/master/helm#how-to-contribue)

gianrubio:
also bump kube-prometheus and it's requirements

MihaiAnei:
I realised I forgot that after the failed travis-ci build, thanks. It should be okay now.

gianrubio:
bumping kube-prometheus version, on green we are ready to merge!!

gianrubio:
thanks @MihaiAnei

andersflemmen:
Does this not break the rules.yaml template? I'm getting errors when using
```
rulesSelector:
  matchLabels:
    role: example-rules
```

MihaiAnei:
@andersflemmen previously you could not use any other selector besides `matchLabels` with `rulesSelector`. Indeed your custom chart values need to be changed as this is a breaking change.
For example, if you need to use a `matchLabels` selector:
```
rulesSelector: {
  matchLabels: [{role: example-rules}]
}
```
or if you need a `matchExpression` (which I use most of the times):
```
rulesSelector: {
  matchExpressions: [{key: prometheus, operator: In, values: [nginx-rules, kube-etcd-rules, any-other-awesome-rules]}]
}
```

andersflemmen:
Hmm, ok, so I have to use json inside my yaml?

andersflemmen:
The error comes from the template file below (rules.yaml), where it seems to expect only single labels.
```
{{- if and .Values.rules.specifiedInValues .Values.rules.value }}
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app: {{ template "prometheus.name" . }}
    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
    prometheus: {{ .Release.Name }}
    release: {{ .Release.Name }}
    {{- if .Values.rulesSelector }}
{{ toYaml .Values.rulesSelector | indent 4 }}
    {{- end }}
  name: prometheus-{{ .Release.Name }}-rules
data:
{{ toYaml .Values.rules.value | indent 2 }}
{{- end }}
```

MihaiAnei:
@andersflemmen Indeed, you are right. I did not see that because I did not have any values in 
```
rules:
  specifiedInValues: true
  value:
```
so using (in custom values yaml)
```
rules:
  specifiedInValues: true
  value:
    groups:
    - name: prometheus.rules
      rules:
      - alert: PrometheusConfigReloadFailed
        expr: prometheus_config_last_reload_successful == 0
        for: 10m
        labels:
          severity: warning
        annotations:
          description: Reloading Prometheus' configuration has failed for {{`{{$labels.namespace}}`}}/{{`{{$labels.pod}}`}}
```
generates
```
# Source: prometheus/templates/rules.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app: prometheus
    chart: prometheus-0.0.24
    heritage: Tiller
    prometheus: prometheus-test
    release: prometheus-test
    matchExpressions:
    - key: prometheus
      operator: In
      values:
      - prometheus-hadm
    
  name: prometheus-prometheus-test-rules
data:
  groups:
  - name: prometheus.rules
    rules:
    - alert: PrometheusConfigReloadFailed
      annotations:
        description: Reloading Prometheus' configuration has failed for {{`{{$labels.namespace}}`}}/{{`{{$labels.pod}}`}}
      expr: prometheus_config_last_reload_successful == 0
      for: 10m
      labels:
        severity: warning
```
which, indeed, is not right. I'll think of a solution and come back with a PR.

PR #1099, opened 2018-03-16
prometheus: Add ability to authenticate with bearer token from file

brancz:
This adds the ability to authenticate against Alertmanager with a bearer token read from file. It allows us to use a ServiceAccount token in a Prometheus Pod to authenticate against Alertmanager with it.

@ant31 @fabxc @derekwaynecarr @ironcladlou

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -497,6 +497,10 @@ func generateAlertmanagerConfig(version semver.Version, am v1.AlertmanagerEndpoi
 		cfg = append(cfg, k8sSDWithNamespaces([]string{am.Namespace}))
 	}
 
+	if am.BearerTokenFile != "" {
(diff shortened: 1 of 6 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -488,6 +488,10 @@ func generateAlertmanagerConfig(version semver.Version, am v1.AlertmanagerEndpoi
 		cfg = append(cfg, k8sSDWithNamespaces([]string{am.Namespace}))
 	}
 
+	if am.BearerTokenFile != "" {
+		cfg = append(cfg, yaml.MapItem{Key: "bearer_token_file", Value: am.BearerTokenFile})
```
ant31: is this configuration key already exist in the Alertmanager ?

brancz: Yep see [alertmanager_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<alertmanager_config>)

PR #1100, opened 2018-03-16
Add StaticConfigs to ServiceMonitors

### LINKED ISSUE #1086: Add static_configs to ServiceMonitor
**What did you do?**
I'd like to use the combination of ServiceMonitors and static configuration for Federation.

**Use Case**: I want to run a Prometheus that scrapes a bunch of exporters and other apps via ServiceMonitors, but at the same time scrape other Prometheus via Federation with an URL like this: https://prometheus.example.com/federate

I expected there to be a static_configs key on the ServiceMonitor's endpoint:
https://prometheus.io/docs/prometheus/latest/federation/#configuring-federation
https://github.com/coreos/prometheus-operator/blob/master/pkg/client/monitoring/v1/types.go#L302-L328

or

I'd like be able to merge a static config with the generated configuration from the operator.

There's currently no way to use Federation for static endpoints.

metalmatze:
I've started working on integrating StaticConfigs into the ServiceMonitors.
Somehow they still don't get appended to the generated configuration.

I also want to discuss the behavior of ServiceMonitors with StaticConfigs:
It seems like we need to disable the kubernetes_sd_config for ServiceMonitors that have a StaticConfig, I think.
Otherwise we need to do a hack in the ServiceMonitors along the lines of:
```yaml
kind: ServiceMonitor
metadata:
  name: prometheuses
  namespace: monitoring
  labels:
    team: example
spec:
  selector:
    matchLabels:
      name: nothing
  namespaceSelector:
    matchNames:
    - nothing
  endpoints:
  - interval: 1m
    path: /federate
    params:
      'match[]':
        - '{__name__="up"}'
        - '{__name__=~"machine_controller.*"}'
    staticConfigs:
      targets:
        - 'source-prometheus-1:9090'
        - 'source-prometheus-2:9090'
        - 'source-prometheus-3:9090'
        - 'https://prometheus.example.com'
```
> Note `nothing` as namespace and `nothing` as service labels.

Thoughts?

Closes #1086


Documentation/api.md
```diff
@@ -148,6 +149,7 @@ Endpoint defines a scrapeable endpoint serving Prometheus metrics.
 | honorLabels | HonorLabels chooses the metric's labels on collisions with target labels. | bool | false |
 | basicAuth | BasicAuth allow an endpoint to authenticate over basic authentication More info: https://prometheus.io/docs/operating/configuration/#endpoints | *[BasicAuth](#basicauth) | false |
 | metricRelabelings | MetricRelabelConfigs to apply to samples before ingestion. | []*[RelabelConfig](#relabelconfig) | false |
+| staticConfigs | StaticConfigs with targets to scrape. | *[StaticConfig](#staticconfig) | false |
```
brancz: it's probably fine to just have `staticTargets` field here

pkg/prometheus/promcfg_test.go
```diff
@@ -248,5 +248,37 @@ func makeServiceMonitors() map[string]*monitoringv1.ServiceMonitor {
...
+	res["servicemonitor5"] = &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "testservicemonitor5",
+			Namespace: "default",
+			Labels: map[string]string{
+				"group": "group4",
+			},
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Selector: metav1.LabelSelector{
```
brancz: we should make this field optional if it's not already, it doesn't make sense that we have to specify this when static targets are given

mikeln:
What is the status of this PR?  Can someone push this through?

metalmatze:
Yep. I wanted to work on it today. Looking into it in a few minutes! :+1:

metalmatze:
I've updated the PR to simply include `StaticTargets`. Additionally the LabelSelector is now gone from the ServiceMonitor in the test.

brancz:
lgtm thanks!

PR #1101, opened 2018-03-16
Create statefulset reconcile errors counter

### LINKED ISSUE #1097: Add statefulset_errors_total metric with alert
**What did you do?**
Run the Operator and probably update the Prometheus CRD.

The StatefulSet should be updated.

The StatefulSet was not updated. I could see this in the logs:

```
E0316 10:50:53.825175       5 operator.go:577] Sync "monitoring/prometheus" failed: updating statefulset failed: StatefulSet.apps "prometheus-prometheus" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden.
```

I'd like to have a metric for this, so we can alert on the `sum(rate(prometheus_operator_statefulset_errors_total[5m]))`. I sometimes don't see that these errors happen in the background and thus think a metric with alert would be perfect here.

metalmatze:
We should probably create some alerts for these two counters. But I just quickly added those two.
What do you think?

Closes #1097

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -60,6 +60,8 @@ type Operator struct {
 
 	queue workqueue.RateLimitingInterface
 
+	statefulsetErrors prometheus.Counter
+
 	config Config
 }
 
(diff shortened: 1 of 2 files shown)


pkg/alertmanager/operator.go
```diff
@@ -143,7 +145,17 @@ func New(c prometheusoperator.Config, logger log.Logger) (*Operator, error) {
 }
 
 func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
-	r.MustRegister(NewAlertmanagerCollector(c.alrtInf.GetStore()))
+	c.statefulsetErrors = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "prometheus_operator",
```
brancz: namespace and subsystem are subject to removal in the client lib, I'd suggest to not start using it

metalmatze:
Oh. I didn't hear about that discussion yet. I have removed `Namespace` and `Subsystem`.

brancz:
lgtm, thanks! As a side note, I think it would be good if we extracted the prometheus metrics into two endpoints 1 for prometheus-operator internal metrics and 1 for prometheus/alertmanager object metrics.

PR #1103, opened 2018-03-18
e2e: ensure C locale when generating random string for namespace

jeffersongirao:
LC_ALL when set takes precedence over LC_CTYPE and might cause
the tr command to fail with 'illegal byte sequence' when
generating the random string.


brancz:
lgtm, thanks!

PR #1105, opened 2018-03-19
Ability to add values to grafana ingress

bradenwright:
This already exists for Prometheus and Alertmanager (under helm).  I also need ability to sign labels for Grafana ingress as we are moving to using traefik ingress.


bradenwright:
@gianrubio my bad, should be all good now.  Thanks!

gianrubio:
thanks @bradenwright

PR #1107, opened 2018-03-19
contrib/kube: Restrict example-app servicemonitor to default ns

mxinden:
The `prometheus-frontend` role of the example app kubeprometheus section
is scoped to the default namespace. Thereby the frontend Prometheus
instance is not able to discover anything outside of the default
namespace. We might as well restrict the front end service monitor to
the default namespace too.

### DIFF
--- a/contrib/kube-prometheus/manifests/examples/example-app/servicemonitor-frontend.yaml
+++ b/contrib/kube-prometheus/manifests/examples/example-app/servicemonitor-frontend.yaml
@@ -11,3 +11,6 @@ spec:
   endpoints:
   - port: web
     interval: 10s
+  namespaceSelector:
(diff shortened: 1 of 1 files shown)


brancz:
Wouldn't this be the same if it's deployed in the default namespace? I feel like we should just set the namespace for all of these objects just to make sure the user actually deploys it in the default namespace. It's an example after all, it is just supposed to serve for demonstration purposes, if people want to use it further, they can investigate. Maybe we should also be more clear in the kube-prometheus docs about this.

mxinden:
Setting the namespace to `default` for all objects is done in https://github.com/coreos/prometheus-operator/pull/1111. 

Any other concerns @brancz?

PR #1112, opened 2018-03-19
Helm: change defaults for secrets from empty map to empty list

Nonymus:
Change default for `secrets` in prometheus-chart from empty map to empty list. Avoids helm complaining about being unable to merge a map and a table.


gianrubio:
Thanks @nonymus

PR #1122, opened 2018-03-20
Support Alertmanager v0.15.0

mxinden:
With Alertmanager v0.15.0 the HA command line flag prefix changes from
"mesh" to "cluster" and peer ports need to be specified. This patch
enables running Alertmanager v0.15.0 with the Prometheus operator
without breaking backward compatibility.

In addition it does the following refactoring:

- Instead of generating the stateful set for an old AM version, and then
addressing all necessary changes up to the newest version, generate an
up to date stateful set version and address any changes necessary for
backward compatibility.

- Add unit tests for the various flag changes in between Alertmanager
version.

- Add v0.15.0 mesh initialization test


mxinden:
@fabxc @brancz Could one of you give this a review?

brancz:
lgtm :+1: nice work!

PR #1123, opened 2018-03-20
Addon resizer memory config hooks

olvesh:
We have been having kube-state crashing on us constantly because of
memory limits. In our cluser with 10 nodes this setting worked, well.

Anyway, you can now tweak the values to what you need.


gianrubio:
thanks @olvesh

PR #1125, opened 2018-03-21
tests: Rename govendor-ensure.sh to golang-dep-ensure.sh

mxinden:
Since we are not using govendor anymore, we might as well rename the
script file.


brancz:
lgtm on green

PR #1126, opened 2018-03-21
helm/exporter-node : add args and volumes configurability to the node-exporter

mbenabda:
i need more configurability at the exporter-node level to enable collectors and ignore filesystem mountpoints.

this PR is not a breaking change, and allows one to add args and volumes to the node-exporter container's spec 

a way of using it would be to add something like the following yaml under the ```exporter-node``` key of ```kube-prometheus```'s values.yaml file
```yaml
extraVolumes:
- name: rootfs
  hostPath:
    path: /

extraVolumeMounts:
- mountPath: /rootfs
  name: rootfs
  readOnly: true

extraArgs: 
- --collector.filesystem.ignored-mount-points="^/(sys|proc|host|dev)($|/)"
```


helm/exporter-node/templates/daemonset.yaml
```diff
@@ -21,14 +21,17 @@ spec:
...
             - --collector.procfs=/host/proc
             - --collector.sysfs=/host/sys
-          {{ else }}
+          {{- else }}
             - --path.procfs=/host/proc
             - --path.sysfs=/host/sys
-          {{ end }}
+          {{- end }}
             - --web.listen-address=0.0.0.0:{{ .Values.service.containerPort }}
+          {{- if .Values.extraVolumeMounts }}
```
gianrubio: should be extraArgs

helm/exporter-node/templates/daemonset.yaml
```diff
@@ -21,14 +21,17 @@ spec:
...
+{{ toYaml .Values.extraArgs | indent 12 }}
```
gianrubio: rather than have extraArgs could you rename this to `.Values.container.args` and set the default values to 
``` 
- --collector.procfs=/host/proc
- --collector.sysfs=/host/sys
```

helm/exporter-node/templates/daemonset.yaml
mbenabda: oops !

helm/exporter-node/templates/daemonset.yaml
mbenabda: ok. it would make sense to do the same with volumes and mounts then

gianrubio: Indeed!

helm/exporter-node/templates/daemonset.yaml
```diff
@@ -21,37 +21,33 @@ spec:
...
-          {{ if lt .Values.image.tag "v0.15.0" }}
-            - --collector.procfs=/host/proc
-            - --collector.sysfs=/host/sys
-          {{ else }}
-            - --path.procfs=/host/proc
-            - --path.sysfs=/host/sys
-          {{ end }}
-            - --web.listen-address=0.0.0.0:{{ .Values.service.containerPort }}
+          - --web.listen-address=0.0.0.0:{{ .Values.service.containerPort }}
+          {{- if .Values.container }}
```
gianrubio: Please use one line if https://stackoverflow.com/a/43026557

mbenabda: the issue with the inline if is that both the operands of the and seem to be evaluated. 
when i set the ```container``` key to ```null``` (let's say someone wants the most basic exporter config possible), i get ```Error: render error executing exporter-node/templates/daemonset.yaml:25:46 at <.Values.container.ar...>: can't evaluate field args in type interface {}```

gianrubio: What about we assume Values.container.args should not be null?

mbenabda: ok, the following would work to disable the default args:
```
container: 
  args: []
```

item for volumes & mounts

gianrubio: There’s no document for now, I’ll write docs as soon I move this chart to the K8s charts upstream

mbenabda:
you're welcome @gianrubio . thanks for merging it !

Issue #695, opened 2017-10-16
Propagate service monitored label


sebglon:
**What did you do?**
my service have labels pfid.
My servicemonitor use label pfid on selector
But those label is not propagate on prometheus metrics

**What did you expect to see?**
i want to have access on all or select of labels available on service to prometheus metrics

brancz:
While this is generally possible, you can also join all labels of a service onto metrics using the metrics available in kube-state-metrics

https://github.com/kubernetes/kube-state-metrics/blob/master/Documentation/service-metrics.md#service-metrics

Does that work for you?

sebglon:
Sorry.
when i am not clear.
I use mysql_exporter on a pod with label pfid
on my service monitor i have matchExpressions
```
spec:
  jobLabel: k8s-app
  selector:
    matchExpressions:
      - { key: pfid, operator: Exists}
```
But when metrics show metric `mysql_exporter_scrapes_total` on prometheus, fid` label used on selector is not available on metrics.
I have only basic label
`
mysql_exporter_scrapes_total{endpoint="mysqld-exporter",instance="10.0.2.12:9104",job="mysqld-exporter",namespace="srnt-sgl1",pod="mysql-2760336166-hg3rf",service="mysqld-exporter"}
`

My genrated config: 
```
- job_name: monitoring/mysqld-exporter/0
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  kubernetes_sd_configs:
  - api_server: null
    role: endpoints
    namespaces:
      names: []
  relabel_configs:
  - source_labels: [__meta_kubernetes_service_label_pfid]
    separator: ;
    regex: .+
    replacement: $1
    action: keep
```

brancz:
Yes, the excerpt of the Prometheus config only says that a target with that meta label will be kept, not that it will have that target after relabeling. Relabeling is exactly the kind of black magic we want to keep out of the Prometheus Operator in terms of configurability.

What you are seeing is the mix of Prometheus labels and Kubernetes labels. While they can be used in a compatible manner, they are not the same.

sebglon:
Yes i'm right with you. but how can i make to configure or request with Kubernetes labels? i have not found information or example.

brancz:
If you have kube-state-metrics running and are scraping its metrics, then you can join the `kube_service_labels` metric onto the `mysql_exporter_scrapes_total` metric.

```
[...]

PR #1129, opened 2018-03-21
Grouped queries to some alerts to get hold of pod/deployment labels

olvesh:
As promised in https://github.com/coreos/prometheus-operator/issues/695#issuecomment-374511921 I have made an attempt to get some deployment and pod labels into the alerts.

Maybe there can be done some more, if so please suggest some changes, and I'll add them. #

### DIFF
--- a/helm/exporter-kube-state/templates/kube-state-metrics.rules.yaml
+++ b/helm/exporter-kube-state/templates/kube-state-metrics.rules.yaml
@@ -3,7 +3,8 @@ groups:
 - name: kube-state-metrics.rules
   rules:
   - alert: DeploymentGenerationMismatch
-    expr: kube_deployment_status_observed_generation != kube_deployment_metadata_generation
+    expr: (kube_deployment_status_observed_generation != kube_deployment_metadata_generation) /
+          on(deployment) group_right() kube_deployment_labels
     for: 15m
     labels:
       severity: warning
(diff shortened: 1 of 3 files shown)


helm/exporter-kube-state/templates/kube-state-metrics.rules.yaml
```diff
@@ -3,7 +3,8 @@ groups:
 - name: kube-state-metrics.rules
   rules:
   - alert: DeploymentGenerationMismatch
-    expr: kube_deployment_status_observed_generation != kube_deployment_metadata_generation
```
gianrubio: Rather than changing this file make the changes [here](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus/assets/prometheus/rules) and run `helm/hack/sync_kube_prometheus.py`

gianrubio: @brancz despite the change in the helm template, are you ok with the changes in kube-prometheus alerts?

brancz: I'm struggling to understand what this gets us at alert time, the alert already says the namespace and deployment this belongs to, how do the additional labels help?

olvesh: An example from our inhouse use is to add a `team` label to the deployments and pods. This is used to route the alert to the correct group in PagerDuty.
This part could also have been done with the use of ServiceMonitor as you (@brancz) suggested in https://github.com/coreos/prometheus-operator/issues/695#issuecomment-339652103. That is not implemented yet I take?

Also labels like `tier`, and other infrastructural mapping labels can be used in alerts to inhibit and group alerts.  

I would love having a simpler way of doing this, and having it upstream until a better way is all I suggest in this PR.

brancz: I think what you're trying to achieve is absolutely in line with what we want to enable with the Prometheus Operator, though I'd prefer if we just get https://github.com/coreos/prometheus-operator/issues/695#issuecomment-339652103 down :slightly_smiling_face: .

olvesh: Alright. I'll let this PR lie around for doc purposes if someone wants to achieve the same while waiting for #695

brancz: I opened #1145 to get this done :slightly_smiling_face:

PR #1135, opened 2018-03-22
Enforce external labels, add Thanos example manifests

fabxc:


pkg/prometheus/statefulset.go
```diff
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strings"
 
+	"k8s.io/api/apps/v1beta1"
```
brancz: with the recently merged v1.9 update this now needs to be v1beta2

pkg/prometheus/statefulset.go
```diff
@@ -567,7 +567,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMaps []
 		},
 		Template: v1.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
-				Labels:      finalLabels,
+				Labels:      c.Labels.Merge(podLabels),
```
brancz: doesn't make a difference, but for consistency this should be `finalLabels`

fabxc:
This ist just a time sink right now. We can look into this again when we are closer to actual demand.

PR #1140, opened 2018-03-23
alertmanager: Allows specifying additional secrets

brancz:
For sidecar proxies it may be desirable to add additional secrets to be mounted so they can be consumed by the sidecar.

@ant31 @fabxc @mxinden

### DIFF
--- a/pkg/alertmanager/statefulset_test.go
+++ b/pkg/alertmanager/statefulset_test.go
@@ -351,6 +351,49 @@ func TestMakeStatefulSetSpecPeerFlagPort(t *testing.T) {
 	}
 }
 
+func TestAdditionalSecretsMounted(t *testing.T) {
+	secrets := []string{"secret1", "secret2"}
+	sset, err := makeStatefulSet(&monitoringv1.Alertmanager{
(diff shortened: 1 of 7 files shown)


pkg/alertmanager/statefulset_test.go
```diff
@@ -351,6 +351,45 @@ func TestMakeStatefulSetSpecPeerFlagPort(t *testing.T) {
...
+	secret1Found = false
+	secret2Found = false
+	for _, v := range sset.Spec.Template.Spec.Containers[0].VolumeMounts {
+		if v.Name == "secret-secret1" && v.MountPath == "/etc/alertmanager/secrets/secret1" {
+			secret1Found = true
+		}
+		if v.Name == "secret-secret2" && v.MountPath == "/etc/alertmanager/secrets/secret2" {
+			secret2Found = true
+		}
+	}
```
mxinden: You are missing the second check here, no?

```
+	if !(secret1Found && secret2Found) {
+		t.Fatal("Additional secrets were not found.")
+	}
```

brancz: Absolutely correct! Nice catch!

PR #1141, opened 2018-03-23
Update To Grafana 5.0.3

michaelpietzsch:


brancz:
could you update all the places where it is used as well?

michaelpietzsch:
Is the upload to Quay automated? ... because sanity checks will fail if the image is not uploaded

brancz:
Currently it is not automated, we're hoping to get our needs into the upstream image eventually. I will push it and re-build, that way it will work :slightly_smiling_face:

brancz:
Grafana has since been upgraded to v5.1.x

PR #1147, opened 2018-03-23
Helm: Add the ability to specify custom configuration files for Grafana

marc-sensenich:
This allows for users to mount arbitrary files via a Secrets file to be able to configure such things as LDAP without being bound to a host for Grafana deployments


helm/grafana/templates/grafana-deployment.yaml
```diff
@@ -43,10 +44,14 @@ spec:
...
           mountPath: /var/lib/grafana
-      {{- if .Values.mountGrafanaConfig }}
+        {{- if .Values.mountGrafanaConfig }}
         - name: grafana-config
           mountPath: /etc/grafana
-      {{- end }}
+        {{- end }}
+        {{- if $useCustomConfig }}
+        - name: grafana-config
+          mountPath: /etc/grafana/conf
```
marc-sensenich: Put this one directory higher so that it does not clash with the existing configuration, and changes to the grafana.ini can be done using the extraEnv by defining `GF_...`

marc-sensenich:
@gianrubio Is there anything that could be done from my end to move this forward? Are there any issues with this approach?

gianrubio:
@marc-sensenich since we are moving the helm charts to the kubernetes upstream I'll ask you to wait until we finish the job. You can track the status of this task on #1154. Thanks for understanding

gianrubio:
Feel free to push this changes as soon we finish the migration

richerve:
@marc-sensenich this is something that I need too because we use LDAP.

Upstream on stable/grafana the recently merged https://github.com/kubernetes/charts/pull/4713 that adds support for custom config but it's not as flexible as your change. Do you mind trying to push your change to https://github.com/kubernetes/charts/tree/master/stable/grafana. I think it should be easy in the current state of that Chart.

marc-sensenich:
@richerve I can look into getting this into the official charts repo, are you going to be doing the work to convert the operator Helm deployment to utilize the official Grafana chart?

richerve:
@marc-sensenich yes, I'm working on that, in any case this change will not affect the migration but I think it's very useful so it's better if it can land upstream.

marc-sensenich:
I opened https://github.com/kubernetes/charts/pull/5136 for specifying `ldap.toml`, given it can be defined as structured data I took that approach instead of the open ended approach from here

PR #1168, opened 2018-04-02
Updated README for removing CRDs generated by v1 api.

rahulwa:
When I started working on this project I figured that a few steps were missing.  I have added them in the PR.


brancz:
I don't quite understand why this is necessary, can you explain?

rahulwa:
Currently `prometheus-operator` generates `alertmanagers.monitoring.coreos.com`, `prometheuses.monitoring.coreos.com ` and `servicemonitors.monitoring.coreos.com ` CRDs but [Removal section on README](https://github.com/coreos/prometheus-operator#removal) is missing the instruction for removing them.

README is the first place to look for while getting started with any project. README should be correct and complete for any instruction that it mentions.

brancz:
If the new entries work, then I believe we can remove the old ones, as each of these should only exist once under one name, so it seems odd that we need to remove them twice

rahulwa:
Agreed. I have removed old entries.

brancz:
Thansk! lgtm :+1:

PR #1170, opened 2018-04-03
update  <Monitoring external etcd> docs 

lth2015:
Update the part of <Monitoring external etcd> docs: tlsConfig


brancz:
The debugging suggestion you gave on the other thread is probably valuable as well, could you add that? Otherwise lgtm

mxinden:
@brancz Do you still want to get this into `master`?

brancz:
Yeah I think it's still worth it. (although we need to also add jsonnet instructions to those docs)

PR #1171, opened 2018-04-03
prometheus: Make relabeling fields optional

### LINKED ISSUE #1167: Bug: sourceLabels should not be required in RelabelConfig
Prometheus Operator v0.17.0, Prometheus v2.2.1.

## Problem
For now, if you try to use `action: labeldrop`, you will get the following error in Prometheus:
```
labeldrop action requires only 'regex', and no other fields"
```

## Service Monitor Example
```
kind: ServiceMonitor
metadata:
  name: node-exporter
  labels:    prometheus: main
spec:
  jobLabel: prometheus-target
  selector:
    matchLabels:
      prometheus-target: node-exporter
  namespaceSelector:
    matchNames:
    - kube-prometheus
  endpoints:
  - port: https
    scheme: https
    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    tlsConfig:
      insecureSkipVerify: true
    metricRelabelings:
    - action: labeldrop
      regex: endpoint
      sourceLabels: [] # <- ugly hack, but whithout it operator ignores whole relabeling
```

## Extra info
* `replacement`, that is also required according documention, should also not be required
* According to prometheus docs all fields are optional.

brancz:
Closes #1167 

@ant31 @fabxc @mxinden

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -245,12 +245,12 @@ RelabelConfig allows dynamic rewriting of the label set, being applied to sample
 
 | Field | Description | Scheme | Required |
 | ----- | ----------- | ------ | -------- |
-| sourceLabels | The source labels select values from existing labels. Their content is concatenated using the configured separator and matched against the configured regular expression for the replace, keep, and drop actions. | []string | true |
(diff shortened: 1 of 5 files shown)


brancz:
@distol good catch, do you want to create a PR with an if condition around that and only create the source_labels field if it's not nil?

distol:
Yes, sure.

PR #1172, opened 2018-04-03
prometheus: Make relabeling field SourceLabel optional

distol:
Fix for #1171 (and #1167).


distol:
@brancz, something like this?

brancz:
yep, lgtm :+1: thanks!

Issue #1174, opened 2018-04-03
helm charts: Add CoreDNS support (feature)


ephur:
**What did you do?**
Deployed kube-prometheus via helm charts

**What did you expect to see?**
Monitoring for my DNS service

**What did you see instead? Under which circumstances?**
No monitoring and a service monitor for a DNS service which I do not use

**Environment**

* Kubernetes version information:
kubectl: 1.9.4
kubernetes: 1.10.0

* Kubernetes cluster kind: 
Custom built

* Manifests:
N/A 

* Prometheus Operator Logs:
N/A

ephur:
Please note, I added a relevant PR: (https://github.com/coreos/prometheus-operator/pull/1175)[https://github.com/coreos/prometheus-operator/pull/1175]

PR #1176, opened 2018-04-03
charts: Add Core DNS Support

ephur:
With CoreDNS becoming a first class citizen it should be an optional part
of the kube-prometheus deployment.

Fixes: #1174

Note: I don't know how this fits into the existing testing pipeline since it includes CoreDNS which may not be part of the e2e testing environment


brancz:
Could you also put a sample service and servicemonitor into the contrib/kube-prometheus/manifest/prometheus-k8s Directory?

brancz:
lgtm from bare manifests side... leaving he charts call up to @gianrubio

gianrubio:
Nice improvement @ephur, thanks!

PR #1187, opened 2018-04-06
Added Pod Security Policies to the HELM Charts

FaKod:
I added a default set of rather restrictive Pod Security Policies (PSP) to the charts.
The Alert Manager is created without a SA, so I had to add a PSP to the default SA.

This change is required once you enable the Admission Controller PodSecurityPolicy.
However, this is a draft to make it work. It should be definitely more configurable.

### DIFF
--- a/helm/alertmanager/values.yaml
+++ b/helm/alertmanager/values.yaml
@@ -175,14 +175,14 @@ global:
 ## Ref: https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/storage.md
 ##
 storageSpec: {}
-  #  volumeClaimTemplate:
-  #    spec:
+#  volumeClaimTemplate:
+#    spec:
 #      storageClassName: gluster
-  #      accessModes: ["ReadWriteOnce"]
-  #      resources:
-  #        requests:
-  #          storage: 50Gi
-  #    selector: {}
+#      accessModes: ["ReadWriteOnce"]
+#      resources:
+#        requests:
+#          storage: 50Gi
+#    selector: {}
 
 # default rules are in templates/alertmanager.rules.yaml
 # prometheusRules: {}
--- a/helm/alertmanager/values.yaml
+++ b/helm/alertmanager/values.yaml
@@ -165,22 +165,24 @@ service:
   ##
   type: ClusterIP
 
-## If true, create & use RBAC resources
+## If true, create & use RBAC resources resp. Pod Security Policies
 ##
-rbacEnable: true
+global:
+  rbacEnable: true
+  pspEnable: true
 
 ## Alertmanager StorageSpec for persistent data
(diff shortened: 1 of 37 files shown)


gianrubio:
@FaKod CI is failling because you forget to bump charts, please read this [guide](https://github.com/coreos/prometheus-operator/tree/master/helm#how-to-contribue)

helm/alertmanager/values.yaml
```diff
@@ -153,22 +153,24 @@ service:
   ##
   type: ClusterIP
 
-## If true, create & use RBAC resources
+## If true, create & use RBAC resources resp. Pod Security Policies
 ##
-rbacEnable: true
+global:
+  rbacEnable: true
```
gianrubio: nice catch

helm/alertmanager/values.yaml
```diff
@@ -153,22 +153,24 @@ service:
...
+  pspEnable: true
 
 ## Alertmanager StorageSpec for persistent data
 ## Ref: https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/storage.md
 ##
 storageSpec: {}
-#  volumeClaimTemplate:
-#    spec:
+  #  volumeClaimTemplate:
```
gianrubio: please rollback the indentation

helm/exporter-node/templates/psp-clusterrole.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.global.pspEnable }}
```
gianrubio: Rbac should be required to create ClusterRole and ClusterRoleBinding please add the line below in all resource added by you

``` {{- if .Values.global.rbacEnable }} ```

FaKod:
@brancz If you enable the PSP admission controller no POD gets created. So at least one PSP for SA default (for the alertmanager) and all the other SAs is needed. If we do not want to have chart dependencies, this is an acceptable way to do it, imho.

brancz:
That makes sense. I wasn't advocating for not having a PSP here, I was saying we should only need 2, 1 for the node-exporter and one for the rest of the components.

FaKod:
@brancz "...one for the rest of the components..." This are all different HELm charts, so you mean one of the same name in all charts (except node-exporter)? Or how do you handle the dependencies?

brancz:
Ah gotcha, sorry you're right, as these are separate charts, there's not really the possibility for a shared PSP. Hmm a shame, but I don't see a better way then this.

gianrubio:
Sorry @FaKod to not merge your PR, I still need few more day to to test this charts

mrueg:
Any chance to get this in soon? Would be super helpful on my deployment!

gianrubio:
I rebased the branch, I'll merge on green

gianrubio:
thanks @FaKod

PR #1193, opened 2018-04-08
alertmanager: Pass pod ip to --cluster.listen-address

mxinden:
Alertmanager needs a routable IP address via the
`--cluster.listen-address` in order for other nodes in an Alertmanager
cluster to be able to reach it. This patch sets the
`--cluster.listen-address` to the k8s pod ip + `6783` port.

Fixes https://github.com/prometheus/alertmanager/issues/1312

//CC @gmauleon


brancz:
lgtm :+1: nice catch

PR #1194, opened 2018-04-09
Fix StaticTargets generation and add unit test

metalmatze:
The generation of StaticTargets was broken.
I've fixed that.

Sadly the unit tests right now doesn't seem to compile the correct YAML.

@brancz @mxinden


brancz:
Can you mark the `staticTargets` field as experimental in the godoc comment? It feels a bit fishy but I haven't thought of a better way yet, I just want to keep the right to modify it should we want to.

metalmatze:
Yep. That's probably better. I've copied the experimental comment from other fields.

PR #1196, opened 2018-04-09
alertmanager: Test if Alertmanagers gossip silences

mxinden:
This patch adds an e2e test. It creates a silence on alertmanager 0, and
checks if that silence is gossiped to alertmanager 1 and 2.

This should prevent regressions like https://github.com/prometheus/alertmanager/issues/1312


brancz:
lgtm :+1:

PR #1205, opened 2018-04-10
Fixed CPU accounting in recording rules

zuzzas:
Currently, node recording rules features incorrect idle CPU accounting. This change aims to fix that.

Also removed negative regex match, since guest CPU time (albeit, rarely present on Kubernetes nodes) contributes to the overall non-idle CPU time.


brancz:
lgtm

PR #1206, opened 2018-04-10
Convert kube-prometheus to jsonnet

brancz:
This PR implements the kube-prometheus stack as jsonnet code.

@ant31 @fabxc @mxinden 

/cc @metalmatze


Documentation/troubleshooting.md
```diff
@@ -34,3 +34,17 @@ Account: [myname@example.org]
...
+### Troubleshooting ServiceMonitor changes
+
+When creating/deleting/modifying `ServiceMonitor` objects it is sometimes not as obvious what piece is not working properly. This section gives a step by step guide how to troubleshoot such actions on a `ServiceMonitor` object.
+
+#### Has my `ServiceMonitor` been picked up by Prometheus?
+
+`ServiceMonitor` objects are selected by the `serviceMonitorSelector` of a Prometheus object. The name of a `ServiceMonitor` is encoded in the Prometheus configuration, so you can simply grep whether it is present there. The configuration generated by the Prometheus Operator is stored in a Kubernetes `Secret`, named after the Prometheus object name prefixed with `prometheus-` and is located in the same namespace as the Prometheus object. For example for a Prometheus object called `k8s` one can find out if the `ServiceMonitor` named `my-service-monitor` has been picked up with:
+
+```
+kubectl -n monitoring get secret prometheus-k8s -ojson | jq -r '.data["prometheus.yaml"]' | base64 -d | grep "my-service-monitor"
```
metalmatze: Nice command. I always simply go to the Prometheus Configuration UI. But I guess then there's the reload in between?

brancz: also this accidentally ended up in this PR I actually wanted to create a separate one for the troubleshooting stuff, but is too much work to remove :grin:

contrib/kube-prometheus/manifests/grafana/grafana-deployment.yaml
```diff
@@ -1,48 +1,59 @@
...
-        volumeMounts:
-        - name: grafana-storage
-          mountPath: /data
-        - name: grafana-datasources
-          mountPath: /grafana/conf/provisioning/datasources
-        - name: grafana-dashboards
-          mountPath: /grafana/conf/provisioning/dashboards
-        - name: grafana-dashboard-definitions-0
-          mountPath: /grafana-dashboard-definitions/0
+      - image: quay.io/coreos/monitoring-grafana:5.0.0
```
metalmatze: You're going from `5.0.3` back to `5.0.0`? Regression?

contrib/kube-prometheus/README.md
```diff
@@ -46,26 +48,69 @@ install
...
 
 To tear it all down again, run:
 
 ```bash
 hack/cluster-monitoring/teardown
 ```
 
+## Customizing
+
+As everyone's infrastructure is slightly different, different organizations have different requirements and so on there may be modifications you want to do on kube-prometheus to fit your need.
```
mxinden: > different organizations have different requirements and so on there may be modifications you want to do on kube-prometheus to fit your need.

The sentence structure is a bit confusing to me. How about:

"different organizations have different requirements. Thereby there may be modifications you want to do on kube-prometheus to fit your need."

contrib/kube-prometheus/README.md
```diff
@@ -46,26 +48,69 @@ install
...
+
+The kube-prometheus stack is intended to be a jsonnet library for organizations to consume and use in their own infrastructure repository. Below is an example how it can be used to deploy the stack properly on minikube.
+
+The three "distribution" examples we have assembled can be found in:
+
+* `example-dist/kubeadm`: contains the kube-prometheus stack with slight modifications to work properly monitoring kubeadm clusters and exposes UIs on NodePorts for demonstration purposes.
+* `example-dist/bootkube`: contains the kube-prometheus stack with slight modifications to work properly clusters created with bootkube.
```
mxinden: s/properly/properly on/

contrib/kube-prometheus/README.md
```diff
@@ -46,26 +48,69 @@ install
...
+* `example-dist/base`: contains the plain kube-prometheus stack for organizations to build on.
```
mxinden: Would you mind mentioning `base` first?

contrib/kube-prometheus/hack/cluster-monitoring/teardown
```diff
@@ -1,30 +1,22 @@
...
-find manifests/prometheus -type f ! -name prometheus-k8s-roles.yaml ! -name prometheus-k8s-role-bindings.yaml -exec kubectl --namespace "$NAMESPACE" delete -f {} \;
-kubectl delete -f manifests/prometheus/prometheus-k8s-roles.yaml
-kubectl delete -f manifests/prometheus/prometheus-k8s-role-bindings.yaml
-kctl delete -f manifests/alertmanager
-
-# Hack: wait a bit to let the controller delete the deployed Prometheus server.
-sleep 5
-
-kctl delete -f manifests/prometheus-operator
+kubectl create namespace monitoring
```
mxinden: This should be a `delete`, right?

mxinden: Same for the rest of the file.

ScottBrenner:
What would we need to do to convert our current kube-prometheus stacks to jsonnet after this goes through?

contrib/kube-prometheus/README.md
```diff
@@ -1,5 +1,7 @@
 # kube-prometheus
 
+> Note that everythin in the `contrib/kube-prometheus/` directory is experimental and may change significantly at any time.
```
ScottBrenner: > Note that everythin**g** in the `contrib/kube-prometheus/` directory is experimental and may change significantly at any time.

brancz:
@ScottBrenner we will still keep the manifests currently present in `manifests/` if people still want to work with the bare manifests, but it will always be a bare minimum of things. In fact they are from now on generated by `example-dist/base/kube-prometheus.jsonnet`. The idea is that people take exactly that jsonnet file and do all the additions they want, that way these files don't have to support any possible customization, as it is necessary with helm, but instead just a minimum common example and then people can add whatever they want with jsonnet. At that point upgrading to new kube-prometheus versions is just a button push away as your changes are code and are automatically applied :slightly_smiling_face: .

contrib/kube-prometheus/hack/example-service-monitoring/deploy
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-kubectl apply -f manifests/examples/example-app
+kubectl apply -f experimental/examples/example-app
```
mxinden: As far as I can tell, this path does not exist. Don't you mean `examples/example-app` here? Same below in `teardown`?

brancz: good catch!

PR #1219, opened 2018-04-13
Enable Thanos sidecar

mxinden:
## prometheus: enforce external labels

Add external labels `prometheus` and `prometheus_replica` via Pod env
variables and Prometheus config expansion.

## prometheus: Drop 'prometheus_replica' label on alerts

For Thanos to differentiate Prometheus replicas and shards, we
introduced the `prometheus-replica` external label. This label makes two
alerts from two Prometheus replicas indentical. Alertmanager doesn't
deduplicate them, resulting in duplicate alert notifications.

This patch drops the `prometheus_replica` label on Prometheus alerts.

In addition it does the following test refactorings:

- In `TestAlertmanagerZeroDowntimeRollingDeployment` poll whether
Prometheus is firing the necessary alert, instead of simply waiting for
a fixed amount of time.

- In `TestAlertmanagerZeroDowntimeRollingDeployment` simulate
Alertmanager update between more recent versions (`v0.13.0` - `v0.14.0`)

- In `TestPrometheusReloadRules` instead of checking the Prometheus log
for a rule reload, load specific rule set and check for firing alert.


contrib/kube-prometheus/manifests/thanos/prometheus-self.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: monitoring.coreos.com/v1
```
brancz: can we move these manifests to `experimental/` to match the changes in [this](https://github.com/coreos/prometheus-operator/pull/1206) PR?

PR #1222, opened 2018-04-14
grafana: count only running pods in utilization calculation

ntfrnzn:
Kubernetes clusters with Jobs may have a significant number
of pods in 'Completed' states.  Only "Running" state should
count against capacity.

### DIFF
--- a/helm/grafana/dashboards/kubernetes-capacity-planning-dashboard.json
+++ b/helm/grafana/dashboards/kubernetes-capacity-planning-dashboard.json
@@ -808,7 +808,7 @@
                         "steppedLine": false,
                         "targets": [
                             {
-                                "expr": "sum(kube_pod_info)",
+                                "expr": "sum(kube_pod_status_phase{phase='Running'})",
                                 "format": "time_series",
                                 "intervalFactor": 2,
                                 "legendFormat": "Current number of Pods",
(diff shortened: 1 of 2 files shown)


helm/grafana/dashboards/kubernetes-capacity-planning-dashboard.json
```diff
@@ -808,7 +808,7 @@
                         "steppedLine": false,
                         "targets": [
                             {
-                                "expr": "sum(kube_pod_info)",
+                                "expr": "sum(kube_pod_status_phase{phase='Running'})",
```
brancz: It feels like it would make sense to show the aggregate of Pods of all phases. So rather something like "sum(kube_pod_status_phase) by(phase)", that way people can spot if there are many pods in other phases.

ntfrnzn: The distribution of pod phases can be very important for looking at workflows, and may deserve a chart of its own.  For capacity planning, however, the only relevant state is "Running" 

As an example, I have a cluster with a large number of completed jobs which, as intended, have not been cleaned up.  The capacity estimate including all pods indicates that the cluster is running at 211% capacity.  Since this is a capacity graph, it seems to indicate that I need to add more computing resources.  In fact, the cluster should display 13% capacity.

For people who are using the cluster as a job processing facility, the number of completed (or failed) pods is relevant.  However, the number of completed pods will probably be much much greater than the number of running pods, and plotting the phase distribution with take some design effort to make sure it is easily interpretable.  Most people, I believe, are not using kubernetes clusters in this way, so the pod phase distribution chart might be something not generally needed.

ntfrnzn:
I'll open a pr in the other repo and close this one.

If I can figure out a nice way to display pod phase I'll suggest that too.

helm/grafana/dashboards/kubernetes-capacity-planning-dashboard.json
brancz: Agreed. For capacity planning it makes most sense only to show running pods.

PR #1223, opened 2018-04-15
Ensure statuses of ctrl-mgr and scheduler are fetched

naseemkullah:
In kubeadm 1.10 (and perhaps earlier versions, not sure), ctrl-mgr and scheduler pods' label key is `component=` rather than `k8s-app=`, which is the selector key used for kube-prometheus exporters for these components.

### DIFF
--- a/contrib/kube-prometheus/docs/kube-prometheus-on-kubeadm.md
+++ b/contrib/kube-prometheus/docs/kube-prometheus-on-kubeadm.md
@@ -42,7 +42,7 @@ schedulerExtraArgs:
   address: 0.0.0.0
 ```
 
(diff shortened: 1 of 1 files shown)


mxinden:
Hi @naseemkullah,

Thanks a bunch for the contribution! We are soon transitioning the `kube-prometheus` project to jsonnet. Once https://github.com/coreos/prometheus-operator/pull/1206 is merged, we could encode the label differences in the `example-dist` script of each distribution (see e.g. [kubeadm](https://github.com/coreos/prometheus-operator/pull/1206/files#diff-05a756d8c3ef1b4d8f4a622df610994d)).

What do you think? Would you like to propose a PR, once #1206 is merged?

naseemkullah:
Hi @mxinden,

Cool! 
Indeed I would, or feel free to on my behalf.

PR #1224, opened 2018-04-17
fix etcd alert rule: InsufficientMembers

Cplo:
(count(up{job="etcd"}) / 2 - 1) 
if count = 3, then (count(up{job="etcd"}) / 2 - 1)  = 0.5
This will also trigger an alarm if only one etcd is down.
So the rules should be written like this:
((count(up{job="etcd"}) - 1) / 2)

### DIFF
--- a/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
+++ b/contrib/kube-prometheus/manifests/prometheus/prometheus-k8s-rules.yaml
@@ -45,7 +45,7 @@ data:
     - name: ./etcd3.rules
       rules:
       - alert: InsufficientMembers
-        expr: count(up{job="etcd"} == 0) > (count(up{job="etcd"}) / 2 - 1)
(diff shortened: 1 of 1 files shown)


brancz:
Could you first send a PR to the etcd repository? The rules here are vendored from there: 

* https://github.com/coreos/etcd/blob/master/Documentation/op-guide/etcd3_alert.rules.yml
* https://github.com/coreos/etcd/blob/master/Documentation/op-guide/etcd3_alert.rules

Thanks!

metalmatze:
This is actually related to https://github.com/etcd-io/etcd/pull/10150 where we're fixing this upstream. The changes made there, will eventually show up within kube-prometheus.
I will leave it open to make sure we merge these into kube-prometheus.

metalmatze:
kube-prometheus just moved to https://github.com/coreos/kube-prometheus. 
See [contrib/kube-prometheus/README.md](https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/README.md) for details.

The fix was merged upstream and is part of kube-prometheus now.

PR #1218, opened 2018-04-13
Helm chart changes

kousik93:
Added additionalServiceMonitorLabels template incase of using custom ServiceMonitorSelector.
Fixed helm chart file permissions by removing executable bit.

PR #1226, opened 2018-04-17
[helm/alertmanager] Choose custom prometheus label value

richerve:
The value of the prometheus label is fixed to `Release.Name`, this means that alertmanager and prometheus should be deployed on the same release. In cases where this is not desirable or possible is useful to be able to specify a custom value for that label.

This can be more flexible allowing any generic set of labels but I didn't want to complicate the PR for now and preferred to maintain compatibility.

### DIFF
--- a/helm/alertmanager/Chart.yaml
+++ b/helm/alertmanager/Chart.yaml
@@ -9,4 +9,4 @@ maintainers:
 name: alertmanager
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.14
+version: 0.0.15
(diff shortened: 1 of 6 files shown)


richerve:
@gianrubio I saw #1218, if that's going to be merged then maybe this can be included there, as it doesn't address this use case specifically. If not this will not cause conflicts anyway.

richerve:
CI seems to be stuck

richerve:
Rebased

gianrubio:
thanks @richerve

PR #1229, opened 2018-04-18
kube-prometheus: Move service monitors to target folder

mxinden:
With 1d00e0ab ("ServiceMonitor: Enable Prometheus to select ServMon
outside own ns") merged: To emphasize the fact, that ServiceMonitors can
be deployed alongside their targets, this patch moves the Alertmanager,
Kube-State-Metrics, Node-Exporter and Prometheus Operator
Service-Monitor into the respective target folder in the kube-prometheus
project.


brancz:
lgtm on green

PR #1230, opened 2018-04-18
[helm/prometheus] Custom ServiceAccountName when rbac disabled

richerve:
On the cases where `global.rbacEnable: false` the `serviceAccountName` attribute of the Prometheus is not defined so is not possible to define a ServiceAccount outside the Chart and let the Prometheus object use it.

This change allows the case where custom rbac objects like ClusterRole, ServiceAccount and bindings are managed outside the Chart (so `global.rbacEnable: false`) and we want that the Prometheus objects uses that externally defined account.

## Specific use case
When installing prometheus on an specific namespace I want to define a custom role that gives access only to resources inside the namespace and with that a custom RoleBinding, so having full control of the ServiceAccount, makes sense to use that account on the Prometheus Chart instead of relying on the preconfigured account.

### DIFF
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -9,4 +9,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.51
+version: 0.0.52
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -9,4 +9,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.52
+version: 0.0.53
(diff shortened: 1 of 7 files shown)


richerve:
@gianrubio can you run the tests again? It seems that was a timeout issue again:

```
2018/04/18 14:41:34 expected 1 Prometheus instances, got 0
--- FAIL: TestPrometheusDiscoverTargetPort (196.19s)
	prometheus_test.go:496: validating Prometheus target discovery failed: timed out waiting for the condition```

gianrubio:
@richerve triggered again

richerve:
@gianrubio, will rebase later

richerve:
@gianrubio rebased

gianrubio:
Please  make the serviceAccount changes following the charts upstream best practices? https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/rbac.md

richerve:
@gianrubio done, Is much cleaner this way.

PR #1232, opened 2018-04-18
Move the initial delay seconds to Readiness, not Liveness.

virusdave:
statefulset: fix which probe has initial delay

The readiness probe is used for the case of a pod being initially ready, not the liveness probe.
Since this is the grace period used for recovering bad metrics data, the grace period for recovery should be applied to readiness, not liveness.

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -490,7 +490,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMaps []
 
 	var livenessProbeHandler v1.Handler
 	var readinessProbeHandler v1.Handler
-	var livenessProbeInitialDelaySeconds int32
+	var livenessFailureThreshold int32
 	if (version.Major == 1 && version.Minor >= 8) || version.Major == 2 {
 		livenessProbeHandler = v1.Handler{
 			HTTPGet: &v1.HTTPGetAction{
@@ -504,7 +504,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMaps []
 				Port: intstr.FromString("web"),
 			},
 		}
-		livenessProbeInitialDelaySeconds = 30
+		livenessFailureThreshold = 6
 	} else {
 		livenessProbeHandler = v1.Handler{
 			HTTPGet: &v1.HTTPGetAction{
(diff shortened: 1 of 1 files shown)


brancz:
I believe this was fine tuned before. IIRC it was because live was needs to be larger because Prometheus 1.x was performing crash recovery, which can take a long time and Prometheus doesn’t starts certain handlers when it has finished,  it it’s possible that we may have missed something, especially as I believe they have been revised in the latest releases.

mxinden:
@virusdave I have been discussing this with @brancz.

> The readiness probe is used for the case of a pod being initially ready, not the liveness probe.
Since this is the grace period used for recovering bad metrics data, the grace period for recovery should be applied to readiness, not liveness.

 First of all thanks a bunch for pointing this out!

Instead of having a `InitialDelaySeconds` delay in the first place, how about simply increasing the `FailureThreshold` instead? This way a quick Prometheus not recovering from bad metrics data is marked `ready` faster, while a Prometheus taking longer for recovery still has the time to do so.

@virusdave What do you think?

For the future me: [Prometheus code](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L495) marking Prometheus as _ready_.

virusdave:
@mxinden That's a great suggestion.
In our case, we were seeing metrics recovery triggering a readiness failure loop.  Having plenty of time but still a rapid 'short circuit on trivial successful startup' is the best of both worlds.  Stay tuned, i'll update this PR.

virusdave:
Updated the PR.

mxinden:
@virusdave Thanks! Would you mind replacing the `livenessProbeInitialDelaySeconds` with an increased `FailureThreshold` as well?

virusdave:
@mxinden Sure thing, done!

virusdave:
So, it looks like my understanding of the relationship between these two probes was a little bit off, which means this change didn't actually solve the problem it was supposed to.

Per the discussion here (https://kubernetes.slack.com/archives/C09NXKJKA/p1525381511000438), it looks like the original change in this PR was the correct one (set a high `InitialDelaySeconds` for `liveness`, no change for `readiness`, leave periods on both the same (and relatively low, for quick responsiveness if they die).

So i propose changing this back, but it apparently led to some test timeout issues.  Thoughts on how to proceed @mxinden @brancz ?

mxinden:
@virusdave Mhh, this is confusing to me. The K8s [documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) says:

> The initialDelaySeconds field tells the kubelet that it should wait 5 second before performing the first probe.

Which I understand as K8s does not check anything before the `initialDelaySeconds` isn't over. Did I misunderstand the docs or the Slack conversation?

In addition as far as I understand the code [here](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L495) Prometheus can take a lot of time between `liveness` and `readiness`, right?

Thanks for coming back to this!

PR #1235, opened 2018-04-19
[helm/prometheus] Custom prometheus label value

richerve:
Similar to #1226, will be handy to specify a custom value for the prometheus label on the prometheus Chart too.

Default value of .Release.Name is kept.

### DIFF
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -9,4 +9,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.50
+version: 0.0.51
--- a/helm/kube-prometheus/requirements.yaml
+++ b/helm/kube-prometheus/requirements.yaml
@@ -6,7 +6,7 @@ dependencies:
     condition: deployAlertManager
 
   - name: prometheus
-    version: 0.0.26
+    version: 0.0.27
     #e2e-repository: file://../prometheus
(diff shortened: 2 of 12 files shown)


gianrubio:
thanks @richerve

richerve:
@gianrubio I've just noticed that using prometheusLabelValue to change the prometheus label renders the service not working because the underlying Pods created by the StatefulSet still have other value for that label so the selector on the Service fails to find the Pods.

Looking at the code: https://github.com/coreos/prometheus-operator/blob/42788a1a1e10f717ce6ef1a27599486333bcf228/pkg/prometheus/statefulset.go#L539-L557

I noticed that the "prometheus" label is hardcoded to p.Name (`podLabels["prometheus"] = p.Name`) even in the case were `podMetadata` is defined on the Prometheus object.

So in order to allow this PR to work, the labels in the code will need to be defined before and let podMetadata to overwrite them if they're present. What do you think?

pesimon:
I've also experience an error in templating, when no `prometheusLabelValue`is not defined 

    can't evaluate field prometheusLabelValue in type interface {}

Assuming this  

        prometheus: {{ .Values.prometheusLabelValue | default $releaseName | quote }}

needs to be change to 

        prometheus: {{ default  $releaseName .Values.prometheusLabelValue  | quote }}

richerve:
@pesimon are you using the latest version? `prometheusLabelValue` should be set to empty on values.yaml, but it must be defined, latest version of the Chart should include that on values.yaml.

Using the `default` function with or without the pipe is the same behavior so that's not the reason.

PR #1253, opened 2018-04-25
service-monitor: Revert static targets configuration in service monitors

mxinden:
With the _additional-scrape-config_ feature (https://github.com/coreos/prometheus-operator/pull/1246) via Kubernetes secrets, the _static-configs_ feature in Service Monitors is redundant.

I suggest reverting the latter one, as _additional-scrape-config_ enables users to do a greater range of custom configurations.

@brancz @metalmatze What do you think?


brancz:
lgtm

metalmatze:
Yes, please revert the _static-configs_, if you found a better way.
Are you reverting that one, or should I send a PR?

mxinden:
@metalmatze This PR reverted both the feature PR, as well as the feature fix PR. Am I missing something else?

metalmatze:
Oh, I somehow missed that this was the reverting PR already. Thanks for taking care! :ok_hand:

PR #1254, opened 2018-04-25
kube-prometheus: Remove hard-coded Prometheus Operator version

mxinden:


contrib/kube-prometheus/jsonnet/prometheus-operator/prometheus-operator-deployment.libsonnet
```diff
@@ -2,7 +2,7 @@ local k = import "ksonnet.beta.3/k.libsonnet";
 local rawVersion = importstr "../../../../VERSION";
 
 local removeLineBreaks = function(str) std.join("", std.filter(function(c) c != "\n", std.stringChars(str)));
-local version = "v0.18.1";//removeLineBreaks(rawVersion);
+local version = "v" + removeLineBreaks(rawVersion);
```
brancz: Problem with the version importstr is that if these files are used as a library, as we're intending it to be used, then users must have the full prometheus operator repository. Ideally I'd like to see each of these components to be a separate packages which are separately configurable, then we can use something like [bumpversion](https://github.com/peritus/bumpversion) to bump the version in the jsonnet package for the prometheus operator on new releases (as a matter of fact we maybe should do that right now instead of this, wdyt?).

mxinden: @brancz I see your point. I am not too keen on introducing a separate tool for a task that can also be achieved via a simple `sed`. For example: `sed -i "s/local version = \".*\";/local version = "v$(cat VERSION)";/" contrib/kube-prometheus/jsonnet/prometheus-operator/prometheus-operator-deployment.libsonnet`.

How about I add the above `sed` command to the `generate-kube-prometheus` target of the top level Makefile?

brancz: Sounds good to me

mxinden:
@brancz would you mind taking another look?

brancz:
lgtm

PR #1257, opened 2018-04-25
Add customizable ingress path to grafana helm chart

aadu:


gianrubio:
Closing this because grafana just got support to path, see [here](https://github.com/coreos/prometheus-operator/blob/master/helm/grafana/templates/ingress.yaml#L27)

PR #1260, opened 2018-04-26
kube-prometheus: fix addon-resizer role binding

bdimcheff:
The addon resizer was referring to an incorrect role binding and was unable to access the API.  This changes the rolebinding name to refer to the actually defined rolebinding.

Without this change, you get a lot of messages like this after deploying to a cluster with rbac enabled:

```
monitoring/kube-state-metrics-5d8784b89d-qw9r2[addon-resizer]: E0426 02:02:19.908809       1 nanny_lib.go:95] pods "kube-state-metrics-5d8784b89d-qw9r2" is forbidden: User "system:serviceaccount:monitoring:kube-state-metrics" cannot get pods in the namespace "monitoring": Unknown user "system:serviceaccount:monitoring:kube-state-metrics"
```

Note that I couldn't easily get the ksonnet stuff building, so I just changed both the jsonnet file and the yaml manually.


bdimcheff:
It looks like this has been an issue since the jsonnet conversion: https://github.com/coreos/prometheus-operator/commit/8b54c34352020dda52fb69d2cba46b44a67cb36e#diff-e77d0212fa06aaf651e1b9c0e1a90771R9

PR #1261, opened 2018-04-26
Bump helm prometheus operator chart

gianrubio:
bump operator to v0.18.1
review Chart.yaml according to helm charts upstream


gianrubio:
Actually I'll wait for #1262 to bump to 0.19.0

Issue #1240, opened 2018-04-23
[Helm] Adding a serviceMonitor to kube-prometheus values throws helm error


hobbsh:
Seems to be an issue with the helm template for serviceMonitors - could it be possible that it is expecting some helm labels for my service, but I didn't deploy my service with helm?

**What did you do?**
Added a service monitor in kube-prometheus values.yml and updated with helm (using Ansible):

**What did you expect to see?**
Expected the serviceMonitor to create on its own

**What did you see instead? Under which circumstances?**
```
{
   "changed":true,
   "cmd":[
      "/usr/local/bin/helm",
      "upgrade",
      "--install",
      "kube-prometheus",
      "coreos/kube-prometheus",
      "--namespace",
      "monitoring",
      "-f",
      "/root/.helm/kube-prometheus-values.yml"
   ],
   "delta":"0:00:01.645034",
   "end":"2018-04-23 02:19:03.529344",
   "msg":"non-zero return code",
   "rc":1,
   "start":"2018-04-23 02:19:01.884310",
   "stderr":"Error: UPGRADE FAILED: render error in \"kube-prometheus/charts/prometheus/templates/servicemonitors.yaml\": template: kube-prometheus/charts/prometheus/templates/servicemonitors.yaml:18:30: executing \"kube-prometheus/charts/prometheus/templates/servicemonitors.yaml\" at <.Values.prometheusLa...>: can't evaluate field prometheusLabelValue in type interface {}",
   "stderr_lines":[
      "Error: UPGRADE FAILED: render error in \"kube-prometheus/charts/prometheus/templates/servicemonitors.yaml\": template: kube-prometheus/charts/prometheus/templates/servicemonitors.yaml:18:30: executing \"kube-prometheus/charts/prometheus/templates/servicemonitors.yaml\" at <.Values.prometheusLa...>: can't evaluate field prometheusLabelValue in type interface {}"
   ]
}
```
**Environment**
Using Ansible (kubespray) to install/upgrade prometheus-operator and kube-prometheus

```
$ sudo helm version
Client: &version.Version{SemVer:"v2.8.1", GitCommit:"6af75a8fd72e2aa18a2b278cfe5c7a1c5feca7f2", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.8.1", GitCommit:"6af75a8fd72e2aa18a2b278cfe5c7a1c5feca7f2", GitTreeState:"clean"}
```
* Kubernetes version information:

```
[...]

PR #1263, opened 2018-04-26
Fix scope when adding serviceMonitors to prometheus

Nonymus:
Adresses #1240 

Local tests with serviceMonitors worked, but I don't know the project's CI well enough to add a real test.

@gianrubio Feel free to throw this away if you already have a better solution. I just needed a quick fix.

### DIFF
--- a/helm/kube-prometheus/requirements.yaml
+++ b/helm/kube-prometheus/requirements.yaml
@@ -6,7 +6,7 @@ dependencies:
     condition: deployAlertManager
 
   - name: prometheus
-    version: 0.0.28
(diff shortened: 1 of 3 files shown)


Nonymus:
```--- FAIL: TestPrometheusDiscovery (1098.08s)
	prometheus_test.go:418: validating Prometheus target discovery failed: timed out waiting for the condition```
Don't see how this could be related to anything in the PR. Anyone fit in the actual go code any ideas?

gianrubio:
thanks @Nonymus

PR #1264, opened 2018-04-26
Fixes a problem with variable scope

zapur1:
The error I was encountering was 

```
...
executing \"kube-prometheus/charts/prometheus/templates/servicemonitors.yaml\" at <.Values.prometheusLa...>: can't evaluate field prometheusLabelValue in type interface {}"
...
```

It was due to the scope change inside the `range` block


gianrubio:
thanks for you pr @zapur1 but #1263 came first so I'm closing this as a duplicate PR

PR #1268, opened 2018-04-26
[helm/grafana] Fix ServiceMonitor label selector

yann-soubeyrand:
Edit the app label selector to be the same as the one in the service template (https://github.com/coreos/prometheus-operator/blob/7d4d93734a31826103137ee3b935388f8db681df/helm/grafana/templates/svc.yaml#L4). Also remove the grafana label selector as it is not present in the service template.


gianrubio:
lgtm on green

gianrubio:
After I rebase this PR the changes were already applied, so I'll mark this PR as duplicate of #1261 

thanks for your time @yann-soubeyrand

PR #1270, opened 2018-04-26
Update pods dashboard

dimm0:
Make namespace, pod and container fields in pods dashboard sorted


brancz:
the python dashboards are only there for migration purposes, we are in the process of migrating everything, we'll have an update for you soon

PR #1271, opened 2018-04-27
Scrape nginx ingress engine

devcfgc:
I was having the same problems reported in this issue https://github.com/coreos/prometheus-operator/issues/617 and I fixed it adding permissions to the cluster role used by prometheus-k8s.


brancz:
Kube prometheus is meant you you to take and adapt, we cannot possibly apart to every use case.

brancz:
I encourage you to modify the manifests to your need, but the set of manifests we provide will stay the most minimal version to get basic cluster monitoring up and running.

devcfgc:
ok thanks =)

PR #1277, opened 2018-04-28
Add leveled logging

brancz:
While debugging something myself I wished for some additional debug logs, so adding leveled logging (and some additional logging, but mostly adapting existing log calls).

@mxinden

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -23,6 +23,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


sgmiller:
Looks great, thanks!

cmd/operator/main.go
```diff
@@ -35,10 +36,28 @@ import (
...
+	"github.com/go-kit/kit/log/level"
+)
+
+const (
+	logLevelAll   = "all"
+	logLevelDebug = "debug"
+	logLevelInfo  = "info"
+	logLevelWarn  = "warn"
+	logLevelError = "error"
+	logLevelNone  = "none"
```
mxinden: Do we really need `All` and `None`? I don't know what should be more verbose than `Debug`. In addition if users want `None` they can simply do `> /dev/null`, right?

brancz: Just going with what the go kit logger has so offer here. I think the none case is legit as it saves actual function calls internally, and all is easier to digest for users when they really want to make sure they're actually getting all logs.

PR #1278, opened 2018-04-28
prometheus: Allow all selector for namespaces

brancz:
The empty selector is meant as the "all" selector in Kubernetes, this change allows using the all selector to be used for the namespace selector of the Prometheus resource.

@mxinden

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1076,8 +1076,8 @@ func (c *Operator) selectServiceMonitors(p *monitoringv1.Prometheus) (map[string
 		return nil, err
 	}
 
-	// If 'ServiceMonitorNamespaceSelector' is empty, only check own namespace.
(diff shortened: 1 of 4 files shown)


pkg/prometheus/operator.go
```diff
@@ -1075,7 +1075,7 @@ func (c *Operator) selectServiceMonitors(p *monitoringv1.Prometheus) (map[string
 	}
 
 	// If 'ServiceMonitorNamespaceSelector' is empty, only check own namespace.
```
mxinden: With this change I find this comment a bit misleading. Would you mind changing it to something like:

```go
// If 'ServiceMonitorNamespaceSelector' is nil (yaml, json: null), only check own namespace. If it is empty, check all namespaces.
```

Same would need to be copied here:
https://github.com/coreos/prometheus-operator/blob/master/pkg/client/monitoring/v1/types.go#L63-L64

PR #1279, opened 2018-04-28
kube-prometheus/jsonnet: Use jsonnet-bundler

brancz:
This PR has some large changes on the kube-prometheus sub-project. It now uses a packaging approach of jsonnet code using [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler). This allows each component to be individually packaged up and then brought together in the now meta-package kube-prometheus. As such, Grafana has already been extracted into it's own package: [kubernetes-grafana](https://github.com/brancz/kubernetes-grafana). All other components are currently still living in this repository but the goal is to eventually make them standalone packages. Additionally the majority of alerting rules and dashboards have been extracted to the [Kubernetes monitoring mixins](https://github.com/kubernetes-monitoring/kubernetes-mixin). Not all alerting rules and dashboards have been migrated there, but I'd prefer to get this merged and continue with the migration piece by piece (which is why dashboards and alerting rules have not been entirely removed from this repository).

@mxinden could you just play through the instructions in the readme against a local minikube as described in the readme? If that works I think we can merge this.

PR #1280, opened 2018-04-28
Corrected API latency recording rule name

jamtur01:
This metric name should be consistent with the 50th and 99th percentile metrics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,3 +153,9 @@ kubectl delete --ignore-not-found customresourcedefinitions \
   1. build the container image with the docker host from within minikube by running `eval $(minikube docker-env)`.
   2. You can build the container using `make container`.
   3. Finally run the e2e tests using `make e2e-test`.
+
+### Updating contributed rules
+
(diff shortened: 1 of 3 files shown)


jamtur01:
It's not clear to me why the Travis test failed...

mxinden:
All the files in `contrib/kube-prometheus/manifests/` are auto generated, based on e.g. the file you edited. Can you run `make generate` on the root level of the Prometheus Operator and check-in the changed files?

jamtur01:
Updated and added a README note explaining the steps involved.

brancz:
#1279 does this as well, so I'd prefer to merge that so we don't create merge conflicts, but let's see if we can get that done today otherwise we'll merge this. Thanks @jamtur01 !

brancz:
This has been fixed on master, and as a side note, the new location of these sources are in the [kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin) repository.

Thanks a lot for taking the initiative though, highly appreciated!

PR #1284, opened 2018-04-30
Fix typo in grafana datasources manifest

serathius:
`s/etitable/editable`

http://docs.grafana.org/administration/provisioning/#example-datasource-config-file

### DIFF
--- a/contrib/kube-prometheus/manifests/grafana-dashboardDatasources.yaml
+++ b/contrib/kube-prometheus/manifests/grafana-dashboardDatasources.yaml
@@ -5,7 +5,7 @@ data:
         "datasources": [
             {
                 "access": "proxy",
-                "etitable": false,
+                "editable": false,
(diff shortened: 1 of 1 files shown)


serathius:
Need help with fixing `./scripts/check-make-generate.sh`

brancz:
All of the generated manifests are being reworked in #1279, I recommend waiting for that change before getting into editing the dashboards as we're moving all the generation to be done by jsonnet.

brancz:
We've finished the whole jsonnet migration, but the error you found here seems to have survived it. To save you further trouble with it I already opened a new PR to fix it in the new location. Thanks a lot for this PR, I would have not caught this in a while!

Closing here, but thanks again for the catch! :slightly_smiling_face: 

For the reference, the new pull request can be found here: https://github.com/brancz/kubernetes-grafana/pull/30

PR #1290, opened 2018-05-01
Added support for multiple hostnames to alertmanager/prometheus/grafana

tmestdagh:
Instead of using the String ``fqdn`` to configure the hostname in the Ingresses, the List ``hosts`` is now used to configure the hostnames for the Ingresses.

### DIFF
--- a/helm/grafana/Chart.yaml
+++ b/helm/grafana/Chart.yaml
@@ -8,4 +8,4 @@ maintainers:
 name: grafana
 sources:
 - https://github.com/coreos/prometheus-operator
-version: 0.0.28
+version: 0.0.29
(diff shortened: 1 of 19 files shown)


helm/grafana/Chart.yaml
```diff
@@ -8,4 +8,4 @@ maintainers:
 name: grafana
 sources:
 - https://github.com/coreos/prometheus-operator
-version: 0.0.28
+version: 0.0.29
```
gianrubio: change to 0.0.30

helm/prometheus/Chart.yaml
```diff
@@ -7,4 +7,4 @@ maintainers:
 name: prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.32
+version: 0.0.33
```
gianrubio: change to 0.0.34

PR #1291, opened 2018-05-01
Improve "Control Plane Components Down" panel

naseemkullah:
We want to count the number of jobs that are down. Also adding the "instant=true" checkbox yields better results.

### DIFF
--- a/helm/grafana/Chart.yaml
+++ b/helm/grafana/Chart.yaml
@@ -8,4 +8,4 @@ maintainers:
 name: grafana
 sources:
 - https://github.com/coreos/prometheus-operator
-version: 0.0.29
+version: 0.0.28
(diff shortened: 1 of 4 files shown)


brancz:
We are in the process of migrating these dashboards to https://github.com/kubernetes-monitoring/kubernetes-mixin, but I'm fine with merging this for now, but just wanted to mention that there is a possibility that maybe we will miss to transfer things during this migration. Sorry and thanks for understanding!

Regarding merging this PR, giving @gianrubio the last call.

naseemkullah:
Sounds good! Thanks.

gianrubio:
I'd prefer to apply the changes in the upstream, otherwise this PR will be overwritten when someone run the [sync_kube_prometheus.py](https://github.com/coreos/prometheus-operator/blob/master/helm/hack/sync_kube_prometheus.py) script

naseemkullah:
Old

PR #1299, opened 2018-05-03
use node_filesystem_avail instead of free 

adamdecaf:
From https://github.com/prometheus/node_exporter/issues/269 it's noted that `node_filesystem_free` can provide a less accurate metric for when a fs write would fail (from full disk). 

I only use the `prometheus-k8s-rules` files, but am changing them all to be consistent across the codebase.


adamdecaf:
I didn't realize these were generated. Where should I go to modify the templates? 

I see `scripts/jsonnet/Dockerfile`, but not where that `assets/` folder comes from.
(for `contrib/kube-prometheus/jsonnet/kube-prometheus.libsonnet`)

brancz:
there are still some migrations to be done, but generally all common kubernetes alerts/dashboards should go into https://github.com/kubernetes-monitoring/kubernetes-mixin

adamdecaf:
@mxinden @brancz So you're both OK with this change? If so cool. I'll split out the changes to whatever repo. Just let me know.

I'm going to search in that kubernetes-mixin repo now.

adamdecaf:
@brancz Is there a file that specifies those deps? I ran `make generate`, but kubernetes-mixin wasn't updated.

brancz:
I actually just updated it :slightly_smiling_face: (see #1318 )

adamdecaf:
Cool, thanks!

Issue #1297, opened 2018-05-03
Wrong Prometheus pod label


vielktus:
**What did you do?**
helm install -n kube-prometheus

**What did you expect to see?**
```
$ kubectl describe pod prometheus-kube-prometheus-prometheus-0
Labels:         app=prometheus
                controller-revision-hash=prometheus-kube-prometheus-prometheus-c6d9ddb8
                prometheus=kube-prometheus-prometheus <----- This should be "kube-prometheus" only
```

**What did you see instead? Under which circumstances?**
prometheus=kube-prometheus-prometheus 

I checked the service and found out that it didn't have any endpoints.
The "kube-prometheus-prometheus" service will chose the pod with label:
```
 Selector:          app=prometheus,prometheus=kube-prometheus
```
But the pod label was wrongly created like I mentioned above.
Did I miss configure anything or the Operator has some issues ?

My current value for "prometheusLabelValue" is "" (blank). I just leave it.

verwilst:
I have the same issue. Kubespray-installed Kubernetes 1.10.2,

Ran following commands:

```
```

verwilst:
Also tried to set a  prometheusLabelValue during helm install, but it doesn't seem to affect the pod label, which remains "kube-prometheus-prometheus".

```
---
prometheus:
   prometheusLabelValue: "prometheus-verwilst"
```

vielktus:
@richerve I might need your help 👍

gianrubio:
@vielktus you can solve this issue removing the selector `prometehus` from the prometheus service. Leave only the selector `app: prometheus`.

I'm gonna raise a PR to fix that.

brancz:
can this be closed as #1302 is merged?

verwilst:
Installed 0.0.36:

kubectl describe svc kube-prometheus-prometheus -n monitoring
<snip>
Labels:            app=prometheus
                   chart=prometheus-0.0.36
                   heritage=Tiller
                   prometheus=kube-prometheus
                   release=kube-prometheus
</snip>
TargetPort:        9090/TCP
Endpoints:         <none>


kubectl describe pod prometheus-kube-prometheus-prometheus-0 -n monitoring
<snip>
Labels:         app=prometheus
                controller-revision-hash=prometheus-kube-prometheus-prometheus-5484998ff9
                prometheus=kube-prometheus-prometheus
                statefulset.kubernetes.io/pod-name=prometheus-kube-prometheus-prometheus-0
</snip>

[...]

PR #1302, opened 2018-05-04
Use 'prometheus.fullname' template when setting prometheus' label

twexler:
I ran into an issue that I believe this fixes, where the `prometheus` chart uses the wrong value for the `prometheus` label.

What I saw:
The `prometheus` service has a selector of `prometheus: kube-prometheus-prometheus`, but the pod created by the `Prometheus` resource has a label of `prometheus: kube-prometheus` (which is just `Release.Name`). This causes the prometheus service to fail to function, because the selector doesn't match the pod's labels.


gianrubio:
CI is failing because you didn't bumped the chart version, I did a commit to fix that. On green I'll merge it. Fix #1297

twexler:
@gianrubio I believe I've fixed it, but it is much more verbose than it used to be. Thoughts?

PR #1305, opened 2018-05-07
Fix prometheus service default pod selector in the prometheus chart

kevinjqiu:
In the case where `prometheusLabelValue` is not specified, the Prometheus CRD [is using a different default label than the service](https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus/templates/prometheus.yaml#L9), resulting in the service not being able to select the right pods.


gianrubio:
Close #1297

gianrubio:
thanks @kevinjqiu

PR #1309, opened 2018-05-07
contrib/grafana-watcher: add deprecation notice

adamdecaf:
https://github.com/coreos/prometheus-operator/issues/1025#issuecomment-369225236 and https://github.com/coreos/prometheus-operator/issues/1251#issuecomment-384079856 mention this, but I didn't see it reflected in the docs.


brancz:
thanks! :+1:

PR #1311, opened 2018-05-07
Simpler and most commonly used memory calculation

alok87:
Fix for https://github.com/coreos/prometheus-operator/issues/1310


brancz:
Alerting rules and dashboard now come from https://github.com/kubernetes-monitoring/kubernetes-mixin

I recommend doing changes there where applicable.

PR #1313, opened 2018-05-08
kube-prometheus: fix alert rule K8SManyNodesNotReady

piglei:
I believe there is a typo in alert rule "K8SManyNodesNotReady" which will results alert like "**0.25%** of Kubernetes nodes are not ready".

Fixed by multiplied the value with 100.


mxinden:
@piglei Thanks for the updates. With the helm file changes, the chart versions need to be updated as well (https://github.com/coreos/prometheus-operator/tree/master/helm#how-to-contribue). Would you mind adjusting that as well?

//CC @gianrubio (helm expert)

gianrubio:
@piglei please read the @mxinden comments and also run [this script](https://github.com/coreos/prometheus-operator/blob/master/helm/hack/sync-repo.sh) to keep grafana dashboards in sync.

piglei:
@gianrubio version bumped. But I don't know how to run `sync-repo.sh` successfully, because it seems that the script needs AWS credentials to sync all the charts to S3?

piglei:
@gianrubio Run `sync_kube_prometheus.py` successfully and seem no difference generated under `helm/exporter-kubelets/templates/`. I guess that is because I manually modified it in [this commit](https://github.com/coreos/prometheus-operator/pull/1313/commits/911d4dc2cac0485a79db8bf1353d508a35ac9474).

But the new file generated did removed some trivial new line breaks at the end of file, so I will commit it anyway.

gianrubio:
thanks @piglei

PR #1314, opened 2018-05-08
e2e/alertmanager: Check if config contains string in ReloadConfig

mxinden:
Instead of comparing full configs, just check for a unique keyword. This
makes the test more robust to upstream format changes.


brancz:
this needs a rebase

brancz:
lgtm :+1: nice

PR #1322, opened 2018-05-09
add prometheus-config-reloader image

davidstack:
add    prometheus-config-reloader =quay.io/coreos/prometheus-config-reloader:v0.0.4  in file 0prometheus-operator-deployment.yaml,so  developers can define their image.


brancz:
Those manifests are generated, which is why the builds are failing. Could you just add the image here: https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-operator/prometheus-operator.libsonnet#L122 and then run `make generate` in either the root of the repo or in `contrib/kube-prometheus`.

brancz:
There is one more tiny generation issue. Could you just run `make generate` in the root of the repository? Thanks! (and sorry for the inconvenience)

PR #1333, opened 2018-05-11
prometheus: Introduce RuleFile Custom Resource Definition 

mxinden:
This patch introduces a new Custom Resource Definition to the
Prometheus Operator - the Rule CRD. It addresses two main
needs:

1. Prometheus (alerting and recording) Rule validation during creation time
via Kubernetes Custom Resource Definition validation.

2. Life-cycle management of Prometheus application Rules alongside the
application itself, inside the applications Kubernetes namespace, not
necessarily the namespace of the scraping Prometheus instance.

A user defines Prometheus alerting and recording Rules via a Kubernetes
Custom Resource Definition. These Custom Resource Definitions can be
fully validated by the Kubernetes API server during creation time via
automatically generated OpenAPI specifications. Instead of the
restriction of a Prometheus instance to only select Rule definitions
inside its own namespace, the Prometheus specification is extended to
also specify namespaces to look for Rule Custom Resource Definitions
outside its own namespace.

---

Dependent technical changes:

- prometheus: Use github.com/jimmidyson/configmap-reload to reload rules

- prometheus: Remove Prometheus Statefulset deletion function. Starting
with K8s >=1.8 this is handled via OwnerReferences.

- prometheus: Do not add rule files checksum to Prometheus configuration
secret

- prometheus: Update StatefulSet only on relevant changes. Instead of
updating the Prometheus StatefulSet on every `sync()` run, only update
it if the input parameters to `makeStatefulSet` change.  Enforce this
via a checksum of the parameters which is saved inside the annotations
of the statefulset.

- e2e/prometheus: Check how often resources (Secret, ConfigMap,
Prometheus CRD, Service) are updated to enforce that Prometheus Operator
only updated created resources if necessary.

- contrib/prometheus-config-reloader: Remove logic to retriev K8s
ConfigMaps. These are mounted into the pod right away now.

[Design Document](https://docs.google.com/document/d/1V5pSP_b3Q7j79-IByr1_p77LRcjGHszkUu0lO09Homs/edit?usp=sharing)


pkg/client/monitoring/v1alpha1/types.go
```diff
@@ -193,6 +193,24 @@ type ServiceMonitorSpec struct {
 	NamespaceSelector NamespaceSelector `json:"namespaceSelector,omitempty"`
 }
 
+// AlertingRuleFile defines alerting rules for a Prometheus instance
+// TODO: Should we go with v1 instead of v1alpha1 right away?
+type AlertingRuleFile struct {
```
ironcladlou: If `AlertingRuleFile` is used only to hold the serialized contents of an alerting rule file (the type for which is not represented in this API), how does `AlertingRuleFile` compare to using a `ConfigMap`?

mxinden: @ironcladlou My plan is to vendor the upstream Prometheus alerting rule struct. Thereby we would have full OpenAPI generated deep validation on creation time (instant feedback). What do you think?

mxinden: I still have a couple `TODO`s on my list :/

https://github.com/coreos/prometheus-operator/pull/1333/files#diff-c8710fb169ff41dfb58c4a26a2f7a755R210

ironcladlou: That makes sense to me- cool!

Documentation/api.md
```diff
@@ -211,7 +211,8 @@ Specification of the desired behavior of the Prometheus cluster. More info: http
 | externalUrl | The external URL the Prometheus instances will be available under. This is necessary to generate correct URLs. This is necessary if Prometheus is not served from root of a DNS name. | string | false |
 | routePrefix | The route prefix Prometheus registers HTTP handlers for. This is useful, if using ExternalURL and a proxy is rewriting HTTP routes of a request, and the actual ExternalURL is still true, but the server serves requests under a different route prefix. For example for use with `kubectl proxy`. | string | false |
 | storage | Storage spec to specify how storage shall be used. | *[StorageSpec](#storagespec) | false |
-| ruleSelector | A selector to select which ConfigMaps to mount for loading rule files from. | *[metav1.LabelSelector](https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#labelselector-v1-meta) | false |
```
brancz: We will need to continue to support this for the time being, I suggest this should select both configmaps and Rule CRDs for the forseeable future and the configmap selector stays local to the namespace and the CRD selector is respective to the ruleNamespaceSelector.

In fact we should actually do a migration where we re-create rule CRDs from the content of the configmaps.

Documentation/api.md
```diff
@@ -211,7 +211,8 @@ Specification of the desired behavior of the Prometheus cluster. More info: http
...
+| alertingRuleFileSelector | A selector to select which AlertingRuleFiles to mount for loading alerting rules from. | *[metav1.LabelSelector](https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#labelselector-v1-meta) | false |
```
brancz: It's not just alerting rules, it's also recording rules, therefore usually referred to as "rule"

cmd/operator/main.go
```diff
@@ -115,6 +115,8 @@ func Main() int {
 		return 1
 	}
 
+	fmt.Println("image: ", cfg.ConfigReloaderImage)
```
brancz: remove debug fmt.Println

pkg/k8sutil/k8sutil.go
```diff
@@ -174,15 +174,16 @@ func GetMinorVersion(dclient discovery.DiscoveryInterface) (int, error) {
...
-		Group:                 group,
-		Kind:                  crdKind.Kind,
-		Version:               monitoringv1.Version,
+		SpecDefinitionName: crdKind.SpecName,
+		EnableValidation:   validation,
+		Labels:             crdutils.Labels{LabelsMap: labels},
+		ResourceScope:      string(extensionsobj.NamespaceScoped),
+		Group:              group,
+		Kind:               crdKind.Kind,
+		// Version:               monitoringv1.Version,
```
brancz: remove comment

pkg/client/monitoring/v1alpha1/types.go
brancz: I already commented on the name above, I think this should probably be "PrometheusRule" instead.

Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM quay.io/prometheus/busybox:latest
 
 ADD .build/linux-amd64/operator /bin/operator
+ADD ./contrib/prometheus-config-reloader/prometheus-config-reloader /bin/prometheus-config-reloader
```
brancz: I've thought of this as well, but I'm not sure I like it, especially if the container image repo is the same, then people will be maximum confused I think. Even today people think the operator runs as a sidecar. It also has a large impact on the container size, the reload container is very very small whereas operator+reloader is a pretty large image.

metalmatze:
I have a general question about this feature:

How does this align with the current work going on to make jsonnet bases monitoring mixins with alerts? Are these features orthogonal? I guess it tries to solves the problem with a different approach, but on the other hand can also be combined?
So I'd have those mixins, make slight adjustments and then generate a CRD from that?

brancz:
@metalmatze they're unrelated, this CRD is merely so that we can collect Prometheus rules from across namespaces similar to cross namespace ServiceMonitors. In fact as these rules are identical to the Prometheus rule definition they are compatible.

brancz:
@mxinden do they work alongside each other? I seem to recall there was some problem as some point. Otherwise v1alpha1 sounds good as it's a new resource.

ant31:
AFAIK, CRD are painful to upgrade (from v1alpha1 to v1), I would save us and users this additional migration.  Also there is no plan to keep both at the same time alive (configmap + crd), so as soon as we release this, the CRD will replace the configmap, then I would go for  v1.
Last The spec itself is well-known, we don't add any feature/fields that exist already in the current configmap.

brancz:
Yeah after thinking about it once more I think putting it in v1 is fine, given that the spec we're using is identical to the Prometheus rule spec, which cannot be broken until the next Prometheus major version anyways.

mxinden:
General design document can be found [here](https://docs.google.com/document/d/1V5pSP_b3Q7j79-IByr1_p77LRcjGHszkUu0lO09Homs/edit?usp=sharing).

Feedback by the community is very welcome.

Dockerfile
mxinden: :+1:

pkg/client/monitoring/v1/types.go
```diff
@@ -397,6 +405,63 @@ type ServiceMonitorList struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of desired alerting rule definitions for Prometheus.
+	Spec RuleFileSpec `json:"spec"`
+}
+
+// RuleFileSpec contains specification parameters for a Rule.
+// +k8s:openapi-gen=true
+type RuleFileSpec struct {
+	// Content of Prometheus rule file
+	Content RuleGroups `json:"content,omitempty"`
```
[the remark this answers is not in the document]
mxinden: True. I will address that.

pkg/client/monitoring/v1/types.go
```diff
@@ -550,3 +615,11 @@ func (l *ServiceMonitor) DeepCopyObject() runtime.Object {
 func (l *ServiceMonitorList) DeepCopyObject() runtime.Object {
 	return l.DeepCopy()
 }
+
+func (l *RuleFile) DeepCopyObject() runtime.Object {
+	panic("DeepCopyObject not implemented for ServiceMonitorList")
```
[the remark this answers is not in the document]
mxinden: Not entirely sure, but we are auto-generating these, right?

I have replaced the panic with `return l.DeepCopy()`.

pkg/prometheus/operator.go
```diff
@@ -470,27 +487,56 @@ func (c *Operator) syncNodeEndpoints() error {
 	return nil
 }
 
+// TODO: All handleXXX do the same
```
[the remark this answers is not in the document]
mxinden: `handleSmonUpdate` and `handleRuleUpdate` seem equal to me, no?

pkg/prometheus/operator.go
```diff
@@ -1117,15 +1102,25 @@ func (c *Operator) selectServiceMonitors(p *monitoringv1.Prometheus) (map[string
...
+		// If Prometheus, RuleFile and ServiceMonitor objects are already registered, we
 		// won't attempt to do so again.
 		return nil
 	}
 
 	crds := []*extensionsobj.CustomResourceDefinition{
 		k8sutil.NewCustomResourceDefinition(c.config.CrdKinds.Prometheus, c.config.CrdGroup, c.config.Labels.LabelsMap, c.config.EnableValidation),
 		k8sutil.NewCustomResourceDefinition(c.config.CrdKinds.ServiceMonitor, c.config.CrdGroup, c.config.Labels.LabelsMap, c.config.EnableValidation),
+		k8sutil.NewCustomResourceDefinition(
+			monitoringv1.CrdKind{
```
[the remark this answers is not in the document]
mxinden: :+1: thanks, I forgot that.

pkg/prometheus/promcfg_test.go
```diff
@@ -150,7 +153,8 @@ alerting:
 	result := string(cfg)
 
 	if expected != result {
-		t.Fatalf("Unexpected result.\n\nGot:\n\n%s\n\nExpected:\n\n%s\n\n", result, expected)
+		pretty.Compare(expected, result)
```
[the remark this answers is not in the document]
mxinden: I share that feeling. On the up side this is only used in testing (unit & e2e) code, no production code.

pkg/prometheus/statefulset.go
```diff
@@ -223,50 +226,8 @@ func (l *ConfigMapReferenceList) Swap(i, j int) {
...
-	return json.Marshal(cml)
-}
-
-func makeConfigSecret(p *monitoringv1.Prometheus, configMaps []*v1.ConfigMap, config Config) (*v1.Secret, error) {
-	b, err := makeRuleConfigMapListFile(configMaps)
-	if err != nil {
-		return nil, err
-	}
-
+// TODO: No need to return an error
```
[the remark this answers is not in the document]
mxinden: :D Fixed it and forgot to remove the TODO :man_facepalming:

pkg/prometheus/rulefile.go
```diff
@@ -0,0 +1,212 @@
...
+				marshalErr = err
+				return
+			}
+			ruleFiles[fmt.Sprintf("%v-%v.yaml", file.Namespace, file.Name)] = string(content)
+		})
+		if err != nil {
+			return ruleFiles, err
+		}
+		if marshalErr != nil {
+			return ruleFiles, marshalErr
```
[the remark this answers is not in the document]
mxinden: I don't understand what you mean.

There are two types of possible errors here:
1. Error thrown by `ListAllByNamespace`
2. Error thrown by `yaml.Marshal`

Not very proud of the structure of my code here. Happy for suggestions.

pkg/prometheus/rulefile.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, ns := range namespaces {
+		var marshalErr error
+		err := cache.ListAllByNamespace(c.ruleFileInf.GetIndexer(), ns, fileSelector, func(obj interface{}) {
+			file := obj.(*monitoringv1.RuleFile)
+			content, err := yaml.Marshal(file.Spec.Content)
+			if err != nil {
+				marshalErr = err
+				return
+			}
+			ruleFiles[fmt.Sprintf("%v-%v.yaml", file.Namespace, file.Name)] = string(content)
```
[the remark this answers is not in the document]
mxinden: I have just started a Prometheus with two unique rule files, with matching rule group names. Both the alerts from file one, as well as file two show up in the Prometheus UI.

pkg/prometheus/operator.go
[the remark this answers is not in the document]
brancz: yes, that's true, we can look into deduping those

pkg/prometheus/promcfg_test.go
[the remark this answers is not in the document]
brancz: yeah, I've seen people do some hacks with a "testing" Gopkg.toml that they run `dep ensure` on in CI but don't check in the vendor directory for it, but I think I'm ok with just going with standard dep usage here ... let's not overcomplicate

mxinden:
@brancz Thanks for all the comments! Would you mind taking another look?

brancz:
Nice job! lgtm :+1:

Documentation/api.md
```diff
@@ -211,7 +216,9 @@ Specification of the desired behavior of the Prometheus cluster. More info: http
...
 | storage | Storage spec to specify how storage shall be used. | *[StorageSpec](#storagespec) | false |
-| ruleSelector | A selector to select which ConfigMaps to mount for loading rule files from. | *[metav1.LabelSelector](https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#labelselector-v1-meta) | false |
+| ruleFileSelector | A selector to select which RuleFiles to mount for loading alerting rules from. | *[metav1.LabelSelector](https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#labelselector-v1-meta) | false |
+| ruleSelector | DEPRECATED with Prometheus Operator 'v0.20.0'. Any value in this field will just be copied to 'RuleFileSelector' field | *[metav1.LabelSelector](https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#labelselector-v1-meta) | false |
+| ruleFileNamespaceSelector | Namespaces to be selected for RuleFiles discovery. If empty, only check own namespace. | *[metav1.LabelSelector](https://v1-6.docs.kubernetes.io/docs/api-reference/v1.6/#labelselector-v1-meta) | false |
```
Atoms: here is written if empy only check own namespace, more down it's written empty will match all objects. so which is it ? :)

gmauleon: Seems to be the former:
https://github.com/mxinden/prometheus-operator/blob/89fc4e306972604eba2dcb961a6d29cc27a668ad/pkg/prometheus/rulefile.go#L103

gmauleon: Most likely to match the behavior of the ServiceMonitor one:
https://github.com/coreos/prometheus-operator/blob/f9676ddc5b55ae47dcdfa43d9290cb28e5504649/pkg/prometheus/promcfg.go#L468

Although @mxinden will confirm but in the RuleFile it seems to implement a standard selector which may offer the possibility to match all namespaces somehow.

Atoms: yes that makes sense :) thanks for explanation

PR #1334, opened 2018-05-11
contrib/kube-prometheus: add `-f` to `rm` in `build.sh`

mxinden:
According to the man pages of `rm` and the `-f` option:
> Attempt to remove the files without prompting for confirma- tion,
> regardless of the file's permissions.  If the file does not exist, do
> not display a diagnostic message or modify the exit status to reflect an
> error.  The -f option overrides any previous -i options.

This patch prevents the `build.sh` script from failing if files do not
exist when attempting to delete them.

### DIFF
--- a/contrib/kube-prometheus/build.sh
+++ b/contrib/kube-prometheus/build.sh
@@ -3,5 +3,5 @@ set -e
 set -x
 
                                                # optional, but we would like to generate yaml, not json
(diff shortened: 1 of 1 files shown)


brancz:
lgtm :+1:

brancz:
actually just seconds after I merged I realized, but can you just re-embed this into the readme in kube-prometheus? looks like our generation make target is not picking that up

PR #1338, opened 2018-05-11
prometheus: Allow specifying additional AlertManager configs via secret

CoufalJa:
Built upon the foundation of #1246, but this time for AlertManager configs. This functionality is important in case you have an external cluster of alert managers or you are working (as us) in a multi-cluster environment.


brancz:
Can you explain what you want this for and why the existing solutions are not sufficient?

CoufalJa:
Sure, I will try to explain with an example:
Imagine you have multiple K8s clusters with prometheus-operator and Prometheus itself deployed in each. Unlike Prometheus It does not make much sense have HA AlertManager deployed in each cluster as well. In order to connect Prometheus provisioned via prometheus-operator with AM you have to use K8s Service (endpoints) for service discovery. That is something you can´t use when connecting outside of the current K8s cluster. You can´t even use Service type ExternalName. The purpose of this PR is to enable external AM use-case. I originally wanted to just support external name services but then I realized that having an option to customize the configuration is much more powerful. Basically, it is the extension of external scrape configs idea.

brancz:
Nice, this generally lgtm. I'd like to have @mxinden give this a review as well just for additional perspective.

And just to make sure, @CoufalJa you did test this in your cluster I'm assuming? :slightly_smiling_face:

brancz:
Thanks for your effort @CoufalJa !

mxinden:
Cool addition. Thanks for the PR @CoufalJa!

hedeesa:
thanks!  is there any doc for how use this feature ? 
I wanna use additionalAlertmanagerConfigs but I dont know how exactly the manifest file should be changed

brancz:
@hedeesa we don't have user guides for this yet, only [API documentation](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#prometheusspec). The mechanism is quite similar to the `additionalAlertmanagerConfigs` feature though, for which we do have a [user guide](https://github.com/coreos/prometheus-operator/blob/master/Documentation/additional-scrape-config.md).

PR #1340, opened 2018-05-14
docs: Run embedmd on both Docs and kube-prometheus README.md

mxinden:
> actually just seconds after I merged I realized, but can you just re-embed this into the readme in kube-prometheus? looks like our generation make target is not picking that up

@brancz Follow up for https://github.com/coreos/prometheus-operator/pull/1334


brancz:
thanks! lgtm :+1:

PR #1345, opened 2018-05-15
Fix kube-prometheus chart not deploying when some features are disabled

lostick:
We encountered the same issue as [#1342](https://github.com/coreos/prometheus-operator/issues/1342).
When installing `kube-prometheus:0.0.66` without Node Exporter or Grafana, it fails as values are not set.


lostick:
Rebased.

gianrubio:
thanks @lostick

PR #1346, opened 2018-05-15
prometheus: Filter alerts from configmaps by cfg.Namespace

jmcfarlane:
Config maps in `configmaps.json` are keyed by namespace. This change
enables all namespaces (constrained by `cfg.Namespace`) to create alerts
(constrained by `ruleSelector`).

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -761,7 +761,7 @@ func (c *Operator) ruleFileConfigMaps(p *monitoringv1.Prometheus) ([]*v1.ConfigM
 		return nil, err
 	}
 
-	cache.ListAllByNamespace(c.cmapInf.GetIndexer(), p.Namespace, ruleSelector, func(obj interface{}) {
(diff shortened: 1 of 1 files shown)


brancz:
Thanks for your contribution, however, this doesn't do the entire trick and has unwanted semantics. As it should still be decidable by the Prometheus object which namespaces rules come from. Actually @mxinden is already working on a solution that solves the same problem as you're intending to solve. I suggest to follow #1333. This will land in the next release.

I will close this here.

jmcfarlane:
@brancz thanks for the response. For the purpose of iteration could you elaborate on the unwanted semantics this (naive) approach has? I ask because we might want to use this approach _now_, and migrate to the improved solution when it's ready (unless the details of the problems this PR introduces make that for sure a terrible idea :)

brancz:
The Prometheus Operator is used in multi tenant environments where a single Prometheus server is shared across multiple teams. In such a case it should not be possible for a single team to submit an invalid config causing the Prometheus server not to reload or worse, when redeployed the Prometheus server may not start in the first place.

jmcfarlane:
Got it, makes perfect sense. Thanks!

PR #1349, opened 2018-05-17
Adapt docs on additional rules and dashboards

brancz:
This PR changes the developing alerts and dashboards documentation to the new jsonnet based approach.

@mxinden @ant31


contrib/kube-prometheus/docs/developing-alerts-and-dashboards.md
```diff
@@ -2,40 +2,222 @@
...
+```
 
-To extend the rules simply add a new `.rules.yaml` file into the `assets/prometheus/rules/` directory and re-generate the manifests. To modify the existing rules, simply edit the respective `.rules.yaml` file and re-generate the manifest.
+### Recording rules
+
+In order to add a recording rule, simply do the same with the `prometheusRules` field.
+
+> Note that rules can just as well be included into this file, using the jsonnet `import` function. In this example it is just inlined in order to demonstrate their use in a single file.
+
+[embedmd]:# (../examples/prometheus-additional-alert-rule-example.jsonnet)
```
mxinden: I guess this should be "prometheus-additional-**recording**-rule-example.jsonnet" and not "prometheus-additional-alert-rule-example.jsonnet", right?

contrib/kube-prometheus/test.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
mxinden: Perfect! Very cool!

contrib/kube-prometheus/docs/developing-alerts-and-dashboards.md
```diff
@@ -2,40 +2,222 @@
```
mxinden: How about renaming this file to `developing-prometheus-rules-and-grafana-dashboards.md` as it includes the recoding rules docs as well?

PR #1360, opened 2018-05-18
prometheus chart: fix identation on serviceMonitorSelector

Serg1i:
Hello,

This one should fix broken yaml in case of using `.Values.prometheusLabelValue` in the prometheus helm chart


gianrubio:
Please bump kube-Prometheus/charts.yaml

gianrubio:
thanks @Serg1i

PR #1361, opened 2018-05-18
Make "Compiling" section easier to execute

luarx:
There is a section in the Makefile to do that instead of execute manually a script.

### DIFF
--- a/contrib/kube-prometheus/README.md
+++ b/contrib/kube-prometheus/README.md
@@ -120,6 +120,8 @@ jsonnet -J vendor -m manifests ${1-example.jsonnet} | xargs -I{} sh -c 'cat $1 |
 
 This script reads each key of the generated json and uses that as the file name, and writes the value of that key to that file.
 
+> You can also run this script executing the command `make generate-raw` from kube-prometheus base directory of this repository but the above option it is recommended so that you run it in your own infrastructure repository.
+
 ## Configuration
 
 A hidden `_config` field is located at the top level of the object this library provides. These are the available fields with their respective default values:


Renol7:
Yes, it could make sense! 
But, in my opinion, embedded scripts have the inconvenient that if the scripts is updated, sometimes people don't remember to update documentation too... 
There are cases where there is no other option that embedded an script, but in this case as exists a Makefile, it encapsulates it and if it is modified the user does not care :)

Despite this,I can adapt it to what you decide!

brancz:
The main point of all of this work, is that this should not be executed from this repository anymore, but instead from your own internal infrastructure repository. That's why intentionally the script is done extremely simple.

I'd prefer if the script stays in the foreground and `make generate-raw` is just a helper in a hint.

luarx:
Great! I have changed that @brancz :)

contrib/kube-prometheus/README.md
```diff
@@ -95,6 +95,8 @@ jsonnet -J vendor -m manifests ${1-example.jsonnet} | xargs -I{} sh -c 'cat $1 |
 
 This script reads each key of the generated json and uses that as the file name, and writes the value of that key to that file.
 
+> You can run also run this script executing the command `make generate-raw` from kube-prometheus base directory of this repository but it is recommended the above option so that you run it in your own infrastructure repository.
```
mxinden: `s/You can run also run/You can also run`

`s/it is recommended the above option/the above option is recommended`

luarx:
Yeah! I agree with you @mxinden! I have already changed that

mxinden:
@luarx sorry to bother you once more here. Could you rebase your changes onto current master. Our CI system doesn't do this before testing, thereby it fails with diffs in the generation step.

luarx:
Yes! There is no problem, I have already done it :)

mxinden:
Thanks @luarx.

PR #1366, opened 2018-05-21
Add instructions for new dashboards in the Grafana helm chart README

aranair:
It wasn't immediately clear how to add new dashboards. Probably took me longer than needed to figure it out so I thought an example or two in the README might help.


gianrubio:
@aranair thanks for your contribution! To fix CI you should bump charts version, please follow [this guide](https://github.com/coreos/prometheus-operator/tree/master/helm#how-to-contribue)

PR #1377, opened 2018-05-23
Remove references to datasources in README for ConfigMap generator

ryanwalls:
The generator script no longer handles datasources.  Updating the README to reflect that.


brancz:
Actually with the re-written jsonnet stack the entire configmap generator has become obsolete. We should remove it soon.

PR #1383, opened 2018-05-25
kube-prometheus: Update kops docs and add Ingress docs

brancz:
This PR ports the old kops documentation and adds docs on how to add ingress for Prometheus/Alertmanager/Grafana. The latter was tested on minikube and on my private cluster.

@mxinden @VinceMD


brancz:
I'm going to go ahead and merge this as there are some re-generate things in here that are stopping other PRs and I'd rather have docs that we can improve than none :slightly_smiling_face: .

VinceMD:
good doc. thanks

PR #1384, opened 2018-05-25
Make the Prometheus & Alertmanager name configurable

metalmatze:
The defaults are still with the same names as before. Prometheus is still called `k8s` and Alertmanager is still `main`. I call them both `metalmatze` in my cluster.
Already got it working. :blush:


metalmatze:
I don't think that the e2e tests fail because of me.

brancz:
Nope you're right, it's all good. (we have a flaky test that we're in the process of fixing)

This lgtm :+1: thanks!

PR #1386, opened 2018-05-25
Adds proxy_url to v1.Endpoint. Fixes #1385

### LINKED ISSUE #1385: Support proxy_url in Endpoint
`v1.types.Endpoint` does not map the Prometheus **proxy_url** option.
I use custom proxy servers to transform non-prometheus endpoints into prometheus scrape-able feeds, so the proxy_url option is super useful. Will submit PR for this.

nickman:
proxy_url support in v1.Endpoint would be super helpful.


pkg/client/monitoring/v1/types.go
```diff
@@ -367,6 +367,8 @@ type Endpoint struct {
 	BasicAuth *BasicAuth `json:"basicAuth,omitempty"`
 	// MetricRelabelConfigs to apply to samples before ingestion.
 	MetricRelabelConfigs []*RelabelConfig `json:"metricRelabelings,omitempty"`
+	//Optional ProxyURL
```
brancz: There is a space missing between `//` and the start of the comment, and this needs a more thorough description of what it does.

pkg/client/monitoring/v1/types.go
```diff
@@ -367,7 +367,7 @@ type Endpoint struct {
 	BasicAuth *BasicAuth `json:"basicAuth,omitempty"`
 	// MetricRelabelConfigs to apply to samples before ingestion.
 	MetricRelabelConfigs []*RelabelConfig `json:"metricRelabelings,omitempty"`
-	//Optional ProxyURL
+	// Optional ProxyURL e.g. http://proxyserver:2195. Directs scrapes to proxy through this endpoint
```
brancz: End sentence with a period. For it to be marked optional in the documentation this needs to be a pointer to a string (*string). Then you can also skip the “Optional” start in the comment.

pkg/prometheus/promcfg.go
```diff
@@ -233,6 +233,10 @@ func generateServiceMonitorConfig(version semver.Version, m *v1.ServiceMonitor,
 	if ep.Path != "" {
 		cfg = append(cfg, yaml.MapItem{Key: "metrics_path", Value: ep.Path})
 	}
+	if ep.ProxyURL != "" {
```
brancz: With a string pointer this will need to be changed to != nil.

brancz:
The build is failing, because some parts of the docs have not been updated, you need to run `make generate` once more since there were modifications.

nickman:
Got it. Fixed. Thanks again.

//Nicholas

brancz:
Perfect, all green now, thanks for your effort!

PR #1391, opened 2018-05-28
e2ePrometheusOnlyUpdatedOnRelevantChanges: Check for range

mxinden:
Instead of checking if a resource was updated exactly x times, check if
the amount of updates is within a range.

The test should prevent repeadet (>100) updates by the Prometheus
Operator. Kubernetes updates resources by itself, e.g. the status
fields. This amount stays within limits. Thereby checking if the
statefulset was updated between 1 and 3 times is enough.


mxinden:
This is a follow up for https://github.com/coreos/prometheus-operator/pull/1382

brancz:
lgtm, thanks for taking care of this! :+1:

PR #1393, opened 2018-05-28
security: Enforce nobody user and read only /

mxinden:
* Make the Prometheus Operator Docker image run as `nobody` by default.
* Disallow privilege escalation via K8s
* Enforce read only root filesystem

//CC @squeed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,4 +2,7 @@ FROM quay.io/prometheus/busybox:latest
 
 ADD operator /bin/operator
 
+# On busybox 'nobody' has uid `65534'
+USER 65534
+
 ENTRYPOINT ["/bin/operator"]
--- a/Documentation/user-guides/cluster-monitoring.md
+++ b/Documentation/user-guides/cluster-monitoring.md
@@ -59,6 +59,7 @@ spec:
       containers:
       - args:
         - --kubelet-service=kube-system/kubelet
+        - -logtostderr=true
         - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
         - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.22.0
         image: quay.io/coreos/prometheus-operator:v0.22.0
@@ -73,6 +74,9 @@ spec:
           requests:
             cpu: 100m
             memory: 50Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
       nodeSelector:
         beta.kubernetes.io/os: linux
       securityContext:
--- a/Documentation/user-guides/getting-started.md
+++ b/Documentation/user-guides/getting-started.md
@@ -115,6 +115,7 @@ spec:
       containers:
       - args:
         - --kubelet-service=kube-system/kubelet
+        - -logtostderr=true
         - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
         - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.22.0
         image: quay.io/coreos/prometheus-operator:v0.22.0
@@ -129,6 +130,9 @@ spec:
           requests:
             cpu: 100m
             memory: 50Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
       nodeSelector:
         beta.kubernetes.io/os: linux
       securityContext:
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -97,6 +97,7 @@ spec:
       containers:
       - args:
         - --kubelet-service=kube-system/kubelet
+        - -logtostderr=true
         - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
         - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.22.0
         image: quay.io/coreos/prometheus-operator:v0.22.0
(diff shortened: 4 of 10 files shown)


hack/generate/prometheus-operator.jsonnet
```diff
@@ -15,6 +15,8 @@ local operatorContainer =
   container.new("prometheus-operator", "quay.io/coreos/prometheus-operator:v" + version) +
   container.withPorts(containerPort.newNamed("http", targetPort)) +
   container.withArgs(["--kubelet-service=kube-system/kubelet", "--config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1"]) +
+  container.mixin.securityContext.withAllowPrivilegeEscalation(false) +
+  container.mixin.securityContext.withReadOnlyRootFilesystem(true) +
```
brancz: Does this collide with PSPs? Generally I'm fine with it, but if this causes the same troubles as specifying the Pod's security context, then I'd rather hold off.

brancz:
Generally lgtm, just want to make sure that these pods still create under a restrictive PSP.

Dockerfile
```diff
@@ -2,4 +2,6 @@ FROM quay.io/prometheus/busybox:latest
 
 ADD .build/linux-amd64/operator /bin/operator
 
+USER nobody
```
mrueg: This might cause problems due to https://github.com/kubernetes/kubernetes/pull/56503 You might want to specify a UID here instead.

hack/generate/prometheus-operator.jsonnet
mxinden: I am not quite sure I understand your comment. Could you elaborate further on this?

`AllowPrivilegeEscalation ` and `ReadOnlyRootFilesystem ` are only options in the container security context, not in the pod security context.

mxinden: Maybe the description of the `SecurityContext` struct helps:

```golang
// SecurityContext holds security configuration that will be applied to a container.
// Some fields are present in both SecurityContext and PodSecurityContext.  When both
// are set, the values in SecurityContext take precedence.
type SecurityContext struct {
```

Dockerfile
mxinden: Good point @mrueg, thanks!

@brancz What do you think?

hack/generate/prometheus-operator.jsonnet
brancz: Yes, I'm referring to the precedence taken here. If a PodSecurityPolicy sets something but the container's SecurityContext specifies something different, then does that prevent the Pod from being scheduled? Also are those not the default settings?

mxinden: Default of [`AllowPrivilegeEscalation`](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation) and [`ReadOnlyRootFilesystem`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#securitycontext-v1-core)  is `false` (see links).

mxinden: Default of [`AllowPrivilegeEscalation`](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation) is allow. Default of [`ReadOnlyRootFilesystem`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#securitycontext-v1-core)  is `false`.

mxinden:
@mrueg Looking into your [comment](https://github.com/coreos/prometheus-operator/pull/1393#discussion_r191226963) again in regards to using the uid instead of the username: The mentioned PR was merged for K8s v1.9.0. The Prometheus upstream images are using usernames instead of uids, which seems to be working fine.

Argument that speaks for using usernames is, that once we switch the base image which might have a different uid for the nobody user, we still end up with the nobody user.

Can you elaborate further why we should use uids over usernames?

Thanks for your help!

mxinden:
@brancz Any further comments? Otherwise I would move forward once the CI turned green.

brancz:
this needs a rebase otherwise lgtm

mrueg:
@mxinden sorry for getting back to this with this delay. My main concern is what was added to kubernetes in https://github.com/kubernetes/kubernetes/pull/56503 If you run it with strict pod security policies that disallow root, you'll likely run into issues. Currently that is not the case for vanilla prometheus-operator https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus-operator/templates/psp.yaml#L32 The kubernetes instance I work with runs the prometheus-operator with a stricter psp, that's why we moved to UIDs in docker images.

mxinden:
Prometheus Operator failing with:

```
W0716 17:07:34.077317       1 listers.go:63] can not retrieve list of objects using index : Index with name namespace does not exist
log: exiting because of error: log: cannot create log: open /tmp/operator.prometheus-operator-745df79676-dnw5s.unknownuser.log.WARNING.20180716-170734.1: read-only file system
```

due to `readOnlyRootFilesystem`.

brancz:
Is this maybe the a behavior of glog?

mxinden:
@brancz This is related to the Kubernetes client-go caching. I will check if this is configurable.

mxinden:
@brancz @squat added an empty dir mounted as `/tmp` fixing the `read-only file system` issue.

Would you mind taking another look?

brancz:
Could we investigate if we can configure glog with never attempting to write to those files? I'm also happy with a "hack" like [this](https://github.com/kubermatic/glog-gokit) as we are using go-kit/log elsewhere already as well.

mxinden:
> Could we investigate if we can configure glog with never attempting to write to those files? I'm also happy with a "hack" like this as we are using go-kit/log elsewhere already as well.

@brancz Good idea, removed the additional volume and made glog log to stderr instead of `/tmp`.

mxinden:
@brancz @squat can you give this another review?

brancz:
lgtm

PR #1395, opened 2018-05-29
Migrate alerts

brancz:
This PR migrates the majority of all alerts previously existing in kube-prometheus to jsonnet.

@mxinden @metalmatze 

As a quick check in, what's missing at this point is:

* The rest of the grafana dashboards to be migrated (specifically from https://github.com/brancz/kubernetes-grafana/commit/1d5427f5cfdd47549a0502f071320312acb20330). The blocker for this is for the Grafana jsonnet stack to put each dashboard into a separate ConfigMap (I hear @metalmatze is working on this).
* etcd rules/alerts/dashboard. These are in progress to be added to the etcd repository directly. Once that's done we will add etcd alerts (and dashboard) back here again as well, but vendored through jsonnet-bundler to stay up to date.

Those are two relatively large ones, but the end is in sight :slightly_smiling_face: .


contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/general.libsonnet
```diff
@@ -0,0 +1,34 @@
...
+              summary: 'Targets are down',
+            },
+            expr: '100 * (count(up == 0) BY (job) / count(up) BY (job)) > 10',
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+          },
+          {
+            alert: 'DeadMansSwitch',
```
metalmatze: We want to definitely have the `TargetDown`, but not necessarily the `DeadMansSwitch`. Can we separate these?

brancz: Hmm I would prefer to keep it. The kube-prometheus stack is a lot about our opinionated way of doing things due to our experience. And in our experience this is an invaluable tool that I cannot recommend highly enough. Aside from that the kube-prometheus stack by default routes it to /dev/null in the alertmanager, so is this really a problem?

I’m fine with having a way to filter it out though, but would prefer if that’s the explicit action rather than having to explicitly add it.

brancz: Moving forward with this for now, but more than happy to change things in follow ups.

jolson490:
Hello. I have some questions about the etcd stuff on the [original comment](https://github.com/coreos/prometheus-operator/pull/1395#issue-191086593) on this PR.

Is [this](https://github.com/coreos/etcd/pull/9640/files) the PR that added the etcd rules/alerts directly to the etcd repo?

May I please ask when will kube-prometheus out-of-the-box once again have etcd rules? 
 * When I deploy [v19](https://github.com/coreos/prometheus-operator/tree/v0.19.0) then [etcd3.rules.yaml](https://github.com/coreos/prometheus-operator/blob/v0.19.0/contrib/kube-prometheus/assets/prometheus/rules/etcd3.rules.yaml) is included in prometheus (e.g. on the `https://<my-prometheus>/rules` page).
 * But deploying [v21](https://github.com/coreos/prometheus-operator/tree/v0.21.0) doesn't setup any etcd rules in prometheus.

PR #1397, opened 2018-05-29
travis: Move e2e tests to first stage

mxinden:
If both the *generate* as well as the *e2e* tests on a patch fail, a
user does not need to wait for the second (fixed) iteration of the
*generate* tests to pass until the user gets the *e2e* failure feedback.


brancz:
:facepalm: ... the amount of time we have wasted on not doing this. Safe enough for me to me to merge considering the gain.

lgtm

PR #1403, opened 2018-05-31
Remove suffix '-prometheus'

greenwind89:
Currently the prometheus service does not have suffix '-prometheus'  but this template added the suffix therefore making Grafana unable to communicate the prometheus.


gianrubio:
lgtm on green

gianrubio:
Sorry, but #1413 was merged before this, anyway thanks for your PR @greenwind89

PR #1409, opened 2018-05-31
kube-prometheus: Remove unnecessary namespace selectors

brancz:
As ServiceMonitors by default select services from it's own namespace, these namespace selectors are unnecessary and in fact would prevent one from getting a functional cluster right now as they have hardcoded `monitoring` namespace in them.

This changes it to use whatever the ServiceMonitor's namespace is.

@mxinden @metalmatze


brancz:
as CI is green I'll move forward with this, as it's a dependency elsewhere

Issue #1412, opened 2018-06-01
Prometheus datasource URL is configured wrongly in graphana


sanderaernouts:
**What did you do?**
```
helm repo add coreos https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/
helm install coreos/prometheus-operator --name prometheus-operator --namespace monitoring --set rbacEnable=false
helm install coreos/kube-prometheus --name kube-prometheus --namespace monitoring --set global.rbacEnable=false
```
Login to Graphana dashboard using username/password

**What did you expect to see?**
Dashboards with data from Prometheus

**What did you see instead? Under which circumstances?**
No data. After inspecting the datasource I noticed this:
![image](https://user-images.githubusercontent.com/6184673/40834496-45c0a046-6591-11e8-9603-209e5e0c3a5a.png)

Whereas the service is named kube-prometheus instead:
![image](https://user-images.githubusercontent.com/6184673/40834540-6b78e014-6591-11e8-9ebb-35fdfe8d8b34.png)

**Environment**

* Kubernetes version information:

	iClient Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.2", GitCommit:"5fa2db2bd46ac79e5e00a4e6ed24191080aa463b", GitTreeState:"clean", BuildDate:"2018-01-18T10:09:24Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.6", GitCommit:"9f8ebd171479bec0ada837d7ee641dec2f8c6dd1", GitTreeState:"clean", BuildDate:"2018-03-21T15:13:31Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
* Kubernetes cluster kind:  
Azure AKS

cheethoe:
when will this be pushed to the coreos helm repo?  I still see 0.0.75 as the latest on the helm repo but this change is for 0.0.78 of kube-prometheus

gianrubio:
The CI is failing in a task before helm sync repo, it'll be sorted today. Sorry for any inconvenience

PR #1413, opened 2018-06-01
fixed prometheus datasource configuration in graphana configmap

sanderaernouts:
When installing the Helm charts the datasource for prometheus in graphana is not configured correctly, see issue #1412. Removed `-prometheus` from datasource URL in Helm chart


gianrubio:
lgtm on green

yurrriq:
This broke my usage of prometheus-operator as a subchart.

kyleterry:
Why was this done? This broke my release.

PR #1416, opened 2018-06-01
*: Update CRDs on boot

brancz:
This PR changes, that CRDs (including their validations) are always updated on boot of the Prometheus Operator.

@mxinden

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -38,6 +38,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


pkg/alertmanager/operator.go
```diff
@@ -534,29 +535,45 @@ func (c *Operator) destroyAlertmanager(key string) error {
...
+		if err != nil && !apierrors.IsNotFound(err) {
+			return errors.Wrapf(err, "getting CRD: %s", crd.Spec.Names.Kind)
+		}
+		if apierrors.IsNotFound(err) {
+			if _, err := crdClient.Create(crd); err != nil {
+				return errors.Wrapf(err, "creating CRD: %s", crd.Spec.Names.Kind)
+			}
+			level.Info(c.logger).Log("msg", "CRD created", "crd", crd.Spec.Names.Kind)
+		}
+		if err == nil {
```
mxinden: Unnecessary check, but I guess improves code readability.

brancz:
I'm not sure an e2e test is appropriate at this point, the thing I would want to test and make sure is that across versions updates work, a simple update is not the regression we're going to have here (not repeatedly at least).

There is a more general e2e test that I would like us to have that would include this: upgrade tests. What do you think of creating an issue for upgrade tests and move forward with this for now?

PR #1423, opened 2018-06-04
doc: Update alerting documentation to rule file crd

mxinden:
Tested on minikube.


brancz:
this needs a `make generate` otherwise lgtm

PR #1424, opened 2018-06-04
Fix conversion of Prometheus timestamp fields

ironcladlou:
Fix a conversion issue which prevents updates to Prometheus resource timestamp
fields where the timestamp is a.) a value type and b.) a read-only API field
(i.e. creationTimestamp).

### DIFF
--- a/pkg/client/monitoring/v1/alertmanager.go
+++ b/pkg/client/monitoring/v1/alertmanager.go
@@ -177,6 +177,20 @@ func UnstructuredFromAlertmanager(a *Alertmanager) (*unstructured.Unstructured, 
 	if err := json.Unmarshal(b, &r.Object); err != nil {
 		return nil, err
 	}
+	// Value-type timestamp fields like ObjectMeta.CreationTimestamp with a zero
+	// value are marshalled as "null" in JSON (rather than omitted) and then
+	// unmarshalled into Unstructured with the key intact and a null value (rather
+	// than being omitted); the net effect is the resulting structs can't be used
+	// to issue a POST because creationTimestamp=null is sent to the server and
+	// fails validation. For example, passing an Alertmanager with a
+	// volumeClaimTemplate can result in an invalid object. This hack simply
+	// removes such timestamp fields manually.
+	//
+	// TODO: reevaluate the use of Unstructured directly here in the context of
+	// the latest dynamic client capabilities; this manual conversion may not be
+	// necessary anymore.
+	unstructured.RemoveNestedField(r.Object, "metadata", "creationTimestamp")
+	unstructured.RemoveNestedField(r.Object, "spec", "storage", "volumeClaimTemplate", "metadata", "creationTimestamp")
 	return &r, nil
 }
 
--- a/pkg/client/monitoring/v1/prometheus.go
+++ b/pkg/client/monitoring/v1/prometheus.go
@@ -176,6 +176,20 @@ func UnstructuredFromPrometheus(p *Prometheus) (*unstructured.Unstructured, erro
 	if err := json.Unmarshal(b, &r.Object); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


pkg/client/monitoring/v1/prometheus_test.go
```diff
@@ -0,0 +1,78 @@
...
+						},
+						"status": map[string]interface{}{},
+					},
+				},
+			},
+		},
+	}
+
+	if e, a := expected.Object, actual.Object; !reflect.DeepEqual(e, a) {
+		t.Fatalf("expected:\n%v\ngot:\n%v", e, a)
```
mxinden: Instead of printing the structs you could use [github.com/kylelemons/godebug/pretty](https://github.com/coreos/prometheus-operator/blob/master/test/e2e/prometheus_test.go#L41) which we have vendored for testing code. Gives you a nice diff (See [example usage](https://github.com/coreos/prometheus-operator/blob/master/test/e2e/prometheus_test.go#L800)).

pkg/client/monitoring/v1/prometheus.go
```diff
@@ -176,6 +176,15 @@ func UnstructuredFromPrometheus(p *Prometheus) (*unstructured.Unstructured, erro
 	if err := json.Unmarshal(b, &r.Object); err != nil {
 		return nil, err
 	}
+	// Value-type timestamp fields like ObjectMeta.CreationTimestamp with a zero
+	// value are marshalled as "null" in JSON (rather than omitted) and then
+	// unmarshalled into Unstructured with the key intact and a null value (rather
+	// than being omitted); the net effect is the resulting structs can't be used
+	// to issue a POST because creationTimestamp=null is sent to the server and
+	// fails validation. This hack simply removes the field manually. Need to
```
mxinden: "Need to [...]": Is that a TODO?

pkg/client/monitoring/v1/prometheus_test.go
ironcladlou: Done, much nicer. Thanks!

pkg/client/monitoring/v1/prometheus.go
ironcladlou: Refactored comment

ironcladlou:
@mxinden 
>To fully understand the initial failure: The Prometheus operator fails to POST a Prometheus CRD struct to the Kubernetes API server?

Actually the place this manifested is in the cluster-monitoring-operator which attempts to update a `Prometheus` using the vendored prometheus-operator client; when specifying a volumeClaimTemplate the update can never succeed because `creationTimestamp` is always set to `null` by the client, causing a rejection by the apiserver.

ironcladlou:
Addressed comments and squashed

ironcladlou:
Hmm, I'll bet Alertmanager has the same issue

ironcladlou:
Alertmanager indeed suffers from the same issue, verified with a copypasta-style unit test. Added a corresponding fix with a cross reference there too.

pkg/client/monitoring/v1/alertmanager.go
```diff
@@ -177,6 +177,9 @@ func UnstructuredFromAlertmanager(a *Alertmanager) (*unstructured.Unstructured,
 	if err := json.Unmarshal(b, &r.Object); err != nil {
 		return nil, err
 	}
+	// TODO: See the corresponding comment in UnstructuredFromPrometheus.
```
mxinden: Would you mind adding the explanation comment here as well? I would prefer duplicate comments over comment referencing. Feel free to still keep a reference like `See 'unstructured.RemoveNestedField' in prometheus.go` here as well, so if we delete one, we keep in mind to delete the other as well.

ironcladlou: Done, and squashed

PR #1426, opened 2018-06-04
helm/grafana: Allow to set custom Service labels

NicolasT:
See: e131164c428a4a6ed3c2b60efb6bc627f3af95dd


NicolasT:
Good point @gianrubio, done now.

PR #1430, opened 2018-06-05
*: Rename RuleFile to PrometheusRule

### LINKED ISSUE #1425: Rename RuleFile to PrometheusRule
Generic CRD names should be avoided. `RuleFile` could be anything and used by other projects and it's not self-documented. 

We can rename it to `PrometheusRule`, which is more specific.

brancz:
As discussed in #1425 this renames the `RuleFile` CRD to `PrometheusRule`.

(mostly opening for CI right now)

@mxinden @ant31 

Closes #1425


mxinden:
@brancz Could you address your [comment](https://github.com/coreos/prometheus-operator/pull/1429#discussion_r193023052) directly here?

brancz:
Already done :slightly_smiling_face:

Once all tests pass I will remove the `prometheusRuleSelector`, so only the `ruleSelector` is left.

mxinden:
@brancz Thanks. Let me know once this is ready for review.

brancz:
@mxinden Ready for review! :slightly_smiling_face: Let's get this release out.

pkg/prometheus/rules.go
```diff
@@ -207,6 +200,6 @@ func checksumRuleFiles(files map[string]string) string {
 	return fmt.Sprintf("%x", sha256.Sum256([]byte(sum)))
 }
 
-func prometheusRuleFilesConfigMapName(prometheusName string) string {
+func prometheusRuleConfigMapName(prometheusName string) string {
 	return "prometheus-" + prometheusName + "-rulefiles"
```
mxinden: Do we want to keep `-rulefiles` here? Could be just `rules` instead?

mxinden: I am also fine leaving it like it is, keeping it as an internal implementation detail.

brancz: I chose to stick with `-rulefiles` because it's very common (even in kube-prometheus this is the case) that people have `<prometheus-name>-rules` as their configmap name today, in which case it will get overridden, so I chose not to and in this case `-rulefiles` is ok to be specific I think.

PR #1432, opened 2018-06-05
prometheus-config-reloader: Hardcode version

mxinden:
The prometheus-config-reloader will be released alongside the
prometheus-operator. Thereby it holds the same version.

For now we will manually hardcode that version. In the future this could
be achieved dynamically via a compiler flags.


brancz:
lgtm on green

PR #1433, opened 2018-06-05
NodeDiskRunningFull ignore etc files mountpoints

atopuzov:
When NodeDiskRunningFull is alarmed 4 alerts are generated. 3 of those are for the following mountpoints: `/etc/resolv.conf`, `/etc/hosts`, `/etc/hostname` and the last one is for the `/` mountpoint.
A good example is https://github.com/coreos/prometheus-operator/issues/1180#issuecomment-382799514

N.B. Seems the automation to update the helm charts is broken /cc @gianrubio so I'm unsure if I should update them manually.

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/node.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/node.libsonnet
@@ -11,7 +11,7 @@
               summary: 'Node disk is running full within 24 hours',
             },
             expr: |||
-              predict_linear(node_filesystem_free{%(nodeExporterSelector)s}[6h], 3600 * 24) < 0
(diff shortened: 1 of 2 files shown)


brancz:
Thanks for the contribution! This lgtm. We'll merge once CI is green. Thanks a lot!

gianrubio:
@atopuzov thanks for let me know, I'm gonna fix the helm sync later

brancz:
Sorry, I lost track of this, CI was green so I merged! :tada:

stevenpall:
Just wondering if this could be updated in the Helm chart as well? I've manually updated it for now. I'm running into similar issues as described [here](https://github.com/coreos/prometheus-operator/issues/1180#issuecomment-382799514) with overzealous (and duplicate) alerts being triggered.

PR #1434, opened 2018-06-05
Remove `description` from the root schema in CRDs

### LINKED ISSUE #1293: k8s 1.10 with CustomResourceSubresources=true prevents operator from creating CRDs
**What did you do?**

Use prometheus-operator on a kubernetes 1.10 cluster with the feature gate CustomResourceSubresources set to true

Use the bundle yaml from the root of the repo: https://github.com/coreos/prometheus-operator/blob/master/bundle.yaml

Operator/Controller is able to create CRDs

The operator fails to create CRDs - see logs below
https://gist.github.com/bsnchan/1f721c6e9f7aa989cb0c070ed305bd02

nikhita:
Fixes  #1293

`description` is not allowed in the root schema in CRDs in Kubernetes v1.10 if subresources are enabled. They are allowed from Kubernetes v1.11.

/assign @brancz @sslavic

### DIFF
--- a/contrib/kube-prometheus/manifests/0prometheus-operator-0alertmanagerCustomResourceDefinition.yaml
+++ b/contrib/kube-prometheus/manifests/0prometheus-operator-0alertmanagerCustomResourceDefinition.yaml
@@ -11,7 +11,7 @@ spec:
   scope: Namespaced
   validation:
     openAPIV3Schema:
-      description: Alertmanager describes an Alertmanager cluster.
+      # description: Alertmanager describes an Alertmanager cluster.
       properties:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
--- a/contrib/kube-prometheus/manifests/0prometheus-operator-0prometheusCustomResourceDefinition.yaml
+++ b/contrib/kube-prometheus/manifests/0prometheus-operator-0prometheusCustomResourceDefinition.yaml
@@ -11,7 +11,7 @@ spec:
   scope: Namespaced
   validation:
     openAPIV3Schema:
-      description: Prometheus defines a Prometheus deployment.
+      # description: Prometheus defines a Prometheus deployment.
       properties:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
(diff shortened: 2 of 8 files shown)


contrib/kube-prometheus/manifests/0prometheus-operator-0alertmanagerCustomResourceDefinition.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   scope: Namespaced
   validation:
     openAPIV3Schema:
-      description: Alertmanager describes an Alertmanager cluster.
+      # description: Alertmanager describes an Alertmanager cluster.
```
nikhita: I commented it out in case it needs to be added again.

nikhita: Can remove it as well, let me know what you prefer. :)

nikhita:
Even after removing this field, if this yaml configuration is used against a 1.10 cluster with subresources enabled, it will still fail validation because `required` is in the root. 1.10 only allows `properties` in the root. 

1.11 allows `properties`, `required` and `description` in the root.

contrib/kube-prometheus/manifests/0prometheus-operator-0alertmanagerCustomResourceDefinition.yaml
mxinden: @nikhita All of these are auto-generated. The source is [here](https://github.com/nikhita/prometheus-operator/blob/1c2f14c2b216fd0ecfa9e9823466f2e133ce0aa1/pkg/client/monitoring/v1/types.go#L457).

I guess removing the comment and running `make generate` would do the job.

Thanks a lot for the help!

brancz:
These are actually generated through `cmd/po-crdgen`, so we should do it there, and in the code where the operator registers the CRDs itself (we give users both options, either manage CRDs yourself or have them be managed by the operator).

nikhita:
@brancz @mxinden thank you, I wasn't aware of that!

I'll get to this tomorrow. It's quite late here and I'm supposed to be on a vacation. :)

brancz:
> I'll get to this tomorrow. It's quite late here and I'm supposed to be on a vacation. :)

No worries at all, happy to do a patch release for this once you're back from vacation :slightly_smiling_face: .

nikhita:
Sorry, this somehow got off my radar.

Looking at the original issue, this seems to be fixed. Is this PR still needed or is it safe to be closed?

brancz:
Yes sorry for dropping the ball @nikhita. Yes this is fixed, so let's close this :slightly_smiling_face: .

PR #1437, opened 2018-06-05
Make component selectorLabel for kube-controller-manager and kube-scheduler overridable

hairmare:
**What did you do?**

I followed the TLDR on several platforms.

Noice graphs in Grafana.

A failed deploy as soon as I got into the whole *self-hosted historically labels the kube-controller-manager & kube-scheduler differently than kubeadm* debacle.

**Environment**

Quite some of them... IMO this affects two use cases
* deploying prom to a cluster where we don't manage the cluster itself but do want to provide a managed prom
* the out of the box experience for users on non *component label scheme* (?) compliant platforms

This aims to fix an issue that affects quite a-lot of kinds of clusters. One of the numerous bad "new user experiences that results in an issue" I found regarding this is https://github.com/coreos/prometheus-operator/issues/618 (it's not related to the charts specifically but there are others).

I'm not touching anything except kube-scheduler & kube-controller-manager since the community seems to have standardised on the `component` label vs. the `k8s-app` label & it only really affects those two components at large.

I think https://github.com/kubernetes/community/issues/300 comes close to addressing the underlying issue when it comes to standardising the the label of (loosely) *kube-system* namespaced components.

Also, since adding new features technically qualifies as a feature I bumped the MINOR version on the charts, I hope you don't mind. SemVer doesn't really care about 0.x versions anyhow.

I'm aware that this is just a quick hack for a quite specific label issue. Personally I think we should standardise on the `component` label scheme and I didn't implement support for overriding this this in the other exporters since I haven't seen them using weird labels in the wild yet.

Let me know if you'd rather have something akin to the new*ish* `serviceMonitorNamespaceSelector` thing in some of the templates and I'll be happy to refactor.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -6,6 +6,8 @@ helm install coreos/prometheus-operator --name prometheus-operator --namespace m
 helm install coreos/kube-prometheus --name kube-prometheus --namespace monitoring
 ````
 
+If you are on a platform that uses the `k8s-app` labeling scheme for the scheduler and controller manager you can `--set exporter-node.selectorComponentMatchKey=k8s-app,exporter-kube-scheduler.selectorComponentMatchKey=k8s-app` to get started.
+
 # How to contribue?
 
 1. Fork the project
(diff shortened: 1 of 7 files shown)


metalmatze:

PR #1439, opened 2018-06-06
Add apiserver config of kubernetes_sd_config to prometheus spec.

whypro:
Add feature: https://github.com/coreos/prometheus-operator/issues/1372


pkg/prometheus/promcfg.go
```diff
@@ -216,12 +218,12 @@ func generateServiceMonitorConfig(version semver.Version, m *v1.ServiceMonitor,
 	switch version.Major {
 	case 1:
 		if version.Minor < 7 {
-			cfg = append(cfg, k8sSDAllNamespaces())
+			cfg = append(cfg, generateK8SSDConfig(nil, nil, nil))
```
mxinden: Am I right that if a user defines a custom API server config, and chooses to use Prometheus < 1.7 that user would not see any log messages, saying that the configuration is ignored? If so, could you add a log line here to inform the user?

whypro: You're right. This configuration is ignored when using Prometheus < 1.7. I will add log messages for this.

pkg/prometheus/operator.go
```diff
@@ -949,7 +949,7 @@ func loadBasicAuthSecret(basicAuth *monitoringv1.BasicAuth, s *v1.SecretList) (B
 
 }
 
-func (c *Operator) loadBasicAuthSecrets(mons map[string]*monitoringv1.ServiceMonitor, remoteReads []monitoringv1.RemoteReadSpec, remoteWrites []monitoringv1.RemoteWriteSpec, s *v1.SecretList) (map[string]BasicAuthCredentials, error) {
+func (c *Operator) loadBasicAuthSecrets(mons map[string]*monitoringv1.ServiceMonitor, remoteReads []monitoringv1.RemoteReadSpec, remoteWrites []monitoringv1.RemoteWriteSpec, apiserverConfig *monitoringv1.APIServerConfig, s *v1.SecretList) (map[string]BasicAuthCredentials, error) {
```
brancz: There is a pull request pending that fixes some behavior here, we should make sure that this still works as expected for both cases. Could you coordinate across the PRs @mxinden ?

brancz: Cross referencing: https://github.com/coreos/prometheus-operator/pull/1550

pkg/prometheus/operator.go
```diff
@@ -982,13 +982,24 @@ func (c *Operator) loadBasicAuthSecrets(mons map[string]*monitoringv1.ServiceMon
...
+				return nil, fmt.Errorf("could not generate basicAuth for remote_write config %d. %s", i, err)
 			} else {
 				secrets[fmt.Sprintf("remoteWrite/%d", i)] = credentials
 			}
 		}
 	}
 
+	// load apiserver basic auth secret
+	if apiserverConfig != nil && apiserverConfig.BasicAuth != nil {
+		if apiserverConfig.BasicAuth != nil {
```
mxinden: Haven't you already checked this in the line above in the second expression of the if clause?

pkg/prometheus/promcfg.go
```diff
@@ -216,12 +219,15 @@ func generateServiceMonitorConfig(version semver.Version, m *v1.ServiceMonitor,
 	switch version.Major {
 	case 1:
 		if version.Minor < 7 {
-			cfg = append(cfg, k8sSDAllNamespaces())
+			if apiserverConfig != nil {
+				log.Printf("custom apiserver config is set but it will not take effect if prometheus version is < 1.7")
```
mxinden: We are not using the standard logger anymore, but instead use [leveled logging](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/operator.go#L61).

For example: https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/operator.go#L731-L733

whypro: Thank you for reply. But I found that functions in `prom_cfg.go` are all "static". They all have no logger object like the one in operator object. I can't simply add a line for log it. I wonder how to use this leveled logging package without create a local logger object nor change these static functions to operator's method. What do you think?

brancz: We should bind these methods to a configGenerator struct that is instantiated in the reconcile loop and that gets the logger passed. Then the bound methods can just use the logger from the struct they’re bound to.

whypro: OK, Shuold I do this in current PR, or create another one?

pkg/prometheus/operator.go
whypro: Sorry, It may be a mistake.

pkg/prometheus/promcfg.go
whypro: Done. cc @brancz

brancz:
There are still failures in the auto generation build. Otherwise this lgtm.

schmitch:
as far as I understand with that pr it would be possible to have 1 k8s cluster with prometheus-operator that could monitor other k8s clusters?

brancz:
@schmitch although that wasn't the use case of the author, theoretically yes, but you still need to ensure some things to be available for that to work:

* a form of authentication is available (bearer token, client TLS certificate, basic auth)
* network access to the nodes and pods of the other cluster

schmitch:
what exactly means network access to the nodes and pods of the other cluster?
wouldn't it be enough to actually have access to the apiserver of the other cluster, I mean most stuff, services/etc. could be proxied through it?

brancz:
What I mean by network access is that the Prometheus server can ping the other Pod. Yes there are of course esoteric ways of proxying traffic, but those are things that I recommend against as in whitebox monitoring you want to have as little indirections between the Prometheus server and the target. Ideally the Prometheus server is also as close to the target as possible.

pkg/prometheus/operator.go
huydinhle: update cross referencing: https://github.com/coreos/prometheus-operator/pull/1619

whypro:
I've fix all build failures of travis ci. Could you take a look again? Thank you very much. @brancz

mxinden:
> as far as I understand with that pr it would be possible to have 1 k8s cluster with prometheus-operator that could monitor other k8s clusters?

@schmitch with this PR you can only configure the Prometheus Operator with one API server. Thereby, yes, it could watch for Service Monitors in the other cluster, but it would also spin up the Prometheus instances in the other cluster. Thereby management (Prometheus Operator) would be in your first cluster and targets and Prometheus would be in the other. In that case the Prometheus Operator would only need access to the master node, not all worker nodes.

whypro:
@mxinden OK. It doesn't matter. I've resolved conflicts.

PR #1443, opened 2018-06-06
kube-prometheus: Extract prometheus-operator jsonnet out of contrib

brancz:
After this PR is merged, it will need a follow up changing the versions to use to be from `master`, since this "package" is not on master (yet).

@mxinden


brancz:
Moving forward with this, as it fixes CI.

PR #1444, opened 2018-06-06
Kube-prometheus: add additional labels for rules configMap

mikeraimondi:
Allow additional labels to be added to the rule configMap, as done in most of the other Helm charts which include rules.


helm/kube-prometheus/values.yaml
```diff
@@ -431,3 +431,7 @@ prometheus:
 # Select Deployed DNS Solution
 deployCoreDNS: false
 deployKubeDNS: true
+
+##Custom Labels to be added to Prometheus Rules ConfigMap
+##
+additionalRulesConfigMapLabels: {}
```
gianrubio: please add an example  explaining how to use this label

mikeraimondi: I added an example. I'm not positive this is what you're looking for, since none of the other Helm charts have an example for this value.

gianrubio:
lgtm on green

mikeraimondi:
PR should be all set. Let me know if I can do anything to get this merged in!

gianrubio:
lgtm on green

PR #1447, opened 2018-06-07
[helm] use the right kube-etcd label

gianrubio:
- Fix helm etcd alerts using the right kube-etcd label 
- ~~This PR also fix the broken CI running `make generate`~~


gianrubio:
travis CI canceled the test but it all green on the test page https://travis-ci.org/coreos/prometheus-operator/builds/389141722

PR #1449, opened 2018-06-07
*: Use all jsonnet from master branch

brancz:
This is a follow up to #1443, changing the version of the prometheus-operator jsonnet package to the now existent `master` version.

@mxinden


brancz:
Moving forward to not block other CI runs.

PR #1450, opened 2018-06-07
kube-prometheus: Add troubleshooting and alertmanager docs

brancz:
@mxinden


mxinden:
Cool! Thanks for the improvement @brancz!

PR #1455, opened 2018-06-08
Support HostPath Volume for Prometheus Storage

opsnull:


brancz:
Thanks for opening this PR!

I'm not sure hostpath is a good idea for a Prometheus server, it's mainly meant for daemons that need actual access to the host filesystem/mount-namespace. What speaks against using local volumes? You get the same I/O characteristics but none of the usability and security problems.

AllaeddineEL:
It will be very useful for picking up secret files e.g. certificates or keys form the host file system, such as in the case of etcd client TLS creds when monitoring etcd.

szediktam:
How about this pr? @opsnull 

Does prometheus-operator support hostpath as a volume resource now ? @brancz

zgfh:
I find a way to use hostpath,no need change code
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: manual
provisioner: manual
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: prometheus-ptest-db-prometheus-ptest-0
  namespace: monitoring
spec:
  accessModes:
  - ReadWriteOnce
  dataSource: null
  resources:
    requests:
      storage: 10Gi
  storageClassName: manual
  volumeMode: Filesystem
---
kind: PersistentVolume
apiVersion: v1
metadata:
  name: prometheus-ptest-db-prometheus-ptest-0
  labels:
    type: local
spec:
  storageClassName: manual
  capacity:
    storage: 10Gi
  accessModes:
    - ReadWriteOnce
  hostPath:
    path: "/data/p8stest"
```
you create path "/data/p8stest" on host
then,create promethues
```
kind: Prometheus
metadata:
  labels:
    prometheus: p8s
  name: ptest
  namespace: monitoring
spec:
  baseImage: harbor.geniusafc.com/quay.io/prometheus/prometheus
  version: v2.5.0
  storage:
    volumeClaimTemplate:
      spec:
        storageClassName: manual
```

zgfh:
@metalmatze  if we can support hostpath ,it may be easy to use (many producted k8s cluster version< 1.14, it not easy to use local volume)

varunsrinath:
Hi Guys

varunsrinath:
for MTLS how do we setup volumes to the host paths where my certificates are located.

PR #1462, opened 2018-06-11
*: Refactor build system

mxinden:
- Move prometheus-config-reloader to cmd/
- Refactor Makefile & contrib/kube-prometheus/Makefile
- Only execute a target if its dependencies changed
  - Create empty target file for docker builds
- Replace promu with plain static `go build`


brancz:
Nice! This has been long long long overdue. This looks fantastic! :+1: lgtm

PR #1466, opened 2018-06-12
kube-prometheus: Use 1 ConfigMap per Grafana dashboard

brancz:
This updates the Grafana stack, most notably, that all dashboards are now in their own ConfigMap.

@mxinden @metalmatze


mxinden:
Restarted CI due to skipped deploy stage failure.

Issue #1435, opened 2018-06-05
Prometheus tagets not being updated after creating a ServiceMonitor


smileisak:
**What did you do?**
Trying to monitor kubernetes service using ServiceMonitor CRD.

deployment.yaml

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: hash-app
spec:
  replicas: 3
  template:
    metadata:
      labels:
        app: hash-app
    spec:
      containers:
      - name: hash-app
        image: smile0x90/hashme
        ports:
        - name: web
          containerPort: 8080
```
service.yaml
```yaml
kind: Service
apiVersion: v1
metadata:
  name: hash-app
  labels:
    app: hash-app
spec:
  selector:
    app: hash-app
  ports:
  - name: web
    port: 8080

```
service-monitor.yaml
```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: hash-app
  labels:
    team: frontend
spec:
  selector:
    matchLabels:
      app: hash-app
  endpoints:
  - port: web
```
`hashme` is go application that exposes `/hash` and `/metrics`

**What did you expect to see?**

Targets should be updated with new created ServiceMonitor in prometheus interface.

**What did you see instead? Under which circumstances?**

No new target created Target 

**Environment**

* Kubernetes version information:
```Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.2", GitCommit:"5fa2db2bd46ac79e5e00a4e6ed24191080aa463b", GitTreeState:"clean", BuildDate:"2018-01-18T10:09:24Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.7", GitCommit:"8e1552342355496b62754e61ad5f802a0f3f1fa7", GitTreeState:"clean", BuildDate:"2017-09-28T23:56:03Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: 

	Azure ACS Service

* Prometheus Operator Logs:

```
ts=2018-06-05T14:46:31.028895466Z caller=operator.go:671 component=prometheusoperator msg="sync prometheus" key=monitoring/kube-prometheus
[...]

PR #1468, opened 2018-06-12
Documentation Update

smileisak:
This PR adds in the documentation the fact the ServiceMonitors must be installed in the same namespace as Prometheus Operator.

Related to this issue: https://github.com/coreos/prometheus-operator/issues/1435

### DIFF
--- a/Documentation/user-guides/getting-started.md
+++ b/Documentation/user-guides/getting-started.md
@@ -202,6 +202,8 @@ spec:
 
 This Service object is discovered by a ServiceMonitor, which selects in the same way. The `app` label must have the value `example-app`.
 
+
+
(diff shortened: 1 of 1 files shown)


mxinden:
Cool, thanks for following up on this.

I would suggest moving this up to line 160.

For example:

```markdown
The Prometheus resource includes a field called `serviceMonitorSelector`, which defines a selection of ServiceMonitors to be used. By default and before the version `v0.19.0`, ServiceMonitors must be installed in the same namespace as the Prometheus instance. With the Prometheus Operator `v0.19.0` and above, ServiceMonitors can be selected outside the Prometheus namespace via the `serviceMonitorNamespaceSelector` field of the Prometheus resource.
```

What do you think?

mxinden:
Closes #1435

PR #1469, opened 2018-06-12
pkg/version: Pass Prometheus Operator version via ldflags

mxinden:
The Prometheus config reloader image is released along with the
Prometheus Operator image, tagged with the same semver version. In order
to default to the Prometheus Operator version if no Prometheus config
reloader image is specified, the content of `VERSION` needs to be passed
during compile time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,10 +27,14 @@ all: format generate build test
 build: operator prometheus-config-reloader
 
 operator: $(GOLANG_FILES)
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o $@ cmd/operator/main.go
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build \
+	-ldflags "-X github.com/coreos/prometheus-operator/pkg/version.Version=$(shell cat VERSION)" \
+	-o $@ cmd/operator/main.go
 
 prometheus-config-reloader:
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o $@ cmd/$@/main.go
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -27,7 +27,9 @@ all: format generate build test
 build: operator prometheus-config-reloader
 
 operator: $(GOLANG_FILES)
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o $@ cmd/operator/main.go
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build \
+	-ldflags "-X github.com/coreos/prometheus-operator/pkg/version.Version=$(shell cat VERSION)" \
```
brancz: I think it would be a good idea to do the same for the config reloader.

cmd/operator/main.go
```diff
@@ -106,6 +111,8 @@ func Main() int {
 	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
 	logger = log.With(logger, "caller", log.DefaultCaller)
 
+	logger.Log("msg", fmt.Sprintf("Starting Prometheus Operator version '%v'.", version.Version))
```
brancz: yay, finally :tada:

Makefile
mxinden: Do I understand correctly, that you want to release the config-reloader along with the Prometheus Operator and the Prometheus Config Reloader? Is that necessary, given the fact, that it never changed so far?

brancz: Sorry I meant that it would be a good idea for the prometheus-config-reloader as well, so we can tell the version from the logs people share with us.

mxinden: Good idea. Adjusted accordingly.

mxinden:
Any other thoughts on this @brancz?

brancz:
lgtm :+1: nice!

PR #1475, opened 2018-06-13
operator: Support json as a log format

mrueg:
It would be nice if all moving parts of the prometheus operator would support json-formatted structured logging.

Providing this as an early stage PR to gather feedback on it.

For upstream prometheus, there's a PR available here: https://github.com/prometheus/common/pull/136/files

* [x] prometheus-config-reloader needs an option as well

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -49,6 +49,11 @@ const (
 	logLevelNone  = "none"
 )
 
+const (
+	logFormatLogfmt = "logfmt"
+	logFormatJson   = "json"
+)
+
 var (
 	cfg                prometheuscontroller.Config
 	availableLogLevels = []string{
@@ -58,6 +63,10 @@ var (
 		logLevelWarn,
 		logLevelError,
 		logLevelNone,
+	}
+	availableLogFormats = []string{
+		logFormatLogfmt,
+		logFormatJson,
 	}
 )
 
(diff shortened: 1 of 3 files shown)


brancz:
Out of curiosity, why would json be a better option than logfmt? logfmt is much more efficiently parsed and ingested by systems like fluentd+elasticsearch, if that's what you're going for.

mrueg:
I agree on the efficiency part, the team I work in agreed on structured json-formatted logging as our standard log output format.

brancz:
I guess json is a reasonable format, but I'd like to keep the choices to a minimum here and not extend this list beyond logfmt and json. Let's add it for the prometheus config reloader as well as you mentioned.

brancz:
> Does it makes sense to unify the CLI frameworks (kingpin, urfave/cli) for operator and config-reloader? Which one is preferred?

Generally I'm all for consistency, but I'd be careful with potential breaking changes. Let's do that in a separate PR.

> Also, do we need a prometheus-config-reloader-args argument in prometheus operator that will configure the config-reloader with the right arguments?

I'd say no, the prometheus-config-reloader is an implementation detail of the Prometheus Operator, we should just configure the prometheus-config-reloader however the prometheus-operator is configured - luckily we control both.

mrueg:
I don't really get why CI is failing here. Anyone can help me out?

brancz:
You can run the tests on a local minikube by building the image with `make image` and then running `make test-e2e`, you can debug further when the tests are running and figure out what's wrong. This looks like Prometheus is not starting up properly for some reason, so I'd be looking for crashlooping or erroring pods in the tests.

mrueg:
@brancz looks like rebasing on latest master fixed the build.

PR #1476, opened 2018-06-13
Added nvme device pattern to disk io rule

dimm0:


brancz:
Thanks for the contribution! We develop, then vendor and use these rules from the [kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin) repository, so we'll need to have these contributions there first.

The changes just need to be done here: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/297f40ffc7dc07b690bf715f3952a8b729ba9197/rules/rules.libsonnet#L250-L285

Once merged there CI will pass here as well :slightly_smiling_face: .

PR #1484, opened 2018-06-14
kube-prometheus: Run node-exporter in host network

mxinden:
Node exporter needs to run in the host network, not in the pod network
in order to pick up network metrics of the node.

Fixes https://github.com/coreos/prometheus-operator/issues/1473


brancz:
lgtm on

PR #1485, opened 2018-06-14
Add sidecar containers field to Prometheus Helm chart

sdurrheimer:
Hi there,

I've been working on using Thanos with the kube-prometheus helm charts but I had to add some features to make it work. So now I'm contributing it back.

It adds:
- Prometheus sidecarsSpec field : necessary to add thanos's sidecar component to a Prometheus instance.
- Prometheus labels field as podMetadata.labels : necessary to make the thanos-peer service to work with thanos's sidecar component.

### DIFF
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.83
+version: 0.0.84
--- a/helm/kube-prometheus/requirements.yaml
+++ b/helm/kube-prometheus/requirements.yaml
@@ -6,7 +6,7 @@ dependencies:
     condition: deployAlertManager
 
   - name: prometheus
-    version: 0.0.44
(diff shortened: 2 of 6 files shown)


brancz:
Note that we will be working on a tighter integration with Thanos (it's one of the things on our list, but our list is loing :slightly_smiling_face: ), but in the mean time the containers field is your friend.

sdurrheimer:
@brancz One missing feature would be the ability to set custom prometheus's cmdlines options like `--storage.tsdb.min-block-duration=2h` and `--storage.tsdb.max-block-duration=2h` which are recommended by the Thanos team.

brancz:
Yeah there is already an issue for this, it’s something we need to fix soon.

gianrubio:
@sdurrheimer could you rebase it from master?

sdurrheimer:
Done

gianrubio:
I hope this will fix CI ;)

brancz:
If it's just the first CI job that fails, you can safely ignore it, https://github.com/coreos/prometheus-operator/pull/1536 is going to fix that.

PR #1492, opened 2018-06-15
*: Add Thanos integration

### LINKED ISSUE #1375: setting --storage.tsdb.min-block-duration and --storage.tsdb.max-block-duration
Hi, I'ld like to specify both of these settings manually but can't find any way in the operator API to do it, is it possible?

brancz:
This pull requests adds a deeper integration for Thanos. Users will simply have to specify their options for the Thanos sidecar via the new `thanos` field. As soon as users add this the respective flags will be added to the sidecar as well as the tsdb configurations as depicted in #1375.

Closes #1375 

@mxinden 

cc @bplotka @sdurrheimer


Makefile
```diff
@@ -36,7 +36,7 @@ prometheus-config-reloader:
 	-ldflags "-X github.com/coreos/prometheus-operator/pkg/version.Version=$(shell cat VERSION)" \
 	-o $@ cmd/$@/main.go
 
-pkg/client/monitoring/v1/zz_generated.deepcopy.go: $(DEEPCOPY_GEN_BINARY)
+pkg/client/monitoring/v1/zz_generated.deepcopy.go: pkg/client/monitoring/v1/types.go $(DEEPCOPY_GEN_BINARY)
```
mxinden: Good catch. Would you mind adding `.header` here as well?

example/thanos/prometheus.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: self
+  labels:
+    prometheus: self
+spec:
+  podMetadata:
+    labels:
+      thanos-peer: 'true'
+  replicas: 2
+  version: v2.2.1
```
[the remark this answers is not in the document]
mxinden: Not sure this should be specified in general, as the Prometheus Operator already [defaults](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L36) to `v2.2.1`.

pkg/client/monitoring/v1/types.go
```diff
@@ -215,6 +221,45 @@ type StorageSpec struct {
...
+// ThanosSpec defines parameters for a Prometheus server within a Thanos deployment.
+// +k8s:openapi-gen=true
+type ThanosSpec struct {
+	// Peers is a DNS name for Thanos to discover peers through.
+	Peers string `json:"peers"`
+	// Version describes the version of Thanos to use.
+	Version string `json:"version"`
+	// Thanos base image if other than default.
+	BaseImage *string `json:"baseImage,omitempty"`
+	// GCS configures use of GCS in Thanos.
```
mxinden: `s/GCS configures use of GCS in Thanos./GCS configures use of Google Cloud Storage in Thanos.

pkg/client/monitoring/v1/types.go
```diff
@@ -215,6 +221,45 @@ type StorageSpec struct {
...
+// ThanosSpec defines parameters for of AWS S3 (or another compatible service)
+// with Thanos.
+// +k8s:openapi-gen=true
+type ThanosS3Spec struct {
+	// S3-Compatible API bucket name for stored blocks.
+	Bucket *string `json:"bucket,omitempty"`
+	// S3-Compatible API endpoint for stored blocks.
+	Endpoint *string `json:"endpoint,omitempty"`
+	// Access key for an S3-Compatible API.
+	AccessKey *string `json:"accessKey,omitempty"`
```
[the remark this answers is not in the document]
mxinden: Further on that, I am not sure we should include secrets inside the Prometheus custom resource. How about specifying a secret for it instead?

test/e2e/prometheus_test.go
```diff
@@ -1070,6 +1070,79 @@ func TestPromOpMatchPromAndServMonInDiffNSs(t *testing.T) {
...
+
+	if _, err := testFramework.CreateServiceAndWaitUntilReady(framework.KubeClient, ns, tservice); err != nil {
+		t.Fatal("Creating prometheus service failed: ", err)
+	}
+
+	if _, err := testFramework.CreateServiceAndWaitUntilReady(framework.KubeClient, ns, qservice); err != nil {
+		t.Fatal("Creating Thanos querier service failed: ", err)
+	}
+
+	time.Sleep(time.Minute)
```
mxinden: Instead of a sleep I would prefer a `wait.Poll` here to prepare for slow test environments.

pkg/client/monitoring/v1/types.go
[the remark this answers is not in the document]
Bplotka: yes, indeed. S3 Secrets should be done through Secret + env var

pkg/client/monitoring/v1/zz_generated.deepcopy.go
```diff
@@ -986,3 +995,132 @@ func (in *TLSConfig) DeepCopy() *TLSConfig {
...
+func (in *ThanosGCSSpec) DeepCopy() *ThanosGCSSpec {
+	if in == nil {
+		return nil
+	}
+	out := new(ThanosGCSSpec)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
Bplotka: `writing into out. in must be non-nil.`

`In must be non-nil` instead? `In` first letter uppercase? I know this is autogenerated, but..

pkg/prometheus/statefulset.go
```diff
@@ -542,6 +542,66 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config) (*appsv1.Stateful
...
+
+	if p.Spec.Thanos != nil {
+		thanosBaseImage := c.ThanosDefaultBaseImage
+		if p.Spec.Thanos.BaseImage != nil {
+			thanosBaseImage = *p.Spec.Thanos.BaseImage
+		}
+
+		thanosArgs := []string{
+			"sidecar",
+			fmt.Sprintf("--cluster.peers=%s", p.Spec.Thanos.Peers),
```
sdurrheimer: Also, the `--cluster.peers` should not be required. In my use-case, the thanos-sidecar is alone and connected by a remote thanos-query --store parameter.

pkg/client/monitoring/v1/types.go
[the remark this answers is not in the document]
ant31: this field could be use to name the kubernetes secret containing the creds.

pkg/client/monitoring/v1/types.go
```diff
@@ -215,6 +221,45 @@ type StorageSpec struct {
...
+	BaseImage *string `json:"baseImage,omitempty"`
+	// GCS configures use of GCS in Thanos.
+	GCS *ThanosGCSSpec `json:"gcs"`
+	// S3 configures use of S3 in Thanos.
+	S3 *ThanosS3Spec `json:"s3"`
+}
+
+// ThanosGCSSpec defines parameters for use of GCS with Thanos.
+// +k8s:openapi-gen=true
+type ThanosGCSSpec struct {
```
ant31: why GCS / S3 distinction?

pkg/prometheus/statefulset.go
brancz: Got it. Thanks for the insight! I've only run with both.

pkg/client/monitoring/v1/types.go
brancz: Mostly because Thanos does, and I don't want to introduce an abstraction that could diverge if thanos diverges.

pkg/client/monitoring/v1/zz_generated.deepcopy.go
brancz: yeah we should report these to the kubernetes/code-generator repo

pkg/client/monitoring/v1/types.go
brancz: I'll make sure to expand acronyms at least once for each of them.

example/thanos/querier-deployment.yaml
```diff
@@ -19,7 +19,7 @@ spec:
     spec:
       containers:
       - name: thanos-query
-        image: improbable/thanos:latest
+        image: improbable/thanos:v0.1.0-rc.0
```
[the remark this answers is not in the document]
brancz: Especially as we use these in our e2e tests I would like to not use a floating version (besides them not existing anymore).

pkg/prometheus/statefulset.go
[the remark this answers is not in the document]
brancz: I need to check but ideally the log levels are identical to prometheus (or map somehow). I'd like that the most. I understand that some people don't want constant debug output of the mesh.

pkg/client/monitoring/v1/types.go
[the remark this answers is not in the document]
brancz: Yes good point, will change this to a `SecretKeySelector`, just like the `additionalScrapeConfigs` field.

pkg/prometheus/statefulset.go
Bplotka: sidecar will never block. It will just print this warning until it will be not alone. Also, we use gossip by default here in operator (since it is scoped to single cluster only for now), so having a mandatory cluster.peers seems fine.

pkg/client/monitoring/v1/types.go
Bplotka: if you want GCS it  requires  different configuration than if you want S3

pkg/prometheus/statefulset.go
sdurrheimer: @Bplotka I had the issue. After our discussion on Slack I did the modifications you recommended and my thanos-query couldn't communicate with the thanos-sidecar (`context deadline exceeded`) when the thanos-sidecar was spamming "I appear to be alone in the cluster". I have then removed the `--cluster.peers`, restart the prometheus+thanos-sidecar and it worked.

brancz: For now I'm marking `Peers` within the `Thanos` config as optional, the entire integration is marked experimental for now so we can still make modifications before we promote it to be stable.

brancz:
@mxinden could you give this a re-review?

test/e2e/prometheus_test.go
```diff
@@ -1070,6 +1070,107 @@ func TestPromOpMatchPromAndServMonInDiffNSs(t *testing.T) {
...
+		// Prometheus replicas by both replicas.
+		expected := 4
+		if result != expected {
+			log.Printf("Unexpected number of results from query. Got %d, expected %d. retrying...\n", result, expected)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatal("Failed to get correct result from Thanos querier: ", err)
```
mxinden: `s/Failed/failed`

PR #1494, opened 2018-06-15
Removed the architecture node selector

dimm0:
Can this be removed? It doesn't work well with podnodeselector admission controller

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
@@ -98,7 +98,6 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
           replicas: $._config.alertmanager.replicas,
           version: $._config.versions.alertmanager,
           baseImage: $._config.imageRepos.alertmanager,
-          nodeSelector: { 'beta.kubernetes.io/os': 'linux' },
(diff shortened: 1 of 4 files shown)


mxinden:
@dimm0 could you elaborate further on the issue?

According to the [K8s docs](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector) you can either annotate each namespace to allow this node selector, or you set a cluster-wide configuration.

Would that not be an option?

dimm0:
I have my own nodeselector set for all system components (`monitor=true`), and so far I haven't found a way to also add yours.

With `scheduler.alpha.kubernetes.io/node-selector: monitor=true,beta.kubernetes.io/os=linux` in namespace label, `monitoring: "monitor=true,beta.kubernetes.io/os=linux"` in /etc/kubernetes/adm-control/admission-control.yaml and 

```  
nodeSelector:
    beta.kubernetes.io/os: linux
```

in yaml definition I get `pod node label selector labels conflict with its namespace whitelist`

Why is this nodeSelector needed at all?

brancz:
> Why is this nodeSelector needed at all?

It was added for this stack to work on mixed os clusters (linux + windows for example).

The os label is standardized for all Kubernetes clusters, even without this issue it is a best practice (maybe even requirement?) for conformant clusters.

dimm0:
I was going to ask whether we could make this configurable, but then I decided to write my own better admission controller.
Thanks all :)

PR #1512, opened 2018-06-20
kube-prometheus: Remove default serviceMonitorsSelector

jdoupe:
The default serviceMonitorsSelector that the prometheus chart provides is sufficient and having one set in kube-prometheus makes it impossible to just get the default behaviour (without actually overriding it).

### DIFF
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.86
+version: 0.0.87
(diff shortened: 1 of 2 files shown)


brancz:
I think this is generally ok, I just feel this is going to confuse people who want to add additional ServiceMonitor objects that are in a different namespace, as the Prometheus server requires a `Role` plus `RoleBinding` per namespace a `ServiceMonitor` exist and is supposed to be picked up from.

brancz:
This needs a rebase (for versioning purposes).

@jdoupe please have a look at the contribution guidelines to see how you need to bump the chart: https://github.com/coreos/prometheus-operator/tree/master/helm#how-to-contribue

Then we can merge this.

jdoupe:
Rebased and chart version bumped.

brancz:
The test failure is a flake. No need to worry about it. This lgtm :+1: . Merging as @gianrubio already gave consent if version is properly bumped.

PR #1493, opened 2018-06-15
RemoteWriteSpec Expose QueueConfig Parameters (#1488)

midnightconman:

PR #1517, opened 2018-06-22
Remote write queue config

brancz:
Supersedes #1493 

@midnightconman


brancz:
@mxinden could you give this a review?

pkg/client/monitoring/v1/zz_generated.deepcopy.go
```diff
@@ -790,6 +806,15 @@ func (in *RemoteWriteSpec) DeepCopyInto(out *RemoteWriteSpec) {
 			**out = **in
 		}
 	}
+	if in.QueueConfig != nil {
+		in, out := &in.QueueConfig, &out.QueueConfig
```
mxinden: Why complicating this via a pointer reference? Why not edit `in.QueueConfig` and `out.QueueConfig` right away? Am I missing something?

brancz: It's generated code. It's so `in` and `out` can be treated the same regardless of their type, and then the generator can generate further code based on the actual type.

Here the line 809 and 810 are generated:

https://github.com/kubernetes/gengo/blob/f659ec9d5207b3bf034eb9b7e22720d360d28712/examples/deepcopy-gen/generators/deepcopy.go#L848-L849

And then recursively `doPointer` is caled by `generateFor` because `QueueConfig` is a pointer type struct.

brancz:
@mxinden could you give this a re-review?

pkg/client/monitoring/v1/zz_generated.deepcopy.go
mxinden: Oh, I missed that. Thanks for clarifying.

Issue #1150, opened 2018-03-26
How to provide imagePullSecrets when using helm operator?


andersflemmen:
What is the best way to provide imagePullSecrets to the different service accounts used by the operator when using helm for deployment? I am currently patching the service accounts manually, but this requires running the deployment script multiple times, as the service accounts are created when installing the chart, and any changes to the service accounts don't seem to be picked up by currently running pods. Is there a better way?

gianrubio:
There's no support for private images for now. I'm wondering what kind of image are you deploying to require to host privately. In my case all prometheus metrics are running in custom build but I always push as public image.

brancz:
This should be supported, different companies have different standards on this. The Prometheus and Alertmanager objects already support it so at just a matter of adding the playing with helm.

andersflemmen:
Yes, it is company standards. Would be really nice to have indeed!

gianrubio:
@andersflemmen would you like to contribute to the project sending a PR to us?

lilic:
Since we don't support the helm chart in this repo anymore, closing this issue. Feel free to open a new one if you think it's not valid, thanks!

PR #1518, opened 2018-06-22
[helm] Add imagePullSecrets option for ServiceAccounts, Prometheus and Alertmanager

andersflemmen:
Gives the option to provide imagePullSecrets in helm values as requested in #1150


andersflemmen:
Sounds reasonable to have a default value yes, but shouldn't it be [] in that case?

gianrubio:
Yeah, go for a list.

gianrubio:
lgtm on green

PR #1520, opened 2018-06-25
Add the possiblity to set resource limits for kube-state-metrics cont…

danielcb:
The current chart only allows setting the resource limits for one of the two containers in the kube-state-metrics deployment. Since we're running bigger clusters and kube-state-metrics gets constantly OOM killed, this introduces a resource entries (without any defaults) in the values and deployment file.


brancz:
this is immediately going to be overridden by the addon-resizer requests/limits, no?

danielcb:
It is, sry. I haven't looked closely at whats happening regarding resources here.

PR #1524, opened 2018-06-25
kube-prometheus/Makefile: Change target output from files to folder

mxinden:
Previously the `manifests` and `vendor` folder content has been build
via `manifests/**` and `vendor/**` target definitions. This results in
two issues:

1. The GNU Make recipe is executed for every single file in the folders.

2. Not all timestamps inside the `manifests` folder would be updated on
every run, thereby needing to be rerun on all following target
executions.

Solution:

1. Define the target based on the folder, not individual files

2. Remove target folder before execution recipe to ensure all timestamps
are updated.


brancz:
Could you rebase this, otherwise lgtm

brancz:
Actually it seems I could resolve the conflicts from the UI.

brancz:
Merging this as it significantly speeds up CI and CI failures are fixed on master.

PR #1527, opened 2018-06-26
fix: wrong port name

oilbeater:
The port create by operator is `web` not `http` which will lead prometheus scraping no data.


gianrubio:
lgtm on green

PR #1536, opened 2018-06-27
kube-prometheus: Regenerate prometheus CRD

brancz:
Follow up to #1517 

@mxinden


brancz:
Moving forward to unblock CI.

Issue #1456, opened 2018-06-09
add additionalAlertManagerConfigs to Helm Chart


alice-sawatzky:
the latest version (0.20.0) introduces the `additionalAlertManagerConfigs` field to the [`PrometheusSpec`](https://github.com/coreos/prometheus-operator/blob/v0.20.0/Documentation/api.md#prometheusspec),  which looks super handy! Unfortunately, it doesn't seem to be added yet to the helm chart. Can we add this?

PR #1537, opened 2018-06-27
Use correct feature flag for additionalAlertManagerConfigs option

danielcb:
There seems to be a bug introduced in commit https://github.com/coreos/prometheus-operator/commit/cab34d7fe1a7e2cc881cfda176b9fa6f24ea3ca6 (fix for #1456)

The config block additionalAlertManagerConfigs in promtheus.yaml gets activated by the flag .Values.additionalScrapeConfigs instead of .Values.additionalAlertManagerConfigs, which leads to errors when enabling additionalScrapeConfigs and not additionalAlertManagerConfigs.


danielcb:
I think I broke the build with the version bump?

danielcb:
sry, should have read it in the first place. I've updated the requested files and rebased on current master.

PR #1550, opened 2018-06-29
[WIP] Load basic-auth secrets from ServiceMonitor namespace

huydinhle:
PR for basic-auth couldn't find secrets
https://github.com/coreos/prometheus-operator/issues/1487


pkg/prometheus/operator.go
```diff
@@ -1000,24 +1002,36 @@ func (c *Operator) createOrUpdateConfigurationSecret(p *monitoringv1.Prometheus)
...
-
-	listSecrets, err := sClient.List(metav1.ListOptions{})
-
+	promSecrets, err := sClient.List(metav1.ListOptions{})
 	if err != nil {
 		return err
 	}
 
-	basicAuthSecrets, err := c.loadBasicAuthSecrets(smons, p.Spec.RemoteRead, p.Spec.RemoteWrite, listSecrets)
+	secretMap := make(map[string]*v1.SecretList)
```
mxinden: `secretMap` seems a bit generic to me. What do you think about `sMonSecretMap`?

mxinden: In addition it seesm like `secretMap` is only used inside of `loadBasicAuthSecrets`. How about retrieving the secrets in there, instead of having the logic distributed between functions?

huydinhle: good call, let me do it

huydinhle:
@mxinden  no problem at all, I will add an e2e test for this

brancz:
@mxinden Both valid PRs in my opinion, but we need to make sure that one doesn't break the other: https://github.com/coreos/prometheus-operator/pull/1439

PR #1552, opened 2018-06-29
kube-prometheus: Add kube-prometheus mixin to configure static etcd

brancz:
This adds an easy to configure mixin to the kube-prometheus library to monitor a static set of etcd nodes.

@mxinden


contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-static-etcd.libsonnet
```diff
@@ -0,0 +1,95 @@
+local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
+
+(import 'etcd-mixin/mixin.libsonnet') + {
```
mxinden: Why do we need the `etcd-mixin` here? It only gives us alerting and dashboarding definitions, right?

brancz: Exactly, it gives us dashboards and alerting rules :slightly_smiling_face: .

PR #1553, opened 2018-06-29
Fix prometheus chart for ruleNamespaceSelector

aleerizw:
Missed the variable in https://github.com/coreos/prometheus-operator/pull/1540


gianrubio:
lgtm on green

gianrubio:
lgtm on green

PR #1556, opened 2018-06-30
Update client libraries to Kubernetes 1.11

tamalsaha:
Client-go v8.0.0 release has introduced a new dynamic client and moved original dynamic client to deprecated-dynamic package . In this pr, I just moved to deprecated-dynamic pkg . If you want to move to new dynamic client in this pr, let me know.
https://github.com/kubernetes/client-go/blob/master/CHANGELOG.md#v800


tamalsaha:
Any idea what I am missing here? I ran the `make generate` command .

brancz:
Yeah we recently reworked our `Makefile` and haven't gotten target dependencies, I think this should fix things:

```
rm pkg/client/monitoring/v1/zz_generated.deepcopy.go pkg/client/monitoring/v1/openapi_generated.go
make pkg/client/monitoring/v1/zz_generated.deepcopy.go pkg/client/monitoring/v1/openapi_generated.go
go install github.com/coreos/prometheus-operator/cmd/po-crdgen
make example/prometheus-operator-crd/**.crd.yaml
make jsonnet/prometheus-operator/prometheus-crd.libsonnet
make generate
```

tamalsaha:
Thanks @brancz . This fixed all tests but one.

brancz:
I fixed the remaining generation issues. This lgtm :+1: thanks a lot for doing this @tamalsaha!

tamalsaha:
Thanks for merging this quickly.

brancz:
Note that there are still difficulties with subresources being enabled by default in Kubernetes 1.11 so the Prometheus Operator will not work on Kubernetes 1.11 yet. It is the problem depicted in #1293 .

tamalsaha:
The issue seems to be `if subresources for custom resources are enabled, only properties can be used at the root of the schema"` . We got a [similar error report ](https://github.com/appscode/voyager/pull/1168)for our Voyager project. 

I think this happens when sub resources is enabled in 1.10 cluster. In 1.11, I see that this was fixed https://github.com/kubernetes/apiextensions-apiserver/commit/5dc100524bb96c95827b77d45af06bc3b595f519 . I have not yet tried on a 1.11 cluster.

I think the fix may to just modify this in the crd-validation binary and leave the comment in the code.

On an application level, I have added a [`UpdateXYZStatus`](https://github.com/appscode/voyager/blob/master/client/clientset/versioned/typed/voyager/v1beta1/util/ingress.go#L82) method that will use `/status` endpoint if subresource is used, otherwise use the `Update()` method. Then used it in rest of the application. Then I have a global flag that user pass to activate sub resource.

I would like to hear/follow what you folks do on this.

PR #1560, opened 2018-07-02
Helm: Support 'external' `node_exporter`s

NicolasT:
Similar to how `exporter-kube-etcd` supports monitoring of etcd clusters
running outside the Kubernetes cluster, this patch permits to configure
`exporter-node` to *not* deploy `node_exporter` as a `DaemonSet` in the
cluster, and instead fetch metrics from pre-provisioned `node_exporter`
instances, e.g. deployed as part of the host OS.


NicolasT:
Prometheus `node_exporter` is not really meant to run inside a Docker container, because this complicates monitoring all volumes in the system, for example (see https://github.com/prometheus/node_exporter#using-docker). Also, deploying `node_exporter` as a *DaemonSet* results in only Kubernetes nodes to be monitored, whilst one can have e.g. the *etcd* cluster running on dedicated systems.

Similar to how it's done in `exporter-kube-etcd`, this PR allows to disable the *DaemonSet*-based deployment, and set up a static list of endpoints (IP addresses) where `node_exporter` is deployed using other means to be monitored.

NicolasT:
These changes conflict on chart versions. I'll fix these once/if the idea and general changes get approved :smiley:

gianrubio:
lgtm on green

NicolasT:
CI seems to be failing on AlertManager configuration or something, not sure how that would relate :thinking:

brancz:
We just merged these tests, if we can’t figure limit what’s happenin quickly we will revert the changes involved. For now ignore those failures when doing helm changes.

gianrubio:
the helm test step was succeed so I'm gonna merge it :)

PR #1561, opened 2018-07-02
*: Fix CRD generation

brancz:
@mxinden @ant31 @tamalsaha


tamalsaha:
LGTM

brancz:
I'm still going to test and make sure sub resources work now. Ideally even enable that by default in the e2e tests.

lucab:
Verbose note for future self-reference.

Linux kernel marks the initial `/` (and submounts, by default) as `MS_SLAVE` propagation, with systemd turning it very early into `MS_SHARED`. This is not true on travis, as it uses an old Ubuntu Trusty (either not booting with systemd or with an old one, I didn't check).
Changing the propagation of `/` recursively later on (as done in this PR) affects all the mounts in the subtree, but other mounts that happened before that (like `/proc` and `/sys`) are only part of their own separate private hierarchy and won't be affected by it. Thus they need to be individually changed as well.

(at least, this is my current understanding as to why three separate `mount` invocations are required, confirmed via `cat /proc/1/mountinfo` at different stages)

PR #1566, opened 2018-07-03
Allow the prometheus service name to be overridden. 

rgardam:
This enables the use case of sub-charting the kube-prometheus chart so that you can define the datasource grafana needs.


rgardam:
Sorry, I was away on holidays without internet. I'll have a look at this now :)

rgardam:
bumped the version. :)

rgardam:
@gianrubio Hey, thanks for the review. I believe i've addressed the requested changes.

helm/prometheus/templates/service.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
 {{- if .Values.service.labels }}
 {{ toYaml .Values.service.labels | indent 4 }}
 {{- end }}
-  name: {{ template "prometheus.fullname" . }}
+  name: {{ .Values.service.nameOverride | default .Release.Name }}
```
gianrubio: does ` "prometheus.fullname"` has the same value of `.Release.Name`?

rgardam: I had a look at how the helper provides `prometheus.fullname` and this would be a truncated version. However, it doesn't seem possible to include the helper function as a default value. 

I'm getting parsing errors. Do you know if this is possible?

`{{ .Values.service.nameOverride | default template "prometheus.fullname" . }}`

rgardam:
I'm closing this PR as I don't see much movement. I currently have a work around. 
I would like to come back to this in the future if possible.

PR #1569, opened 2018-07-03
helm: Allow setting pod selector on etcd

vsliouniaev:
Kubeadm uses different labels when it creates the pods. Other exporters allow selecting these, but etcd is set up by kubeadm has an additional discrepancy


gianrubio:
lgtm on green

vsliouniaev:
I don't think this is related to the changes

gianrubio:
> I don't think this is related to the changes
What do you mean?

brancz:
I think he meant the CI failures, I restarted the failed job, yes it's a flake that we are in the process of fixing :slightly_smiling_face:

PR #1579, opened 2018-07-04
README: Update target for container image build

mrueg:
Target seems to have changed


mxinden:
Right, thanks for the fix @mrueg!

brancz:
Thanks!

Issue #1169, opened 2018-04-02
Support Thanos Deployments


wleese:
Support deploying Thanos along with Prometheus, to provide a default solution to long term metric storage and horizontal scaling.

https://github.com/improbable-eng/thanos

brancz:
We've already worked out a PoC for this, and will in the upcoming weeks provide more details! :slightly_smiling_face: It's coming!

dhain:
Very interested in this. Any links to where this dev is happening? Happy to try out pre-release.

brancz:
Actually we have released more integrated support for this in the latest release (v0.21.0).

API documentation is here:
* https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#prometheusspec
* https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#thanosspec

If anyone would like to write up a user guide on how to use this, that would be fantastic!

deltasquare4:
I tried `v0.21.0` out on GCP, but I think it has a problem:

* Default thanos sidecar container does not have a way to override the default path, which is `./data` ([Source](https://github.com/improbable-eng/thanos/blob/v0.1.0-rc.1/cmd/thanos/sidecar.go#L43)), but should be `/prometheus` ([Source](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L39))
* The `/prometheus` volume is not being mounted inside thanos container, so thanos is not able to access prometheus data.

mxinden:
Closing here. @dhain feel free to reopen if you have further questions.

@deltasquare4 Thanks a lot for your help!

PR #1583, opened 2018-07-05
Mount prometheus data volume into thanos sidecar (#1169)

deltasquare4:
As described in https://github.com/coreos/prometheus-operator/issues/1169#issuecomment-402659888, thanos sidecar container did not have prometheus data directory mounted inside it.

* Mounted prometheus data volume inside thanos on `/prometheus` path
* Override default `tsdb.path` directory to `storageDir` of prometheus operator (default `/prometheus`)


brancz:
lgtm :+1: thanks a lot!

deltasquare4:
Thank you for looking into it quickly.

PR #1586, opened 2018-07-05
[helm/prometheus] Rename extra rules

richerve:
- If having a release that contains `prometheus` then those extra rules ConfigMap names will be something like `prometheus-prometheus-rules`
- Having the `rules` suffix clashes with the ConfigMap name for the default prometheus rules.

### DIFF
--- a/helm/prometheus/templates/rules.yaml
+++ b/helm/prometheus/templates/rules.yaml
@@ -8,10 +8,11 @@ metadata:
     heritage: {{ .Release.Service }}
     prometheus: {{ .Values.prometheusLabelValue | default .Release.Name | quote }}
     release: {{ .Release.Name }}
+    role: alert-rules
     {{- if .Values.rules.additionalLabels }}
 {{ toYaml .Values.rules.additionalLabels | indent 4 }}
     {{- end }}
-  name: prometheus-{{ .Release.Name }}-rules
+  name: {{ template "prometheus.fullname" . }}-extra-rules
 spec:
   groups:
 {{ toYaml .Values.rules.value | indent 4 }}
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.96
+version: 0.0.97


richerve:
Timing out issues again?
```
--- FAIL: TestPrometheusMultiplePrometheusRulesDifferentNS (318.10s)
	prometheus_test.go:623: waiting for alert 'first-alert' to fire: timed out waiting for the condition: expected 1 query result but got 0
=== CONT  TestPrometheusMultiplePrometheusRulesSameNS
--- FAIL: TestPrometheusRulesExceedingConfigMapLimit (330.48s)
	prometheus_test.go:665: waiting for alert 'my-alert-0' to fire: timed out waiting for the condition: expected 1 query result but got 0
```

brancz:
These Tests just got merged. Please ignore them if they are flaking. cc @mxinden

helm/prometheus/templates/rules.yaml
```diff
@@ -8,10 +8,11 @@ metadata:
     heritage: {{ .Release.Service }}
     prometheus: {{ .Values.prometheusLabelValue | default .Release.Name | quote }}
     release: {{ .Release.Name }}
+    role: alert-rules
```
vsliouniaev: Could these be `recording-rules` as well as `alerting-rules` ?

brancz: That is correct.

richerve: Totally agree, that labels is used on all other configmap definitions of rules. Formally at some point this Chart should migrate to define proper PrometheusRules objects instead and at that point this labels can be "fixed".

vsliouniaev: @richerve Would your suggestion be to simply remove the `role: alert-rules` label entirely at that point?

richerve: Yes because at the end `PrometheusRule` is an explicit CRD so there will be no more "ConfigMap that serves as Prometheus rule definition"

vsliouniaev: Thanks for that :-) Pr's been merged so we don't need to add this role any longer

richerve: @vsliouniaev indeed the role label will not be needed but the name of the rule (now PrometheusRule) is the main purpose of these PR.

richerve:
Rebased on master

PR #1590, opened 2018-07-06
helm: kube-prometheus: add deployPrometheus toggle

### LINKED ISSUE #1578: Add deployPrometheus value to helm/kube-prometheus
Right now we use kube-prometheus to deploy the bundle of charts that a complete Prometheus setup consists of (Prometheus, Alertmanager, Node-exporters etc etc).

However, not all options defined in the template for `prometheus` exist in the kube-prometheus chart (such as remoteWrite, https://github.com/coreos/prometheus-operator/issues/352 ).

We deploy Grafana separately since we need values not available in kube-prometheus. It would be nice to be able to NOT deploy Prometheus as part of kube-prometheus as well, and do a separate deploy while waiting for kube-prometheus to reach value-parity with the underlying charts.

StianOvrevage:
helm: kube-prometheus: Added deployPrometheus toggle value

This allows kube-prometheus to install all components except Prometheus so that Prometheus can be installed separately when greater customization on Prometheus are needed.

Fixes #1578

### DIFF
--- a/helm/kube-prometheus/Chart.yaml
+++ b/helm/kube-prometheus/Chart.yaml
@@ -7,4 +7,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/coreos/prometheus-operator
-version: 0.0.95
+version: 0.0.96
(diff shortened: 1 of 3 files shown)


brancz:
The test failure can be safely ignored. We're working on it, but it does not concern helm at all.

gianrubio:

PR #1601, opened 2018-07-10
alertmanager/statefulset: Move gossip port to 9094

### LINKED ISSUE #1574: Wrong port for Alertmanager >=v0.15.0
Starting with Alertmanager version 0.15.0 there are different ports being used for gossiping, but even for Alertmanager >=v0.15.0 versions we still add the old ports to the StatefulSet. Here:

https://github.com/coreos/prometheus-operator/blob/041acb182d8362d7758297e80669ebb3b1e3a4b6/pkg/alertmanager/statefulset.go#L255-L270

This is a bug and should be fixed for any version >=v0.15.0. I'm not 100% certain but I think the port that is now used is `9094` for both TCP and UDP.

@mxinden

mxinden:
The Alertmanager upstream default port is 9094. This patch brings the
Proemtheus Operator project inline with the upstream default.

Fixes #1574

### DIFF
--- a/example/networkpolicies/alertmanager.yaml
+++ b/example/networkpolicies/alertmanager.yaml
@@ -31,8 +31,10 @@ spec:
           values:
           - main
     ports:
-    - port: 6783
+    - port: 9094
       protocol: tcp
+    - port: 9094
+      protocol: udp
   podSelector:
     matchLabels:
       alertmanager: main
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -38,6 +38,7 @@ const (
 	alertmanagerConfDir    = "/etc/alertmanager/config"
 	alertmanagerConfFile   = alertmanagerConfDir + "/alertmanager.yaml"
 	alertmanagerStorageDir = "/alertmanager"
+	alertmanagerGossipPort = 9094
 )
 
 var (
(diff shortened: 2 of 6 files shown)


example/networkpolicies/alertmanager.yaml
```diff
@@ -31,9 +31,9 @@ spec:
           values:
           - main
     ports:
-    - port: 6783
+    - port: 9094
```
brancz: memberlist uses TCP and UDP, no?

pkg/alertmanager/statefulset.go
```diff
@@ -152,8 +153,8 @@ func makeStatefulSetService(p *monitoringv1.Alertmanager, config Config) *v1.Ser
 				},
 				{
 					Name:       "mesh",
-					Port:       6783,
-					TargetPort: intstr.FromInt(6783),
+					Port:       alertmanagerGossipPort,
+					TargetPort: intstr.FromInt(alertmanagerGossipPort),
```
brancz: I think we will need both TCP and UDP declarations here.

pkg/alertmanager/statefulset.go
```diff
@@ -253,13 +254,13 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 	podLabels["alertmanager"] = a.Name
 
 	for i := int32(0); i < *a.Spec.Replicas; i++ {
-		amArgs = append(amArgs, fmt.Sprintf("--cluster.peer=%s-%d.%s.%s.svc:6783", prefixedName(a.Name), i, governingServiceName, a.Namespace))
+		amArgs = append(amArgs, fmt.Sprintf("--cluster.peer=%s-%d.%s.%s.svc:%v", prefixedName(a.Name), i, governingServiceName, a.Namespace, alertmanagerGossipPort))
```
squat: we normally template as `%d` for integers, no?

brancz:
Closing as #2644 is merged.

PR #1610, opened 2018-07-12
bugfix: fix alertmanager port number label name

cofyc:
Pod number label is `__meta_kubernetes_pod_container_port_number` not `__meta_kubernetes_container_port_number`, see https://prometheus.io/docs/prometheus/latest/configuration/configuration/ and prometheus code https://github.com/prometheus/prometheus/blame/master/discovery/kubernetes/pod.go#L143.

This is reason why port int does not work in this issue: https://github.com/coreos/prometheus-operator/issues/1356.


brancz:
Wow, nicely spotted! This lgtm :+1:

PR #1616, opened 2018-07-13
Configure kube-state-metrics

directionless:
As I work with kube-state-metrics in a large cluster, I found I needed to make some adjustments.

- Expose the collectors, allowing one to configure exclusions. 
- Expose resource limits, allowing large adjustments.

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
@@ -4,6 +4,22 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
   _config+:: {
     namespace: 'default',
 
+    kubeStateMetrics+:: {
+      // when this is an empty string, you get the default set
+      collectors: '',
+      scrapeTimeout: '',
+      scrapeInterval: '30s',
+    },
+
+    resizer+:: {
+      kubeStateMetrics+:: {
+        cpu: '100m',
+        extraCpu: '2m',
+        memory: '150Mi',
+        extraMemory: '30Mi',
+      },
+    },
+
     versions+:: {
       kubeStateMetrics: 'v1.3.1',
       kubeRbacProxy: 'v0.3.1',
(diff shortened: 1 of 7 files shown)


directionless:
This is prompted by https://github.com/kubernetes/kube-state-metrics/issues/493

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -4,6 +4,22 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
   _config+:: {
     namespace: 'default',
 
+    kubeStateMetrics+:: {
+      // when this is an empty string, you get the default set
+      collectors: '',
+      scrapeTimeout: '',
```
brancz: I just checked, the default scrape timeout is `10s`, I think we can have this default to the value of `scrapeInterval` for kube-state-metrics. And generally I think `1m` interval is probably a more universal choice. Even `2m` is still fine with the dashboards and alerting rules that we currently have (and more generally more than `2m` scrape interval is not advised).

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -4,6 +4,22 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
...
+      scrapeInterval: '30s',
+    },
+
+    resizer+:: {
```
brancz: I think it would be more natural if we fold this into the `kubeStateMetrics` configuration.

directionless: This was the bit I was stylistically unsure on. I bias towards `kubeStateMetrics` as well, but I followed the versions/imageRepos pattern. I'll push an update moving it!

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
directionless: I'm not sure if you're suggesting I merge `scrapeTimeout` and `scrapeInterval` into a single variable, or if I should keep them split and set them to better initial values. (Happy to do either, and I don't have much preference)

brancz: I think we should keep them split but have them default on the same value.

brancz:
Thanks for contributing this! I just re-generated to satisfy CI. Thanks!

directionless:
Awesome, thank you!

PR #1622, opened 2018-07-16
bundle.yaml: Bump Prometheus Operator memory request and limit

mxinden:
When handling big Kubernetes objects, marshalling objects is memory
intense. This can be reproduced with the end-to-end test
`TestPrometheusRulesExceedingConfigMapLimit`. This patch doubles the
memory request and limit of the Prometheus Operator deployment to 100mb
and 200mb.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -20,10 +20,7 @@ import (
 	"os"
 	"testing"
 
-	"github.com/coreos/prometheus-operator/pkg/k8sutil"
 	operatorFramework "github.com/coreos/prometheus-operator/test/framework"
-
-	"k8s.io/api/core/v1"
 )
 
 var framework *operatorFramework.Framework
@@ -38,8 +35,8 @@ func TestMain(m *testing.M) {
 	flag.Parse()
 
 	var (
-		err  error
-		code int = 0
+		err      error
+		exitCode int
 	)
 
 	if framework, err = operatorFramework.New(*ns, *kubeconfig, *opImage); err != nil {
@@ -47,26 +44,8 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
-	err = k8sutil.WaitForCRDReady(framework.MonClientV1.Prometheuses(v1.NamespaceAll).List)
-	if err != nil {
-		log.Printf("Prometheus CRD not ready: %v\n", err)
-		os.Exit(1)
-	}
-
-	err = k8sutil.WaitForCRDReady(framework.MonClientV1.ServiceMonitors(v1.NamespaceAll).List)
-	if err != nil {
-		log.Printf("ServiceMonitor CRD not ready: %v\n", err)
-		os.Exit(1)
-	}
-
-	err = k8sutil.WaitForCRDReady(framework.MonClientV1.Alertmanagers(v1.NamespaceAll).List)
-	if err != nil {
-		log.Printf("Alertmanagers CRD not ready: %v\n", err)
-		os.Exit(1)
-	}
-
 	defer func() {
-		if code != 0 {
+		if exitCode != 0 {
 			if err := framework.PrintEvents(); err != nil {
 				log.Printf("failed to print events: %v", err)
 			}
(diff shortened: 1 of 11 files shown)


mxinden:
Memory profile of Prometheus Operator running `TestPrometheusRulesExceedingConfigMapLimit`: [file.pdf](https://github.com/coreos/prometheus-operator/files/2197767/file.pdf)

test/e2e/main_test.go
```diff
@@ -77,11 +56,31 @@ func TestMain(m *testing.M) {
...
+	exitCode = m.Run()
+
+	// Check if Prometheus Operator ever restarted.
+	restarts, err := framework.GetRestartCount(framework.Namespace.Name, framework.OperatorPod.Name)
+	if err != nil {
+		log.Printf("failed to retrieve restart count of Prometheus Operator pod: %v", err)
+		exitCode = 1
+	}
+	if len(restarts) != 1 {
+		log.Printf("expected to have 1 container but got %v", len(restarts))
```
squat: consider using `%d` to format integers

test/e2e/main_test.go
```diff
@@ -77,11 +56,31 @@ func TestMain(m *testing.M) {
...
+		exitCode = 1
+	}
+	for _, restart := range restarts {
+		if restart != 0 {
+			log.Printf(
+				"expected Prometheus Operator to never restart during entire test execution but got %v restarts",
```
squat: same here

test/framework/alertmanager.go
```diff
@@ -279,8 +278,12 @@ func (f *Framework) GetSilences(ns, n string) ([]amAPISil, error) {
 	return getSilencesResponse.Data, nil
 }
 
+// WaitForAlertmanagerConfigToContainString retrieves the Alertmanager
+// configuration via the Alertmanager's API and checks if it contains the given
+// string.
 func (f *Framework) WaitForAlertmanagerConfigToContainString(ns, amName string, expectedString string) error {
```
squat: since all of the parameters are strings, we could simplify the signature as:
```go
func (f *Framework) WaitForAlertmanagerConfigToContainString(ns, amName, expectedString string) error
```

test/framework/alertmanager.go
```diff
@@ -290,10 +293,14 @@ func (f *Framework) WaitForAlertmanagerConfigToContainString(ns, amName string,
 			return true, nil
 		}
 
-		log.Printf("\n\nExpected:\n\n%#+v\n\nto contain:\n\n%#+v\n\n", config.Data.ConfigYAML, expectedString)
-
 		return false, nil
 	})
+
+	if err != nil {
+		return fmt.Errorf("failed to wait for alertmanager config to contain '%v': %v: %v", expectedString, err, pollError)
```
squat: should we use `%q` instead of `'%v'` to format the `expectedString`?

test/framework/framework.go
```diff
@@ -165,6 +165,26 @@ func (f *Framework) setupPrometheusOperator(opImage string) error {
...
+	}
+
+	err = k8sutil.WaitForCRDReady(f.MonClientV1.ServiceMonitors(v1.NamespaceAll).List)
+	if err != nil {
+		return errors.Wrap(err, "ServiceMonitor CRD not ready: %v\n")
+	}
+
+	err = k8sutil.WaitForCRDReady(f.MonClientV1.Alertmanagers(v1.NamespaceAll).List)
+	if err != nil {
+		return errors.Wrap(err, "Alertmanagers CRD not ready: %v\n")
```
squat: Is it multiple Alertmanagers or just one?

test/framework/pod.go
```diff
@@ -42,3 +42,20 @@ func (f *Framework) PrintPodLogs(ns, p string) error {
...
 }
+
+// GetRestartCount returns a map of container names and their restart counts for
+// a given pod.
+func (f *Framework) GetRestartCount(ns, podName string) (map[string]int32, error) {
+	restarts := map[string]int32{}
+
+	pod, err := f.KubeClient.CoreV1().Pods(ns).Get(podName, metav1.GetOptions{})
+	if err != nil {
+		return map[string]int32{}, errors.Wrap(err, "failed to retrieve pod to get restart count")
```
squat: you could return `restarts` since you already instantiated one empty map. or better yet, just return `nil, err`. In Go, slices and maps are both pointer types so you can return nil in their place. If this function returns an error, there is no guarantee that the map should be iterable.

test/e2e/main_test.go
mxinden: Using `%v` uses the default format of the `Stringer` interface, thereby a base10, right?

test/framework/alertmanager.go
mxinden: Oh cool, wasn't aware of that.

mxinden:
Thanks for the review @squat, would you mind taking another look?

test/e2e/main_test.go
squat: Sure, but `%d` is designed for integers. Why should we be less specific?

mxinden:
Thanks for the help. Changed to `%d`, anything else missing?

brancz:
I think something went wrong with rebasing here, the commit message doesn't seem to be in line with the changes.

squat:
lgtm once rebase is fixed :+1:

mxinden:
Thanks @brancz @squat for the feedback.

PR #1624, opened 2018-07-17
Update dead links to latest version of Kubernetes docs

the-smooth-operator:


brancz:
Thanks for updating this! You will also need to re-generate to update the API docs with this. Just run `make generate-in-docker` and commit the changes.

The-smooth-operator:
Unfortunately running `make generate-in-docker` seems not to be regenerating the docs.

I can see in the Makefile:81 `generate: ... Documentation/*`, however in the log of the build I can't find any reference to it (only to compatibility.md)

mxinden:
@The-smooth-operator Thanks for the contribution.

https://github.com/coreos/prometheus-operator/pull/1625 should fix your issue. Once it is merged, could you rebase and run `make generate` again? If that is not fixing it `rm $GOPATH/bin/po-docgen && make generate` should do.

PR #1625, opened 2018-07-17
Makefile: Properly rebuild po-docgen on src changes

mxinden:


Makefile
```diff
@@ -78,7 +78,7 @@ hack/prometheus-config-reloader-image: cmd/prometheus-config-reloader/Dockerfile
 ##############
 
 .PHONY: generate
-generate: pkg/client/monitoring/v1/zz_generated.deepcopy.go pkg/client/monitoring/v1/openapi_generated.go jsonnet/prometheus-operator/**-crd.libsonnet bundle.yaml kube-prometheus Documentation/*
+generate: pkg/client/monitoring/v1/zz_generated.deepcopy.go pkg/client/monitoring/v1/openapi_generated.go jsonnet/prometheus-operator/**-crd.libsonnet bundle.yaml kube-prometheus $(shell find Documentation -type f)
```
squat: Now that you are replacing the splats with shell invocations, consider replacing `jsonnet/prometheus-operator/**-crd.libsonnet` as well?

mxinden: @squat Good point, will follow up.

Makefile
```diff
@@ -215,7 +215,7 @@ $(JB_BINARY):
 $(PO_CRDGEN_BINARY): cmd/po-crdgen/main.go pkg/client/monitoring/v1/openapi_generated.go
 	go install github.com/coreos/prometheus-operator/cmd/po-crdgen
 
-$(PO_DOCGEN_BINARY): cmd/po-docgen/**.go
+$(PO_DOCGEN_BINARY): $(shell find cmd/po-docgen -type f)
```
brancz: I think this also has a dependency on `pkg/client/monitoring/v1/types.go`.

mxinden:
@brancz @squat Thanks for the feedback. Could you take another look?

brancz:
lgtm on green

brancz:
There are some jsonnet changes that were just merged, which is why the first job fails, just re-generate locally and add an additional commit.

PR #1627, opened 2018-07-17
Fix find commands in Makefile.

whypro:
I run `make build` command in Mac OS. But it is failed with errors below:
```
find: illegal option -- t
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
```

I install gnu find and try again. It also doesn't work and the output is: 

```
find -type f paths must precede expression
```

So I move `[path]` in front of `-type f` of `find` command in Makefile to solve this problem.


brancz:
cc @mxinden

PR #1632, opened 2018-07-17
helm: Add grafana ldap configuration

dzoeteman:
Allow LDAP secret for grafana LDAP authentication. This only allows a secret to be provided, not created, since it's not a great idea to have this in a configmap or checked in to your code at all, since it contains a bind user's password.


metalmatze:

PR #1634, opened 2018-07-18
Change rules to list

vlal:
* With new CRDs the value should be a list. Otherwise helm complains


vlal:
@gianrubio Any updates?

vsliouniaev:
Fixes https://github.com/coreos/prometheus-operator/issues/2025

PR #1640, opened 2018-07-19
pkg: add flag to toggle CRD creation in operator

squat:
In certain Prometheus Operator deployment scenarios it is desirable to
manage CRD creation outside of the operator. Likewise, it can be
desirable to scope the permissions of the Prometheus Operator so that it
does not have cluster-level access. This commit enables operation in
these situations by adding a flag to configure whether or not the
Prometheus Operator should try to create CRDs itself.


brancz:
lgtm :+1: nice work!

brancz:
We will need to cherry-pick these changes into the `release-0.22` release branch and cut a `v0.22.1` release.

squat:
regenerated contrib dir to make CI happy

PR #1655, opened 2018-07-20
Make localhost for local pods configurable via operator flag

allen13:
Add flag for fixing localhost resolve issue: https://github.com/coreos/prometheus-operator/issues/1646


brancz:
Thanks for reporting and fixing this! :slightly_smiling_face: We have the same case in the Alertmanager: https://github.com/coreos/prometheus-operator/blob/aa0166b0ee0c7efd8b99b9d9f08807036626fa5c/pkg/alertmanager/statefulset.go#L209

Would you mind fixing that as well? Thanks!

allen13:
Changed local reload host to simply localhost in an attempt to generalize as I added the variable to the Thanos sidecar as well. I did a search for any additional places where this might help as well.

brancz:
lgtm :+1: thanks!

PR #1656, opened 2018-07-20
pkg,test: fix gofmt and much golint

squat:
This commit fixes the gofmt warning and many of the golint warnings. We
still need to add comments for many exported types.

cc @mxinden @brancz


pkg/prometheus/rule_configmap_migration.go
```diff
@@ -97,10 +97,12 @@ func (c *Operator) getRuleCMs(ns string, cmLabelSelector *metav1.LabelSelector)
 	}
 
 	configMaps := []*v1.ConfigMap{}
-
 	err = cache.ListAllByNamespace(c.cmapInf.GetIndexer(), ns, cmSelector, func(obj interface{}) {
 		configMaps = append(configMaps, obj.(*v1.ConfigMap))
 	})
+	if err != nil {
```
squat: this was an actual bug; we were ignoring the error returned by `cache.ListAllByNamespace`

squat:
@mxinden I think it does make sense. Obvisously, this one-off PR won’t work well if we don’t because we’ll always be playing catch-up. Enforcing zero warnings is also a good way to remind ourselves not to export everything by default. Right now there are several undocumented functions that are depended on by other projects, eg cluster-monitoring-operator. We should create an issue to add linting to our test pipeline.

PR #1661, opened 2018-07-23
Only alert for nodes that currently exist

aleerizw:
Based on the discussion in https://github.com/coreos/prometheus-operator/issues/1180


brancz:
lgtm :+1: thanks!

PR #1666, opened 2018-07-23
Fix Node full alert condition

aleerizw:


aleerizw:
This fixes tla regression caused by https://github.com/coreos/prometheus-operator/pull/1639 ( < 0 ) was mistakenly removed.

faheem-cliqz:
This caused 100s of alerts on our side for node full, but was a test cluster, so all is well :).

PR #1671, opened 2018-07-23
Fix typo, point to helm install

mrueg:
If people look at the readme they most likely have no clue what helm is, give them a small pointer.


brancz:
Thanks!

PR #1678, opened 2018-07-24
Format yaml files correctly starting with ---

zioproto:
This is very important for people that want to integrate
these files in scripts and playbooks that call:

    cat *.yaml | kubectl apply -f -

The current syntax is robust against `kubectl apply -f .`
but a lot of scripts out there use the former solution
because for example in the folder they have also other files
that are not yaml.


brancz:
This fails because all manifests in `contrib/kube-prometheus/manifests` are auto-generated and checked in CI, that they are in the expected state. I don't think that adding arbitrary dividers makes sense here, just to make one of the possible scripts work, especially as we already have a script that does exactly this without having to modify the manifests themselves:

https://github.com/coreos/prometheus-operator/blob/d1d3ab12dec3c67ab12f531e53dba8772d7fd863/hack/concat-kubernetes-manifests.sh#L6

I recommend just using that :slightly_smiling_face: .

PR #1679, opened 2018-07-24
prometheus: Do not migrate rule config map for statefulset

mxinden:
If a user sets the RuleSelector property of a Prometheus custom resource
to an empty object (`{}`), all config maps in that namespace will be
selected for migration, including the Prometheus config map used to
mount all PrometheusRules into the Prometheus statefulset. This patch
filters the above mentioned config map out.

Fixes https://github.com/coreos/prometheus-operator/issues/1676


brancz:
This lgtm, but it would be good to have a confirmation from @carlpett .

mxinden:
Will direct this PR to `release-0.15` instead of `master`.

pkg/prometheus/rule_configmap_migration.go
```diff
@@ -102,7 +102,26 @@ func (c *Operator) getRuleCMs(ns string, cmLabelSelector *metav1.LabelSelector)
...
+	// to an empty object (`{}`), all config maps in that namespace will be
+	// selected for migration, including the Prometheus config map used to mount
+	// all PrometheusRules into the Prometheus statefulset. The following filters
+	// this config map out.
+	cmsNotGeneratedByPO := []*v1.ConfigMap{}
+	for _, cm := range configMaps {
+		migrate := true
+		for key, value := range cm.Labels {
+			if key == managedByOperatorLabel && value == managedByOperatorLabelValue {
+				migrate = false
```
squat: just one small optimization: you could add a `break` here to exit the loop early

squat:
lgtm on green and pending @carlpett's thumbsup

carlpett:
Hah, here I sat waiting for Travis to run the "push docker" step. But it doesn't actually push on PRs.
Reporting back soon, building locally...

carlpett:
Yep, looks good here! :+1:

PR #1681, opened 2018-07-24
cmd/operator: add process collector

squat:
As noted in
https://github.com/openshift/cluster-monitoring-operator/pull/52#discussion_r204747221,
the Prometheus Operator should also register process metrics.

cc @brancz


brancz:
lgtm on green

PR #1682, opened 2018-07-24
Update GKE kubelet scraping docs

directionless:
Chatting in https://github.com/coreos/prometheus-operator/issues/1623 it sounds like prior to 1.11, downgrading to https _is_ the correct approach.

This adds the jsonnet approach to to the docs, and adds a reference to it from the README

### DIFF
--- a/contrib/kube-prometheus/docs/GKE-cadvisor-support.md
+++ b/contrib/kube-prometheus/docs/GKE-cadvisor-support.md
@@ -1,7 +1,22 @@
-# Kubelet / cAdvisor special configuration updates for GKE 
+# Kubelet / cAdvisor special configuration updates for GKE
 
-In order to allow Prometheus to access the endpoints provided by the kubelet/cAdvisor on GKE we have to downgrade the scheme to HTTP (from HTTPS).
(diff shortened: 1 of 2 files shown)


contrib/kube-prometheus/docs/GKE-cadvisor-support.md
```diff
@@ -1,7 +1,22 @@
...
+You can configure this behavior through kube-prometheus with:
+```
+local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
+    (import 'kube-prometheus/kube-prometheus-insecure-kubelet.libsonnet') +
+	{
+        _config+:: {
+		...
+		}
+    };
+{ ['kube-state-metrics-' + name]: kp.kubeStateMetrics[name] for name in std.objectFields(kp.kubeStateMetrics) }
```
brancz: this line is probably more confusing then helpful here, I think we can just remove it

directionless: Done!

I wasn't sure how much jsonnet knowledge to assume in these doc snippets.

brancz:
Thanks!

PR #1697, opened 2018-07-25
client-go/informers: Add namespace indexer to SharedIndexInformers

mxinden:
`client-go/tools/cache.ListAllByNamespace` tries to search the given
cache by the namespace index. If that index is not present, it logs a
warning message and iterates over the objects itself [1].

Initially to remove the log line, but as a side effect also to improve query
speed, this patch adds the namespace index to all used SharedIndexInformers in
the Prometheus and Alertmanager operator.

[1]
```golang
items, err := indexer.Index(NamespaceIndex, &metav1.ObjectMeta{Namespace: namespace})
if err != nil {
	// Ignore error; do slow search without index.
	glog.Warningf("can not retrieve list of objects using index : %v", err)
	for _, m := range indexer.List() {
```

Fixes https://github.com/coreos/prometheus-operator/issues/1694

//CC @prune998


brancz:
Nice! lgtm :+1: 

Also that was a nice deep dive into the informer infrastructure! :smile:

PR #1698, opened 2018-07-25
Adding a nodeSelector to the exporter-node daemonset

dwardu89:
The DaemonSet yaml for the node exporter only allows tolerations. However when taints are not being used on nodes, the requirement for a nodeSelector arises.


dwardu89:
@brancz is any possibility of looking at the pull request yourself please?

Issue #1673, opened 2018-07-23
Thanos options for S3: s3.encrypt-sse as environment var


astub:
Is there a way we can open up the Thanos spec to add environment variables flags? It would be really nice as Thanos adds more options or if they change. 

I really need to be able to enable s3.encrypt-sse, but there appears to be no way of doing that at the moment. Correct me if i'm wrong.

brancz:
That's something that we would have to add to the `ThanosSpec` here: https://github.com/coreos/prometheus-operator/blob/b0f98a839fd3b9e04fad5602e426bda6f99fc904/pkg/client/monitoring/v1/types.go#L258-L271

And then based on that add the flag like here: https://github.com/coreos/prometheus-operator/blob/b0f98a839fd3b9e04fad5602e426bda6f99fc904/pkg/prometheus/statefulset.go#L634-L636

Would you like to create a pull request for this? :slightly_smiling_face:

astub:
should this issue be closed?

mxinden:
Yes, thanks for the reminder @astub and the contribution.

PR #1700, opened 2018-07-25
thanos-s3 spec added encrypt-sse

astub:
updated thanos sidecar spec to be able to use server side encrypt.

Issue #1673


brancz:
lgtm :+1: thanks!

PR #1701, opened 2018-07-26
[helm] PrometheusLabelValue Can Change Prometheus Name

midnightconman:
Older versions of the operator had a different way of naming resources created. Previously the prometheus pods were called `mcon-prometheus-0`, now they are called `prometheus-mcon-prometheus-0`... this change allows for the older behavior, if using `.Values.prometheusLabelValue`.


gianrubio:

PR #1707, opened 2018-07-26
pkg: return err

squat:
We are mistakingly not returning an error and allowing a function to
continue.

cc @mxinden @brancz


mxinden:
Looks good to me. Great catch. Thanks for the fix.

Does that relate to the cluster monitoring operator failures today?

squat:
@mxinden unfortunately I think not :/ I did find it while hunting down the cause of the CMO issues.

PR #1710, opened 2018-07-27
alertmanager: Configurable data retention

### LINKED ISSUE #1055: Alertmanager's data.retention option is not configurable
Unless I am mistaken, there is no way to customize the `--data.retention` flag for the Alertmanager `StatefulSet` which defaults to 120 hours.

https://github.com/coreos/prometheus-operator/blob/dedb1f41e6c9229af05b88e16dc6cced39e2f0d6/pkg/alertmanager/statefulset.go#L165

Ideally, the `Alertmanager` CRD should support the same `retention` field as the Prometheus CRD.
For example, to set `retention: 30d`.

This is mostly for capacity planning on the `storage` configuration and the `requests.storage` value of the `volumeClaimTemplate`.

osela:
Fixes #1055.
I used the default data retention of Alertmanager as the default for the operator. Let me know if it doesn't make sense.


osela:
@mxinden Added unit tests for the retention option for both Alertmanager and Prometheus

PR #1718, opened 2018-07-31
prometheus: Improve metrics and stop reconciling on no update

mxinden:
- Introduce the 'prometheus_operator_triggered_total'
Prometheus metric counter, differentiated in label dimensions:
  - triggered_by
    - monitoringv1.PrometheusesKind
    - monitoringv1.ServiceMonitorsKind
    - monitoringv1.PrometheusRuleKind
    - "ConfigMap"
    - "Secret"
    - "StatefulSet"
    - "Namespace"
  - action
    - "add"
    - "delete"
    - "update"

- Add increase hooks of above mentioned metric to all Kubernetes
informer handlers and add simple log line on debug level.

- Add condition to Kubernetes informer update handlers to only preceed
when previous resource version is not equal to new resource version.
This will prevent triggering a reconciliation loop on informer resyncs.

Relates to https://github.com/coreos/prometheus-operator/issues/1659.

### DIFF
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -98,7 +98,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
                             policyRule.withResources([
                               'namespaces',
                             ]) +
-                            policyRule.withVerbs(['list', 'watch']);
+                            policyRule.withVerbs(['list', 'watch', 'get']);
 
       local rules = [apiExtensionsRule, monitoringRule, appsRule, coreRule, podRule, routingRule, nodeRule, namespaceRule];
 
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -98,7 +98,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
                             policyRule.withResources([
                               'namespaces',
                             ]) +
-                            policyRule.withVerbs(['list', 'watch', 'get']);
+                            policyRule.withVerbs(['list', 'watch']);
 
       local rules = [apiExtensionsRule, monitoringRule, appsRule, coreRule, podRule, routingRule, nodeRule, namespaceRule];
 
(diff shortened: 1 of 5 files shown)


mxinden:
This should partially reduce the CPU footprint of the Prometheus Operator. I can not tell if it eliminates the root cause for https://github.com/coreos/prometheus-operator/issues/1659.

pkg/prometheus/operator.go
```diff
@@ -72,6 +72,12 @@ type Operator struct {
 
 	statefulsetErrors prometheus.Counter
 
+	// triggerByCounter is a map of maps of Prometheus counters. The first map
+	// represents each Kubernetes object, the second map represents its actions
+	// (add, delete, update). These metrics help understand what triggers the
+	// Prometheus Operator to reconcile its created objects
+	triggerByCounter map[string]map[string]prometheus.Counter
```
brancz: What you are doing here is equivalent to a `CounterVec`, we should use that instead.

pkg/prometheus/operator.go
```diff
@@ -622,14 +697,72 @@ func (c *Operator) enqueue(obj interface{}) {
 	c.queue.Add(key)
 }
 
-// enqueueForNamespace enqueues all Prometheus object keys that belong to the given namespace.
-func (c *Operator) enqueueForNamespace(ns string) {
-	cache.ListAll(c.promInf.GetStore(), labels.Everything(), func(obj interface{}) {
+// enqueueForNamespace enqueues all Prometheus object keys that belong to the
+// given namespace or select objects in the given namespace.
+func (c *Operator) enqueueForNamespace(nsName string) {
+	nsObject, exists, err := c.nsInf.GetStore().GetByKey(nsName)
```
mxinden: We could also get the namespace via kclient, happy for thoughts on this.

brancz: I think it's fine like this

jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -98,7 +98,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
                             policyRule.withResources([
                               'namespaces',
                             ]) +
-                            policyRule.withVerbs(['list', 'watch']);
+                            policyRule.withVerbs(['list', 'watch', 'get']);
```
brancz: I don't see why this is necessary

brancz:
I don't think the rbac changes are necessary, otherwise lgtm

mxinden:
Thanks for the review @brancz, removed the rbac changes.

Issue #1319, opened 2018-05-08
Rule evaluation failures for alertmanager.rules


bgagnon:
Two of the `kube-prometheus` rules from `alertmanager.rules` fail to evaluate: `AlertmanagerConfigInconsistent` and `AlertmanagerDownOrMissing`.

The error message is `many-to-many matching not allowed: matching labels must be unique on one side`.

Here are the full log entries:

```
level=warn
ts=2018-05-08T19:09:30.012163531Z
caller=manager.go:339
component="rule manager"
group=alertmanager.rules
msg="Evaluating rule failed"
rule="alert: AlertmanagerConfigInconsistent\nexpr: count_values by(service) (\"config_hash\", alertmanager_config_hash) / on(service)\n  group_left() label_replace(prometheus_operator_alertmanager_spec_replicas, \"service\",\n  \"alertmanager-$1\", \"alertmanager\", \"(.*)\") != 1\nfor: 5m\nlabels:\n  severity: critical\nannotations:\n  description: The configuration of the instances of the Alertmanager cluster `{{$labels.service}}`\n    are out of sync.\n  summary: Configuration out of sync\n" err="many-to-many matching not allowed: matching labels must be unique on one side"
```
and:
```
level=warn
ts=2018-05-08T19:09:30.013795976Z
caller=manager.go:339
component="rule manager"
group=alertmanager.rules
msg="Evaluating rule failed"
rule="alert: AlertmanagerDownOrMissing\nexpr: label_replace(prometheus_operator_alertmanager_spec_replicas, \"job\", \"alertmanager-$1\",\n  \"alertmanager\", \"(.*)\") / on(job) group_right() sum by(job) (up) != 1\nfor: 5m\nlabels:\n  severity: warning\nannotations:\n  description: An unexpected number of Alertmanagers are scraped or Alertmanagers\n    disappeared from discovery.\n  summary: Alertmanager down or missing\n" err="many-to-many matching not allowed: matching labels must be unique on one side"
```

bgagnon:
I wish I knew how to fix this and submit a PR... but my PromQL is weak at best :/ I _think_ the issue is with the `group_right()` clause, but I'm not sure what the problem is exactly.

nabadger:
@bgagnon did you manage to resolve this in the end? 

We are hitting this as well.

It seems to be related to `prometheus_operator_alertmanager_spec_replicas` 

[...]

PR #1721, opened 2018-07-31
Update Alertmanager-check rules to support multiple namespaces

nabadger:
This helps to resolve an issue identified in #1319 

It's feasible to have multiple `Alertmanager` CRD's in different namespaces. If you do, the alert-rule fails with `many-to-many matching not allowed: matching labels must be unique on one side`.

In order for this PR to work fully, you also need to set `honorLabels: true` for the prometheus-operator `ServiceMonitor` (will attach PR soon)


brancz:
This needs a re-generate run on top of is, just like @mxinden explained here: https://github.com/coreos/prometheus-operator/pull/1722#issuecomment-409486865

mxinden:
@nabadger I am not entirely sure how the helm charts are generated, but I would greatly appreciate, if you could also add this change to the `.libsonnet` file [here](https://github.com/coreos/prometheus-operator/blob/8cd343a9132170c2e644b73ee8bf792d1a53f8f0/contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/alertmanager.libsonnet#L14) and follow the [contribution guide](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus#contributing).

Thanks for the help.

nabadger:
@mxinden - following the contrib docs, I run into issues here. 

I've made changes to the file in question, committed and pushed the file, but then when I run `make generate-in-docker` it eventually fails with:

```
HEAD is now at 2af790e8... make format generate-in-docker
rm -rf manifests
./build.sh
+ set -o pipefail
+ rm -rf manifests
+ mkdir manifests
+ jsonnet -J vendor -m manifests example.jsonnet
+ xargs '-I{}' sh -c 'cat {} | gojsontoyaml > {}.yaml; rm -f {}' -- '{}'
RUNTIME ERROR: couldn't open import "kube-prometheus/kube-prometheus.libsonnet": no match locally or in the Jsonnet library paths.
	example.jsonnet:1:13-63	thunk <kp>
	example.jsonnet:7:81-83	thunk <o>
	std.jsonnet:1153:24
	std.jsonnet:1153:5-33	function <anonymous>
	example.jsonnet:7:64-99	thunk <a>
	example.jsonnet:7:1-101	function <anonymous>
	example.jsonnet:7:1-101
Makefile:30: recipe for target 'manifests' failed
make[1]: *** [manifests] Error 1
make[1]: Leaving directory '/go/src/github.com/coreos/prometheus-operator/contrib/kube-prometheus'
Makefile:93: recipe for target 'kube-prometheus' failed
make: *** [kube-prometheus] Error 2
```

squat:
that error suggests that the external jsonnet dependencies were not installed. running `make vendor` should fix this, though this target should have been executed as a dependency of the `generate` target as well.

brancz:
I think this might be an unfortunate commit that you are basing this work on. Last week we for a short amount of time were pinning wrong versions of jsonnet bundler and jsonnet dependencies. Just to make sure, could you rebase your changes?

paskal:
@gianrubio can this be merged?

brancz:
There are 53 commits in this PR, can you distill them into the important commit, as it seems most commits do not belong in this PR.

nabadger:
Yeah it's been a while - I'll rebase from master. There's actually conflicts because some of the metric names have changed, `prometheus_operator_alertmanager_spec_replicas`  -> `prometheus_operator_spec_replicas` for example.

PR #1730, opened 2018-08-01
add serviceMonitorNamespaceSelector to the prometheus jsonnet library

nrobert13:
to implement the feature from PR #1227


contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
```diff
@@ -167,6 +167,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
           baseImage: $._config.imageRepos.prometheus,
           serviceAccountName: 'prometheus-' + $._config.prometheus.name,
           serviceMonitorSelector: selector.withMatchExpressions({ key: 'k8s-app', operator: 'Exists' }),
+          serviceMonitorNamespaceSelector: selector.withMatchExpressions({ key: 'prometheus', operator: 'In', values: ['yes', 'true'] }),
```
brancz: the default here should probably be `{}`, which just selects all namespaces

brancz:
Thanks! Did you want this to end up in the rendered manifests as well? In that case you need to run `jb update` in the `contrib/kube-prometheus/` directory, and then run `make generate-in-docker` from the root of the repo and then commit the changes. If not I'm happy to merge this :slightly_smiling_face: .

brancz:
Perfect :+1:

PR #1708, opened 2018-07-27
Prune/jsonnet namespace roles

prune998:
This PR replace the Jsonnet config that create the RBAC roles for `default` and `kube-system` namespaces by a loop that will generate the RBAC rules for a list of namespaces.
The default list includes `default`, `kube-system` and the Prometheus install namespace (`monitoring` by default).

I also added a simple doc file.

PR #1733, opened 2018-08-01
Change roleBindingSpecificNamespaces to use $._config.namespace

vasrem:
There is an issue which leads to the creation of RoleBindings which are correlated with ServiceAccounts which don't exist. This change will allow each RoleBinding to be attached to the ServiceAccount created on [this](https://github.com/coreos/prometheus-operator/blob/bb1966b2a1dc737d67f8f72983c747f0b5732947/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet#L29-L33) section.

The issue was created after merging #1708.


brancz:
You're absolutely correct. Could you just run `make generate-in-docker` from the root of the repo and commit the changes? Thanks!

vasrem:
I have already run this but I get nothing new to commit.

brancz:
I pushed another commit on top of your branch that re-generates and updates the jsonnet dependencies. This should be good now, let's just see CI pass :slightly_smiling_face: .

brancz:
Oh. It actually looks like we broke jsonnet-bundler today. It's time to sleep over here. I'll try to fix everything in the morning.

brancz:
Thanks a lot for taking care of this @vasrem! Highly appreciated!

PR #1734, opened 2018-08-02
storage docs

kanolato:
The `storageClassName` field was missing


mxinden:
Thanks for the PR! What is the reasoning for deleting the other properties?

kanolato:
@mxinden  just to match the example https://github.com/coreos/prometheus-operator/blob/master/example/storage/persisted-prometheus.yaml

kanolato:
I added a similar line on 18, to do the same on the storageClass.

brancz:
Thanks!

PR #1737, opened 2018-08-02
prometheus: Use hashstructure package for input checksum

mxinden:
Instead of marshalling all StatefulSet inputs to JSON and hashing that
string, use hashstructure [1] package. In addition do not generate new
rule config maps if underlying rules did not change.

Related to https://github.com/coreos/prometheus-operator/issues/1659.

[1] github.com/mitchellh/hashstructure


Gopkg.lock
```diff
@@ -2,476 +2,379 @@
 
 
 [[projects]]
-  digest = "1:8e47871087b94913898333f37af26732faaab30cdb41571136cf7aec9921dae7"
```
mxinden: These were introduced by https://github.com/coreos/prometheus-operator/pull/1691. I am using dep `v0.4.1`. `v0.5.0` has just been released. Let's see if CI complains.

pkg/prometheus/operator.go
```diff
@@ -969,18 +968,22 @@ func (c *Operator) sync(key string) error {
...
-		struct {
-			P monitoringv1.Prometheus
-			C Config
-			R []string
-		}{p, c, ruleConfigMapNames},
+	hash, err := hashstructure.Hash(struct {
+		P monitoringv1.Prometheus
+		C Config
+		R []string `hash:"set"`
+	}{p, c, ruleConfigMapNames},
```
s-urbaniak: sorry for a newbie question: does `ruleConfigMapNames` always maintain the set invariant? Do we make sure this is the case already?

pkg/prometheus/operator.go
```diff
@@ -969,18 +968,22 @@ func (c *Operator) sync(key string) error {
 }
 
 func createSSetInputChecksum(p monitoringv1.Prometheus, c Config, ruleConfigMapNames []string) (string, error) {
```
s-urbaniak: minor nit: the function states it creates a checksum, while we calculate a hash. While similar in nature, the purpose of a checksum (error detection) vs. a hash (uniformity) is slightly different. Maybe we can change the func name to reflect this fact?

pkg/prometheus/operator.go
```diff
@@ -969,18 +968,22 @@ func (c *Operator) sync(key string) error {
...
+		C Config
+		R []string `hash:"set"`
+	}{p, c, ruleConfigMapNames},
+		nil,
 	)
 	if err != nil {
-		return "", errors.Wrap(err, "failed to marshal Prometheus CRD and config to json")
+		return "", errors.Wrap(
+			err,
+			"failed to calculate combined hash of Prometheus CRD, config and"+
```
squat: nit: i think gofmt likes to put whitespace between operators and operands

pkg/prometheus/rules.go
```diff
@@ -224,15 +222,14 @@ func makeRulesConfigMaps(p *monitoringv1.Prometheus, ruleFiles map[string]string
 		{},
 	}
 	currBucketIndex := 0
-	sortedNames := sortKeyesOfStringMap(ruleFiles)
 
-	for _, filename := range sortedNames {
+	for filename, filecontent := range ruleFiles {
 		// If rule file doesn't fit into current bucket, create new bucket
-		if bucketSize(buckets[currBucketIndex])+len(ruleFiles[filename]) > maxConfigMapDataSize {
+		if bucketSize(buckets[currBucketIndex])+len(filecontent) > maxConfigMapDataSize {
```
squat: I think we need to run gofmt on this file. Is that possible?

pkg/prometheus/operator.go
squat: what does the extra `S` mean in this func name?

pkg/prometheus/operator.go
```diff
@@ -969,18 +968,22 @@ func (c *Operator) sync(key string) error {
...
+				" rule config map names",
```
squat: nit: in kubernetes it is a `configmap`, not a `config map`

pkg/prometheus/operator.go
mxinden: Not sure I understand the question correctly: Treating `ruleConfigMapNames` as a set instead of an array is enforced via the struct tag `R []string `hash:"set"``.

pkg/prometheus/operator.go
mxinden: Good idea. Changed across the project.

pkg/prometheus/operator.go
mxinden: Thought so as well, but running `[space]+` through `gofmt` results in `+`.

pkg/prometheus/rules.go
mxinden: My editor does so on every save. Does it complain on your version?

pkg/prometheus/operator.go
s-urbaniak: Sorry for being imprecise: do we make sure that `ruleConfigMapnames` never contains duplicate enries? If they do, they would hash to the same value when treated as a set of strings.

pkg/prometheus/operator.go
s-urbaniak: maybe a shorter error message would suffice, i.e. `hashing of prometheus CRD, config, and configmap names failed`?

mxinden:
@squat replaced all `config map` with `ConfigMap` in the entire repo. Thanks for the hint.

pkg/prometheus/operator.go
s-urbaniak: I guess `createSSetInputChecksum` expands to `createS[tateful]SetInputChecksum`. How about introducing:

```
type statefulSetConfig struct {
  prom monitoringv1.Prometheus
  operatorConfig Config
  ruleConfigMapNames []string
}
```
and implement a `Hash() (string, error)` as well as a `StatefulSet() (*appsv1.StatefulSet, error)`  method which would also reduce the amount of parameters: https://github.com/coreos/prometheus-operator/blob/cc98874a6df78cfb0d0fdb8edac3ae8d0ef68739/pkg/prometheus/statefulset.go#L80-L86

pkg/prometheus/rules.go
squat: looks like it's context dependent [0]

[0] https://play.golang.org/p/9ZbTzwNUHU7

pkg/prometheus/operator.go
mxinden: Yes we do here: https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/rules.go#L241

In addition it is enforced by Kubernetes, as there can not be multiple ConfigMaps with the same name in the same namespace.

pkg/prometheus/operator.go
mxinden: Shortened :+1:

pkg/prometheus/operator.go
mxinden: I went through a couple of iterations on this idea, but ended up bursting the scope of this PR. I will follow up with you on this @s-urbaniak if you don't mind.

> what does the extra S mean in this func name?

@squat `SSet` stands for a StatefulSet.

s-urbaniak:
Of course, sorry for distracting from the actual PR target. We can pair any
time on this one.

Max Inden <notifications@github.com> schrieb am Do., 2. Aug. 2018, 19:51:

PR #1744, opened 2018-08-03
prometheus: Fix wrong metric increase

mxinden:
Instead of increasing the Prometheus Rule trigger counter, the
ServiceMonitor trigger counter was increased.


brancz:
Oops. Good catch. lgtm on green

mxinden:
@brancz My bad. Thanks for the quick review.

PR #1749, opened 2018-08-05
pkg: don't rely on ns informer if ns selected

squat:
If the operator is not watching all namespaces, then the namespace
informer is not started. This means that all operations that depend on
listing namespaces based on the informer will not work. This commit
modifies those operations so that they work in a single-namespace
configuration.

cc @mxinden @brancz 

~~Note: still needs a little more tweaking; there is some configuration that is causing the Prometheus-k8s instances to crashloop.~~ Disregard, crashlooping was due to a manually deployed alerting rule that was missing quotes around a string.


squat:
@mxinden it’s the error we were discussing Friday: the Prometheus container itself is failing to parse the configuration and errororing. I’m doing some digging to determine what config change could be causing the error.

brancz:
If possible can you share the generated config plus the error you are getting?

PR #1750, opened 2018-08-06
pkg/prometheus: Remove rule ConfigMap migration

mxinden:
Remove Prometheus rule ConfigMap live migration code introduced in #1399
[1].

[1] https://github.com/coreos/prometheus-operator/pull/1399

I would like to remove the live migration with the next release. What do you think?


anandsinghkunwar:
Was there a specific reason why this was removed? Is there a way to do it by some flag maybe in the newer versions? 
I am planning to bump from 0.19 to the 0.25, should I follow 0.19=>0.20=>0.25?

brancz:
There is still the command line tool to perform migrations available. It was mentioned in the release notes of the release that intoduced the Prometheus rule object.

We just removed the migration to keep the Prometheus operator start up less involved.

PR #1761, opened 2018-08-07
Fix typos on Alertnames

sfrique:


contrib/kube-prometheus/manifests/prometheus-rules.yaml
```diff
@@ -849,7 +849,7 @@ spec:
       for: 10m
       labels:
         severity: warning
-    - alert: PrometheusTargetScapesDuplicate
+    - alert: PrometheusTargetScrapesDuplicate
```
brancz: As this is generated from the jsonnet files, you need to either bump [the commit used to generate](https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/jsonnetfile.lock.json#L11), or revert this to get CI to be green. If you want to bump the commit, then you need to do this in a separate commit on top of your current commit, and put your current commit in there.

sfrique:
@brancz Thanks!
Fixed.

brancz:
Perfect :+1:

PR #1772, opened 2018-08-08
Prune/jsonnet roles

prune998:
When overriding the `config.prometheus.namespaces` value, the namespace used to deploy the Operator is not added to the list. 
This PR solve this issue


brancz:
In case someone looks at this PR and wonders why it was closed, the jsonnet code already adds the "current" namespace: https://github.com/coreos/prometheus-operator/blob/416ed2b8d17b8122be354acae08c43fa2c7565b4/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet#L24

prune998:
@brancz the whole point here is, if you specifically set the `namespaces`, let's say to add your `dev` namespace, if will override the defaults values you're showing here.
This PR was to *FORCE* the addition of the `$._config.namespace` to the user-provided list.

You're free to re-open the PR or make it clearly clear in the documentation before some users opens issue about that behaviour :)

prune998:
I closed this PR as it was mixing too much stuff. The clean one is https://github.com/coreos/prometheus-operator/pull/1773#issuecomment-412494829

PR #1773, opened 2018-08-08
SpecificNamespaces includes config.namespace

prune998:
When overriding the `config.prometheus.namespaces` value with your own, the namespace used to deploy the Operator is not added to the list of `roles`/`roleBindings`.
This PR solve this issue


brancz:
`$._config.prometheus.namespaces` already contains the "own" namespace, can you elaborate what's going wrong in your setup?

prune998:
@brancz the whole point here is, if you specifically set the namespaces, let's say to add your dev namespace, if will override the defaults values you're showing here.
This PR was to FORCE the addition of the $._config.namespace to the user-provided list.

brancz:
In such a case you should add your additional namespace, not remove the old ones, meaning something like this:

```
(import "kube-prometheus/kube-prometheus.libsonnet") + 
... +
... + {
  _config+:: {
    prometheus+:: {
      namespaces+: ['dev'],
    },
  },
}
```

prune998:
Ok thanks @brancz, I wasn't aware of this variables options... this needs to be added to the documentation then (maybe it is, but it would be welcomed to be added in this specific case).

You can safely close the PR then.

squat:
closing now that documentation in #1786 has landed

PR #1784, opened 2018-08-13
jsonnet: consistent dash in flags

squat:
The Prometheus Operator manifests all use double dashes for long flags
exept for the `logtostderr` flag, which uses one dash. This commit
adjusts the Prometheus Operator libsonnet to fix this. The
kube-prometheus project will need to be bumped later to pick up this
change.


brancz:
lgtm :+1:

mxinden:
@squat didn't know double dash `--logtostderr` was available :+1:  thanks for the cleanup!

PR #1785, opened 2018-08-13
pkg/*/statefulset.go: Do not mutate shared object

mxinden:
Users have reported high CPU usage of the Prometheus Operator when
adding an annotation to a Prometheus object. The Operator would update
the respective StatefulSet in an infinite loop.

Whether a given StatefulSet needs updating is determined by the hash of
the inputs needed to generate the StatefulSet, which is calculated and
then attached to the StatefulSet as an annotation. On subsequent
reconciliations this hash is compared to the hash of the new inputs.

The function to build the StatefulSet definition is passed the
Prometheus object. This is done by value, not by reference. This does
not enforce a deep copy but merely a shallow copy. In the build function
the new StatefulSet would inherit the annotation map of the Prometheus
object. Next the input hash would be added to this map, resulting in
both the Statefulset having the hash annotation, as intended, as well as
the Prometheus object (same map, shared as a reference).

On subsequent reconciliations the same Prometheus object is used to
calculate the input hash, this time accidentally containing the has
annotation from the previous run. Even though the actual inputs never
changed, this results in a new hash, thereby updating the StatefulSet,
...

The solution is to deep copy the Prometheus object before using it in
the StatefulSet build function, thereby never mutating the annotations
of the Prometheus object. Same measure is taken for the Alertmanager
StatefulSet build function.

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -693,6 +693,12 @@ func TestPrometheusOnlyUpdatedOnRelevantChanges(t *testing.T) {
 	name := "test"
 	prometheus := framework.MakeBasicPrometheus(ns, name, name, 1)
 
+	// Adding an annotation to Prometheus lead to high CPU usage in the past
+	// updating the Prometheus StatefulSet in a loop (See
+	// https://github.com/coreos/prometheus-operator/issues/1659). Added here to
+	// prevent a regression.
+	prometheus.Annotations["test-annotation"] = "test-value"
+
 	ctx, cancel := context.WithCancel(context.Background())
 
 	type versionedResource interface {
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -170,6 +170,10 @@ func makeStatefulSetService(p *monitoringv1.Alertmanager, config Config) *v1.Ser
 }
 
 func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.StatefulSetSpec, error) {
+	// Before editing 'a' create deep copy, to prevent side effects. For more
+	// details see https://github.com/coreos/prometheus-operator/issues/1659
+	a = a.DeepCopy()
+
 	tag := a.Spec.Version
 	if a.Spec.Tag != "" {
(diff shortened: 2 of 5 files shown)


mxinden:
Related to https://github.com/coreos/prometheus-operator/issues/1659

mxinden:
I would create a patch release once this is merged. What do you think @brancz @squat @s-urbaniak?

brancz:
> I would create a patch release once this is merged. What do you think @brancz @squat @s-urbaniak?

Sounds good.

squat:
This commit breaks the following test: https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset_test.go#L57

This is expected, because the test checks that labels and annotations are identical on the prometheus object and the statefulset, even though the statefulset gets new annotations: https://github.com/coreos/prometheus-operator/blob/33a84937182c3f2b159dd6a5e9501afc11826d87/pkg/prometheus/statefulset.go#L174.

We need to adjust the test to check that all prometheus object annotation are included in the statefulset annotations, not that they are equal.

test/e2e/prometheus_test.go
```diff
@@ -693,6 +693,15 @@ func TestPrometheusOnlyUpdatedOnRelevantChanges(t *testing.T) {
 	name := "test"
 	prometheus := framework.MakeBasicPrometheus(ns, name, name, 1)
 
+	// Adding an annotation to Prometheus lead to high CPU usage in the past
+	// updating the Prometheus StatefulSet in a loop (See
+	// https://github.com/coreos/prometheus-operator/issues/1659). Added here to
+	// prevent a regression.
+	if prometheus.Annotations == nil {
```
s-urbaniak: rather doing this here, I would advocate to do this in `framework.MakeBasicPrometheus(...)` so the caller does not need to worry about initialization.

s-urbaniak:
> I would create a patch release once this is merged. What do you think @brancz @squat @s-urbaniak?
sounds fine with me :heavy_check_mark:

squat:
lgtm on green!

mxinden:
Thanks for the feedback!

PR #1790, opened 2018-08-13
*: Cut v0.23.1 back to master

mxinden:
Let me know if you want a full merge instead of the cherry-pick.


mxinden:
@brancz @squat should I also update kube-prometheus?

PR #1806, opened 2018-08-17
kube-prometheus: bind mount host rootfs into node exporter

### LINKED ISSUE #569: node-exporter could not export all mountpoints
node-exporter in this yaml  :

prometheus-operator/contrib/kube-prometheus/manifests/node-exporter/node-exporter-daemonset.yaml

volumes and mountpoints doesn't content : 

   "/:/rootfs:ro" 

not like node-exporter:
https://github.com/prometheus/node_exporter

why not add the to it?

s-urbaniak:
Fixes https://github.com/prometheus/node_exporter/issues/474 in the operator
Fixes #569 (by making the mount explicit)

### DIFF
--- a/Documentation/user-guides/cluster-monitoring.md
+++ b/Documentation/user-guides/cluster-monitoring.md
@@ -166,6 +166,8 @@ spec:
         - --web.listen-address=127.0.0.1:9101
         - --path.procfs=/host/proc
         - --path.sysfs=/host/sys
+        - --collector.filesystem.ignored-mount-points=^/(dev|proc|sys|var/lib/docker/.+)($|/)
+        - --collector.filesystem.ignored-fs-types=^(autofs|binfmt_misc|cgroup|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|mqueue|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|sysfs|tracefs)$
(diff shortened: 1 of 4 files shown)


squat:
A lightbulb just turned on: we should keep in mind that the `mountPropagation` field is a beta feature and consider its use when making decisions about taking the operator to 1.0.0.

s-urbaniak:
@squat good point, yes :+1:

brancz:
> A lightbulb just turned on: we should keep in mind that the mountPropagation field is a beta feature and consider its use when making decisions about taking the operator to 1.0.0.

kube-prometheus just happens to be part of this repository but is not versioned along with the Prometheus Operator so I don't think it's relevant. It is relevant however for the whole monitoring stack to be considered stable.

s-urbaniak:
hmm, I believe CI failed, because I rewrote the commit message which in turn changed the commit hash, let me rebundle the jsonnet files.

s-urbaniak:
CI resolved @brancz @squat PTAL

AndorCS:
Looks like you forgot to specify `--path.rootfs` command-line argument.

Andor:
Looks like you forgot to specify `--path.rootfs` command-line argument.

brancz:
@Andor you're absolutely right, there is already a pull request out to fix this: https://github.com/coreos/prometheus-operator/pull/2285

PR #1813, opened 2018-08-21
*: add multi-namespace support to Prometheus Operator

squat:
This is a large PR broken into two logical commits that can be separate PRs if needed.
I have tested these changes on an Openshift cluster as part of https://github.com/openshift/cluster-monitoring-operator with several different configurations:
* single namespace, i.e. `--namespaces=x`
* multiple namespaces, i.e. `--namespaces=x,y` and `--namespaces=x --namespaces=y`
* all namespaces, i.e. no `--namespaces` flag so `v1.NamespaceAll` is used

First:
This commit adds multi-namespace support to the Prometheus Operator. PO
now allows all informers to react to changes to multiple, individually
configured, namespaces, i.e. without specifying v1.NamespaceAll. This is
important for cases where the Prometheus Operator does not have
privileges to List all namespaces but should still administer multiple
namespaces. This capability is accomplished by abstracting the
cache.ListerWatchers for the various informers the PO uses.

Second:
This commit ensures that all informer caches are synced before the
eventhandlers are added. This is important whenever one informer's
eventhandler depends on the cache of a different informer. Without this,
eventhandlers will complain that certain objects don't exist. This is
important now that the namespace informer still runs for a single
namespace, where previously it was only run for v1.NamespaceAll.
Otherwise, e.g. the secret eventhandler could fail until the namespace
handler had synced.

cc @brancz @mxinden @s-urbaniak

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -55,6 +55,40 @@ const (
 )
 
 var (
+	ns namespaces
+)
+
+type namespaces map[string]struct{}
+
+// Set implements the flagset.Value interface.
+func (n namespaces) Set(value string) error {
+	ns := strings.Split(value, ",")
+	if n == nil {
+		n = namespaces{}
+	}
+	for i := range ns {
+		n[ns[i]] = struct{}{}
+	}
+	return nil
+}
+
+// String implements the flagset.Value interface.
+func (n namespaces) String() string {
+	return strings.Join(n.asSlice(), ",")
+}
+
+func (n namespaces) asSlice() []string {
+	var ns []string
+	for k := range n {
+		ns = append(ns, k)
+	}
+	if len(ns) == 0 {
+		ns = append(ns, v1.NamespaceAll)
+	}
+	return ns
+}
+
+var (
 	cfg                prometheuscontroller.Config
 	availableLogLevels = []string{
 		logLevelAll,
@@ -88,7 +122,7 @@ func init() {
 	flagset.StringVar(&cfg.AlertmanagerDefaultBaseImage, "alertmanager-default-base-image", "quay.io/prometheus/alertmanager", "Alertmanager default base image")
 	flagset.StringVar(&cfg.PrometheusDefaultBaseImage, "prometheus-default-base-image", "quay.io/prometheus/prometheus", "Prometheus default base image")
 	flagset.StringVar(&cfg.ThanosDefaultBaseImage, "thanos-default-base-image", "improbable/thanos", "Thanos default base image")
-	flagset.StringVar(&cfg.Namespace, "namespace", v1.NamespaceAll, "Namespace to scope the interaction of the Prometheus Operator and the apiserver.")
+	flagset.Var(ns, "namespaces", "Namespaces to scope the interaction of the Prometheus Operator and the apiserver.")
 	flagset.Var(&cfg.Labels, "labels", "Labels to be add to all resources created by the operator")
 	flagset.StringVar(&cfg.CrdGroup, "crd-apigroup", monitoringv1.Group, "prometheus CRD  API group name")
 	flagset.Var(&cfg.CrdKinds, "crd-kinds", " - EXPERIMENTAL (could be removed in future releases) - customize CRD kind names")
@@ -99,7 +133,7 @@ func init() {
 	flagset.StringVar(&cfg.LogFormat, "log-format", logFormatLogfmt, fmt.Sprintf("Log format to use. Possible values: %s", strings.Join(availableLogFormats, ", ")))
 	flagset.BoolVar(&cfg.ManageCRDs, "manage-crds", true, "Manage all CRDs with the Prometheus Operator.")
 	flagset.Parse(os.Args[1:])
-
+	cfg.Namespaces = ns.asSlice()
 }
 
 func Main() int {
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -140,16 +140,6 @@ func New(c prometheusoperator.Config, logger log.Logger) (*Operator, error) {
 		}),
 		&appsv1.StatefulSet{}, resyncPeriod, cache.Indexers{},
 	)
-	o.alrtInf.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    o.handleAlertmanagerAdd,
-		DeleteFunc: o.handleAlertmanagerDelete,
-		UpdateFunc: o.handleAlertmanagerUpdate,
-	})
-	o.ssetInf.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    o.handleStatefulSetAdd,
-		DeleteFunc: o.handleStatefulSetDelete,
-		UpdateFunc: o.handleStatefulSetUpdate,
-	})
 
 	return o, nil
 }
@@ -164,6 +154,45 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
 		c.statefulsetErrors,
 		NewAlertmanagerCollector(c.alrtInf.GetStore()),
 	)
+}
+
+// waitForCacheSync waits for the informers' caches to be synced.
+func (c *Operator) waitForCacheSync(stopc <-chan struct{}) error {
+	ok := true
+	informers := []struct {
+		name     string
+		informer cache.SharedIndexInformer
+	}{
+		{"Alertmanager", c.alrtInf},
+		{"StatefulSet", c.ssetInf},
+	}
+	for _, inf := range informers {
+		if !cache.WaitForCacheSync(stopc, inf.informer.HasSynced) {
+			level.Error(c.logger).Log("msg", fmt.Sprintf("failed to sync %s cache", inf.name))
+			ok = false
+		} else {
+			level.Debug(c.logger).Log("msg", fmt.Sprintf("successfully synced %s cache", inf.name))
+		}
+	}
+	if !ok {
+		return errors.New("failed to sync caches")
+	}
+	level.Info(c.logger).Log("msg", "successfully synced all caches")
+	return nil
+}
+
+// addHandlers adds the eventhandlers to the informers.
+func (c *Operator) addHandlers() {
+	c.alrtInf.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.handleAlertmanagerAdd,
+		DeleteFunc: c.handleAlertmanagerDelete,
+		UpdateFunc: c.handleAlertmanagerUpdate,
+	})
+	c.ssetInf.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.handleStatefulSetAdd,
+		DeleteFunc: c.handleStatefulSetDelete,
+		UpdateFunc: c.handleStatefulSetUpdate,
+	})
 }
 
 // Run the controller.
@@ -202,6 +231,10 @@ func (c *Operator) Run(stopc <-chan struct{}) error {
 
 	go c.alrtInf.Run(stopc)
 	go c.ssetInf.Run(stopc)
+	if err := c.waitForCacheSync(stopc); err != nil {
+		return err
+	}
+	c.addHandlers()
 
 	<-stopc
 	return nil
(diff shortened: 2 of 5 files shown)


pkg/listwatch/listwatch.go
```diff
@@ -0,0 +1,194 @@
...
+		metaObj, err := meta.ListAccessor(list)
+		if err != nil {
+			return nil, err
+		}
+		for _, item := range items {
+			l.Items = append(l.Items, runtime.RawExtension{Object: item.DeepCopyObject()})
+		}
+		resourceVersions = append(resourceVersions, metaObj.GetResourceVersion())
+	}
+	l.ListMeta.ResourceVersion = strings.Join(resourceVersions, "/")
```
brancz: let's put a comment on here why we're doing this

pkg/alertmanager/operator.go
```diff
@@ -198,6 +231,10 @@ func (c *Operator) Run(stopc <-chan struct{}) error {
 
 	go c.alrtInf.Run(stopc)
 	go c.ssetInf.Run(stopc)
+	if err := c.waitForCacheSync(stopc); err != nil {
+		return err
+	}
+	c.addHandlers()
```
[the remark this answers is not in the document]
squat: yes, when a new handler is added to an informer, the handler is called with the contents of the informer's store:
https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L338-L342

pkg/listwatch/listwatch.go
```diff
@@ -0,0 +1,180 @@
...
+				Into(result)
+			if err != nil {
+				return nil, err
+			}
+			list.Items = append(list.Items, *result)
+		}
+		return list, nil
+	}
+	watchFunc := func(_ metav1.ListOptions) (watch.Interface, error) {
+		return watch.NewFake(), nil
```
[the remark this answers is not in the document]
squat: :+1:

pkg/listwatch/listwatch.go
squat: :+1:

pkg/alertmanager/operator.go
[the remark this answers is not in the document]
brancz: perfect thanks for clearing that up :+1:

pkg/listwatch/listwatch.go
```diff
@@ -0,0 +1,204 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/cache"
+)
+
+// NewUnprivilegedNamespaceListWatchFromClient mimics cache.NewListWatchFromClient.
```
mxinden: Tiny nit: I would prefer comments to be wrapped at 80 chars.

Side note: Thanks for the detailed comments! Very much appreciate it.

pkg/listwatch/listwatch.go
```diff
@@ -0,0 +1,204 @@
...
+		options.FieldSelector = fieldSelector.String()
+	}
+	return NewFilteredUnprivilegedNamespaceListWatchFromClient(c, namespaces, optionsModifier)
+}
+
+// NewFilteredUnprivilegedNamespaceListWatchFromClient mimics cache.NewUnprivilegedNamespaceListWatchFromClient.
+// It allows for the creation of a cache.ListWatch for namespaces from a client that does not have `List` privileges.
+// If the slice of namespaces contains only v1.NamespaceAll, then this func assumes that the client
+// has List and Watch privileges and returns a regular cache.ListWatch, since there is no other way to get all namespaces.
+func NewFilteredUnprivilegedNamespaceListWatchFromClient(c cache.Getter, namespaces []string, optionsModifier func(options *metav1.ListOptions)) *cache.ListWatch {
```
mxinden: Does this function need to be exported?

squat: No, strictly speaking it does not. I just wanted to keep this as consistent as possible with the mentioned functions in https://github.com/kubernetes/client-go/blob/master/tools/cache/listwatch.go. There are two exported funcs, `New*` and `NewFiltered*` and the former wraps the latter. If you feel strongly about it then I am happy to change it

pkg/listwatch/listwatch.go
squat: wrapped all comments in the files I touched :+1:

squat:
@mxinden sure, let's add some e2e tests tomorrow :+1:

cmd/operator/main.go
```diff
@@ -54,6 +54,33 @@ const (
 	logFormatJson   = "json"
 )
 
+type namespaces map[string]struct{}
+
+// Set implements the flagset.Value interface.
```
s-urbaniak: very small nit, instead of putting this into godoc, you could
```
var _ flag.Value = (namespaces)(nil)
```
to let the compiler check that `namespaces` implements `flag.Value`

cmd/operator/main.go
```diff
@@ -54,6 +54,33 @@ const (
...
+func (n namespaces) Set(value string) error {
+	ns := strings.Split(value, ",")
+	for i := range ns {
+		n[ns[i]] = struct{}{}
```
s-urbaniak: nit: this will panic if `n` is uninitialized.

cmd/operator/main.go
```diff
@@ -88,7 +115,8 @@ func init() {
 	flagset.StringVar(&cfg.AlertmanagerDefaultBaseImage, "alertmanager-default-base-image", "quay.io/prometheus/alertmanager", "Alertmanager default base image")
 	flagset.StringVar(&cfg.PrometheusDefaultBaseImage, "prometheus-default-base-image", "quay.io/prometheus/prometheus", "Prometheus default base image")
 	flagset.StringVar(&cfg.ThanosDefaultBaseImage, "thanos-default-base-image", "improbable/thanos", "Thanos default base image")
-	flagset.StringVar(&cfg.Namespace, "namespace", v1.NamespaceAll, "Namespace to scope the interaction of the Prometheus Operator and the apiserver.")
+	ns := namespaces{}
```
s-urbaniak: ah, that prevents panicing ;-)  Wouldn't `var ns namespaces` look cleaner? That would require lazy initialization in `Set` as described above though.

pkg/alertmanager/operator.go
```diff
@@ -162,6 +156,45 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
 	)
 }
 
+// waitForCacheSync waits for the informers' caches to be synced.
+func (c *Operator) waitForCacheSync(stopc <-chan struct{}) error {
+	ok := true
+	informers := map[cache.SharedInformer]bool{
```
s-urbaniak: `cache.SharedInformer` is an interface. We are not in control over the implementing underlying type. If the underlying type uses an incomparable type (i.e. a composited map or slice) this will result in a runtime panic.

I am new to informers, but can we construct a natural key (struct) with values we have under our control?

pkg/listwatch/listwatch.go
```diff
@@ -0,0 +1,214 @@
...
+	return newMultiWatch(mlw, resourceVersions, options)
+}
+
+// multiWatch abstracts multiple watch.Interface's, allowing them
+// to be treated as a single watch.Interface.
+type multiWatch struct {
+	result   chan watch.Event
+	stopped  bool
+	stoppers []func()
+	sync.Mutex
```
s-urbaniak: this embeds `sync.Mutex` into `multiWatch`, effectively exporting `Lock()` und `Unlock()` to the outside world of but you are using those methods only internally (which is good).

Can we make a private `mu sync.Mutex // protects the stopped field only` so those methods don't escape and consumers of `multiWatch` don't even try to lock/unlock externally?

pkg/prometheus/operator.go
```diff
@@ -300,6 +296,70 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
 	)
 }
 
+// waitForCacheSync waits for the informers' caches to be synced.
+func (c *Operator) waitForCacheSync(stopc <-chan struct{}) error {
+	ok := true
+	informers := map[cache.SharedInformer]bool{
```
s-urbaniak: see comment above for using `cache.SharedInformer` as an interface.

cmd/operator/main.go
squat: What is the nit here? A map will also panic if you attempt to set a value when it is uninitialized. The solution is to `make(map...)` or `m := map...{}`. I think the same applies here, no?

pkg/alertmanager/operator.go
squat: OK sounds good

pkg/listwatch/listwatch.go
squat: `multiWatch` is an unexported type so only this package can access `Lock` and `Unlock`. Furthermore, anything in this package has access to unexported struct fields, so there is no added security, no?

Embedding the struct this way was mainly for convenience so if you feel strongly, I can make the mutex private.

cmd/operator/main.go
s-urbaniak: sounds good, agreed :+1:

cmd/operator/main.go
s-urbaniak: Sorry for being imprecise, the nit is that the following will panic:
```
var ns namespaces
ns.Set("foo")
```
Your suggestion is spot on, we may lazy initialize the map here in `Set` to make it safer.

pkg/alertmanager/operator.go
brancz: I think we can actually describe each informer with a string, so this shouldn't be much of an issue

pkg/listwatch/listwatch.go
brancz: Agreed with @s-urbaniak

cmd/operator/main.go
squat: :+1: plz see the updated code

pkg/listwatch/listwatch.go
squat: :+1:

pkg/alertmanager/operator.go
```diff
@@ -162,6 +156,49 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
...
 }
 
+// waitForCacheSync waits for the informers' caches to be synced.
+func (c *Operator) waitForCacheSync(stopc <-chan struct{}) error {
+	ok := true
+	type informerSynced struct {
+		name string
+		sync cache.InformerSynced
+	}
+	informers := map[*informerSynced]bool{
```
squat: The map cannot use an anonymous struct (or even a named struct type) as its key because the `sync` field is not comparable. Even if the field was `informer cache.SharedInformer` then we'd be back at the original issue Sergiusz raised where we'd only know if the field was comparable at runtime. Instead, we can use a pointer, whose address is comparable.

pkg/prometheus/operator.go
```diff
@@ -300,6 +296,74 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
...
 }
 
+// waitForCacheSync waits for the informers' caches to be synced.
+func (c *Operator) waitForCacheSync(stopc <-chan struct{}) error {
+	ok := true
+	type informerSynced struct {
+		name string
+		sync cache.InformerSynced
+	}
+	informers := map[*informerSynced]bool{
```
squat: same as above :)

squat:
thanks for taking a look @s-urbaniak. Plz take another look when you can.

pkg/alertmanager/operator.go
s-urbaniak: Suggestion: let's omit the map and go with a simpler slice:
```
	type namedInformer struct {
		name     string
		informer cache.SharedIndexInformer
	}

	informers = append(informers, namedInformer{"Alertmanager", c.alrtInf})
	informers = append(informers, namedInformer{"StatefulSet", c.alrtInf})

	for _, inf := range informers {
		if !cache.WaitForCacheSync(stopc, inf.informer.HasSynced) {
			level.Error(c.logger).Log("msg", fmt.Sprintf("failed to sync %s cache", inf.name))
			ok = false
		} else {
			level.Debug(c.logger).Log("msg", fmt.Sprintf("successfully synced %s cache", inf.name))
		}
	}
```

squat:
@s-urbaniak thanks for the thorough review!

brancz:
lgtm :+1: nice work!

PR #1814, opened 2018-08-21
pkg/prometheus: fix marshalling of additionalAlertManagerConfig when building the prometheus config

crandles:
While trying to use the `additionalAlertManagers` field, I noticed this error from the operator:
```
E0821 18:48:21.780219       1 operator.go:798] Sync "my-namespace/my-prometheus" failed: creating config failed: generating config failed: unmarshalling additional alert manager configs failed: yaml: unmarshal errors:
  line 1: cannot unmarshal !!map into []yaml.MapSlice
```
Added a test and fixed the marshalling bug.


crandles:
Turns out my additionalAlertManagers config was broken.

PR #1816, opened 2018-08-22
Don't abort kubelet endpoints update because of nodes without addresses

### LINKED ISSUE #1815: Syncing kubelet endpoints fails
**What did you do?**
Add new nodes to cluster, on which we had a working prometheus-operator setup.

kube-system/kubelet `Endpoints` object updated with new nodes.

`ts=2018-08-21T10:48:46.709178411Z caller=operator.go:342 component=prometheusoperator msg="syncing nodes into Endpoints object failed" err="synchronizing kubelet endpoints object failed: updating kubelet endpoints object failed: Endpoints \"kubelet\" is invalid: subsets[0].addresses[10].ip: Invalid value: \"k8s-master-24872480-0\": must be a valid IP address, (e.g. 10.9.8.7)"`

Note that all other Prometheus config updates work fine.

**Environment**
Azure, acs-engine

0.17.0 (Yes, it is old. We have some migration pains to handle before upgrading to the latest goodness)

Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-07T23:08:19Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}

* Kubernetes cluster kind: 
acs-engine

* Prometheus Operator Logs:
See above. Repeats every three minutes.

carlpett:
If a node doesn't have an IP address, currently all endpoint updating is
aborted. This change lets the addressable nodes get updated, and increments an
error counter for alerting on this condition.

Fixes #1815

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -75,7 +75,8 @@ type Operator struct {
 	// times Prometheus Operator was triggered to reconcile its created objects. It
 	// is split in the dimensions of Kubernetes objects and corresponding actions
 	// (add, delete, update).
-	triggerByCounterVec *prometheus.CounterVec
+	triggerByCounterVec     *prometheus.CounterVec
+	nodeAddressLookupErrors prometheus.Counter
 
 	host                   string
 	kubeletObjectName      string
@@ -292,10 +293,15 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
 	},
 		[]string{"triggered_by", "action"},
 	)
+	c.nodeAddressLookupErrors = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_node_address_lookup_errors_total",
+		Help: "Number of times a node IP address could not be determined",
+	})
 
 	r.MustRegister(
 		c.statefulsetErrors,
 		c.triggerByCounterVec,
+		c.nodeAddressLookupErrors,
 		NewPrometheusCollector(c.promInf.GetStore()),
 	)
 }
@@ -417,8 +423,6 @@ func (c *Operator) reconcileNodeEndpoints(stopc <-chan struct{}) {
 // nodeAddresses returns the provided node's address, based on the priority:
 // 1. NodeInternalIP
 // 2. NodeExternalIP
-// 3. NodeLegacyHostIP
-// 3. NodeHostName
 //
 // Copied from github.com/prometheus/prometheus/discovery/kubernetes/node.go
 func nodeAddress(node v1.Node) (string, map[v1.NodeAddressType][]string, error) {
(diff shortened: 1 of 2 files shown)


carlpett:
Ping @brancz @mxinden

pkg/prometheus/operator.go
```diff
@@ -472,21 +497,14 @@ func (c *Operator) syncNodeEndpoints() error {
 		return errors.Wrap(err, "listing nodes failed")
 	}
 
-	for _, n := range nodes.Items {
-		address, _, err := nodeAddress(n)
-		if err != nil {
-			return errors.Wrapf(err, "failed to determine hostname for node (%s)", n.Name)
+	addresses, errs := getNodeEndpoints(nodes)
```
squat: small nit: it's a little funny that the return value of `getNodeEndpoints` is named `addresses`. I think either the func or the value should be renamed so they agree

carlpett: Good point, I'll fix

mxinden:
Thanks for the patch @carlpett and the improved instrumentation.

As this is a bug-fix, would you mind pointing your PR against [`release-0.23`](https://github.com/coreos/prometheus-operator/tree/release-0.23)? We will get it into `master` once we cut the Prometheus Operator `v0.23.2` release.

carlpett:
@mxinden Hm, I've actually never had to change the merge target in Github before. Is it even possible? Pushed the new history now, but it still wants to merge into master. Do I need to create a new PR?

carlpett:
Found it :)

brancz:
We should probably rebase the changes on top of head of `release-0.23` additionally, that will also re-trigger CI.

Thanks for all your work @carlpett, it's very highly appreciated!

carlpett:
Hm, did I manage to rebase it wrong? The commit graph looks right here at least

carlpett:
>Have you tried turning it off and on again?

:slightly_smiling_face:

pkg/prometheus/operator.go
```diff
@@ -472,21 +497,14 @@ func (c *Operator) syncNodeEndpoints() error {
...
 
-	for _, n := range nodes.Items {
-		address, _, err := nodeAddress(n)
-		if err != nil {
-			return errors.Wrapf(err, "failed to determine hostname for node (%s)", n.Name)
+	addresses, errs := getNodeAddresses(nodes)
+	if len(errs) > 0 {
+		for _, err := range errs {
+			level.Warn(c.logger).Log("err", err)
+			c.nodeAddressLookupErrors.Inc()
```
brancz: this should be `c.nodeAddressLookupErrors.Add(float64(len(errs)))`

carlpett: Interesting, I started with that, then got a comment from a colleague that it was hard to read. Figured that for something run every 3 mins or so it wouldn't matter performance-wise.

Changed it back, but would be interesting to know your perspective on this over the other? Performance, readability, something else?

brancz:
Perfect! Thanks a ton! We should get a patch release out tomorrow. :tada:

PR #1820, opened 2018-08-23
kube-prometheus: fix root typo in node-exporter

s-urbaniak:
Currently, we are mounting `/root` from the host as `/host/root`.
Instead simply `/` from the host should be mounted.

This fixes it


brancz:
lgtm on green

PR #1829, opened 2018-08-27
pkg/prometheus: test promcfg's generation of the additionalAlertmanagerConfigs

crandles:
I wrote a test to validate the additionalAlertmanagerConfigs. I thought there was a bug originally (there isn't), but this test seems useful to have.


brancz:
Agreed! Thanks for adding this test! lgtm :+1:

Issue #1835, opened 2018-08-29
Thanos sidecar is not starting when nodes have multiple network interfaces or non standard private network ip


LucasBoisserie:
**What did you do?**

Start thanos sidecar from prometheus operator

**What did you expect to see?**

Thanos sidecar starting

**What did you see instead? Under which circumstances?**

Thanos sidecar CrashLoopBackOff

**Environment**

* Prometheus Operator version:
```
quay.io/coreos/prometheus-operator:0.23.2
```

* Kubernetes version information:

```
$ kubectl version                                                                                                         
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-08T16:31:10Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.2", GitCommit:"81753b10df112992bf51bbc2c2f85208aad78335", GitTreeState:"clean", BuildDate:"2018-04-27T09:10:24Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: 

custom

* Manifests:

```
docker images :
thanos:v0.1.0-rc.2
configmap-reload:v0.0.1
prometheus-config-reloader:v0.23.2
prometheus:v2.3.1
```

* Thanos-sidecar Logs:

```
level=error ts=2018-08-29T09:13:36.163715175Z caller=main.go:160 msg="running command failed" err="join cluster: create memberlist: Failed to get final advertise address: No private IP address found, and explicit IP not provided"
```

At Thanos sidecar startup, it starts a gossip cluster and need to know pod's ip. 
To find pod's ip, it will take the first network interface which is of type IP, forwardable, up and respect RFC 6890(private network).
In some kubernetes deployment, nodes can have multiple networks interfaces (public, admin, storage, ...) and pod network IP range can be non standard which makes crashed the thanos sidecar on start in both cases.

Thanos has an option to set directly the pod's ip for gossip cluster and grpc.
```
--cluster.address=$(POD_IP)
--grpc-address=$(POD_IP)
```
[...]

PR #1836, opened 2018-08-29
Set thanos cluster and grpc ip from pod.ip

LucasBoisserie:
This changes set cluster.address and grpc-address from pod.ip.
It's disabled automatic pod ip detection which is not working, on node with multi network interfaces or non standart private network

issue details: #1835 

Signed-off-by: Calu <lucas.boisserie@gmail.com>


brancz:
lgtm :+1: thanks a lot for the contribution!

PR #1837, opened 2018-08-30
kube-prometheus: PVC (using persistent storage for the prom tsdb) example/documentation

jolson490:
Resolves #1819

PR #1846, opened 2018-09-03
Makefile: Always generate api.md on travis

metalmatze:
In PR #1837 only the generated api.md has been changed. Everyone generating now will see a diff with the lines removed. 
To make sure we _always_ generate the api.md and similar docs on travis, we introduce the `-B` flag to the make target which will always run the task.
`$(MFLAGS)` passes the flags from make to its sub-make command.

https://stackoverflow.com/questions/816370/how-do-you-force-a-makefile-to-rebuild-a-target/816400#816400

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ generate-in-docker: hack/jsonnet-docker-image
 	--rm \
 	-u=$(shell id -u $(USER)):$(shell id -g $(USER)) \
 	-v `pwd`:/go/src/github.com/coreos/prometheus-operator \
-	po-jsonnet make generate
+	po-jsonnet make $(MFLAGS) generate
 
 .PHONY: kube-prometheus
 kube-prometheus:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ jobs:
   include:
   - stage: Sanity check and tests
   # Check generated contents are up to date and code is formatted.
-    script: make format generate-in-docker && git diff --exit-code
+    script: make -B format generate-in-docker && git diff --exit-code
   - script: cd contrib/kube-prometheus && make test-in-docker
   # Build Prometheus Operator rule config map to rule file crds cli tool
   - script: cd cmd/po-rule-migration && go install


Makefile
```diff
@@ -91,7 +91,7 @@ generate-in-docker: hack/jsonnet-docker-image
 	--rm \
 	-u=$(shell id -u $(USER)):$(shell id -g $(USER)) \
 	-v `pwd`:/go/src/github.com/coreos/prometheus-operator \
-	po-jsonnet make generate
+	po-jsonnet make $(MFLAGS) generate
```
mxinden: Maybe worth adding a small comment on what `$(MFLAGS)` does in this particular case.

brancz:
lgtm

metalmatze:
Merging this will break `master` due to the fact that the api.md is currently broken and this change detects it.

brancz:
> Merging this will break master due to the fact that the api.md is currently broken and this change detects it.

Wouldn't rebasing and re-generating fix that?

metalmatze:
Yes, I am going to update the changes to types.go so that the generated api.md will stay the same.

pkg/client/monitoring/v1/openapi_generated.go
```diff
@@ -1217,7 +1217,7 @@ func schema_pkg_client_monitoring_v1_PrometheusSpec(ref common.ReferenceCallback
 					},
 					"retention": {
 						SchemaProps: spec.SchemaProps{
-							Description: "Time duration Prometheus shall retain data for. Default is '24h'.",
+							Description: "Time duration Prometheus shall retain data for. Default is '24h', and must match the regular expression `[0-9]+(mssmhdwy)` (milliseconds seconds minutes hours days weeks years).",
```
mxinden: Sorry for holding back your PR with this unrelated issue, but shouldn't the RegEx be `[0-9]+(ms|s|m|h|d|w|y)`?

metalmatze: Yup, good catch. `mssmhdwy` does not make sense in the regex context. I simply copied the sentences and expected them to be good after getting through review already 😇

PR #1847, opened 2018-09-04
pkg/client: Add sha field to PrometheusSpec

### LINKED ISSUE #1703: Ability to specify sha of container image
As image tags are floating values, it would be nice if we could allow specifying the exact sha of an image to use, as that way we can ensure which exact image is used.

RE: #1644

cc @allisonking

metalmatze:
> As image tags are floating values, it would be nice if we could allow specifying the exact sha of an image to use, as that way we can ensure which exact image is used.

This will add the sha field to the PrometheusSpec allowing users to specify the exact sha of a container image to be deployed.

Closes #1703


Documentation/api.md
```diff
@@ -259,6 +259,7 @@ PrometheusSpec is a specification of the desired behavior of the Prometheus clus
 | serviceMonitorNamespaceSelector | Namespaces to be selected for ServiceMonitor discovery. If nil, only check own namespace. | *[metav1.LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#labelselector-v1-meta) | false |
 | version | Version of Prometheus to be deployed. | string | false |
 | tag | Tag of Prometheus container image to be deployed. Defaults to the value of `version`. | string | false |
+| sha | Sha of Prometheus container image to be deployed. Defaults to the value of `version`. Similar to a tag, but the sha explicitly deploys an immutable container image. | string | false |
```
mxinden: Maybe worth mentioning that the chain of precedence here is sha, then tag and finally version, right?

metalmatze: Yup. I've create a comment in the source, but this should be in the docs for sure 👍

squat:
looks like one of the e2e tests flaked. I am re-running the tests now

mxinden:
> Would you mind adding the same to the Thanos and Alertmanager spec? I think these should be consistent.

Would you mind doing a follow up @metalmatze?

PR #1855, opened 2018-09-05
*: Update to Go 1.11

metalmatze:
I've searched for all places where a Go version was specified. This should be all of them.


metalmatze:
Seems like some parts of Go regarding generating code have changed. I'll update this PR so it contains the new style.

metalmatze:
Apparently Kubernetes' code generation and Go 1.11 fmt do some things differently. 
https://travis-ci.org/coreos/prometheus-operator/jobs/424756201#L2075
Either the Code generation needs to be updated to match Go 1.11 or we ignore the file when formatting on our side. Maybe we wait with this PR? WDYT?

metalmatze:
Finally all formatting issues are fixed. 
After generating code we now always run `go fmt` with our current Go version on these files and thus are not depending on Kubernetes' Go version.

Issue #1463, opened 2018-06-11
include __meta_kubernetes_pod_node_name in the relabel


huydinhle:
I notice that we have relabeled  `__meta_kubernetes_namespace`, `__meta_kubernetes_pod_name`, `__meta_kubernetes_service_name` in our metrics. 

Is it possible to include the `__meta_kubernetes_pod_node_name` in there as well ? We want this information for our graphs. 

I found that the cleanest way to do this is to add `__meta_kubernetes_pod_node_name` in prometheus.cfg
```
  // Relabel namespace and pod and service labels into proper labels.
  relabelings = append(relabelings, []yaml.MapSlice{
    yaml.MapSlice{
      {Key: "source_labels", Value: []string{"__meta_kubernetes_namespace"}},
      {Key: "target_label", Value: "namespace"},
    },
    yaml.MapSlice{
      {Key: "source_labels", Value: []string{"__meta_kubernetes_pod_name"}},
      {Key: "target_label", Value: "pod"},
    },
    yaml.MapSlice{
      {Key: "source_labels", Value: []string{"__meta_kubernetes_service_name"}},
      {Key: "target_label", Value: "service"},
    },
    yaml.MapSlice{
      {Key: "source_labels", Value: []string{"__meta_kubernetes_pod_node_name"}},
      {Key: "target_label", Value: "node"},
    },
  }...)
```

Can you let me know maybe there is an better way to add relabel_config in our Prometheus

huydinhle:
closing this issue because of this one https://github.com/coreos/prometheus-operator/issues/1166

PR #1856, opened 2018-09-05
pkg/prometheus: add node_name to labels

vsliouniaev:
As per @huydinhle's suggestion in #1463 :

When doing Prometheus debugging, correlating the pod with the host node is useful for disk metrics. To do this it is necessary to get information about the node name.


brancz:
The `node` is not an identifying label for a target, and therefore doesn't belong on the set of labels identifying the target. You can get this information at query time, via the `kube_pod_info` metric, which you can join onto other metrics if you need this information, or when debugging simply querying that one metric.

vsliouniaev:
Thanks for that!

PR #1861, opened 2018-09-06
test/e2e: Add multi namespace end-to-end test

mxinden:
Refactor test framework to support multi namespace end-to-end tests. Add
basic multi namespace end-to-end test:

- Creating two namespaces with one firing Prometheus alerting rule each.
- Create Prometheus which selects Prometheus rules out of both namespaces.
- Configure Prometheus Operator to only watch one namespace.

=> Expect Prometheus to only fire one alert.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -21,17 +21,27 @@ import (
 	"testing"
 
 	operatorFramework "github.com/coreos/prometheus-operator/test/framework"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 )
 
-var framework *operatorFramework.Framework
-
-// Basic set of e2e tests for the operator:
-// - config reload (with and without external url)
+var (
+	framework *operatorFramework.Framework
+	opImage   *string
+)
 
 func TestMain(m *testing.M) {
-	kubeconfig := flag.String("kubeconfig", "", "kube config path, e.g. $HOME/.kube/config")
-	opImage := flag.String("operator-image", "", "operator image, e.g. quay.io/coreos/prometheus-operator")
-	ns := flag.String("namespace", "prometheus-operator-e2e-tests", "e2e test namespace")
+	kubeconfig := flag.String(
+		"kubeconfig",
+		"",
+		"kube config path, e.g. $HOME/.kube/config",
+	)
+	opImage = flag.String(
+		"operator-image",
+		"",
+		"operator image, e.g. quay.io/coreos/prometheus-operator",
+	)
 	flag.Parse()
 
 	var (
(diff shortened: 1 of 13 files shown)


brancz:
Generally this looks great! There seem to be some failures though and we were discussing if we could maybe find a way to make sure that the map of subtests is consistent with the ones available. I don't mind if we skip that in the scope of this PR (actually I would prefer if it doesn't grow any bigger), but let's make sure to create an issue to track that we do this.

test/e2e/main_test.go
```diff
@@ -21,66 +21,142 @@ import (
...
 
-	exitCode = m.Run()
+	ns := ctx.CreateNamespace(t, framework.KubeClient)
+
+	err := framework.CreatePrometheusOperator(ns, *opImage, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// t.Run blocks until f returns or calls t.Parallel to become a parallel
```
metalmatze: > cks until f return
What do you mean with f? Framework?

mxinden: `f` is the function passed as the second argument to `t.Run`. Good question. I have adjusted the description.

mxinden:
@brancz ensuring all end-to-end tests are run is tracked here: https://github.com/coreos/prometheus-operator/issues/1874

@brancz @metalmatze can I get another review / approval?

squat:
I'm sorry this comes so late but I was just thinking about this some more I think it would be nice to do the following:
* Creating __three__ namespaces with one firing Prometheus alerting rule each.
* Create Prometheus which selects Prometheus rules out of __all__ namespaces.
* Configure Prometheus Operator to only watch __two__ namespaces.
=> Expect Prometheus to only fire __two__ alerts.

The reason for this is that the multi-namespace ListWatchs optimizes for the single namespace case [0] and treats it differently from watching multiple namespaces. It would be nice to have an _explicitly_ multi-namespace PO e2e test to validate this entire codepath.

[0] https://github.com/coreos/prometheus-operator/blob/master/pkg/listwatch/listwatch.go#L91-L95

squat:
We can also make this change in a follow up of course

mxinden:
@squat excellent hint. Forgot about that. I will add that in this PR.

test/e2e/prometheus_test.go
```diff
@@ -1339,6 +1339,155 @@ func TestPrometheusGetBasicAuthSecret(t *testing.T) {
...
+// It checks whether it ignores rules that are not in the watched namespaces of the
+// Prometheus Operator. The Prometheus Operator internally treats watching a
+// single namespace different than watching multiple namespaces, hence the two
+// sub-tests.
+func testOperatorNSScope(t *testing.T) {
+	name := "test"
+	firtAlertName := "firstAlert"
+	secondAlertName := "secondAlert"
+
+	t.Run("SingleNS", func(t *testing.T) {
```
mxinden: I have split this test up into two subtests. One with the Prometheus Operator watching a single namespace, one watching two namespaces.

mxinden:
@squat would you mind taking another look?

mxinden:
@squat ping

squat:
travis is complaining about a newline on the hack/generate/jsonnetfile.json. Otherwise looks good!

metalmatze:
Oh, yes. I recently merged a PR in the jsonnet-bundler to add a newline at the end of `jsonnetfile.json`. @mxinden can you please run `go get github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb`. Sorry and thanks!

PR #1865, opened 2018-09-06
contrib/kube-prometheus: Use default timezone in dashboards

metalmatze:
In the [kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin) we now override the dashboard's timezone to default to what is configured in Grafana itself. This updates the dependency and the generated configmap.

@brancz @squat


squat:
xref: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/78

brancz:
lgtm on green

PR #1871, opened 2018-09-07
Skip malformed rules when processing PrometheusRule CRDs

dcondomitti:
In clusters without CRD validation (pre-1.9) invalid PrometheusRules can halt processing of all rules if one is malformed. It's not ready to be merged because I'm seeing EOF errors in the watch loop after dropping a malformed rule. This could be already happening as I wasn't watching for it before but I'm assuming it's due to the `watch.Deleted`. I'm still not exactly sure how that process works. If anyone has suggestions on how to debug that I'd appreciate it.

@mxinden This is the patch we implemented for https://github.com/coreos/prometheus-operator/issues/1831.

### DIFF
--- a/pkg/client/monitoring/v1/prometheusrule.go
+++ b/pkg/client/monitoring/v1/prometheusrule.go
@@ -16,6 +16,7 @@ package v1
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/pkg/errors"
 
@@ -53,6 +54,12 @@ type prometheusrules struct {
 	client     dynamic.ResourceInterface
 	crdKind    CrdKind
 	ns         string
+}
+
+type rawPrometheusRuleList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []json.RawMessage `json:"items"`
 }
 
 func newPrometheusRules(r rest.Interface, c *dynamic.Client, crdKind CrdKind, namespace string) *prometheusrules {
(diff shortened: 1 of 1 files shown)


mxinden:
Thanks for looking into it and also proposing a patch.

I still think that this introduces atomicity concerns, as mentioned in https://github.com/coreos/prometheus-operator/issues/1831#issuecomment-417675077:

> In regards to preceding with all valid rules in case one rule is malformed in a rule group: I would like to bring up the issue of atomicity here. Say a user wants to create two new rules, one malformed recording rule and one valid alerting rule depending on the former recording rule. Skipping the recording and creating the alerting rule would result in a failing alerting rule evaluation.

In addition implementing custom marshalers only enforces, that rules adhere to the struct properties. It does not enforce each expression string to be a valid Prometheus expression. If an expression is invalid, it will pass through the Prometheus Operator but will be rejected at Prometheus along with all other rules included in the update, bringing rule updating at n the Prometheus side to a halt.

This we can fix via custom admission controllers (See https://github.com/coreos/prometheus-operator/issues/1831#issuecomment-418347686).

@dcondomitti Let us know what you think.

squat:
Until we implement the custom admission controllers, we could consider using the control loop to write a status condition to the PrometheusRules object that lists the rules that failed to parse.

metalmatze:
Thinking about this again, I'd personally much rather like the whole generation to fail and then have some alerting rules on the rules generation errors, so that one can immediately see if it fails.

michaelbannister:
Here and elsewhere there are several remarks about implementing custom admission controllers or validating webhooks. Is there a specific issue for this (so I can subscribe to see progress on it)?

domino7:
This kind of validation would be very useful, I observed this issue when rule with malformed syntax was added to the pool (earlier correctly tested with too old promtool version), as such:
- all new rules were omitted and not added to configuration
- Prometheus cannot restart / start up until malformed rules are removed

From that perspective the metric exposing configuration changes errors is must have.

In regard to the atomicity concern, when applied at namespace level it would allow to avoid cross-projects and cross-teams impacts. Very bare minimum could be to validate atomicity initially at PrometheusRule (single object) level, as optional step activated via dedicated Prometheus flag (to keep backward compatibility with current behavior if someone desire full atomicity / current behavior).

Any chance for further work on that PR @dcondomitti?

PR #1875, opened 2018-09-10
Add configuration for priority class to be assigned to pods

aveshagarwal:
@sjenning 

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1624253


brancz:
How critical is this to go through for v3.11? I'm asking as we would need to cut a new release for this and let everything trickle down into the openshift release, in which case we would have to request an extension for code freeze.

PR #1876, opened 2018-09-11
pkg/alertmanager: Add additionalPeers to alertmanager CRD for configuring extra peers

### LINKED ISSUE #1052: proposal: support configuring AlertManager mesh peers
I would like to deploy an AlertManager mesh across multiple Kubernetes clusters. To do so, this would require some extra configuration of the `-mesh.peer` flag.

AM doesn't support service discovery [yet](https://github.com/prometheus/alertmanager/issues/645) -- ideally we would add an AlertingSpec, to the AlertmanagerSpec that would discover the other endpoints, similar to Prometheus.

Until that is available, would it be possible to support a list of initial peers to the AlertmanagerSpec?

```
kind: Alertmanager
...
spec:
  initialPeers:
  - fqdn-01.somewhere.com:1234
  - alertmanager-multi-cluster-fqdn.com
```

This could be used in conjunction with the peers automatically added via the operator when multiple replicas are defined.

crandles:
Resolves #1052 

New config for supplying extra peers to AlertManager


metalmatze:
Please rebase this branch to the current `master`. Thanks!

crandles:
Will resubmit after I do a little rebasing.

crandles:
See #1878

PR #1880, opened 2018-09-11
pkg/prometheus: add podTargetLabels to CRD for sourcing additional labels from pods

crandles:
Something like this was suggested in #1604 for promoting pod labels like service labels are via `targetLabels`.


crandles:
Will do

crandles:
Added a test for PodTargetLabels as well as TargetLabels, since they are pretty similar.

crandles:
I can't tell why this test is failing. I tried to regen docs, but there were no changes.

brancz:
You need to rebase and then re-generate. Sorry recent merges caused a conflict.

brancz:
lgtm

PR #1881, opened 2018-09-11
Add square brackets around pod IPs

### LINKED ISSUE #1858: Alertmanager fails to start when POD_IP is IPv6
I attempted to deploy kube-prometheus on my ipv6 enabled kubernetes cluster.  When alertmanager starts, it crashes immediately with the dreaded "too many colons" error.   If POD_IP is ipv6 it needs to be enclosed in square brackets.  

https://github.com/coreos/prometheus-operator/blob/4253f01455fa7a578d45dcf43e251c5c229c3351/pkg/alertmanager/statefulset.go#L198

Based on a quick grep, the issue probably exists in more than this one spot.  I can't think of a quick way to fix this.  If anyone has an idea, I'm happy to work on a PR.

azenk:
IPv6 pod IPs need to be enclosed in square brackets.  This closes #1858.

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -195,7 +195,7 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 
 	amArgs := []string{
 		fmt.Sprintf("--config.file=%s", alertmanagerConfFile),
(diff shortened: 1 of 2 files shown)


brancz:
lgtm :+1:

mxinden:
Thanks @azenk for the contribution!

PR #1886, opened 2018-09-12
Fixing build.sh in kube-prometheus README

dmcnaught:
Fixing https://github.com/coreos/prometheus-operator/issues/1885

### DIFF
--- a/contrib/kube-prometheus/README.md
+++ b/contrib/kube-prometheus/README.md
@@ -140,7 +140,7 @@ rm -rf manifests
 mkdir manifests
 
                                                # optional, but we would like to generate yaml, not json
-jsonnet -J vendor -m manifests "${1-example.jsonnet}" | xargs -I{} sh -c 'cat {} | gojsontoyaml > {}.yaml; rm -f {}' -- {}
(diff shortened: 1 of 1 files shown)


brancz:
I think that’s actually correct. $1 is the first argument passed to the script and the rest means, that if it’s not specified it will default to example.jsonnet

mxinden:
@dmcnaught for example:

```shell
➜  /tmp cat test.sh 
echo "The first argument was"
echo ${1-no argument}
➜  /tmp ./test.sh      
The first argument was
no argument
➜  /tmp ./test.sh hallo
The first argument was
hallo
➜  /tmp 
```

Not sure if this differs on Mac though.

metalmatze:
I've tested this on my Mac and its behaviour exactly the same as documented by @mxinden.

dmcnaught:
Ah. I forgot it could be that. I’m using zsh and it didn’t work. I can look into it more later today.

metalmatze:
I tested that again on macOS with zsh (previously with bash).

tesh.sh:
```
#!/bin/zsh
echo "The first argument was"
echo ${1-no argument}
```

`$ zsh ./test.sh`
```
The first argument was
no argument
```

`$ zsh ./test.sh hello`
```
The first argument was
hello
```

I'd say that it works as expected on linux and macOS and with bash and zsh, thus closing.
If you happen to find that it's still an issue, please reopen.
Thanks.

PR #1891, opened 2018-09-13
kube-prometheus: Add Prometheus Operator alerts

metalmatze:
We are missing some alerts for reconciling errors that might happen when running the Prometheus Operator. This PR will add these alerts to kube-prometheus.

/cc @brancz @squat


contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/prometheus-operator.libsonnet
```diff
@@ -0,0 +1,50 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'prometheus-operator',
+        rules: [
+          {
+            alert: 'PrometheusOperatorAlertmanagerReconcileErrors',
+            expr: |||
+              rate(prometheus_operator_alertmanager_reconcile_errors_total{%(prometheusOperatorSelector)s}[5m]) > 0.01
```
brancz: I’m fairly certain 1% is going to trigger on any error occurring. Let’s bump this to 10% I’d say, or do a 3% warning and 10% critical.

metalmatze:
I've updated this to alert on 10% reconciling errors. I don't think we need to split these up in warning and critical.

contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/prometheus-operator.libsonnet
```diff
@@ -0,0 +1,50 @@
...
+          {
+            alert: 'PrometheusOperatorAlertmanagerReconcileErrors',
+            expr: |||
+              rate(prometheus_operator_alertmanager_reconcile_errors_total{%(prometheusOperatorSelector)s}[5m]) > 0.1
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'Errors while reconciling Alertmanager in {{$labels.namespace}} namespace.',
```
squat: Nit: in the kubernetes-monitoring mixins the template variables all have spaces between the brackets and the variable name (look at the generated alerting rules below. Functionally of course it it the same but it would be nice to standardize

squat: Also, the kubernetes monitoring mixins all capitalize the kubernetes object types, so “namespace” becomes “Namespace”, “daemonset” becomes “DaemonSet” etc

metalmatze:
I've addressed your points @squat. ☺️

PR #1892, opened 2018-09-13
contrib/kube-prometheus: Fix outdated alerts to use message annotation

metalmatze:
We don't use the summary and description annotations in our alerts anymore. This will remove the last alerts that are still using the old annotations.

@brancz


brancz:
Oops, nice catch, just need a re-generate.

squat:
That one CI failure is unrelated. Travis is picking up a branch named the same as this PR ref but the branch is gone. The actual CI tests passed. Merging.

Issue #689, opened 2017-10-11
Consider setting sample_limit to mitigate overload


julianvmodesto:
`sample_limit` is available in the scrape config as a means of mitigating Prometheus overload:

> Thus, the reasons for the inability to evict enough chunks include the following:
> ...
>    3. There are too many active time series, which results in too many open head chunks.
> ...
> To mitigate an unplanned explosion of the number of series, you can limit the number of samples per individual scrape (see sample_limit in the scrape config). If the number of active time series exceeds the number of memory chunks the Prometheus server can afford, the server will quickly throttle ingestion as described above. The only way out of this is to give Prometheus more RAM or reduce the number of time series to ingest.

https://prometheus.io/docs/operating/storage/

This is limit that would be high enough to only hit during an emergency. See https://github.com/prometheus/prometheus/issues/2137 for reference.

Perhaps a reasonable value for this is `sample_limit: 1000` for scrape configs.

brancz:
I think this is a valid setting on a `ServiceMonitor` endpoint. kube-state-metrics easily exports thousands of metrics in a single scrape, so I'm not sure whether we should be posing any default on a user, but I'm happy to have it configurable

midnightconman:
I haven't been able to find this in the `PrometheusSpec` or `ServiceMonitorSpec` objects. Could we add this to the `PrometheusSpec` object, so people could impose scrape limits at the prometheus cluster level?

brancz:
`sample_limit` is configured on a per job basis, so this should be configured in the `ServiceMonitorSpec` or even the ServiceMonitor's Endpoints. Ultimately this should be configured with defaults through Kubernetes LimitRanges in combination with ResourceQuotas (though I don't know whether either of those are extensible today), so I'm not entirely sure whether we should have the field in the `PrometheusSpec`. It definitely should be available in the ServiceMonitor.

brancz:
It is now possible to configure an enforced sample limit on Prometheus side, as well as a sample limit on ServiceMonitors/PodMonitors. In the future we may also add exceptions to the enforced limit for example for the kube-state-metrics case mentioned above.

PR #1895, opened 2018-09-13
configure sample_limit per ServiceMonitor

mkabischev:
The problem is described in #689. We want to have possibility to limit max number of series to mitigate prometheus overload.

There is no default value like 1000 (as proposed in issue), so you need to set it explicitly.


brancz:
You know I could have sworn that we already had this feature. Thanks a lot for contributing! This lgtm :+1:

mkabischev:
@brancz thanks!
Can you suggest when will be the next release?

Issue #1548, opened 2018-06-28
Ability to add custom relabeling to prometheus configuration


d-ulyanov:
**What did you do?**
I would like to add some labels to targets which I have on endpoint but not on k8s service.
Example:
I have kubelet job (from operator examples), that scrapes kubelets. 
Targets have label (before relabeling) 
`__meta_kubernetes_endpoint_address_target_name: <myprettyhostname>`

**What did you expect to see?**
I would like to have this label in scraped metrics. It could be done via relabeling, but I can't configure custom relabeling for job.

**What did you see instead? Under which circumstances?**
It would be great to add ability to configure relabel configs.

* Kubernetes version information:
```
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.4", GitCommit:"5ca598b4ba5abb89bb773071ce452e33fb66339d", GitTreeState:"clean", BuildDate:"2018-06-06T15:22:13Z", GoVersion:"go1.9.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.4", GitCommit:"5ca598b4ba5abb89bb773071ce452e33fb66339d", GitTreeState:"clean", BuildDate:"2018-06-06T08:00:59Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

brancz:
Relabeling is a concept that is very hard to grasp for most users, if possible I'd like to avoid adding relabeling as much as possible, as it will cause a ton of inconsistencies across environments making alerting rules and dashboard sharing much much harder.

In this case I think it's actually valid to simply always relabel the endpoints' `__meta_kubernetes_endpoint_address_target_name` meta label.

d-ulyanov:
Hi @brancz !
Now I've solved an issue with additionalScrapeConfig (it exists only for forwarding target_name label)

Maybe we can just add ability to list endpoint label names that would be transferred to target? 
We can make PR for this.

d-ulyanov:
Hi, @brancz 
What do you think about this?

brancz:
I'm not entirely sure if this is possible but if so I would like to try and always relabel

* `__meta_kubernetes_endpoint_address_target_name` -> `target_name`
* `target_name` -> `pod` (if `__meta_kubernetes_endpoint_address_target_kind` == `Pod`)
[...]

PR #1896, opened 2018-09-14
Target name relabeling

mkabischev:
This PR implements solution to relabel `__meta_kubernetes_endpoint_address_target_name ` to `pod`/`node` label proposed in #1548.


brancz:
Thank you so much for this! Did you already try whether kubelet's would be tagged with the `node` label?

brancz:
Also I'm not entirely sure the Pod part is actually necessary as we already do that kind of a relabelling rule.

mkabischev:
@brancz, yeah, I've tested it :)
screen shot 2018-09-17 at 16 56 05

I removed old relabelling for pod, so now they works the same way.

brancz:
Really awesome! Thanks a lot!

> I removed old relabelling for pod, so now they works the same way.

Whoops, I missed that.

lgtm :+1: 

(still giving others a chance to review)

PR #1898, opened 2018-09-14
[WIP] contrib/kube-prometheus: Use mixtool to build manifests

metalmatze:
With the initial release of the [mixtool](https://github.com/metalmatze/mixtool) we can start replacing some of the scripts with a much simpler CLI UX.

This is just the beginning of this PR.


contrib/kube-prometheus/README.md
```diff
@@ -101,7 +101,11 @@ In order to update the kube-prometheus dependency, simply use the jsonnet-bundle
 
 ### Compiling
 
-e.g. of how to compile the manifests: `./build.sh example.jsonnet`
+Compiling is done with [mixtool](https://github.com/metalmatze/mixtool) which bundles [go-jsonnet](https://github.com/google/go-jsonnet) and adds some more convenience helpers like converting to YAML and linting the output.
+
+> `mixtool` can be installed with `go get -u github.com/metalmatze/mixtool/cmd/mixtool`
+
+Example of how to compile the manifests: `mixtool build -m manifests/ example.jsonnet`
```
brancz: <3

brancz: maybe we should default to outputting json and make yaml output an argument

brancz:
really awesome!

contrib/kube-prometheus/README.md
metalmatze: `--yaml` is already a flag. It is simply `true` by default. We can easily flip it. Though I think that most people want YAML. Maybe we should add a `--json` flag which is false by default instead. Might make more sense from a UX perspective.

brancz: I think we need to be careful not to mix kubernetes and mixtool concerns. mixtool really should just be optimized for mixins.

metalmatze: Yep, totally. But I still think that most people want to output YAML these days.
Moving the discussion to the mixtool itself: https://github.com/metalmatze/mixtool/issues/2

metalmatze: Should we merge this PR nevertheless? WDYT?

brancz: as long as mixtool is a moving target in terms of what it renders, I'd rather not as that could cause builds to be non-reproducible if mixtool master changes

contrib/kube-prometheus/Makefile
```diff
@@ -23,12 +23,12 @@ generate-in-docker: ../../hack/jsonnet-docker-image
 
 generate: manifests **.md
 
-**.md: $(EMBEDMD_BINARY) $(shell find examples) build.sh example.jsonnet
+**.md: $(EMBEDMD_BINARY) $(shell find examples) example.jsonnet
 	$(EMBEDMD_BINARY) -w `find . -name "*.md" | grep -v vendor`
 
-manifests: vendor example.jsonnet build.sh
+manifests: vendor example.jsonnet
```
mxinden: This target depends on the `mixtool` binary. Can you add a dependency like:

```make
$(JB_BINARY):
  go get -u github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
```

metalmatze:
This is blocked at the moment, until we figure out the way forward with the mixtool.

PR #1911, opened 2018-09-18
Fix the syntax of code block in table row in .md Markdown file.

jolson490:
Previously these two rows were not rendering properly on github.com.


mxinden:
`API.md` is autogenerated based on https://github.com/coreos/prometheus-operator/blob/master/pkg/client/monitoring/v1/types.go. Probably best to change there, even though that would add a bit of noise to the go code. Any alternatives?

metalmatze:
Any updates on generating this? :blush:

PR #1914, opened 2018-09-19
contrib/kube-prometheus: Update README with up-to-date default versions

dlespiau:
With those changes, diff does show any differences between generated manifests
with and without the version override in a top level jsonnet file.

### DIFF
--- a/contrib/kube-prometheus/README.md
+++ b/contrib/kube-prometheus/README.md
@@ -187,13 +187,13 @@ These are the available fields with their respective default values:
     namespace: "default",
 
     versions+:: {
-        alertmanager: "v0.15.0",
-        nodeExporter: "v0.15.2",
+        alertmanager: "v0.15.2",
+        nodeExporter: "v0.16.0",
         kubeStateMetrics: "v1.3.1",
         kubeRbacProxy: "v0.3.1",
         addonResizer: "1.0",
-        prometheusOperator: "v0.18.1",
-        prometheus: "v2.2.1",
+        prometheusOperator: "v0.23.2",
+        prometheus: "v2.3.2",
     },
 
     imageRepos+:: {


dlespiau:
For reference, the diff with the current documented default versions:

```diff
$ diff -u manifests.orig/ manifests
diff -u manifests.orig/0prometheus-operator-deployment.yaml manifests/0prometheus-operator-deployment.yaml
--- manifests.orig/0prometheus-operator-deployment.yaml	2018-09-19 11:28:45.117884294 +0100
+++ manifests/0prometheus-operator-deployment.yaml	2018-09-19 11:46:18.143833290 +0100
@@ -20,8 +20,8 @@
         - --kubelet-service=kube-system/kubelet
         - --logtostderr=true
         - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
-        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.23.2
-        image: quay.io/coreos/prometheus-operator:v0.23.2
+        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.18.1
+        image: quay.io/coreos/prometheus-operator:v0.18.1
         name: prometheus-operator
         ports:
         - containerPort: 8080
diff -u manifests.orig/alertmanager-alertmanager.yaml manifests/alertmanager-alertmanager.yaml
--- manifests.orig/alertmanager-alertmanager.yaml	2018-09-19 11:28:45.125884381 +0100
+++ manifests/alertmanager-alertmanager.yaml	2018-09-19 11:46:18.151833388 +0100
@@ -11,4 +11,4 @@
     beta.kubernetes.io/os: linux
   replicas: 3
   serviceAccountName: alertmanager-main
-  version: v0.15.2
+  version: v0.15.0
diff -u manifests.orig/node-exporter-daemonset.yaml manifests/node-exporter-daemonset.yaml
--- manifests.orig/node-exporter-daemonset.yaml	2018-09-19 11:28:45.173884900 +0100
+++ manifests/node-exporter-daemonset.yaml	2018-09-19 11:46:18.219834220 +0100
@@ -21,7 +21,7 @@
         - --path.sysfs=/host/sys
         - --collector.filesystem.ignored-mount-points=^/(dev|proc|sys|var/lib/docker/.+)($|/)
         - --collector.filesystem.ignored-fs-types=^(autofs|binfmt_misc|cgroup|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|mqueue|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|sysfs|tracefs)$
-        image: quay.io/prometheus/node-exporter:v0.16.0
+        image: quay.io/prometheus/node-exporter:v0.15.2
         name: node-exporter
         resources:
           limits:
diff -u manifests.orig/prometheus-prometheus.yaml manifests/prometheus-prometheus.yaml
--- manifests.orig/prometheus-prometheus.yaml	2018-09-19 11:28:45.185885029 +0100
+++ manifests/prometheus-prometheus.yaml	2018-09-19 11:46:18.231834367 +0100
@@ -25,4 +25,4 @@
   serviceAccountName: prometheus-k8s
   serviceMonitorNamespaceSelector: {}
   serviceMonitorSelector: {}
-  version: v2.3.2
+  version: v2.2.1
```

brancz:
Yes you’re totally right that we should automate this. I have some ideas but they are as you said a lot more involved. Merging this for now. Thanks a lot for the contribution! :)

metalmatze:
Maybe we can actually create a file that has this default configuration set explicitly and then use embedmd to keep the README.md up to date?

dlespiau:
I was thinking along the lines of:
- a bit of jsonnet that dump the _config object
- a way to template the README.md file so template + _config -> README.md
- a test that ensure the README.md has been regenerated when something in _config changes.

brancz:
Yep I was thinking the same as @dlespiau. It's not super easy, but totally possible. It's all hidden fields and objects and we need to recursively traverse the config and convert it to non-hidden fields (for jsonnet to actually render it) and then somehow (maybe regex replace is good enough) replace all `:` with `+::` for all non-value fields. Then we can use [embedmd](https://github.com/campoy/embedmd) which we already use all over the place to include it into the readme with the `generate` make target.

dlespiau:
Opened https://github.com/coreos/prometheus-operator/issues/1924 to capture the ideas there.

PR #1925, opened 2018-09-21
pkg/k8sutil: add test case for secret name sanitization

squat:
cc @mxinden @metalmatze 

addresses: https://github.com/coreos/prometheus-operator/pull/1907#pullrequestreview-157673507


metalmatze:
lgtm

mxinden:
Thanks for the follow up @squat!

PR #1929, opened 2018-09-24
*: Update to Prometheus 2.4.0

mxinden:
Supersedes https://github.com/coreos/prometheus-operator/pull/1894.

### DIFF
--- a/Documentation/compatibility.md
+++ b/Documentation/compatibility.md
@@ -35,3 +35,6 @@ The versions of Prometheus compatible to be run with the Prometheus Operator are
 * v2.2.1
 * v2.3.1
 * v2.3.2
+* v2.4.0
+* v2.4.1
+* v2.4.2
--- a/Documentation/user-guides/cluster-monitoring.md
+++ b/Documentation/user-guides/cluster-monitoring.md
@@ -422,7 +422,7 @@ spec:
   serviceAccountName: prometheus-k8s
   serviceMonitorNamespaceSelector: {}
   serviceMonitorSelector: {}
-  version: v2.3.2
+  version: v2.4.2
 ```
 
 > Make sure that the `ServiceAccount` called `prometheus-k8s` exists and if using RBAC, is bound to the correct role. Read more on [RBAC when using the Prometheus Operator](../rbac.md).
--- a/contrib/kube-prometheus/manifests/prometheus-prometheus.yaml
+++ b/contrib/kube-prometheus/manifests/prometheus-prometheus.yaml
@@ -25,4 +25,4 @@ spec:
   serviceAccountName: prometheus-k8s
   serviceMonitorNamespaceSelector: {}
   serviceMonitorSelector: {}
-  version: v2.3.2
+  version: v2.4.2
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -358,8 +358,6 @@ func testAMZeroDowntimeRollingDeployment(t *testing.T) {
 	ns := ctx.CreateNamespace(t, framework.KubeClient)
 	ctx.SetupPrometheusRBAC(t, ns, framework.KubeClient)
 
-	alertName := "ExampleAlert"
-
 	whReplicas := int32(1)
 	whdpl := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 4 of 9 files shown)


mxinden:
Updating ` testAMZeroDowntimeRollingDeployment` to use Alertmanager v0.15.2 fails. An Alertmanager joining a given cluster does not properly catch up on which alerts have already been sent. I will debug this further upstream.

Documentation/compatibility.md
```diff
@@ -35,3 +35,4 @@ The versions of Prometheus compatible to be run with the Prometheus Operator are
 * v2.2.1
 * v2.3.1
 * v2.3.2
+* v2.4.0
```
mrueg: Probably want to add 2.4.1 and 2.4.2 here as well

Documentation/user-guides/cluster-monitoring.md
```diff
@@ -422,7 +422,7 @@ spec:
   serviceAccountName: prometheus-k8s
   serviceMonitorNamespaceSelector: {}
   serviceMonitorSelector: {}
-  version: v2.3.2
+  version: v2.4.0
```
mrueg: Use 2.4.2 here

contrib/kube-prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -25,4 +25,4 @@ spec:
   serviceAccountName: prometheus-k8s
   serviceMonitorNamespaceSelector: {}
   serviceMonitorSelector: {}
-  version: v2.3.2
+  version: v2.4.0
```
mrueg: Use 2.4.2 here

Documentation/compatibility.md
deejay1: I'd stick just with 2.4.2 due to `[BUGFIX] Handle WAL corruptions properly prometheus/tsdb#389`

metalmatze: 2.4.2 is the new default Prometheus version, this is just the compatibility matrix of supported versions by the operator.

metalmatze:
I think the e2e tests are still failing. IIRC @mxinden wanted to look into these, but I might be wrong here.

mxinden:
@brancz @metalmatze Preparing a rebase and re-generate locally right now. Looking into the e2e failures afterwards.

mxinden:
I can reproduce `TestAllNS/x/PromReloadConfig` failure locally.

mxinden:
@brancz @metalmatze would you mind giving this another look?

test/framework/alertmanager.go
```diff
@@ -255,6 +256,27 @@ func (f *Framework) CreateSilence(ns, n string) (string, error) {
 	return createSilenceResponse.Data.SilenceID, nil
 }
 
+func (f *Framework) SendAlertToAlertmanager(ns, n string) error {
+	var postAlertResp amAPIPostAlertResp
+	request := ProxyPostPod(f.KubeClient, ns, n, "web", "api/v1/alerts", `
+[{"labels":{"alertname":"ExampleAlert","prometheus":"my-prometheus"},"annotations":{},"startsAt":"2018-09-17T16:27:45.70751247Z","endsAt":"2030-09-17T16:31:45.80751247Z","generatorURL":"http://prometheus-test-0:9090/graph?g0.expr=vector%281%29\u0026g0.tab=1"}]
```
brancz: let's not hardcode the dates here, even if it's in 12 years

test/e2e/alertmanager_test.go
```diff
@@ -472,32 +470,43 @@ inhibit_rules:
...
+		go func() {
+			ticker := time.NewTicker(100 * time.Millisecond)
+			failures := 0
+			for {
+				select {
+				case <-ticker.C:
+					err := framework.SendAlertToAlertmanager(ns, "alertmanager-rolling-deploy-"+strconv.Itoa(replica))
+					if err != nil {
+						failures++
+						// Allow 30 (~3 Seconds) failures during Alertmanager rolling update.
```
brancz: actually any amount of failures is fine as long as the result in the end is still correct

mxinden: The test can be split up into three phases:

1. Send alerts and expect a single notification.

2. Roll out new version of Alertmanager.

3. Send alerts and expect no new notifications.

If sending alerts between two and three fails, we will not see any new notifications, thereby the test might give a false positive.

Does that make sense @brancz?

brancz: yep, makes perfect sense

test/framework/alertmanager.go
mxinden: Agree, that approach is a bit hacky. Updated.

brancz:
lgtm :+1: nice!

PR #1934, opened 2018-09-25
update Thanos to v0.1.0

marcin-je:
This updates default version of Thanos to v0.1.0. The update is motivated by the ability to use EC2 node's IAM policy to access S3 bucket rather than passing AWS Keys into the configuration.

v0.1.0 is not backward-compatible - see https://github.com/improbable-eng/thanos/releases/tag/v0.1.0

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -35,7 +35,7 @@ import (
 const (
 	governingServiceName     = "prometheus-operated"
 	DefaultPrometheusVersion = "v2.3.2"
-	DefaultThanosVersion     = "v0.1.0-rc.2"
+	DefaultThanosVersion     = "v0.1.0"
 	defaultRetention         = "24h"
 	storageDir               = "/prometheus"
 	confDir                  = "/etc/prometheus/config"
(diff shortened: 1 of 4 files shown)


metalmatze:
I'll test this tomorrow. 
If somebody wants to jump on it in the meantime, I don't have a problem with it. 😊

pkg/prometheus/statefulset.go
```diff
@@ -599,7 +599,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		thanosArgs := []string{
 			"sidecar",
 			fmt.Sprintf("--prometheus.url=http://%s:9090", c.LocalHost),
-			fmt.Sprintf("--tsdb.path=%s", storageDir),
+			fmt.Sprintf("--data-dir=%s", storageDir),
```
metalmatze: This change is wrong. The flag only changed for the store:
> [breaking flag compatibility] Changed store gateway tsdb.path flag to data-dir

I've downloaded the 0.1.0 binary and run `thanos sidecar -h` where I can see `tsdb.path` as flag.

metalmatze:
An updated version of this has been merged. Thanks for your contribution @marcin-je :+1: 

Superseded by https://github.com/coreos/prometheus-operator/pull/1934

PR #1935, opened 2018-09-25
contrib/kube-prometheus: Set antiAffinity for Prometheus & Alertmanager

### LINKED ISSUE #1930: anti-affinity mixin for kube-prometheus
Currently the kube-prometheus stack comes with no anti-affinity rules defined for Alertmanagers or Prometheus servers, which could cause them to all be scheduled onto the same nodes. If that node goes down, then all of monitoring would be down. This is a situation we should allow to be easily avoided.

I propose to add a mixin for the kube-prometheus stack, so that no two Alertmanagers and Prometheus servers created through the same object are scheduled onto the same node.

@mxinden @squat @s-urbaniak @metalmatze

metalmatze:
This will set podAntiAffinity for Prometheus & Alertmanager signaling the kube scheduler to schedule pods on different nodes.
I've tested these updated mixins in my private cluster. Works great and as expected.
I'm kind of suprised we didn't had this. 😇 

Closes #1930 

/cc @brancz

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-anti-affinity.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-anti-affinity.libsonnet
@@ -1,0 +1,39 @@
+local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
+local statefulSet = k.apps.v1beta2.statefulSet;
+local affinity = statefulSet.mixin.spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecutionType;
+local matchExpression = affinity.mixin.podAffinityTerm.labelSelector.matchExpressionsType;
+
+{
+  local antiaffinity(key, values) = {
+    affinity: {
+      podAntiAffinity: {
+        preferredDuringSchedulingIgnoredDuringExecution: [
+          affinity.new() +
+          affinity.withWeight(100) +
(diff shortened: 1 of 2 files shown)


contrib/kube-prometheus/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
```diff
@@ -100,6 +104,25 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
           baseImage: $._config.imageRepos.alertmanager,
           nodeSelector: { 'beta.kubernetes.io/os': 'linux' },
           serviceAccountName: 'alertmanager-' + $._config.alertmanager.name,
+
+          affinity: {
```
brancz: Although I understand why we would want this by default, the generated manifests now result in something that won't schedule on minikube.

We should just make this a mixin, so users can just do

```
(import "kube-prometheus.libsonnet") +
(import "kube-prometheus-anti-affinity.libsonnet)
```

metalmatze:
Alright, I've reworked this to be a file `kube-prometheus-anti-affinity.libsonnet` that can be imported to extend Prometheus & Alertmanager to add anti affinity.

brancz:
lgtm :+1: nice!

mxinden:
@metalmatze can you create a PR on the cluster-monitoring-operator to also apply `kube-prometheus-anti-affinity.libsonnet`?

mxinden:
And link the following Bugzilla ticket in the description https://bugzilla.redhat.com/show_bug.cgi?id=1635802

PR #1938, opened 2018-09-25
contrib/kube-prometheus fix KubeQuotaExceeded potential division by zero

lefeverd:
This fixes an issue with the KubeQuotaExceeded alert when having some hard limit set to 0 (for instance services.loadbalancers set to 0).
The alert was triggering with +Inf or NaN results.

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
@@ -1,0 +1,64 @@
+local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
+local service = k.core.v1.service;
+local servicePort = k.core.v1.service.mixin.spec.portsType;
+
+
+{
+  _config+:: {
+    versions+:: {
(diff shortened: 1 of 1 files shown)


brancz:
Thanks for the PR! That alert is actually defined in the kubernetes monitoring mixin, meaning we have to do the change there first: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/770d35207fdfebf23cfaa178049e684674fec387/alerts/resource_alerts.libsonnet#L80

Then we can update the kubernetes-mixin dependency we use here and re-generate the alert. Would you like to take care of that?

brancz:
If you want to do the PR against the other repo soonish I also happy with leaving this one open and just updating it once the PR lands in the kubernetes-mixin.

lefeverd:
Already merged: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/95

metalmatze:
I've just merged https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/95 so this PR should be good to update.
Please run `jb update`(where jb is https://github.com/jsonnet-bundler/jsonnet-bundler) in `contrib/kube-prometheus` and then run `make generate` in the root of this repo. 
Thanks.

PR #1946, opened 2018-09-27
Set Owner References on Created Services

### LINKED ISSUE #1945: No Owner Reference on Prometheus Service
**What did you do?**
Created `Prometheus` instance

`Service` with `ownerReference` back to the created `Prometheus` instance

`Service` with no `ownerReferences`

**Environment**

`quay.io/coreos/prometheus-operator@sha256:3daa69a8c6c2f1d35dcf1fe48a7cd8b230e55f5229a1ded438f687debade5bcf`

```
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.0", GitCommit:"91e7b4fd31fcd3d5f436da26c980becec37ceefe", GitTreeState:"clean", BuildDate:"2018-06-27T20:17:28Z", GoVersion:"go1.10.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2018-09-19T23:27:19Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: 
OpenShift 3.11

* Manifests:
n/a

* Prometheus Operator Logs:
n/a

alecmerdler:
### Description

When creating the `Service` for a `Prometheus`/`Alertmanager` instance, the Operator will now correctly update `metadata.ownerReferences`.

Resolves #1945

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -133,6 +133,7 @@ func CreateOrUpdateService(sclient clientv1.ServiceInterface, svc *v1.Service) e
 		}
 	} else {
 		svc.ResourceVersion = service.ResourceVersion
+		svc.SetOwnerReferences(mergeOwnerReferences(service.GetOwnerReferences(), svc.GetOwnerReferences()))
 		_, err := sclient.Update(svc)
 		if err != nil && !apierrors.IsNotFound(err) {
 			return errors.Wrap(err, "updating service object failed")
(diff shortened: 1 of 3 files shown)


brancz:
This service is not unique to a Prometheus object, it's one Service per namespace, so this is actually a bit more complicated. We need to list all Prometheus objects from the cache and set all Prometheus objects as owners.

pkg/k8sutil/k8sutil.go
```diff
@@ -133,6 +133,7 @@ func CreateOrUpdateService(sclient clientv1.ServiceInterface, svc *v1.Service) e
 		}
 	} else {
 		svc.ResourceVersion = service.ResourceVersion
+		svc.SetOwnerReferences(mergeOwnerReferences(service.GetOwnerReferences(), svc.GetOwnerReferences()))
```
brancz: This way we're never pruning non-existant prometheus/alertmanager owner references right? I think we should.

alecmerdler: @brancz An undocumented feature of Kubernetes GC is that it will actually remove the `ownerReference` from the child object once it is deleted, so we don't need to do this ourselves. We actually have a [test for this in OLM](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/gc_e2e_test.go#L22).

brancz: Woah, I had no idea. Thanks for sharing!

PR #1950, opened 2018-09-28
remove duplicate alertmanager down alert

squat:
cc @brancz @sichvoge 

This commit removes the handwritten Alertmanager down alert since there is already an automatically generated `X down` alert for every job.


brancz:
lgtm

PR #1954, opened 2018-09-28
pkg/prometheus: Update Thanos to 0.1.0

metalmatze:
This supersedes #1934, as I can't push to that branch.
The only change is reverting to using `tsdb.path` for the sidecar.

I've tested this branch running on my cluster and so far it seems good.


brancz:
lgtm on green

PR #1956, opened 2018-09-28
Remote read config generator

### LINKED ISSUE #1955: bearer_token is missing from remote_read config
**What did you do?**
Created a pometheus object in k8s with remoteRead and RemoteWrite enabled using bearerToken

The generated prometheus.conf should contain bearer_token for both remote_read and remote_write 

Bearer_token is missing from remote_read

**Environment**

	master

kovszilard:
Fixed generateRemoteReadConfig function to generate bearer_token

Fixes #1955


kovszilard:
@metalmatze looks like, I copied it from there

PR #1958, opened 2018-10-01
objectify alertmanager config

nrobert13:
this way it will be possible to merge configs

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
@@ -14,22 +14,31 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
 
     alertmanager+:: {
       name: $._config.alertmanager.name,
-      config: |||
-        global:
-          resolve_timeout: 5m
-        route:
-          group_by: ['job']
-          group_wait: 30s
-          group_interval: 5m
-          repeat_interval: 12h
-          receiver: 'null'
-          routes:
-          - match:
-              alertname: DeadMansSwitch
-            receiver: 'null'
-        receivers:
-        - name: 'null'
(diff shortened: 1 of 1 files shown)


contrib/kube-prometheus/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
```diff
@@ -38,7 +47,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
     secret:
       local secret = k.core.v1.secret;
 
-      secret.new('alertmanager-' + $._config.alertmanager.name, { 'alertmanager.yaml': std.base64($._config.alertmanager.config) }) +
+      secret.new('alertmanager-' + $._config.alertmanager.name, { 'alertmanager.yaml': std.base64(std.manifestYamlDoc($._config.alertmanager.config)) }) +
```
metalmatze: This would be a breaking change for all the people that have overwritten their Alertmanager config.
Not really sure if we want to do this, even though I like the change personally.

nrobert13:
@metalmatze , I guess the second commit should assure backward compatibility ...

metalmatze:
Yes, that looks good from the perspective of going forward. Not sure if we want to maintain both options though. What do others think?

mxinden:
Using json over yaml here sounds like a great idea to me, as it enables all the benefits jsonnet can bring us here. Not a strong opinion though.

metalmatze:
I think we can merge this later, right? I quite like it this way too.

mxinden:
@metalmatze this is kube-prometheus only. As we are only releasing the Prometheus Operator I don't think this needs to be included in the v0.24 milestone.

brancz:
Any reason why we're not merging this?

metalmatze:
I think we can merge! 😊

metacyclic:
Will do @brancz, thanks!

PR #1967, opened 2018-10-02
remove aggressive log line 

### LINKED ISSUE #1966: reduce alertmanager controller's aggressive logging
**What did you see**

This line being printed 50+ times per second in operator logs:

`level=info ts=2018-10-02T12:25:57.013722259Z caller=operator.go:402 component=alertmanageroperator msg="sync alertmanager" key=namespace/pod`

**What did you expect to see**

The above line being printed way less times. It floods 99.9% of the operator's logs.

on master:
https://github.com/coreos/prometheus-operator/blob/master/pkg/alertmanager/operator.go#L445

**Environment**

any

	0.23.2 / latest

	any / 1.9.10

* Prometheus Operator Logs:

```
level=info ts=2018-10-02T12:25:57.013722259Z caller=operator.go:402 component=alertmanageroperator msg="sync alertmanager" key=namespace/pod

```

nikopen:
Resolves #1966 

If needed, an 'apparition' can be printed every 10 seconds or so whether it syncs without any problems or not?

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -442,8 +442,6 @@ func (c *Operator) sync(key string) error {
 		return nil
 	}
 
-	level.Info(c.logger).Log("msg", "sync alertmanager", "key", key)
-
 	// Create governing service if it doesn't exist.
 	svcClient := c.kclient.Core().Services(am.Namespace)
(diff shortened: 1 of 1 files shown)


metalmatze:
I don't think we should remove this log line as the log level can be changed with a CLI flag.

PR #1970, opened 2018-10-03
Add an example of aggregated ClusterRoles for the operator CRDs

astefanutti:


mxinden:
Hi @astefanutti,

Thanks a bunch for the documentation contribution. This is very much appreciated! Great that you have already used the `[embedmd]` feature. 

As we would threat `example/rbac/prometheus-operator-crd/prometheus-operator-crd-cluster-roles.yaml` as an example only I don't think we need to write them in jsonnet like we do with all the other yaml files in this repo. What are thoughts on this by others? (// @s-urbaniak @metalmatze @squat )

metalmatze:
Yes, as far as I can see none of those examples are somehow generated in `examples/`.
LGTM

mxinden:
Everything inside `example/rbac/prometheus-operator` is generated via the Prometheus Operator jsonnet library @metalmatze.

mxinden:
Thanks @astefanutti for the detailed docs!

PR #1974, opened 2018-10-03
operator: Add watch-secrets flag to enable/disable secret watching

davidhao3300:
This allows for more restricted RBAC in which the Prometheus Operator does not have
view access to all secrets in the cluster.

We would like to use the Prometheus Operator with an RBAC scheme that more strictly regulates what secrets the Operator can view/watch. There were two key places I found where Secrets viewing weren't necessary for those who don't use Basic Auth:
 - Listing secrets in namespaces containing ServiceMonitors, regardless of whether any of the ServiceMonitors specify any basic auth configs.
- Watching secrets for updates to trigger a Prometheus refresh

The first point's behavior has been changed to looping through the service monitors, and if basic auth is configured, list the secrets for the namespace like before. This allows us to minimize how many namespace's secrets we list, and ultimately avoid it entirely if no ServiceMonitors specify basic auth configs.

The second point is addressed by adding a flag to toggle watching secrets. Setting --watch-secrets to false means that the operator will not attempt to watch all the secrets, at the cost of the Prometheus Operator not triggering a refresh if a secret is updated. This cost is documented in the Getting Started guide, and it's a cost we're willing to take in exchange for restricting the Operator's RBAC permissions.

In the long-run, it might be better to watch/stop watching namespace secrets as service monitors with basic auth configs come and go, but that'll require mucking around with the multi-namespace logic and keeping track of when we can stop watching a namespace

### DIFF
--- a/Documentation/user-guides/getting-started.md
+++ b/Documentation/user-guides/getting-started.md
@@ -61,6 +61,12 @@ rules:
   - ""
   resources:
   - configmaps
+  # You may choose to restrict this to the namespaces the Prometheus objects are in instead of all the namespaces
+  # by setting --watch-secrets=false in the operator and configuring the proper RBAC in a Role. The operator will need
+  # access to all verbs for the Secrets in the namespaces that the Prometheus objects are in, and view acces in namespaces
+  # where ServiceMonitors specify Basic Auth configurations.
+  # By doing so, the operator will not watch secrets in all namespaces at the cost of not refreshing the Prometheus configuration
+  # upon relevant secret changes, such as when a Basic Auth secret is altered.
   - secrets
   verbs:
   - '*'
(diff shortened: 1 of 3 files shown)


metalmatze:
I think I like the idea of only watching secrets when necessary.
Yet, it only makes sense when we also have RBAC rules that turn off access to these secrets when the operator is not watching.

squat:
Generally, I do like the idea of tightening down the operator, however this introduces some odd edge cases that I suspect could cause users to experience undesired behavior. If you want to run the prometheus operator with fewer privileges, why not run the operator with the `--namespaces=x,y,z` flag to constrain it to only watch the handful of namespaces you care about? This way, the rbac roles need only give the operator access to those namespaces.

davidhao3300:
My impression of the --namespaces flag is that it causes the Operator to not watch ServiceMonitors in namespaces out of that collection.
This our desired setup:
The Prometheus Operator has all rights to secrets in the namespaces with Prometheus objects in them.
The Prometheus Operator watches all ServiceMonitors in all namespaces.

For our cluster, we have customized the rbac scheme documented in this repo so that the operator does not have cluster-wide admin permissions.

Perhaps it would be better to instead have a flag that specifies what namespaces the operator is allowed to watch secrets in?

davidhao3300:
I can also split this PR into 2 parts, since I believe changing the "Listing secrets in namespaces containing ServiceMonitors, regardless of whether any of the ServiceMonitors specify any basic auth configs." behavior is desirable, regardless of how we address watching secrets in all namespaces

PR #1983, opened 2018-10-08
*: Update to Prometheus v2.4.3

mrueg:


mxinden:
@mrueg could you run `cd contrib/kube-prometheus && jb update && make generate-in-docker` and commit the resulting changes? We missed this in our docs so far. Will be fixed in https://github.com/coreos/prometheus-operator/pull/1989.

mrueg:
@mxinden done.

metalmatze:
Our current workflow is kind of broken and the jsonnetfile.lock.json needs to reference one of the previous commits. Squashing the commits in this PR broke it, I guess.

I've gone ahead and pushed a fix with the correct version reference in the `jsonnetfile.lock.json`. I hope that's ok.

metalmatze:
Yep. Thanks @mrueg, it will be in the next release :relaxed:

mrueg:
Thanks @metalmatze for the quick fix. I think it would be cool to have a single make target that does that (ideally compares it to latest commit in remote master) and runs jb inside docker.

PR #1984, opened 2018-10-08
Add triggerBy metric to alertmanager and overhault metrics in general

metalmatze:
As discussed with @mxinden we want to have more metrics from the Alertmanager controller to better debug issues like #1966 in the future. Hence we add the `prometheus_operator_triggered_total` to the Alertmanager controller as well.

While working on adding this metric, we felt that the overall structure of metric names doesn't feel Prometheus idiomatic. Therefore, we renamed a few of them in this PR as well, to be consistent overall.

This is a **breaking change** for 4 metrics that are now combined within 2 metrics with additional labels.

```diff
- prometheus_operator_alertmanager_reconcile_errors_total
- prometheus_operator_prometheus_reconcile_errors_total
+ prometheus_operator_reconcile_errors_total{controller="alertmanager"}
+ prometheus_operator_reconcile_errors_total{controller="prometheus"}

- prometheus_operator_alertmanager_spec_replicas
- prometheus_operator_prometheus_spec_replicas
+ prometheus_operator_spec_replicas{controller="alertmanager"}
+ prometheus_operator_spec_replicas{controller="prometheus"}

prometheus_operator_node_address_lookup_errors_total

# This now for alertmanager as well and thus we introduced the labels there too
+ prometheus_operator_triggered_total{controller="alertmanager"}
+ prometheus_operator_triggered_total{controller="prometheus"}
```

/cc @squat @s-urbaniak


pkg/alertmanager/operator.go
```diff
@@ -146,12 +146,13 @@ func New(c prometheusoperator.Config, logger log.Logger) (*Operator, error) {
...
 func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
 	c.statefulsetErrors = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "prometheus_operator_alertmanager_reconcile_errors_total",
+		Name: "prometheus_operator_reconcile_errors_total",
 		Help: "Number of errors that occurred while reconciling the alertmanager statefulset",
 	})
 
 	r.MustRegister(
 		c.statefulsetErrors,
+		c.triggerByCounterVec,
```
squat: nit: this should be in the next commit

pkg/alertmanager/operator.go
```diff
@@ -150,6 +151,12 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
 		Help: "Number of errors that occurred while reconciling the alertmanager statefulset",
 	})
 
+	c.triggerByCounterVec = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "prometheus_operator_triggered_total",
+		Help: "Number of times a Kubernetes object add, delete or update event" +
+			" triggered Prometheus Operator to reconcile an object",
```
squat: tiny nit: could we make this a full sentence and add a `the` between `triggered` and `Prometheus Operator` so that it reads more cleanly?

squat:
This looks great overall and is a welcome change! One question, what is the reason for bumping the version of client_golang in this PR? Can we do it in a different pr to scope down the change set?

pkg/alertmanager/operator.go
metalmatze: Are you talking about commit or PR? Because all commits will be squashed in the end. :blush:

squat:
Ok great; the new version is needed for that feature. :+1:

pkg/alertmanager/operator.go
squat: commit. cool as long as it's squashed then sounds good :)

metalmatze:
The `prometheus_operator_spec_replicas` was actually broken and the operator thus panicking. It's fixed by created the metric with the same help text and the same set of labels.
This time I won't be able to squash as the `jb update` now relies on one of the commits.
Mergable on green from my side.

metalmatze:
Please don't yet merge this PR as it is. I was talking to Björn yesterday and currying (wrapping) the counters might be actually better. I'll investigate this first thing tomorrow.

metalmatze:
Alright, this is ready for review again.
We're now using currying for the counters that are shared by the controllers. The metrics that the controllers themselves register still use the wrapped registry.

cmd/operator/main.go
```diff
@@ -196,8 +189,39 @@ func Main() int {
 		return 1
 	}
 
-	po.RegisterMetrics(r)
-	ao.RegisterMetrics(r)
+	reconcileErrors := prometheus.NewCounterVec(prometheus.CounterOpts{
```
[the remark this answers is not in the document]
metalmatze: Personally I don't do these postfixes, but we should be consistent and thus will add it. :+1:

pkg/prometheus/operator.go
```diff
@@ -276,27 +276,18 @@ func New(conf Config, logger log.Logger) (*Operator, error) {
...
-func (c *Operator) RegisterMetrics(r prometheus.Registerer) {
-	c.statefulsetErrors = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "prometheus_operator_prometheus_reconcile_errors_total",
-		Help: "Number of errors that occurred while reconciling the alertmanager statefulset",
-	})
+func (c *Operator) RegisterMetrics(r prometheus.Registerer, reconcileErrors *prometheus.CounterVec, triggerBy *prometheus.CounterVec) {
+	c.reconcileErrors = reconcileErrors
+	c.triggerByCounterVec = triggerBy
+
+	c.reconcileErrors.With(prometheus.Labels{}).Add(0)
```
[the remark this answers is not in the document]
metalmatze: Yep. I wanted to test if the metric is still working and saw it does not appear. Therefore we should init the counter with 0. (it's a best practice AFAIK anyway)

cmd/operator/main.go
```diff
@@ -27,18 +27,17 @@ import (
...
-
-	"github.com/coreos/prometheus-operator/pkg/alertmanager"
+	alertmanagercontroller "github.com/coreos/prometheus-operator/pkg/alertmanager"
 	"github.com/coreos/prometheus-operator/pkg/api"
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/client/monitoring/v1"
 	prometheuscontroller "github.com/coreos/prometheus-operator/pkg/prometheus"
 	"github.com/coreos/prometheus-operator/pkg/version"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
```
squat: we did something a little funny here. normally we like to keep imports separated by origin-type:
stdlib, vendored, and project. it should be like it was before but with go-kit placed along with the other vendored dependencies

squat: we can clean this up later though

pkg/alertmanager/operator.go
```diff
@@ -321,7 +313,7 @@ func (c *Operator) processNextWorkItem() bool {
 		return true
 	}
 
-	c.statefulsetErrors.Inc()
+	c.reconcileErrors.With(prometheus.Labels{}).Add(0)
```
squat: This should be `Inc()`, no?

metalmatze: Yes, that's a bug! :clap: Thanks.

metalmatze:
I've corrected the sorting of imports and fixed the bug that only did `Add(0)` instead of `Inc()`. Squashed everything in the end.

cmd/operator/main.go
```diff
@@ -196,8 +190,39 @@ func Main() int {
...
+		prometheus.WrapRegistererWith(prometheusLabels, r),
+		reconcileErrorsCounter.MustCurryWith(prometheusLabels),
+		triggerByCounter.MustCurryWith(prometheusLabels),
+	)
+
+	alertmanagerLabels := prometheus.Labels{"controller": "alertmanager"}
+	ao.RegisterMetrics(
+		prometheus.WrapRegistererWith(alertmanagerLabels, r),
+		reconcileErrorsCounter.MustCurryWith(alertmanagerLabels),
+		triggerByCounter.MustCurryWith(prometheusLabels),
```
squat: We should have `alertmanagerLabels` here

PR #1993, opened 2018-10-10
Add the ability to mount CA certificate bundles from secrets

vatine:


vatine:
This is to resolve https://github.com/coreos/prometheus-operator/issues/1992, which has been stumping us for quite a while now.

vatine:
Yes, but I need to mount a private CA certificate bundle, rather than a client certificate and I cannot see a good way of getting the normal secrets list sticking my CA cert bundle where the TLS library picks this up.

brancz:
Actually the solution to this in Prometheus is to specify the ca file in the configuration, you can do so via the ServiceMonitor.

vatine:
Cool, this seems to be available in Alertmanager as well. Let me give that a quick test and hopefully we can close this once I've seen it work.

quick691fr:
> Actually the solution to this in Prometheus is to specify the ca file in the configuration, you can do so via the ServiceMonitor.

Can you be more explicit on how to achieve that ? 

Thanks and happy new year

brancz:
I think something went wrong when updating your PR. Can you rebase it?

brancz:
Actually looking at this PR. You can accomplish the same already with volumes and the containers array no?

vatine:
Sorry, I thought this PR had been closed and the update was me trying to freshen my fork. I will close this now, as it is old, and not needed.

PR #1994, opened 2018-10-10
add ThanosSpec.ClusterAdvertisePort and ThanosSpec.GRPCAdvertisePort

wklken:
for advertise the sidecar via service:NodePort with `externalTrafficPolicy: Local`
for build a global scale thanos cluster on top of multiple k8s clusters
with prometheus-operator

------

relate to: https://github.com/improbable-eng/thanos/issues/454

- we have deployed the prometheus-operator in each k8s cluster, and we need to maintain all the prometheus via thanos.
- thanos is based on gossip, so the sidecar should advertise itself
- while all the components are not in the same cluster, the sidecar can advertise via NodePort. e.g. 
- use `service:[type=NodePort,  nodePort=AdvertisePort]`, use the `externalTrafficPolicy: Local`, the requests to `NodeIP:AdvertisePort` will route to the sidecar on the same Node.


wklken:
I just found the proposal https://github.com/improbable-eng/thanos/issues/484 and the related pull requests, most of them were merged recently.

Closed :)

PR #1997, opened 2018-10-10
operator: Remove unnecessary secret-list calls

davidhao3300:
Avoid asking for secrets when none of the ServiceMonitors in the namespace
configure basic auth.

Broken out from https://github.com/coreos/prometheus-operator/pull/1974

The change by itself doesn't allow us to reduce the RBAC permissions we need to allow for the Operator to work, but it does make future work around this easier.


metalmatze:
LGTM

davidhao3300:
Could someone with permissions merge the PR?

pkg/prometheus/operator.go
```diff
@@ -1196,24 +1196,23 @@ func (c *Operator) loadBasicAuthSecrets(
...
 	secrets := map[string]BasicAuthCredentials{}
 	for _, mon := range mons {
 		for i, ep := range mon.Spec.Endpoints {
 			if ep.BasicAuth != nil {
-				credentials, err := loadBasicAuthSecret(ep.BasicAuth, sMonSecretMap[mon.Namespace])
+				smNamespace := mon.Namespace
+				// Load namespace's secrets if not previously fetched
+				if sMonSecretMap[smNamespace] == nil {
+					msClient := c.kclient.CoreV1().Secrets(smNamespace)
+					listSecrets, err := msClient.List(metav1.ListOptions{})
```
squat: one problem I see with this current patch is that if a single ServiceMonitor has multiple endpoints (which they often do), we will be listing all the secrets for the same namespace multiple times when there is no need. We should either keep the slice of secrets in a variable in the in the scope of the `mon` for loop or always list from the object cache so we are not making list calls to the API every time.

davidhao3300: could you explain why that is? As we iterate through the endpoints of a single ServiceMonitor, the first time we notice the namespace hasn't been queried for secrets, fetch the secrets, and then store in the smonSecretMap on line 1212. The next endpoint iteration should notice the cached secrets

squat: As I commented and Frederic reiterated, we should only list the secrets from the object cache rather than directly querying the API

davidhao3300: got it, I see what you mean now

redbaron: IMHO `List` permission is excessive, SMon references concrete object names, so prometheus operators should  `c.kclient.CoreV1().Secrets(smNamespace).Get()` them.

Authors of SMon CR can then provide rolebinding for promether-operator to read  just given secrets without allowing `List`ing all secrets in the namespace

redbaron: Another take on the same problem: https://github.com/coreos/prometheus-operator/pull/2202

PR #1999, opened 2018-10-11
Changelog: Add note about changed namespaces flag in v0.24

metalmatze:
As @squat mentioned in https://github.com/coreos/prometheus-operator/pull/1995#discussion_r224437818 we should make it obvious that the `--namespace` flag changed to `--namespaces` and now takes a comma-separated list of namespaces as string.

/cc @squat @mxinden

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,15 @@ Alertmanager controller.
 
 For a full description see: https://github.com/coreos/prometheus-operator/pull/1984#issue-221139702
 
+In order to support multiple namespaces, the `--namespace` flag changed to `--namespaces`
+and accepts and comma-separated list of namespaces as a string.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -14,12 +14,15 @@ Alertmanager controller.
 
 For a full description see: https://github.com/coreos/prometheus-operator/pull/1984#issue-221139702
 
+Due to supporting multiple namespaces the `--namespace` flag changed to `--namespaces` and
+accepts and comma-separated list of namespaces as string.
```
squat: Maybe:
```
In order to support multiple namespaces, the `--namespace` flag changed to `--namespaces` and accepts and comma-separated list of namespaces as a string.
```

PR #2001, opened 2018-10-11
Fix api docs not escaping pipe character in tables

jfchevrette:
I've found that the pipe characters in `[0-9]+(ms|s|m|h|d|w|y)` as seen in markdown tables generated from types were actually parsed as column separators by Github's markdown parser.

This PR adds an escape function to po-docgen and uses it to escape pipe characters when generating a markdown table.

Ran `make generate` to rebuild the docs following this fix.


squat:
Rather than adding a new function and a new formatting point for fields, we should escape the `|` character in the same place where we escape all the other characters in the Godoc body https://github.com/coreos/prometheus-operator/blob/master/cmd/po-docgen/api.go#L184-L188

PR #2004, opened 2018-10-12
contrib/kube-prometheus: Update to use prometheus-operator v0.24

metalmatze:
After releasing the operator yesterday kube-prometheus also should use the new operator version.

/cc @squat


contrib/kube-prometheus/README.md
```diff
@@ -222,7 +222,7 @@ These are the available fields with their respective default values:
         kubeStateMetrics: "v1.3.1",
         kubeRbacProxy: "v0.3.1",
         addonResizer: "1.0",
-        prometheusOperator: "v0.23.2",
+        prometheusOperator: "v0.24.2",
```
squat: we should have 0.24.0 here

PR #2013, opened 2018-10-15
Prevent TargetPort from being set by default

LiliC:
Currently despite the `omitEmpty` field, it was always set to 0. This prevents setting the value when we don't explicitly set it. When creating the resource programmatically you would end up with `TargetPort: 0`, this prevents that. Added a small test to make sure the change works.


squat:
@LiliC thanks for contributing this! CI is failing because we need to re-generate some documentation and the client. Can you run `make generate-in-docker`?

LiliC:
@squat Done. PTALA, thanks!

PR #2016, opened 2018-10-16
truncate volume name when it exceed 63 chars

khevse:
Failed to create a pvc(gluster)
```
Warning

ProvisioningFailed  3s    persistentvolume-controller  
Failed to provision volume with StorageClass "gluster": 
failed to create volume: failed to create endpoint/service monitoring/glusterfs-dynamic-prometheus-mon-prometheus-db-prometheus-mon-prometheus-0: 
error creating service: Service "glusterfs-dynamic-prometheus-mon-prometheus-db-prometheus-mon-prometheus-0" is invalid: 
metadata.name: Invalid value: "glusterfs-dynamic-prometheus-mon-prometheus-db-prometheus-mon-prometheus-0": 
must be no more than 63 characters
```

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -199,14 +199,12 @@ func NewCustomResourceDefinition(crdKind monitoringv1.CrdKind, group string, lab
 func SanitizeVolumeName(name string) string {
 	name = strings.ToLower(name)
 	name = invalidDNS1123Characters.ReplaceAllString(name, "-")
-	if len(name) > validation.DNS1123LabelMaxLength {
-		name = name[0:validation.DNS1123LabelMaxLength]
-	}
-	return strings.Trim(name, "-")
+
+	return truncateVolumeName(name)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/k8sutil/k8sutil.go
```diff
@@ -203,3 +204,64 @@ func SanitizeVolumeName(name string) string {
 	}
 	return strings.Trim(name, "-")
 }
+
+// TruncateVolumeName truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec)
+func TruncateVolumeName(name string) string {
```
squat: This function duplicates most of the functionality from the `SanitizeVolumeName` func above. That func makes a volume name DNS1123 compliant. These funcs should either be reconciled or, we should just use that one wholesale where this new one was used.

khevse: I agree. I fixed my mistake.

leeyan85:
Is the issue resolved ? It makes me mad

khevse:
> Is the issue resolved ? It makes me mad
Good. I will test another solution and report the result.

khevse:
A conflict can appear with other names after truncating. Upgrade is not planned.

PR #2020, opened 2018-10-16
contrib/kube-prometheus: Bump the prometheus-operator jsonnet dep

dlespiau:
kube-prometheus now contains a script, sync-to-internal-registry.jsonnet, that
depends on commit 8c6a68760010347134e41f3aa3d73c68eb094a1b to work. Bump the prometheus-operator hash accordingly.

EDIT: removed the kube-prometheus bump and amended the commit, because it was actually a downgrade, not a bump.

### DIFF
--- a/contrib/kube-prometheus/jsonnetfile.lock.json
+++ b/contrib/kube-prometheus/jsonnetfile.lock.json
@@ -68,7 +68,7 @@
                     "subdir": "jsonnet/prometheus-operator"
                 }
             },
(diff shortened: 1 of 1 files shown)


dlespiau:
This is followup of PR #1960. The bump couldn't be done there (at least that's what I remember) because the prometheus-operator jsonnet files are git cloned from the upstream URL and the needed hash only existed in the PR, not upstream master.

brancz:
Thanks! :+1:

PR #2022, opened 2018-10-16
Allow passing additional alert relabel config

eyalzek:
for `prometheus.yaml` configuration file. As discussed in https://github.com/coreos/prometheus-operator/issues/1805


brancz:
You need to run `make generate-in-docker` to re-generate all the generated files, it's a big part of why the CI jobs are failing.

brancz:
Sorry that was a bug we merged and immediately fixed yesterday, but it seems your changes were done right after we introduced it. Could you just rebase your changes on top of latest master?

eyalzek:
I'm definitely based on latest master, but the command is still failing the same

eyalzek:
I guess that other people have `MFLAGS` defined somewhere? Since in my shell it's empty, the expression `"${MFLAGS[@]}"` evaluates to an empty string which fails the command. Unquoting it works for me, though I'm not sure what would be the consequences in case `MFLAGS` is set to a string with whitespace, so I won't commit that change. Maybe worth taking a look.

eyalzek:
ping @brancz :)

eyalzek:
ping @brancz :)

eyalzek:
ping @brancz

brancz:
sorry I was on PTO, this lgtm :) thanks!

PR #2024, opened 2018-10-17
add configFile reference to alertmanager

lswith:
This addresses https://github.com/coreos/prometheus-operator/issues/2003


brancz:
I think I would prefer just `config` or `configSecret`, otherwise this just needs a `make generate-in-docker` :slightly_smiling_face: .

lswith:
Updated to use `configSecret`

pkg/client/monitoring/v1/types.go
```diff
@@ -629,6 +629,8 @@ type AlertmanagerSpec struct {
 	// object, which shall be mounted into the Alertmanager Pods.
 	// The Secrets are mounted into /etc/alertmanager/secrets/<secret-name>.
 	Secrets []string `json:"secrets,omitempty"`
+	// ConfigSecret is the secret file to use for the alertmanager configuration
```
mxinden: `s/secret file/Kubernetes secret/` would read better to me, what do you think?

lnovara:
@lswith do you have any updates about this PR? 

We are Kustomize heavy users and it would be really nice to have Alertmanager get secrets in a more transparent way.

Let me know if I can be of any help.

Thanks.

brancz:
For what its worth we're happy to review a PR if anyone else wants to pick this up! :slightly_smiling_face:

PR #2026, opened 2018-10-18
kube-prometheus: Example jsonnet ingress, add externalURL

kskewes:
It took me a bit of searching to find that `externalURL` is set in the `prometheus` object/kind spec.
Per the comment below have added the suggested merge into the example ingress.
https://github.com/coreos/prometheus-operator/pull/1838#issuecomment-417747427

One could then duplicate the whole lot for alertmanager. I can add Alertmanager to PR if not too verbose?

### DIFF
--- a/contrib/kube-prometheus/docs/exposing-prometheus-alertmanager-grafana-ingress.md
+++ b/contrib/kube-prometheus/docs/exposing-prometheus-alertmanager-grafana-ingress.md
@@ -24,7 +24,8 @@ htpasswd -c auth <username>
 
 In order to use this a secret needs to be created containing the name of the `htpasswd`, and with annotations on the Ingress object basic auth can be configured.
 
-[embedmd]:# (../examples/ingress.jsonnet)
+Also, the applications provide external links to themselves in alerts and various places. When an ingress is used in front of the applications these links need to be based on the external URL's. This can be configured for each application in jsonnet.
+
 ```jsonnet
 local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
 local secret = k.core.v1.secret;
@@ -38,6 +39,13 @@ local kp =
   {
     _config+:: {
       namespace: 'monitoring',
+    },
+    prometheus+:: {
+      prometheus+: {
+        spec+: {
+          externalURL: 'http://prometheus.example.com',
+        },
+      },
     },
     ingress+:: {
       'prometheus-k8s':
@@ -73,7 +81,7 @@ k.core.v1.list.new([
 ])
 ```
 
-In order to expose Alertmanager and Grafana, simply create additional fields containing an ingress object, but simply pointing at the `alertmanager` or `grafana` instead of the `prometheus-k8s` Service. Make sure to also use the correct port respectively, for Alertmanager it is also `web`, for Grafana it is `http`.
+In order to expose Alertmanager and Grafana, simply create additional fields containing an ingress object, but simply pointing at the `alertmanager` or `grafana` instead of the `prometheus-k8s` Service. Make sure to also use the correct port respectively, for Alertmanager it is also `web`, for Grafana it is `http`. Be sure to also specify the appropriate external URL. 
 
 In order to render the ingress objects similar to the other objects use as demonstrated in the [main readme](../README.md#usage):
 
@@ -89,3 +97,5 @@ In order to render the ingress objects similar to the other objects use as demon
 ```
 
 Note, that in comparison only the last line was added, the rest is identical to the original.
+
+See (../examples/ingress.jsonnet) for an example implementation.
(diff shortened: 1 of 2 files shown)


metalmatze:
I'm fine with adding the alertmanager part in here too. Thanks!

kskewes:
After adding Alertmanager's `externalURL` it made sense to also add in Grafana's `root_url`.
Having only Prometheus object seemed light so added the appropriate ingress objects (it could be looped for brevity if one has an opinionated URL naming scheme).

Unfortunately I didn't see the failed test results until submitted this commit (44fd8f8).
Travis is pending now and it will probably fail at the documentation step again.

The [guide here](https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/docs/exposing-prometheus-alertmanager-grafana-ingress.md) is great and might make my changes to ingress.jsonnet example a step too far, removing exercise for user. Again I saw the guide after these edits.

Perhaps I should strip changes right back to just `externalURL` and `root_url` additions and instead add an appropriate section into the guide linked above?

Open to direction.

contrib/kube-prometheus/examples/ingress.jsonnet
```diff
@@ -11,10 +11,71 @@ local kp =
...
+    // Configure External URL's per application
+    alertmanager+:: {
+      alertmanager+: {
+        spec+: {
+          externalURL: 'http://alertmanager.example.com',
+        },
+      },
+    },
+    grafana+:: {
+      config: {
```
metalmatze: Please use the pluses all the way down so it merges with other configs.
```diff
-config: {
+config+: {
```

kskewes: This tripped me up initially when I started customizing Grafana. 

`config: null` is set [upstream](https://github.com/brancz/kubernetes-grafana/blob/master/grafana/grafana.libsonnet#L26) so if you don't override with `config: { ...` then building fails with the error:
`RUNTIME ERROR: binary operator + requires matching types, got null and object.`

I modified my local grafana.libsonnet to `config: {},` and this then allows use of pluses in `ingress.jsonnet` as types will match.
I think the Grafana config section [here](https://github.com/brancz/kubernetes-grafana/blob/master/grafana/grafana.libsonnet#L33) can be rewritten without the `!=null` (and match dashboards, etc). 

I will push commit here with pluses and open a PR to Grafana that is linked here.

kskewes: Unfortunately setting with pluses with a simplified `example.jsonnet` `_config` of: 
```
 _config+:: {                                                                                                                                                                                                                                                                                                               
    namespace: 'monitoring',                                                                                                                                                                                                                                                                                                 
    grafana+:: {                                                                                                                                                                                                                                                                                                             
      config+: {                                                                                                                                                                                                                                                                                                             
        sections+: {                                                                                                                                                                                                                                                                                                         
          server+: {                                                                                                                                                                                                                                                                                                         
            root_url: 'http://grafana.example.com/',                                                                                                                                                                                                                                                                         
          },                                                                                                                                                                                                                                                                                                                 
        },                                                                                                                                                                                                                                                                                                                   
      },                                                                                                                                                                                                                                                                                                                     
    },                                                                                                                                                                                                                                                                                                                       
  },                
```

results in the following error:
```
RUNTIME ERROR: binary operator + requires matching types, got null and object.
        example.jsonnet:(4:16)-(12:6)   object <anonymous>
        /mnt/data/github/my-kube-prometheus/vendor/grafana/grafana.libsonnet:33:9-33    thunk <a>
```

`config: null` set [here](https://github.com/brancz/kubernetes-grafana/blob/master/grafana/grafana.libsonnet#L26)

With the following changes to [`grafana.libsonnet`](https://github.com/brancz/kubernetes-grafana/blob/master/grafana/grafana.libsonnet) however we can compile with both empty and custom `Grafana.config`:

L26: `      config: {},`
L33: `[if std.length($._config.grafana.config) > 0 then 'config']:`

As this affects custom Grafana config for anyone, I think a PR to brancz's Grafana would be good.
The LDAP if statement could be changed in the same way. 

I've done a commit to this repo with the pluses and can do required PR to brancz's Grafana in the next few days.

metalmatze:
I'm fine with the changes and adjusting the docs instead.

metalmatze:
I'm fine with the changes and adjusting the docs instead.

contrib/kube-prometheus/docs/exposing-prometheus-alertmanager-grafana-ingress.md
```diff
@@ -89,3 +97,5 @@ In order to render the ingress objects similar to the other objects use as demon
 ```
 
 Note, that in comparison only the last line was added, the rest is identical to the original.
+
+See (../examples/ingress.jsonnet) for an example implementation.
```
metalmatze:
```suggestion
See [ingress.jsonnet](../examples/ingress.jsonnet) for an example implementation.
```

brancz:
I'd say let's move forward and @kskewes can do a follow up to fix the link :slightly_smiling_face: .

PR #2028, opened 2018-10-18
*: Add ability to mount custom ConfigMaps

brancz:
@squat @mxinden @s-urbaniak @metalmatze


pkg/client/monitoring/v1/types.go
```diff
@@ -629,6 +637,10 @@ type AlertmanagerSpec struct {
 	// object, which shall be mounted into the Alertmanager Pods.
 	// The Secrets are mounted into /etc/alertmanager/secrets/<secret-name>.
 	Secrets []string `json:"secrets,omitempty"`
+	// ConfigMaps is a list of ConfigMaps in the same namespace as the Alertmanager
+	// object, which shall be mounted into the Alertmanager Pods.
+	// The ConfigMaps are mounted into /etc/alertmanager/configmaps/<configmap-name>.
```
mxinden: Missing some details mentioned in the `PrometheusSpec.ConfigMaps` comment:

```
ConfigMap changes after initial creation of a Prometheus object are not reflected in the running Pods. To change the ConfigMaps mounted into the Prometheus Pods, the object must be deleted and recreated with the new list of ConfigMaps.
```

hack/generate-in-docker.sh
```diff
@@ -17,8 +17,9 @@ if hash getenforce 2> /dev/null && getenforce | grep 'Enforcing' > /dev/null; th
   VOLUME_OPTIONS=":Z"
 fi
 
+# shellcheck disable=SC2068
 docker run \
     --rm \
     -u="$(id -u "$USER")":"$(id -g "$USER")" \
     -v "${SCRIPTDIR}/..:/go/src/github.com/coreos/prometheus-operator${VOLUME_OPTIONS}" \
-    po-jsonnet make "${MFLAGS[@]}" generate
+    po-jsonnet make ${MFLAGS[@]} generate
```
mxinden: What was the issue here?

pkg/client/monitoring/v1/types.go
brancz: I actually tested this, and this statement became false when we introduced the reconciliation on illegal updates. Will remove where applicable.

hack/generate-in-docker.sh
brancz: in this case we actually want the flags to expand and if there is nothing, with quotes the command will be `make "" generate`, and `""` is not a valid make target.

brancz:
@s-urbaniak you're right I'll add some tests to this. I tested it manually though and it works like charm.

nikopen:
@brancz does this allow alertmanager to mount its config as a configmap and override the 'secret' requirement?

PR #2039, opened 2018-10-22
[WIP] contrib/kube-prometheus: Add Thanos store StatefulSet

metalmatze:
The current Thanos extension misses the store as StatefulSet.

This PR is currently WIP as we need to discuss the way we want to handle secrets for GCS and S3 configuration-wise in kube-prometheus.

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
@@ -57,8 +57,56 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
     thanosPeerService:
       local thanosPeerPort = servicePort.newNamed('cluster', 10900, 'cluster');
       service.new('thanos-peers', { 'thanos-peer': 'true' }, thanosPeerPort) +
+      service.mixin.metadata.withNamespace($._config.namespace) +
       service.mixin.spec.withType('ClusterIP') +
       service.mixin.spec.withClusterIp('None'),
 
+    thanosStoreStatefulSet:
+      local statefulSet = k.apps.v1beta2.statefulSet;
+      local volume = statefulSet.mixin.spec.template.spec.volumesType;
+      local container = statefulSet.mixin.spec.template.spec.containersType;
+      local containerEnv = container.envType;
+      local containerVolumeMount = container.volumeMountsType;
+
+      local labels = { app: 'thanos', 'thanos-peer': 'true' };
+
+      local c =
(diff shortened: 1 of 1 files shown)


brancz:
looks like this needs resolving some conflicts

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
```diff
@@ -57,8 +57,56 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
...
+      local statefulSet = k.apps.v1beta2.statefulSet;
+      local volume = statefulSet.mixin.spec.template.spec.volumesType;
+      local container = statefulSet.mixin.spec.template.spec.containersType;
+      local containerEnv = container.envType;
+      local containerVolumeMount = container.volumeMountsType;
+
+      local labels = { app: 'thanos', 'thanos-peer': 'true' };
+
+      local c =
+        container.new('thanos-store', 'improbable/thanos:v0.1.0') +
```
elisiano: should the container image be `$._config.imageRepos.thanos + ':' + $._config.versions.thanos` instead? (like the `thanosQueryContainer` for example)

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
```diff
@@ -57,8 +57,56 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
...
+        container.withArgs([
+          'store',
+          '--log.level=debug',
+          '--data-dir=/var/thanos/store',
+          '--cluster.peers=thanos-peers.' + $._config.namespace + '.svc.cluster.local:10900',
```
elisiano: I personally wouldn't use the `.cluster.local` suffix. Some users might have customized that.

elisiano:
This implementation forces users to have s3 key/secret. 
Some users might actually be using `kube2iam` and use a role annotation so that the pod has access to the bucket without needing any credentials (https://gist.github.com/elisiano/a0c8cf83788f08bac577f25bcf830d10#file-kube-prometheus-thanos-compact-libsonnet-L41-L46 ). I was wondering if we could merge the logic in order to support both scenarios.

brancz:
@elisiano I think your request is reasonable, we should integrate it.

metalmatze:
Yep. This PR is work in progress and was merely a copy from what I've quickly written to get the things up in my cluster. It also leaves out GCS buckets entirely.

metalmatze:
I'm adding the Thanos store StatefulSet with the update to v0.2 in #2275.

PR #2042, opened 2018-10-24
helm:* remove all helm

squat:
Remove all helm now that Prometheus Operator is in the official helm chart repo.

cc @metalmatze @ant31 @brancz


squat:
Need to scope this down to the one helm chart that was moved

PR #2055, opened 2018-10-26
Update kube-rbac-proxy

brancz:
As I was already at it I also updated kube-state-metrics as we are not using the latest version yet.

@metalmatze @mxinden @squat


brancz:
I tested the stack on minikube and could not spot any differences (and am not aware of breaking changes in the release), but we should definitely think about how we could exercise the stack on minikube in general.

PR #2057, opened 2018-10-26
Add AlertmanagerMembersInconsistent alerting rule

metalmatze:
If the Alertmanager peers have a partition within one cluster we'd like to alert on that fact.

This rule will compare the reported number of members by each peer against the total number (count) of peers in the cluster.

/cc @mxinden


mxinden:
```
Unit Testing:  tests.yaml
  SUCCESS
```

Cool. Would you mind adding an additional commit updating the `/manifests` directory?

PR #2058, opened 2018-10-26
kube-prometheus: Change K8sCertificateExpirationNotice description

madstacks:
Update the `K8sCertificateExpirationNotice` alert summary and description to
more accurately reflect the meaning of the metric. This metric is used to
observe the expiration of client certificates used for authentication. It
does not mean that the kube-apiserver's server certificate is expiring
which can be implied by the existing description.

### DIFF
--- a/helm/exporter-kubernetes/templates/kubernetes.rules.yaml
+++ b/helm/exporter-kubernetes/templates/kubernetes.rules.yaml
@@ -94,15 +94,15 @@ groups:
     labels:
       severity: warning
     annotations:
-      description: Kubernetes API Certificate is expiring soon (less than 7 days)
-      summary: Kubernetes API Certificate is expiering soon
+      description: A client certificate used for authentication is expiring soon (less than 7 days)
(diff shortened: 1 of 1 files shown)


brancz:
You're absolutely right! Could you please also open a PR to change the description here: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/system_alerts.libsonnet#L135-L157 ? Besides that, the helm charts here are not maintained anymore, as they have been migrated to the upstream stable charts repo. It only continues to exist here as a form of migration for existing users, but it will soon be removed (basically only waiting for a go from the chart maintainer @gianrubio, then they will be removed from this repo).

madstacks:
Thanks @brancz, I will open another PR to kubernetes-mixin. Shall I go ahead and close this PR in favor of a PR to the helm chart?

madstacks:
Closing this PR in favor of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/141 and https://github.com/helm/charts/pull/10840

PR #2069, opened 2018-10-29
kube-prometheus: docs - fix example link

kskewes:
See: https://github.com/coreos/prometheus-operator/pull/2026#discussion_r228917452


squat:
thanks @kskewes!

PR #2071, opened 2018-10-30
pkg/api: expose alertmanager status

squat:
cc @brancz I think that as long as we are exposing statuses via the operator's API, we should also expose Alertmanager statuses.

cc @brancz @metalmatze

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -504,16 +504,19 @@ func ListOptions(name string) metav1.ListOptions {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


brancz:
I totally forgot we do this. Actually I think we should remove the other endpoint. We initially did this to circumvent that TPRs/CRDs used to not be able to have a status subresource, but that's possible now, so we should just remove this and actually write the status into the Prometheus/Alertmanager status subresources.

squat:
Sounds good. I started working on t by saw that Alec just opened #2074. I’ll close this :)

PR #2074, opened 2018-11-01
[WIP] Update Status on Operator Custom Resources

alecmerdler:
### Description

Uses the [Kubernetes CRD `status` sub-resource](https://github.com/kubernetes/kubernetes/pull/55168) to reflect the status for `Prometheus`, `PrometheusRule`, `Alertmanager`, and `ServiceMonitor` instances.

Addresses #887

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -61,6 +61,7 @@ var (
 	prometheusRoute = regexp.MustCompile("/apis/monitoring.coreos.com/" + v1.Version + "/namespaces/(.*)/prometheuses/(.*)/status")
 )
 
+// TODO(alecmerdler): Remove this in favor of writing `status` subresource
 func (api *API) Register(mux *http.ServeMux) {
 	mux.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
 		if prometheusRoute.MatchString(req.URL.Path) {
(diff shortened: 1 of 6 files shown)


squat:
Remember that to first enable the status subresource, you need to add it to the CRD when registering it [0]:
```yaml
...
spec:
  subresources:
    # status enables the status subresource.
    status: {}
```

Recent versions of Antoine’s crdutils package, which we use internally to generate our CRDs, have support for this.

[0] https://github.com/coreos/prometheus-operator/blob/master/pkg/k8sutil/k8sutil.go#L183-L195

brancz:
Thanks a ton for starting this!

As we've never exposed the types, I think it's fair to make some changes to the status struct, I think we should essentially take the status struct of the statefulset for Prometheus and Alertmanager and sync the status of the underlying statefulsets into the prom/am status. Updating the status can then be done outside of the actual sync.

In the future we could think about more advanced statuses like alertmanager cluster status, but that would require a secure channel of communication between the prometheus operator and the alertmanager pods, so that's quite a bit out there I think.

alecmerdler:
@brancz Exactly, the difficulty comes from converting from _on-demand_ status to _in-sync_ status. Reusing the existing types and functions was the goal of this PR; simply adding an `UpdateStatus()` call in the `sync()` method seemed like the best first step to realizing fully supported statuses in the Prometheus Operator.

jan--f:
I suppose this can be closed since #887 has been closed by now? cc. @slashpai @simonpasquier

slashpai:
https://github.com/prometheus-operator/prometheus-operator/issues/887 closed in-favour of https://github.com/prometheus-operator/prometheus-operator/issues/3335. We have implemented Status subresource only for Prometheus in https://github.com/prometheus-operator/prometheus-operator/pull/4580. We still need to do for AlertManager, ServiceMonitor, PodMonitor etc. 

Since this PR is a bit old I will wait for the author to update back to see if interested to continue on work on this :)

PR #2082, opened 2018-11-02
contrib/kube-prometheus: Add Prometheus Adapter to stack

metalmatze:
kube-prometheus already scrapes cAdvisor and knows all the metrics necessary for the metrics.k8s.io API.
The Prometheus Adapter implements this interface and forwards the request from Kubernetes to Prometheus while translating to PromQL.

This is still WIP:

1. The container image used in this deployment is not yet available and needs to be locally build
2. Before deploying this, the `gencerts.sh` inside `experimental/custom-metrics-api` needs to be run. We should probably integrate this into the Prometheus Adapter.

/cc @brancz @s-urbaniak

### DIFF
--- a/contrib/kube-prometheus/README.md
+++ b/contrib/kube-prometheus/README.md
@@ -45,12 +45,15 @@ This stack is meant for cluster monitoring, so it is pre-configured to collect m
 
 ## Prerequisites
 
-You will need a Kubernetes cluster, that's it! By default it is assumed, that the kubelet uses token authN and authZ, as otherwise Prometheus needs a client certificate, which gives it full access to the kubelet, rather than just the metrics. Token authN and authZ allows more fine grained and easier access control.
+You will need a Kubernetes cluster, that's it! By default it is assumed, that the kubelet uses token authentication and authorization, as otherwise Prometheus needs a client certificate, which gives it full access to the kubelet, rather than just the metrics. Token authentication and authorization allows more fine grained and easier access control.
 
 This means the kubelet configuration must contain these flags:
 
 * `--authentication-token-webhook=true` This flag enables, that a `ServiceAccount` token can be used to authenticate against the kubelet(s).
 * `--authorization-mode=Webhook` This flag enables, that the kubelet will perform an RBAC request with the API to determine, whether the requesting entity (Prometheus in this case) is allow to access a resource, in specific for this project the `/metrics` endpoint.
+
+This stack provides [resource metrics](https://github.com/kubernetes/metrics#resource-metrics-api) by deploying the [Prometheus Adapter](https://github.com/DirectXMan12/k8s-prometheus-adapter/).
+This adapter is an Extension API Server and Kubernetes needs to be have this feature enabled, otherwise the adapter has no effect, but is still deployed.
 
 ### minikube
 
@@ -155,6 +158,7 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') + {
 { ['kube-state-metrics-' + name]: kp.kubeStateMetrics[name] for name in std.objectFields(kp.kubeStateMetrics) } +
 { ['alertmanager-' + name]: kp.alertmanager[name] for name in std.objectFields(kp.alertmanager) } +
 { ['prometheus-' + name]: kp.prometheus[name] for name in std.objectFields(kp.prometheus) } +
+{ ['prometheus-adapter-' + name]: kp.prometheusAdapter[name] for name in std.objectFields(kp.prometheusAdapter) } +
 { ['grafana-' + name]: kp.grafana[name] for name in std.objectFields(kp.grafana) }
 ```
 
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
@@ -1,0 +1,198 @@
+local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
+
+{
+  _config+:: {
+    namespace: 'default',
+
+    versions+:: {
+      prometheusAdapter: 'v0.3.0',
+    },
+
+    imageRepos+:: {
+      prometheusAdapter: 'quay.io/coreos/k8s-prometheus-adapter-amd64',
+    },
+
+    prometheusAdapter+:: {
+      name: 'prometheus-adapter',
+      labels: { name: $._config.prometheusAdapter.name },
+      config: |||
+        resourceRules:
+          cpu:
+            containerQuery: sum(rate(container_cpu_usage_seconds_total{<<.LabelMatchers>>}[1m])) by (<<.GroupBy>>)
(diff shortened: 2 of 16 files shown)


brancz:
@elisiano the adapter contains more than just custom metrics, it also contains the resource metrics API and the external metrics API is in progress, however, I do agree we should go with the project's name `k8s-prometheus-adapter`.

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
```diff
@@ -0,0 +1,199 @@
...
+          window: 1m
+      |||,
+    },
+  },
+
+  prometheusAdapter+:: {
+    apiService:
+      {
+        apiVersion: 'apiregistration.k8s.io/v1beta1',
+        kind: 'APIService',
```
s-urbaniak: What happens if the user doesn't have aggregation enabled? I guess there are quite some deployments out there which don't have aggregation configured.

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
```diff
@@ -0,0 +1,199 @@
...
+          name: 'v1beta1.metrics.k8s.io',
+        },
+        spec: {
+          service: {
+            name: $.prometheusAdapter.service.metadata.name,
+            namespace: $._config.namespace,
+          },
+          group: 'metrics.k8s.io',
+          version: 'v1beta1',
+          insecureSkipTLSVerify: true,
```
s-urbaniak: `insecureSkipTLSVerify: true` sounds like a bummer, is there no possibility to inject the API server's CA bundle?

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
metalmatze: True, so we should make this an optional component that can be added to the manifests, similar to how we do it with Thanos or the anti affinity?

s-urbaniak: Sounds like a great idea! :+1: Some simple "opt-in" mechanism. If we have a precedence already, even better :-)

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
metalmatze: I'm not sure. This whole component is basically a copy of [contrib/kube-prometheus/experimental/custom-metrics-api/](https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/experimental/custom-metrics-api/custom-metrics-apiservice.yaml#L11) which a bunch of changes.

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
brancz: It's been beta since 1.7, and since the data served is a totally different source than let's say etcd, there is no other way. tl;dr if you don't have api aggregation, your cluster won't work properly :slightly_smiling_face: .

s-urbaniak: oki, in that case should we mention this prerequisite in the readme then?

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
brancz: In environments where we have more control over this we can (eg. the openshift serving certs controller can inject the `caBundle` into the APIService object). I also talked to the cert-manager folks this morning and they are also not opposed to having functionality like that.

contrib/kube-prometheus/README.md
```diff
@@ -45,13 +45,16 @@ This stack is meant for cluster monitoring, so it is pre-configured to collect m
...
+You will need a Kubernetes cluster, that's it! By default it is assumed, that the kubelet uses token authentication and authorization, as otherwise Prometheus needs a client certificate, which gives it full access to the kubelet, rather than just the metrics. Token authentication and authorization allows more fine grained and easier access control.
 
 This means the kubelet configuration must contain these flags:
 
 * `--authentication-token-webhook=true` This flag enables, that a `ServiceAccount` token can be used to authenticate against the kubelet(s).
 * `--authorization-mode=Webhook` This flag enables, that the kubelet will perform an RBAC request with the API to determine, whether the requesting entity (Prometheus in this case) is allow to access a resource, in specific for this project the `/metrics` endpoint.
 
+This stack provides [resource metrics](https://github.com/kubernetes/metrics#resource-metrics-api) by deploying the Prometheus Adapter.
+This adapter is an Extension API Server and Kubernetes needs to be have this feature enabled.
```
brancz: ", otherwise the adapter has no effect, but is still deployed."

contrib/kube-prometheus/README.md
```diff
@@ -45,13 +45,16 @@ This stack is meant for cluster monitoring, so it is pre-configured to collect m
...
+You will need a Kubernetes cluster, that's it! By default it is assumed, that the kubelet uses token authentication and authorization, as otherwise Prometheus needs a client certificate, which gives it full access to the kubelet, rather than just the metrics. Token authentication and authorization allows more fine grained and easier access control.
 
 This means the kubelet configuration must contain these flags:
 
 * `--authentication-token-webhook=true` This flag enables, that a `ServiceAccount` token can be used to authenticate against the kubelet(s).
 * `--authorization-mode=Webhook` This flag enables, that the kubelet will perform an RBAC request with the API to determine, whether the requesting entity (Prometheus in this case) is allow to access a resource, in specific for this project the `/metrics` endpoint.
 
+This stack provides [resource metrics](https://github.com/kubernetes/metrics#resource-metrics-api) by deploying the Prometheus Adapter.
```
brancz: maybe add a link to the adapter

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
```diff
@@ -0,0 +1,198 @@
...
+      local containerVolumeMount = container.volumeMountsType;
+
+      local c =
+        container.new($._config.prometheusAdapter.name, $._config.imageRepos.prometheusAdapter + ':' + $._config.versions.prometheusAdapter) +
+        container.withArgs([
+          '--cert-dir=/var/run/serving-cert',
+          '--config=/etc/adapter/config.yaml',
+          '--logtostderr=true',
+          '--metrics-relist-interval=1m',
+          '--prometheus-url=http://prometheus-k8s.monitoring.svc:9090/',
```
elisiano: I think the prometheus-url should not be hardcoded but use the `$._config` variables for the prometheus name and namespace if possible.

brancz:
lgtm :+1:

PR #2088, opened 2018-11-05
Add support for Kubespray clusters

aveyrenc:
Add config files and documentation for Kubespray clusters.

The discovery services are almost the same as the kubeadm ones, the only change is the selector's name. Maybe someone jsonnet fluent could factor them?

### DIFF
--- a/contrib/kube-prometheus/README.md
+++ b/contrib/kube-prometheus/README.md
@@ -328,7 +328,7 @@ Jsonnet is a turing complete language, any logic can be reflected in it. It also
 
 ### Cluster Creation Tools
 
-A common example is that not all Kubernetes clusters are created exactly the same way, meaning the configuration to monitor them may be slightly different. For [kubeadm](examples/jsonnet-snippets/kubeadm.jsonnet) and [bootkube](examples/jsonnet-snippets/bootkube.jsonnet) and [kops](examples/jsonnet-snippets/kops.jsonnet) clusters there are mixins available to easily configure these:
+A common example is that not all Kubernetes clusters are created exactly the same way, meaning the configuration to monitor them may be slightly different. For [kubeadm](examples/jsonnet-snippets/kubeadm.jsonnet), [bootkube](examples/jsonnet-snippets/bootkube.jsonnet), [kops](examples/jsonnet-snippets/kops.jsonnet) and [kubespray](examples/jsonnet-snippets/kubespray.jsonnet) clusters there are mixins available to easily configure these:
 
 kubeadm:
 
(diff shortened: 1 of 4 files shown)


contrib/kube-prometheus/README.md
```diff
@@ -328,7 +328,7 @@ Jsonnet is a turing complete language, any logic can be reflected in it. It also
 
 ### Cluster Creation Tools
 
-A common example is that not all Kubernetes clusters are created exactly the same way, meaning the configuration to monitor them may be slightly different. For [kubeadm](examples/jsonnet-snippets/kubeadm.jsonnet) and [bootkube](examples/jsonnet-snippets/bootkube.jsonnet) and [kops](examples/jsonnet-snippets/kops.jsonnet) clusters there are mixins available to easily configure these:
+A common example is that not all Kubernetes clusters are created exactly the same way, meaning the configuration to monitor them may be slightly different. For [kubeadm](examples/jsonnet-snippets/kubeadm.jsonnet) and [bootkube](examples/jsonnet-snippets/bootkube.jsonnet) and [kops](examples/jsonnet-snippets/kops.jsonnet) and [kubespray](examples/jsonnet-snippets/kubespray.jsonnet) clusters there are mixins available to easily configure these:
```
metalmatze:
```suggestion
A common example is that not all Kubernetes clusters are created exactly the same way, meaning the configuration to monitor them may be slightly different. For [kubeadm](examples/jsonnet-snippets/kubeadm.jsonnet), [bootkube](examples/jsonnet-snippets/bootkube.jsonnet), [kops](examples/jsonnet-snippets/kops.jsonnet) and [kubespray](examples/jsonnet-snippets/kubespray.jsonnet) clusters there are mixins available to easily configure these:
```

metalmatze:
LGTM. Just one grammar nitpick.
You can also fix that later on, if you want.

contrib/kube-prometheus/README.md
aveyrenc: Just applied your suggested grammar fix.

brancz:
CI fails, because the "current" version of kube-prometheus configured does not contain the kubespray changes. You just need to run `jb update` in the `contrib/kube-prometheus/` directory and do a separate commit. Sorry for the inconvenience, we're looking into how to solve this better, but this extra step is necessary for the time being.

aveyrenc:
The `jb update` has updated all dependencies and Grafana dashboards definitions have changed.

Should I add those changes to my P/R or bump only the `kube-prmetheus` dependency?

metalmatze:
For simplicity on your side, I would say to only bump the kube-prometheus dependency. Otherwise you might need to update and re-generate more things.

brancz:
lgtm thanks for the addition!

PR #2091, opened 2018-11-05
Update alertmanager retention documentation

### LINKED ISSUE #2085: Alertmanager retention description incorrect
**What did you do?**
Set the alertmanager spec retention to 7d

The alertmanager pod to start up with a 7 day retention parameter. According to the spec this should be allowed: https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#alertmanagerspec

The pod crashlooping with 
```
alertmanager: error: time: unknown unit d in duration 170d, try --help
```
The accepted units are limited to `(ms|s|m|h|)`

**Prometheus Operator version:**
 v0.25.0

Opened an issue https://github.com/prometheus/alertmanager/issues/1605 since this seems like an omission in alertmanager

vsliouniaev:
- fixes #2085

Relates to https://github.com/helm/charts/issues/8967, https://github.com/prometheus/alertmanager/issues/1605


brancz:
Sorry for the inconvenience, you need to run `jb update` in the `contrib/kube-prometheus/` directory to update the dependency for the CI generated `manifests/` directory and commit that change into a separate commit.

vsliouniaev:
Sure thing, I was checking on exactly what to do there but my local env has some broken configuration on it that caused the build failure as well. I'll remove the WIP when I'm sure the tests are flaking as opposed to me doing the wrong thing ;-)

vsliouniaev:
If the relative path isn't an option then it looks like the full update will need to be done in 2 separate PRs

brancz:
The relative paths doesn’t work, I’m sorry for not noticing before, we actually only bump the Prometheus operator dependency on actually tagged releases as that’s consistent with the Prometheus operator version that will run, so not updating kube Prometheus is fine in this PR.

vsliouniaev:
@brancz This should be fine now, sorry it took so long for me to get to it

brancz:
No worries at all, thanks for taking care of it! :+1:

Issue #2094, opened 2018-11-06
Incorrect message in alerts KubeJobFailed and KubeJobCompletion


yuripastushenko:
**What did you do?**
run a kubernetes job with name "my-job" in "default" namespace that is taking more than 30 minutes.

**What did you expect to see?**
Alert:
Job default/my-job is taking more than 30 minutes to complete.

**What did you see instead? Under which circumstances?**
Job /kube-state-metrics is taking more than 30 minutes to complete.

* Prometheus Operator version:
	0.24.0 

I find out that it is mistake in alert in 
contrib/kube-prometheus/manifests/prometheus-rules.yaml

squat:
should be fixed by https://github.com/coreos/prometheus-operator/pull/2096

PR #2095, opened 2018-11-06
Incorrect message in alerts KubeJobFailed and KubeJobCompletion #2094

yuripastushenko:
Fixed message for alert KubeJobCompletion and KubeJobFailed.

The message was:
`Job /kube-state-metrics is taking more than 30 minutes to complete.`

The new message is:
`Job default/my-job is taking more than 30 minutes to complete.`

### DIFF
--- a/contrib/kube-prometheus/manifests/prometheus-rules.yaml
+++ b/contrib/kube-prometheus/manifests/prometheus-rules.yaml
@@ -698,7 +698,7 @@ spec:
         severity: warning
     - alert: KubeJobCompletion
       annotations:
-        message: Job {{ $labels.namespaces }}/{{ $labels.job }} is taking more than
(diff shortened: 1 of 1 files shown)


brancz:
These alerts are not defined in this repository, they are just a dependency that happens to be used here, therefore this needs to be fixed here: https://github.com/kubernetes-monitoring/kubernetes-mixin

metalmatze:
This has been fixed upstream in the monitoring-mixins now.
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/115
We should create another PR to update the dependency.

metalmatze:
Just found out that https://github.com/coreos/prometheus-operator/pull/2096 is taking care of updating the jsonnet dependencies already.

PR #2099, opened 2018-11-07
[WIP] Migrate to go-modules

kuznero:
This is WIP, but feedback is welcome!


brancz:
in a general sense, we're more than happy to migrate if it works :slightly_smiling_face:

kuznero:
> in a general sense, we're more than happy to migrate if it works slightly_smiling_face

:) It was pretty simple, but I'm afraid I have replaced a bit too much ;) it builds, but I didn't get to running tests, and didn't change Makefile, etc. I might consider re-opening it later.

brancz:
that would be great, if you get it to work!

PR #2105, opened 2018-11-07
[WIP] Migrate to go-modules

kuznero:
Related to previous attempt #2099.
Any advise on how to perform testing of such a change are very much welcome!

Ping @brancz.

PR #2107, opened 2018-11-07
Migrate to go-modules

kuznero:
Another version of  #2105. Ping @brancz. This version is less noisy as root module is named such that it was not necessary to change any of the imports.


kuznero:
Green this time around ;)

brancz:
Even with go-mods we would like to keep the `vendor/` directory and the related verification tooling, in order to do offline, immutable builds.

kuznero:
@brancz I think all the necessary stuff is still in the vendor folder.

metalmatze:
I looked into this again. Is there something blocking us from using Go modules? @brancz @squat @mxinden?
Otherwise I can continue with this, as I have a rebased branch on my machine.

mxinden:
No reason to not use go modules from my side. :+1:

brancz:
I've had some difficulties with Kubernetes libs and go mods. My experience has been converting from dep/glide works just fine, but starting from scratch/updating doesn't necessarily work. Can you confirm that those two also work?

metalmatze:
Alright cool. I'll make another branch from the updated branch of this and check how an update of Kubernetes works out. If that's ok, then I will update this PR and merge Go modules. :+1:

brancz:
My experience is ~2 months old, so may very well be fixed in newer go versions. I have no objections.

metalmatze:
This is superseeded by https://github.com/coreos/prometheus-operator/pull/2571. Thanks for the taking the time nevertheless!

PR #2109, opened 2018-11-08
*: Remove attempting to set "secure" security context

### LINKED ISSUE #2102: Instance creation fails with security context issues
**What did you do?**

* Install OKD 3.11
* Deploy OLM
* Create a subscription to the Prometheus Operator
* Run through the "getting started" section

* A working deployment

* The operator successfully deploys
* The prometheus stateful set fails with the following error:

~~~
create Pod prometheus-prometheus-0 in StatefulSet prometheus-prometheus failed error: pods "prometheus-prometheus-0" is forbidden: unable to validate against any security context constraint: [fsGroup: Invalid value: []int64{2000}: 2000 is not an allowed group spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000340000, 1000349999] spec.containers[1].securityContext.securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000340000, 1000349999] spec.containers[2].securityContext.securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000340000, 1000349999]]
~~~

**Environment**

	0.22.2

~~~
$ oc version
oc v3.10.0+dd10d17
kubernetes v1.10.0+b81c8f8
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://foobar:8443
openshift v3.11.0+aeff40b-47
kubernetes v1.11.0+d4cacc0
~~~

* Kubernetes cluster kind: 

	OKD 3.11

* Prometheus Operator Logs:

```
level=info ts=2018-11-07T15:43:06.141416547Z caller=operator.go:737 component=prometheusoperator msg="sync prometheus" key=hono/prometheus
W1107 15:43:06.141507       1 listers.go:63] can not retrieve list of objects using index : Index with name namespace does not exist
W1107 15:43:06.141556       1 listers.go:63] can not retrieve list of objects using index : Index with name namespace does not exist
W1107 15:43:06.151212 1 listers.go:63] can not retrieve list of objects using index : Index with name namespace does not exist
```

brancz:
This was previously a very poor attempt at achieving something similar to what PodSecurityPolicies are for, therefore I am removing this code. There are no compatibility issues, as for any supported volume types, the kubelet sets the ownership recursively for all files of a volume before mounting.

Fixes #2102

@mxinden @metalmatze @squat @s-urbaniak '

/cc @ctron

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -128,7 +128,6 @@ func init() {
 	flagset.StringVar(&cfg.CrdGroup, "crd-apigroup", monitoringv1.Group, "prometheus CRD  API group name")
 	flagset.Var(&cfg.CrdKinds, "crd-kinds", " - EXPERIMENTAL (could be removed in future releases) - customize CRD kind names")
 	flagset.BoolVar(&cfg.EnableValidation, "with-validation", true, "Include the validation spec in the CRD")
-	flagset.BoolVar(&cfg.DisableAutoUserGroup, "disable-auto-user-group", false, "Disables the Prometheus Operator setting the `runAsUser` and `fsGroup` fields in Pods.")
 	flagset.StringVar(&cfg.LocalHost, "localhost", "localhost", "EXPERIMENTAL (could be removed in future releases) - Host used to communicate between local services on a pod. Fixes issues where localhost resolves incorrectly.")
 	flagset.StringVar(&cfg.LogLevel, "log-level", logLevelInfo, fmt.Sprintf("Log level to use. Possible values: %s", strings.Join(availableLogLevels, ", ")))
 	flagset.StringVar(&cfg.LogFormat, "log-format", logFormatLogfmt, fmt.Sprintf("Log format to use. Possible values: %s", strings.Join(availableLogFormats, ", ")))
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -79,7 +79,6 @@ type Config struct {
 	CrdKinds                     monitoringv1.CrdKinds
 	CrdGroup                     string
 	EnableValidation             bool
-	DisableAutoUserGroup         bool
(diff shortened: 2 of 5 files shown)


metalmatze:
After updating the Prometheus Operator to v0.26 I got this error in my setup:
```
level=error ts=2018-12-03T11:43:20.312345098Z caller=main.go:625 err="opening storage failed: open /prometheus/wal/00000733: permission denied"
```

As the securityContext is not set by default anymore, but all my files are still written with user 1000 and group 2000 I needed to set the security like:
```diff
  ...
  prometheus+: {
    prometheus+: {
      spec+: {
        ...
+       securityContext+: {
+         fsGroup: 2000,
+         runAsNonRoot: true,
+         runAsUser: 1000,
+       },
  ...
```

brancz:
@metalmatze do you think we should do a migration to explicitly set it to that? Sounds like a lot of volume types don’t do this (although the kubelet should be doing it).

MarkDeckert:
Also this breaks common PodSecurityPolicy rules that require explicitly setting runAsUser in the deployment.

I believe this is the default in openshift and gce, and most likely also in many bare-metal or self-managed k8s instances.

fsGroup would always need be set as well for mounted volumes although PodSecurityPolicy wouldn't explicitly require this setting in the yaml, it clearly would break PVs depending on what runAsUser is set to.

metalmatze:
@brancz what do you mean by a migration? We could simply set this in jsonnet and have it people override as needed.
It might be the case, that the kubelet doesn't care about my volumes as mine are CIS ones, but just a guess.

brancz:
Actually I'm happy with the jsonnet solution as well.

metalmatze:
Just letting you know in this issue that we merged a jsonnet/yaml based solution into the manifests: https://github.com/coreos/prometheus-operator/pull/2206

PR #2120, opened 2018-11-12
contrib/kube-prometheus: Mount tmpfs to /tmp in Prometheus Adapter

metalmatze:
This will make sure that there's always a /tmp in the container.

Related to https://github.com/coreos/prometheus-operator/issues/2116.

/cc @s-urbaniak @brancz


brancz:
lgtm

squat:
this is such a classic k8s burn. we should make a framed photo :p

PR #2127, opened 2018-11-13
Makefile: use only the first path on GOPATH

pgier:
The build currently fails if the GOPATH includes multiple paths separated by a colon.


brancz:
Thanks! :+1:

PR #2131, opened 2018-11-14
contrib/kube-prometheus: Update experimental Custom Metrics adapter example

metalmatze:
As the Prometheus Adapter is now merged into kube-prometheus we can update the experimental example of Custom Metrics to make use of this.

/cc @s-urbaniak @brancz


brancz:
How about creating an experimental marked mixin in jsonnet?

metalmatze:
That also sounds good for the future! :+1:

brancz:
let's do it as a follow up

PR #2134, opened 2018-11-14
kube-prometheus: Add documentaton to README on the Pod Anti-Affinity mixin

jeromefroe:
Currently there is no documentation on the Pod Anti-Affinity mixin which makes it
difficult to find. To address this, this PR adds a section to the README on how one
can use it.


brancz:
Nice! :+1:

PR #2146, opened 2018-11-20
node-exporter RBAC proxy listen on Pod IP

squat:
This commit adjusts the RBAC proxy for the node-exporter DaemonSet to
only listen on the Pod IP. It also adjusts the ports used by the
node-exporter Pod so that both containers are listening on 9100. The
actual node-exporter listens on 127.0.0.1:9100, while the RBAC proxy
listens on <PODIP>:9100. This ensures that port 9101 is not taken on
the host networking namespace.

cc @brancz @s-urbaniak re: our conversation from earlier in the day.
I think that putting the node-exporter on hostPort 9100 and exposing
the RBAC proxy on the containerPort 9100 is a better option than
taking up another port for the proxy. With this new configuration, the
proxy listens on the Pod IP (i.e. the node's IP since we are on host network)
and the node exporter listen's on the node's localhost. The service NATs
to podIP:9100 so it connects to the RBAC proxy. I tested this on my cluster
and Prometheus correctly scrapes the proxy on HTTPS.

### DIFF
--- a/Documentation/user-guides/cluster-monitoring.md
+++ b/Documentation/user-guides/cluster-monitoring.md
@@ -163,7 +163,7 @@ spec:
     spec:
       containers:
       - args:
-        - --web.listen-address=127.0.0.1:9101
+        - --web.listen-address=127.0.0.1:9100
         - --path.procfs=/host/proc
         - --path.sysfs=/host/sys
         - --collector.filesystem.ignored-mount-points=^/(dev|proc|sys|var/lib/docker/.+)($|/)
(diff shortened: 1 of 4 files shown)


brancz:
I honestly don’t know why we didn’t do this in the first place, I’d say let’s do this for all our uses of the proxy.

brancz:
But let’s take care of the other ones in follow up pull requests.

squat:
> I honestly don’t know why we didn’t do this in the first place, I’d say let’s do this for all our uses of the proxy.

Yeah, that's what I was thinking too. We should standardize this so that all the proxies listen on the same canonical URL that the components typically listens on.

PR #2149, opened 2018-11-21
kube-prometheus fix coredns to work with default recommended setup

brancz:
This PR adapts our default configuration for monitoring coredns to match the recommended default install on kubernetes as specified in the coredns repository: https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed#L150-L172

@metalmatze @squat @mxinden @s-urbaniak

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
@@ -386,11 +386,9 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
           },
         },
         spec: {
-          jobLabel: 'k8s-app',
           selector: {
             matchLabels: {
-              'k8s-app': 'coredns',
-              component: 'metrics',
+              'k8s-app': 'kube-dns',
             },
           },
           namespaceSelector: {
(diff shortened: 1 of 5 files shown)


squat:
agh thank you! I was just going to post a PR when I saw yesterday that scraping of coredns was failing on my PR. It wouldn't be a bad idea to also file a PR against coredns/deployment to upgrade the prom semantics

contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
```diff
@@ -400,7 +398,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
           },
           endpoints: [
             {
-              port: 'http-metrics',
+              port: 'metrics',
```
squat: we're selecting the `metrics` port, but the service doesn't specify this named port [0]. 

[0] https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed#L166-L172

brancz: The pod does though: https://github.com/coredns/deployment/blob/bbf9284adc6bace3caf35a38a131baeda92dee18/kubernetes/coredns.yaml.sed#L122

brancz: Nevertheless I think you're right it would make sense to add it to the service upstream.

BrianChristie: I've sent a PR to coredns/deployment to add the port to the service: https://github.com/coredns/deployment/pull/112

PR #2154, opened 2018-11-22
[WIP] */*: add timeout for the k8s API

s-urbaniak:
Currently, we don't set any timeouts.
This fixes it by setting the following timeouts:
- Sets the timeout on each REST call, see
k8s.io/client-go/rest#Request.Timeout
- Sets the global http timeout on k8s's http client, see
https://github.com/kubernetes/client-go/blob/8abb21031259350aad0799bb42ba213ee8bb3399/rest/config.go#L211-L216

_Note:_ This needs some testing, before merging.

/cc @brancz @squat


metalmatze:
Do we actually need to pass the timeout every single time? Isn't that something that can be done when creating the client once?

s-urbaniak:
@metalmatze yes and no, the http client is initialized only once. But a per-request timeout is also being set, effectively setting its context, see https://github.com/kubernetes/kubernetes/pull/70998.

squat:
@s-urbaniak the per request timeout sets a timeout query parameter so that the server can can close long requests, but arguably that’s not what we really care about here, no? Our case is that the server is unable to close the timeouts because it’s not actually up and running yet so the query parameter doesn’t really help. Maybe we could instead simplify this PR to only set the timeout on the client config OR we could bump the client-go dep to something that includes that new commit and sets all automatically. Otherwise I think we may be complicating our codebase and adding debt for little or no benefit.

s-urbaniak:
@squat agreed, this is starting to become a rabbit hole. I also discussed with @brancz adding client generation here. For now I'll open a separate PR on the http client timeout only and leave this PR open for client generation.

s-urbaniak:
closing in favor of #2163

PR #2157, opened 2018-11-22
cmd/operator: add http client timeout

s-urbaniak:
This is the lightweight version of
https://github.com/coreos/prometheus-operator/pull/2154.

This should be good enough for http client based timeouts.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -26,6 +26,7 @@ import (
 	"os/signal"
 	"strings"
 	"syscall"
+	"time"
 
 	alertmanagercontroller "github.com/coreos/prometheus-operator/pkg/alertmanager"
 	"github.com/coreos/prometheus-operator/pkg/api"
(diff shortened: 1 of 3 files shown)


cmd/operator/main.go
```diff
@@ -132,6 +133,7 @@ func init() {
 	flagset.StringVar(&cfg.LogLevel, "log-level", logLevelInfo, fmt.Sprintf("Log level to use. Possible values: %s", strings.Join(availableLogLevels, ", ")))
 	flagset.StringVar(&cfg.LogFormat, "log-format", logFormatLogfmt, fmt.Sprintf("Log format to use. Possible values: %s", strings.Join(availableLogFormats, ", ")))
 	flagset.BoolVar(&cfg.ManageCRDs, "manage-crds", true, "Manage all CRDs with the Prometheus Operator.")
+	flagset.DurationVar(&cfg.Timeout, "timeout", 32*time.Second, "The timeout for remote calls towards the k8s API.")
```
squat: Nit: wording could be “The timeout for remote calls against the Kubernetes API.”

pkg/prometheus/operator.go
```diff
@@ -147,6 +147,11 @@ type Config struct {
 	LogLevel                     string
 	LogFormat                    string
 	ManageCRDs                   bool
+
+	// The maximum length of time to wait before giving up on a server request towards the k8s API.
```
squat: Same wording a/towards/against/

s-urbaniak:
@squat trying to do that as we speak, not yet tested on my cluster.

s-urbaniak:
hrm, setting a global timeout on the http client is apparently not good for watches :-(

```
E1122 14:27:46.344972       1 streamwatcher.go:109] Unable to decode an event from the watch stream: net/http: request canceled (Client.Timeout exceeded while reading body)
```

s-urbaniak:
closing this PR, as this is a dead-end unfortunately. We need per-use-case timeout settings.

PR #2161, opened 2018-11-22
contrib/kube-prometheus: Add prometheusURL to adapter's jsonnet config

metalmatze:
It should be easier to configure the Prometheus Adapater's prometheusURL from jsonnet. Hence, we move the prometheusURL into the _config and reference it as argument in the Deployment.

As this is a nop for the manifests themselves, nothing needs to be generated.

/cc @s-urbaniak @brancz


s-urbaniak:
looks good :+1:

brancz:
lgtm :+1:

PR #2163, opened 2018-11-26
pkg/client: use kubernetes client generation

s-urbaniak:
Currently, prometheus operator uses custom hand-crafted client code to talk to the Kubernetes API. This is fine, as prometheus operator was one of the first operators out there, but time has moved on.

The upstream kubernetes project proposes operators to use https://github.com/kubernetes/code-generator instead of custom hand-written code.

As we currently also have an issue in OpenShift, where prometheus operator does not create prometheus statefulsets, this PR introduces client code generation based on the above upstream framework.

It is a starting point to investigate the current flakes in OpenShift, but also aligns the code here with other 
operators which moved to client generation.


s-urbaniak:
@brancz @squat @mxinden I am getting there, the remaining e2e failures are pretty mechanical, PTAL for an initial pass.

s-urbaniak:
@brancz fyi, we triggered three times CI to at least have some idea, that this does not introduce flakes.

cmd/operator/main.go
```diff
@@ -125,7 +126,7 @@ func init() {
 	flagset.StringVar(&cfg.ThanosDefaultBaseImage, "thanos-default-base-image", "improbable/thanos", "Thanos default base image")
 	flagset.Var(ns, "namespaces", "Namespaces to scope the interaction of the Prometheus Operator and the apiserver.")
 	flagset.Var(&cfg.Labels, "labels", "Labels to be add to all resources created by the operator")
-	flagset.StringVar(&cfg.CrdGroup, "crd-apigroup", monitoringv1.Group, "prometheus CRD  API group name")
+	flagset.StringVar(&cfg.CrdGroup, "crd-apigroup", monitoring.GroupName, "prometheus CRD  API group name")
```
brancz: for what it's worth, if this makes any problems this might as well be removed, it was introduced as an experimental flag that may be removed at any point

brancz:
Wow, thanks for taking care of this so fast! lgtm :+1:

cmd/operator/main.go
brancz: but it seems that it didn't cause that much complications, so let's keep it as is

s-urbaniak: :+1: I stumbled across and was having raised eye-brows if this actually works, when set to something else. I didn't touch it for backwards compatibility.

PR #2169, opened 2018-11-27
Change examples to be more robust against customizations

redbaron:
Explicitly reference values we are interesting in (jsonnet style)
instead of just following conventions (helm style)


brancz:
Perfect, you nailed exactly why jsonnet is awesome! :)

Thanks! Lgtm :+1:

PR #2170, opened 2018-11-28
fix syntax error in prometheus.libsonnet

fstab:
Got the following error when calling jsonnet:

```
vendor/kube-prometheus/prometheus/prometheus.libsonnet:43:44-47 Expected token then but got (IDENTIFIER, "and")

    [if $._config.prometheus.rules != null and $._config.prometheus.rules != {} then "rules"]:
```

Changing `and` to `&&` fixes that.


brancz:
Thanks! lgtm :+1:

PR #2174, opened 2018-11-28
Remove GOARCH=amd64 to prevent cross-compiling go binaries

jeffdyoung:
Remove GOARCH=amd64 to prevent cross-compiling go binaries for multi-arch container builds.

@brancz


brancz:
lgtm :+1: thanks!

PR #2183, opened 2018-11-30
kube-prometheus: bump prometheus-operator

s-urbaniak:
cc @brancz @mxinden @squat


contrib/kube-prometheus/jsonnet/kube-prometheus/jsonnetfile.json
```diff
@@ -51,4 +51,4 @@
             "version": "master"
         }
     ]
-}
```
brancz: I think you might have an old version of `jb`, I think this might fail CI.

s-urbaniak: the _one_ time I use vim instead of emacs ;-)

PR #2187, opened 2018-11-30
Additional scrape config docs

jmreicha:
More docs


brancz:
Wow, this is really nice, thank you so much! It’s clearly written, to the point and all things that are important are noted. Nothing else I can say, lgtm :+1:.

PR #2191, opened 2018-12-03
kube-prometheus: Fix ExternalUrl key typo in ingress example

kskewes:
Per: https://github.com/coreos/prometheus-operator/pull/2172


brancz:
lgtm on green

PR #2202, opened 2018-12-06
Load ServiceMon secrets directly

redbaron:
Alternative implementation for same problem as in https://github.com/coreos/prometheus-operator/pull/1997

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1160,39 +1160,39 @@ func loadAdditionalScrapeConfigsSecret(additionalScrapeConfigs *v1.SecretKeySele
 	return nil, nil
 }
 
-func loadBasicAuthSecretFromAPI(basicAuth *monitoringv1.BasicAuth, c corev1client.CoreV1Interface, ns string, cache map[string]*v1.Secret) (BasicAuthCredentials, error) {
+func extractCredKey(secret *v1.Secret, sel v1.SecretKeySelector, cred string) (string, error) {
+	if s, ok := secret.Data[sel.Key]; ok {
+		return string(s), nil
+	} else {
+		return "", fmt.Errorf("secret %s key %q in secret %q not found", cred, sel.Key, sel.Name)
+	}
+}
+
+func getCredFromSecret(c corev1client.SecretInterface, sel v1.SecretKeySelector, cred string, cacheKey string, cache map[string]*v1.Secret) (_ string, err error) {
+	var s *v1.Secret
+	var ok bool
+
+	if s, ok = cache[cacheKey]; !ok {
+		if s, err = c.Get(sel.Name, metav1.GetOptions{}); err != nil {
+			return "", fmt.Errorf("unable to fetch %s secret %q: %s", cred, sel.Name, err)
+		}
+		cache[cacheKey] = s
+	}
+	return extractCredKey(s, sel, cred)
+}
+
+func loadBasicAuthSecretFromAPI(basicAuth *monitoringv1.BasicAuth, c corev1client.CoreV1Interface, ns string, cache map[string]*v1.Secret) (_ BasicAuthCredentials, err error) {
 	var username string
 	var password string
-	var s *v1.Secret
-	var err error
-	var ok bool
 
 	sClient := c.Secrets(ns)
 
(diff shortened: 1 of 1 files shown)


pkg/prometheus/operator.go
```diff
@@ -1159,28 +1160,60 @@ func loadAdditionalScrapeConfigsSecret(additionalScrapeConfigs *v1.SecretKeySele
 	return nil, nil
 }
 
+func extractCredKey(secret *v1.Secret, sel v1.SecretKeySelector, cred string) (string, error) {
+	if s, ok := secret.Data[sel.Key]; ok {
+		return string(s), nil
+	} else {
```
squat: nit: there is no need for this `else` here since you are returning early

pkg/prometheus/operator.go
```diff
@@ -1159,28 +1160,60 @@ func loadAdditionalScrapeConfigsSecret(additionalScrapeConfigs *v1.SecretKeySele
...
+	if s, ok = cache[cacheKey]; !ok {
+		if s, err = c.Get(sel.Name, metav1.GetOptions{}); err != nil {
+			return "", fmt.Errorf("unable to fetch %s secret %q: %s", cred, sel.Name, err)
+		}
+		cache[cacheKey] = s
+	}
+	return extractCredKey(s, sel, cred)
+}
+
+func loadBasicAuthSecretFromAPI(basicAuth *monitoringv1.BasicAuth, c corev1client.CoreV1Interface, ns string, cache map[string]*v1.Secret) (_ BasicAuthCredentials, err error) {
```
squat: nit: we are standardized around not using named returns for this project

pkg/prometheus/operator.go
```diff
@@ -1205,25 +1238,12 @@ func (c *Operator) loadBasicAuthSecrets(
...
-		}
-	}
-
 	secrets := map[string]BasicAuthCredentials{}
+	nsSecretCache := make(map[string]*v1.Secret)
 	for _, mon := range mons {
 		for i, ep := range mon.Spec.Endpoints {
 			if ep.BasicAuth != nil {
-				credentials, err := loadBasicAuthSecret(ep.BasicAuth, sMonSecretMap[mon.Namespace])
+				credentials, err := loadBasicAuthSecretFromAPI(ep.BasicAuth, c.kclient.CoreV1(), mon.Namespace, nsSecretCache)
```
squat: Rather than unnecessarily GETing secrets from the API when we could be GETing them from the in-memory secrets cache.

redbaron: This PR is a preparation for https://github.com/coreos/prometheus-operator/pull/2197 , with endgoal of reducing scope of RBAC permissions required for namespaces with ServiceMonitor and PrometheusRule CRs

brancz:
I generally like the approach, I'm a little concerned about the amount of calls this may produce against the Kubernetes API, but as this is bound to the number of secrets specified in servicemonitors I think it's ok for now. The only thing that seems a little strange is that users would have to bind the role to access these secrets to the prometheus operator. How do they know what to bind it to?

pkg/prometheus/operator.go
redbaron: updated

pkg/prometheus/operator.go
redbaron: updated

redbaron:
>  but as this is bound to the number of secrets specified in servicemonitors I think it's ok for now. 

there is a per-iteration caching to to try to minimize it, I doubt it is going to be effective though

> The only thing that seems a little strange is that users would have to bind the role to access these secrets to the prometheus operator service account. How do they know what to bind it to?

Doesn't prometheus-operator update `.status` field? It can add error/success reports there so that users can know if their SMon was processed and what went wrong.

pkg/prometheus/operator.go
squat: Got it :+1:  there was no mention of 2197 so it was not clear why the lister was not used

redbaron:
I think all comments are addressed now, is it a good time to merge it? I have more PRs regarding reducing RBAC scope and working within restricted namespaces and would like to keep number of them in-flight to a bare minimum

brancz:
Sorry for taking so long, KubeCon kept me busy real good. This lgtm :+1:

Issue #2204, opened 2018-12-07
Prometheus error when using latest image from upstream (2.6.0-rc)


carlosedp:
**What did you do?**
Launched a kube-prometheus stack using custom images built from latest source.

**What did you expect to see?**
Stack up

**What did you see instead? Under which circumstances?**
Error while launching prometheus container inside prometheus-k8s-0 pod. The error states duplicate parameters:

```
Pod: prometheus-k8s-0

Error parsing commandline arguments: flag 'web.console.templates' cannot be repeated
usage: prometheus [<flags>]

The Prometheus monitoring server

Flags:
  -h, --help                     Show context-sensitive help (also try
                                 --help-long and --help-man).
      --version                  Show application version.
...
```

The pod is ran with correct arguments:

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  annotations:
    prometheus-operator-input-hash: "15217728158418187864"
  creationTimestamp: "2018-12-06T20:17:09Z"
  generation: 1
  labels:
    prometheus: k8s
  name: prometheus-k8s
  namespace: monitoring
  ownerReferences:
  - apiVersion: monitoring.coreos.com/v1
    blockOwnerDeletion: true
    controller: true
    kind: Prometheus
    name: k8s
    uid: e864c605-f993-11e8-b511-3ea301e3f9d0
  resourceVersion: "186482"
  selfLink: /apis/apps/v1/namespaces/monitoring/statefulsets/prometheus-k8s
  uid: e8d7b90d-f993-11e8-b511-3ea301e3f9d0
spec:
  podManagementPolicy: OrderedReady
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: prometheus
      prometheus: k8s
  serviceName: prometheus-operated
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: prometheus
        prometheus: k8s
    spec:
      containers:
      - args:
        - --web.console.templates=/etc/prometheus/consoles
        - --web.console.libraries=/etc/prometheus/console_libraries
        - --config.file=/etc/prometheus/config_out/prometheus.env.yaml
        - --storage.tsdb.path=/prometheus
        - --storage.tsdb.retention=24h
        - --web.enable-lifecycle
        - --storage.tsdb.no-lockfile
        - --web.route-prefix=/
        image: carlosedp/prometheus:v2.5.0
        imagePullPolicy: IfNotPresent
...
```

[...]

PR #2205, opened 2018-12-07
Fix prometheus start for version 2.6.0 onwards

carlosedp:
Due to the changes on Prometheus Dockerfile (https://github.com/prometheus/prometheus/pull/4796), the arguments and CMD are merged into ENTRYPOINT. This breaks the launch from prometheus-operator causing duplicate arguments.


carlosedp:
More info on the issue https://github.com/coreos/prometheus-operator/issues/2204

simonpasquier:
Note that the final release will probably preserve the old behavior. See https://github.com/prometheus/prometheus/pull/4976

metalmatze:
Closing this for the same reason as closing #2204. If we need these changes after all, we can still reopen. Thanks you nonetheless!

PR #2206, opened 2018-12-07
Set the securityContext for Prometheus and Alertmanager in manifests

metalmatze:
As #2109 removed setting the securityContext in the reconciliation loop of the operator and this broke quite a few people we want to set this as default in the manifests again.

/cc @brancz @squat


squat:
looks nice to me :) :+1:

brancz:
The reason I was suggesting the migration within the Prometheus operator is that we can eventually remove the migration, but at what point can we remove this from kube Prometheus? Because this seems like an odd default to suggest to people.

PR #2208, opened 2018-12-09
Replace SharedIndexInformers with nonshared version of them

redbaron:
Informers were not really shared at all, so to convey
what is  actually happening clearer, use of less powerfull
abstractions is desirable.

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -52,6 +52,16 @@ const (
 	resyncPeriod = 5 * time.Minute
 )
 
+type indexedInformer struct {
+	informer cache.Controller
+	index    cache.Indexer
+}
+
+type informer struct {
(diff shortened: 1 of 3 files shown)


brancz:
I realize we don't do this today, but the statefulset and secrets informers should actually be shared between the prometheus and alertmanager controllers.

brancz:
That sounds good to me.

metalmatze:
The e2e tests seem to timeout. I can't see anything in the Travis logs that seems obvious. Maybe debugging this on a machine with `make test-e2e` makes sense?!

redbaron:
I'll try, is  it just minikube it needs?

redbaron:
It doesn't make it more tidy or performant, so I am closing it, it doesn't worth it

PR #2209, opened 2018-12-10
contrib/kube-prometheus: Set Prometheus Adapter maxSurge

metalmatze:
We want to make sure that at least one Prometheus Adapter is always running while the next one starts. Therefore we're setting the maxSurge to 1 and maxUnavailable to 0.

/cc @squat @brancz @s-urbaniak


squat:
@s-urbaniak that is correct. These are the necessary semantics since what we want to prevent is the case where 0 instances are run.

PR #2212, opened 2018-12-11
kube-prometheus: Add poddisruptionbudgets to ksm cluster role 

mxinden:
With https://github.com/kubernetes/kube-state-metrics/commit/40f293dbde2f69ad0b5ae5875e2f19fed0b749e5 kube-state-metrics exposes metrics about poddisruptionbudgets. Hence kube-state-metrics needs the permissions to collect those metrics via its cluster role.

Thanks @metalmatze for the hint!

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
@@ -41,11 +41,11 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
 
     clusterRole:
       local clusterRole = k.rbac.v1.clusterRole;
-      local policyRule = clusterRole.rulesType;
+      local rulesType = clusterRole.rulesType;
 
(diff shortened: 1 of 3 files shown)


contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -41,11 +41,11 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
 
     clusterRole:
       local clusterRole = k.rbac.v1.clusterRole;
-      local policyRule = clusterRole.rulesType;
+      local rulesType = clusterRole.rulesType;
```
mxinden: Renaming `policyRule` to `rulesType` to reuse word later on for the actual rules for the *policy* API group.

mxinden:
Only updated kube-prometheus dependency in `jsonnet.lock.json`.

metalmatze:
After applying this new ClusterRole I don't see any messages in the logs anymore and the `kube_poddisruptionbudget*` metrics now show up! :+1: :100:

PR #2215, opened 2018-12-12
add the name of the operator namespace to the clusterrolebinding

nrobert13:
in order to be able to run multiple operators in the same cluster

### DIFF
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -28,7 +28,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
       local clusterRoleBinding = k.rbac.v1.clusterRoleBinding;
 
       clusterRoleBinding.new() +
-      clusterRoleBinding.mixin.metadata.withName('prometheus-operator') +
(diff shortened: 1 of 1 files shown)


metalmatze:
For this change to succeed you need to run `jb update` inside `contrib/kube-prometheus`. Then run `make generate` from the root of this repository again (so `cd ../..` out of `contrib/kube-prometheus`). Now the files in `contrib/kube-prometheus/manifests` should be updated. Commit all changes and push them here.

brancz:
I'm not sure this is a generally applicable thing. I think you should just apply your environment's opinion in your jsonnet files.

nrobert13:
@brancz, this would be required in order to manage multiple prometheus instances in different namespaces. or do you have a better option for this?

brancz:
My point is you can just do this customization in your own setup.

```
{
  prometheusOperator+: {
    clusterRoleBinding+: clusterRoleBinding.mixin.metadata.withName('prometheus-operator-' + $._config.namespace),
  },
}
```

nrobert13:
@brancz, I see your point. This is what I do currently, but if we all need to do the same thing to run multiple instances of prometheus in one cluster, I thought maybe it would make sense to get it upstream, unless there's a better way to do it....

PR #2216, opened 2018-12-12
add namespace to the component selectors

nrobert13:
in order to select the right components in case multiple operators are running in the same cluster


metalmatze:
Nice.
I think we should be consistent and also add namespaces to the selectors for other components.

I'm wondering why we don't yet select on the namespaces @brancz @mxinden. I know Tom Wilkie always selects the namespaces.

mxinden:
Adding namespace selector sounds good to me. @brancz what do you think?

brancz:
lgtm

brancz:
Personally I even think we could add the `service` label as well. As a side note, I think this is a very positive development, I think the `job` label is long due to loose relevance.

PR #2217, opened 2018-12-12
Avoids values from other namespaces in pods graphs

solsson:
For example if the same StatefulSet name is used in different namespaces,
graphs could show resources from one namespace with requests or limits from another.

### DIFF
--- a/contrib/kube-prometheus/manifests/grafana-dashboardDefinitions.yaml
+++ b/contrib/kube-prometheus/manifests/grafana-dashboardDefinitions.yaml
@@ -6195,7 +6195,7 @@ items:
                           "steppedLine": false,
                           "targets": [
                               {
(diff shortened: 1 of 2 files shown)


metalmatze:
Thank you for this contribution.
The manifests folder gets generated with jsonnet and the dashboards are actually defined in the [kubernetes-monitoring/kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin) project.
It would be great if you could create a PR against that repository.
[This](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/dea7e9907daab8e84d4b87d2067869074c9ad275/dashboards/pods.libsonnet) is the file you want to update. Once the PR is merged there we can update kube-prometheus' dependency and have this in kube-prometheus as well.

solsson:
I made another PR. Didn't find the dashboard under ./helm in the other repo, so maybe that should be merged here? I can revert the second commit in that case.

metalmatze:
With https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/134 merged, this should trickle down to kube-prometheus within the next days. Thanks for taking care of this @solsson! :+1:

PR #2220, opened 2018-12-13
Use Parallel podManagementPolicy in the alert manager

kargakis:
@squat this is one change. It seems that prometheus already uses Parallel in master. Now, how can I make the v3.11 version of the monitoring operator use Parallel for both statefulsets?


kargakis:
Closing in favor of coreos/prometheus-operator#2221

PR #2222, opened 2018-12-13
pkg/prometheus: Do not pass podManagementPolicy and stick with Parallel

metalmatze:
With the discussion happening in #2221 we stumbled upon the fact that we already create the Prometheus StatefulSet with `podManagementPolicy: Parallel`. However, we never see this being applied with the StatefulSets themselves and the Pods are managed with `OrderedReady`.

As it turns out we pass in an empty string to `makeStatefulSet` so that the StatefulSet doesn't contain this key value pair in the applied struct. Kubernetes defaults to `OrderedReady`.

https://github.com/coreos/prometheus-operator/compare/master...metalmatze:prometheus-parallel-pod-management?expand=1#diff-94707ea811ea06be1af9eb4049a2e13bL1018

@squat and I can't find a reason anymore why we should pass this down and now always set `podManagementPolicy: Parallel`. The StatefulSet can't be updated with this field anyway and would need to be recreated.

@brancz, do you still know of a reason to keep it as is, with passing it as argument to `makeStatefulSet`?

Also we should be fine with using `Parallel` by default for Prometheus. Alertmanager should stay `OrderedReady` as written in #2221.

/cc @mxinden @s-urbaniak @kargakis @mjudeikis @jim-minter

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -86,7 +86,6 @@ var (
 
 func makeStatefulSet(
 	p monitoringv1.Prometheus,
-	previousPodManagementPolicy appsv1.PodManagementPolicyType,
 	config *Config,
 	ruleConfigMapNames []string,
 	inputHash string,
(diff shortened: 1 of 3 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -211,10 +210,6 @@ func makeStatefulSet(
 		statefulset.Spec.VolumeClaimTemplates = append(statefulset.Spec.VolumeClaimTemplates, pvcTemplate)
 	}
 
-	// Updates to statefulset spec for fields other than 'replicas',
-	// 'template', and 'updateStrategy' are forbidden.
-	statefulset.Spec.PodManagementPolicy = previousPodManagementPolicy
```
mxinden: I am a bit confused. Given that Kubernetes defaults to `orderedready` and the fact that we would like `parallel`, don't we need to set `parallel` here?

mxinden: Ah, nevermind, this is set in https://github.com/coreos/prometheus-operator/blob/7d73503f0f0be5565b0e017fea04a6c753c24432/pkg/prometheus/statefulset.go#L753

PR #2226, opened 2018-12-14
[thanos] support objstore.config=type:s3 for v0.2.0

gangseok514:
from v0.2.0, thanos removed --s3.xxx options and added objstore

thanos value in prometheus crd changes to objstore values in thanos-sidecar

Signed-off-by: gangseok.lee <gangseok.lee@samsung.com>

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -35,7 +35,7 @@ import (
 const (
 	governingServiceName     = "prometheus-operated"
 	DefaultPrometheusVersion = "v2.5.0"
-	DefaultThanosVersion     = "v0.1.0"
+	DefaultThanosVersion     = "v0.2.1"
 	defaultRetention         = "24h"
 	storageDir               = "/prometheus"
 	confDir                  = "/etc/prometheus/config"
@@ -698,17 +698,23 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		}
 
 		if p.Spec.Thanos.S3 != nil {
+			objstore := []string{"--objstore.config=type: S3"}
+			objstore = append(objstore, "config:")
+
 			if p.Spec.Thanos.S3.Bucket != nil {
-				thanosArgs = append(thanosArgs, fmt.Sprintf("--s3.bucket=%s", *p.Spec.Thanos.S3.Bucket))
+				objstore = append(objstore, fmt.Sprintf("  bucket: %s", *p.Spec.Thanos.S3.Bucket))
 			}
 			if p.Spec.Thanos.S3.Endpoint != nil {
-				thanosArgs = append(thanosArgs, fmt.Sprintf("--s3.endpoint=%s", *p.Spec.Thanos.S3.Endpoint))
+				objstore = append(objstore, fmt.Sprintf("  endpoint: %s", *p.Spec.Thanos.S3.Endpoint))
 			}
 			if p.Spec.Thanos.S3.Insecure != nil && *p.Spec.Thanos.S3.Insecure {
-				thanosArgs = append(thanosArgs, "--s3.insecure")
+				objstore = append(objstore, "  insecure: true")
 			}
 			if p.Spec.Thanos.S3.SignatureVersion2 != nil && *p.Spec.Thanos.S3.SignatureVersion2 {
-				thanosArgs = append(thanosArgs, "--s3.signature-version2")
+				objstore = append(objstore, "  signature-version2: true")
+			}
+			if p.Spec.Thanos.S3.EncryptSSE != nil && *p.Spec.Thanos.S3.EncryptSSE {
+				objstore = append(objstore, "  encrypt-sse: true")
 			}
 			if p.Spec.Thanos.S3.AccessKey != nil {
 				envVars = append(envVars, v1.EnvVar{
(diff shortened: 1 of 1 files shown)


gangseok514:
@metalmatze Hello, This is my first PR. would you review this? Thanks!

gangseok514:
@metalmatze Hello, I'd like to hear some little advice. :)

abursavich:
I would suggest leaving the existing code as is, marking the old fields as deprecated, and adding a new field to the ThanosSpec, e.g.:

```go
type ThanosSpec struct {
    // ... other fields omitted ...
    // Deprecated: Use ObjectStorageConfig for thanos v0.2.0+
    GCS *ThanosGCSSpec `json:"gcs,omitempty"`
    // Deprecated: Use ObjectStorageConfig for thanos v0.2.0+
    S3 *ThanosS3Spec `json:"s3,omitempty"`
    // Object storage configuration.
    ObjectStorageConfig *v1.SecretKeySelector `json:"objectStorageConfig,omitempty"`
}
```

```go
if p.Spec.Thanos.ObjectStorageConfig != nil {
    envVars = append(envVars, v1.EnvVar{
        Name: "OBJECT_STORAGE_CONFIG",
        ValueFrom: &v1.EnvVarSource{
            SecretKeyRef: p.Spec.Thanos.ObjectStorageConfig,
        },
    })
    thanosArgs = append(thanosArgs, "--objstore.config", "$(OBJECT_STORAGE_CONFIG)")
}
```

The main benefit is that prometheus-operator doesn't have to explicitly support every thanos storage backend (GCS, S3, Azure, Swift, etc.).

It also allows users to manage their own upgrade path. It would be possible for someone to pick up a new version of prometheus-operator and stick with an older version of thanos until they're ready. 

Eventually, the `GCS` and `S3` fields could be removed from `ThanosSpec` an older versions of thanos can be abandoned.

abursavich:
Actually, I found an existing issue for Azure storage that supports something like what I suggested: https://github.com/coreos/prometheus-operator/issues/2133#issuecomment-438989379

mmack:
This works for me.
Dockerimage: https://cloud.docker.com/u/mmack/repository/docker/mmack/prometheus-operator
Commit: https://github.com/mackIOConsulting/prometheus-operator/commit/10481d5c5622149beee3275428882438d72659a6

pkg/prometheus/statefulset.go
```diff
@@ -35,7 +35,7 @@ import (
 const (
 	governingServiceName     = "prometheus-operated"
 	DefaultPrometheusVersion = "v2.5.0"
-	DefaultThanosVersion     = "v0.1.0"
+	DefaultThanosVersion     = "v0.2.0"
```
bwplotka: Can we use 0.2.1?

pkg/prometheus/statefulset.go
```diff
@@ -699,9 +705,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 						SecretKeyRef: p.Spec.Thanos.S3.AccessKey,
 					},
 				})
-			}
-			if p.Spec.Thanos.S3.EncryptSSE != nil && *p.Spec.Thanos.S3.EncryptSSE {
-				thanosArgs = append(thanosArgs, "--s3.encrypt-sse")
+				objstore = append(objstore, "  access_key: $(S3_ACCESS_KEY)")
```
bwplotka: I guess those are defined? They must be in pod spec env vars for container.

metalmatze:
I'll create a unit test for the generation.
I was thinking about generating a secret with the config would be helpful, as that secret could also be used by the store component for example.

bwplotka:
Up to you, we use compacted  format as it is just nicer to read all configuration from flags. (:

abursavich:
Please let users create their own secret with the configuration that they reference in the ThanoSpec. Users can reference the same secret for the store and compactor, but most importantly, it decouples the thanos storage backend config from the prometheus-operator config. Thanos will be able to support arbitrary backends without needed to push changes to prometheus-operator.

pkg/prometheus/statefulset.go
gangseok514: the above of it...

```
                          if p.Spec.Thanos.S3.AccessKey != nil {
				envVars = append(envVars, v1.EnvVar{
					Name: "S3_ACCESS_KEY",
					ValueFrom: &v1.EnvVarSource{
						SecretKeyRef: p.Spec.Thanos.S3.AccessKey,
					},
				})
			}
```

is defiend.

pkg/prometheus/statefulset.go
gangseok514: Done!

gangseok514:
I think @abursavich's approach(https://github.com/coreos/prometheus-operator/pull/2264) is better than my changes. But the approach contains "not secret" data. containing only secret data in Secret resource is more proper use.

gangseok514:
@metalmatze I think that separating the secret from prometheus-operator is better. because as @abursavich  said that the secret will be shared by thanos-store/compactor.

mmack:
Yeah this is the better approach!

brancz:
cc @metalmatze

metalmatze:
Perfect. Referecing a secret with the config seems like a good idea, agree. 
Let's continue with #2264 and discuss there.

Thanks for the initiative @gangseok514 :+1:

PR #2230, opened 2018-12-17
contrib: add quotes when value begins with colon

kyletru:
This makes the yaml parsable by ruby which tries to interpret strings
beginning with : as symbols


metalmatze:
Interesting that this fails to parse in Ruby.
Sadly, we generate those files with jsonnet and we use https://github.com/brancz/gojsontoyaml to convert the internal JSON to YAML. That binary uses https://github.com/ghodss/yaml to the YAML part.
Maybe opening an issue with them might start a conversation in the right place.

kyletru:
Thanks, I'll follow up with the other projects and close.

PR #2243, opened 2018-12-21
write compressed prometheus config secret

wleese:
Thanos reloader now transparently gunzips secrets to configEnvsubstFilename.
New version wants ruleDir as []string. It doesn't seem we're using this feature anyway, so I removed it.


metalmatze:
gzipping the config sounds like a good improvement indeed.
Looking at the changes, I was wondering why you actually pulled in the thanos dependency, when you simply reference `encoding/zip`. I couldn't find any references and was probably missing them.

wleese:
@metalmatze I did this for the reloader (./vendor/github.com/improbable-eng/thanos/pkg/reloader/reloader.go), which was updated to be able to unpack gzipped config files. Was this unnecessary (golang rookie here)? I also had to add the constraint because an "dep ensure -update github.com/improbable-eng/thanos@cd7963372a11cea93a832584be099d2bfa0ac19a" failed otherwise.

brancz:
run `git clean -xfd && make generate-in-docker` in the root of the repo please and commit the changes :slightly_smiling_face:

the e2e failure needs some test adaptation as we check for the plaintext config to be equal to something which isn't the case anymore

brancz:
This should do it (feel free to use a newer version of kubernetes of course :smile:  ). https://github.com/coreos/prometheus-operator#running-end-to-end-tests-on-local-minikube-cluster

test/e2e/prometheus_test.go
```diff
@@ -290,13 +291,19 @@ scrape_configs:
...
+	firstConfigCompressed := buf.Bytes()
+
 	cfg := &v1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: fmt.Sprintf("prometheus-%s", name),
 		},
 		Data: map[string][]byte{
-			"prometheus.yaml": []byte(firstConfig),
-			"configmaps.json": []byte("{}"),
+			"prometheus.yaml.zip": firstConfigCompressed,
```
brancz: this should be `prometheus.yaml.gz` no?

test/e2e/prometheus_test.go
```diff
@@ -332,12 +339,17 @@ scrape_configs:
...
+	}
+	secondConfigCompressed := buf.Bytes()
+
 	cfg, err := framework.KubeClient.CoreV1().Secrets(ns).Get(cfg.Name, metav1.GetOptions{})
 	if err != nil {
 		t.Fatal(errors.Wrap(err, "could not retrieve previous secret"))
 	}
 
-	cfg.Data["prometheus.yaml"] = []byte(secondConfig)
+	cfg.Data["prometheus.yaml.zip"] = secondConfigCompressed
```
brancz: same as above

wleese:
e2e fixed

brancz:
Awesome! Thanks a lot for taking care of this! lgtm :+1:

wleese:
Thanks again - I really appreciate all the help! @brancz

PR #2244, opened 2018-12-22
Fix dashboard title typo

arsane:
The dashboard is actually about CPU usage.


metalmatze:
the helm charts are not longer maintained here, please submit this pull request to the upstream stable charts repo where this chart has been migrated to

PR #2252, opened 2018-12-28
add service monitor for grafana

jouve:


brancz:
The grafana definition is from https://github.com/brancz/kubernetes-grafana, could you add the servicemonitor there in jsonnet and then integrate that here? Thanks!

jouve:
hi,

i updated the PR to use jsonnet.

if I move this to grafana, there will be some circular dependencies: kube-prometheus needs grafana and grafana will need kube-prometheus ?

brancz:
yes I know what you mean, this works just as well I think :+1: thanks!

PR #2257, opened 2019-01-01
Change label k8s-app to component

okamototk:
Kubernetes installed by kubeadm uses component label for scheduler and controller-manager.
Default label should be component instead of k8s-app.


brancz:
these helm chart files are no longer maintained, please change them in the upstream stable charts repo

PR #2259, opened 2019-01-03
Use nobody

mmack:
For kubernetes clusters which do not allow running images as root user this is required. 
For all others it does not hurt though.

### DIFF
--- a/cmd/prometheus-config-reloader/Dockerfile
+++ b/cmd/prometheus-config-reloader/Dockerfile
@@ -2,4 +2,8 @@ FROM quay.io/prometheus/busybox:latest
 
 ADD prometheus-config-reloader /bin/prometheus-config-reloader
 
+RUN chown nobody:nogroup /bin/prometheus-config-reloader
+
+USER nobody
+
 ENTRYPOINT ["/bin/prometheus-config-reloader"]


metalmatze:
The StatefulSet that the Prometheus Operator generates currently uses a securityContext to set this:
```
  securityContext:
    fsGroup: 2000
    runAsNonRoot: true
    runAsUser: 1000
```
This also includes this container as far as I  can see.

mmack:
It's the sidecar that creates the problem, not prometheus itself.

```
  prometheus-config-reloader:
    Container ID:  
    Image:         quay.io/coreos/prometheus-config-reloader:v0.26.0
    Image ID:      
    Port:          <none>
    Host Port:     <none>
    Command:
      /bin/prometheus-config-reloader
    Args:
      --log-format=logfmt
      --reload-url=http://localhost:9090/-/reload
      --config-file=/etc/prometheus/config/prometheus.yaml
      --config-envsubst-file=/etc/prometheus/config_out/prometheus.env.yaml
    State:          Waiting
      Reason:       CreateContainerConfigError
    Ready:          False
    Restart Count:  0
    Limits:
      cpu:     50m
      memory:  50Mi
    Requests:
      cpu:     50m
      memory:  50Mi
    Environment:
      POD_NAME:  prometheus-lxos-dev-prometheus-0 (v1:metadata.name)
    Mounts:
      /etc/prometheus/config from config (rw)
      /etc/prometheus/config_out from config-out (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from admin-token-2msh2 (ro)
```

Error:
```
Warning  Failed     5m6s (x5 over 5m32s)  kubelet, hostname  Error: container has runAsNonRoot and image will run as root
  Warning  Failed     5m6s (x5 over 5m34s)  kubelet, hostname  Error: container has runAsNonRoot and image has non-numeric user (nobody), cannot verify user is non-root

```

metalmatze:
I just double checked, the securityContext also sets the user id for the sidecars.
But I'm fine with running `prometheus-config-reloader` as nobody by default, in case somebody doesn't have the context set.

brancz:
lgtm

marshallford:
I'm still running into this issue: https://github.com/helm/charts/issues/20793

PR #2263, opened 2019-01-06
Fix: Override the url of Prometheus datasource.

Lentil1016:
#### What this pull request do

Hi, in [brancz/kubernetes-grafana](https://github.com/brancz/kubernetes-grafana.git) the url of prometheus datasource ~~is~~ used to be `http://prometheus-k8s.monitoring.svc:9090` by default. And kube-prometheus use this value without override it, which caused grafana not be able to access prometheus when `_config.namepsace` is solved to some value other than `monitoring`, like `kube-system`.

This PR override the url  of `prometheus datasource` config in [brancz/kubernetes-grafana](https://github.com/brancz/kubernetes-grafana.git)  with `'http://prometheus-k8s.' + $._config.namespace + '.svc:9090'`


metalmatze:
With https://github.com/brancz/kubernetes-grafana/pull/56 being merged, we don't need to explicitly set this here now. Instead we need to update the revision for the kubernetes-grafana dependency. Then the change in kubernetes-grafana should take the correct namespace into account.

Lentil1016:
Cool, I'll close this PR for it's not necessary anymore.

metalmatze:
I was about to push the same. :smile:  But I guess you already have done the same :man_shrugging: 
I can confirm that it works now. Changing the namespace to `kube-system` in `contrib/kube-prometheus/example.jsonnet` and then generating the manifests folder with `./build.sh` results in an updated datasource with the correct URL :+1: 
We can merge this, when Travis is green.

Thanks

PR #2265, opened 2019-01-07
Added AlertResendDelay to Prometheus resource

zuzzas:
Added the option to Prometheus custom object, since there wasn't one.

Fixes https://github.com/coreos/prometheus-operator/issues/2253


pkg/apis/monitoring/v1/types.go
```diff
@@ -118,6 +118,8 @@ type PrometheusSpec struct {
 	ScrapeInterval string `json:"scrapeInterval,omitempty"`
 	// Interval between consecutive evaluations.
 	EvaluationInterval string `json:"evaluationInterval,omitempty"`
+	// Minimum amount of time to wait before resending an alert to Alertmanager.
+	AlertResendDelay string `json:"alertResendDelay,omitempty"`
```
brancz: as there are multiple `-rules.alert` flags, could you add two config structs for this?

basically so we end up with

```
rules:
  alert:
    resendDelay: ...
```

brancz:
just one comment otherwise this is looking pretty good, thanks for taking care of it!

zuzzas:
@brancz, it seems that everything is per your desires. Can you review the naming of structs and comments, please? Never been good with those thing.

brancz:
Looks good! Thanks! :+1:

PR #2266, opened 2019-01-08
*: Add image fields with highest precedence

brancz:
This pull request adds `image` fields to AlertmanagerSpec, PrometheusSpec and ThanosSpec as well as appropriate handling in the statefulset logic. In hindsight, we should have probably never introduced `baseImage`, `tag` and `sha` fields, but it doesn't seem like a huge maintenance burden.

@metalmatze @squat @mxinden @s-urbaniak


metalmatze:
So the tl;dr is: Whenever the image field is set it will override everything else, correct?

brancz:
Correct.

PR #2267, opened 2019-01-08
Gopkg.toml: Remove unnecessary override

LiliC:
'github.com/ericchiang/k8s' is not used anywhere, so the override is not
needed.


brancz:
golang dependency management :upside_down_face: 

thanks!

PR #2269, opened 2019-01-08
Lint YAML against our CRD defintions

metalmatze:
For now this is a standalone binary that I simply use by running `go run ./cmd/lint prometheus.yaml`.
It will then log errors if the YAML (that is first converted to JSON so that we can use `decoder.
DisallowUnknownFields()`) find something that is wrong.

Example:
```
$ go run ./cmd/lint contrib/kube-prometheus/manifests/prometheus-prometheus.yaml                                                                                                  

2019/01/08 12:40:18 prometheus is invalid: json: unknown field "nodeSelectors"
exit status 1
```

It should help us debug and identify typos and other mistakes more quickly.
We can also integrate the functionality some other place, but I also like the standalone binary.

/cc @brancz @s-urbaniak  @squat

### DIFF
--- a/cmd/lint/main.go
+++ b/cmd/lint/main.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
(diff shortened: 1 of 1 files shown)


brancz:
nice :+1: lgtm

PR #2271, opened 2019-01-08
Add SessionAffinity ClusterIP to prometheus service

BennX:
This PR adds SessionAffinity ClusterIP to the prometheus service to prevent grafana from querying different instances when loading/updating metrics.


brancz:
lgtm :+1: thanks!

PR #2283, opened 2019-01-10
kube-prometheus: Added ability to configure nodeExporter port

jonomacd:
We have a conflict on port 9100 so we must change the port which node exporter runs on. I've not used jsonnet before so there is a good change I've done something silly but this seems to wfm.

### DIFF
--- a/contrib/kube-prometheus/README.md
+++ b/contrib/kube-prometheus/README.md
@@ -312,6 +312,10 @@ These are the available fields with their respective default values:
       cpuPerNode: '2m',
       memoryPerNode: '30Mi',
     },
+
+    nodeExporter+:: {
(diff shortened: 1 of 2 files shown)


metalmatze:
Looks just the way I would have done the change! :+1: 
Now that you've updated the jsonnet files, you need to update kube-prometheus' dependencies by running `jb update` and generate all the assets in this repository by running `make clean generate` from the top of this repository.
You can install jb by following this guide: https://github.com/jsonnet-bundler/jsonnet-bundler

brancz:
No need to re-generate as the change doesn't modify anything in the generated manifests checked into this repository.

Thanks! lgtm :+1:

PR #2284, opened 2019-01-10
pkg/prometheus: support min_shards configuration of the queueConfig

crandles:
This is a new config in prometheus v2.6.0


brancz:
Thanks! lgtm :+1: 

As a side note and just for your information, the majority of how remote write works within Prometheus is being changed, as a result flags may be deprecated, but I don't know the final state yet (RE: https://github.com/prometheus/prometheus/pull/4588).

pkg/prometheus/promcfg.go
```diff
@@ -816,6 +816,10 @@ func (cg *configGenerator) generateRemoteWriteConfig(version semver.Version, spe
 				queueConfig = append(queueConfig, yaml.MapItem{Key: "capacity", Value: spec.QueueConfig.Capacity})
 			}
 
+			if spec.QueueConfig.MinShards != int(0) {
```
brancz: this needs a version guard

brancz:
Thanks! lgtm :+1:

PR #2285, opened 2019-01-11
fix node_filesystem only collect mountpoint '/' metrics

liuxu623:


brancz:
This file is generated through jsonnet, you will also need to add that line here: 

https://github.com/coreos/prometheus-operator/blob/8254bb3aafca4df8735e8df644ef99448bb01ba3/contrib/kube-prometheus/jsonnet/kube-prometheus/node-exporter/node-exporter.libsonnet#L88

Andor:
Let's merge it as soon as it will pass all the tests.

brancz:
@liuxu623 I hope you don't mind, I added a commit that does some (confusing) bumps and re-generates the auto-generated parts of this repo. This should make CI pass now :slightly_smiling_face: .

liuxu623:
@brancz Thank you very much🙂.

metalmatze:
Awesome!

s-urbaniak:
good catch indeed, thanks! :+1:

PR #2286, opened 2019-01-11
Update README.md

wleese:
Capitalization and hyphen change. Using the old example results in a broken minikube:

```
✗ minikube status
host: Running
kubelet: Running
apiserver: Stopped
```
https://github.com/kubernetes/minikube/issues/2798


brancz:
Thanks! lgtm :+1:

PR #2287, opened 2019-01-11
contrib/kube-prometheus: Update kube-state-metrics to v1.5.0

metalmatze:
Update to latest kube-state-metrics release


brancz:
needs a generate from the root directory, otherwise lgtm

PR #2292, opened 2019-01-11
Fix for omitting source_labels where they are unnecessary

### LINKED ISSUE #2218: Using `labeldrop` on remoteWrite config breaks Prometheus
**What did you do?**

I supplied a remoteWrite config like so:
```
  remoteWrite:
    - url: "http://influxdb:8086/api/v1/prom/write?db=prometheus"
      writeRelabelConfigs:
        - action: labeldrop
          regex: prometheus
        - action: labeldrop
          regex: prometheus_replica
```

I expected this to configure Prometheus' remote write config so that it would drop the existing `prometheus` and `prometheus_replica` labels from metrics that are scraped from another Prometheus.

Under all circumstances, I got this error from Prometheus:

```
Error loading config couldn't load configuration (--config.file=/etc/prometheus/config_out/prometheus.env.yaml): parsing YAML file /etc/prometheus/config_out/prometheus.env.yaml: labeldrop action requires only 'regex', and no other fields
```

**Environment**

	0.25.0

	1.10.6

* Kubernetes cluster kind: 

	KOPS on AWS

alebsack:
I've tried my hand at the go workflow and applied the fix outlined by @ViniciusSouza in https://github.com/coreos/prometheus-operator/issues/2218#issuecomment-449349056

Fixes #2218

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -750,8 +750,10 @@ func (cg *configGenerator) generateRemoteWriteConfig(version semver.Version, spe
 		if spec.WriteRelabelConfigs != nil {
 			relabelings := []yaml.MapSlice{}
 			for _, c := range spec.WriteRelabelConfigs {
-				relabeling := yaml.MapSlice{
-					{Key: "source_labels", Value: c.SourceLabels},
+				relabeling := yaml.MapSlice{}
+
+				if c.SourceLabels != "" {
+					relabeling = append(relabeling, yaml.MapItem{Key: "source_labels", Value: c.SourceLabels})
 				}
 
 				if c.Separator != "" {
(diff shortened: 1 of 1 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -750,8 +750,10 @@ func (cg *configGenerator) generateRemoteWriteConfig(version semver.Version, spe
 		if spec.WriteRelabelConfigs != nil {
 			relabelings := []yaml.MapSlice{}
 			for _, c := range spec.WriteRelabelConfigs {
-				relabeling := yaml.MapSlice{
-					{Key: "source_labels", Value: c.SourceLabels},
+				relabeling := yaml.MapSlice{}
+
+				if c.SourceLabels != "" {
```
brancz: `.SourceLabels` is a slice of strings, not a single string, so this needs to be:


```suggestion
				if len(c.SourceLabels) > 0 {
```

alebsack:
@brancz github made this too easy. 👍

brancz:
Awesome :+1: lgtm

PR #2296, opened 2019-01-14
Fix node exporter crash looping

BrianChristie:
Seeing this error on all the node_exporter pods after updating kube-prometheus:
`node_exporter: error: unknown long flag '--path.rootfs', try --help`

https://github.com/coreos/prometheus-operator/pull/2285 added `--path.rootfs=/host/root` to the nodeExporter args, however the nodeExporter default version is `v0.16.0` which doesn't support this flag. 

Support for `--path.rootfs` was added in node exporter v0.17.0 https://github.com/prometheus/node_exporter/pull/1058.

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/node-exporter/node-exporter.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/node-exporter/node-exporter.libsonnet
@@ -5,7 +5,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
     namespace: 'default',
 
     versions+:: {
-      nodeExporter: 'v0.16.0',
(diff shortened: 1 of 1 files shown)


metalmatze:
I've confirmed that this fixes current master on my minikube cluster. I'll merge this as a hot fix for `master`. Thanks.

complex64:
Ran into the same issue, thanks for the fix and the quick merge!

Could this somehow be covered by tests?

brancz:
fwiw I think we could have some rudimentary e2e tests for kube-prometheus, that checks that no alerts are firing for example (which should cover all components being up)

metalmatze:
Agreed, something like that is already implemented in our cluster monitoring operator. Maybe we can backport it here too? What do you think @mxinden?

PR #2297, opened 2019-01-14
contrib/kube-prometheus: Add ServiceMonitor for thanos-peers

metalmatze:
As of now we start Thanos as sidecar and provide the store as StatefulSet as well as the querier as Deployment. However we don't collect any metrics from these components.

This PR adds a ServiceMonitor collecting metrics from Services that have `thanos-peers: true` and that then matches pods having `thanos-peers: true` too. Thus, we can scrape sidecars, stores an queriers with one ServiceMonitor. 
Tested on my minikube cluster.

/cc @brancz @squat @s-urbaniak


brancz:
lgtm :+1:

PR #2299, opened 2019-01-15
contrib/kube-prometheus: Add Thanos compactor as StatefulSet

metalmatze:
This will add the [Thanos compactor](https://github.com/improbable-eng/thanos/blob/master/docs/components/compact.md) as a StatefulSet next to the other components. The Thanos compactor exposes metrics on port `10902` but the other ports for gossip etc are not needed.

It would be good to somehow limit the number of replicas this StatefulSet can run to exactly 0 or 1.
Even if the user wants to `kubectl scale statefulset thanos-compactor --replicas 2` it should fail, as the can be at most one compactor running per object storage.
Any ideas on how to achieve this?

/cc @brancz @squat @s-urbaniak @mxinden @bwplotka

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
@@ -139,6 +139,42 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
         volume.fromEmptyDir('data'),
       ]),
 
+    serviceMonitorThanosCompactor:
+      {
+        apiVersion: 'monitoring.coreos.com/v1',
+        kind: 'ServiceMonitor',
+        metadata: {
+          name: 'thanos-compactor',
+          namespace: $._config.namespace,
+          labels: {
+            'k8s-app': 'thanos-compactor',
+          },
+        },
+        spec: {
+          jobLabel: 'k8s-app',
+          endpoints: [
+            {
+              port: 'http',
+              interval: '30s',
+            },
+          ],
+          selector: {
+            matchLabels: {
+              app: 'thanos-compactor',
+            },
+          },
+        },
+      },
+
+    thanosCompactorService:
+      service.new(
+        'thanos-compactor',
+        { app: 'thanos-compactor' },
+        servicePort.newNamed('http', 9090, 'http'),
+      ) +
+      service.mixin.metadata.withNamespace($._config.namespace) +
+      service.mixin.metadata.withLabels({ app: 'thanos-compactor' }),
+
     thanosCompactorStatefulset:
       local statefulSet = k.apps.v1beta2.statefulSet;
       local volume = statefulSet.mixin.spec.template.spec.volumesType;
@@ -146,7 +182,7 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
       local containerEnv = container.envType;
       local containerVolumeMount = container.volumeMountsType;
 
-      local labels = { app: 'thanos', 'thanos-peers': 'true' };
+      local labels = { app: 'thanos-compactor' };
 
       local c =
         container.new('thanos-compactor', $._config.imageRepos.thanos + ':' + $._config.versions.thanos) +
@@ -155,6 +191,7 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
           '--log.level=debug',
           '--data-dir=/var/thanos/store',
           '--objstore.config=$(OBJSTORE_CONFIG)',
+          '--wait',
         ]) +
         container.withEnv([
           containerEnv.fromSecretRef(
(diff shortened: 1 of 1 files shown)


brancz:
I think it's fair to expect a user to understand the stack enough to make this decision themselves (at least as long as we don't fully manage this stack, if we do so, then an/the operator could do this). This generally looks good to me.

Can you just explain a bit more about this being a StatefulSet but having emptydir? Is it so we ensure on rollout the instance is fully terminated before we continue? Is it so storage gets re-attached on rollout?

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
```diff
@@ -138,5 +138,44 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
...
+      local containerEnv = container.envType;
+      local containerVolumeMount = container.volumeMountsType;
+
+      local labels = { app: 'thanos', 'thanos-peers': 'true' };
+
+      local c =
+        container.new('thanos-compactor', $._config.imageRepos.thanos + ':' + $._config.versions.thanos) +
+        container.withArgs([
+          'compact',
+          '--log.level=debug',
```
bwplotka: Are we sure debug is what all user wants? Might be bit spammy?

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
```diff
@@ -138,5 +138,44 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
...
       ]),
+
+    thanosCompactorStatefulset:
+      local statefulSet = k.apps.v1beta2.statefulSet;
+      local volume = statefulSet.mixin.spec.template.spec.volumesType;
+      local container = statefulSet.mixin.spec.template.spec.containersType;
+      local containerEnv = container.envType;
+      local containerVolumeMount = container.volumeMountsType;
+
+      local labels = { app: 'thanos', 'thanos-peers': 'true' };
```
bwplotka: It is not thanos-peer. It does not care about any gossip/dns/fileSD configuration. It does not participate in any Thanos inter communication and I guess that's where thanos-peers label is used?

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
```diff
@@ -138,5 +138,44 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
...
+
+      local c =
+        container.new('thanos-compactor', $._config.imageRepos.thanos + ':' + $._config.versions.thanos) +
+        container.withArgs([
+          'compact',
+          '--log.level=debug',
+          '--data-dir=/var/thanos/store',
+          '--objstore.config=$(OBJSTORE_CONFIG)',
```
bwplotka: I think we need `--wait` as well for continous server-like behaviour?

bwplotka:
Yea, I think blocking user for doing more than one replica would be nice, but hard to achieve.

s-urbaniak:
@metalmatze the only idea I have here to enforce that invariant is to have a sidecar proxying requests to thanos. This proxy would block requests on stateful replicas >0.

bwplotka:
@s-urbaniak we are talknig here about compactor. Again, as mentioned earlier: https://github.com/coreos/prometheus-operator/pull/2299#discussion_r248114178

It does not communicate with Querier, so what exactly we want to block? What proxy do you mean?

s-urbaniak:
@bwplotka gah, you're right, ignore my noise then :+1:

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
metalmatze: You're right, as this was just a hacky way to enable the metrics from that endpoint yesterday. I need to add another ServiceMonitor to match that separately, I guess. Thankfully this is still a WIP PR right now :blush:

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
metalmatze: This makes sense, otherwise it will be crashlooping. Thanks for pointing that out, exactly the feedback I was looking for! :100:

metalmatze:
@brancz 
> Can you just explain a bit more about this being a StatefulSet but having emptydir? Is it so we ensure on rollout the instance is fully terminated before we continue? Is it so storage gets re-attached on rollout?

We should use the same storage class as the Prometheuses in kube-prometheus use. Will update. 
Thanos docs say:
> The compactor needs local disk space to store intermediate data for its processing. Generally, about 100GB are recommended for it to keep working as the compacted time ranges grow over time.
https://github.com/improbable-eng/thanos/blob/master/docs/components/compact.md

bwplotka:
Yea, pvc is the recommendation I think. I guess operator cannot include pvc out of the box?

metalmatze:
PTAL @brancz

contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
brancz: I think I'd prefer to have this be done on an individual basis, and not have the `--log.level` flag at all by default. Feel free to have the removal from all components be a follow up though.

brancz:
Nice work! lgtm :+1:

PR #2314, opened 2019-01-18
contrib/kube-prometheus: Make debugging logs for Thanos configurable

metalmatze:
This will move the `--log.level=debug` argument behind a config boolean which is set to false by default.


brancz:
hmm .. I personally think this is probably best solved by just having a mixin that can be added on top, instead of being a configuration

PR #2318, opened 2019-01-21
pkg/prometheus: guard against nil targetPort

squat:
This commit guards against a nil pointer dereference caused when an
Endpoint's port is 0 and the targetPort is nil.

Fixes: https://github.com/coreos/prometheus-operator/issues/2156

cc @brancz @metalmatze


brancz:
lgtm on green :+1:

PR #2319, opened 2019-01-21
jsonnet/prometheus-operator: delete services 

squat:
This commit adds the delete verb to the services rule of the prometheus
operator role. It also gives the prometheus operator the power to get,
delete, create, and update the finalizers subresource of services.

fixes: https://github.com/coreos/prometheus-operator/issues/2279

cc @brancz @metalmatze

### DIFF
--- a/Documentation/rbac.md
+++ b/Documentation/rbac.md
@@ -61,11 +61,13 @@ rules:
   - ""
   resources:
   - services
+  - services/finalizers
   - endpoints
   verbs:
   - get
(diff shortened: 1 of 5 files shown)


squat:
hmmm the only place I know it's written down is the issue comment https://github.com/coreos/prometheus-operator/issues/2279#issuecomment-453115851

metalmatze:
@squat, that will do. Thanks!

PR #2321, opened 2019-01-21
Add README badge for Slack channel

metalmatze:
As we now have a dedicated channel on Kubernetes Slack, why not link to it from the README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Prometheus Operator
 [![Build Status](https://travis-ci.org/coreos/prometheus-operator.svg?branch=master)](https://travis-ci.org/coreos/prometheus-operator)
 [![Go Report Card](https://goreportcard.com/badge/coreos/prometheus-operator "Go Report Card")](https://goreportcard.com/report/coreos/prometheus-operator)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,7 @@
 # Prometheus Operator
 [![Build Status](https://travis-ci.org/coreos/prometheus-operator.svg?branch=master)](https://travis-ci.org/coreos/prometheus-operator)
 [![Go Report Card](https://goreportcard.com/badge/coreos/prometheus-operator "Go Report Card")](https://goreportcard.com/report/coreos/prometheus-operator)
+[![Slack](https://img.shields.io/badge/join%20slack-%23prometheus--operator-brightgreen.svg)](http://slack.k8s.io/)
```
squat: should the link be https? Also, the link is to the general k8s slack, can we somehow link directly to the right channel?

squat:
ah ok well, if it's not possible too bad :/ but i don't think it's worth investing much time in

metalmatze:
I don't see a way to forward people directly into the channel we want. :confused:

PR #2322, opened 2019-01-21
contrib/kube-prometheus: Update kubernetes mixin for fixed CPUThrottlingHigh

metalmatze:
This fixed the CPUThrottlingHigh which would trigger two alerts for just one container in some cases:
See https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/137

/cc @squat @brancz @mxinden


brancz:
lgtm on green :+1:

PR #2323, opened 2019-01-21
kube-prometheus: Remove enableAdminAPI configuration option

mxinden:
Manifest generation fails if this option is not set. Instead, remove the
option and require users to patch the object in case they need it.

See https://github.com/coreos/prometheus-operator/pull/2300#issuecomment-456113717 for more details.


mxinden:
Guess regeneration is not necessary. Will do that in a follow up PR anyways, bumping Alertmanager to v0.16.0.

PR #2325, opened 2019-01-22
 kube-prometheus: Update Alertmanager to v0.16.0 

mxinden:


squat:
hmm does this need a regeneration?

brancz:
it needs a regenerate from the root of the repo

squat:
There were two failures here:
```
            --- FAIL: TestAllNS/x/AMMeshInitialization/amVersion (316.06s)
                alertmanager_test.go:240: timed out waiting for the condition
        --- FAIL: TestAllNS/x/AMMeshAfterRollingUpdate (346.06s)
            alertmanager_test.go:285: timed out waiting for the condition
```

mxinden:
Both `AMMeshInitialization` and `AMMeshAfterRollingUpdate` could use a `wait`, giving Alertmanager a bit more time to form the mesh. In case we see this issue more often I can follow up with a patch for the two tests.

PR #2327, opened 2019-01-23
Use kubectl apply instead of create

chaosaffe:
Resources can be created _and_ updated with `apply` without receiving errors


brancz:
If I recall correct replaced it with apply because some of our objects were too large to fit into the "last applied" annotation. Could you try applying multiple times and see if this is still the case? If not I'm happy to change this back (I believe it used to be the Grafana dashboards that used to be in one giant ConfigMap).

chaosaffe:
I don't experience the mentioned behaviour on _v1.13.2_ , but it may still be the case  for older clusters.

brancz:
That's validation enough for me :)

In that case can you remove the `|| true` and `2>/dev/null || true`, as those were only added to safely run `create` multiple times without a non 0 exit code.

brancz:
Thanks!

PR #2331, opened 2019-01-23
Update kube-rbac-proxy and limit the ciphers being used for h2

metalmatze:
This updates kube-rbac-proxy to v0.4.1 and sets it to log output to stderr. We also limit the number of ciphers even more to be h2 compatible.

Depends on kube-rbac-proxy v0.4.1 being releasd. See https://github.com/brancz/kube-rbac-proxy/pull/31

/cc @brancz @squat @mxinden


contrib/kube-prometheus/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -145,6 +146,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
       local proxySelfMetrics =
         container.new('kube-rbac-proxy-self', $._config.imageRepos.kubeRbacProxy + ':' + $._config.versions.kubeRbacProxy) +
         container.withArgs([
+          '--logtostderr',
```
mxinden: For https://github.com/brancz/kube-rbac-proxy/ would it make sense to add this flag as a default to the Dockerfile?

metalmatze: Good idea, that should work. Is it a common thing to do when using glog? Not sure if in a future release we would get rid of glog entirely?

brancz: I'm not a huge fan of adding flags into entrypoints, it gives the notion of this being the default for the binary when it's not ... we made this mistake in kube-state-metrics and are stuck with it

PR #2334, opened 2019-01-24
Change prometheus service to headless service

benjaminhuo:
Signed-off-by: Benjamin <benjamin@yunify.com>
Statefulset should use headless service


metalmatze:
Can you elaborate a bit more on the reasoning behind that change? Just as a heads up - I'm not sure we actually want this. :wink:

benjaminhuo:
@metalmatze finally We found this PR is not necessary because the prometheus-operated service is what we want :)

PR #2335, opened 2019-01-24
kube-prometheus: Add e2e tests

mxinden:


contrib/kube-prometheus/tests/e2e/main_test.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func TestQueryPrometheus(t *testing.T) {
+	t.Parallel()
+	queries := []struct {
+		query   string
+		expectN int
+	}{
+		{
+			// 	query:   `up{job="node-exporter"} == 1`,
```
mxinden: Prometheus fails to scrape these on my local minikube. Still need to dig further. Not sure we should block on that though.

metalmatze: Have you made sure to run minikube with the additional flags? https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus#minikube

brancz:
Let's start with this and add more as follow ups.

PR #2338, opened 2019-01-24
*: cut v0.28.0

metalmatze:
Still WIP as we're waiting for https://github.com/coreos/prometheus-operator/pull/2331.

I will regenerate all things, when we have agreed on the CHANGELOG.md and we're ready to merge.

/cc @brancz @mxinden @squat


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 0.28.0 / 2019-01-24
+
+* [FEATURE] CLI tool to lint YAML against CRD defintions (#2269)
```
mxinden: Not sure this should be part of the Prometheus Operator changelog, as it seems to be a tool on the side?

metalmatze: Yep, it's a separate tool, however as it is directly related to the CRDs and to validating those, therefore I feel like it's a valid addition to the changelog. No hard feelings though.

mxinden: I am fine with either way.

metalmatze:
We need to add https://github.com/coreos/prometheus-operator/pull/2307 now.

brancz:
Are we waiting on anything here? If not let's remove the `[WIP]` and merge :slightly_smiling_face: 

lgtm :+1:

metalmatze:
Sorry, I restarted the tests and then went for lunch hoping it would be green when coming back. Which it is. :+1:

PR #2340, opened 2019-01-25
test: Return pollError when waiting for Alertmanager mesh

metalmatze:
We want to know a bit more about the internal error that occurred while waiting for the Alertmanager mesh to get ready.

While working on this change, I also saw several calls to `Core()` which is deprecated and updated these.

/cc @mxinden

### DIFF
--- a/test/framework/alertmanager.go
+++ b/test/framework/alertmanager.go
@@ -189,7 +189,8 @@ func amImage(version string) string {
 }
 
 func (f *Framework) WaitForAlertmanagerInitializedMesh(ns, name string, amountPeers int) error {
-	return wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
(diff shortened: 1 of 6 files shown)


test/framework/alertmanager.go
```diff
@@ -337,7 +345,7 @@ func (f *Framework) WaitForAlertmanagerConfigToContainString(ns, amName, expecte
 	})
 
 	if err != nil {
-		return fmt.Errorf("failed to wait for alertmanager config to contain %q: %v: %v", expectedString, err, pollError)
+		return fmt.Errorf("failed to wait for alertmanager config to contain %q: %v", expectedString, err)
```
metalmatze: I don't think we need the pollError here as we already have `expectedString` inside the returned error.

PR #2345, opened 2019-01-25
pass prometheus route prefix to thanos sidecar

### LINKED ISSUE #2344: Thanos sidecar does not get configured properly when using routePrefix
**What did you do?**
Deployed prometheus with a routePrefix option

Thanos query sidecar get updated with the proper prometheus-url argument.

Thanos sidecar did not respect the prometheus route prefix and was thus unable to function as expected

ephur:
Fixes #2344

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -21,7 +21,7 @@ import (
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1beta2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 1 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -658,7 +658,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 
 		thanosArgs := []string{
 			"sidecar",
-			fmt.Sprintf("--prometheus.url=http://%s:9090", c.LocalHost),
+			fmt.Sprintf("--prometheus.url=http://%s:9090%s", c.LocalHost, webRoutePrefix),
```
metalmatze: I suggest to handle `webRoutePrefix` similar to other parts in this function, where we use `path.Clean`.
```suggestion
			fmt.Sprintf("--prometheus.url=http://%s:9090%s", c.LocalHost, path.Clean(webRoutePrefix)),
```

https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L429

pkg/prometheus/statefulset.go
```diff
@@ -21,7 +21,7 @@ import (
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1beta2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
metalmatze:
```suggestion
	"k8s.io/api/core/v1"
```

ephur: Not sure why vscode keeps trying to make this change when go fmt does not, reverted!

metalmatze:
Thanks!

PR #2347, opened 2019-01-28
extend configuration for Grafana config file

widgetpl:
I have found that current approach with `hostPath` as volume mount for grafana configuration file is not enough so I decided to extend it with more options like `secret` and `configmap`.
Please review my changes as I don't have any experience with writing helm's charts and go templating.

### DIFF
--- a/helm/grafana/values.yaml
+++ b/helm/grafana/values.yaml
@@ -38,8 +38,18 @@ global:
 extraVars:
 
 ## Change to true override Grafana's default config.
+## There are 3 possibilities for config type: hostpath, configmap or secret
 ## Make sure grafana.ini is present on /etc/grafana
-mountGrafanaConfig: false
+mountGrafanaConfig:
+  enabled: false
+  # type: hostpath
+
+  # hostPath:
+  #   path: /etc/grafana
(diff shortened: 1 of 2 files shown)


helm/grafana/values.yaml
```diff
@@ -38,8 +38,18 @@ global:
...
-mountGrafanaConfig: false
+mountGrafanaConfig:
+  enabled: false
+  # type: hostpath
+
+  # hostPath:
+  #   path: /etc/grafana
+  #   type: Directory
+
+  # secretName: "configSecretName"
```
widgetpl: not sure if this one should look like
```
secretName:
  name: "configSecretName"
```

helm/grafana/values.yaml
```diff
@@ -38,8 +38,18 @@ global:
...
+  # configmapName: "configConfigmapName"
```
widgetpl: and this one:
```
configmapName:
  name: "configConfigmapName"
```

metalmatze:
Yes, the helm chart has been migrated to https://github.com/helm/charts/tree/master/stable/prometheus-operator

PR #2349, opened 2019-01-28
 contrib/kube-prometheus: remove node role

squat:
This commit removes get/list/watch on nodes for the Prometheus-k8s
instance, as Prometheus pods do not need that privilege for anything.

cc @brancz @metalmatze @mxinden


mxinden:
>  Wondering how that got in there grin Thanks for the fix.

@metalmatze maybe we have been using it to discover the kubelet?

metalmatze:
Good point, @mxinden. I just removed nodes (like this PR does) from the roles in my cluster and try to find out if scraping and discovering nodes still works.

metalmatze:
It seems that discovering nodes is done via ClusterRole and it still works on my cluster without these rules. I would say that we can go ahead with this! :+1:

mxinden:
@squat debugging that will hopefully improve once https://github.com/coreos/prometheus-operator/pull/2341 merged.

PR #2350, opened 2019-01-28
Aggregate prometheus-crd-view role to admin

aditya-konarde:
In case the `view` is not aggregated to `admin`, you would need to grant a user both `admin` and `view` roles on the same namespace to be able to edit and get the resources respectively. This means that an `admin` on the namespace would be able to `create` the resource object, but not `get` it. 

This change aggregates the view to the admin role, so that the admin can now `get` on the resource as well.

### DIFF
--- a/example/rbac/prometheus-operator-crd/prometheus-operator-crd-cluster-roles.yaml
+++ b/example/rbac/prometheus-operator-crd/prometheus-operator-crd-cluster-roles.yaml
@@ -4,6 +4,7 @@ metadata:
   name: prometheus-crd-view
   labels:
     rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
 rules:
 - apiGroups: ["monitoring.coreos.com"]
   resources: ["alertmanagers", "prometheuses", "prometheusrules", "servicemonitors"]
--- a/example/rbac/prometheus-operator-crd/prometheus-operator-crd-cluster-roles.yaml
+++ b/example/rbac/prometheus-operator-crd/prometheus-operator-crd-cluster-roles.yaml
@@ -3,8 +3,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: prometheus-crd-view
   labels:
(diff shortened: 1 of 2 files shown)


aditya-konarde:
Also need to update the docs here: https://github.com/coreos/prometheus-operator/blob/master/Documentation/rbac-crd.md

aditya-konarde:
CC @squat @jfchevrette

example/rbac/prometheus-operator-crd/prometheus-operator-crd-cluster-roles.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: prometheus-crd-view
   labels:
     rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```
aweiteka: Also add `rbac.authorization.k8s.io/aggregate-to-edit: "true"`

aditya-konarde:
@aweiteka updated. 

Also thanks @metalmatze for the pointers :)

aweiteka:
These are docs/examples. How do these land in an openshift cluster? Are they defined in cluster-monitoring-operator? In other words, what else do we need to change to see these land in an openshift cluster?

metalmatze:
I'm not entirely sure myself. @brancz should know more.
For now I'm going to merge this change, as it looks good to me from the code side of things.
Nonetheless, feel free to continue the discussion afterwards.

brancz:
This does not get created by cluster-monitoring-operator today at all, as this so far has been a concern for OLM to create on behalf of users when installing a prometheus-operator. I'm not opposed to having it created by cluster-monitoring, but we would need to make sure there are no conflicts with OLM.

jfchevrette:
To my understanding and in my specific case there is no cluster-monitoring-operator involved. We're instantiating prometheus-operator from OLM (0.8.x) and the CR roles are not aggregated/propagated to the view/edit/admin cluster roles properly. I believe this PR fixes that.

From the looks of it we will need to wait for a new release to be cut and then update or add this release in an OLM catalog so that we can deploy it.

Does that sound right?

brancz:
That was exactly my thinking, yes.

PR #2352, opened 2019-01-28
kube-prometheus: ability to configure annotations on prometheus service

philpearl:
We're configuring an ingress in GCP, and to do some advanced things we need to add annotations to the backing services.

I'm an absolute ksonnet newbie, so apologies in advance if I'd done something silly.


brancz:
The principle here is pretty much the same as I commented here: https://github.com/coreos/prometheus-operator/pull/2351#issuecomment-460382527

PR #2371, opened 2019-02-04
Bump config reloader resources

kaarolch:
Hi, 
I still observe that alert manager trigger high CPU through for three configurator pods:
* `config-reloader` - alertmanager
* `prometheus-config-reloader` and `rules-configmap-reloader` - prometheus. 

Maybe instead bum those limits every time we can expose them as argument?

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -462,8 +462,8 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 						},
 						Resources: v1.ResourceRequirements{
 							Limits: v1.ResourceList{
-								v1.ResourceCPU:    resource.MustParse("50m"),
-								v1.ResourceMemory: resource.MustParse("10Mi"),
+								v1.ResourceCPU:    resource.MustParse("100m"),
+								v1.ResourceMemory: resource.MustParse("15Mi"),
 							},
 						},
 					},
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -617,8 +617,8 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 			VolumeMounts: []v1.VolumeMount{},
 			Resources: v1.ResourceRequirements{
 				Limits: v1.ResourceList{
-					v1.ResourceCPU:    resource.MustParse("25m"),
-					v1.ResourceMemory: resource.MustParse("10Mi"),
+					v1.ResourceCPU:    resource.MustParse("50m"),
+					v1.ResourceMemory: resource.MustParse("15Mi"),
 				},
 			},
 		}
@@ -856,7 +856,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 						VolumeMounts: configReloadVolumeMounts,
 						Resources: v1.ResourceRequirements{
 							Limits: v1.ResourceList{
-								v1.ResourceCPU:    resource.MustParse("50m"),
+								v1.ResourceCPU:    resource.MustParse("100m"),
 								v1.ResourceMemory: resource.MustParse("50Mi"),
 							},
 						},


brancz:
I believe @squat was just about to look into this, I was wondering if we should maybe make this a flag on the prometheus operator process, that could have these defaults, but make it configurable and allow it to be disabled by setting it to 0. What do you think?

kaarolch:
@brancz options that you mentioned sound great. If no one is currently working on it, I can try to add this feature.

brancz:
@metalmatze what do you think of the above suggestion?

metalmatze:
I'm fine with adding this as flags. Please go ahead and update this PR or create a new one. :+1:

RaynDol:
HI,

we encounter this issue lately in K8S using the prom operator , any suggestion how to resolve this ?
it was stable in the last 4 month and in the last week we got the following issue, i think its releted,
isn't it ? 
OCI runtime create failed: container_linux.go:348: starting container process caused "process_linux.go:402: container init caused "process_linux.go:367: setting cgroup config for procHooks process caused \"failed to write 10485760 to memory.limit_in_bytes: write /sys/fs/cgroup/memory/kubepods/burstable/pod972cf87b-3192-11e9-bae2-

brancz:
We've merged https://github.com/coreos/prometheus-operator/pull/2403 in which case this is obsolete/covered by that PR. Closing here, but thanks a lot for taking the effort to prepare the PR!

RaynDol:
@brancz  -  is this issue related to my mine ?  if yes when we can expect new release ? Thanks!

brancz:
It is related to it, yes. I think https://github.com/coreos/prometheus-operator/issues/2409 describes pretty well what you are seeing.

RaynDol:
@brancz  - Thanks for the quick response, our deployment look like 
,,,
....
      - args:
        - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.26.0

```

Should I change something in the above config  order to resolve it ?  other thing ? 

Thank you!

brancz:
We're going to have to do a new release to solve this for you. Sorry for the inconvenience. We will do our best to get a release out with this fix, we understand it's an issue.

RaynDol:
@brancz Thanks a lot!  
do you know perhaps when it will be available to use ?

brancz:
I can't promise a date, but we are working on it, typically it doesn't take too long, but we maintain many other projects so things sometimes get in the way :slightly_smiling_face: .

RaynDol:
@brancz  - Thank you very much, looking forward to get it.  :) 
meanwhile (since we are blocked) if there is some workaround it will be great if you can provide reference

brancz:
We build `master` images that are generally not recommended for use, but could be a mean for you to try out the state on master, and use them until we actually do a proper versioned release :slightly_smiling_face: . Keep in mind that there is 0 compatibility between master and a release.

brancz:
(sorry misclicked)

RaynDol:
OK thank you !

tjholwik:
HI,

We are running into the same issue in Kubernetes , @brancz  - can you please why this happen? AFAIK we didn't change any Prometheus version/ config lately ...and we want to avoid such of issue in the future. 

BR,
T.J

brancz:
@tjholwik this particular issue got much worse with the recent runc CVE mitigation, we're releasing better defaults and the option to configure this.

tjholwik:
@brancz  - it's related to this ? 
https://www.twistlock.com/2019/02/11/how-to-mitigate-cve-2019-5736-in-runc-and-docker/ 

my question is how does it happen if we didn't  change anything ?  any hint will be appreciated  or maybe how to avoid it in the future. since currently this is blocking us...

Cheers,
tj

metalmatze:
runc now copies the binaries when starting and that creates an memory overhead that we see now.
https://github.com/opencontainers/runc/commit/0a8e4117e7f715d5fbeef398405813ce8e88558b
>  There are quite a few circumstances where /proc/self/exe pointing to a
pretty important container binary is a _bad_ thing, so to avoid this we
have to make a copy (preferably doing self-clean-up and not being
writeable).

Not sure if anyone can avoid something like that. :confused:

tjholwik:
@metalmatze  - we really appreciate if you could provide a new release, the Prometheus and the alert manager are not working well, if there is a workaround please advice.

PR #2381, opened 2019-02-05
Update grafana

brancz:
@metalmatze @squat @s-urbaniak @mxinden


metalmatze:
Do you want to help beta testing Grafana? Just asking why we should move to a beta release. Happy to do it though. Not sure if it is good as default for kube-prometheus.

brancz:
We've regularly done this with Prometheus and Alertmanager as well, plus I would like to make it easier to try out the query browser and logging interfaces.

PR #2384, opened 2019-02-05
Fix links

Arslanbekov:
Fix links to official Prometheus documentation


brancz:
there are some generated files that still need updating, just run `make clean generate-in-docker` in the root of the repo

brancz:
Awesome! Thanks!

PR #2387, opened 2019-02-06
Reduce metrics

brancz:
As described in https://github.com/google/cadvisor/issues/1925, the container network metrics are disabled and therefore cause more confusion than gain as they are always set to 0. This PR removes those at ingestion time, and also drops unnecessary high cardinality metrics from the Kubernetes API.

@metalmatze @mxinden @squat @s-urbaniak


s-urbaniak:
LGTM :ok_hand:

PR #2411, opened 2019-02-15
prometheus: make replicas external label name configurable

forkbomber:
This change enables existing Thanos deployments which adopted any label name other than `prometheus_replica` to migrate to Prometheus Operator without losing of Thanos deduplication features.

Defaults to `prometheus_replica` for backward compatibility.


brancz:
Nice, thanks! lgtm :+1:

PR #2416, opened 2019-02-18
[helm] Remove

vsliouniaev:
As per the suggestion here: https://github.com/coreos/prometheus-operator/pull/2242#issuecomment-451988560 removing Helm from this repository.

Added some migration documentation. While there is no direct path to migrate, added the most interesting aspect of migrating data.

This should go some way to satisfying
https://github.com/helm/charts/issues/9108


vsliouniaev:
As there are no changes to anything but helm, the build failure should probably be ignored. The fix for the overall build failures are included in https://github.com/coreos/prometheus-operator/pull/2415

s-urbaniak:
@vsliouniaev this seems indeed the way forward. Do you mind to remove the helm references from `.travis.yml`?

https://github.com/coreos/prometheus-operator/blob/d177599bac58db58fc4fdd0761e1e11861b4cc8f/.travis.yml#L24

vsliouniaev:
Thanks @s-urbaniak I think I got everything

helm/README.md
```diff
@@ -1,15 +1,79 @@
...
+It is still possible to run multiple prometheus instances on a single cluster - you will need to disable the parts of the chart you do not wish to deploy.
+
+Issues and pull requests should be tracked using the [helm/charts](https://github.com/helm/charts) repository.
 
 You can check out the tickets for this change [here](https://github.com/coreos/prometheus-operator/issues/592) and [here](https://github.com/helm/charts/pull/6765)
 
----
+## Changes
+The chart has 3 dependencies, that can be seen in the chart's requirements file:
+https://github.com/helm/charts/blob/master/stable/prometheus-operator/requirements.yaml
```
s-urbaniak: I understand we want to document the upstream helm chart changes but I am a bit worried this will skew over time. Can we maybe point the upstream chart's README instead?

vsliouniaev: Thanks for the suggestion. I've moved the migration section to that chart and updated the link

brancz:
Thanks for taking care of this! lgtm :+1: 

As a side note, it would be very helpful to us if the upstream chart was named kube-prometheus instead of prometheus-operator, as the prometheus-operator itself is just about managing Prometheus servers, whereas kube-prometheus is the whole monitoring stack. In fact we are planning to extract kube-prometheus into its own repository and then the confusion would be even bigger. This change would be much appreciated!

vsliouniaev:
I am concerned this will create further confusion for everyone already using the stable/prometheus-operator chart. There is also a stable/prometheus already.

I am not certain as to the process for doing this, since removing a chart from stable seems like it is going directly against what the stable designation is intended to behave like.

In my opinion the right thing to do is create documentation with the various install options / related projects.

PR #2418, opened 2019-02-19
*: Cut 0.29.0

metalmatze:
People are asking for a new release with the configurable CPU limits.

We should probably merged #2415 before doing the release. I just wanted to prepare it already.

/cc @brancz @mxinden @s-urbaniak @squat


s-urbaniak:
@metalmatze this looks good, but it seems examples need to be bumped too.

metalmatze:
Thanks for the hint. I ran `make generate` and it didn't generate everything... Running `make generate-in-docker` now.

PR #2421, opened 2019-02-20
Upgrade prometheus-adapter APIService to v1

mgoodness:
`APIRegistration` went v1 in Kubernetes v1.10. Upgrading seems safe.

Also:

- remove unnecessary namespace from prometheus-adapter ClusterRoleBinding
- update Grafana dashboardDefinitions (side effect of `jb update`)

Signed-off-by: Michael Goodness <mike.goodness@ticketmaster.com>

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
@@ -50,7 +50,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
   prometheusAdapter+:: {
     apiService:
       {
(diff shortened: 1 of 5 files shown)


brancz:
Generally this sounds and looks good to me.

I think some jsonnet deps were outdated, can you just run `make clean generate-in-docker` from the root of the repo once more? Thanks!

mgoodness:
Dunno what's going on with the build, but I'm happy to help if I can.

brancz:
My apologies, this is an artifact of us having a dependency on our own repo. You need to run `make clean generate-in-docker` and then commit all of that into (yet another) new commit. This is because the jsonnet dependency then points at the "last" commit, so amending or rebasing gives the commit a new commit id.

mgoodness:
Not your fault; I'd previously appended, not realizing the self-dependency. I'll do things the right way going forward. Have a few more kustomize-related things in mind.

brancz:
This lgtm :+1: thanks!

PR #2427, opened 2019-02-22
Operator labels

mgoodness:
This adds a common `app.kubernetes.io/name: prometheus-operator` label to all resources, as [recommended](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) by SIG-Apps.

It also replaces the `k8s-app` label used in the Deployment and Service. That's a potentially disruptive change, since a Deployment's `spec.selector` is immutable. If prometheus-operator is already deployed, it will have to be deleted first.

Not sure how to get new `0prometheus-operator` manifests generated for kube-prometheus.

### DIFF
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -5,7 +5,11 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
     namespace: 'default',
 
     prometheusOperator+:: {
-      labels: { 'apps.kubernetes.io/name': 'prometheus-operator' },
+      labels: {
+        'apps.kubernetes.io/name': 'prometheus-operator',
+        'apps.kubernetes.io/component': 'controller',
+        'apps.kubernetes.io/version': $._config.versions.prometheusOperator,
+      },
     },
 
     versions+:: {
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -2,8 +2,11 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
 
 {
   _config+:: {
-    labels: { 'apps.kubernetes.io/name': 'prometheus-operator' },
     namespace: 'default',
+
+    prometheusOperator+:: {
+      labels: { 'apps.kubernetes.io/name': 'prometheus-operator' },
+    },
 
     versions+:: {
       prometheusOperator: 'v0.29.0',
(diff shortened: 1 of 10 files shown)


jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -2,6 +2,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
 
 {
   _config+:: {
+    labels: { 'apps.kubernetes.io/name': 'prometheus-operator' },
```
brancz: this should be nested in a `prometheusOperator` object

brancz:
I think I'm generally ok with this, as users can revert to the "current" way by setting the config.

ant31:
@mgoodness I would add all the rec labels, at least: [name, part-of, instance]
fixes: https://github.com/coreos/prometheus-operator/issues/1441

```
app.kubernetes.io/name: prometheus-operator
app.kubernetes.io/part-of: kube-prometheus
app.kubernetes.io/component: "controller"
app.kubernetes.io/version: $version
app.kubernetes.io/instance: '1'
app.kubernetes.io/managed-by: 'prometheus-operator'
```


Then for other components deployed by . the kube-prometheus stack, like AlertManager:
alertmanager:
```
app.kubernetes.io/name: alertmanager
app.kubernetes.io/part-of: kube-prometheus
app.kubernetes.io/component: "backend"
app.kubernetes.io/version: $alertmanager-version
app.kubernetes.io/instance: '1'
app.kubernetes.io/managed-by: 'helm'  # 'kubectl' ...
```

grafana:
```
app.kubernetes.io/name: grafana
app.kubernetes.io/part-of: kube-prometheus
app.kubernetes.io/component: "front-end"
app.kubernetes.io/version: $grafana-version
app.kubernetes.io/instance: '1'
app.kubernetes.io/managed-by: 'helm?'  
```


UI/Dashboards will soon expose all the labels created this way and also it makes it easier to use with the [ApplicationCRD](https://github.com/kubernetes-sigs/application) 
```
# get all resources related to the monitoring stack:
kubectl get all -l app.kubernetes.io/part-of=kube-prometheus 
```

jsonnet/prometheus-operator/prometheus-operator.libsonnet
mgoodness: 👍

mgoodness:
I'd intended to save the other components for a separate PR, but will add them here.

ant31:
@mgoodness 
my comment was not to about adding labels to all other components in this PR, only to add the extra labels
```
app.kubernetes.io/name: 
app.kubernetes.io/part-of: 
app.kubernetes.io/component:
app.kubernetes.io/version: 
app.kubernetes.io/instance:
app.kubernetes.io/managed-by: 
```

mgoodness:
@ant31 
Understood. I'll add the other labels to prometheus-operator here and stick with my original plan for a second, kube-prometheus PR.

Now, about those labels...  I think `app.kubernetes.io/instance:` and `app.kubernetes.io/managed-by:` want to be set by whatever installs them (Helm, Argo CD, Weave Flux, etc.). Should I default them to something, or leave them out and trust that the installing tool will do its job? [Argo CD](https://github.com/argoproj/argo-cd), for example, adds/set them on each resource. FWIW, they're also quite keen on jsonnet.

brancz:
This lgtm :+1: thanks for the work!

Giving @ant31 and @mxinden the last call.

brancz:
Thanks!

abursavich:
I got here from the v0.30.0 release notes and noticed that all of the labels are using `apps.kubernetes.io`, but the recommended labels use `app.kubernetes.io` (singular app).

brancz:
That makes you’re absolutely right. Would you like to fix this?

abursavich:
Sure, I'll send a PR.

PR #2429, opened 2019-02-25
node-exporter tolerate any taint

brancz:
Node-exporter should be running on any linux node, no exception.

@mxinden @s-urbaniak 

cc @sichvoge

### DIFF
--- a/Documentation/user-guides/cluster-monitoring.md
+++ b/Documentation/user-guides/cluster-monitoring.md
@@ -221,8 +221,10 @@ spec:
         runAsUser: 65534
       serviceAccountName: node-exporter
       tolerations:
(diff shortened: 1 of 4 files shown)


brancz:
> Btw. how to we prevent node exporter being scheduled on a windows or arm machine?

We already have a [nodeSelector for linux](https://github.com/coreos/prometheus-operator/blob/e381c49b5d25a802bfcde9fb36333587987d06e3/contrib/kube-prometheus/manifests/node-exporter-daemonset.yaml#L71), as node-exporter technically builds on other platforms just fine, I would like to hold off on specifying the architecture here, and let multi-arch images become a thing.

PR #2430, opened 2019-02-25
Add prometheusExternalLabelName field to Prometheus object

forkbomber:
This commit adds prometheusExternalLabelName field to Prometheus object.

The string value of this field is used to customize the name of Prometheus external label holding namespace and name of Prometheus object.

When the field is set to empty string (`""`) it would not be included into a rendered external_labels dictionary of prometheus.yaml.

When the field is missing (`null`) it defaults to `prometheus` for backward compatibility.


forkbomber:
I'd like to bring up a concern related to rendering of Prometheus external labels from the perspective of interface consistency and complexity:

While this PR on its own a must-have for pre-existing Thanos-enabled HA deployments to have a smooth transition to Prometheus Operator (due to the fact that Thanos HA capabilities are sensitive to any discrepancy with global labels among replicas) it introduces discontinuity with previously introduced field `replicaExternalLabelName` (controlling the name of the key to holding a specific StatefulSet replica name in Prometheus external labels) (#2411).

I can think of 3 options to approach this discontinuity:

1. Leave as-is: both fields sound similar and are designed for similar use-cases, but for stated use-case, only one of them need to have a capability to be skipped/excluded entirely (`prometheus` which this PR is addressing). 

1. Update `replicaExternalLabelName` field handling, where the empty string would be interpreted as a request _not_ to add "replica" global label, thus bringing up behavior up to `prometheus` global label (which this PR is proposing). This would be a breaking change (as of now empty string in `replicaExternalLabelName` is interpreted as `prometheus_replica`)

1. move both "replica" and "prometheus" global labels into a default value of externalLabels map:

    ```
    externalLabels:
        prometheus: $POD_NAMESPACE/$STATEFULSET_NAME
        replica: $POD_NAME
    ```

    These environment variables would be exposed to a prometheus-config-reloader container via 
    Downward API (directly via `metadata` ref or via pod labels which already contain these values) or  
    just set statically set by the operator (directly or via annotation).

In my opinion, option 3 would result in the best UX, but is breaking change; option 2 would be a minor improvement to UX and a less breaking change; while option 1 is the most pragmatic minimum viable fix to the problem at hand with no breakage and least amount of code.

brancz:
Sorry for taking so long. I think 2 is the best and most consistent option here.

forkbomber:
Thank you for taking time to review the PR.
Would it make sense to handle a PR implementing option 2 separately from this one?

Due to a minor breaking change we might want to release them separately targeting different releases, this PR might go into a next _patch_, while change to `replicaExternalLabelName` would be better suited for the next _minor_ release to signify possible breakage.

brancz:
I don't see it as a breaking change, it's an extension of the functionality of an existing field. Values that were set previously still work, only empty strings change their meaning which didn't make sense to be set previously anyways, so I'm ok with adding it in this PR.

forkbomber:
Added implementation of option #2 via a [separate commit](2430/commits/410721d5ff5f5b7f6b4945288f23070f5efbec8f).

brancz:
lgtm :+1: 

leaving last call up to @mxinden @squat @s-urbaniak @metalmatze

PR #2431, opened 2019-02-25
Fix operator selector

mgoodness:
#2427 introduced a bug to the prometheus-operator Deployment. Including the version in the selector will result in an orphaned ReplicaSet when the version is changed. This separates labels into two objects, leaving `apps.kubernetes.io/version` out of the one used by the selector.


brancz:
lgtm :+1: thanks for the quick follow up!

PR #2436, opened 2019-02-26
configurable thanos replicas and resource-quotas

Hashfyre:
This makes adds two sections in `config`:
 - replicas:
     - thanosQuery: p
     - thanosCompact: q
     - thanosStore: r
 - resources:
     - thanosQuery: { requests: {cpu: x, memory: y}, limits: {cpu: X, memory: Y}}
     - thanosCompact: {...}
     - thanosStore: {...}

based on which:
 - replicas for each component are configured:
    - using: `local replicas = $._config.replicas.<componentName>;`
    - passing to `deployment.new('<label>', replicas, ...)`
 - resource quota for each component are configured:
   - `container.mixin.resources.withRequests()`
   - `container.mixin.resources.withLimits()`

Adds documentation at [Documentation/thanos.md](Documentation/thanos.md)

### DIFF
--- a/Documentation/thanos.md
+++ b/Documentation/thanos.md
@@ -51,6 +51,41 @@ Let's assume you saved this file to `/tmp/thanos-config.yaml`. You can use the f
 kubectl -n monitoring create secret generic thanos-objstore-config --from-file=thanos.yaml=/tmp/thanos-config.yaml
 ```
 
+### Configuring Thanos Replicas
+You can use the config block to run custom number of replicas of each Thanos component
+
+```
+{
+  _config+:: {
+    replicas+:: {
+      thanosQuery: 1,
+      thanosCompact: if thanosFederationRole =="cluster" then 1 else 0,
+      thanosStore: 1,
+    },
+    ...
+```
+
+### Configuring Thanos Resource Quotas
+You can use the config block to run Thanos components with custom resource quotas
+
+```
+{
+  _config+:: {
+    resources+:: {
+      thanosQuery+:: {
+        requests: {
+          cpu: '1',
+          memory: '4Gi',
+        },
+        limits: {
+          cpu: '1',
+          memory: '4Gi',
+        },
+      },
+      thanosCompact+:: {},
+      thanosStore+:: {},
+```
+
 ### Prometheus Custom Resource with Thanos Sidecar
 
 The `Prometheus` CRD has support for adding a Thanos sidecar to the Prometheus
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
@@ -10,6 +10,53 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
     imageRepos+:: {
       thanos: 'improbable/thanos',
     },
+    replicas+:: {
+      thanosQuery: 1,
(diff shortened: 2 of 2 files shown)


Hashfyre:
cc @metalmatze could you take a look at this?

brancz:
This shouldn't need to be necessary, as you can patch these changes on top of the existing jsonnet out of tree. For the sidecar that would be something along the lines of:

```jsonnet
{
  prometheus+:: {
    prometheus+: {
      spec+: {
        thanos+: {
          resources: {
            requests: {
              cpu: $._config.resources.thanosSideCar.requests.cpu,
              memory: $._config.resources.thanosSideCar.requests.memory,
            },
            limits: {
              cpu: $._config.resources.thanosSideCar.limits.cpu,
              memory: $._config.resources.thanosSideCar.limits.memory,
            },
          },
        },
      },
    },
  },
}
```

Hashfyre:
Correct, that's possible for the sidecar.
But, for the other components, would we be able to set the resource quota without:
```
        container.mixin.resources.withRequests({ cpu: requestsCPU, memory: requestsMEM }) +
        container.mixin.resources.withLimits({ cpu: limitsCPU, memory: limitsMEM });
```

since `local container = container.new()` is used in the code. I've tried patching this, but it did not work. either for resource-quotas or trying to replace the hardcoded replica `1` in
`deployment.new('thanos-query', 1, thanosQueryContainer, podLabels)`

brancz:
Haven't tried this exact patch, but we do this type of strategy all over the place where we use kube-prometheus as a dependency:

```jsonnet
{
  thanos+:: {
    thanosQueryDeployment+: {
        spec+: {
          template+: {
            spec+: {
              containers:
                std.map(
                  function(c)
                    if c.name == 'thanos-query' then
                      c +
                      container.mixin.resources.withRequests({ cpu: requestsCPU, memory: requestsMEM }) +
                      container.mixin.resources.withLimits({ cpu: limitsCPU, memory: limitsMEM }),
                  super.containers,
                ),
            },
          },
        },
    },
  },
}
```

For what it's worth, I'm happy to have snippets like this documented or separate patches that can be additionally applied. Like:

```jsonnet
//... my setup
+ thanos.query.withResources(request, limit)
```

We don't have precedence for this yet, but it's a viable strategy I think.

Hashfyre:
I'll try this patch out. But yes, this needs documentation. I still believe adding
```
container.mixin.resources.withRequests()
container.mixin.resources.withLimits()
```
and,
```
deployment.new('thanos-query', replicas, thanosQueryContainer, podLabels)`
```

with all three directly configurable from `_config+::` offers a much nicer user-experience along with the documentation.

brancz:
I'd prefer a default where possible plus the "separate patches" strategy I described above. I'd like to avoid putting any customization into the config that can be done otherwise.

Hashfyre:
I'm getting `RUNTIME ERROR: attempt to use super when there is no super class.` with the above patch. I'll try to find out the fix.

brancz:
@Hashfyre can you share the whole code you're attempting to execute?

Hashfyre:
@brancz I would not be able to share the whole code, I'll post a small testable unit in the thread.

brancz:
@Hashfyre that works, thanks!

Hashfyre:
Sorry for the delay, I'll post the failing code snippet in a day's time.

Hashfyre:
@brancz this PR can be closed. Since the jsonnet patch with `filter.map()` works.

PR #2441, opened 2019-02-28
Allow browsing of downsampled thanos data in Grafana

Hashfyre:
As per https://github.com/improbable-eng/thanos/issues/601
We need to enable the `'--query.auto-downsampling'` flag
So that grafana can browse data beyond the Raw Data period of prometheus/sidecar

Dashboard showing 70d data of prometheus and alertmanager config-reloads
screen shot 2019-02-28 at 3 08 01 pm


brancz:
looks good to me on green

brancz:
thanks!

Issue #2444, opened 2019-02-28
Feature Request: Ability to inject environment variables in the prometheus pod


sgmitchell:
### Background
I attempted to add a scrape config for google cloud instances [`gce_sd_config`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#gce_sd_config) using an additional scrap config. The prometheus statefulset loaded this config successfully but failed to discover any targets because of an authentication issue.

The prometheus documentation notes that credentials will be found using the standard google SDK process by searching for:
1. a well known environment variable
2. a well know file path
3. the GCE metadata endpoint (if available)

### Workaround
There are 2 workarounds I could find.

My current workaround is to build a custom docker image and adding a secret to the prometheus object:

```
FROM quay.io/prometheus/prometheus:v2.7.1
ENV GOOGLE_APPLICATION_CREDENTIALS=/etc/prometheus/secrets/googleCreds/creds.json
```
```
apiVersion: v1
kind: Secret
...
data:
  creds.json: $B64_ENCODED_CREDS
```
```
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
...
spec:
  image: $MY_CUSTOM_IMAGE
  secrets:
  - googleCreds
  ...
```

Alternatively, [this line](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L739) _seems_ to indicate that I could set the environment variable using the deprecated thanos GCS spec

### Proposed Change
Modify the operator to allow for injecting secrets in to the prom, either by adding a new top level element `env` to the prometheus spec or extending the existing `containers` spec to allow for patching the the generated containers.

I'll have a PR ready for the patch containers solution soon

sterchelen:
@sgmitchell Any news about your proposal ?

sgmitchell:
#2445 implements this and should be merged in soon

swtch1:
#2445 was merged.  This issue should likely be closed if this is the recommended pattern.

PR #2445, opened 2019-03-01
Add the ability to modify arbitrary parts of the operator generated containers

sgmitchell:
This change allows the existing `containers` array in the Prometheus spec to overwrite other containers with the same name, even the ones generated by the operator.

This works by using kubernetes StrategicMergePatch (the same method used by `kubectl patch`) to patch the containers generated by the operator with any containers defined in `.spec.containers` that share the same name (`prometheus`, `prometheus-config-reloader`, `rules-configmap-reloader`, and `thanos-sidecar`)

**Example Usage**
Modifying the prometheus container with an environment variable results in no new containers and a change to the prometheus container
```
kind: Prometheus
...
spec:
  containers:
  - name: "prometheus"
    env:
    - name: A
      value: B
...
```

Fix for: #2444


brancz:
I’m inclined to accept this, even though it has been a non goal since the start of the Prometheus operator. I’m ok with it under the condition of 1) the other maintainer are and 2) adding into the documentation, that making use of this voids any support for upgrades and the Prometheus operator team will not help debugging customized setups.

brancz:
cc @mxinden @metalmatze @squat @s-urbaniak

sgmitchell:
100% agreed that doing this should "void the support warranty". I was initially planning on adding  line that logged a warning when containers were overridden to help prevent accidental overrides, but it wasn't in line with the rest of the code. If anyone has any suggestions about that, I'd be happy to take a crack at implementing

brancz:
I think the warning is not necessary, the user intentionally engages in doing this, so I think they need to know what they're doing anyways.

Ping @mxinden @metalmatze @squat @s-urbaniak once more to see what they think.

s-urbaniak:
I am generally also fine to override, but I am also concerned that using this mechanism internal containers could be overridden by accident.

Just to play a little bit the devil's advocate: This PR solves the issue of injecting a env var by allowing custom images to be set in the prometheus operator, which is an orthogonal solution.

@sgmitchell untill I miss something obvious, can't you mount the secret into the well-known path as documented in https://prometheus.io/docs/prometheus/latest/configuration/configuration/#gce_sd_config directly instead of building a custom image?
> a JSON file in the well-known path $HOME/.config/gcloud/application_default_credentials.json

sgmitchell:
> can't you mount the secret into the well-known

Unfortunately not, unless I overlooked something (which is entirely possible).

The well known path will resolves to `/home/.config/gcloud/application_default_credentials.json` in quay.io/prometheus/prometheus:v2.7.1 and operator mounts all [configmaps](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L524) and [secrets](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L506) under a [fixed directory](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L45) in `/etc/prometheus`.

brancz:
For what it's worth, I do think that separately whatever functionality you need we should ensure to have as a first class feature, rather than making users do random patches. In regard to your problem, I think we should allow specifying those credentials either way. Whether it's via an environment variable or another way, shouldn't really be relevant I think.

sgmitchell:
I agree that having environment variables in the CRD would put more guardrails around the process. I shied away from that for this more general approach assuming that it's impossible to predict all of the use cases people might have and that not all of them should be included as a field.
Personally, I was surprised that I couldn't find any issues or feature requests for something like this, so it felt a little presumptive to add a whole new field to the crd spec.

metalmatze:
Agreed, if the additional env vars solve the initial problem we should try to delay the additional container change. If we encounter a problem in the future, we can still add it.

sgmitchell:
Sounds like there's a consensus which is fine by me.

Thoughts on what this new field should look like?
Some different ways I can think of (in order of increasing flexibility/danger) are:
* `googleApplicationCredentials string` that would just add the `GOOGLE_APPLICATION_CREDENTIALS` environment variable with the value if set
* `env []EnvVar` would add environment variables to the prometheus container (or maybe all containers)
* `env map[string][]EnvVar` would add environment variables to the same container name as the key

brancz:
I think the first is probably the most canonical thing we do in the prometheus operator, but having that point at a Secret within the same namespace as the Prometheus object. As a side note: didn't Thanos unify the way these configurations are specified (cc @metalmatze) ?

sgmitchell:
I know thanos has the concept of the object store config but for [GCS](https://github.com/improbable-eng/thanos/blob/master/pkg/objstore/gcs/gcs.go#L29) it only takes a bucket. For [S3](https://github.com/improbable-eng/thanos/blob/master/pkg/objstore/s3/s3.go#L37) the config contains all of the needed access credentials.
I feel like there used to be an example manifest in the repo with a GCS bucket that used the `GOOGLE_APPLICATION_CREDENTIALS` environment variable but couldn't find it.

sgmitchell:
Closing in favor of the narrower scope in #2499

brancz:
As commented in https://github.com/coreos/prometheus-operator/pull/2499 I'm happy to re-open this and move forward.

sgmitchell:
Reopening per discussion in #2499

sgmitchell:
From the [other pr](https://github.com/coreos/prometheus-operator/pull/2499#issuecomment-484512020):
>> I like this patch, because we can explicitly let the user know that they're going into unsupported territory.
>
> How would you like to tackle that? Add documentation somewhere? I think it should be part of this change then.
-@metalmatze 

I think documentation would be the best solution, possibly with a warning as backup. There are currently 4 possible conflicting names: `prometheus`, `prometheus-config-reloader`, `rules-configmap-reloader`, and `thanos-sidecar`. It shouldn't be too tedious to keep track of them.

The containers object doesn't seem so the commonly used or advocated so developers that take advantage of it would almost certainly be reading the docs to discover it. A very quick google search didn't yield any blogs or tutorials that use the container type so the risk of blind copy/paste should be low

If we also think that warning are a must, we can add a warning [here](https://github.com/coreos/prometheus-operator/blob/b6e2e41b3ab1b57614703f765c48799f054262f3/pkg/prometheus/operator.go#L1039) that might look something like:
```
for _, c := range p.Spec.Containers {
  if c.Name == "prometheus" || ... {
    level.Warn(c.logger).Log("msg", "StatefulSet has mergepatched container", "container", c.Name)
  }
}
```

brancz:
Yep, I think it's relatively reasonable to track. This does still need extra documentation stating, that this is unsupported territory, where we may break naming and or remove containers at our will at any time without notice.

metalmatze:
We might even want to fail hard or create a metric for when there are extra containers with the same name. Otherwise a silent fail will be not obvious enough.

brancz:
The point of this PR is that fields of those containers can be modified, as the merge-patch is applied on matching names.

metalmatze:
Ok. I wasn't exactly sure about that. SGTM.

sgmitchell:
I've validated that my rebase runs properly in my environment.
The one test failure looks like alertmanager clustering failed to start. Could someone rerun that? I'm reasonably confident that it's a flake and not related to my changes.

brancz:
Sorry this is a test flake that we've been fighting for a while. We're hoping to have this fixed soon, for now we'll have to re-run.

sgmitchell:
No big deal. Those kind of integration tests are never perfect.

Let me know if you want any changes to the documentation. Otherwise, I'll just ping the thread periodically until we get past the flake

brancz:
I'm happy with the documentation as is. @metalmatze what do you think?

metalmatze:
LGTM. 

Nitpick would be to add more information about how the merging works, like is the original container merged with the new container spec, when the name is the same? Do we want some more user facing documentation, in `Documentation/`?
All of this can be solved in a next PR, so I'm fine moving forward with this one! :+1:

metalmatze:
This seems to need a rebase.

brancz:
lgtm thanks for the effort and for bearing with us! :) :tada:

sgmitchell:
Not a problem. Happy to contribute.

sterchelen:
@brancz @sgmitchell Do you know when a new image tag will be pushed with this feature :) ?

metalmatze:
I might do a release for the Operator today, but can't promise anything yet.

Shini31:
Hello,

So what is the method for adding gcloud credentials into prometheus?

Regards

acondrat:
@Shini31 In case you are using the [prometheus-operator](https://hub.helm.sh/charts/stable/prometheus-operator) Chart, add/adjust the following settings in your values.yaml

```
  prometheus:
    prometheusSpec:
      secrets:
        - prometheus-service-account
      containers:
        - name: prometheus
          env:
            - name: GOOGLE_APPLICATION_CREDENTIALS
              value: /etc/prometheus/secrets/prometheus-service-account/credentials.json
```

FelixRodriguezJara:
Hello. I have gone through the same issue. In my case, it is thanos-sidecar which needs the variable. I have managed to get the environment variable inside the thanos-sidecar container as explained by @acondrat. I have created a secret called gcs-credentials, like this:

kubectl -n observability create secret generic gcs-credentials --from-file=gcs-credentials.json 

In credentials.json, I have included what I´ve got from GCP. I´ve tested this credentials somewhere else and they work. Then, I have added the secret to the prometheus specs:
---
    secrets:
      - gcs-credentials
---
    containers:
      - name: thanos-sidecar
        env:
          - name: GOOGLE_APPLICATION_CREDENTIALS
            value: /etc/prometheus/secrets/secret-gcs-credentials/gcs-credentials.json
---
 However, the mount path isn´t reachable by the container, see below. How can I make it available for the thanos-sidecar container?

# kubectl logs prometheus-prom-op-prometheus-operato-prometheus-0 -n observability thanos-sidecar
level=info ts=2019-10-16T11:51:31.822968432Z caller=flags.go:90 msg="StoreAPI address that will be propagated through gossip" address=172.26.0.12:10901
level=info ts=2019-10-16T11:51:31.838634538Z caller=main.go:256 component=sidecar msg="disabled TLS, key and cert must be set to enable"
level=info ts=2019-10-16T11:51:31.838693504Z caller=factory.go:39 msg="loading bucket configuration"
sidecar command failed: create GCS client: dialing: google: error getting credentials using GOOGLE_APPLICATION_CREDENTIALS environment variable: open /etc/prometheus/secrets/secret-gcs-credentials/gcs-credentials.json: no such file or directory

Thank you very much!

PR #2454, opened 2019-03-04
pkg/prometheus: watch configmaps having the prometheus-name selector

s-urbaniak:
Currently, we are watching all configmaps amongst all configured tenant namespaces. This is very costly in certain deployment use cases where we have:

a) many namespace tenants
b) kubernetes controllers running in those namespace doing leader election using k8s intrinsics

Here, we can observe quite a high CPU usage of the prometheus operator of up to 15% CPU "idle" usage:
image

A profile revealed most CPU time spent in the reconciliation loop:
```
(pprof) top10 -cum
Showing nodes accounting for 0.03s, 0.32% of 9.33s total
Dropped 359 nodes (cum <= 0.05s)
Showing top 10 nodes out of 271
      flat  flat%   sum%        cum   cum%
         0     0%     0%      6.11s 65.49%  github.com/coreos/prometheus-operator/pkg/prometheus.(*Operator).processNextWorkItem
         0     0%     0%      6.11s 65.49%  github.com/coreos/prometheus-operator/pkg/prometheus.(*Operator).sync
         0     0%     0%      6.11s 65.49%  github.com/coreos/prometheus-operator/pkg/prometheus.(*Operator).worker
         0     0%     0%      3.81s 40.84%  github.com/coreos/prometheus-operator/pkg/prometheus.(*Operator).createOrUpdateConfigurationSecret
```

In the above scenario we have a lot of configmap updates, namely `*-lock` configmaps which are being used inside kubernetes to do leader election. In the environment under test (OpenShift), we have up to 3 configmap updates per second:

image

This fixes the above by watching for configmaps having the prometheus-name label selector only. CPU time is reduced to ~0,7% idle CPU usage:
image

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -33,7 +33,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	appsv1 "k8s.io/api/apps/v1beta2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	extensionsobj "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


brancz:
lgtm :+1: thanks!

PR #2455, opened 2019-03-05
Set "instance" as node hostname instead of IP

Tri0L:
Add relabelings for set "instance" as node hostname instead of IP.


brancz:
This goes against Prometheus best practices and also doesn't make sense, as the `instance` label is always supposed to reflect the identity of the target, and when it's the node name, this is not unique to the pods on the node anymore.

PR #2456, opened 2019-03-05
Update Prometheus Operator in kube-prometheus to v0.29.0

friedrich-at-adobe:


friedrich-at-adobe:
Just realized per https://github.com/coreos/prometheus-operator/issues/1757#issuecomment-411034013  that kube-prometheus is not versioned.

PR #2461, opened 2019-03-06
Document components that can't be scraped on EKS

sverch:
The kube-scheduler and kube-cluster-manager are not visible from a managed EKS cluster.  They are likely run inside the managed control plane as a black box.  This means you cannot use Prometheus to scrape these components and must rely on your cloud provider for metrics.


brancz:
lgtm :+1: thanks!

PR #2470, opened 2019-03-10
Fix recommended labels

pst:
According to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/, the recommended labels are prefixed `app.kubernetes.io`, not `apps.kubernetes.io`. I'm assuming this is a simple oversight, not a disagreement with the recommended upstream labels.


ant31:
configuration/manifest should be regenereted too

brancz:
In a second commit just run

```
make clean generate-in-docker
```

PR #2471, opened 2019-03-10
Fix PrometheusTSDBWALCorruptions alert

dghubble:
* `tsdb_wal_corruptions_total metric` name should be `prometheus_tsdb_wal_corruptions_total`

### DIFF
--- a/contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/prometheus.libsonnet
+++ b/contrib/kube-prometheus/jsonnet/kube-prometheus/alerts/prometheus.libsonnet
@@ -109,7 +109,7 @@
               summary: 'Prometheus write-ahead log is corrupted',
             },
             expr: |||
-              tsdb_wal_corruptions_total{%(prometheusSelector)s} > 0
+              prometheus_tsdb_wal_corruptions_total{%(prometheusSelector)s} > 0
             ||| % $._config,
             'for': '4h',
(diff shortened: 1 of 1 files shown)


dghubble:
I'd regenerate yaml files in a separate commit, but somewhere in the tangle of (assumed sudoless) docker and relative references the `../../hack/jsonnet-docker-image` Makefile step errors on symlinks. I've reached my timebox.

brancz:
yes, jsonnet only changes don't necessarily need re-gen, also I'm happy to inform you that we are thinking of ways to remove generation entirely and only make it a release artifact (that can only happen after we move this to a separate repo though)

brancz:
Thanks @dghubble!

dghubble:
:wave: Hi!

@s-urbaniak It should change https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/prometheus-rules.yaml#L954 and thanks @brancz!

Manifest generation in-repo does seem tricky. I suppose you're wanting to still have rendered manifests so the result of libsonnet changes can be seen, but generating in PRs is an extra step too. Separately, I've had good experiences switching docker-based artifact creation to buildah / buildah unshare to drop the (effective) root requirement, though the tool is probably less familiar to many. But if its part of releases I suppose contributors don't deal with it at all. Thanks for the updates, looking forward to how that goes!

PR #2473, opened 2019-03-11
contrib/kube-prometheus: Update Prometheus, Thanos & Grafana

metalmatze:
Update Prometheus, Thanos and Grafana's default versions.
This PR is mostly here, so that we can update Grafana from beta to stable.

/cc @brancz


example/rbac/prometheus-operator/prometheus-operator-deployment.yaml
```diff
@@ -2,23 +2,18 @@ apiVersion: apps/v1beta2
 kind: Deployment
 metadata:
   labels:
-    apps.kubernetes.io/component: controller
```
brancz: I think you might have missed a jsonnet dependency update

PR #2488, opened 2019-03-14
Disable KubeClientCertificateExpiration when metric 0

robinro:
The alert should not fire if the metric value is exactly 0, i.e. no expiration duration was recorded by the server. This happens for example with aws-iam-authenticator on EKS.

### DIFF
--- a/contrib/kube-prometheus/manifests/prometheus-rules.yaml
+++ b/contrib/kube-prometheus/manifests/prometheus-rules.yaml
@@ -792,7 +792,7 @@ spec:
           in less than 7 days.
         runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubeclientcertificateexpiration
       expr: |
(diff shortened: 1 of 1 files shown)


robinro:
Related to (but not fixing all aspects of) https://github.com/coreos/prometheus-operator/issues/1843

brancz:
This alert is defined in the [kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin) it needs to be fixed there first, and then we can update the jsonnet dependency in this repository to that version.

llamahunter:
Created https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/177

metalmatze:
As far as I can see we now have two versions of this PR over at the mixin repository:
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/176
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/177

However somebody created a fix with a different query. If you could take a look at this one too, that would be helpful!
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/163

robinro:
@metalmatze @llamahunter @brancz Thanks for helping out. I didn't see kubernetes-monitoring/kubernetes-mixin#163 before. That solves the problem and is imo a better solution than the one I proposed.

PR #2494, opened 2019-03-16
Update travis to latest minikube, k8s, jsonnet

goll:
Attempt to update travis scripts to use latest versions of minikube (v0.35.0), kubernetes (v1.13.4) and jsonnet (v0.12.1). 
* `dist: xenial` is required for the default minikube bootstrapper (kubeadm) as it depends on systemd
* bump go to latest `1.11.x`
* enable caching of go build cache to speed up builds
* enable go build cache when running `po-jsonnet` container
* remove redundant `kube-dns` check from e2e scripts
* `create_minikube.sh`: use minikube config to set defaults, add `kube-addon-manager` check
* `Dockerfile`: use latest jsonnet `v0.12.1`, replace wget with curl

### DIFF
--- a/contrib/kube-prometheus/Makefile
+++ b/contrib/kube-prometheus/Makefile
@@ -18,6 +18,7 @@ generate-in-docker: ../../hack/jsonnet-docker-image
 	--rm \
 	-u=$(shell id -u $(USER)):$(shell id -g $(USER)) \
 	-v $(shell dirname $(dir $(abspath $(dir $$PWD)))):/go/src/github.com/coreos/prometheus-operator/ \
+	-v $(shell go env GOCACHE):/.cache/go-build \
 	--workdir /go/src/github.com/coreos/prometheus-operator/contrib/kube-prometheus \
 	po-jsonnet make generate
 
@@ -43,7 +44,7 @@ test: $(JB_BINARY)
 	./test.sh
 
 test-e2e:
-	go test -timeout 55m -v ./tests/e2e
+	go test -timeout 55m -v ./tests/e2e -count=1
 
 test-in-docker: ../../hack/jsonnet-docker-image
 	@echo ">> Compiling assets and generating Kubernetes manifests"
@@ -51,6 +52,7 @@ test-in-docker: ../../hack/jsonnet-docker-image
 	--rm \
 	-u=$(shell id -u $(USER)):$(shell id -g $(USER)) \
 	-v $(shell dirname $(dir $(abspath $(dir $$PWD)))):/go/src/github.com/coreos/prometheus-operator/ \
+	-v $(shell go env GOCACHE):/.cache/go-build \
 	--workdir /go/src/github.com/coreos/prometheus-operator/contrib/kube-prometheus \
 	po-jsonnet make test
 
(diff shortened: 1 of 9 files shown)


metalmatze:
Really cool! Thanks for the effort to keep it up to date!
I'm generally fine with these changes.

Maybe @mxinden wants to take another look at the testing and `.travis.yml` file especially.

contrib/kube-prometheus/tests/e2e/travis-e2e.sh
```diff
@@ -12,9 +12,6 @@ SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
 
 "${SCRIPT_DIR}"/../../../../scripts/create-minikube.sh
 
-# waiting for kube-dns to be ready
-JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'; until kubectl -n kube-system get pods -lk8s-app=kube-dns -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do sleep 1;echo "waiting for kube-dns to be available"; kubectl get pods --all-namespaces; done
```
mxinden: :+1: for moving this into the start script.

contrib/kube-prometheus/Makefile
```diff
@@ -18,6 +18,7 @@ generate-in-docker: ../../hack/jsonnet-docker-image
 	--rm \
 	-u=$(shell id -u $(USER)):$(shell id -g $(USER)) \
 	-v $(shell dirname $(dir $(abspath $(dir $$PWD)))):/go/src/github.com/coreos/prometheus-operator/ \
+	-v $(shell go env GOCACHE):/.cache/go-build \
```
mxinden: As far as I know, go cache invalidation only happens based on golang code changes. If we change the environment itself, tests would not be rerun.

Thereby I find this a bit risky compared to the time saved through it. What do you think?

contrib/kube-prometheus/Makefile
```diff
@@ -51,6 +52,7 @@ test-in-docker: ../../hack/jsonnet-docker-image
...
```
mxinden: As far as I know, go cache invalidation only happens based on golang code changes. If we change the environment itself, tests would not be rerun.

Thereby I find this a bit risky compared to the time saved through it. What do you think?

goll: Thank you for the comment, I agree as I had the same doubts, in the end I added the cache to the containers because go 1.12 requires the build cache.

Would invalidating the test cache solve this issue?
AFAIK to disable test caching you only need the `-count=1` flag:
https://golang.org/cmd/go/#hdr-Testing_flags
```
When 'go test' runs in package list mode, 'go test' caches successful package test results to avoid unnecessary repeated running of tests. To disable test caching, use any test flag or argument other than the cacheable flags. The idiomatic way to disable test caching explicitly is to use -count=1. 
```

goll:
@mxinden following my last comment, disabled `go test` cache everywhere with `-count=1`, so the `GOCACHE` should only be used by build/install cmds, providing future compatibility with go 1.12 and hopefully improving job times, what do you think, would this solve the CI caching issues?

squat:
tests failed due to the known flake:
```
            alertmanager_test.go:600: One notification expected, but 3 received.
                
                Alertmanager Notification Payload Received
                Alertmanager Notification Payload Received
                Alertmanager Notification Payload Received
```
restarting the one failed job

brancz:
Curious as restarting for the second time caused the exact same failure. Kicking it one more time, if this doesn't work, we should look into it further.

brancz:
Ok it seems we were just lucky enough to hit the same flake twice :)

brancz:
Thanks @goll! :+1: :tada:

PR #2499, opened 2019-03-19
pkg/prometheus: Add new field googleApplicationCredentials

sgmitchell:
Add a new field `googleApplicationCredentials` that mounts a secret and
sets the environment variable GOOGLE_APPLICATION_CREDENTIALS to its path.

Follow up to discussion in #2445


brancz:
There seems to be some formatting that CI wants, just run

```
make clean format generate-in-docker
```

That will also generate the required API documentation.

pkg/prometheus/statefulset.go
```diff
@@ -47,6 +47,7 @@ const (
 	configFilename                  = "prometheus.yaml.gz"
 	configEnvsubstFilename          = "prometheus.env.yaml"
 	sSetInputHashName               = "prometheus-operator-input-hash"
+	googleCredsEnvar                = "GOOGLE_APPLICATION_CREDENTIALS"
```
metalmatze: Let's simply call it this:
```suggestion
	googleCredentials                = "GOOGLE_APPLICATION_CREDENTIALS"
```

sgmitchell:
Hmm. That last failure looks like a flake.
```
--- FAIL: TestAllNS/x/AMMeshAfterRollingUpdate (52.06s)
            alertmanager_test.go:285: failed to wait for initialized alertmanager mesh: the server is currently unable to handle the request (get pods alertmanager-test-1): <nil>
```

Can someone with permissions rerun that build.

brancz:
@metalmatze was thinking of making this an explicit configuration on thanos, then the env var injection wouldn't be necessary, I think I like that solution even better.

metalmatze:
Yes, sorry for not communicating here earlier. I've talk to the Thanos maintainers and they are fine with having the GCS configuration in the Thanos configuration. Then we only need to care about one generic config file (which is a Secret for us) and not add different ones for each object storage.
I'll link the PR here once I have it up.

metalmatze:
https://github.com/improbable-eng/thanos/pull/963 has been merged and we should now simply wait for a new relase of Thanos.
Once that is out, there should be no changes needed to the Prometheus Operator to get Thanos up and running on GCS now.

philpearl:
What about cases where you need GOOGLE_APPLICATION_CREDENTIALS but you aren't using Thanos? For example the scenario in https://github.com/coreos/prometheus-operator/issues/2444 - trying to run a gce-sd-config scrape?

sgmitchell:
@metalmatze I agree with @philpearl that even with [improbable-eng/thanos#963](https://github.com/improbable-eng/thanos/pull/963) the operator will still need a code change.

While the referenced PR gives us some consistent structure for how to define the service account, we still need to add the environment variable to the prometheus container. The [current code](https://github.com/coreos/prometheus-operator/blob/9b61c8fc2dd6f65937af26943a714ce5aee84aa4/pkg/prometheus/statefulset.go#L727) only applies to the thanos sidecar and will have no impact on prometheus service discovery

brancz:
Prometheus stance is that everything should be configurable via the config file, if it's not then that needs to be fixed. Any environment variable configuration is purely coincidental, and may be broken at any time.

philpearl:
In this case that would be an odd thing to do. All GCP APIs are accessed the same way - either by picking up credentials from the GCE credentials API, by looking for a file via GOOGLE_APPLICATION_CREDENTIALS or by reading $HOME/.config/gcloud/application_default_credentials.json. In the kubernetes case the recommended way is to mount a secret containing the credentials and point to it with GOOGLE_APPLICATION_CREDENTIALS. All the GCP API libraries from Google are built to pick up this information automatically from the environment. This is unlikely to change or break.

A GCP developer would not expect to reference GCP credentials in a config file, and I expect the devs behind gce_sd_config will think the same way. I don't think asking them to change is the right thing to do.

brancz:
> All the GCP API libraries from Google are built to pick up this information automatically from the environment. This is unlikely to change or break.

This doesn't change the Prometheus project's stance. A precedence case is for example http proxy, even if there is a widely used convention of respecting the `HTTP_PROXY` env var. Prometheus broke in favor of the explicit configuration in it's config file. This is for consistency in configuration, to have a single place of stable and reliable configuration.

Regardless, this is something that needs to be clarified upstream. As it stands, we're not implementing something in a stable API, that has no stability guarantee on Prometheus.

philpearl:
GOOGLE_APPLICATION_CREDENTIALS is in the doc for gce_sd_config. I don't understand what further clarification is needed. https://prometheus.io/docs/prometheus/latest/configuration/configuration/#gce_sd_config

I'm happy to open a prometheus issue, but I'm struggling with the content beyond "when you said this, did you really mean it?"

brancz:
Interesting, thanks for sharing. This is not usually how we do things in regard to configuration on Prometheus. I think in that case maybe having an env var array to be injected into the prometheus container would be the appropriate solution. I'll talk to the other prometheus team members though to see what they think.

sgmitchell:
Since this is proving to be a more contentious issue than I had hoped and prometheus service discovery seems to be against any changes without proper integration tests ([discussion here](https://groups.google.com/forum/#!topic/prometheus-developers/3B3jBsErK5M)), I'd like to re-float my initial PR #2445 that allowed arbitrary modifications to the generated operator containers

brancz:
Sorry for being so late to answer this. I think I agree. There will always be small things to fiddle with, which doesn't necessarily mean we support them. I like this patch, because we can explicitly let the user know that they're going into unsupported territory.

metalmatze:
Alright then, I'm fine with having it then too. I hoped the Thanos change would resolve the situation, but I can see how it is still limiting.

> I like this patch, because we can explicitly let the user know that they're going into unsupported territory.

How would you like to tackle that? Add documentation somewhere? I think it should be part of this change then.

metalmatze:
Ok. Got it. Seems like I mixed it up with all the issues I've open in my window to review. :sweat_smile:

Issue #2457, opened 2019-03-05
`additionalScrapeConfigs.optional` doesn't unblock cluster creation with nonexistent secret


davidquarles:
**What did you do?**
I created a promethei with `additionalScrapeConfigs.optional: true`, pointing to a nonexistent secret.

**What did you expect to see?**
A new prometheus-operated statefulset with additional scrape configs injected if/when the specified secret/key exists.

**What did you see instead? Under which circumstances?**
A new `prometheus` was created, but the absence of the "optional" secret actually blocked statefulset creation.

**Additional context:**
I've written a custom controller that watches namespaces within our clusters and automatically injects a `prometheus` into each as they are spun up.  We want to functionally shard by namespace to reduce the footprint of our ingestion layer, to segment compute/storage costs by namespace, etc... with a unified cluster-level query + rule-eval layer via thanos.

I have this wired up successfully today, with the caveat that a small handful of namespaces ship high-cardinality metrics that we want to aggregate pre-ingestion.  In the absence of any match-based filtering in the ingestion layer, it seems like federating to lower-tier promethei that don't participate in the thanos cluster is a viable solution.  That said, I'd like to provide a mechanism to (optionally) use `additionalScrapeConfigs`.

Statefulsets are created if I create that secret with an empty value for the specified key, but I'd like to just leave factor that secret out of the control loop altogether and let teams create it if/when they deem that pattern necessary.  Unless I'm missing something and this functionality in fact exists, it seems like I'll have to write more nuanced logic around that secret creation, such that we create, but do not mutate.  I am of course open to any/all design feedback, ideas, etc.  Thank you!

**Environment**

* Prometheus Operator version:

`quay.io/coreos/prometheus-operator:v0.29.0`

* Kubernetes version information:

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.7", GitCommit:"65ecaf0671341311ce6aea0edab46ee69f65d59e", GitTreeState:"clean", BuildDate:"2019-01-24T19:32:00Z", GoVersion:"go1.10.7", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #2511, opened 2019-03-26
pkg/prometheus: optional secret in scrapeconfig

squat:
This commit enables the creation of a Prometheus when
additionalScrapeConfigs is defined but points to a non-existent secret
and `additionalScrapeConfigs.optional` is set to `true`.

Formerly, the creation of the Prometheus would error; this change adds
support to respect the `optional` field.

Fixes: #2457


davidquarles:
Thank you so much @squat!

PR #2522, opened 2019-03-29
add support for initContainers in Alertmanager and Prometheus

### LINKED ISSUE #1888: Support initContainers on `kind: Prometheus`
This is missing as in some cases when using `hostPath` volumes with non-root it is important to be able to `chown -R ...` on the init container.

### LINKED ISSUE #2708: Support Volumes on `kind: Prometheus`
It would be useful if we had the ability to express ```Volumes``` in the Prometheus Operator's CRD ```kind: Prometheus```.  These volumes would be useful for passing data between ```initContainer``` instances that are part of a Prometheus instance pod.  A particular use case for this is when trying to use InitContainers to log into and fetch secrets from Vault, then insert them into template files (the 'vault-auth/consul-template' process).  Secrets could include things such as TLS certs, as well as login creds for scrap targets, etc.

InitContainers themselves are not currently supported in the operator's Prometheus object, either.  However, there is an existing issue that seeks to address that:
https://github.com/coreos/prometheus-operator/issues/1888

tongpu:
Fixes #1888 
Fixes #2708


tongpu:
I need to say that I haven't yet tested if it works as expected, I was already happy to see it compile. :D

brancz:
Can you elaborate on the use case for this? Because the hostPath argument from #1888 is not a valid one in my opinion. hostPath should never be used to run Prometheus, if you really want to consume local storage then do so via the now stable local-volume feature.

tongpu:
@brancz In our case we're facing permission issues with PVs created via [trident](https://github.com/NetApp/trident), the storage orchestrator for NetApp. PVs are always created with root:root permissions and because of that the alertmanager and prometheus pods fail to start because of the following error:
```
Error: failed to start container "prometheus": Error response from daemon: oci runtime error: container_linux.go:317: starting container process caused "chdir to cwd (\"/prometheus\") set in config.json failed: permission denied"
```
Because of that we would like to be able to add an initContainer to the pod definition that fixes these permissions for us, before starting the main application containers. Sadly trident doesn't yet have support for fsGroup, which could possibly also fix this issue.

Currently the only possible solutions to run promtheus/alertmanager instances created by prometheus-operator in our environment is to use `runAsNonroot: false` and `runAsUser: 0` to run all services as root.

tongpu:
I have the feeling that the limitation is already on the NetApp side, because NFS volumes are always rolled out with root:root and that default can't be changed.

I don't see why it would be an issue to expose a feature that the pod API already provides in the prometheus-operator API, when the same this is already being done with the ability to configure additional containers in the pod.

brancz:
> I don't see why it would be an issue to expose a feature that the pod API already provides in the prometheus-operator API, when the same this is already being done with the ability to configure additional containers in the pod.

This argument would hold for exposing _any_ field of the Pod API, but that's not the point of the prometheus-operator, the point is to automate. There are two issues here, 1) Prometheus explicitly advises [against usage of NFS](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects), as we've seen a ton of inconsistencies and many implementations are known to yield incorrect behavior 2) even if using NFS and using an init container, you still need to run the init container as root to chown the directory "correctly", while running any process of a Prometheus pod as root would be a very bad idea.

If you know the permissions to be provisioned up front, then I recommend you to prepare your volumes before as well, outside of the scope of the Prometheus Operator.

erkiesken:
FYI the freshly created PV permission issue that is fixable with `chown -R 65534:65534 /prometheus` in initContainer also happens on GKE every time new prometheus instance with new PVC/PV is created.

So getting initContainer support into Prometheus CRD is very welcome for GKE users too.

tongpu:
I still believe that this would be a welcome feature that allows users to fix e.g. permissions issues that seem to be rather common. I'm with @brancz that the issues should be fixed upstream, but this feature gives users the possibility to fix the issues when they're running into them.

brancz:
There are other solutions as well, such as preparing the volume beforehand, before the Prometheus/Alertmanager Pod even mount it.

rchapple:
@tongpu -- do you still intend to merge these changes?  They look like they are valid/useful and are very close to mergable.  Would it be much effort to resolve these conflicts?  Thanks!!

tongpu:
@rchapple I'm willing to fix any merge conflicts as longs as a maintainer tells me that the changes are going to get merged in the end.
@brancz What your take on that?

rchapple:
@tongpu -- I think we might want to touch up the API docs for the added spec params, in order to satisfy @brancz's request that we steer users toward just using the `initContainers` for secret fetch use cases.

See this comment:
https://github.com/coreos/prometheus-operator/issues/2708#issuecomment-522904062

tongpu:
@brancz -- I've rebased my changes and made the intended use cases for `initContainers` clear in the documentation. Feel free to propose any changes to the documentation text if you don't feel comfortable with it.
Are there any tests you would like me to add for the feature?

tongpu:
@rchapple -- I already rebased my changes this afternoon but somehow `make generate-in-docker` is currently failing on me. I'll check again later.

tongpu:
@rchapple -- Finally managed to get the config and API generator to work again.
@brancz -- Would love to get this merged before another set of merge conflicts.

brancz:
lgtm :+1:

PR #2526, opened 2019-04-02
Fix node:node_disk_saturation rules to return value in seconds

nabadger:
Resolves https://github.com/coreos/prometheus-operator/issues/2525

This is a re-generation of jsonnet for kubernetes-mixin.

Let me know if this have not generated correctly. I found that running `jb update` updated all jsonnet deps (which is a lot of changes not related to this fix). I minimized this by only updating the jsonnetlock file dep.  for kubernetes-mixin and running `jb install` instead.


brancz:
lgtm on green

brancz:
Thanks @nabadger !

PR #2528, opened 2019-04-02
Fix prometheus adapter double accounting

brancz:
This should fix https://bugzilla.redhat.com/show_bug.cgi?id=1669410

@mxinden @metalmatze @squat @s-urbaniak @paulfantom


metalmatze:
Restarted flaky tests.

PR #2536, opened 2019-04-04
VolumeClaimTemplate is silently ignored

auhlig:
As outlined in issue https://github.com/coreos/prometheus-operator/issues/2535


auhlig:
fails due to unrelated alertmanager mesh test: https://travis-ci.org/coreos/prometheus-operator/jobs/515669485#L661

PR #2543, opened 2019-04-08
Set Alertmanager session affinity to ClientIP

brancz:
@mxinden @metalmatze @s-urbaniak @squat @paulfantom

### DIFF
--- a/contrib/kube-prometheus/jsonnetfile.lock.json
+++ b/contrib/kube-prometheus/jsonnetfile.lock.json
@@ -8,7 +8,7 @@
                     "subdir": "contrib/kube-prometheus/jsonnet/kube-prometheus"
                 }
             },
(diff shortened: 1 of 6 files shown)


contrib/kube-prometheus/jsonnetfile.lock.json
```diff
@@ -28,7 +28,7 @@
                     "subdir": ""
                 }
             },
-            "version": "0669b548b8bc981f2676e7ec70c8f4a05fa39aa7"
+            "version": "3991660410dab201bb1f60b84d26e027c6c877e4"
```
mxinden: Is this necessary?

brancz: Not strictly, but we need to update anyways.

PR #2550, opened 2019-04-10
Node clock skew and node network interface alerts

brancz:
@mxinden @metalmatze @s-urbaniak @squat @paulfantom


s-urbaniak:
restarted flaky AM, we need to look into this :thinking:

mxinden:
> restarted flaky AM, we need to look into this thinking

@s-urbaniak investigation is happening here: https://github.com/coreos/prometheus-operator/pull/2544

Issue #2273, opened 2019-01-09
Prometheus Operator should validate Prometheus Rule expression syntax


mikelduke:
**What did you do?**
Create a PrometheusRule with an invalid prometheus expression, like missing quotes or unmatched parentheses in the query.

**What did you expect to see?**
Prometheus Operator should syntax check and reject the invalid rule before adding it to the configmap of Prometheus Rules and log a helpful message with at least the namespace name, PrometheusRule object name, and error information. Ouput similar to promtools check rules CLI would be great.

**What did you see instead? Under which circumstances?**
Prometheus Operator adds the invalid rule to the configmap. This causes Prometheus to fail to reload the config file, while it keeps using the previous version. Finding the offender can be a pain and users must wait until the broken rule is fixed before seeing newer rule updates. If the Prometheus pod bounces, it cannot restart due to the invalid config.

mikelduke:
It would be even better if the api server rejected the creation or update of a PrometheusRule object with bad rules.

brancz:
Agreed. We've been thinking to add a validating webhook that validates the PromQL expressions. Biggest problem with this is cert setup is somewhat difficult so we've been discussing what we can do with the cert-manager people.

Hashfyre:
This issue also has taken down entire prometheus monitoring stack.

For us the bad rule had:
-  `for: "2"` in stead of a `for: 2m`
- `(expr) > 10m` in stead of a `(expr) > 10`

We ended up losing monitoring data for around an hour.

Hashfyre:
OK, I found out a way to do CI validation of `PrometheusRules` manifests.
`cat ./rulefile.yaml| yq --yaml-output .spec | promtool check rules /dev/stdin`
uses: 
 - `pip install yq`
 - `go get github.com/prometheus/prometheus/cmd/promtool`

brancz:
I recommend @Hashfyre's solution until the validating webhook is implemented (if anyone would like to get started on this, we're happy to provide some guidance).

mikelduke:
> OK, I found out a way to do CI validation of `PrometheusRules` manifests.
> `cat ./rulefile.yaml| yq --yaml-output .spec | promtool check rules /dev/stdin`
> uses:
> 
> * `pip install yq`
> * `go get github.com/prometheus/prometheus/cmd/promtool`

[...]

PR #2551, opened 2019-04-10
PrometheusRule admission webhooks

vsliouniaev:
Add admission webhooks to prometheus operator that prevent invalid rules from making it into the cluster and potentially breaking a prometheus.

There are two webhooks
- validating prevents pushing a badly formatted rule and provides feedback about each failing rule like promtool does.
- mutating adds an annotation to each rule indicating it was validated

Adding validation logic to the mutating webhook doesn't allow to provide the same kind of nicely formatted information to the user, so using two separate hooks. Since the mutating hook always runs first, and incorrect configuration could theoretically result in a rule annotated as validated to not actually be validated but I think that's rather unlikely.

More testing has shown that a single pod should really be enough to deal with validation but willing to add more functionality to run the webhooks externally

- fixes: #2273

Looking for guidance on how to proceed with this - I don't have a lot of experience in go, so any and all feedback is welcome


brancz:
This is really really awesome! I have a massive backlog of things to read/review. I haven't forgotten about this, it's really really awesome, thank you for working on this!

pkg/api/api.go
```diff
@@ -64,6 +64,7 @@ var (
 )
 
 func (api *API) Register(mux *http.ServeMux) {
+	mux.HandleFunc("healthz", ok)
```
brancz: did you mean this to be `/healthz` ?

cmd/operator/main.go
```diff
@@ -115,6 +133,8 @@ func init() {
 	flagset.StringVar(&cfg.TLSConfig.CertFile, "cert-file", "", " - NOT RECOMMENDED FOR PRODUCTION - Path to public TLS certificate file.")
 	flagset.StringVar(&cfg.TLSConfig.KeyFile, "key-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to private TLS certificate file.")
 	flagset.StringVar(&cfg.TLSConfig.CAFile, "ca-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to TLS CA file.")
+	flagset.StringVar(&cfg.HostTLSConfig.CertFile, "host-cert-file", "", "Path to certificate file to host web endpoint")
+	flagset.StringVar(&cfg.HostTLSConfig.KeyFile, "host-key-file", "", "Path to key file to host web endpoint")
```
brancz: I wonder if we should even handle this here or just ask people to put the [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) in front of this. Getting into TLS is hairy business.

pkg/api/api.go
vsliouniaev: Was going off the convention of using healthz in other projects. Is there a better convention to use?

brancz: I was just wondering if the preceding `/` is necessary or not

cmd/operator/main.go
vsliouniaev: I'm certain that I don't know all the cases you're thinking of, but my intention here was to provide the simplest way to do this set-up. I think that if someone wants to perform certificate rotation they can just restart the pod, since it doesn't have a requirement for being up 100% of the time.

A case I can think of that would become awkward is if the validation endpoint should be segregated from the rest of the Prometheus Operator API, with one requiring being served over TLS and the other not. 

There isn't anything preventing someone from using a proxy in front of prometheus operator to achieve this if they want to.

Could you suggest some other reasons for why this is a bad idea?

pkg/admission/admission.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	rules := &PrometheusRules{}
+	if err := json.Unmarshal(ar.Request.Object.Raw, rules); err != nil {
+		level.Info(a.logger).Log("msg", "Cannot unmarshal rules from spec", "err", err)
+		(*a.validationErrorsCounter).Inc()
+		return toAdmissionResponseFailure("Cannot unmarshal rules from spec", []error{err})
+	}
+
+	_, errors := rulefmt.Parse(rules.Spec.Raw)
```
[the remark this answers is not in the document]
vsliouniaev: I was under the impression that rules are added here in a completely 1:1 format with Prometheus. If not, then it would become pretty awkward to perform any kind of validation on these before pushing them to the cluster.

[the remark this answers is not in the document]
brancz: Yeah for now this is definitely fine.

cmd/operator/main.go
brancz: This would need the ability to allow TLS version, cipher suites, key rotation/reload, and many other things that are easy to get wrong. As a matter of fact we have already done those mistakes in kube-rbac-proxy.

> A case I can think of that would become awkward is if the validation endpoint should be segregated from the rest of the Prometheus Operator API, with one requiring being served over TLS and the other not.

Agreed, I think it should all be one. I don't see how that wouldn't be the case here.

> I think that if someone wants to perform certificate rotation they can just restart the pod, since it doesn't have a requirement for being up 100% of the time.

I think this dramatically changes with this being a blocking call to CRUD PrometheusRule objects.

brancz:
No worries! Thanks again for this work, it's really awesome!

vsliouniaev:
Looking at  kube-rbac-proxy, I didn't see an option to disable client-auth, so I went with `square/ghostunnel` that should be providing the same thing.

Do let me know if I'm off the mark here and I'll try to correct it

brancz:
Generally this looks really awesome, and I'm inclined to merge it as is. I'm just wondering what the intention for the mutating webhook adding the annotation is. Is it to identify rules that were created before the webhook was put in place?

brancz:
(this will also need a rebase and `go mod tidy`, but otherwise lgtm)

vsliouniaev:
> Generally this looks really awesome, and I'm inclined to merge it as is. I'm just wondering what the intention for the mutating webhook adding the annotation is. Is it to identify rules that were created before the webhook was put in place?

Yes, I think you suggested doing something like this here: https://github.com/coreos/prometheus-operator/issues/2273#issuecomment-469602658

brancz:
Ok, makes sense, just wanted to confirm. lgtm after rebase and go mod tidying

vsliouniaev:
I think this is correctly rebased and ready for a final check.

brancz:
This is really awesome! Thanks a ton!

vsliouniaev:
Thanks for taking the time to review it

brancz:
Sorry for taking so long, it's been a busy couple of weeks and then holidays and vacation. But so happy this is finally a reality! :tada:

brancz:
These are validation webhooks for Kubernetes.

omerlh:
Is there a reason this is not part of the mixin? I want to have the webhook as part of the operator deployment. Do I need to do something besides the webhook configuration?

brancz:
I think it’s largely the cert setup that is essentially not existent but I think it’s possible to configure skipping tls verification in which case I’d be happy to have it added!

omerlh:
On Kubernetes? I'm not sure... I think you must have TLS in order to create a webhook. We can add something like [this controller](https://github.com/open-policy-agent/gatekeeper/blob/50933a08be119de4cf5bdf1f088be523f24b6111/pkg/webhook/certs.go) that refresh the certs and update the webhook accordingly. What do you think?

vsliouniaev:
You could also use this: https://github.com/jet/kube-webhook-certgen, which may be even more lightweight

brancz:
I don’t think I would want to take any opinion on cert management by default. Can we just have the prometheus operator create a TLS server with a self signed cert if none is given?

omerlh:
Helm can generate certs each time it generates a template, with jsonnet it will be harder - but possible

brancz:
I meant the prometheus operator itself would do that just like every other kube component does. It would require code changes to the prometheus operator but I feel that’s the best thing to do here.

omerlh:
Let's file a new issue?

brancz:
Yes that sounds good

omerlh:
There is a similar issue already: coreos/prometheus-operator#2916

PR #2556, opened 2019-04-13
kube-prometheus: Add missing adapter to build snippet

jgustie:
Just noticed a difference between this file and the `example.jsonnet`.


metalmatze:
We just moved kube-prometheus back to https://github.com/coreos/kube-prometheus.
I'll move your PR there! :)

metalmatze:
I've recreated this PR with @jgustie as author. See this one: https://github.com/coreos/kube-prometheus/pull/67

PR #2560, opened 2019-04-16
*: Allow disabling limits on sidecars

brancz:
This PR introduces the ability to fully disable limits on the sidecar containers next to Prometheus and Alertmanager.

@mxinden @squat @s-urbaniak @metalmatze @paulfantom

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -407,6 +407,14 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 		})
 	}
 
+	resources := v1.ResourceRequirements{Limits: v1.ResourceList{}}
+	if config.ConfigReloaderCPU != "0" {
+		resources.Limits[v1.ResourceCPU] = resource.MustParse(config.ConfigReloaderCPU)
+	}
+	if config.ConfigReloaderMemory != "0" {
+		resources.Limits[v1.ResourceMemory] = resource.MustParse(config.ConfigReloaderMemory)
+	}
+
 	terminationGracePeriod := int64(0)
 	finalLabels := config.Labels.Merge(podLabels)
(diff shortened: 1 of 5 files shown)


pkg/alertmanager/statefulset.go
```diff
@@ -407,6 +407,16 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 		})
 	}
 
+	var resources v1.ResourceRequirements = v1.ResourceRequirements{}
+	if config.ConfigReloaderCPU != "0" && config.ConfigReloaderMemory != "0" {
```
squat: What if we want no CPU limit but do want a memory limit? I expect this may cause some confusion if both have to be set in order to be applied.

brancz: Agreed. Will change.

brancz: Changed to set them individually. Please take a look :)

pkg/prometheus/statefulset.go
```diff
@@ -628,12 +628,23 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
...
+		}
+
+		if c.ConfigReloaderCPU != "0" {
+			container.Resources.Limits[v1.ResourceCPU] = resource.MustParse(c.ConfigReloaderCPU)
+		}
+		if c.ConfigReloaderMemory != "0" {
+			container.Resources.Limits[v1.ResourceMemory] = resource.MustParse(c.ConfigReloaderMemory)
+		}
+
+		if c.ConfigReloaderCPU != "0" && c.ConfigReloaderMemory != "0" {
```
squat: we can safely eliminate this block now, right?

brancz: :facepalm:

brancz: fixed

PR #2567, opened 2019-04-22
Update bundle.yaml

castironclay:
Add specific callout for base image. This will help to show which images to modify when possibly using this manifest on different architectures. Having this commit present would have helped me when modifying this manifest for use with ARM. Please consider.


brancz:
This file is generated. You'll need to modify the jsonnet that generates this, but I feel maybe documenting the flags explicitly would be a better option here no?

castironclay:
I think that would be helpful.

PR #2568, opened 2019-04-23
pkg/alertmanager: Drop support for Alertmanager < v0.15.0

mxinden:
Alertmanager < v0.15 is not supported by the Prometheus upstream team.
In addition Alertmanager < v0.15.0 use a different clustering library.
Supporting two clustering libraries (< v0.15.0, >= v0.15.0) in the
Prometheus Operator prooved to be too much effort for the team.


brancz:
I think we can just drop it from the tests, to not block us. Best effort support as in leaving there what we have doesn't hurt I feel (if someone wants to go ahead and fix it that would be ok no?).

mxinden:
@brancz I would prefer to only offer what we support. I am not aware of a reason why someone is blocked on upgrading to >v0.15.0. Are you?

(Not a strong opinion. Your call :)

brancz:
The problem is that people may be on the old version of alertmanager and upgrading to the new version of prometheus operator and it silently failing. Unfortunately we don't maintain a compatibility matrix as we do for Prometheus so it's not clearly documented, but do you maybe just want to document in the compatibility document that any <v0.15.0 support is purely best effort and not guaranteed?

mxinden:
Resurrected version logic and added compatibility notice. Can you have another look @brancz?

Documentation/compatibility.md
```diff
@@ -44,3 +44,9 @@ The versions of Prometheus compatible to be run with the Prometheus Operator are
 * v2.6.1
 * v2.7.0
 * v2.7.1
+
+## Alertmanager
+
+Any version < v0.15.0 is supported on a best effort basis without end-to-end
```
metalmatze: I'd describe it the other way around. Something like:
> We only support Alertmanager v0.16 and above. Everything before that is on a best effort basis.

mxinden: Sounds good. Changed.

brancz:
This needs a re-gen and the compatibility doc is 100% generated so you need to modify the binary built to generate the compatibility docs.

Otherwise lgtm (e2e Tests seem to pass reliably so yay :) )

PR #2569, opened 2019-04-23
pkg/apis: update PodSecurityContext docs

squat:
The logic for handling PodSecurityContext has changed and the docs need
to catch up.

Fixes: https://github.com/coreos/prometheus-operator/commit/ce3b98df3e7f615d18273ff4ae3a08c7641a3339#commitcomment-33240896


brancz:
lgtm on green

PR #2579, opened 2019-04-30
Clarify docs on how Endpoints are discovered

redbaron:


Documentation/api.md
```diff
@@ -489,7 +489,7 @@ ServiceMonitorSpec contains specification parameters for a ServiceMonitor.
 | targetLabels | TargetLabels transfers labels on the Kubernetes Service onto the target. | []string | false |
 | podTargetLabels | PodTargetLabels transfers labels on the Kubernetes Pod onto the target. | []string | false |
 | endpoints | A list of endpoints allowed as part of this ServiceMonitor. | [][Endpoint](#endpoint) | true |
-| selector | Selector to select Endpoints objects. | [metav1.LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#labelselector-v1-meta) | true |
+| selector | Selector to select Service objects to get Endpoints from. | [metav1.LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#labelselector-v1-meta) | true |
```
brancz: It's the other way around. The selector selects `Endpoints` objects, and if there is a respective `Service` object to the `Endpoints` object then that will be used for additional metadata. The point is, this will also work for just `Endpoints` objects, the respective `Service` object is great for additional metadata but it will still work without.

redbaron:
That is not what I found in code, hence this doc clarification.

https://github.com/coreos/prometheus-operator/blob/410721d5ff5f5b7f6b4945288f23070f5efbec8f/pkg/prometheus/promcfg.go#L351

-------- Original Message --------

brancz:
Oh interesting, you're totally right. Sorry! Thanks for the addition!

brancz:
This needs a `make clean generate-in-docker` from the root of the repo to re-generate docs.

Documentation/api.md
brancz: this needs to be modified in the `pkg/apis/.../types.go` file, and then re-generated

pgier:
@redbaron Are you still interested in fixing this?  The doc change would need to be made in `pkg/apis/v1/types.go` and then run `make generate` to update the generated files.

PR #2585, opened 2019-05-05
[Fix] Corrected word spelling error

crazyacking:
@metalmatze

### DIFF
--- a/Documentation/user-guides/exposing-prometheus-and-alertmanager.md
+++ b/Documentation/user-guides/exposing-prometheus-and-alertmanager.md
@@ -179,7 +179,7 @@ This example was tested with the [NGINX Ingress Controller][nginx-ingress]. For 
 kubectl create -f https://raw.githubusercontent.com/kubernetes/ingress/master/examples/deployment/nginx/nginx-ingress-controller.yaml
 ```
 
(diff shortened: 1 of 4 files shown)


Documentation/user-guides/exposing-prometheus-and-alertmanager.md
```diff
@@ -179,7 +179,7 @@ This example was tested with the [NGINX Ingress Controller][nginx-ingress]. For
 kubectl create -f https://raw.githubusercontent.com/kubernetes/ingress/master/examples/deployment/nginx/nginx-ingress-controller.yaml
 ```
 
-> Be certain to evaluate available Ingress controllers for the specifics of your  production environment. The NGINX Ingress Controller may or may not be suitable. Consider other solutions, like HA Proxy, Træfɪk, GCE, or AWS.
+> Be certain to evaluate available Ingress controllers for the specifics of your  production environment. The NGINX Ingress Controller may or may not be suitable. Consider other solutions, like HA Proxy, Traefik, GCE, or AWS.
```
elisiano: I believe this could be either way (look at the logo on their web page: https://traefik.io/ )

crazyacking: Well, you are right, but the 'æ' symbol will show an exception in some operating systems, I recommend using the standard name :)

brancz:
Thanks!

PR #2591, opened 2019-05-08
Config reloader pod ordinal number

vsliouniaev:
This PR introduces a new `STATEFULSET_ORDINAL_NUMBER` environment variable in the prometheus config reloader, by regexing the numbers from the end of  `POD_NAME`. This allows a prometheus configuration with
```
externalLabels:
  prometheus_replica: $(STATEFULSET_ORDINAL_NUMBER)
```

The motivation for this is our current setup:
- Multiple prometheuses running in the same cluster, deployed into the same namespace
- All write to the same long-term storage database
- Remote-read from the database is used to execute rules, pulling data from multiple prometheuses, but requiring distinction between the replicas

### DIFF
--- a/cmd/prometheus-config-reloader/main_test.go
+++ b/cmd/prometheus-config-reloader/main_test.go
@@ -1,0 +1,31 @@
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+var cases = []struct {
+	in  string
+	out string
+}{
+	{"prometheus-0", "0"},
+	{"prometheus-1", "1"},
+	{"prometheus-10", "10"},
+	{"prometheus-10a", ""},
+	{"prometheus1", "1"},
+	{"pro-10-metheus", ""},
+}
+
+func TestCreateOrdinalEnvVar(t *testing.T) {
+	for _, tt := range cases {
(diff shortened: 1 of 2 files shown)


brancz:
Could you elaborate why the current set of external label replacement is not sufficient for this? For differentiation it already suffices for your use case as far as I can tell.

vsliouniaev:
I need to have a consistent external label value, that does not change with the name of the `prometheus` but is still able to differentiate between replicas. The key of the label itself doesn't matter, but the value needs to be consistent between separate operated prometheuses

cmd/prometheus-config-reloader/main_test.go
```diff
@@ -0,0 +1,31 @@
+package main
```
brancz: this needs a license header

brancz:
That makes sense. I'm ok with this.

brancz:
Thanks!

caarlos0:
are there any plans for a new release soon?

brancz:
There should be one happening some time next week.

caarlos0:
awesome, thanks!

PR #2595, opened 2019-05-09
Add prometheus-operator Service and Servicemonitor to Bundle

aditya-konarde:
Background:

Currently, the Service and ServiceMonitor for the Prometheus Operator are not added to the generated bundle.yaml. 

The issue with the Service and ServiceMonitor not being bundled is seen when a cluster has multiple namespaces with the prometheus-operator pods running, and you'd have to manually create a Service in front of each of the operator pods.

Changes in this PR:
- Add the Service and ServiceMonitor object already present in the jsonnet to the bundle

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -147,3 +147,41 @@ metadata:
     apps.kubernetes.io/version: v0.29.0
   name: prometheus-operator
   namespace: default
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    apps.kubernetes.io/component: controller
+    apps.kubernetes.io/name: prometheus-operator
+    apps.kubernetes.io/version: v0.29.0
+  name: prometheus-operator
+  namespace: default
+spec:
+  endpoints:
+  - honorLabels: true
+    port: http
+  selector:
+    matchLabels:
+      apps.kubernetes.io/component: controller
+      apps.kubernetes.io/name: prometheus-operator
+      apps.kubernetes.io/version: v0.29.0
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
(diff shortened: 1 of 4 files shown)


aditya-konarde:
@metalmatze

aditya-konarde:
Updated to remove the extra `ServiceMonitor` from the bundle.

brancz:
I’m not sure how good this is being in the bundle.yaml, as it’s going to cause people just starting with the Prometheus operator (who that file is aimed it) to get errors when applying as the servicemonitor CRD is not registered yet.

metalmatze:
Ugh, you're absolutely right. The Service should be included. I'll test this against a new kind cluster a couple of times to get a feeling for this. Worst case we remove the ServiceMonitor again.

aditya-konarde:
+1, good catch @brancz

brancz:
I'd say let's keep it in the jsonnet files, there it makes sense, even the sample directory, but not generate it into the `bundle.yaml`, that should be fairly easy to adapt to.

metalmatze:
Ok, never works. We need to revert the ServiceMonitor. 
@brancz , do you mean to comment out the line with ServiceMonitor? 

https://github.com/coreos/prometheus-operator/blob/0c603de008188d5ccbc20df5108d1ad0df186d89/hack/generate/prometheus-operator-rbac.jsonnet#L9

brancz:
I'm thinking, just filter out the service monitor file here: https://github.com/coreos/prometheus-operator/blob/0c603de008188d5ccbc20df5108d1ad0df186d89/hack/generate-bundle.sh#L9

metalmatze:
Created https://github.com/coreos/prometheus-operator/pull/2597 with a fix to exclude the ServiceMonitor file from the bundle.yaml.

PR #2597, opened 2019-05-10
Ignore ServiceMonitor in bundle.yaml

metalmatze:
Users installing the bundle.yaml most likely do not yet have the
ServiceMonitor CRD available in their cluster, therefore having this
object in the bundle.yaml would break them.

/cc @brancz @aditya-konarde


brancz:
lgtm :+1:

aditya-konarde:
I agree, thanks @metalmatze

PR #2600, opened 2019-05-13
Alertmanager: Use /-/healthy and /-/ready for probes .

desaintmartin:
Solves https://github.com/coreos/prometheus-operator/issues/2599

Exists in alertmanager since early 2018: https://github.com/prometheus/alertmanager/issues/991


brancz:
Nice thanks! lgtm on green :+1:

PR #2605, opened 2019-05-15
*: fix sig-apps recommended labels

abursavich:
Per https://github.com/coreos/prometheus-operator/pull/2427#issuecomment-491923629, this fixes the SIG Apps [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) from `apps.kubernetes.io/...` to `app.kubernetes.io/...` (app is singular).

It also updates the `app.kubernetes.io/version` label to `v0.30.0` in the places it was still `v0.29.0`.

It might should be noted that changing selectors requires recreating Deployments.

### DIFF
--- a/Documentation/rbac.md
+++ b/Documentation/rbac.md
@@ -19,9 +19,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   labels:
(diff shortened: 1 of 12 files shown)


brancz:
Thanks!

brancz:
lgtm

PR #2615, opened 2019-05-25
Prometheus selector CLI switch

redbaron:
It is often convenient to limit operator to particular Prometheus objects, comes handy in testing or in some namespace specific situations, where "local" Prometheus runs alongside cluster-wide Prometheus.

This adds `--prometheus-instance-selector` switch with a syntax same as `kubectl get -l`.

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -182,6 +182,10 @@ func New(conf Config, logger log.Logger) (*Operator, error) {
 		return nil, errors.Wrap(err, "instantiating monitoring client failed")
 	}
 
+	if _, err := labels.Parse(conf.PromSelector); err != nil {
+		return nil, errors.Wrap(err, "can not parse prometheus selector value")
+	}
+
 	kubeletObjectName := ""
 	kubeletObjectNamespace := ""
 	kubeletSyncEnabled := false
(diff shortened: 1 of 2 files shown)


pkg/prometheus/operator.go
```diff
@@ -213,9 +218,13 @@ func New(conf Config, logger log.Logger) (*Operator, error) {
 		listwatch.MultiNamespaceListerWatcher(c.config.Namespaces, func(namespace string) cache.ListerWatcher {
 			return &cache.ListWatch{
```
brancz: There is already a type for filtered list watches, we should re-use that: https://github.com/kubernetes/client-go/blob/7b18d6600f6b0022e31c46b46875beffd85cc71a/tools/cache/listwatch.go#L81

redbaron: Interesting.

it doesn't use typed clients, can it be a problem? 

also what would be a resource name? just "prometheus"? , but where apiversion is specified then?

brancz: hmm you're right .. doesn't look like we can easily wrap this, given those circumstances what you did here lgtm

neoludo:
Need the same for alertmanager, will create another Issue

Issue #2498, opened 2019-03-19
Old version of configmap-reload referenced


nfisher:
Looking at this line it appears the configmap-reload and digging into the quay repo it appears the image is 2y old. Implies the running binary contains some CVE's that have recently been patched in Go.
https://github.com/coreos/prometheus-operator/blob/master/bundle.yaml#L117

Should this be updated to the version below or a new build generated and re-pushed (e.g. 0.2.1-go112)?
https://hub.docker.com/r/jimmidyson/configmap-reload/tags

brancz:
Could you do a pull request replacing it with the latest upstream image? If tests just pass I'm happy to bump it, and remove us having to sync with upstream.

pgier:
Fixed by #2952

PR #2617, opened 2019-05-27
Update configmap-reload image from upstream

qw1mb0:
fix: #2498

### DIFF
--- a/Documentation/user-guides/getting-started.md
+++ b/Documentation/user-guides/getting-started.md
@@ -132,7 +132,7 @@ spec:
       - args:
         - --kubelet-service=kube-system/kubelet
         - --logtostderr=true
-        - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
+        - --config-reloader-image=jimmidyson/configmap-reload:v0.2.2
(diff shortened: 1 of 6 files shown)


brancz:
while I understand this, I'd prefer if we could move all of these into the prometheus operator repo, as it's very hard to reason about what is happening when this is an external container

qw1mb0:
How can i do this? I can make a separate directory with Dockerfile:
```
FROM jimmidyson/configmap-reload:v0.2.2
```

Then you will need to configure auto build? Or how to do it right?

brancz:
I'd be ok with that. We already have two dockerfiles, I'd be ok with adding this one as well. Or we could consolidate and merge the functionality from configmap-reload into the sidecar whose code is already in this repo.

mrueg:
@qw1mb0 are you still planning to work on this issue?

pgier:
Closing this because a similar update was merged in #2952

xUnholy:
Was this forgotten to be updated in the helm charts? https://github.com/helm/charts/blob/bd697f1f5b248fda127ffa0b3dfaa2fdac0f55fc/stable/prometheus-operator/values.yaml#L1268

PR #2620, opened 2019-05-29
Add doc re community helm chart

vsliouniaev:
Should at least add an explanation here about what the differences are, I am talking with the other maintainers about options around renaming the chart.


brancz:
I still think it should be renamed but this is at least better than the current situation.

PR #2623, opened 2019-06-01
Remove gossip from thanos sidecar v0.5.0-rc.0+

### LINKED ISSUE #2517: Thanos gossip deprecated
Thanos has deprecated the gossip clustering.  Wanted to make sure maintainers were aware of the impending change.

see:
https://github.com/improbable-eng/thanos/blob/master/docs/proposals/approved/201809_gossip-removal.md

https://github.com/improbable-eng/thanos/issues/734

abursavich:
In thanos v0.4, gossip was deprecated and disabled by default. With thanos v0.4, prometheus-operator users could (must?) migrate to gossip-less thanos while prometheus-operator continued to provide the extra (ignored) cluster args to the sidecar.

In thanos v0.5, the gossip code is removed along with its associated flags, which is now a fatal error to continue providing. 

This removes the sidecar's gossip/cluster args and the container's cluster port starting at thanos v0.5.0-rc.0.

Fixes #2517

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -680,27 +680,32 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 			thanosImage = *p.Spec.Thanos.Image
 		}
 
+		thanosVersion := semver.MustParse(strings.TrimPrefix(*p.Spec.Thanos.Version, "v"))
+		clusterSupport := thanosVersion.LT(semver.MustParse("0.5.0-rc.0"))
+
 		thanosArgs := []string{
 			"sidecar",
 			fmt.Sprintf("--prometheus.url=http://%s:9090%s", c.LocalHost, path.Clean(webRoutePrefix)),
 			fmt.Sprintf("--tsdb.path=%s", storageDir),
-			fmt.Sprintf("--cluster.address=[$(POD_IP)]:%d", 10900),
 			fmt.Sprintf("--grpc-address=[$(POD_IP)]:%d", 10901),
 		}
 
-		if p.Spec.Thanos.Peers != nil {
-			thanosArgs = append(thanosArgs, fmt.Sprintf("--cluster.peers=%s", *p.Spec.Thanos.Peers))
+		if clusterSupport {
+			thanosArgs = append(thanosArgs, fmt.Sprintf("--cluster.address=[$(POD_IP)]:%d", 10900))
+			if p.Spec.Thanos.Peers != nil {
+				thanosArgs = append(thanosArgs, fmt.Sprintf("--cluster.peers=%s", *p.Spec.Thanos.Peers))
+			}
+			if p.Spec.Thanos.ClusterAdvertiseAddress != nil {
+				thanosArgs = append(thanosArgs, fmt.Sprintf("--cluster.advertise-address=%s", *p.Spec.Thanos.ClusterAdvertiseAddress))
+			}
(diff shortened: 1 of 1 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -680,27 +680,32 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 			thanosImage = *p.Spec.Thanos.Image
 		}
 
+		thanosVersion := semver.MustParse(strings.TrimPrefix(*p.Spec.Thanos.Version, "v"))
+		clusterSupport := thanosVersion.LT(semver.MustParse("0.5.0-rc.0"))
```
brancz: I personally think we should just ignore this. We'll just support v0.5.0 and above from here. For exactly these reasons we marked this experimental.

abursavich: There have been a lot of changes to thanos, but prometheus-operator still supports v0.1.0, but I agree that we should limit backwards compatibility. My personal preference might be to support the latest and previous minor versions.

How about we move forward with this less controversial change, and I'll prepare a follow-up that changes the ThanosSpec (deprecated options) and removes all backwards compatibility? Then that change can be debated and move slower.

brancz: sounds good to me

pkg/prometheus/statefulset.go
```diff
@@ -680,27 +680,32 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
...
+		if clusterSupport {
+			thanosArgs = append(thanosArgs, fmt.Sprintf("--cluster.address=[$(POD_IP)]:%d", 10900))
+			if p.Spec.Thanos.Peers != nil {
+				thanosArgs = append(thanosArgs, fmt.Sprintf("--cluster.peers=%s", *p.Spec.Thanos.Peers))
+			}
+			if p.Spec.Thanos.ClusterAdvertiseAddress != nil {
+				thanosArgs = append(thanosArgs, fmt.Sprintf("--cluster.advertise-address=%s", *p.Spec.Thanos.ClusterAdvertiseAddress))
+			}
+			if p.Spec.Thanos.GrpcAdvertiseAddress != nil {
+				thanosArgs = append(thanosArgs, fmt.Sprintf("--grpc-advertise-address=%s", *p.Spec.Thanos.GrpcAdvertiseAddress))
```
brancz: we still need this even without clustering

abursavich: The grpc listen address (--grpc-address) is still provided unconditionally in the `thanosArgs` initialization, but advertising addresses are part of the gossip protocol (to tell other members how to contact you when joining).

--grpc-advertise-address removed: https://github.com/improbable-eng/thanos/commit/37f89adfd678c0e263a136da34aafe213e88bc24#diff-78f75f79fe3e933fafa9b97d9c6db643L49
--grpc-address remains: https://github.com/improbable-eng/thanos/commit/37f89adfd678c0e263a136da34aafe213e88bc24#diff-78f75f79fe3e933fafa9b97d9c6db643R19

brancz: :man_facepalming: you're totally right, sorry

brancz:
Thanks! lgtm :+1:

brancz:
While merging I remembered, can you bump the thanos dependency as well? We make use of some of the code, and I'd prefer it to be consistent.

abursavich:
> While merging I remembered, can you bump the thanos dependency as well? We make use of some of the code, and I'd prefer it to be consistent.

Will do.

PR #2629, opened 2019-06-05
thanos: remove deprecated features and support for older versions

abursavich:
Thanos support is explicitly flagged as experimental in the documentation which states that it may change significantly without deprecation notice or backward compatibility in any release.

Thanos configuration has changed significantly throughout its early iterations. Most notably, object store configuration was replaced in v0.2.0 and gossip was deprecated in v0.4.0 (removed in v0.5.0). Retaining complete backwards compatibility in prometheus-operator for all version of thanos has unnecessarily complicated its API and implementation.

Per https://github.com/coreos/prometheus-operator/pull/2623#discussion_r290654738, this drops support for older versions and simplifies the CRDs.


abursavich:
I'll bump the dependencies and DefaultThanosVersion to v0.5.0 once it's released.

cc @brancz

abursavich:
Fix for e2e tests coming soon... It's taking me longer than I would like to admit to get minikube running on my machine ;x

brancz:
Thank you for the awesome work! Lgtm :+1:

Issue #2630, opened 2019-06-06
Support hostNetwork api in prometheus and alertmanager crds


huangjiuyuan:
**What is missing?**
Support hostNetwork api in prometheus and alertmanager crds.

**Why do we need it?**
In our large scale Kubernetes cluster with VPC network, Prometheus need to monitor external Etcd clusters. However, Prometheus could not communicate with external Etcd clusters due to some restrictions of VPC network. We have a work-around by using host network for Prometheus.

metalmatze:
Can you explain how moving Prometheus to the hostNetwork mitigates your problem? Shouldn't a change in etcd solve your problems instead?
Not sure we really need this in the Prometheus Operator, but I might be missing something right now.

brancz:
If this is something you need to do, then I'd prefer if you deploy a proxy that has hostNetwork that forwards the request to etcd instead of exposing a potentially privilege escalation enabling feature in the prometheus operator.

huangjiuyuan:
@metalmatze  We use two VPC networks for a cluster. One for Pod IP and one for Node IP. The VPC IP assigned to Pod cannot connect to external etcd clusters, yet the VPC IP assigned to Node (the IP bound to APIServer, Controller Manager, Scheduler, Kubelet, etc.) can because we do not allow our users to visit any external management services outside Kubernetes clusters for securoty reasons. We have admisson control webhooks to forbidden any privileged Pod creations by common users. Components such as Prometheus, Servicebrokers, Node Problem Detector, ELK and deployed on dedicated machines and maintained by us. Moving Prometheus to the hostNetwork allows Prometheus to use the Node VPC IP and connect to external Etcd clusters.

huangjiuyuan:
@brancz Thank you for your advice. However, introducing a proxy would increase the complexity and using hostNetwork for Prometheus does not compromise the security in my opinion. I have noticed that securityContext is supported in the API document. Setting securityContext and setting hostNetwork both escalate privilege level of Prometheus operator.

fyery-chen:
@metalmatze We also need this feature, so when it can be available?

silenceper:
@metalmatze  We also need this feature too.

We customized the network in Kubernetes to isolate the pod network from the host network.
And we want prometheus to use the host network.

Haleygo:
@metalmatze ,We also need this feature too.
[...]

PR #2631, opened 2019-06-06
Support hostNetwork api in prometheus and alertmanager crds

huangjiuyuan:
Support hostNetwork api in prometheus and alertmanager crds.

Related issue: https://github.com/coreos/prometheus-operator/issues/2630


metalmatze:
We should first discuss if we need this in #2630.

3Golds:
When is this feature available?

Or there are other ways to support 'hostNetwork: true'?

simonpasquier:
superseded by https://github.com/prometheus-operator/prometheus-operator/pull/5010

PR #2638, opened 2019-06-11
ServicemonitorKey

auhlig:
Use the ServiceMonitor key as the default job name as proposed by @brancz in https://github.com/coreos/prometheus-operator/pull/2566#discussion_r291619343.

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -492,9 +492,8 @@ func (cg *configGenerator) generateServiceMonitorConfig(version semver.Version, 
 	// as a fallback the port number.
 
 	relabelings = append(relabelings, yaml.MapSlice{
-		{Key: "source_labels", Value: []string{"__meta_kubernetes_service_name"}},
 		{Key: "target_label", Value: "job"},
-		{Key: "replacement", Value: "${1}"},
+		{Key: "replacement", Value: fmt.Sprintf("%s/%s", m.GetNamespace(), m.GetName())},
 	})
 	if m.Spec.JobLabel != "" {
 		relabelings = append(relabelings, yaml.MapSlice{
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -497,10 +497,8 @@ scrape_configs:
     target_label: env
     regex: (.+)
     replacement: ${1}
-  - source_labels:
-    - __meta_kubernetes_service_name
-    target_label: job
-    replacement: ${1}
+  - target_label: job
+    replacement: default/testservicemonitor1
   - target_label: endpoint
     replacement: web
 alerting:
@@ -618,10 +616,8 @@ scrape_configs:
     target_label: env
     regex: (.+)
     replacement: ${1}
-  - source_labels:
-    - __meta_kubernetes_service_name
-    target_label: job
-    replacement: ${1}
+  - target_label: job
+    replacement: default/testservicemonitor1
   - target_label: endpoint
     replacement: web
 alerting:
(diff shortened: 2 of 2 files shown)


brancz:
I think here the service key actually makes a lot of sense, it's a better and more granular grouping than the ServiceMonitor. The thing that I did mention we should have is that the `container` label should always be relabelled onto the target.

primeroz:
Is there any hope in getting something similar to this merged ? 

I have been looking for this so that i could match the `job` label on my metrics to what my `promtail loki` set ( which is very similar to what podmonitor does, so `namespace/name` ) 

I ended up for now having a label on my service with the format i need and set it through `jobLabel` but of course is much higher maintenance than a default like this.

I guess this being a breaking change ( and forcing anyone relying on the previous behavior to set a specific jobLabel) could be an issue. 
Would we consider setting this behind either a control flag or a control key in the `serviceMonitor` spec ?

brancz:
@primeroz arbitrary job labels can already be relabeled onto any target via relabeling rules. You can also change the loki config btw :)

primeroz:
> You can also change the loki config btw :)

@brancz Fair enough :) 

I guess this is really just my use case but my problem is that `promtail` does not really have an operator ( yet ? ) so i don't have the relabeling flexibility i get from the `prometheus-operator`  so i have a way more static config that is sort of like

1. If `name label is present on the pod` -> `job=namespace/<NAME LABEL>`
2. If `name label is not present on the pod but app label is` -> `job=namespace/<app LABEL>`
3. If neither `name or app label is not present on the pod` -> `job=<SUPER CATCH ALL CASE>`

The `namespace/<LABEL>` is actually needed because otherwise i could have multiple pods with the same label in different namespaces all grouped in the same job ( which would make it hard in grafana to use the job label to _jump_ from _metrics_ to _logs_ 

I tried using `jobLabel` in the `serviceMonitor` but unfortunately a label can't contain `/` so that was not very good.

I did end up adding the following bit to all endpoints in all my servicemonitor 

```
      relabelings:
        - sourceLabels:
            - __meta_kubernetes_namespace
            - __meta_kubernetes_pod_label_app_kubernetes_io_name
          targetLabel: job
          separator: /
```

which is doing the job ... but is very hard maintenance. I have loads of serviceMonitor in different repos ... I know, our fault :)

 Since i saw that the `podMonitor` does indeed set the default job label to `namespace/CRDName` i thought that maybe the `serviceMonitor` could do that too ... and then i found this MR :) 

But is fine if something like this is not planned to go through ... i just wanted to ping on this MR since it would be exactly what i needed ! 

thanks

brancz:
hmm I was under the impression that if you don't set `jobLabel` on the ServiceMonitor you get `namespace/<ServiceMonitorName>` is that not the case?

primeroz:
For the _serviceMonitor` at the moment the default label is the value `__meta_kubernetes_service_name` see https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/prometheus/promcfg.go#L1056

which essentially turns into 
```
  - source_labels: [__meta_kubernetes_service_name]
    separator: ;
    regex: (.*)
    target_label: job
    replacement: ${1}
    action: replace
```

while podMonitor does https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/prometheus/promcfg.go#L597 which is `fmt.Sprintf("%s/%s", m.GetNamespace(), m.GetName())` 

In my ideal world the second part would be configurable using the `jobLabels` ... :)

primeroz:
> while podMonitor does https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/prometheus/promcfg.go#L597 which is fmt.Sprintf("%s/%s", m.GetNamespace(), m.GetName())
> In my ideal world the second part would be configurable using the jobLabels ... :)

I can work on this if we think is something we could merge in ... I assume this would all have to be behind some sort of control flag for the operator to avoid breaking everyone's `Rules` and `Dashboard` with an operator update

PR #2644, opened 2019-06-14
alertmanager/statefulset: move gossip port to 9094

### LINKED ISSUE #1574: Wrong port for Alertmanager >=v0.15.0

mickael-carl:
Alertmanager default port for clustering is now 9094 over both tcp and udp. This patch brings that change to the operator.
More or less an up-to-date version of #1601, as rebasing that one would have been painful.

Fixes #1574

### DIFF
--- a/Documentation/network-policies.md
+++ b/Documentation/network-policies.md
@@ -5,7 +5,7 @@
 
 # Network policies
 
-[Network policies](https://kubernetes.io/docs/user-guide/networkpolicies/) allow you easily restrict the ingress traffic between pods using [k8s labels](https://kubernetes.io/docs/user-guide/labels/). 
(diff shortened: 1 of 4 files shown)


brancz:
Sorry for taking so long, thanks!

lgtm :+1:

mickael-carl:
Thanks! Probably #1601 can be closed as well then?

PR #2652, opened 2019-06-19
Bump manifest version

brancz:
I'm not exactly sure why, I went down to re-generating ksonnet.4 but for some reason `withReadOnlyRootFilesystem` is not available anymore. In order to move forward I removed it for now.

@paulfantom @squat @metalmatze @s-urbaniak

### DIFF
--- a/example/rbac/prometheus-operator/prometheus-operator-deployment.yaml
+++ b/example/rbac/prometheus-operator/prometheus-operator-deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
@@ -40,7 +40,6 @@ spec:
             memory: 100Mi
         securityContext:
           allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: true
       nodeSelector:
         beta.kubernetes.io/os: linux
       securityContext:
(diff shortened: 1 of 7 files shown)


example/rbac/prometheus-operator/prometheus-operator-deployment.yaml
```diff
@@ -40,7 +40,6 @@ spec:
             memory: 100Mi
         securityContext:
           allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: true
```
s-urbaniak: According to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#securitycontext-v1-core, `readOnlyRootFilesystem` defaults to `false`.

Aren't we lowering the security bar here? :thinking:

s-urbaniak:
that's a bummer that this setting disappeared :thinking: just one nit regarding `readOnlyRootFilesystem` else :+1:

example/rbac/prometheus-operator/prometheus-operator-deployment.yaml
brancz: We’re still enforcing an entirely different set of things through PSPs/SCCs. I do agree that I wish this was available but I feel I’ve done the due diligence on figuring out why this disappeared, but I can’t figure it out. To be able to move forward for now I removed it.

PR #2671, opened 2019-07-13
feat: Make Prometheus port name configurable

### LINKED ISSUE #2401: Prometheus service/pod port name, change to "http"
It is not possible at the moment to modify the port name that the Operator is creating, for example by default (hard coded) it is set to "web" on port 9090, this cause some conflicts if one is trying to use Istio as a service mesh.
Best practice is to use  port names starting with "http-" such as "http-prometheus" to avoid conflicts.

mikebryant:
Allow changing the port name to http, to be compatible with istio's port naming requirements

Fixes #2401


mikebryant:
I'm confused as to why the tests fail. Help would be appreciated :)

brancz:
```
prometheus_test.go:1459: waiting for targets timed out. 0 of 1 targets found. the server is currently unable to handle the request (get services prometheus-operated:web)
```

That's probably the problem. I realize it'll be a little disruptive, but why not change the default? `http` is arguably better in every way than `web`.

mikebryant:
I didn't change the default because you said not to here: https://github.com/coreos/prometheus-operator/issues/2401#issuecomment-464777473

I'm confused as to why the e2e test can't get prometheus-operated:web - in theory my changes should have had no effect by keeping the default. I guess I have a mistake somewhere? Not sure where though...

mikebryant:
Aha, I've figured out the mistake I was making - the Service part doesn't go through the same bit where the other defaults are set

mikebryant:
Tests currently appear to be failing due to Travis-CI instability

Documentation/api.md
```diff
@@ -143,6 +143,7 @@ AlertmanagerSpec is a specification of the desired behavior of the Alertmanager
 | containers | Containers allows injecting additional containers. This is meant to allow adding an authentication proxy to an Alertmanager pod. | []v1.Container | false |
 | priorityClassName | Priority class assigned to the Pods | string | false |
 | additionalPeers | AdditionalPeers allows injecting a set of additional Alertmanagers to peer with to form a highly available cluster. | []string | false |
+| portName | Port name used for the pods and governing service This defaults to web | string | false |
```
brancz: `This` -> `this`

mikebryant: Ah, this is from a two line comment in the api definition. Will fix now

mikebryant: Fixed! :)

brancz:
Thanks!

PR #2679, opened 2019-07-18
feat: Change governing service port names to match istio standard

mikebryant:
Istio requires that service port names follow a convention, and in particular that if two services point at the same pod (e.g. if a user makes a service with a ClusterIP in addition to the governing service) that the name's have the same protocol. By not changing the pod port name this preserves backwards compatibility with existing users services

Fixes https://github.com/coreos/prometheus-operator/issues/2401


brancz:
Can you explain how this relates to #2671 ?

mikebryant:
> Can you explain how this relates to #2671 ?

Alternative approach, instead of making it configurable just change the default

PR #2683, opened 2019-07-22
Provide option to turn on WAL compression

csmarchbanks:
This allows users to enable wal compression (or disable if compression becomes default in a later release).


brancz:
Awesome! Thanks! lgtm :+1:

PR #2685, opened 2019-07-24
scripts: Bump minikube version

LiliC:
All the tests are failing, as suggested in https://github.com/coreos/prometheus-operator/pull/2684#issuecomment-514260866 bumping minikube version should do the trick.


LiliC:
I didn't bump the kubernetes version in this PR, but might be good to do at some point, as it's still using version 1.13.4.

cc @brancz PTAL

s-urbaniak:
lgtm :+1:

PR #2692, opened 2019-07-26
add extra notes for crd-apigroup flag

RiRa12621:
this fixes https://github.com/coreos/prometheus-operator/issues/2691


brancz:
I think this actually doesn’t work anymore since @s-urbaniak changed the api generation to the upstream generators. I think this should probably just be removed or marked deprecated.

RiRa12621:
ACK, closing

PR #2705, opened 2019-08-06
add flag watch interval for prometheus config reloader

huanggze:
The original logic for watching config file changes is via a regular [ticker and fsnotify watcher](https://github.com/coreos/prometheus-operator/blob/master/vendor/github.com/improbable-eng/thanos/pkg/reloader/reloader.go#L164-L165). I oberseved a long waiting time for reloader being triggered (might up to 2 min)

It would be better if we can customize the time interval of ticker to allow more frequent cfg file re-reading, since now it defaults to [3 min](https://github.com/coreos/prometheus-operator/blob/master/vendor/github.com/improbable-eng/thanos/pkg/reloader/reloader.go#L108).

Signed-off-by: huanggze <loganhuang@yunify.com>

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -133,6 +133,7 @@ func init() {
 	// the Prometheus Operator version if no Prometheus config reloader image is
 	// specified.
 	flagset.StringVar(&cfg.PrometheusConfigReloaderImage, "prometheus-config-reloader", fmt.Sprintf("quay.io/coreos/prometheus-config-reloader:v%v", version.Version), "Prometheus config reloader image")
+	flagset.StringVar(&cfg.PrometheusConfigReloaderWatchInterval, "prometheus-config-reloader-interval", "3m", "Prometheus config file watch interval")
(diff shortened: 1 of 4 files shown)


huanggze:
The original logic for watching config file changes is via a regular [ticker and fsnotify watcher](https://github.com/coreos/prometheus-operator/blob/master/vendor/github.com/improbable-eng/thanos/pkg/reloader/reloader.go#L164-L165). I oberseved a long waiting time for reloader being triggered (might up to 2 min)

It would be better if we can customize the time interval of ticker to allow more frequently reloading, since now it is default to [3 min](https://github.com/coreos/prometheus-operator/blob/master/vendor/github.com/improbable-eng/thanos/pkg/reloader/reloader.go#L108).

brancz:
Does this actually improve anything for you? The long reload time is not actually coming from the reload waiting too long, it's from Kubernetes re-mounting the secret only every couple of minutes as far as I know. So this shouldn't actually change anything.

PR #2718, opened 2019-08-14
generate client .

Ghostbaby:


LiliC:
@Ghostbaby Can you explain why this is needed. Thanks!

Ghostbaby:
@LiliC Ok, when other operators want to call the client to create a ServiceMonitor resource, they need to register the client.
`// newReconciler returns a new reconcile.Reconciler
func newReconciler(mgr manager.Manager) reconcile.Reconciler {
	cli, err := clientcmd.NewClient()
	if err != nil {
		return nil
	}

	mcli, err := monitoringclient.NewRegistry(mgr)
	if err != nil {
		return nil
	}

	return &ReconcileMongodbCluster{Client: mgr.GetClient(), scheme: mgr.GetScheme(), recorder: mgr.GetRecorder("mongodbcluster-controller"), clientcmd: cli, poclient: mcli}
}`

pkg/client/generic_client.go
```diff
@@ -0,0 +1,45 @@
...
+	MonitoringClient monitoringclientset.Interface
+}
+
+// NewForConfig creates a new Clientset for the given config.
+func newForConfig(c *rest.Config) (*GenericClientset, error) {
+	kubeClient, err := kubeclientset.NewForConfig(c)
+	if err != nil {
+		return nil, err
+	}
+	monitoringClient, err := monitoringclientset.NewForConfig(c)
```
LiliC: This is essentially all you need to connect to the monitoring client, so I don't think this helper function is really needed.

Ghostbaby:
ok

LiliC:
@Ghostbaby so you don't spend time next time, maybe open an issue next time so we can discuss, and then we are happy if you work on the implementation as well! :) Always happy to have contributions!

PR #2722, opened 2019-08-20
pkg/listwatch: fix listwatch wrapping regression

squat:
Whenever a multi-listwatch contains only one listwatch, it
would simply use the underlying listwatch without any wrapper. This
improves performance and eliminates unnecessary go routines. This commit
fixes a regression [0] where a denylist-listwatch would wrap underlying
listwatches even if the denylist was empty.

[0] https://github.com/coreos/prometheus-operator/pull/2710/files#diff-b33bee110e95fb2f22d9da56b1630d89R68-R74

cc @s-urbaniak @brancz @LiliC @metalmatze


brancz:
lgtm :+1:

s-urbaniak:
good catch, thanks :+1:

PR #2723, opened 2019-08-20
pkg/listwatch: ensure namespace objects are denied

squat:
The original implementation of the denylist-listwatch does not properly
filter out namespace objects as they are not namespaced. Nevertheless,
the denylist-listwatch is applied to the multi-namespace listwatch. This
means that changes to blacklisted namespaces would be handled by
unsuspecting informers.

This commit ensures that namespaces are correctly filtered.

cc @s-urbaniak @brancz @LiliC @metalmatze

### DIFF
--- a/pkg/listwatch/namespace_denylist.go
+++ b/pkg/listwatch/namespace_denylist.go
@@ -20,6 +20,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


brancz:
good catch! lgtm :+1:

pkg/listwatch/namespace_denylist.go
```diff
@@ -179,3 +179,10 @@ func newDenylistWatch(l log.Logger, denylist map[string]struct{}, next watch.Int
 
 	return proxy
 }
+
+func getNamespace(obj metav1.Object) string {
```
LiliC: Nit: Could we just add a short function comment for this, mainly mentioning that this ensures filtering out `Namespace` objects correctly. Or rename to `getNamespaceName()` as we are not getting the Namespace object. WDYT?

pkg/listwatch/namespace_denylist_test.go
```diff
@@ -92,8 +160,7 @@ func TestDenylistList(t *testing.T) {
 					t.Error(err)
 					return
 				}
-
-				got = append(got, acc.GetNamespace())
+				got = append(got, getNamespace(acc))
```
LiliC: Bit below https://github.com/coreos/prometheus-operator/blob/28db7c57b645a9f4827e5922a558765ac6b33a3c/pkg/listwatch/namespace_denylist_test.go#L179 we are using the `acc.GetNamespace` should we use the `getNamespace()` there as well?

pkg/listwatch/namespace_denylist.go
squat: added a comment (:

s-urbaniak:
nice, thanks!

PR #2724, opened 2019-08-20
pkg/listwatch: allowed/denied>whitelist/blacklist

squat:
This commit renames the allowed and denied names to the more common
whitelist and blacklist, respectively. It does not affect the logic at
all, but rather is purely cosmetic.

cc @LiliC @brancz

### DIFF
--- a/pkg/listwatch/blacklist.go
+++ b/pkg/listwatch/blacklist.go
@@ -28,39 +28,39 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-// denylistListerWatcher implements cache.ListerWatcher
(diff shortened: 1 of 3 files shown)


LiliC:
It was renamed as per this comment in the original PR https://github.com/coreos/prometheus-operator/pull/2710#issuecomment-519531051

> We're trying to move away from "blacklist" and "whitelist", I'd prefer if we went with either just as the flags say "NoNamespaces" and "Namespaces" or "AllowList" and "DenyList".

squat:
sounds fine to me. this needs a clean up as right now it is quite clunky and feels inconsistent.

PR #2728, opened 2019-08-21
Add ListenLocal option for Thanos spec to listen on local loopback

### LINKED ISSUE #2435: Thanos not working with istio
Tried running thanos with promethues operator . We are also using istio in our k8 cluster . The issue i am facing is with istio side car thanos is not working . 

Reason :  the argument generated by prometheus-operator for thanos sidecar is 
--grpc-address=[$(POD_IP)]:10901

But istio forwards traffic to localhost . Due to which istio is not able to forward traffic to thanos side car 

Is there any way to solve this ?

gjtempleton:
Resolves #2435

Solution basically copied wholesale from #1051

Adds a test to ensure both flags are successfully switched to local loopback listening on setting the option.


pkg/prometheus/statefulset_test.go
```diff
@@ -848,3 +848,30 @@ func TestWALCompression(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unexpected error while making StatefulSet: %v", err)
+	}
+	foundGrpcFlag := false
+	foundHTTPFlag := false
+	for _, flag := range sset.Spec.Template.Spec.Containers[2].Args {
+		if flag == "--grpc-address=127.0.0.1:10901" {
+			foundGrpcFlag = true
+		}
+		if flag == "--http-address=127.0.0.1:10902" {
```
metalmatze: I think this still needs to be on POD_IP for Prometheus to be able to scrape it?

LiliC:
cc @bwplotka do you mind having a look as well, thanks!

pkg/prometheus/statefulset.go
```diff
@@ -701,17 +701,21 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		if p.Spec.Thanos.Image != nil && *p.Spec.Thanos.Image != "" {
 			thanosImage = *p.Spec.Thanos.Image
 		}
+		thanosArgs := []string{
```
bwplotka: Small nit: 

Why not just do:
```
                         Args: []string{	
				"sidecar",	
				fmt.Sprintf("--prometheus.url=http://%s:9090%s", c.LocalHost, path.Clean(webRoutePrefix)),	
				fmt.Sprintf("--tsdb.path=%s", storageDir),	
				fmt.Sprintf("--grpc-address=%s:10901", bindAddress),	
				fmt.Sprintf("--http-address=%s:10902",	bindAddress),
			},
```
and find proper `bindAddress`  beforehand? Might more consistent with what we had before and as a result bit more readable (:

LiliC:
> Plus, can we add CHANGELOG on top of this? (:

@bwplotka We create the CHANGELOG entries while doing the [release](https://github.com/coreos/prometheus-operator/blob/master/RELEASE.md#prepare-your-release). :) So no need to do one here.

pkg/prometheus/statefulset.go
gjtempleton: Thanks, good point, reads far better now!

In terms of `0.0.0.0` vs `127.0.0.1` I swayed towards the more limited change, and it also matched the existing behaviour in #1051 - as far as I'm aware there's no limit/issue caused by listening on `0.0.0.0` instead, and it does match the default Thanos behaviour. If we decide to go with that instead I'd like to rename the variable to be more accurate.

LiliC:
As @bwplotka is on PTO, can you have another look at this @metalmatze so we can get this merged. :)

pkg/prometheus/statefulset.go
metalmatze: Then let's go with `127.0.0.1`. We just won't be able to have any probes on those ports.
But there might be other ways to add them for those containers: https://github.com/coreos/prometheus-operator/pull/2763

metalmatze:
This PR needs a rebase. Then lgtm

gjtempleton:
Thanks for the reviews and feedback. Now rebased.

bwplotka:
Thanks!

PR #2729, opened 2019-08-23
WIP: pkg/alertmanager: Add ConfigMapsMounts

LiliC:
Currently we always mount the ConfigMap in the location `"/etc/alertmanager/configmaps/"`, this PR adds specifying the mount location along with the ConfigMap. 

ConfigMapsMounts is a list of ConfigMaps in the same namespace as the Alertmanager object, which shall be mounted into the Alertmanager Pods. The ConfigMaps are mounted into mounted name specified.

Opening this PR early to see if this feature is possible. Would still like to add tests for this before merging. :) 

cc @brancz @s-urbaniak


brancz:
I'm thinking that if we do this, we're better off just fully introducing `Volumes`.

brancz:
as a matter of fact we already have half of that PR from the community: https://github.com/coreos/prometheus-operator/pull/2734

LiliC:
@brancz that PR is to add it to the Prometheus CR, we want to add it to the alertmanager one.

LiliC:
Closing in favour of https://github.com/coreos/prometheus-operator/pull/2737

PR #2755, opened 2019-09-11
Add VolumeMounts to Alertmanager

LiliC:
Because the Alertmanager statefulset already contains the container named `alertmanager` and we just do an append instead of a merge when creating the Statefulset, we then end up with two of the same named containers, so adding VolumeMounts does not work.

Note: I want to add this for prometheus resource, but its not urgent, as there the above problem is not present, as we do a merge of the containers, not just an append.

cc @s-urbaniak


s-urbaniak:
looks great! do you think we can add an e2e test for this?

s-urbaniak:
(which we can also address as a follow-up)

s-urbaniak:
lgtm :+1:

LiliC:
> looks great! do you think we can add an e2e test for this?

Sure! Not too familiar with the e2e tests in this repo, what kind of test case did you have in mind?

PR #2761, opened 2019-09-17
prometheus-config-reloader: Remove unneeded Ownership change

mrueg:
As far as I can see, changing the ownership of the p-c-m binary should be unnessecary.
Droping the chown removes a potential attack vector as a non-privileged user
might be able alter content of the binary during runtime.


brancz:
lgtm thanks!

PR #2763, opened 2019-09-17
Add exec probes with wget localhost:9090 to Prometheus if listenLocal

metalmatze:
We still want to probe Prometheus, even if it's only listening on `localhost:9090`. The only real option here is to run `wget -q localhost:9090` inside the container, so we actually have the right loopback.

/cc @aditya-konarde @LiliC @kakkoyun @s-urbaniak @paulfantom

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -596,6 +596,8 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		fmt.Sprintf("--config-envsubst-file=%s", path.Join(confOutDir, configEnvsubstFilename)),
 	}
 
+	const localProbe = `if [ -x "$(command -v curl)" ]; then curl %s; elif [ -x "$(command -v wget)" ]; then wget -q %s; else exit 1; fi`
+
 	var livenessProbeHandler v1.Handler
 	var readinessProbeHandler v1.Handler
 	var livenessFailureThreshold int32
@@ -603,11 +605,12 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		{
 			healthyPath := path.Clean(webRoutePrefix + "/-/healthy")
 			if p.Spec.ListenLocal {
+				localHealthyPath := fmt.Sprintf("http://localhost:9090%s", healthyPath)
 				livenessProbeHandler.Exec = &v1.ExecAction{
 					Command: []string{
-						"wget",
-						"-q",
-						fmt.Sprintf("http://localhost:9090%s", healthyPath),
+						"sh",
+						"-c",
+						fmt.Sprintf(localProbe, localHealthyPath, localHealthyPath),
 					},
 				}
 			} else {
@@ -620,13 +623,15 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		{
 			readyPath := path.Clean(webRoutePrefix + "/-/ready")
 			if p.Spec.ListenLocal {
+				localReadyPath := fmt.Sprintf("http://localhost:9090%s", readyPath)
 				readinessProbeHandler.Exec = &v1.ExecAction{
 					Command: []string{
-						"wget",
-						"-q",
-						fmt.Sprintf("http://localhost:9090%s", readyPath),
+						"sh",
+						"-c",
+						fmt.Sprintf(localProbe, localReadyPath, localReadyPath),
 					},
 				}
+
 			} else {
 				readinessProbeHandler.HTTPGet = &v1.HTTPGetAction{
 					Path: readyPath,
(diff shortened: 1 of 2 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -600,19 +600,43 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
...
-			HTTPGet: &v1.HTTPGetAction{
-				Path: path.Clean(webRoutePrefix + "/-/healthy"),
-				Port: intstr.FromString(p.Spec.PortName),
-			},
+		{
+			healthyPath := path.Clean(webRoutePrefix + "/-/healthy")
+			if p.Spec.ListenLocal {
+				livenessProbeHandler.Exec = &v1.ExecAction{
+					Command: []string{
+						"wget",
```
s-urbaniak: nit: how can we make sure this is available in all images?

pkg/prometheus/statefulset.go
```diff
@@ -626,21 +650,17 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		livenessFailureThreshold = 60
 	}
 
-	var livenessProbe *v1.Probe
-	var readinessProbe *v1.Probe
-	if !p.Spec.ListenLocal {
```
LiliC: Should we be removing this? I would maybe instead of considering adding a new field, wdyt? As you can't opt-out of this now if I understand correctly.

pkg/prometheus/statefulset.go
metalmatze: It is a posix thing. I'm afraid we can't really make sure. However, even the busybox based Prometheus image had wget and not curl.
Unfournately, I just saw that it's exactly the other way around for RHEL based images... 

:roll_eyes:

pkg/prometheus/statefulset.go
metalmatze: Not sure I understand your question. This if basically moved up a bit to the place where we actually create those handlers. :slightly_smiling_face:

pkg/prometheus/statefulset.go
metalmatze: Discussing offline, we need to somehow check for either of wget or curl being available and then do the probe. We should fail if none available.

pkg/prometheus/statefulset.go
LiliC: discussed offline :)

pkg/prometheus/statefulset.go
LiliC: Not sure if we should fail, thats not backwards compatible. Also failing implies the liveness probe failed. hmm I guess curl and wget should cover 90% of usecases? For the rest can't we add a way to disable the adding of liveness probe?

metalmatze: I'd argue that we would want to fail hard in this case, rather than silently not probe at all.

s-urbaniak: crazy idea: can we implement a "wget"/"curl" request functionality in the prometheus operator binary itself as a separate subcommand? Shipping the operator binary is the lowest common denominator anyways.

metalmatze: Great idea :flushed: :smile: :blush: 
Shouldn't be too hard indeed.

I'll give it a shot.

metalmatze: Wait. Actually that doesn't work. We need something inside the Prometheus Container, not the Operator's Container.

s-urbaniak: ohhh, ok, right. maybe we can persuade upstream then? :) (not sure about this)

LiliC: I think we are fine with checking for wget first, if not there, using curl, and failing. Is there any image we know of that doesn't have either one of those?

metalmatze:
So. I have come up with a lovely "small" shell script that checks for `curl` or `wget` uses any of those or fails entirely if none is available.

Excerpt from prometheus-k8s-1 applied YAML
```yaml
...
    livenessProbe:
      exec:
        command:
        - sh
        - -c
        - if [ -x "$(command -v curl)" ]; then curl http://localhost:9090/-/healthy;
          elif [ -x "$(command -v wget)" ]; then wget -q http://localhost:9090/-/healthy;
          else exit 1; fi
      failureThreshold: 6
      periodSeconds: 5
      successThreshold: 1
      timeoutSeconds: 3
    readinessProbe:
      exec:
        command:
        - sh
        - -c
        - if [ -x "$(command -v curl)" ]; then curl http://localhost:9090/-/ready;
          elif [ -x "$(command -v wget)" ]; then wget -q http://localhost:9090/-/ready;
          else exit 1; fi
      failureThreshold: 120
      periodSeconds: 5
      successThreshold: 1
      timeoutSeconds: 3
...
```

brancz:
lgtm

PR #2773, opened 2019-09-23
Fix examples

maxlim0:
- fix ServiceAccount name according to bundle.yaml

### DIFF
--- a/Documentation/additional-scrape-config.md
+++ b/Documentation/additional-scrape-config.md
@@ -56,7 +56,7 @@ metadata:
     prometheus: prometheus
 spec:
   replicas: 2
-  serviceAccountName: prometheus
+  serviceAccountName: prometheus-operator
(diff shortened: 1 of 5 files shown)


brancz:
Can you explain these changes? The Prometheus instances are intended to have their own service account.

maxlim0:
The bundle.yaml created ServiceAccont: prometheus-operator
In examples we using SA = prometheus, in this case, I can't use examples to correct provision pods.

brancz:
A separate ServiceAccount needs to be provisioned for the Prometheus instances.

maxlim0:
Should bundle.yaml be compatible with the examples?

brancz:
Yes, but they are distinct responsibilities so we'll never have prometheus and prometheus operator use the same serviceaccount.

PR #2778, opened 2019-09-26
update k8s client to 1.16

kinarashah:


brancz:
Awesome, thank you! :tada:

alena1108:
@brancz is there a chance this fix can be included to the next planned release? And if yes, what will be its date

brancz:
Do you have a concrete issue that this fixes? This largely seems like housekeeping to me. We don't have scheduled releases, but typically release every couple of weeks.

PR #2816, opened 2019-10-17
add field ignoreNamespaceSelectors to Prometheus CRD

pgier:
Adds a new field `ignoreNamespaceSelectors` to the Prometheus type which
prevents ServiceMonitors and PodMonitors from collecting
endpoints which are not in the current namespace.


pgier:
Still need to add tests and possibly some additional behaviour.

brancz:
Lgtm once some unit tests covering the generation are added. Nice work!

LiliC:
@pgier is this ready for review? :)

pgier:
@LiliC I just added a unit test, so I think it's ready for review.  This also includes the refactoring from https://github.com/coreos/prometheus-operator/pull/2817, so either that one can be merged first or if this one is merged first, then that one can be closed.

pgier:
@paulfantom Unfortunately, I think even without the refactoring commit it will conflict with #2822.  Unless I just leave the pre-1.7 code as-is and don't implement the new field for the earlier versions.

@LiliC @brancz Should I remove the changes in this PR related to pre-1.7 prometheus versions?

brancz:
lgtm, no let's rather do the pre-1.7 things in the 1.x removal PR.

PR #2820, opened 2019-10-18
Add EnforcedNamespaceLabel to Prometheus CR

LiliC:
This introduces `EnforcedNamespaceLabel` which when set does the following:
- In `PodMonitor` and `ServiceMonitor` for the case of `MetricRelabelConfigs` skips generating the relabilings if the `EnforcedNamespaceLabel` is the same value as the `TargetLabel`
- For `PodMonitor` and `ServiceMonitor` for the case of `RelabelConfigs` the Namespace of the object is added as a relabaling as the last item.
- For Rules we inject a new label `EnforcedNamespaceLabel` key with vlaue of the Objects Namespace. And add the same to the labels.

cc @brancz Can you have a look at the TODOs I left, its more like questions, thanks!


Documentation/api.md
```diff
@@ -395,6 +395,7 @@ PrometheusSpec is a specification of the desired behavior of the Prometheus clus
 | portName | Port name used for the pods and governing service. This defaults to web | string | false |
 | arbitraryFSAccessThroughSMs | ArbitraryFSAccessThroughSMs configures whether configuration based on a service monitor can access arbitrary files on the file system of the Prometheus container e.g. bearer token files. | [ArbitraryFSAccessThroughSMsConfig](#arbitraryfsaccessthroughsmsconfig) | false |
 | overrideHonorLabels | OverrideHonorLabels if set to true overrides all user configured honor_labels. If HonorLabels is set in ServiceMonitor or PodMonitor to true, this overrides honor_labels to false. | bool | false |
+| enforcedNamespaceLabel | EnforcedNamespaceLabel adds a namespace label of origin for each alert and metric that is user created. | string | false |
```
LiliC: I had no clue how to phrase this nicely and descriptive enough, suggestions welcome! :)

go.mod
```diff
@@ -85,6 +38,7 @@ require (
 )
 
 replace (
+	github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.0.0-20190818123050-43acd0e2e93f
```
LiliC: This was needed as for some reason the require changed, so I just changed it back to before. The change was also done in github.com/openshift/prom-label-proxy to match the deps.

go.mod
```diff
@@ -3,74 +3,27 @@ module github.com/coreos/prometheus-operator
...
-	github.com/grpc-ecosystem/grpc-opentracing v0.0.0-20180507213350-8e809c8a8645 // indirect
-	github.com/hashicorp/consul v1.4.4 // indirect
 	github.com/hashicorp/go-version v1.1.0
 	github.com/improbable-eng/thanos v0.3.2
-	github.com/jackc/fake v0.0.0-20150926172116-812a484cc733 // indirect
-	github.com/jackc/pgx v3.2.0+incompatible // indirect
 	github.com/jsonnet-bundler/jsonnet-bundler v0.1.0
-	github.com/knz/strtime v0.0.0-20181018220328-af2256ee352c // indirect
 	github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348
-	github.com/leanovate/gopter v0.2.4 // indirect
```
LiliC: I ran `go mod tidy` 🤷‍♀

LiliC: Seems like this might be because I switched to 1.13 go version, as the CI which is on 1.12 fails. Installing older version and trying again.

Documentation/api.md
brancz: I think I like this name.

pkg/prometheus/promcfg.go
```diff
@@ -844,6 +873,21 @@ func (cg *configGenerator) generateServiceMonitorConfig(
 	return cfg
 }
 
+func enforceNamespaceLabel(namespace, enforcedNamespaceLabel string) (relabeling yaml.MapSlice) {
+	if enforcedNamespaceLabel == "" {
+		return relabeling
+	}
+
+	if namespace == "" {
+		// TODO: what do we do here?
```
brancz: this is impossible, no need to handle

pkg/prometheus/promcfg.go
```diff
@@ -844,6 +873,21 @@ func (cg *configGenerator) generateServiceMonitorConfig(
 	return cfg
 }
 
+func enforceNamespaceLabel(namespace, enforcedNamespaceLabel string) (relabeling yaml.MapSlice) {
```
brancz: I think it would be a bit more natural not to do a named return here and pass the existing relabeling configs in here and append the label enforcing instead of having to deal with a default value return value.

pkg/prometheus/promcfg.go
```diff
@@ -844,6 +873,21 @@ func (cg *configGenerator) generateServiceMonitorConfig(
...
+	if enforcedNamespaceLabel == "" {
+		return relabeling
+	}
+
+	if namespace == "" {
+		// TODO: what do we do here?
+	}
+
+	relabeling = append(relabeling, yaml.MapItem{Key: "source_labels", Value: namespace})
```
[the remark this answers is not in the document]
LiliC: you mean `target_labels` instead of `source_labels `?

pkg/prometheus/rules_test.go
```diff
@@ -19,10 +19,114 @@ import (
...
+
+		Expected string
+	}
+
+	testcases := []testCase{
+		{
+			PromRuleSpec: monitoringv1.PrometheusRuleSpec{
+				Groups: []monitoringv1.RuleGroup{
+					{
+						Name: "alertname",
```
brancz: so far we only have alert rule examples here, can we also make sure to test this with recording rules?

pkg/prometheus/rules_test.go
```diff
@@ -19,10 +19,114 @@ import (
...
+						Rules: []monitoringv1.Rule{
+							{
+								Alert: "alert",
+								Expr:  intstr.FromString("node_cpu_seconds_total{job=\"node-exporter\", mode=\"idle\"}"),
```
brancz: let's also test an example where a user does a query that technically queries another "namespace"

LiliC: Good point!! Will do!

LiliC:
@brancz PTAL

cc @s-urbaniak @paulfantom

Documentation/api.md
LiliC: Not name, I mean the description suggestions of what this does :)

s-urbaniak: maybe instead of "adds" we could say explicitely say "enforces"? The only question mark I had was "of origin". Maybe we could also summarize the edge cases in the description too (as is done in the PR title)?

pkg/prometheus/promcfg.go
```diff
@@ -224,12 +224,28 @@ func (cg *configGenerator) generateConfig(
...
+			scrapeConfigs = append(scrapeConfigs,
+				cg.generateServiceMonitorConfig(
+					version,
+					sMons[identifier],
+					ep, i,
+					apiserverConfig,
+					basicAuthSecrets,
+					bearerTokens,
+					p.Spec.OverrideHonorLabels,
+					p.Spec.EnforcedNamespaceLabel))
```
s-urbaniak: not part of this PR, but I think we have already many precedences wher we could benefit from the "functional API" idiom instead of having very long constructor methods?

pkg/prometheus/promcfg.go
```diff
@@ -537,6 +555,9 @@ func (cg *configGenerator) generatePodMonitorConfig(version semver.Version,
 		for _, c := range ep.RelabelConfigs {
 			relabelings = append(relabelings, generateRelabelConfig(c))
 		}
+		// Because of security risks, whenever enforcedNamespaceLabel is set, we want to append it to the
```
s-urbaniak: can we (shortly) summarize what those security risks are? just to answer the question what threat model we protect against.

pkg/prometheus/rules.go
```diff
@@ -203,6 +207,39 @@ func (c *Operator) selectRules(p *monitoringv1.Prometheus, namespaces []string)
...
+			for ri, r := range group.Rules {
+				if len(promRule.Groups[gi].Rules[ri].Labels) == 0 {
+					promRule.Groups[gi].Rules[ri].Labels = map[string]string{}
+				}
+				promRule.Groups[gi].Rules[ri].Labels[enforcedNsLabel] = ns
+
+				expr := r.Expr.String()
+				parsedExpr, err := promql.ParseExpr(expr)
+				if err != nil {
+					return "", err
```
s-urbaniak: admittedly, I haven't checked the rest of the code, but we add some `errors.Wrap`ish context when returning the error? It could ease debugging in log files.

s-urbaniak:
this looks great, i really just have nits! :tada:

Documentation/api.md
LiliC: Does this sgty @s-urbaniak @brancz 
```
EnforcedNamespaceLabel enforces adding a namespace label of origin for each alert
and metric that is user created. The label value will always be the namespace of the object that is
being created.
```

We have very short descriptions, I think the rest belongs in the docs IMO. I can do that in a follow up, if thats okay with you.

pkg/prometheus/promcfg.go
LiliC: Agreed!

pkg/prometheus/promcfg.go
LiliC: It already says that it ensures it overrides any other relabeling. What did you have in mind to add?

s-urbaniak: if that is all, then this is fine.

Documentation/api.md
s-urbaniak: sounds good! :+1:

LiliC:
@brancz @s-urbaniak Rebased and addressed the comments, PTAL, thanks!

brancz:
lgtm on green

s-urbaniak:
indeed, lgtm on green too :+1:

PR #2826, opened 2019-10-21
docs: Update URL references to api-conventions.md

AngusDavis:
The comments in types.go point to 3 different URLs for k8s API conventions
and sadly only 1 of them is currently active. Update all references to point
to its current home.


AngusDavis:
Looks like I need to run `make generate-in-docker` - running that now.

brancz:
Thanks!

PR #2830, opened 2019-10-22
pkg/prometheus/promcfg.go: Only append relabelings if EnforcedNamespaceLabel value is set

LiliC:
Only append relabelings if EnforcedNamespaceLabel value is set.

cc @s-urbaniak @brancz

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -552,11 +552,10 @@ func (cg *configGenerator) generatePodMonitorConfig(
 		for _, c := range ep.RelabelConfigs {
 			relabelings = append(relabelings, generateRelabelConfig(c))
 		}
-		// Because of security risks, whenever enforcedNamespaceLabel is set, we want to append it to the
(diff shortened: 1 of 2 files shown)


brancz:
lgtm on green

LiliC:
Not sure what we could add to our e2e tests to improve them? As our openshift e2e tests caught this. I added a unit test that failed before this change.

brancz:
We could run promtool checks against all generated configs?

LiliC:
Not sure our e2e tests cover this case, because the pods would not be up and running as prometheus would reject the invalid configuration.

brancz:
Oops I understand now. https://github.com/coreos/prometheus-operator/blob/be1c7825adbfe0cd6e3bf94dec77060be968f752/pkg/prometheus/promcfg.go#L557

This line needs to be outside of the if statement. We need to enforce no matter if there are relabeling rules defined or not by the user.

LiliC:
@brancz good catch! PTAL, thanks!

brancz:
lgtm on greene

s-urbaniak:
lgtm :+1:

PR #2841, opened 2019-10-30
docs: fix typo in api.md

greenstatic:


pgier:
Thanks for the fix!  `Documentation/api.md` is actually autogenerated from types.go, would you mind fixing the typo in `pkg/apis/monitoring/v1/types.go`, and then regenerating?  That should update all the generated files which contain the typo.
```bash
make clean generate-in-docker
```

greenstatic:
@pgier here you go :)

Issue #2824, opened 2019-10-21
Creation/Update of Prometheus object throws error on volume claime template creationTimeStamp being null


pault84:
**What happened?**
Writing a container that deploys our monitoring stack. 
Updated to Operator version 0.33.0 from 0.29.0 so i can deploy on K8S 1.16
We allow users to select our product as a Persistent Volume underneath so we can persist historical data.

in 0.29.0 this deployed fine. creation and update went through and prometheus comes up.
in 0.31.0 this barfs on creation / updating of the prometheus object saying:

 validation failure list:\nspec.storage.volumeClaimTemplate.metadata.creationTimestamp in body must be of type string: \"null\""  

As far as i know, you should never pass in this field as it's generated by the system.

**Did you expect to see some different?**
I expected to see my prometheus pods coming online.

**How to reproduce it (as minimally and precisely as possible)**:
I can't give you my container yet but in my estimation.
Create a Prometheus object with a persistent volume claim template.
Don't pass a creationDate in the metadata.

**Environment**
4 Node k8s cluster
K8S 1.14.7
Ubuntu 18.04

* Prometheus Operator version:
v0.31.0

* Kubernetes version information:
1.14.7

* Kubernetes cluster kind:
kubeadm

* Manifests:

VolumeClaimTemplate without a creation date timestamp.

* Prometheus Operator Logs:
```
[...]

PR #2857, opened 2019-11-15
Fix StatefulSet being needlessly re-created

vsliouniaev:
When using `storage.volumeClaimTemplate` there are some scenarios that require setting `creationTimestamp`

There have been two issues reported around this:
- https://github.com/coreos/prometheus-operator/issues/2824
- https://github.com/coreos/prometheus-operator/issues/2399

An interesting side-effect of this happening is that the timestamp is propagated into the statefulset configuration as well, but this doesn't happen transparently from an operational standpoint - the stateful set is deleted and re-created because updates are forbidden.

Regardless of the implementation of someone's work-around for this (either setting the timestamp to `metav1.Now()` all the time, or only setting it if it is unset) someone moving to an environment where this is required from one where it was not will experience at least one statefulset re-creation per prometheus / alertmanager configuration as the timestamp is changed from `null` to `<something>` on the statefulset.

In environments where a cluster has a lot of Prometheuses on it, and they use PVs this can produce quite a lot of needless pod and cloud-provider storage detach/attach churn.

1. I realize this isn't the optimal solution - we should really just drop the requirement to set this timestamp in the first place - but that fix could take some time, depending on the approach.
2. This probably needs some integration tests, but I'm not sure how to do this comprehensively

---

I also think it's a good idea to provide a metric for the operator's delete-and-recreate behaviour, as it can affect SLAs

cc @s-urbaniak 

Looking for feedback on this so I can proceed in the correct direction

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -132,6 +132,7 @@ func makeStatefulSet(am *monitoringv1.Alertmanager, old *appsv1.StatefulSet, con
 		})
 	} else {
 		pvcTemplate := storageSpec.VolumeClaimTemplate
+		pvcTemplate.CreationTimestamp = metav1.Time{}
 		if pvcTemplate.Name == "" {
 			pvcTemplate.Name = volumeName(am.Name)
 		}
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -217,6 +217,7 @@ func makeStatefulSet(
 		})
 	} else {
 		pvcTemplate := storageSpec.VolumeClaimTemplate
+		pvcTemplate.CreationTimestamp = metav1.Time{}
 		if pvcTemplate.Name == "" {
 			pvcTemplate.Name = volumeName(p.Name)
 		}


s-urbaniak:
@vsliouniaev thanks a lot for the PR, it makes a lot of sense to me :+1: I can surely help in writing an e2e test.

fwiw we are in the process of moving to a controller-tools based CRD generation lead by @pgier in #2855. My hope is that we will be able to successfully set the `nullable` property once controller-tools lands and then can get rid of the `metav1.Time{}` hack.

s-urbaniak:
LGTM once green, but maybe @brancz can give it another pass.

vsliouniaev:
Thanks @s-urbaniak. I'm not certain what the approach to take with a test on this is, since creating something that just validates a statefulset is not deleted and re-created when this change is applied seems pretty artificial and I'm wondering if I could do more.

brancz:
I think doing this in the mean time until we can have the field be properly nullable is reasonable, but not sure how to test this either. I think I’m ok with just going ahead and merging this.

rodrigc:
@vsliouniaev @brancz does this fix #2824 ?

vsliouniaev:
It does not. However it prevents further issues with propagating that timestamp to the StatefulSet. This PR makes the creation timestamp be ignored

PR #2863, opened 2019-11-17
Rename `mesh` ports to be prefixed with protocol

mnaser:
At the moment, it's not possible to rename the port names for the mesh ports.  While that behavior would be nice, it would involve a lot of work.

By swapping the names around, we're able to cater to Istio's naming requirements without affecting the user.


brancz:
As this would really only effect Alertmanager instances in the same mesh and they get updated at the same time I think I'm fine with this change. Anything you can think of @simonpasquier ? If not then I think we can move ahead with this.

simonpasquier:
LGTM. As far as I can see, that change is a no-op for the operator since the port name isn't used anywhere.

simonpasquier:
Thanks!

PR #2870, opened 2019-11-19
Docs: Update Getting-Started examples

gjtempleton:
Updates getting-started example yaml scripts to match
their declared purpose - creates 2 new snippets to
achieve this.

Signed-off-by: GuyTempleton <guy.templeton@skyscanner.net>

PR #2874, opened 2019-11-20
Update getting-started.md

elanv:
It looks like incorrect example for podMonitorSelector.
Fixed the example template.


elanv:
This looks duplicated to #2870

PR #2875, opened 2019-11-20
Add listenAddress option to the services than bind to an IP

fernandocarletti:
This adds the option to use a custom bind address to Alertmanager, Prometheus and Thanos.

The current listenLocal behavior was kept and can be used together with this new options. It is completely backward compatible.

For Alertmanager cluster, it is now possible to use listenLocal and ListenAddress as well.


brancz:
Can you elaborate on the use case(s) for this and why the current behavior is not sufficient?

PR #2879, opened 2019-11-22
Fix plural in metadata.name

lucasmaheo:
This fix allowed me to ```kubectl apply -f ``` the resulting file, while I was getting errors without:
```
The CustomResourceDefinition "prometheus.monitoring.coreos.com" is invalid: metadata.name: Invalid value: "prometheus.monitoring.coreos.com": must be spec.names.plural+"."+spec.group
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,8 +148,8 @@ generate-in-docker:
 
 $(CRD_YAML_FILES): $(CONTROLLER_GEN_BINARY) $(TYPES_V1_TARGET)
 	$(CONTROLLER_GEN_BINARY) crd paths=./pkg/apis/monitoring/v1 output:crd:dir=./example/prometheus-operator-crd
-	cat ./example/prometheus-operator-crd/monitoring.coreos.com_prometheus.yaml | \
-	sed s/plural\:\ prometheus/plural\:\ prometheuses/ \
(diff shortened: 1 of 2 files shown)


pgier:
Thanks for reporting this!  This file is actually generated, and unfortunately the generation doesn't automatically use the correct plural.  You can regenerate this file using the Makefile.
```bash
rm example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
make example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
``` 
Looks like I missed this in the recent build process update.  Can you also update the Makefile to use the correct plural when regenerating the file?
https://github.com/coreos/prometheus-operator/blob/11111cbd3cc9ef1659890a16d6e62c57342cd116/Makefile#L149 
One way to do it is to pipe through another `sed` command.

pgier:
Sorry, I should have mentioned before, you'll also need to run `make generate-in-docker` to update the generated libsonnet files, and then commit the updates.

lucasmaheo:
Yeah I figured I was missing some building process. It should be fixed. Sorry I could not get to it earlier. I made another clean pull request, as this one was done using direct GitHub edits. 
https://github.com/coreos/prometheus-operator/pull/2885

PR #2886, opened 2019-11-27
Add an option to disable compaction

renjugeo:
For resource-constrained clusters with limited cpu we should provide an option to disable compaction. Thanos overrides the compaction settings but i think we should have a general override option as mentioned in #1375. The main reason for doing this is keeping storage for some time but having to restrict the CPU usage to an absolute bare minimum. This feature is effectively providing you the ability to trade off storage and query performance for regular high CPU load.


renjugeo:
cc: @vsliouniaev

vsliouniaev:
@brancz Does this seem like a good reason to add this feature? Would be quite helpful for our case

PR #2887, opened 2019-11-28
pkg/prometheus: add metrics about node endpoints synchronization

simonpasquier:
We have no way currently to detect that the Prometheus operator can't synchronize the nodes' endpoints.

Ref. https://bugzilla.redhat.com/show_bug.cgi?id=1695910

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -84,6 +84,8 @@ type Operator struct {
 	// corresponding actions (add, delete, update).
 	triggerByCounter        *prometheus.CounterVec
 	nodeAddressLookupErrors prometheus.Counter
+	nodeEndpointSyncs       prometheus.Counter
+	nodeEndpointSyncErrors  prometheus.Counter
 
 	host                   string
 	kubeletObjectName      string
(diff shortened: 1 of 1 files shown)


pkg/prometheus/operator.go
```diff
@@ -352,9 +354,19 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer, reconcileErrorsCount
 		Name: "prometheus_operator_node_address_lookup_errors_total",
 		Help: "Number of times a node IP address could not be determined",
 	})
+	c.nodeEndpointSyncs = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_node_syncs_total",
+		Help: "Number of times node endpoints have been synchronized",
+	})
+	c.nodeEndpointSyncErrors = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_node_syncs_failed_total",
+		Help: "Number of times node endpoints couldn't be synchronized",
```
LiliC: Nit, wdyt of this for the help text: "Number of node endpoints synchronisation failures."

pkg/prometheus/operator.go
```diff
@@ -352,9 +354,19 @@ func (c *Operator) RegisterMetrics(r prometheus.Registerer, reconcileErrorsCount
 		Name: "prometheus_operator_node_address_lookup_errors_total",
 		Help: "Number of times a node IP address could not be determined",
 	})
+	c.nodeEndpointSyncs = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_node_syncs_total",
+		Help: "Number of times node endpoints have been synchronized",
```
LiliC: Nit, wdyt of this for the help text: "Number of node endpoints synchronisation."

LiliC: On second thought can't we just have number of successes instead of total? What is the advantage of having total number, versus the failure/success metrics.

simonpasquier: The advantage is that you can compute the rate of failures more easily: `rate(foo_failed_total[5m])/rate(foo_total[5m])`
vs.
`rate(foo_failed_total[5m])/(rate(foo_failed_total[5m]) + rate(foo_success_total[5m])`

Julius explains it [here](https://youtu.be/_MNYuTNfTb4?t=543).

brancz:
lgtm

pkg/prometheus/operator.go
LiliC: Thanks for the explanation!

Issue #2727, opened 2019-08-21
Bad PrometheusRule prevented prometheus operator from starting


corlettb:
The prometheus operator was failing to start with the following error:

```
msg="Unhandled error received. Exiting..." 
err="creating CRDs failed: waiting for PrometheusRule crd failed: timed out waiting for Custom Resource: failed to list CRD: v1.PrometheusRuleList.Items: []*v1.PrometheusRule: v1.PrometheusRule.Spec: v1.PrometheusRuleSpec.Groups: []v1.RuleGroup: v1.RuleGroup.Rules: []v1.Rule: v1.Rule.Annotations: ReadString: expects \" or n, but found f, error found in #10 byte of ...|summary\":false},\"exp|..., bigger context ...|un successfully in the last 24 hours.\",\"summary\":false},\"expr\":\"job:last_success:hours_ago{ exported|..."
```

This turned out to be a problem with the following rule:

```
---
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
...
spec:
  groups:
    - name: fakerule.rules
      rules:
        - alert: SomeAlert
          expr: someexpression > 0
          for: 1m
          labels:
            severity: critical
          annotations:
            summary: No
            description: some alert
```

The problem part of this is the "summary: No". No is special in yaml and will be resolved as a boolean that is false.

Now granted this is bad config, but the operator should deal with this better.

- The operator should log the namespace and path for the invalid crd in the error message.
- A bad prometheusrule should not prevent the operator from starting. Invalid rules should just be skipped over.
- The operator should update a "status" field on the crd to inform the user that their prometheusrule has been rejected and why.
- The operator should expose a metric to prometheus to show failing rules.

allen13:
We have been successfully using the built-in admission webhook in production for the past month now. See implementation details here: https://github.com/helm/charts/tree/master/stable/prometheus-operator/templates/prometheus-operator/admission-webhooks

caarlos0:
similar issue here, but maybe a bit worse:

if the operator is already running and the config is updated to a "bad" one, it keeps running, but throwing that error every time and basically not working.

also no metrics are exposed about this, I expected that `prometheus_operator_rule_validation_errors_total` would increase - or maybe another metrics.

IMHO, increasing that counter would be a good way to inform that things are not working, as I just leave our prometheus broken for a few days without realizing it :( 

if you think this is a good idea and can point me more or less to where I should look at, I can work on a PR 👍 

[...]

PR #2893, opened 2019-12-03
*: instrument operator's list and watch operations

simonpasquier:
Related to #2727 

There's no way to detect that the operator fails to list and watch resources. This PR adds metrics on which we could alert.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -1,0 +1,147 @@
+// Copyright 2019 The prometheus-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package operator
+
+import (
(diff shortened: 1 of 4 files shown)


pkg/operator/operator.go
```diff
@@ -0,0 +1,120 @@
...
+	reconcileErrorsCounter prometheus.Counter
+	stsDeleteCreateCounter prometheus.Counter
+	// triggerByCounter is a set of counters keeping track of the amount
+	// of times Prometheus Operator was triggered to reconcile its created
+	// objects. It is split in the dimensions of Kubernetes objects and
+	// corresponding actions (add, delete, update).
+	triggerByCounter *prometheus.CounterVec
+}
+
+func NewMetrics(name string, r prometheus.Registerer) *Metrics {
```
LiliC: Nit: Can we add function comments.

pkg/operator/operator.go
```diff
@@ -0,0 +1,120 @@
...
+	ret, err := i.next.List(options)
+	if err != nil {
+		i.failed.Inc()
+	}
+	return ret, err
+}
+
+// Watch implements the cache.ListerWatcher interface.
+func (i *instrumentedListerWatcher) Watch(options metav1.ListOptions) (watch.Interface, error) {
+	return i.next.Watch(options)
```
LiliC: It would make sense to also add the same metrics for watch, not necessarily to do it in this PR, we can also open an issue for it. WDYT?

LiliC:
BTW feel free to ping us in PRs directly, as I tend to overlook things easier if I am not directly pinged on.

simonpasquier:
PR updated: comments have been added and watch metrics are implemented too.

PR #2901, opened 2019-12-06
use the latest version of prometheus in compatibility matrix as default

yeya24:

Use the latest version of Prometheus in compatibility matrix instead of v2.7.1

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -34,7 +34,6 @@ import (
 
 const (
 	governingServiceName            = "prometheus-operated"
-	DefaultPrometheusVersion        = "v2.7.1"
 	DefaultThanosVersion            = "v0.8.1"
 	defaultRetention                = "24h"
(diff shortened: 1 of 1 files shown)


brancz:
Could you run `make --always-make format generate-in-docker` and commit your changes? That should make the first CI job pass.

yeya24:
I am sorry I have no idea about how to fix this e2e test.
```
FAIL: TestAllNS/y/PromArbitraryFSAcc/allowed-tls-file (307.18s)
                prometheus_test.go:1981: waiting for targets timed out. 0 of 1 targets found. timed out waiting for the condition
```

brancz:
I re-triggered CI but it appears to be failing consistently. Could you try running the tests locally and see if you can figure out what's happening?

brancz:
@pgier could you check out what went wrong here because this seems odd and we intend to ship this version of prometheus so it would be nice to validate that this works :)

PR #2906, opened 2019-12-11
Fix unmarshalling crd yaml

pgier:
Some fields were being left out of the created CRDs due to missing information during yaml unmarshalling.


brancz:
lgtm

PR #2914, opened 2019-12-13
deprecate baseImage, tag, and sha fields in custom resources

### LINKED ISSUE #2911: Deprecate baseImage, tag, and sha fields from CRDs
With the addition of the `image` field in #2266 to the prometheus, alertmanager, and thanos specs, it seems that we no longer need the separate `baseImage`, `tag`, and `sha` fields.  I'm thinking that if these fields are set, we can print a warning that the Image field should be used instead.  Then we could remove these fields at some point down the road.

pgier:
fixes #2911


brancz:
lgtm, we will need to make sure to announce this in the release notes as well.

PR #2933, opened 2020-01-03
Remove .promu.yml

simonpasquier:
`promu` isn't used anymore so we can just remove `.promu.yml` to avoid confusion.


brancz:
Looks like there was a network issue on CI. Retriggred it.

PR #2936, opened 2020-01-07
Set alertmanager --cluster.peer as fully qualified name

### LINKED ISSUE #2934: DNS definition by Kubernetes affects alertmanager
**What happened?**
Alertmanager `cluster.peer` (added by prometheus-operator) does not translate as a fully qualified name but as:

```
alertmanager-central-alertmanager-0.alertmanager-operated.default.svc.default.svc.cluster.local SOA ns.dns.cluster.local
alertmanager-central-alertmanager-0.alertmanager-operated.default.svc.svc.cluster.local SOA ns.dns.cluster.local
```

DNS resolution inside a container - like any Linux system - is driven by the /etc/resolv.conf config file. The default Kubernetes dnsPolicy is `ClusterFirst` which means any DNS query will be routed to dnsmasq running in the kube-dns pod inside the cluster which - in turn - will route the request to kube-dns application if the name ends with a cluster suffix or to the upstream DNS server otherwise.

The /etc/resolv.conf file inside each container will look like this by default:
```
nameserver xxx.xxx.xxx.xxx
search namespace.svc.cluster.local svc.cluster.local cluster.local eu-central-1.compute.internal
options ndots:5
```

A fully qualified name is a name for which no local search will be executed and the name will be treated as an absolute one during the resolution. Prometheus-operator should add it like         `- --cluster.peer=alertmanager-central-alertmanager-0.alertmanager-operated.default.svc.:9094`
instead of 
`- --cluster.peer=alertmanager-central-alertmanager-0.alertmanager-operated.default.svc:9094`
(dot was added to the end, in order to treat it as FQDN)

I would expect it to resolve as a fully qualified name

0.32

1.15.6

* Kubernetes cluster kind:
kops

* Manifests:
```
Alertmanager statefulset created by the `prometheus-operator`
```
    spec:
      containers:
      - args:
        - --config.file=/etc/alertmanager/config/alertmanager.yaml
        - --cluster.listen-address=[$(POD_IP)]:9094
        - --storage.path=/alertmanager
        - --data.retention=744h
        - --web.listen-address=:9093
        - --web.route-prefix=/
        - --log.format=json
        - --cluster.peer=alertmanager-central-alertmanager-0.alertmanager-operated.default.svc:9094
```

### PR DESCRIPTION
**Pull request by oded-dd**
Fix #2934

A fully qualified name is a name for which no local search will be executed and the name will be treated as an absolute one during the resolution. By convention, DNS software consider a name fully qualified if ends with a full stop (.) and non fully qualified otherwise.

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -324,7 +324,7 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 	podLabels["alertmanager"] = a.Name
 
 	for i := int32(0); i < *a.Spec.Replicas; i++ {
-		amArgs = append(amArgs, fmt.Sprintf("--cluster.peer=%s-%d.%s.%s.svc:9094", prefixedName(a.Name), i, governingServiceName, a.Namespace))
+		amArgs = append(amArgs, fmt.Sprintf("--cluster.peer=%s-%d.%s.%s.svc.:9094", prefixedName(a.Name), i, governingServiceName, a.Namespace))
 	}
 
 	for _, peer := range a.Spec.AdditionalPeers {

### CODE REVIEW

**Comment by s-urbaniak** [comment]
thanks for the contribution, lgtm :+1:

**Comment by s-urbaniak** [comment]
hmm, I guess we have a regression with this:
```

        --- FAIL: TestAllNS/y/PromAlertmanagerDiscovery (328.20s)
            prometheus_test.go:1119: failed to wait for an Alertmanager cluster (alertmanager-test-0) with 3 instances to become ready: failed to wait for initialized alertmanager cluster: timed out waiting for the condition: failed to get correct amount of peers, expected 3, got 1, addresses [{01DXZJZWSVQ5W0YH5YJHVSH0DD 172.17.0.7:9094}]
```

**Comment by oded-dd** [comment]
Ran over the tests and I fail to understand how it is related :(

**Comment by s-urbaniak** [comment]
I restarted the build, maybe we have a flake here :thinking:

**Comment by brancz** [comment]
It looks like the alertmanagers cannot discover each other with this patch. I was actually wondering why we can use this notation when the full dns name should include the cluster suffix no? (Like “cluster.local”)

**Comment by elisiano** [comment]
I don't think this will work. as you said a name with a final dot is fully qualified, which means either it exists or resolution will fail.

But there are no domain ending in `.svc.` by default inside kubernetes.

So what happens without the dot is that the resolution fails and then all other domains inside the `search` domains option of the resolver are tried until a record is found.

```
 ➭ kubectl exec alertmanager-main-0 -c alertmanager cat /etc/resolv.conf
nameserver 100.64.0.10
search monitoring.svc.cluster.local svc.cluster.local cluster.local us-west-2.compute.internal
options ndots:5
```
in this case, at the third retry the domain `cluster.local` will be appended to the original record for resolution and you will have a match.

if you want to use a fqdn, you would need to add `.cluster.local.` but this means that this won't work if someone customizes the default internal domain (unless this is made configurable and defaulting to `.cluster.local.`).

oded-dd:
@elisiano Great. Got it.

Default internal domain is basically added by kubernetes, so if I understand correctly your suggestion should solve it.

brancz:
It's a bit of a leaking abstraction, but I feel that it then should be configurable on the alertmanager object, and if people want they can specify it with an ending colon forcing this.

pgier:
I believe this is resolved by #3138

Issue #2642, opened 2019-06-14
Add StatefulSetMetadata to PrometheusSpec


Devin-Holland:
**What is missing?**
Missing the ability to apply labels and annotations to the generated StatefulSet from the PrometheusSpec

**Why do we need it?**
Being able to tack on labels to the StatefulSet will allow you to ignore the StatefulSet when using the `--prune` flag and `-l` flag on `kubectl apply` so that the StatefulSet does not get deleted every time you re-apply the kubernetes yaml.

brancz:
I felt like we had something like this already. Maybe we’re transferring the labels from the Prometheus object to the statefulset?

Devin-Holland:
There is a feature on the Prometheus object to add `podMetaData` which will allow you to add additional labels to the pod, but for the statefulset the labels and from the Prometheus object are directly copied over and you can't add new labels or annotations to it.

brancz:
Why can’t you accomplish what you are aiming for by adding the label/annotation you want onto the Prometheus object, hence it gets propagated to the statefulset?

Devin-Holland:
So if you run `--purge` on your `kubectl apply` with that label that gets propagated to the stateful set then when kubernetes compares the cluster to your definition yaml it sees the statefulset but it isn't in your definition so it deletes the statefulset.

Example: You have a yaml definition that creates a deployment for Prometheus Operator and then you have a Prometheus object, and a random deployment object in the definition as well. Using Kustomize you apply a common label to all of your objects `purgeMe=true`. You are applying this to a fresh cluster. You want to use `--purge -l purgeMe=true` on your apply so that if you re-apply the yaml without the random deployment in it then the deployment will get deleted since it will have the label but no longer be in your definition yaml. But if you cannot overwrite the purgeMe=true on the stateful set it will get deleted every time you apply the yaml with the purge even though you don't want it to be.

brancz:
That makes perfect sense, we should absolutely have he expected behavior here. How does the ReplicaSet do this in relation to the deployment? Or even more fundamentally pods? Ideally I’d want us to have the expected behavior out of the box, without having to do any custom labeling or anything.

Devin-Holland:
By default the labels on the Prometheus object are not copied to the ReplicaSet or the Pods so prune will not attempt to remove them. You can add labels to the Pods through the podMetadata on the Prometheus object.

brancz:
[...]

PR #2944, opened 2020-01-08
fix generated statefulset being pruned by kubectl

juliantaylor:
When the operator transfers all labels and annotations from the source
prometheus object to the generated statefulset this also kubectl managed
annotations.
These annotations will cause the generated statefulset to be pruned when
the source prometheus is applied with kubectl apply --prune.
Cleanup of the statefulset is managed by the garbage collection via
owner references of the prometheus object (which can be pruned).

To fix this do not transfer any kubectl.kubernetes.io/ annotations to
the generated statefulset.

Closes gh-2642

Signed-off-by: Julian Taylor <juliantaylor108@gmail.com>

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -170,11 +170,19 @@ func makeStatefulSet(
 	}
 
 	boolTrue := true
+	// do not transfer kubectl annotations to the statefulset so it is not
+	// pruned by kubectl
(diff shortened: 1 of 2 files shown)


brancz:
There are some formatting/generation issues. Could you run “make clean generate-in-docker” and commit the changes? Thanks! :)

brancz:
Thanks! Awesome that this finally is in place! :)

PR #2955, opened 2020-01-14
OWNERS: add krasi-georgiev and paulfantom

pgier:
brings the OWNERS file in sync with openshift/prometheus-operator

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,18 +1,22 @@
 reviewers:
-  - brancz
-  - metalmatze
-  - s-urbaniak
-  - squat
-  - LiliC
-  - bwplotka
-  - pgier
-  - simonpasquier
+- brancz
+- bwplotka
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -1,18 +1,22 @@
 reviewers:
-  - brancz
-  - metalmatze
-  - s-urbaniak
-  - squat
-  - LiliC
-  - bwplotka
-  - pgier
-  - simonpasquier
+- brancz
```
s-urbaniak: hmm, is the indentation right?

pgier: yeah, for lists in yaml it's not required to indent.  I just copied the file contents from the openshift repo (https://github.com/openshift/prometheus-operator/blob/master/OWNERS).  You can blame @simonpasquier for changing the indent :)

PR #2958, opened 2020-01-15
prometheus-operator: set config-reloader containers resources

elisiano:
Currently when passing `--config-reloader-(cpu|memory)` values those are
used only for limits. This PR sets the same values for requests.
I wasn't sure if the options for the requests should have been passed
with separate options but I thought this was a good compromise in the
meantime.

The initial question was brought up in https://github.com/coreos/kube-prometheus/issues/376


elisiano:
the first time the tests failed because of formatting and that was fixed.
Now though I'm not sure what the real problem is, I might need some guidance.

elisiano:
I think the failure was a fluke while getting one of the dependencies. I've executed the same command on a linux host (`make --always-make format generate-in-docker && git diff --exit-code"`) and it finishes correctly.

If there's anything that needs to be done on my side please let me know.

Issue #2196, opened 2018-12-05
Make promArg "--storage.tsdb.min-block-duration" configurable


thomschke:
**What did you do?**
Storing historical timeseries (see prometheus/prometheus#4759)

**What did you expect to see?**
I need a prometheus instance started with `--storage.tsdb.min-block-duration` more then 2 hours.

**What did you see instead? Under which circumstances?**
This argument has a fixed value of `2h`.

https://github.com/coreos/prometheus-operator/blob/d37bc833ab4691fe67e17da69893f8a314728eed/pkg/prometheus/statefulset.go#L740

**Environment**

* Prometheus Operator version:

	0.26.0

brancz:
It is not recommended to modify the min block duration, it is specifically chosen to be optimized on all dimensions. Two hours is the optimal point where compression is best. And increasing the time means your Prometheus server will use significantly more memory (as the 2h time frame is held in memory in form of the write-ahead-log).

thomschke:
I understand, but it's the only chance to store historical timeseries at the moment. 

I know my use case is not very common, but at the moment I can't use your operator and have to build my own helm chart :-(

brancz:
What do you mean by store historical timeseries? Are you talking about bulk inserting data?

thomschke:
Yes.

brancz:
Bulk inserting is not a supported feature in Prometheus, and this “workaround” is truly a workaround at best. If that’s what you want to do I would recommend preparing the Prometheus tsdb locally on your computer and then starting Prometheus with the produced database in its persistent volume.

For what it’s worth I’ve heard there is some work on backfilling data being done, but no ETA for this feature.

thomschke:
Thanks for your informations.

wangycc:
Hi @brancz I also want to modify the "--storage.tsdb.min-block-duration" parameter value.
Whenever the data flushes to disk, it will exceed the disk IO load. I want to reduce the flush interval.

abhi4u1947:
Still open issue:

https://github.com/prometheus-operator/prometheus-operator/issues/4414#issuecomment-2402832211

PR #2965, opened 2020-01-17
Add an option to pass block Min/Max duration to prometheus

Cheshirez:
Running prometheus on some unstable environment we want to reduce flushing intervals, doing it with smaller chunks to prevent possible data loss like was mentioned in #2196 comments. For that case would be good to have possibility for tweaking `--storage.tsdb.min-block-duration` and `--storage.tsdb.max-block-duration` from prometheus-operator where not using Thanos compaction.


brancz:
Those flags only exist for performance tests for the prometheus team, they are not safe to be used with any other value than the default. I am very hesitant about adding this. Producing non 2h blocks will cause other problems.

vsliouniaev:
We have use-cases where we use prometheus to scrape and remote-write data while storing nothing locally. Setting block sizes to 15m has allowed us to optimize this use-case. Would there be any interest in supporting this use-case? We're currently using the container strategic merge to support this.

brancz:
> We're currently using the container strategic merge to support this.

This is the right thing to do for now. There is actually work on Prometheus itself happening that reduces the amount of data held in memory ([release candidate for the version that does this is out already](https://github.com/prometheus/prometheus/releases/tag/v2.19.0-rc.0)), I would rather focus the attention to make Prometheus better than expose configuration that is never intended to be used by end users.

Kampe:
This is potentially necessary to set for Thanos Receive components it seems as it only accepts blocks up to 1 hour old. 

https://github.com/thanos-io/thanos/issues/2114

Kampe (approved):

ahmgeek:
@brancz I think this PR is blocking https://github.com/bitnami/charts/issues/3843, can we somehow get it merged or at least assets of what can be done?

brancz:
Any configuration that disables compaction [will do](https://github.com/prometheus-operator/prometheus-operator/blob/7c6a68ba8cde4bacdd26e927e08405820339fac5/pkg/prometheus/statefulset.go#L808-L810) what Thanos wants there, which can either be by explicitly setting the [`.spec.disableCompaction` field to `true`](https://github.com/prometheus-operator/prometheus-operator/blob/7c6a68ba8cde4bacdd26e927e08405820339fac5/pkg/prometheus/statefulset.go#L687), or by [configuring an object storage configuration](https://github.com/prometheus-operator/prometheus-operator/blob/7c6a68ba8cde4bacdd26e927e08405820339fac5/pkg/prometheus/statefulset.go#L752-L780). We have a [user-guide](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/thanos.md#prometheus-custom-resource-with-thanos-sidecar) and [API documentation](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#thanosspec) for that.

I don't see how https://github.com/bitnami/charts/issues/3843 needs this PR.

simonpasquier:
It should be possible to use `spec.additionalArgs` to inject the necessary arguments into the Prometheus container.

matan-legit:
@simonpasquier it's not supported, max block duration is hard coded

PR #2978, opened 2020-01-28
reduce verbosity of deprecation checks

pgier:
I was getting tired of seeing the deprecation warnings in the logs on every sync.  This changes the prometheus and alertmanager operators to only warn when a Prometheus or Alertmanager resource is added or updated.


brancz:
lgtm on green

PR #2980, opened 2020-01-29
refactor update handling of prometheus statefulsets

pgier:
Add a new hash annotation that tracks the state of the
statefulset spec separately from the inputs (Prometheus, Config,
ConfigMaps).  This hash annotation is added immediately
after the statefulset is created, and is checked for changes
during updates to detect if there were manual updates to the
statefulset spec.

This prevents an issue (#2950) where the statefulset is
continously deleted and then recreated in kubernetes v1.17
due to a mismatch between the hash annotation and the state
of the statefulsetspec.  The issue only occurs in kubernetes
v1.17 because the api is more strict about what parts of a
statefulset can be updated.

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -31,7 +31,6 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/mitchellh/hashstructure"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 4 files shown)


pgier:
During testing I found that the prometheus operator still goes through a few iterations of updates to sync everything, but it eventually reaches a stable state.  One possible way to improve this is to just give up on trying to update statefulsets in place, and just always delete when there is a change.

s-urbaniak:
I guess we have a little failure in unit tests:
```
 {
- prometheus-operator-input-hash: "",
+ prometheus-operator-input-hash: "10689339640183164791",
  testannotation: "testannotationvalue",
 }
--- FAIL: TestStatefulSetLabelingAndAnnotations (0.00s)
    statefulset_test.go:76: Annotations are not properly being propagated to the StatefulSet
FAIL
```

vsliouniaev:
> During testing I found that the prometheus operator still goes through a few iterations of updates to sync everything, but it eventually reaches a stable state. One possible way to improve this is to just give up on trying to update statefulsets in place, and just always delete when there is a change.

This means that all pods will get killed and re-created, resulting in down time. For larger deployments this can last many minutes

brancz:
@pgier can probably elaborate but my suspicion is that this doesn't impact the pods being rolled out but just annotations on the statefulset itself.

s-urbaniak:
I am wondering if we should work around this at all as @soltysh acknowledged there is something potentially broken in 1.17 as @pgier already reported in https://github.com/kubernetes/kubernetes/issues/87583.

brancz:
Interesting. Let’s see what the workloads people on kube have to comment about this.

pgier:
@vsliouniaev @brancz It would mean that pods would also be deleted and then recreated by kubernetes because of the ownership relation with the stateful set, so it sounds like that's not a good option.

The problem is that kube 1.17 has stricter update rules for statefulsets which means many changes that were able to successfully update in 1.16 will fail to update in v1.17, which then forces a delete and recreation.

It seems like we will need more sophisticated merge logic for checking what's changed from the previous statefulset to decide whether to update or delete.

pgier:
I improved the implementation of this a bit so it seems to fix #2950 and should be slightly better than before regarding the number of updates and syncs that are performed.  I'll do a bit more testing, but I think this is ok to merge.

pgier:
Closing in favor (#2987) which is simpler solution focused just on fixing #2950

vsliouniaev:
@pgier I mean that deleting the statefulset would cause an outage. Deleting pods one at a time is expected

PR #2984, opened 2020-01-29
improve thanosruler tests

pgier:
Add basic rules test for thanos.  Fix e2e test to call thanos e2e test.


pgier:
@s-urbaniak @brancz LGTY?

PR #2985, opened 2020-01-29
Bump thanos reloader package

yeya24:

This PR bumps the version of thanos/pkg/reloader package, which is used in prometheus-config-reloader.

The previous used reloader package `github.com/improbable-eng/thanos v0.3.2` exists some bugs, which were fixed in later thanos release. So I think it is OK to update the dependency here.


brancz:
lgtm on green

PR #2993, opened 2020-01-31
Support thanos tracing

yeya24:

Fixes https://github.com/coreos/prometheus-operator/issues/2992

Already tested in my local environment.

But since the current thanos-ruler spec doesn't support additional sidecar container, so I only test thanos sidecar tracing and it works.


yeya24:
CI is broken, but seems irrelevant

Documentation/api.md
```diff
@@ -647,6 +647,7 @@ ThanosSpec defines parameters for a Prometheus server within a Thanos deployment
 | resources | Resources defines the resource requirements for the Thanos sidecar. If not provided, no requests/limits will be set | [v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#resourcerequirements-v1-core) | false |
 | objectStorageConfig | ObjectStorageConfig configures object storage in Thanos. | *[v1.SecretKeySelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#secretkeyselector-v1-core) | false |
 | listenLocal | ListenLocal makes the Thanos sidecar listen on loopback, so that it does not bind against the Pod IP. | bool | false |
+| tracingConfig | TracingConfig configures tracing in Thanos. | *[v1.SecretKeySelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#secretkeyselector-v1-core) | false |
```
brancz: as this config is not stable on thanos can you add that this is experimental?

yeya24: Thanks. Updated.

yeya24: Should we add a version check? Since the tracing flag is added in v0.6.0

pgier: I think we are not planning to support older versions of Thanos at this point, so should be ok without a version check for this.

PR #2994, opened 2020-02-01
Add thanos-ruler rules configmap reloader

yeya24:
Add thanos rules configmap reloader. Use the same logic with the reloader for Prometheus rules.

Already tested in my local environment.


brancz:
cc @bwplotka is there no thanos built in for this?

bwplotka:
Not really. It behaves like Prometheus, so it allows SIGHUP/reload API and that's it. Maybe we should add reloader straight away to ruler, but given that Prometheus is not doing it, and you need such reloader anyway... I might think this change is fair.

pgier:
LGTM

PR #2997, opened 2020-02-03
add container customization to thanos-ruler

pgier:
Adds the Containers and InitContainers fields to ThanosRuler allowing
customization of the generated StatefulSet/Pod config.


pgier:
cc @brancz

yeya24:
I think this feature is important for adding sidecar containers to thanos ruler, like adding jager-agent

pgier:
@LiliC It matches the config of the prometheus and alertmanager operators, allows adding sidecar containers as @yeya24 mentioned, and also allows config of any CLI args which don't have explicit fields.  @s-urbaniak had asked me to add this in the original implementation, but for some reason I missed it.

LiliC:
> and also allows config of any CLI args which don't have explicit fields.

yes this is why I asked, if we have some specific args in mind we should add those to the spec, but can be done in a follow up.

PR #2998, opened 2020-02-03
pkg/prometheus/statefulset.go: Bump versions of prometheus in CompatibilityMatrix

LiliC:
Is it possible we remove this for a solution that would not need updating this? 🤔 

cc @coreos/team-monitoring


brancz:
We could sync it automatically from github releases but that would have the same issues as using “latest” type images etc. I think the way we have it is fine.

Lgtm on green

PR #3000, opened 2020-02-03
Support thanos ruler replica label

### LINKED ISSUE #2995: Thanos ruler cannot upload blocks
**What happened?**

The thanos ruler cannot upload blocks to object storage because of no external label specified in Ruler.

**Potential fix**

Add `--label=ruler-replica=$(POD_NAME)`  to ruler args.

**Environment**

* Prometheus Operator version: master

    `kubectl version`
    

* Kubernetes cluster kind:

    insert how you created your cluster: kops, bootkube, tectonic-installer, etc.

* Manifests:

```
```

* Prometheus Operator Logs:

```
```

yeya24:

Fixes #2995

cc @pgier


pkg/thanos/statefulset.go
```diff
@@ -37,6 +37,7 @@ const (
 	defaultPortName           = "web"
 	defaultRetention          = "24h"
 	defaultEvaluationInterval = "15s"
+	defaultReplicaLabelName   = "ruler_replica"
```
pgier: I think we should default this to "thanos_ruler_replica" just to make it clearer, WDYT?

yeya24: It is definitely better. Thanks!

pkg/thanos/statefulset.go
```diff
@@ -159,6 +160,29 @@ func makeStatefulSetSpec(tr *monitoringv1.ThanosRuler, config Config, ruleConfig
...
+	// TODO: Users can add other labels to ruler.
+	if replicaLabelName != "" {
+		trCLIArgs = append(trCLIArgs, fmt.Sprintf(`--label=%s="$(POD_NAME)"`, replicaLabelName))
+	}
+
+	// TODO: Users can add other alert drop labels to ruler.
+	dropLabels := []string{replicaLabelName}
+	for _, lb := range dropLabels {
+		if lb != "" {
+			trCLIArgs = append(trCLIArgs, fmt.Sprintf("--alert.label-drop=%s", lb))
```
yeya24: Is there any case that we don' t want to drop the replica label in the alerts? @pgier

pgier: Not that I can think of.  Maybe @brancz or @simonpasquier have other ideas.

pkg/thanos/statefulset.go
```diff
@@ -158,7 +159,37 @@ func makeStatefulSetSpec(tr *monitoringv1.ThanosRuler, config Config, ruleConfig
...
+	if tr.Spec.ReplicaLabelName != nil {
+		if *tr.Spec.ReplicaLabelName != "" {
+			replicaLabelName = *tr.Spec.ReplicaLabelName
+		} else {
+			replicaLabelName = ""
+		}
+	}
+	// TODO: Users can add other labels to ruler.
+	if replicaLabelName != "" {
+		trCLIArgs = append(trCLIArgs, fmt.Sprintf(`--label=%s="$(POD_NAME)"`, replicaLabelName))
```
yeya24: Do we need to check whether object config is set or not? Since if we add the `thanos_ruler_replica` label by default, then all thanos-query may require to add `--query.replica-label=thanos_ruler_replica` @pgier

pkg/thanos/statefulset.go
brancz: I think we always want to drop them.

brancz:
I think there can be multiple labels just like external labels. And if not then we should implement that on Thanos.

yeya24:
> I think there can be multiple labels just like external labels. And if not then we should implement that on Thanos.

Yes, Thanos supports multiple labels. And I think we should make this configurable in Thanos spec. If it is what we want, I can open another PR for it.

Another question: do we need to drop every label we configured in alert?

brancz:
> Another question: do we need to drop every label we configured in alert?

I would say so yes.

brancz:
On second thought, I don't think the duplication is that bad and I can imagine setups where certain labels (like tenancy labels) may want to be preserved.

pgier:
I'm thinking that instead of the `replicaLabelName` field we should just have `labels` and `alertLabelDrops` to match the CLI params of thanos rule.  That way the user has more flexibility.  And they can have the same default settings as in the current PR.

yeya24:
> I'm thinking that instead of the `replicaLabelName` field we should just have `labels` and `alertLabelDrops` to match the CLI params of thanos rule. That way the user has more flexibility. And they can have the same default settings as in the current PR.

I am thinking having all `replicaLabelName`, `labels`, `alertLabelDrops`.  So if we don't have `replicaLabelName`, you mean that the replica label name is not configurable by users and we just hardcode `thanos_ruler_replica`?

pgier:
> @pgier At first I am thinking having all replicaLabelName, labels, alertLabelDrops. So if we don't have replicaLabelName, you mean that the replica label name is not configurable by users and we just hardcode thanos_ruler_replica?

What I mean is that if the user doesn't configure `labels` or `alertLabelDrops`, then we would default to passing `--label=thanos_ruler_replica=${POD_NAME}` and `--alert.label.drop=thanos_ruler_replica` to the CLI args.  If the user sets `labels`, which could be a map, then we would not set any default labels, and just pass the labels to thanos ruler as given by the user.  In that case the user would be responsible for setting the replica label(s) as needed.  The `alert.label.drops` could default to whatever the label names are, otherwise set to whatever the user provided if they set that field.

yeya24:
> > @pgier At first I am thinking having all replicaLabelName, labels, alertLabelDrops. So if we don't have replicaLabelName, you mean that the replica label name is not configurable by users and we just hardcode thanos_ruler_replica?
> 
> What I mean is that if the user doesn't configure `labels` or `alertLabelDrops`, then we would default to passing `--label=thanos_ruler_replica=${POD_NAME}` and `--alert.label.drop=thanos_ruler_replica` to the CLI args. If the user sets `labels`, which could be a map, then we would not set any default labels, and just pass the labels to thanos ruler as given by the user. In that case the user would be responsible for setting the replica label(s) as needed. The `alert.label.drops` could default to whatever the label names are, otherwise set to whatever the user provided if they set that field.

Thanks for clarifying. I agree it is a better approach. I will update my PR as you said.

yeya24:
Although this one is merged, there is still something needs to be improved. And I don't update the docs about how the labels work in ruler, maybe I can open a follow up PR to improve it.

brancz:
Sorry maybe a bit hasty. Please feel free to improve :)

PR #3003, opened 2020-02-03
consolidate logic for determining image urls and versions

pgier:
This change introduces three new CLI flags for specifying default image URLs (`alertmanager-default-image`, `prometheus-default-image`, and `thanos-default-image`) and deprecates the similar `-base-image` CLI args.  This follows along with the previous changes (#2914) which deprecated the use of baseImage, tag, and sha fields.

This also moves the all the default image versions into the `operator` package.  This helps maintain the Thanos image version for example, which currently has a two defaults, one in the thanos operator package and one in the prometheus operator package (for the sidecar).

This also moves the logic for figuring out the container image URLs into the `operator` package to ensure consistency in how the image URL and version is determined using a combination of the various config parameters (image, baseimage, version, tag, etc).  If we remove the deprecated fields at some point in the future, then it should be possible to simplify this quite a bit.

I tried to maintain consistent behaviour with the previous versions and added some new tests to verify output, but it's possible that I missed some edge cases where a user sets an unusual combination of CLI baseImage, custom resource base image, version, tag etc.  But I think it shouldn't be difficult to either fix or provide a workaround if those cases come up.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -182,6 +182,16 @@ func init() {
 	if len(cfg.Namespaces.ThanosRulerAllowList) == 0 {
 		cfg.Namespaces.ThanosRulerAllowList = cfg.Namespaces.AllowList
 	}
+
+	if cfg.AlertmanagerDefaultBaseImage != operator.DefaultAlertmanagerBaseImage {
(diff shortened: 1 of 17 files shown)


pgier:
on hold for now, I think the logic needs to be improved here

pgier:
@brancz PTAL, I've update this one, I think it should be ok to merge now.

brancz:
Sorry looks like this has conflicts now :/ otherwise lgtm

pgier:
Closing this one as it has been mostly replaced by other PRs

PR #3006, opened 2020-02-04
Upgrade jsonnet bundler

pgier:
Not sure if we want to start upgrading projects yet to jb v0.2.0, but I thought I would create this PR while I was thinking about it after some testing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/protobuf v1.3.2
 	github.com/hashicorp/go-version v1.1.0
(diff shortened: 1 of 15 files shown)


pgier:
@metalmatze says we should start using jsonnet-bundler v0.2.0, so I think this should be ok to merge on green

brancz:
lgtm on green. There is even a newer version coming soon that will feature canonical import paths, but one after another :)

elisiano:
somewhat related to this, can a `--version` flag be added to `jb`? right now it's confusing for end users to know what they installed.

pgier:
@elisiano There is now a `--version` flag in v0.2.0, just not in v0.1.0.

PR #3008, opened 2020-02-04
Release 0.35.1

pgier:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Next release
 
+## 0.35.1 / 2020-01-13
+
+* [ENHANCEMENT] Restrict api extension RBAC rules (#2981)
```
LiliC: hmm did we merge this into 0.35 branch? 🤔 

In patch release/z-stream we don't want ENHANCEMENT or feature work.

pgier: It's a permission fix, so it's not really a new feature.  I wasn't sure whether to call it a bugfix or enhancement, and decided on enhancement since I didn't see a specific bug referenced.  But probably @s-urbaniak has more info.

LiliC: I would say its CHANGE then? 🤷‍♀ :)

pgier: Updated, thanks!

PR #3013, opened 2020-02-05
fix date in CHANGELOG for v0.35.1

pgier:
I noticed immediately after releasing that the date was incorrect in the changelog :(


brancz:
lgtm on green

PR #3018, opened 2020-02-08
Update comments from prometheus to thanos ruler

yeya24:

Updated some comments in Thanos Ruler.

### DIFF
--- a/pkg/thanos/rules.go
+++ b/pkg/thanos/rules.go
@@ -75,7 +75,7 @@ func (o *Operator) createOrUpdateRuleConfigMaps(t *monitoringv1.ThanosRuler) ([]
 		level.Debug(o.logger).Log(
 			"msg", "no PrometheusRule changes",
 			"namespace", t.Namespace,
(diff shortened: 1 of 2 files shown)


pkg/thanos/rules.go
```diff
@@ -75,7 +75,7 @@ func (o *Operator) createOrUpdateRuleConfigMaps(t *monitoringv1.ThanosRuler) ([]
 		level.Debug(o.logger).Log(
 			"msg", "no PrometheusRule changes",
 			"namespace", t.Namespace,
-			"prometheus", t.Name,
+			"thanos", t.Name,
```
yeya24: Not sure, but which one is better `thanos` or `thanos ruler`?

pgier: I think either is fine here.  I've been trying to use `thanos` in places where we could potentially re-use logic in the future for additional thanos components, and `thanos-ruler` when the code is specific just to thanos-ruler.

PR #3027, opened 2020-02-15
Set gRPC endpoint when listen local is specified in thanos ruler

yeya24:

Thanos ruler now has https://github.com/coreos/prometheus-operator/issues/2435 problem as well so it should also set gRPC endpoint to 127.0.0.1

Basically copied from https://github.com/coreos/prometheus-operator/pull/2728.


brancz:
How is the setup going to work in this case, as in how would a querier communicate with a ruler? I feel this should probably be two different fields to control them individually.

PR #3030, opened 2020-02-17
fix alertmanager statefulset being pruned by kubectl

### LINKED ISSUE #3029: Alertmanager statefulset pruned unnecessarily
**What happened?**

When kubectl apply --prune is run against a system deployed with prometheus-operator  the alertmanager stateful set gets pruned every time

Alertmanager stateful set not pruned

This is the same issue as https://github.com/coreos/prometheus-operator/issues/2642, but for the alertmanager stateful set instead of the prometheus stateful set.

alexconlin:
kubectl-managed annotations will cause the generated statefulset to be pruned
when the source alertmanager is applied with kubectl apply --prune.

To fix this do not transfer any kubectl.kubernetes.io/ annotations to
the generated statefulset.

This is the same problem as fixed in #2944 but for alertmanager rather than
prometheus.

Fixes #3029

Signed-off-by: Alex Conlin-Oakley <alexconlin@gmail.com>

### DIFF
--- a/pkg/alertmanager/statefulset_test.go
+++ b/pkg/alertmanager/statefulset_test.go
@@ -25,6 +25,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kylelemons/godebug/pretty"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/alertmanager/statefulset_test.go
```diff
@@ -53,8 +62,14 @@ func TestStatefulSetLabelingAndAnnotations(t *testing.T) {
 
 	require.NoError(t, err)
 
-	if !reflect.DeepEqual(labels, sset.Labels) || !reflect.DeepEqual(annotations, sset.Annotations) {
-		t.Fatal("Labels or Annotations are not properly being propagated to the StatefulSet")
+	if !reflect.DeepEqual(labels, sset.Labels) {
+		fmt.Println(pretty.Compare(labels, sset.Labels))
```
simonpasquier: `t.Log(pretty.Compare(labels, sset.Labels))` instead

pkg/alertmanager/statefulset_test.go
```diff
@@ -53,8 +62,14 @@ func TestStatefulSetLabelingAndAnnotations(t *testing.T) {
...
+		t.Fatal("Labels are not properly being propagated to the StatefulSet")
+	}
+
+	if !reflect.DeepEqual(expectedAnnotations, sset.Annotations) {
+		fmt.Println(pretty.Compare(expectedAnnotations, sset.Annotations))
```
simonpasquier: same thing

PR #3035, opened 2020-02-22
Simplify multi-arch building.

NicolasFloquet:
I found it hard to build prometheus-operator for arm64 target.
I'm suggesting this little change allowing the use of variables OS and ARCH to specify the target architecture.

Those variables are passed to golang through GOOS and GOARCH variable, and to docker build through build-args.


pgier:
@NicolasFloquet Would you mind rebasing to get rid of the extra merge commit?

PR #3038, opened 2020-02-24
Fix flaky test in thanos ruler

yeya24:

Flaky test found in https://travis-ci.org/coreos/prometheus-operator/builds/654476206?utm_source=github_status&utm_medium=notification. 
We need to sort the results because the map is unordered.


yeya24:
cc @pgier

PR #3039, opened 2020-02-25
Make the ports on the kubelet Service and Endpoints match.

coderanger:
Prometheus only actually uses the Endpoints but it helps humans to have them be the same.


pgier:
LGTM

Issue #3041, opened 2020-02-25
podMetadata feature no longer functional


coderanger:
**What happened?**

All edits to the `podMetadata` field of a Prometheus object are discarded by kube-apiserver.

**Did you expect to see something different?**

Data should be stored by kube-apiserver.

**How to reproduce it (as minimally and precisely as possible)**:

Try to set a label or annotation using podMetadata.

**Environment**

* Prometheus Operator version: 0.36.0

* Kubernetes version information: 1.17.2

* Kubernetes cluster kind: kops

* Manifests:

```
podMetadata:
  labels:
    foo: bar
```

* Prometheus Operator Logs: N/A (data never makes it that far.

**Anything else we need to know?**:

This was caused by a mix of #2855 where controller-tools didn't recurse into the ObjectMeta struct so it was changed in the generated CRD to be just a plain `type: object` with no fields. And then #2903 which disabled saving unknown fields so any data set on `podMetadata` will be discarded by apiserver since it has no declared fields. The same issue might be affecting other fields, I didn't do a check of the full diff from #2855.

Getting controller-tools to recurse outside of the current package is an old bug, we've run into it on our operators a few times, but hopefully an upgrade there would fix it? If not, I'm sure Solly could assist.

pgier:
I tried updating to the latest version of controller-tools (v0.2.5) but it seems this is still an issue (https://github.com/kubernetes-sigs/controller-tools/issues/385).  There is a possible fix available (https://github.com/kubernetes-sigs/controller-tools/pull/395), so a local patch might be an option until it's fixed in controller-tools master.

pgier:
I created #3043 to test out the upstream patch.  @coderanger can you give that one a try to see if it fixes the issue?

PR #3042, opened 2020-02-25
Update kube dependencies v0.17.3

pgier:


pgier:
> You mean 1.17.3?

The tags for the various libraries seem to use v0.17.3 even though it corresponds to kube 1.17.3.  I'm not sure why (https://github.com/kubernetes/api/releases/tag/v0.17.3)

> Whats the reason behind this upgrade?

This version of apimachinery is required for the latest controller-tools (v0.2.5) which I was trying to upgrade to see if it fixes #3041.  Also the upgrade allows us to remove several of the `require` directives in go.mod.

PR #3043, opened 2020-02-25
Preserve pod metadata in custom resources

### LINKED ISSUE #3041: podMetadata feature no longer functional

pgier:
This applies a patch (https://github.com/kubernetes-sigs/controller-tools/pull/395) the controller-gen binary to generate validation for PodMetadata and fix #3041.  Since this is not merged yet into upstream controller-tools, I'm not sure if we want to merge it here.  So for now this PR is just for testing.


coderanger:
Built and deployed, and it seems happy now :) Thanks! If you don't want to go with a pre-release patch to c-t, another option would be to just make our own struct there instead of directly using ObjectMeta. Which might help anyway since only a few fields from ObjectMeta really make sense there?

pgier:
> another option would be to just make our own struct there instead of directly using ObjectMeta.

@coderanger That sounds like it could work, and it's similar to some of the comments in controller-tools about how they are planning to handle ObjectMeta, so hopefully somewhat in-line with their future changes.

pgier:
@coderanger It seems we are only using the Labels and Annotations from the ObjectMeta type.  The rest is just ignored anyway.  So I changed the type of PodMetadata to a new custom type PodMeta which includes only Labels and Annotations.  WDYT?

coderanger:
@pgier Heh, was just finishing up a patch to do the same to try and save you the trouble, but 👍 from me. I don't think any other fields make sense to be use controlled. In some galaxy brain mode technically you could allow for non-controller owner refs but I think the use case there is too wonky.

PR #3045, opened 2020-02-27
fix thanos-ruler status reporting

pgier:
Make sure the labels applied to the pods match the labels which
the status function uses for searching.


s-urbaniak:
lgtm :tada:

PR #3051, opened 2020-02-28
support Prometheus v2.16.0

### LINKED ISSUE #3050: Upgrade to v2.16.0 image with s390x support
We are testing EAP on s390x architecture and we have test where EAP and Prometheus are used together. We use this prometheus operator to start up Prometheus.

Please provide release of Prometheus Operator with s390x image support v2.16.0 [2].

We believe it is enough to update [1]. Also consider to expose [1] to any arbitrary image version, which would also help in our case.

[1] https://github.com/coreos/prometheus-operator/blob/release-0.29/pkg/prometheus/statefulset.go#L89
[2] https://quay.io/repository/prometheus/prometheus?tab=tags

pgier:
Fixes #3050

### DIFF
--- a/Documentation/compatibility.md
+++ b/Documentation/compatibility.md
@@ -51,6 +51,7 @@ The versions of Prometheus compatible to be run with the Prometheus Operator are
 * v2.11.0
 * v2.14.0
 * v2.15.2
+* v2.16.0
 
 ## Alertmanager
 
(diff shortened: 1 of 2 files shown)


s-urbaniak:
lgtm :+1:

bwplotka:
:+1:

PR #3057, opened 2020-03-02
only set thanos web.route-prefix if it was set by user

pgier:
Due to a problem in the way thanos handles a web.route-prefix set
to '/' (https://github.com/thanos-io/thanos/issues/2207), the value
 should only set if the user has set a value for it.  If it's set
to empty string, no argument should be passed, so that thanos
uses it's own default value.


pgier:
@yeya24 Does this look ok to you?

PR #3058, opened 2020-03-02
add web.external-prefix to thanos-ruler

pgier:
Adds an option to set the CLI arg 'web.external-prefix' on the
ThanosRuler type.


s-urbaniak:
lgtm

PR #3064, opened 2020-03-05
Clarify that Endpoint.targetPort is pod port

whut:
Now it is more clear that `port` is name of port from the service, while targetPort is the name of port from the target pod.


brancz:
This is generated documentation. Could you also add this in the go structs of the API?

pgier:
@whut Can you update types.go (https://github.com/coreos/prometheus-operator/blob/master/pkg/apis/monitoring/v1/types.go#L589 and maybe also https://github.com/coreos/prometheus-operator/blob/master/pkg/apis/monitoring/v1/types.go#L657), and then run `make generate-in-docker`.  That should sync the source and the docs.

whut:
Done for ServiceMonitor.

In case of PodMonitor, it looks like `port` and `targetPort` here does the same thing: https://github.com/coreos/prometheus-operator/blob/v0.37.0/pkg/prometheus/promcfg.go#L475-L482 (except `port` does not support a number). This makes sense as PodMonitor is all about pods, so no service port. I believe one of the two should be deprecated, IMO `targetPort` is term from Service, so this one should be dropped.

PR #3065, opened 2020-03-05
Fix constant Thanos Ruler Statefulset generation

### LINKED ISSUE #3056: ThanosRuler restarts instead of reloading with new PrometheusRules
**What happened?**
When adding new PrometheusRule objects, the ThanosRuler pod will restart instead of reloading. 

The ThanosRuler pod remain running and a call to /-/reload so that the new rules are found.

Create a running ThanosRuler pod with some PrometheusRules. Add multiple new PrometheusRule objects and see if the pod terminates.

We're using PersistentVolumes. It seems the pod keeps running if one PrometheusRule is added, but if multiple are created at once it terminates.
This issue also seems to occur with bulk edits/deletes of PrometheusRule objects

jalev:
Fixes #3056

Basically the Thanos Ruler statefulset object is constantly being updated whenever a PrometheusRule object undergoes a CRUD operation. This PR makes it so that:

- We no longer look for the specific name of the ConfigMap created, instead using globs to match zero or more ConfigMaps. The container arguments will now not get updated if one or more configmap names are created.
- We do a check on the StatefulSet that the operator wants to generate for Thanos Ruler. If it does not need to be updated, it won't be.

### DIFF
--- a/pkg/thanos/operator.go
+++ b/pkg/thanos/operator.go
@@ -741,9 +741,9 @@ func (o *Operator) createCRDs() error {
 func createSSetInputHash(tr monitoringv1.ThanosRuler, c Config, ruleConfigMapNames []string, ss interface{}) (string, error) {
 	hash, err := hashstructure.Hash(struct {
 		TR monitoringv1.ThanosRuler
-		C Config
-		S interface{}
-		R []string `hash:"set"`
+		C  Config
+		S  interface{}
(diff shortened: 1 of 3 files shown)


pgier:
Thanks for working on this!  Looks good so far, I think it just needs a `make format` to fix the formatting in a couple places for the CI to pass.

PR #3067, opened 2020-03-06
pkg/thanos: allow multiple alertmanager urls to be specified

pgier:
The 'thanos rule' command allows multiple alertmanagers.url args to be
specified, so this changes the ThanosRuler custom resource spec to
match.


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Next release
 
+* [CHANGE] Changed ThanosRuler custom resource field alertmanagersURL type from string to []string` (#3067)
```
lilic: Didn't know we started writing next release as part of features? Do we have it document somewhere?

Documentation/api.md
```diff
@@ -710,7 +710,7 @@ ThanosRulerSpec is a specification of the desired behavior of the ThanosRuler. M
 | listenLocal | ListenLocal makes the Thanos ruler listen on loopback, so that it does not bind against the Pod IP. | bool | false |
 | queryEndpoints | QueryEndpoints defines Thanos querier endpoints from which to query metrics. Maps to the --query flag of thanos ruler. | []string | true |
 | queryConfig | Define configuration for connecting to thanos query instances. If this is defined, the QueryEndpoints field will be ignored. Maps to the `query.config` CLI argument. Only available with thanos v0.11.0 and higher. | *[v1.SecretKeySelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#secretkeyselector-v1-core) | false |
-| alertmanagersUrl | Define URL to send alerts to alertmanager.  For Thanos v0.10.0 and higher, AlertManagersConfig should be used instead. Maps to the `alertmanagers.url` arg. | string | false |
+| alertmanagersUrl | Define URL(s) to send alerts to alertmanager.  For Thanos v0.10.0 and higher, AlertManagersConfig should be used instead.  Note: this field will be ignored if AlertManagersConfig is specified. Maps to the `alertmanagers.url` arg. | []string | false |
```
lilic: Nit:
```suggestion
| alertmanagersUrl | Define URLs to send alerts to Alertmanager. For Thanos v0.10.0 and higher, AlertManagersConfig should be used instead. Note: this field will be ignored if AlertManagersConfig is specified. Maps to the `alertmanagers.url` arg. | []string | false |
```

One question about the "this field will be ignored if AlertManagersConfig is specified", do we do this or does thanos do this?

CHANGELOG.md
pgier: We don't have any official practice around this AFAIK, I just wanted to make sure it was included in the next changelog since it is a breaking change.

Documentation/api.md
pgier: It's done by both us and Thanos.  The statefulset generation will only add the `--alertmanager.url` flag if `alertmanagersConfig` is not set.  This matches what thanos does.  Although maybe it would be better for us to just pass all the config to Thanos to avoid any possible behaviour inconsistencies.  WDYT?

lilic: Depends, what was the reason behind it in the first place, to handle it depending on alertmanagersConfig?

pgier: I'm don't think I put a lot of thought into it, probably just trying to match what thanos ruler is expecting.  Maybe even a warning would be good if both are specified since that's most likely a configuration error.

lilic: Yeah lets do it in a follow up I would say, or if someone hits the issue.

PR #3068, opened 2020-03-06
pkg/thanos: add queryConfig to ThanosRuler spec

pgier:
This allows TLS config using the 'query.config' CLI arg to
thanos ruler.


s-urbaniak:
lgtm :tada:

Issue #3059, opened 2020-03-03
Need TLS-config for Thanos


Zenlil:
**What is missing?**
Options to enable TLS-config in the Prometheus-CRD (i.e the --grpc-server-tls-* options) for the Thanos Sidecar

**Why do we need it?**
To allow multi-cluster prometheus with thanos-sidecar serve grpc to Thanos-Query in a different k8-cluster. GRPC won't work properly cross-cluster (using nginx-ingress as example) without TLS

**Environment**
2 k8-cluster
one with Thanos Store+Query+Compactor + Prom-Op with sidecar
The other one with only Prom-Op with sidecar

* Prometheus Operator version:
0.37.0

brancz:
I think this is reasonable to add. Would you like to do a PR to add this? :)

mikonse:
This feature would also be essential to have for the ThanosRuler-CRD as thanos only supports full tls encryption/authentication if one component of the thanos system is using it (e.g. sidecar-query).

Zenlil:
My attempt at solving this is actually on Thanos itself with a small change: https://github.com/thanos-io/thanos/pull/2227

lilic:
I pinged Bartek on the PR, if it doesn’t get accepted then let’s implement it in the CRD.

bwplotka:
I Will comment on the PR, TL;DR I think we can consider it in Thanos, however still config for `mTLS` would really nice to have in prometheus operator anyway!

> This feature would also be essential to have for the ThanosRuler-CRD as thanos only supports full tls encryption/authentication if one component of the thanos system is using it (e.g. sidecar-query).

Hopefully, this is will be fixed soon: https://github.com/thanos-io/thanos/pull/2226 Wonder if that affects the format of the TLS config.

lilic:
> Wonder if that affects the format of the TLS config this issue consider.

I mean that will be merged in the future release, so for now I would suggest we add support for the TLS config fields into the CRD natively, but something to consider!

Plan to start working on this soon, but if you want @mikonse or @Zenlil to add this support to prom-operator instead, you are welcome to.

PR #3069, opened 2020-03-08
Thanos ruler additional args

pgier:
This is an attempt to help with forward/backward compatibility in case CLI arguments change between Thanos releases.


Documentation/api.md
```diff
@@ -727,6 +727,7 @@ ThanosRulerSpec is a specification of the desired behavior of the ThanosRuler. M
 | alertDropLabels | AlertDropLabels configure the label names which should be dropped in ThanosRuler alerts. If `labels` field is not provided, `thanos_ruler_replica` will be dropped in alerts by default. | []string | false |
 | externalPrefix | The external URL the Thanos Ruler instances will be available under. This is necessary to generate correct URLs. This is necessary if Thanos Ruler is not served from root of a DNS name. | string | false |
 | routePrefix | The route prefix ThanosRuler registers HTTP handlers for. This allows thanos UI to be served on a sub-path. | string | false |
+| additionalArgs | AdditionalArgs is a list of CLI arguments which should be added to the automatically configured arguments when running the Thanos Ruler instance. | []string | false |
```
lilic: hmm do we have a precedence for this?

> This is an attempt to help with forward/backward compatibility in case CLI arguments change between Thanos releases.

We make no guarantees, or rather we can be as stable as Thanos is, so don't think we should care about compatibility that much, same as for other resources we provision. WDYT?

mikonse:
Having the ability of adding additional arguments would immediately solve #3059, as this would enable configuring of TLS certificates for the thanos ruler and additional mounted volumes.

lilic:
I understand the importance just saying so far from what I remember we always just implemented any arguments as fields, worked for all of our resources so far.

Cc @s-urbaniak thoughts?

s-urbaniak:
@mikonse configuring TLS for the thanos-sidecar is possible today by patching the `thanos-sidecar` container, we do this here: https://github.com/s-urbaniak/cluster-monitoring-operator/blob/5ac6e400673bf2824df2af36f79574829af1f46b/assets/prometheus-k8s/prometheus.yaml#L115-L127

@pgier as @lilic mentioned i believe this should already be possible by patching the existing container, no?

mikonse:
@s-urbaniak that is true (an I'm using that exact approach currently), however it is not possible to configure tls paramters for the thanos ruler component if deploying it using the ThanosRuler CRD.
Since thanos query currently only supports full tls encryption on all stores it is using or none at all it is necessary to configure tls on the ruler as well.

lilic:
> configure tls paramters for the thanos ruler component if deploying it using the ThanosRuler CRD.

then lets add that param, or is that not possible?

s-urbaniak:
@lilic that is afaik not possible right now, and @pgier is going to add support for this.

lilic:
Yep, for now I would suggest closing this PR and adding native support for any missing arguments we need. We can add support for the TLS, you are welcome to open a PR for that @mikonse, in the mean time there is a known workaround for this as you mentioned you are already using it.

pgier:
> @pgier as @lilic mentioned i believe this should already be possible by patching the existing container, no?

It's possible, it's just a bit difficult because patching container args does an overwrite instead of a merge.  So it means you can't just add a single unsupported arg to your Thanos ruler, you have to re-specify all the necessary arguments, and basically ignore any configured arguments.

pgier:
Closing this since we're not planning to support arbitrary args for now.

PR #3075, opened 2020-03-10
GRPC TLS config for Thanos Ruler and Sidecar

### LINKED ISSUE #3059: Need TLS-config for Thanos

pgier:
Fixes #3059 
I used the existing TLSConfig type which is currently used by several of the other custom resources.
For simplicity, this currently only supports using the `*File` fields of TLSConfig.  In the future we could also add support for the `SecretOrConfigMap` fields.


example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
```diff
@@ -1698,6 +1698,126 @@ spec:
...
+                  properties:
+                    configMap:
+                      description: ConfigMap containing data to use for the targets.
+                      properties:
+                        key:
+                          description: The key to select.
+                          type: string
+                        name:
+                          description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                            TODO: Add other useful fields. apiVersion, kind, uid?'
```
lilic: TODO? 🤔

pkg/prometheus/statefulset.go
```diff
@@ -775,17 +775,31 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
...
 		}
 
+		thanosArgs := []string{"sidecar",
+			fmt.Sprintf("--prometheus.url=http://%s:9090%s", c.LocalHost, path.Clean(webRoutePrefix)),
+			fmt.Sprintf("--tsdb.path=%s", storageDir),
+			fmt.Sprintf("--grpc-address=%s:10901", bindAddress),
+			fmt.Sprintf("--http-address=%s:10902", bindAddress),
+		}
+
+		if p.Spec.Thanos.GRPCServerTLSConfig != nil {
```
lilic: What are we configuring here?

pkg/thanos/statefulset.go
```diff
@@ -259,6 +259,19 @@ func makeStatefulSetSpec(tr *monitoringv1.ThanosRuler, config Config, ruleConfig
 		})
 	}
 
+	if tr.Spec.GRPCServerTLSConfig != nil {
+		tls := tr.Spec.GRPCServerTLSConfig
+		if tls.CertFile != "" {
```
lilic: What happens if one of these fields is empty, should we just let it error out on thanos ruler side?

example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
pgier: This is coming from the Kube API, I haven't looked into this too deeply, but will likely go away in the next update of the crd generation tools.

pkg/prometheus/statefulset.go
pgier: This is the config for the Thanos Sidecar attached to Prometheus.  It has the same tls options as ruler.

pkg/thanos/statefulset.go
pgier: Yes, I was thinking we can just pass along the parameters as they are.  I wasn't sure how thanos would handle a partial set of parameters, so I didn't yet add any validation.

bwplotka: Ruler checks that

lilic: Then this looks good to me, as thanos validates we don't want to do any validations.

bwplotka: All components use this: https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/tls/options.go#L17

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
```diff
@@ -4486,6 +4486,128 @@ spec:
...
+                          properties:
+                            key:
+                              description: The key to select.
+                              type: string
+                            name:
+                              description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                TODO: Add other useful fields. apiVersion, kind, uid?'
+                              type: string
+                            optional:
+                              description: Specify whether the ConfigMap or its key
```
bwplotka: missing trailing period

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
```diff
@@ -4486,6 +4486,128 @@ spec:
                 baseImage:
                   description: Thanos base image if other than default.
                   type: string
+                grpcServerTlsConfig:
+                  description: 'GRPCServerTLSConfig configures the GRPC server from
```
bwplotka: I think it's called `gRPC`? 
image

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
```diff
@@ -4486,6 +4486,128 @@ spec:
...
+                    which Thanos Querier reads recorded rule data. Note: currently
```
bwplotka:
```suggestion
                    which Thanos Querier reads recorded rule data. Note: Currently
```

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
```diff
@@ -4486,6 +4486,128 @@ spec:
...
+                    only the CAFile, CertFile, and KeyFile fields are supported. Maps
```
bwplotka: What else is not? :thinking:

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
```diff
@@ -4486,6 +4486,128 @@ spec:
...
+                            key:
+                              description: The key of the secret to select from.  Must
+                                be a valid secret key.
+                              type: string
+                            name:
+                              description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                TODO: Add other useful fields. apiVersion, kind, uid?'
+                              type: string
+                            optional:
+                              description: Specify whether the Secret or its key must
```
bwplotka: ditto

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
```diff
@@ -4486,6 +4486,128 @@ spec:
...
+                          properties:
+                            key:
+                              description: The key to select.
+                              type: string
+                            name:
+                              description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                TODO: Add other useful fields. apiVersion, kind, uid?'
+                              type: string
+                            optional:
+                              description: Specify whether the ConfigMap or its key
```
bwplotka: ditto

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
```diff
@@ -4486,6 +4486,128 @@ spec:
...
+                            key:
+                              description: The key of the secret to select from.  Must
+                                be a valid secret key.
+                              type: string
+                            name:
+                              description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                TODO: Add other useful fields. apiVersion, kind, uid?'
+                              type: string
+                            optional:
+                              description: Specify whether the Secret or its key must
```
bwplotka: ditto

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
bwplotka: also I think we can remove this line. We even have `insecureSkipVerify`

example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
```diff
@@ -1698,6 +1698,126 @@ spec:
                 under. This is necessary to generate correct URLs. This is necessary
                 if Thanos Ruler is not served from root of a DNS name.
               type: string
+            grpcServerTlsConfig:
+              description: 'GRPCServerTLSConfig configures the GRPC server from which
+                Thanos Querier reads recorded rule data. Note: currently only the
+                CAFile, CertFile, and KeyFile fields are supported. Maps to the ''--grpc-server-tls-*''
```
bwplotka: ditto

example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
```diff
@@ -1698,6 +1698,126 @@ spec:
...
+                    key:
+                      description: The key of the secret to select from.  Must be
+                        a valid secret key.
+                      type: string
+                    name:
+                      description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                        TODO: Add other useful fields. apiVersion, kind, uid?'
+                      type: string
+                    optional:
+                      description: Specify whether the Secret or its key must be defined
```
bwplotka: ditto everywhere in this file

pkg/apis/monitoring/v1/thanos_types.go
```diff
@@ -165,6 +165,11 @@ type ThanosRulerSpec struct {
 	ExternalPrefix string `json:"externalPrefix,omitempty"`
 	// The route prefix ThanosRuler registers HTTP handlers for. This allows thanos UI to be served on a sub-path.
 	RoutePrefix string `json:"routePrefix,omitempty"`
+	// GRPCServerTLSConfig configures the GRPC server from which Thanos Querier reads
+	// recorded rule data.
+	// Note: currently only the CAFile, CertFile, and KeyFile fields are supported.
```
bwplotka: ditto

pkg/prometheus/statefulset.go
```diff
@@ -775,17 +775,31 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
...
+			tls := p.Spec.Thanos.GRPCServerTLSConfig
+			if tls.CertFile != "" {
+				thanosArgs = append(thanosArgs, "--grpc-server-tls-cert="+tls.CertFile)
+			}
+			if tls.KeyFile != "" {
+				thanosArgs = append(thanosArgs, "--grpc-server-tls-key="+tls.KeyFile)
+			}
+			if tls.CAFile != "" {
```
bwplotka: what about `insecureSkipVerify`? We had this somewhere defined

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
pgier: This one is coming from the K8S libraries so would have to be fixed there.

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
pgier: Fixed

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
pgier: Done

example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
pgier: Done

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
pgier: I'm re-using the TLSConfig type which is used in a few places to configure TLS for prometheus endpoints.  In addition to the file options, it also includes fields `CA`, `Cert`, and `KeySecret` which allow direct usage of secrets or config maps instead of needing to manually mount the secret.  So we can add support for this to the Thanos types at a later point if we want.

The type also includes insecureSkipVerify which is not available as a CLI arg to Thanos so it's ignore if it's set here.

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
pgier: Coming from k8s api.

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
pgier: Coming from k8s api.

pkg/prometheus/statefulset.go
pgier: It's used by Prometheus endpoint config but not by the Thanos components.

example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
bwplotka: :+1:

pkg/prometheus/statefulset.go
bwplotka: ah thanks

PR #3078, opened 2020-03-12
pkg/prometheus: deprecate pod monitor targetPort

### LINKED ISSUE #3071: PodMonitor PodMetricsEndpoint port and targetPort do the same thing
Follow up from PR #3064 

Looks like port and targetPort from PodMonitor does the same thing: https://github.com/coreos/prometheus-operator/blob/v0.37.0/pkg/prometheus/promcfg.go#L475-L482 (except port does not support a number). This makes sense as PodMonitor is all about pods, so no service port, like in ServiceMonitor. I believe one of the two should be deprecated, IMO targetPort is term from Service, so this one should be dropped.

pgier:
The split between `port` and `targetPort` makes sense for `ServiceMonitors` because this determines if we filter endpoints based on `endpoint_port_name` or `pod_container_port_name`.  However, for `PodMonitors` this split doesn't make sense because there is no `endpoint_port_name`.  So we only need a single port config field.

Fixes #3071


s-urbaniak:
LGTM :+1:

PR #3080, opened 2020-03-12
pkg/alertmanager: mergepatch alertmangaer containers

pgier:
Instead of just appending user defined containers to the alertmanager
pods, use k8sutil.MergePatchContainers so that the default containers
can be modified.

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -471,6 +471,52 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 	terminationGracePeriod := int64(120)
 	finalLabels := config.Labels.Merge(podLabels)
 
+	defaultContainers := []v1.Container{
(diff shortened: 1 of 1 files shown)


s-urbaniak:
lgtm modulo a question: does this change the current standard behavior in any way, i.e. are existing declarations potentially affected somehow?

pgier:
The current behaviour is to just append the list of containers.  With the mergepatch change we still append containers unless the name matches an existing container.  The only cases where this would affect an existing configwould be where one or more of the patch containers matches the name of one of the default containers `alertmanager` or `config-reloader`.  Which might cause an error currently, since I'm not sure kube allows two containers with the same name?  So I don't think there is a significant risk of breaking existing configs.  

Also, this brings the alertmanager use of the containers field in-line with prometheus and thanos ruler, so the behaviour is more consistent between the operators.

s-urbaniak:
LGTM, thanks for the rationale!

PR #3084, opened 2020-03-14
consolidate container image defaults

pgier:
This moves the default image versions and urls from multiple
packages into a single location in the operator package.
This has the advantage of easier maintenance and ensuring that
the default Thanos sidecar image doesn't become out of
sync with the default image used for Thanos Ruler.

This is a subset of #3003, to try to reduce the complexity of that one.


pgier:
cc @coreos/team-monitoring

s-urbaniak:
/lgtm
that is a very good idea, thank you!

PR #3091, opened 2020-03-20
pkg/apis: Omit QueryEndpoints if empty

lilic:
As queryEndpoints is not required and query can be configured via other
fields such as queryConfig, so this should have a possibility to be
empty.

cc @coreos/team-monitoring


s-urbaniak:
lgtm

PR #3092, opened 2020-03-20
Release 0.38

pgier:


CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 ## Next release
 
+## 0.38.0 / 2020-03-20
+
 * [CHANGE] Changed ThanosRuler custom resource field alertmanagersURL type from string to []string` (#3067)
+* [CHANGE] Deprecate PodMonitor targetPort field (#3071)
+* [FEATURE] Add queryConfig field to ThanosRuler spce (#3068)
```
lilic:
```suggestion
* [FEATURE] Add queryConfig field to ThanosRuler spec (#3068)
```

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
...
+* [FEATURE] GRPC TLS config for Thanos Ruler and Sidecar (#3059)
+* [FEATURE] MergePatch Alertmanager containers (#3080)
+* [FEATURE] Add VolumeMounts to Prometheus custom resources (#2871)
+* [ENHANCEMENT] Update Thanos to v0.11.0 (#3066)
+* [ENHANCEMENT] Clarify that Endpoint.targetPort is pod port (#3064)
+* [ENHANCEMENT] Omit QueryEndpoints if empty (#3091)
```
lilic: This is bugfix IMO, as it doesn't work right now if you leave out QueryEndpoints empty, but works in thanos ruler.

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 ## Next release
 
+## 0.38.0 / 2020-03-20
+
 * [CHANGE] Changed ThanosRuler custom resource field alertmanagersURL type from string to []string` (#3067)
+* [CHANGE] Deprecate PodMonitor targetPort field (#3071)
```
lilic: This links to a closed issuee 🤔 
https://github.com/coreos/prometheus-operator/issues/3071

lilic: Did you mean https://github.com/coreos/prometheus-operator/pull/3078?
Note discussion here, should that have been merged 🤔 https://github.com/coreos/prometheus-operator/issues/3089
cc @brancz

CHANGELOG.md
pgier: Fixed, thanks!

CHANGELOG.md
pgier: Change to BUGFIX, thanks!

CHANGELOG.md
pgier: #3071 was intentional since that is the reported issue that prompted the change.  I was thinking about whether we should put the issue number or pr number or both in the changelog entries.  I think it makes sense to put the issue number since that usually describes the reason for the change from the user's perspective, but maybe I we should also include the PR number?

The change doesn't break anything in the API, it just discourages use of the field by marking it deprecated in the docs and logging a warning when it's used.  So I think it's safe to include in the release.

lilic: Sorry my mistake, thought we removed it.

Makes sense but can we add both link to issue and PR the change was made in to the changelog line.

pgier: Done!

PR #3101, opened 2020-03-26
use tolerant parsing of version strings

pgier:
From the godoc:

ParseTolerant allows for certain version specifications that do not
strictly adhere to semver specs to be parsed by the semver library.
It currently trims spaces, removes a "v" prefix, and adds a
0 patch number to versions with only major and minor components specified.


brancz:
Nice!

PR #3105, opened 2020-03-29
Do not modify the PrometheusRule cache object

lilic:
As we modify the PrometheusRule when generating the content, we need to make first a DeepCopy of the PrometheusRule object to be able to modify in-place. Currently we were modifying the cached PrometheusRule object, which caused changes to not propagate to the orginal object.

Side note: we need to refactor the prometheus and thanos ruler common code to avoid not missing fixing these kids of errors in the future!

cc @coreos/team-monitoring


simonpasquier:
:+1:

s-urbaniak:
lgtm

PR #3112, opened 2020-04-02
Thanos sidecar log level and format

### LINKED ISSUE #3077: Extend Thanos sidecar container args without replacing them
**What is missing?**
The ability to add extra args to containers (eg. PrometheusSpec.spec.containers) without having to replace the `args` array completely.

**Why do we need it?**
We wanted to enable JSON logging format to Thanos (`--log.format=json`), but the only way to do this is to replace the entire `args` array. At that point, the value added by the operator is significantly reduced.

I think this is simply a consequence of the strategic merge patch. Both cases would be valid -- replacing the `args` array, or extending it.

pgier:
Fixes #3077


brancz:
needs rebase, otherwise lgtm

PR #3115, opened 2020-04-02
Retain metadata for embedded PVC objects

### LINKED ISSUE #3093: Prometheus PVC metadata ignored
**What happened?**

I have metadata (name and labels) set on on Prometheus PVC.
Once applied, the metadata property is empty

I expect the metadata to be filled.

It worked with v0.34.0 : 

```
user@master-1:~$ kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.34/bundle.yaml
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply
clusterrolebinding.rbac.authorization.k8s.io/prometheus-operator configured
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply
clusterrole.rbac.authorization.k8s.io/prometheus-operator configured
deployment.apps/prometheus-operator created
serviceaccount/prometheus-operator created
service/prometheus-operator created


user@master-1:~$ kubectl apply -f prometheus.yaml
prometheus.monitoring.coreos.com/k8s created

user@master-1:~$ kubectl get prometheuses.monitoring.coreos.com k8s -o yaml
kind: Prometheus
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"monitoring.coreos.com/v1","kind":"Prometheus","metadata":{"annotations":{},"name":"k8s","namespace":"default"},"spec":{"storage":{"volumeClaimTemplate":{"metadata":{"label":{"test":"value"},"name":"prometheus-k8s-pv-claim"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"4G"}},"selector":{"matchLabels":{"app.kubernetes.io/component":"prometheus"}},"storageClassName":"glusterfs"}}}}}
  creationTimestamp: "2020-03-22T11:07:04Z"
  generation: 1
  name: k8s
  namespace: default
  resourceVersion: "74683"
  selfLink: /apis/monitoring.coreos.com/v1/namespaces/default/prometheuses/k8s
  uid: 50ca5b16-484f-4c17-bfbb-a09f60d11942
spec:
  storage:
    volumeClaimTemplate:
      metadata:
        label:
          test: value
        name: prometheus-k8s-pv-claim
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 4G
        selector:
          matchLabels:
            app.kubernetes.io/component: prometheus
        storageClassName: glusterfs
```

And starting with v0.35.0 the metadata field is empty : 

```
user@master-1:~$ kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/release-0.35/bundle.yaml
clusterrolebinding.rbac.authorization.k8s.io/prometheus-operator created
clusterrole.rbac.authorization.k8s.io/prometheus-operator created
deployment.apps/prometheus-operator created
serviceaccount/prometheus-operator created
service/prometheus-operator created

user@master-1:~$ kubectl apply -f prometheus.yaml
prometheus.monitoring.coreos.com/k8s created

user@master-1:~$ kubectl get prometheuses.monitoring.coreos.com k8s -o yaml
kind: Prometheus
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"monitoring.coreos.com/v1","kind":"Prometheus","metadata":{"annotations":{},"name":"k8s","namespace":"default"},"spec":{"storage":{"volumeClaimTemplate":{"metadata":{"label":{"test":"value"},"name":"prometheus-k8s-pv-claim"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"4G"}},"selector":{"matchLabels":{"app.kubernetes.io/component":"prometheus"}},"storageClassName":"glusterfs"}}}}}
  creationTimestamp: "2020-03-22T11:07:58Z"
  generation: 1
  name: k8s
  namespace: default
  resourceVersion: "74920"
  selfLink: /apis/monitoring.coreos.com/v1/namespaces/default/prometheuses/k8s
  uid: 49e7b0da-2a89-416d-a7e3-47015f419b14
spec:
  storage:
    volumeClaimTemplate:
      metadata: {}
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 4G
        selector:
          matchLabels:
            app.kubernetes.io/component: prometheus
        storageClassName: glusterfs
```

(same result with v0.38.0)

**Environment**

* Prometheus Operator version: v0.35.0 and v0.38.0

* Kubernetes version information: v1.17.4

* Kubernetes cluster kind: kubeadm

* Manifests:

My minimal prometheus.yaml : 

```
kind: Prometheus
metadata:
  name: k8s
spec:
  storage:
    volumeClaimTemplate:
      metadata:
        name: prometheus-k8s-pv-claim
        label:
          test: value
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 4G
        selector:
          matchLabels:
            app.kubernetes.io/component: prometheus
        storageClassName: glusterfs

```

pgier:
This allows nested PersistentVolumeClaim fields to be validated and maintain metadata information such as labels and annotations.

Fixes #3093


brancz:
This is a breaking change, I think think we can do this. Existing manifests will be invalid.

pgier:
@brancz maybe we need to create a v2 CRD?  That could also allow us to use the v1 kube crd api as described in #2912

joejulian:
> This is a breaking change, I don’t think we can do this. Existing manifests will be invalid.

Existing manifests are already invalid. Right now any manifest that expected to use a metadata.name is having that data silently discarded which is, in my view, far worse than an existing manifest failing validation after the repair.

It doesn't break installed resources (internally, api-machinery already has the apiVersion and kind), but yes, it will prevent a yaml file that doesn't have those fields from being installed as the validation will fail explaining that they need the apiVersion and kind. The error returned is quite specific about what needs to be changed in the manifest. If using the helm chart (currently also broken) adding those fields to the template will allow helm users to use the repaired CRD seamlessly.

brancz:
> Existing manifests are already invalid. Right now any manifest that expected to use a metadata.name is having that data silently discarded which is, in my view, far worse than an existing manifest failing validation after the repair.

Then we accidentally did a breaking change, which would be a bug :) 

I would prefer if we could get back to a state where providing these rather obscure fields is not necessary, just like in the StatefulSet's volumeClaimTemplate itself.

pgier:
I've updated the PR with a different approach using a custom embedded pvc type.  This is similar to the previous fix to the pod metadata and should allow us to maintain backwards compatibility and also fix the missing metadata.
@brancz @joejulian WDYT?

brancz:
I still don’t understand why other kube resources can do this and we can’t. I feel we’re just missing the right annotation.

pgier:
@brancz I'm not sure how other resources are doing this, other than some custom handling of the objects in the kube API maybe?  Prior to the switch to using kube-builder/controller-gen our generation tool was generating the full specification for the metadata (https://github.com/coreos/prometheus-operator/pull/2855/files#diff-194feaadcbc280d6b40a3945a47a40abL3871).  But controller-gen strips out the nested metadata (https://github.com/kubernetes-sigs/controller-tools/issues/385)

brancz:
We've had a number of CRD problems lately, I think I'd like to resolve those first before we continue to do a lot of changes like this.

brancz:
We discussed our options with some other kube folks and looks like this is actually our best bet given the state of the tooling. This lgtm once CI is green.

gracedo:
@pgier @brancz Any chance we can get this moving along? Thanks!

pgier:
I'll rebase this one and try to get it merged shortly after #3155 is merged

pgier:
@lilic I added an e2e test, and also tested manually that going from the current CRD to this updated CRD seems to work fine.  The Prometheus object using the current CRD lost the labels and annotations, but after updating the CRD and then re-applying the prometheus config, the labels and annotations were correctly retained.

lilic:
> and then re-applying the prometheus config, the labels and annotations were correctly retained.

What do you mean by prometheus config?

pgier:
@lilic I mean I reapplied the Prometheus custom resource.  Because the existing resource, created while the current CRD was in place, didn't retain the labels and annotations.

lilic:
@pgier got it, did it retain the rest of the information?

pgier:
@lilic yes, everything else was retained as far as I could tell.

brancz:
I'm curious about something. Could we test adding for example `apiVersion` to a CR with the current CRD (_and_ prometheus-operator version .. let's say just use latest release). And then "upgrade" to head of this PR. Somehow I expected the prometheus operator to fail to parse that type of CR as it's already persisted with that data, but the struct now doesn't contain it anymore.

joejulian:
The operator uses the struct. This only changes how the struct is converted to and from yaml.

pgier:
@brancz `apiVersion` and `kind` are part of the `TypeMeta` which is not changing here, so all the versions should be compatible with each other.  We're not removing any fields from the CRD, only adding them.  I tested this to verify that using the newer prometheus operator struct is able to parse prometheus resources both with and without the apiVersion and kind fields.

One case I wasn't sure about was if we had a pre controller-gen (v0.35.0) CRD which includes some of the extra `ObjectMeta` fields such as `deletionGracePeriodSeconds`, would that cause the new prometheus-operator to fail to parse because those fields aren't recognized in the struct.  So I tested this, and the new prometheus-operator seems to parse this fine even with the extra fields.  The extra fields are just ignored which I think is the behaviour that we want.  Also when applying this new CRD over a pre-controller-gen CRD, the extra fields are removed from any existing prometheus resources, and the annoations and labels are preserved.

So

lilic:
lgtm 👍 thanks!

brancz:
thanks for double checking, this is a sensitive part that we've had many problems with in the past, that's he only reason I want to make sure we do our best not to introduce new regressions :)

c10l:
Any chance this will be backported to v0.38? We're currently on kubernetes 1.15 so we're unable to upgrade to prometheus-operator v0.39.

lilic:
@c10l this should be in 0.38 release already https://github.com/coreos/prometheus-operator/commit/0fc8b1beacc5f961eb5e52ae68fea23737adec0f Are you saying you are still hitting this problem? If so can you open a new issue, thank you so much!

lambdanis:
I see this fix was backported to 0.38 and older release branches, are you planning to create patch releases for them?

simonpasquier:
There might be a 0.38.2 release at some point because it's the last version to support Kubernetes <= 1.15 (see https://github.com/prometheus-operator/prometheus-operator/pull/3300 to follow the discussion).

lambdanis:
Nice, thanks @simonpasquier !

PR #3123, opened 2020-04-06
Fix inaccuracy in ruleSelector description

damienleger:
Current documentation implies that ruleSelector is for alert rules but there is two kind of rules in prometheus: alert rules & recording rules. I think this is a inaccurate description or at least a misleading one.


pgier:
Thanks for the contribution!  These docs are automatically generated from the type definitions (https://github.com/coreos/prometheus-operator/blob/master/pkg/apis/monitoring/v1/types.go#L196).  Can you update the go sources and then regenerate with `make generate-in-docker`?

damienleger:
This is done according to your instructions @pgier 🙂

pgier:
lgtm, thanks!

PR #3126, opened 2020-04-06
pkg/thanos: fix thanos-ruler-operated service

simonpasquier:
This change uses the named port instead of the numerical value for the
governing service. It also ensures that the container port isn't exposed
when ListenLocal is true.


s-urbaniak:
lgtm

PR #3128, opened 2020-04-06
prometheus-config-reloader: support for watching rules directory

leoluk:
Add --rules-dirs to expose the directory watching feature of the reloader library.

### DIFF
--- a/cmd/prometheus-config-reloader/main.go
+++ b/cmd/prometheus-config-reloader/main.go
@@ -51,6 +51,8 @@ func main() {
 	cfgSubstFile := app.Flag("config-envsubst-file", "output file for environment variable substituted config file").
 		String()
 
+	rulesDir := app.Flag("rules-dir", "Rules directory to watch non-recursively").Strings()
+
 	createStatefulsetOrdinalFrom := app.Flag(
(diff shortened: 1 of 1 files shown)


pgier:
Thanks for the contribution, but this is currently failing CI.  I'm not sure why it's failing, but I'm guessing this is because when no `rulesDir` is set, the call to `reloader.New()` is passing a slice with a single empty string value instead of an empty slice.  You might need to just add a check for empty string and pass an empty slice in that case.

leoluk:
Yup, that's what happened. Hooray for tests!

leoluk:
Checks are now passing, there was a much easier way to do this using Kingpin's repeatable arguments.

brancz:
I feel like I’m missing context. Is there an issue or something why we needed this?

leoluk:
Without this flag, the reloader is only able to catch modifications to prometheus.yml, not individual rules files. I'm using the reloader as a standalone container - how does the operator solve this?

brancz:
We use the configmap reloader instead. But we did say that one day we wanted to merge them back together.

brancz:
RE https://github.com/coreos/prometheus-operator/issues/1580

PR #3133, opened 2020-04-08
Remove status field from CRDs

yann-soubeyrand:
Having the status field on CRDs prevents continuous delivery tools from
converging when applying them.


pgier:
These CRDs are generated by the controller-gen command (https://github.com/kubernetes-sigs/controller-tools), so the tool would need to be configured to avoid generating these fields.  I'm not sure if this is currently possible with controller-gen.

yann-soubeyrand:
Yes I've just seen the issue you opened for creationTimestamp.

mcristina422:
I've updated controller-tools with a potential fix https://github.com/kubernetes-sigs/controller-tools/pull/433

yann-soubeyrand:
> I've updated controller-tools with a potential fix [kubernetes-sigs/controller-tools#433](https://github.com/kubernetes-sigs/controller-tools/pull/433)

Thanks a lot!

PR #3134, opened 2020-04-08
cmd/operator: add tls config for web server

pgier:


brancz:
We need to also allow rotating the various certificates as well as various other TLS configuration knobs. I recommend having a look at https://github.com/observatorium/observatorium/pull/24/files which largely does the same thing.

We should also have an e2e test for at least the basic functionality of this.

pgier:
@brancz Added the reloader and updated the tests, PTAL

pkg/operator/server.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		level.Info(logger).Log("msg", "TLS disabled key and cert must be set to enable")
+
+		return nil, nil
+	}
+
+	level.Info(logger).Log("msg", "enabling server side TLS")
+
+	tlsCert, err := tls.LoadX509KeyPair(certFile, keyFile)
```
brancz, line 56: the certificates are already loaded with the GetCertificate method, why do we need to load them explicitly here?

pgier: You're  right, they will be loaded by the reloader later, although it might be useful to load them here during startup just to make sure they are valid files?

brancz: We have no guarantee that that will continue to be the case, I'd rather have the behavior be consistent, and always log. Plus I think if a certificate is set the go TLS listener will not even attempt to load the certificate via the `GetCertificate` function.

pgier: Ok, sounds good.  I think your suggestion also makes sense in case the secrets are not immediately mounted to the container.  I was hoping to activate TLS automatically without needing another flag by detecting if the relevant files are available, but it seems like it would be better to just have an explicit TLS activation flag, so I'll add that.

I tested that the listener will still reload even if the cert fields are set, but I agree it's better to just always reload the same way for consistency.

brancz: Nice :+1:

PR #2554, opened 2019-04-12
Remove kube-prometheus and add README with explanation

metalmatze:
As discussed in #2553, we want to move kube-prometheus out of this repository.

This PR deletes the entire `contrib/kube-prometheus` folder and adds a README.md with some explanation.
It's probably easiest to review the new [README on my branch](https://github.com/metalmatze/prometheus-operator/blob/move-kube-prometheus/contrib/kube-prometheus/README.md).

As discussed with @brancz, once we merge this PR we will close all related kube-prometheus issues and PRs asking the original authors to recreate those at [coreos/kube-prometheus](https://github.com/coreos/kube-prometheus).
And as we are on it, we can just close all Helm related issues and PRs too.

/cc @mxinden @squat @paulfantom @s-urbaniak

PR #3136, opened 2020-04-09
restore storage.md file

sdarwin:
After looking for documentation about persistent storage, it was surprising to see the files had existed and then been deleted.   Thankfully, git is persistent storage.  :-)

From https://github.com/coreos/prometheus-operator/issues/2905
"It seems like we should bring those docs back."  
"Would you like to shoot us a PR with fixed links re-introducing these docs?"


pgier:
Thanks for working on this!  LGTM.  Looks like this doc was removed in #2554 but maybe they never made it over to kube-prometheus repo.  I think we are over due for an update of the docs in this repo in general.

PR #3138, opened 2020-04-11
Allow setting the cluster domain so alertmanager makes fewer DNS calls

### LINKED ISSUE #2934: DNS definition by Kubernetes affects alertmanager

SerialVelocity:
Fixes #2934. This is a more complete version of #2936.

There are two changes in this PR:
 - Allow specifying the cluster domain.
    - This allows using FQDNs for referencing the alertmanager so that the DNS search domain isn't used.
    - This is something that is global to the cluster so I placed it in the operator configuration instead of alertmanager configuration.
 - Remove the `.svc` extension on the alertmanager url when he cluster domain is not specified.
    - The default search domain is `.svc.<cluster-domain>` so this reduces the amount of DNS calls when the cluster domain isn't specified from 6 to 2 (1 for ipv4, 1 for ipv6).

(cc @oded-dd)

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -151,6 +151,7 @@ func init() {
 	flagset.Var(&cfg.CrdKinds, "crd-kinds", " - EXPERIMENTAL (could be removed in future releases) - customize CRD kind names")
 	flagset.BoolVar(&cfg.EnableValidation, "with-validation", true, "Include the validation spec in the CRD")
 	flagset.StringVar(&cfg.LocalHost, "localhost", "localhost", "EXPERIMENTAL (could be removed in future releases) - Host used to communicate between local services on a pod. Fixes issues where localhost resolves incorrectly.")
+	flagset.StringVar(&cfg.ClusterDomain, "cluster-domain", "", "The domain of the cluster. This is used to generate service FQDNs. If this is not specified, DNS search domain expansion is used instead.")
 	flagset.StringVar(&cfg.LogLevel, "log-level", logLevelInfo, fmt.Sprintf("Log level to use. Possible values: %s", strings.Join(availableLogLevels, ", ")))
 	flagset.StringVar(&cfg.LogFormat, "log-format", logFormatLogfmt, fmt.Sprintf("Log format to use. Possible values: %s", strings.Join(availableLogFormats, ", ")))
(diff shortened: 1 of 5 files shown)


brancz:
lgtm

brancz:
no idea why travis ci is not reporting status back

PR #3143, opened 2020-04-15
Add a disable subPath flag for prometheus-db volume

BarthV:
Hi !
We're facing an annoying issue because a subPath for prometheus-db is hardcoded for a long (long) time in this operator. While I completely understand the reasons and history for this to be, in some cases (like rancher + local-path-provisionner, or any other local path implementations ..) we want to be able to disable the use of such subPath.

Some workarounds have already been submitted, but they are now deprecated & I'd rather see some native options to disable subPath instead of applying a kubectl patch which is not really compliant with our gitops policy.

more infos : 
* https://vividcode.io/prometheus-operator-persistence-with-rancher-local-path-provisioner/

This should help fixing : 
* https://github.com/rancher/local-path-provisioner/issues/4
* https://github.com/rancher/rancher/issues/14836
* https://github.com/rancher/rancher/issues/17256
* + some other related issues on prometheus & openstack bug tracker

Don't hesitate to ask me some modifications & user documentation updates too (if it's relevant for you) !

thanks a lot


iwin4aids:
This issue annoy me for a long time.
In my test case, this prometheus operator can not work with any local provisioner provider~

brancz:
I dug around, and the original issue why we introduced this in the first place was: https://github.com/coreos/prometheus-operator/issues/29. Since the root cause was actually fixed in Prometheus, I wonder if we could even make this the new default, and only retain the subpath for those that have it set (for backward compatibility).

BarthV:
I didn't wanted to change the default behavior at this point, but I definitively agree on the idea to disable subPath by default at some point. We should poke some maintainers about that !

GlusterFS & NFS (and some) other compatibility historically required subPath, but I think it should remains an edge case option rather than the default option.

brancz:
> We should poke some maintainers about that !

I'm a maintainer of both projects :)

I think I'm ok with the change as is. Could you just mark the field as deprecated immediately? That way when we come around to do a major cleanup of things (maybe a new API version), that we know to take care of the behavior. Thanks!

BarthV:
:+1: I'm doing that

BarthV:
Added a deprecation notice as comment
`types.go`
```
// Deprecated: subPath usage will be disabled by default in a future release, this option will become unnecessary.
// DisableMountSubPath allows to remove any subPath usage in volume mounts.
DisableMountSubPath bool `json:"disableMountSubPath,omitempty"`
```

brancz:
Perfect :+1: thank you! :tada:

PR #3144, opened 2020-04-15
Support naming of remote write queues

### LINKED ISSUE #2932: Support prometheus v2.15 remote_write queue name config
**What is missing?**
Prometheus recently added support for giving a remote_write queue a name. We have no way to leverage this field when using the prometheus-operator.
https://github.com/prometheus/prometheus/pull/6043

**Why do we need it?**
Simply put: This is configuration on Prometheus that is impossible to configure using the prometheus-operator. 

In my case, I have multiple remote-write queues sending to the same URL, for reasons. Prometheus 2.15 doesn't allow this without disambiguating them. From the perspective of Prometheus, perhaps I'm using it wrong. Regardless, the first point stands.

**Environment**

* Prometheus Operator version: quay.io/coreos/prometheus-operator:v0.31.1

csmarchbanks:
Since version 2.15.0 of Prometheus users have been able to specify the
name of a remote write queue. This helps to differentiate queues in
metrics and logs where the default value is a hash.

Closes #2932


pkg/apis/monitoring/v1/types.go
```diff
@@ -483,25 +487,29 @@ type QueueConfig struct {
...
 // +k8s:openapi-gen=true
 type RemoteReadSpec struct {
-	//The URL of the endpoint to send samples to.
+	// The URL of the endpoint to send samples to.
 	URL string `json:"url"`
-	//An optional list of equality matchers which have to be present
+	// The name of the remote read queue, must be unique if specified. The name
+	// is used in metrics and logging in order to differentiate read
+	// configurations.  Only valid in Prometheus versions 2.15.0 and newer.
+	Name string `json:"name,omitempty"`
```
brancz, line 495: this still needs wiring up to the config generation

csmarchbanks: Do'h, added the wiring for read and write cases. Do you think this is worth a test as well? I am not seeing any other remote specific tests.

brancz: I'm happy with this as is. Thanks!

PR #3145, opened 2020-04-15
Backport #3125, #3129 and #3105 to release-0.38

simonpasquier:


pgier:
I believe the failing CI is due to using a different version of go-bindata.  Probably just need to delete the local one and then run `make generate` again.

simonpasquier:
@lilic I have no strong preference but maybe simpler to do it in this PR so I'll send an update.

@pgier thanks, that was it.

PR #3154, opened 2020-04-20
*: update kubernetes deps to v1.18

### LINKED ISSUE #3146: Migrate kubernetes clients to 1.18 release
Since v1.18.x is out, we should update the client-go, k/api, and k/apimachinery dependencies

To help with the migration, as there were breaking changes we should make use of https://github.com/kubernetes-sigs/clientgofix

joelanford:
This commit bumps the kubernetes version dependencies from v1.17
to v1.18.

The kubernetes client-go library introduced breaking
changes in v1.18 by adding context.Context to many of its function
signatures. The kubernetes-sigs/clientgofix tools was used in this
commit to automatically add the necessary context.TODO() arguments
to all client-go usage in the repo.

Fixes #3146


joelanford:
Yep! Looks like I need to make sure my go-bindata and generator binaries are up-to-date.

lilic:
Thanks, this lgtm on green and after rebase.

cc @coreos/team-monitoring any comments?

brancz:
This lgtm as well, just one thing: could you open an issue to replace the `context.TODO()`s with actual contexts? We should eventually pipe through a real context, but I'm fine with having `context.TODO()` for now as it doesn't make it worse than it was before, but opens the door for improvement.

brancz:
Thanks! 🎉

PR #3160, opened 2020-04-23
Allow setting the alertmanagers cluster.advertiseAddress

auhlig:
This PR allows setting the Alertmanagers `--cluster.advertiseAddress` via `Spec.ClusterAdvertiseAddress` if it's being used with a public address. Additional background can be found [here](https://github.com/prometheus/alertmanager/issues/1434).


pgier:
LGTM, I think this just needs a rebase

brancz:
I can see the use in non Kubernetes environments, but could you explain how you intend to use this in Kubernetes?

auhlig:
Rebased. 
We have 2 instances of the Alertmanager (AM) running in an HA setup. 
To avoid disruptions in the alerting chain due to problems in a single Kubernetes cluster, we're 2 separate clusters (one AM per cluster). 
The addresses being used by both are non-RFC1918. 
Thus we require to the `--cluster.advertiseAddress` being configurable.

sev3ryn:
@brancz, @pgier  shouldn't this parameter be be a slice of strings with size equal to number of replicas. Advertise address couldn't be same for all the replicas

TheMeier:
Is there any way to make this work for a scenario with 2  K8S clusters running with a alertmanager custom resource with 2 replicas in each K8S cluster?
How can you set separate IPs for the 2 pods, given that there is one dedicated Ingress/Metallb/etc external external IP per Pod provided.
Also, is there even any way to automatically fetching the assigned external IPs via discovery any discovery mechanism?

The only sensible way IMHO would be that the operator manages the Resource that provides the external ip, then the operator can inject the IPs from the status of the managed resource sinto the pod spec, eg as env vars:
ALERTMANAGER_CLUSTER_EXTERNAL_IP_0,ALERTMANAGER_CLUSTER_EXTERNAL_IP_1

PR #3163, opened 2020-04-24
Add ability to override global EnforcedNamespaceLabel setting in PrometheusRuleSpec

### LINKED ISSUE #2940: Supporting EnforcedNamespaceLabel in PrometheusRuleSpec
Currently , EnforcedNamespaceLabel is defined in PrometheusSpec which is global to all PrometheusRules.

Maybe we can add label similar to EnforcedNamespaceLabel in PrometheusRuleSpec , this way user can define their own rules in their own namespace.

I can help to create a PR if my proposal sounds reasonable .

Thanks
Ben

sev3ryn:
IgnoreNSLabelEnforcement parameter of  PrometheusRuleSpec allows to override
global setting EnforcedNamespaceLabel of PrometheusSpec.
This is useful for cases when some alerts needs to be defined once
for all namespaces

Fixes #2940


brancz:
See my comment at https://github.com/coreos/prometheus-operator/issues/2940#issuecomment-618837518

PR #3171, opened 2020-04-27
Add Custom Resource Definitions to bundle.yaml

dgrisonnet:
Follow-up PR to the following [discussion](https://github.com/coreos/prometheus-operator/commit/05a8934c678134bffc83c401b091e6d0df651d6e#r38725452).

cc @brancz


lilic:
Thanks, we also have to add the v1 and v1beta1 CRD manifests, correct?

dgrisonnet:
Yes, but this PR is mainly meant to fix what was missing in #3155. I'll move to v1beta1 in another PR.

lilic:
> Yes, but this PR is mainly meant to fix what was missing in #3155. I'll move to v1beta1 in another PR.

Yes, agreed! sgtm 👍

PR #3179, opened 2020-04-28
pkg/thanos: fix typo in statefulset informer

simonpasquier:
Spotted this typo while working on another issue...

### DIFF
--- a/pkg/thanos/operator.go
+++ b/pkg/thanos/operator.go
@@ -190,7 +190,7 @@ func New(conf prometheusoperator.Config, logger log.Logger, r prometheus.Registe
 	)
 
 	o.ssetInf = cache.NewSharedIndexInformer(
-		listwatch.MultiNamespaceListerWatcher(o.logger, o.config.Namespaces.PrometheusAllowList, o.config.Namespaces.DenyList, func(namespace string) cache.ListerWatcher {
+		listwatch.MultiNamespaceListerWatcher(o.logger, o.config.Namespaces.ThanosRulerAllowList, o.config.Namespaces.DenyList, func(namespace string) cache.ListerWatcher {
 			return cache.NewListWatchFromClient(o.kclient.AppsV1().RESTClient(), "statefulsets", namespace, fields.Everything())
 		}),
 		&appsv1.StatefulSet{}, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},


pkg/thanos/operator.go
```diff
@@ -190,7 +190,7 @@ func New(conf prometheusoperator.Config, logger log.Logger, r prometheus.Registe
 	)
 
 	o.ssetInf = cache.NewSharedIndexInformer(
-		listwatch.MultiNamespaceListerWatcher(o.logger, o.config.Namespaces.PrometheusAllowList, o.config.Namespaces.DenyList, func(namespace string) cache.ListerWatcher {
```
lilic, line 193: Interesting find! I would say this is a bug, if this is indeed the case, should it maybe go in the previous release? I have been seeing weird log issues about namespaces not allowed, curious if this is the trigger for it!

simonpasquier: > I would say this is a bug, if this is indeed the case, should it maybe go in the previous release?

I can do that. FWIW the problem arises when you configure different values for the `--prometheus-instance-namespaces` and `--thanos-ruler-instance-namespaces` flags (which we don't in CMO). In this case, prometheus-operator wouldn't reconcile immediately on changes to the Thanos Ruler statefulset (IIUC).

> I have been seeing weird log issues about namespaces not allowed, curious if this is the trigger for it!

If you're talking about the `get namespace to enqueue ThanosRuler instances failed: namespace \"openshift-user-workload-monitoring\" does not exist` error message then it is the bug I'm chasing down but it isn't related to this fix.

lilic: > If you're talking about the get namespace to enqueue ThanosRuler instances failed: namespace \"openshift-user-workload-monitoring\" does not exist error message then it is the bug I'm chasing down but it isn't related to this fix.

ha yes, was hoping it was this.

> In this case, prometheus-operator wouldn't reconcile immediately on changes to the Thanos Ruler statefulset (IIUC).

yep, my thinking as well, I would say we should at least have it in the previous release, not sure when thanos ruler was introduced or if its worth backporting, but it would cause problems to folks who configure different namespaces. Leaving judgment to you if it's worth backporting.

lilic:
lgtm 👍

pkg/thanos/operator.go
simonpasquier: See https://github.com/coreos/prometheus-operator/issues/3181 for the `get namespace to enqueue Prometheus instances failed` errors. I'm working on a fix.

Given we're one week away from the next RC, I'm not sure it's worth backporting.

PR #3182, opened 2020-04-29
*: create separate namespace informers if needed

### LINKED ISSUE #3181: "namespace <...> does not exist" errors in the logs
**What happened?**

Seeing errors like this in the logs when starting prometheus-operator with different values for the `--prometheus-instance-namespaces` and `--namespaces` arguments.

```
level=error ts=2020-04-29T09:04:15.585008464Z caller=operator.go:907 component=prometheusoperator msg="get namespace to enqueue Prometheus instances failed: namespace \"default\" does not exist"
level=error ts=2020-04-29T09:04:15.585030323Z caller=operator.go:907 component=prometheusoperator msg="get namespace to enqueue Prometheus instances failed: namespace \"default\" does not exist"
```

Yes, no error.

**Environment**

`master`

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.1", GitCommit:"7879fc12a63337efff607952a323df90cdc7a335", GitTreeState:"clean", BuildDate:"2020-04-08T17:38:50Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:50:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:

minikube

* Manifests:

```
...
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/component: controller
      app.kubernetes.io/name: prometheus-operator
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/component: controller
        app.kubernetes.io/name: prometheus-operator
        app.kubernetes.io/version: v0.38.0
    spec:
      containers:
      - args:
        - --logtostderr=true
        - --config-reloader-image=jimmidyson/configmap-reload:v0.3.0
        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.38.0
        - --prometheus-instance-namespaces=default
        - --namespaces=ns1
        image: quay.io/coreos/prometheus-operator:master
        imagePullPolicy: IfNotPresent
        name: prometheus-operator
...
```

* Prometheus Operator Logs:

See above

simonpasquier:
Service/pod monitors and rules resources watched by the operator
can live in different namespaces than the configmaps and secrets
associated to the Prometheus custom resources.

Before this change, we had only one namespace informer scoped to the
monitor/rule namespaces. Whenever a change of configmap or secret was
detected, the operator couldn't retrieve the associated namespace
because it didn't exist in the monitor/rule namespaces and the
reconciliation wouldn't happen.

With this change, separate namespace informers are used when the
monitors/rules and prometheus namespaces aren't identical.

Closes #3181

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -38,7 +38,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/watch"
@@ -264,17 +263,6 @@ func (c *Operator) enqueue(obj interface{}) {
 	}
 
 	c.queue.Add(key)
-}
-
-// enqueueForNamespace enqueues all Alertmanager object keys that belong to the
-// given namespace.
-func (c *Operator) enqueueForNamespace(ns string) {
-	cache.ListAll(c.alrtInf.GetStore(), labels.Everything(), func(obj interface{}) {
-		am := obj.(*monitoringv1.Alertmanager)
-		if am.Namespace == ns {
-			c.enqueue(am)
-		}
-	})
 }
 
 // worker runs a worker thread that just dequeues items, processes them
--- a/pkg/thanos/operator.go
+++ b/pkg/thanos/operator.go
@@ -62,11 +62,12 @@ type Operator struct {
 	mclient monitoringclient.Interface
 	logger  log.Logger
 
-	thanosRulerInf cache.SharedIndexInformer
-	nsInf          cache.SharedIndexInformer
-	cmapInf        cache.SharedIndexInformer
-	ruleInf        cache.SharedIndexInformer
-	ssetInf        cache.SharedIndexInformer
+	thanosRulerInf   cache.SharedIndexInformer
+	nsThanosRulerInf cache.SharedIndexInformer
+	nsRuleInf        cache.SharedIndexInformer
+	cmapInf          cache.SharedIndexInformer
+	ruleInf          cache.SharedIndexInformer
+	ssetInf          cache.SharedIndexInformer
 
 	queue workqueue.RateLimitingInterface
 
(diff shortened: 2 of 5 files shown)


pkg/alertmanager/operator.go
```diff
@@ -266,17 +265,6 @@ func (c *Operator) enqueue(obj interface{}) {
 	c.queue.Add(key)
 }
 
-// enqueueForNamespace enqueues all Alertmanager object keys that belong to the
-// given namespace.
-func (c *Operator) enqueueForNamespace(ns string) {
```
lilic, line 271: Is this just not used anywhere?

pkg/thanos/operator.go
```diff
@@ -739,10 +756,18 @@ func needsUpdate(pod *v1.Pod, tmpl v1.PodTemplateSpec) bool {
 	return false
 }
 
+func (o *Operator) enqueueForThanosRulerNamespace(nsName string) {
+	o.enqueueForNamespace(o.nsThanosRulerInf.GetStore(), nsName)
```
lilic, line 760: What happens if this is called but we never call this line before:
> 		go o.nsThanosRulerInf.Run(stopc)

pkg/alertmanager/operator.go
simonpasquier: yes :) I guess it was copied/pasted at some point from prometheus.

pkg/thanos/operator.go
simonpasquier: In this case, `o.nsThanosRulerInf` is equal to `o.nsRuleInf` for which `Run()` is called.

simonpasquier:
> Curious is there an e2e test for this we have around namespace informers where we could add a test case for when two different ns are added for each CR, just so we don't regress in the future?

Since the issue triggered only errors in the logs, I couldn't imagine a test that would detect this. We have e2e tests checking that `-alertmanager-instance-namespaces` and `-prometheus-instance-namespaces` work as expected but we have nothing for `-thanos-ruler-instance-namespaces`...

s-urbaniak:
let's merge then :-)

Issue #3137, opened 2020-04-10
Switch to apiextensions.k8s.io/v1 for CRDs


simonpasquier:
As reported in https://github.com/coreos/prometheus-operator/issues/3122, the latest version of sigs.k8s.io/controller-tools (>= v0.28.0) generate CustomResourceDefinition objects that aren't compatible with Kubernetes >= v1.18.

The [recommended](https://github.com/coreos/prometheus-operator/issues/3122#issuecomment-610421183) approach is to switch from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`.

lilic:
cc @dgrisonnet you said you were interested in some prom-operator things to work on, this is a good one for first start!

dgrisonnet:
Looks interesting, I'll tackle it :)

PR #3187, opened 2020-04-30
Remove apiextensions.k8s.io/v1beta1 CRDs

dgrisonnet:
Follow-up PR from this [discussion](https://github.com/coreos/prometheus-operator/issues/3178#issuecomment-621128023).

cc @pgier @brancz @lilic


s-urbaniak:
lgtm

benjaminhuo:
@dgrisonnet This mean prometheus operator v0.39 and above only supports K8s 1.16 +?
then what this PR used for? https://github.com/coreos/prometheus-operator/pull/3175 , it generates both v1 and v1beta1 CRDs

Thanks
Ben

dgrisonnet:
Yes, prometheus operator v0.39+ will only support k8s 1.16+.

The original plan, when we wanted to update CRDs to v1 was to preserve compatibility with the 4 latest versions of Kubernetes. Meaning that we needed to add a mechanism to generate both v1 and v1beta1 CRDs.

Unfortunately, we later found out that keeping v1beta1 was dragging us down when updating controller-tools, which was the main purpose of updating CRDs #3137. So in the end, we preferred getting rid of v1beta1 in favor of updating controller-tools and thus have further validations on the CRDs.

benjaminhuo:
Got it , thanks very much for these background info.

Ben

PR #3207, opened 2020-05-11
add ability to exclude rules from namespace label enforcement

sev3ryn:
fixes https://github.com/coreos/prometheus-operator/issues/2940


Documentation/api.md
```diff
@@ -314,6 +315,17 @@ PodMonitorSpec contains specification parameters for a PodMonitor.
 
 [Back to TOC](#table-of-contents)
 
+## PromRuleExcludeConfig
```
brancz, line 318: We tend to not abbreviate, especially for types.

sev3ryn:
fixed test cases and added tests for new functionality

sev3ryn:
I didn't noticed that changed functions are also used for Thanos rules - will fix it as soon as I can

brancz:
Actually now that you say it, we're going to have to add the same functionality to the Thanos ruler.

sev3ryn:
added functionality to ThanosRuler. 
I adhered style of operator - by this I mean almost "copy-paste" pieces of code between prometheus/rules and thanos/rule. Shouldn't I move nsLabelEnforcementExcludeList and its methods to separate package to follow DRY principle?

If not, could you please elaborate why - just to educate myself.

brancz:
> Shouldn't I move nsLabelEnforcementExcludeList and its methods to separate package to follow DRY principle?

Please feel free to do so, we wanted to do so anyways.

sev3ryn:
moved namespace label enforcement to shared library. Naming suggestions are welcome

sev3ryn:
@brancz ^^

brancz:
sorry looks like this needs a rebase

sev3ryn:
rebased

pkg/apis/monitoring/v1/types.go
```diff
@@ -315,6 +319,15 @@ type PrometheusSpec struct {
 	QueryLogFile string `json:"queryLogFile,omitempty"`
 }
 
+// PrometheusRuleExcludeConfig enables users to configure excluded PrometheusRule names and their namespaces
+// to be ignored while enforcing namespace label for alerts and metrics.
+type PrometheusRuleExcludeConfig struct {
+	// RuleNamespace - namespace of excluded rule
+	RuleNamespace string `json:"ruleNamespace,omitempty"`
```
brancz, line 326: this shouldn't be `omitempty` I think, as that marks the field as not required, but if an object of this type is specified then the namespace and name are mandatory, just the list itself is optional

pkg/namespace-labeler/labeler.go
```diff
@@ -0,0 +1,106 @@
...
+	excludeList     map[string]map[string]struct{}
+}
+
+// New - creates new Labeler
+// enforcedNsLabel - label name to be enforced for namespace
+// excludeConfig - list of namespace + PrometheusRule names to be excluded while enforcing adding namespace label
+func New(enforcedNsLabel string, excludeConfig []monitoringv1.PrometheusRuleExcludeConfig) *Labeler {
+
+	if enforcedNsLabel == "" || len(excludeConfig) == 0 {
+		return &Labeler{enforcedNsLabel: enforcedNsLabel}
```
brancz, line 39: can we add a comment here that this is essentially a no-op labeler? Just for me in 3months to understand what this is :)

pkg/prometheus/rules.go
```diff
@@ -208,34 +216,8 @@ func (c *Operator) selectRules(p *monitoringv1.Prometheus, namespaces []string)
...
-					return "", errors.Wrap(err, "failed to inject labels to expression")
-				}
-
-				promRule.Groups[gi].Rules[ri].Expr = intstr.FromString(parsedExpr.String())
-			}
-		}
-	}
+func generateContent(promRule monitoringv1.PrometheusRuleSpec) (string, error) {
+
 	content, err := yaml.Marshal(promRule)
```
brancz, line 221: I don't see us doing this previously, so I'm fine with this being a follow up, but we need to discard the `partial_response_strategy` (also shouldn't be snake case).

sev3ryn: you do - see [pkg/prometheus/rules.go#L214](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/rules.go#L214)

snake case is an issue of Thanos unfortunately. See [partial-response](https://github.com/thanos-io/thanos/blob/v0.4.0-rc.0/docs/components/rule.md#partial-response)

but it is an issue - I'm discarding this field for Thanos Rules also what I shouldn't do - will fix it

brancz:
Awesome! Thank you so much for this!

PR #3210, opened 2020-05-12
Update exposing-prometheus-and-alertmanager.md for nginx ingress 0.22.0+

LeoQuote:
Starting in Version 0.22.0, ingress definitions using the annotation nginx.ingress.kubernetes.io/rewrite-target are not backwards compatible with previous versions. In Version 0.22.0 and beyond, any substrings within the request URI that need to be passed to the rewritten path must explicitly be defined in a capture group.


brancz:
Thanks!

PR #3214, opened 2020-05-12
Add nodes/metrics to allowable resources

jpreese:
Pulled in the `RBAC` yaml for the prometheus-operator and saw that it couldn't target our kube state / node exporter metrics, complaining about a 403.


pgier:
Looks like this failed CI because it needs an update to the generated docs.
@jpreese can run `make generate` and commit the generated doc updates?

jpreese:
@pgier pushed!

PR #3225, opened 2020-05-14
remove maxRetries for queueConfig

0x46616c6b:
Prometheus removed the property `max_retries` since a while from the remote_write properties (see https://github.com/prometheus/prometheus/pull/5649/files). This PR reflects this removal in Prometheus.


pgier:
Thanks for the contribution!  Looks like this needs a `make generate` to update all the generated files and pass CI.
In our compatibility matrix chart we do still support older versions of Prometheus (https://github.com/coreos/prometheus-operator/blob/master/Documentation/compatibility.md#prometheus), so I'm not sure we want to remove it at this point, @brancz WDYT?

0x46616c6b:
Okay, maybe otherwise a deprecation notice would be helpful. Ran into that problem today and prometheus crashed with active "maxRetries" setting.

brancz:
We can't remove it from the CR, that's a breaking change to the v1 API. We can exclude the config option though in the generated config if the version of the prometheus server is >=2.11.

0x46616c6b:
Okay, make sense. Fully understand to support that for near future.

PR #3229, opened 2020-05-16
0.38 support absent_over_time - update to prometheus 2.18.1

### LINKED ISSUE #3087: Admission controller - absent_over_time parses as a failure
**What happened?**

When creating a PrometheusRule with an `absent_over_time` function it gets rejected by the admission controller as bad. This function was released in v2.16.0. Updating the prometheus dependency will get the latest version of rulefmt which will resolve this issue. https://github.com/coreos/prometheus-operator/blob/master/go.mod#L39

Prometheus Operator version:
v0.37.0

vsliouniaev:
The compatibility matrix indicates that prometheus-operator 0.38 should be compatible with prometheus 2.16, which has the [`absent_over_time`](https://prometheus.io/docs/prometheus/2.16/querying/functions/#absent_over_time) function.

Since this is the last release of prometheus-operator that supports kubernetes < 1.16 I think it might be worth updating the prometheus dependency to the current latest current release. Updating to 2.16 looks to be largely the same in terms of the necessary code changes.

Requires: https://github.com/openshift/prom-label-proxy/pull/28
Fixes #3087


brancz:
Happy with this, but leaving last call up to @lilic @pgier and or @simonpasquier

pkg/prometheus/rules.go
```diff
@@ -16,6 +16,7 @@ package prometheus
 
 import (
 	"fmt"
+	"github.com/prometheus/prometheus/promql/parser"
```
pgier, line 19: This seems like it moved to the wrong place, was this an automated format?

go.mod
```diff
@@ -8,34 +8,37 @@ require (
...
 	k8s.io/klog v1.0.0
 	sigs.k8s.io/controller-tools v0.2.4
 )
 
 replace (
-	github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.0.0-20190818123050-43acd0e2e93f
+	// Temporary until https://github.com/openshift/prom-label-proxy/pull/28 gets merged
+	github.com/openshift/prom-label-proxy => github.com/vsliouniaev/prom-label-proxy v0.0.0-20200518104441-4fd7fe13454f
+	// Prometheus 2.18.1
+	github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.0.0-20200507164740-ecee9c8abfd1
```
pgier, line 42: Is the replace directive still necessary here?  I'm wondering if we can just set the correct version in the import.

pgier: I guess we won't be able to remove this until prometheus starts including a v2/v3 in their go.mod file.

pkg/prometheus/rules.go
vsliouniaev: It was! fixed now

go.mod
vsliouniaev: Thanks for the pointer, looks like it's possible to set this in imports after all.

go.mod
```diff
@@ -8,34 +8,36 @@ require (
...
-	github.com/oklog/run v1.0.0
+	github.com/oklog/run v1.1.0
 	github.com/openshift/prom-label-proxy v0.1.1-0.20191016113035-b8153a7f39f1
-	github.com/pkg/errors v0.8.1
-	github.com/prometheus/client_golang v1.2.1
-	github.com/prometheus/prometheus v2.3.2+incompatible
-	github.com/stretchr/testify v1.4.0
+	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.5.1
+	github.com/prometheus/prometheus v1.8.2-0.20200507164740-ecee9c8abfd1
```
pgier, line 23: I wish there was a better way to do this, it seems weird to show the version as 1.8.2 but point to 2.18.1.  I guess we can fix this in a separate PR if there is any better way to do it.

simonpasquier: AFAIK there's nothing we can do here for now because `github.com/prometheus/prometheus` doesn't respect the Go modules contract (e.g. the module name should be `github.com/prometheus/prometheus/v2`). There's an upstream effort to split the Prometheus code base into smaller modules which would be tagged as v0.x to avoid  (similar to `k8s.io/client-go` for instance) but there's nothing available yet.

Issue #3223, opened 2020-05-14
Admission hooks block rules that promtool treats as valid


vsliouniaev:
#### What happened?
Submitting a rule with a non-string label passes promtool but fails admission hooks:

```
labels:
  product: 123
  highPriority: false
```

#### Did you expect to see something different?
There should be parity between what is allowed to simplify creating CI pipelines.

#### Suggested solution
It should be possible to use the mutating hook we already have to alter the rules so that label values are always strings. This is what happens when they are loaded in Prometheus anyway.

I can see an objection to this in CI terms because the resources you get in the cluster are not the same as you are applying but there is definite precedent for this with mutating hooks. If necessary this could be made optional, but I think the mutating behaviour should be the default.

brancz:
Can we first assess why this works in promtool and not in the webhook? If that doesn't help then I'm ok with this solution.

vsliouniaev:
It's due to the definition of the `PrometheusRule` CRD only allowing strings. Prometheus using yaml instead of json natively just treats this as strings. 

At first the bad rule would just break the operator, then additional CRD-related validation was added here: https://github.com/coreos/prometheus-operator/commit/1ac10b0be2123c22ffe16f71a83640855865f534.

vsliouniaev:
There is actually an additional issue with `absent_over_time` not being an accepted, which would need a prometheus dependency update.

PR #3230, opened 2020-05-16
Mutate rule labels, annotations to strings

vsliouniaev:
Prometheus natively unmarshals rule label and annotation values to strings, kubernetes does not. This creates an inconsistency in behaviour in cases where someone runs `promtool` on the contents of the `PrometheusRules.spec` and then tries to apply it. For example, this is fine in promtool but rejected by the operator webhook:

```yaml
groups:
  - name: invalid-in-operator
    interval: 15s
    rules:
    - alert: ValidInPrometheus
      expr: absent_over_time(up[5m])
      labels:
        bool: true
      annotations:
        num: 1
```

This PR updates the already existing mutating webhook, to patch annotation and label values to strings before this becomes a problem in the validating webhook.

There should be no negative impact due to the "unexpected" nature of this alteration:
- Established invalid rules would fail to be loaded by the operator anyway
- Previously rejected rules will be patched to a valid state

---

- Also update to using `admission/v1` because this is now available and doesn't have any impact.
- Fixes: #3223


brancz:
Very awesome! Lgtm but giving others a chance to review as well @lilic @simonpasquier @pgier

simonpasquier:
Don't you need to update [pkg/operator/defaults.go](https://github.com/coreos/prometheus-operator/blob/502d2d825923933ee7dfb0b26a746f823cff7855/pkg/operator/defaults.go#L34) too? Otherwise LGTM.

vsliouniaev:
> Don't you need to update [pkg/operator/defaults.go](https://github.com/coreos/prometheus-operator/blob/502d2d825923933ee7dfb0b26a746f823cff7855/pkg/operator/defaults.go#L34) too? Otherwise LGTM.

That got picked up when I rebased this pull request

PR #3231, opened 2020-05-16
Update dependencies to prometheus 2.18

vsliouniaev:
The compatibility matrix indicates that prometheus-operator 0.39 should be compatible with prometheus 2.17.2, which has the `absent_over_time` function.

Requires: https://github.com/openshift/prom-label-proxy/pull/28
Relates to #3229


lilic:
Thanks, lets merge for now until we find a new home for the prom-label-proxy so openshift releases don't block it. cc @brancz as discussed.

PR #3232, opened 2020-05-17
docs: add a user-guide for the admission webhook

### LINKED ISSUE #2916: Consider documenting the admission webhook
**What is missing?**
Currently the operator has an awesome admission webhook to validate PrometheusRules and making sure they won't break Prometheus' configuration reload. However, there are very few mentions of that webhook in the project (the most obvious clue being a mention in the changelog, an existing issue and some tests).

**Why do we need it?**
If not specifically looking for it, as an end-user it's quite difficult to figure out this feature exists. It's also not fully trivial to setup (needs TLS, hence a sidecar such as Ghostunnel is also required).

I'd be willing to just write some doc about it, though I am unsure where would be the best fit (maybe a user-guide and a small mention in the README?)

mickael-carl:
Fixes #2916

### DIFF
--- a/Documentation/user-guides/webhook.md
+++ b/Documentation/user-guides/webhook.md
@@ -1,0 +1,74 @@
+# Admission webhooks
+
+This document describes how to set up an admission webhook to validate
+PrometheusRules, and thus preventing Prometheus from loading invalid
+configuration.
+
+## Prerequisites
+
+This guide assumes that you have already [deployed the Prometheus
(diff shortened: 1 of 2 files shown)


mickael-carl:
Potentially might need an update with #3230 since that updates the webhook to use `admission/v1` instead of `admission/v1beta1`?

brancz:
Doesn’t need to block this I think.

brancz:
Thanks a lot! :)

PR #3241, opened 2020-05-25
prometheus: Implement sharding mechanism

### LINKED ISSUE #3130: Horizontal scaling via sharding
**What is missing?**
Horizontal Scaling via Sharding

**Why do we need it?**
Right now we are collecting approx. 2 Million series with 2 hour blocks and using thanos for long term storage and deduping at query time. Despite these small block sizes, the ram usage is around 11 gigs. We run in HA mode so multiple prometheus pods all have about the same ram usage.

As we scale out more metrics, it will become unfeasible to keep scaling vertically (currently they are on 32 gig boxes so we do have room).

**Environment**
Kubernetes 1.15.7
Thanos 0.10.1

* Prometheus Operator version: quay.io/coreos/prometheus-operator:v0.37.0

I believe I know a way to fix implement this simply enough. We can add an option to the operator like so:

```yaml
sharding:
  enabled: true
  numShards: 2
  hashKey: __address__
```

In turn, this would add something like this to **each** of the jobs in the prometheus config:

> On prometheus-replica-0:

```yaml
- job_name: "foo"
  metrics_path: "/metrics"
  kubernetes_sd_config:
  - role: endpoints
  relabel_configs:
      modulus:       2
      target_label:  __tmp_hash
      action:        hashmod
    - source_labels: [__tmp_hash]
      regex:         ^0$
      action:        keep
```

> On prometheus-replica-1:

```yaml
- job_name: "foo"
  metrics_path: "/metrics"
  kubernetes_sd_config:
  - role: endpoints
  relabel_configs:
      modulus:       2
      target_label:  __tmp_hash
      action:        hashmod
    - source_labels: [__tmp_hash]
      regex:         ^1$
      action:        keep
```

HA would still be available by having more replicas than shards. In the above example that shards using modulus 2, `prometheus-replica-2` config would match `prometheus-replica-0` config.

Deduping would be done by thanos and/or alertmanager.

To merge results, federation can be enabled and then the master can scrape them and consolidate **OR** use a third party tool like thanos to do this for you.

I would be happy to implement this feature if this proposal is accepted.

brancz:
This implements the long overdue sharding mechanism that allows for easily sharding a Prometheus cluster.

This is fully backward compatible, but slightly changes the significance of the `replicas` field in the spec. Instead of being blindly copied from the Prometheus spec to the StatefulSet spec, now the replicas in the StatefulSet are calculated by: `# of shards * # of replicas`, essentially meaning that `.spec.replicas` is now "how many replicas of each shard to create".

~This is still fully contained in one StatefulSet. An instance knows it's shard by looking at its ordinal within the statefulset. An example of a 2 replicas and 2 shards configuration would result in the following pods:~

~prometheus-0 -> shard 0~
~prometheus-1 -> shard 0~
~prometheus-2 -> shard 1~
~prometheus-3 -> shard 1~

Edit Nov 2:

This creates a statefulset per shard, the "0" shard defaulting to the naming before sharding was introduced to make everything backward compatible. An example of a 2 replicas and 2 shards configuration would result in the following pods:

* prometheus-0 -> shard 0, replica 0
* prometheus-1 -> shard 0, replica 1
* prometheus-shard-1-0 -> shard 1, replica 0
* prometheus-shard-1-1 -> shard 1, replica 1

This and various other configurations are present in the unit tests.

As noted in the comments, this does not take care of any resharding work, it's only about sharding the scrape work.

Closes #3130 #2590 

@lilic @pgier @simonpasquier @s-urbaniak @paulfantom @metalmatze


brancz:
If everyone is happy with this approach and design I'll happily add design and user docs for this.

vsliouniaev:
Wow! This would probably help us quite a bit!

I'm wondering how a single statefulset will work with node drains and pod disruption budgets - is it possible that a whole shard will go down simultaneously?

brancz:
Pod spread and PDB are great points, I need to think about those a bit more. It wouldn't be difficult to extract things into separate statefulsets. The logic is roughly the same (maybe even easier as it wouldn't be ordinal based). For `shards==1` we would just do exactly what we do today and only for `shards>1` would we create more.

pkg/prometheus/promcfg.go
```diff
@@ -853,6 +871,19 @@ func (cg *configGenerator) generateServiceMonitorConfig(
...
 }
 
+func addressSharding(relabelings []yaml.MapSlice, shards int32) []yaml.MapSlice {
+	return append(relabelings, yaml.MapSlice{
+		{Key: "source_labels", Value: []string{"__address__"}},
+		{Key: "target_label", Value: "__hash"},
+		{Key: "modulus", Value: shards},
+		{Key: "action", Value: "hashmod"},
+	}, yaml.MapSlice{
+		{Key: "source_labels", Value: []string{"__hash"}},
```
simonpasquier, line 881: (nit) `__tmp_hash` instead of `__hash` as per Prometheus documentation on relabeling?

brancz:
I'm going to try and explore ways to do this with multiple statefulsets, as with a single statefulset we can't in a meaningful way perform pod spreading and PodDisruptionBudget. Thanks to @vsliouniaev for pointing this out. Will put this in WIP until I have something that can be reviewed.

temujin9:
@brancz We currently shard with stock Prometheus, and want to move to the operator. Following this ticket, please let me know if I can help with testing anything.

brancz:
Gave this another attempt. Now each statefulset represents a shard and replicas just sets the replicas in each of those shards. There are e2e tests and they all pass, so I think this is ready for first rounds of reviews! :)

lilic:
Can you rebase, thanks!

lilic:
> continuous-integration/travis-ci Expected — Waiting for status to be reported

Seems strange, either travis or github failed, can you push again to retrigger, thanks!

s-urbaniak:
i think this is a legit failure:
```
 --- FAIL: TestAllNS/y/ShardingProvisioning
```

brancz:
Yeah this is definitely a legit failure, I need to just find time to finish this up :)

s-urbaniak:
it's green! :green_heart: :tada:

brancz:
This is ready for a review now. Instead of one statefulset sharding within itself, we now have one statefulset per shard, the "-shard-0` being the current "default", thus it it's skipped. Eg. a Prometheus object called "main" with 1 shard creates the statefulset `prometheus-main`, whereas with 2 shards it creates `prometheus-main` and `prometheus-main-shard-1`.

s-urbaniak:
this is looking great! :tada: just some questions:

1. `__address__` is currently hard-coded. Can you elaborate on why this is the most sane default? Does it make sense to make it user-configurable as suggested in https://github.com/prometheus-operator/prometheus-operator/issues/3130#issue-595406026 if users want to segment according to their own logical shard keys?

2. I believe, we should add some documentation, especially around the rationale for using the static shard key and the logic of statefulset naming as outlined in https://github.com/prometheus-operator/prometheus-operator/pull/3241#issuecomment-673419178

Documentation/api.md
```diff
@@ -493,7 +493,8 @@ PrometheusSpec is a specification of the desired behavior of the Prometheus clus
...
 | imagePullSecrets | An optional list of references to secrets in the same namespace to use for pulling prometheus and alertmanager images from registries see http://kubernetes.io/docs/user-guide/images#specifying-imagepullsecrets-on-a-pod | [][v1.LocalObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#localobjectreference-v1-core) | false |
-| replicas | Number of instances to deploy for a Prometheus deployment. | *int32 | false |
+| replicas | Number of replicas of each shard to deploy for a Prometheus deployment. Number of replicas multiplied by shards is the total number of Pods created. | *int32 | false |
+| shards | Number of shards to distribute targets onto. Number of replicas multiplied by shards is the total number of Pods created. Note that scaling down shards will not reshard data onto remaining instances, it must be manually moved. Increasing shards will not reshard data either but it will continue to be available from the same instances. To query globally use Thanos sidecar and Thanos querier or remote write data to a central location. | *int32 | false |
```
s-urbaniak, line 497: is it worth mentioning that we use `__address__` label as the static shard key?

brancz: yes makes sense.

brancz:
1) Choosing `__address__` statically was simply a matter of not exposing configuration until there is a very good reason to do so. There is no practical reason why any two targets should have the same address (with the very rare exception of multiple metrics endpoints on the same listener, but even that should disappear in the distribution). I'm happy to reconsider this if we have a sound use case, but wanted to start with the least amount of configuration necessary.
2) Agreed, some documentation is deserved here. I would think a user guide would be best, as the API docs already explain the feature, or do you think it would make more sense as part of the design doc?

lilic (commented):
Do you mind rebasing, thanks!

mcavoyk:
@brancz @s-urbaniak Besides rebasing, is there any other work needed to move this PR forward? I'm very interested in trying out this feature :smile:

brancz:
This should be ready for review again. The rebase was a bit messy, so I will also review this myself (but unit and e2e tests are passing so that gives me some degree of confidence :) ).

Documentation/api.md
```diff
@@ -493,7 +493,8 @@ PrometheusSpec is a specification of the desired behavior of the Prometheus clus
 | image | Image if specified has precedence over baseImage, tag and sha combinations. Specifying the version is still necessary to ensure the Prometheus Operator knows what version of Prometheus is being configured. | *string | false |
 | baseImage | Base image to use for a Prometheus deployment. Deprecated: use 'image' instead | string | false |
 | imagePullSecrets | An optional list of references to secrets in the same namespace to use for pulling prometheus and alertmanager images from registries see http://kubernetes.io/docs/user-guide/images#specifying-imagepullsecrets-on-a-pod | [][v1.LocalObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#localobjectreference-v1-core) | false |
-| replicas | Number of instances to deploy for a Prometheus deployment. | *int32 | false |
+| replicas | Number of replicas of each shard to deploy for a Prometheus deployment. Number of replicas multiplied by shards is the total number of Pods created. | *int32 | false |
```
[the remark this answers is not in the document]
brancz: I think we really only need `shards` to be experimental. This was already here and supported, and didn't really change except when `shards` is configured.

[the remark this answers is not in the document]
brancz: Oh, and I added "EXPERIMENTAL" as the first thing in the `shards` description :)

brancz:
I did a round of fixes, I think this should be good for everyone to review :)

pkg/prometheus/operator.go
```diff
@@ -1295,31 +1341,35 @@ func PrometheusStatus(ctx context.Context, kclient kubernetes.Interface, p *moni
...
+
+		for _, pod := range pods.Items {
+			ready, err := k8sutil.PodRunningAndReady(pod)
+			if err != nil {
+				return nil, nil, errors.Wrap(err, "cannot determine pod ready state")
 			}
-			continue
+			if ready {
+				res.AvailableReplicas++
+				// TODO(fabxc): detect other fields of the pod template
```
lilic, line 1362: Do you think this TODO is still applicable?

pkg/prometheus/operator_test.go
```diff
@@ -138,3 +138,67 @@ func TestGetNodeAddresses(t *testing.T) {
...
+		expected string
+	}{
+		{
+			name:     "namespace/test",
+			shard:    0,
+			expected: "namespace/prometheus-test",
+		},
+		{
+			name:     "namespace/test",
+			shard:    1,
```
lilic, line 193: Would a shard with -1 value be also a good thing to add as a test case?

pkg/prometheus/promcfg.go
```diff
@@ -1137,6 +1153,19 @@ func getLimit(user uint64, enforced *uint64) uint64 {
 	return user
 }
 
+func addressSharding(relabelings []yaml.MapSlice, shards int32) []yaml.MapSlice {
```
lilic, line 1156: Nit: Seems like a confusing function name to me, had to look at the function to understand what it did, maybe something more self-descriptive like `injectSharding` or `generateSharding` wdyt?

Would a unit test be good here for this function to cover some possible edge cases, wdyt?

pkg/prometheus/statefulset.go
```diff
@@ -51,21 +51,42 @@ const (
...
 		managedByOperatorLabel: managedByOperatorLabelValue,
 	}
-	probeTimeoutSeconds int32 = 3
+	shardLabelName                = "operator.prometheus.io/shard"
+	prometheusNameLabelName       = "operator.prometheus.io/name"
+	probeTimeoutSeconds     int32 = 3
 )
 
+func expectedStatefulSetShardNames(
+	p *monitoringv1.Prometheus,
```
lilic, line 68: Nit: Do we need this format, it only has one var, we can have it all in one line?

brancz:
I also updated the PR description to reflect the latest implementation state.

pkg/prometheus/statefulset.go
```diff
@@ -871,6 +902,14 @@ func prefixedName(name string) string {
 	return fmt.Sprintf("prometheus-%s", name)
 }
 
+func prometheusName(name string, shard int32) string {
```
lilic, line 905: Nit: Could we add this function closer to the one where it's used, or vice versa, we only seem to use it in expectedStatefulSetShardNames?

pkg/prometheus/operator.go
```diff
@@ -1165,64 +1187,88 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	}
 
 	ssetClient := c.kclient.AppsV1().StatefulSets(p.Namespace)
+
 	// Ensure we have a StatefulSet running Prometheus deployed.
```
lilic, line 1191: Lets adjust this to say:
```suggestion
	// Ensure we have a StatefulSet running Prometheus deployed and that StatefulSet names are created correctly.
```

pkg/prometheus/operator.go
```diff
@@ -1295,31 +1341,35 @@ func PrometheusStatus(ctx context.Context, kclient kubernetes.Interface, p *moni
...
-	if err != nil {
-		return nil, nil, errors.Wrap(err, "retrieving stateful set failed")
-	}
-
-	res.Replicas = int32(len(pods.Items))
 
 	var oldPods []v1.Pod
-	for _, pod := range pods.Items {
-		ready, err := k8sutil.PodRunningAndReady(pod)
+	expected := expectedStatefulSetShardNames(p)
```
lilic, line 1346: Do we need this check again?

test/e2e/prometheus_test.go
```diff
@@ -1938,6 +1938,151 @@ func testPromSharedResourcesReconciliation(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = framework.KubeClient.AppsV1().StatefulSets(ns).Get(context.TODO(), fmt.Sprintf("prometheus-%s", p.Name), metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = wait.Poll(time.Second, 1*time.Minute, func() (bool, error) {
```
lilic, line 2072: We timeout after 1 second?

test/e2e/prometheus_test.go
```diff
@@ -1938,6 +1938,151 @@ func testPromSharedResourcesReconciliation(t *testing.T) {
...
+			ContainerName: "prometheus",
+			CaptureStdout: true,
+			CaptureStderr: true,
+			Stdin:         nil,
+		})
+		if err != nil {
+			t.Fatalf("Failed to read config from pod %q: %v", p.pod, err)
+		}
+		if !strings.Contains(stdout, p.expectedShardConfigSnippet) {
+			t.Fatalf("expected shard config to be present for %v but not found in config:\n\n%s\n\nexpected to find:\n\n%s", p.pod, stdout, p.expectedShardConfigSnippet)
```
lilic, line 2009: Nit:
```suggestion
			t.Fatalf("Expected shard config to be present for %v but not found in config:\n\n%s\n\nexpected to find:\n\n%s", p.pod, stdout, p.expectedShardConfigSnippet)
```

test/e2e/prometheus_test.go
```diff
@@ -1938,6 +1938,151 @@ func testPromSharedResourcesReconciliation(t *testing.T) {
...
+			PodName:       p.pod,
+			ContainerName: "prometheus",
+			CaptureStdout: true,
+			CaptureStderr: true,
+			Stdin:         nil,
+		})
+		if err != nil {
+			t.Fatalf("Failed to read config from pod %q: %v", p.pod, err)
+		}
+		if !strings.Contains(stdout, p.expectedShardConfigSnippet) {
```
lilic, line 2008: Is this e2e test enough to test the sharding feature, my only worry here is we don't test the sharding feature just Prometheus configuration and even that not in full just that the parts of config are there. Is there another way we could do an e2e test?

test/framework/service.go
```diff
@@ -17,11 +17,12 @@ package framework
 import (
 	"context"
 	"fmt"
-	"k8s.io/apimachinery/pkg/util/yaml"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/yaml"
+
```
lilic, line 23: Nit: lets remove the new line, at least I don't see the need for it?

test/framework/pod.go
```diff
@@ -60,3 +68,59 @@ func (f *Framework) GetPodRestartCount(ns, podName string) (map[string]int32, er
...
+	PodName       string
+	ContainerName string
+	Stdin         io.Reader
+	CaptureStdout bool
+	CaptureStderr bool
+	// If false, whitespace in std{err,out} will be removed.
+	PreserveWhitespace bool
+}
+
+// ExecWithOptions executes a command in the specified container,
```
lilic, line 85: Is this borrowed from Kubernetes? :) If yes let's mention that? Anything against importing it instead?

lilic (commented):
Amazing work, mainly nits and questions from my side, nothing blocking! 👏

pkg/prometheus/operator.go
brancz: Not really, will remove.

test/e2e/prometheus_test.go
s-urbaniak: rather after 1 minute no?

pkg/prometheus/operator_test.go
brancz: I don't think so. The code tested here is guarded by code that tests that shards are always >=1.

test/e2e/prometheus_test.go
s-urbaniak: and poll once per second

s-urbaniak:
LGTM modulo @lilic 's review comments, thank you! :tada:

pkg/prometheus/promcfg.go
brancz: Agreed, something with "generate" in the function name would make more sense.

brancz: There are not really any edge cases here though that aren't covered elsewhere already, as this is really only text templating.

brancz: Renaming to `generateAddressShardingRelabelingRules`

pkg/prometheus/statefulset.go
brancz: I tend to prefer this, as parameters tend to be ever-growing. No strong opinion though.

pkg/prometheus/statefulset.go
brancz: Makes sense, will do. I'll also rename it to `prometheusNameByShard`, which makes it a bit clearer what it does.

pkg/prometheus/operator.go
brancz: I wanted status to be entirely detached from the "actionable code", kind of like a black-box or unit test, testing for a specific pre-calculated result.

test/e2e/prometheus_test.go
brancz: The function signature is:

```go
func Poll(interval, timeout time.Duration, condition ConditionFunc) error {
```

So we're checking once every second for 1 minute.

test/e2e/prometheus_test.go
brancz: 1) Sharding is already tested in Prometheus, it feels off to double test it here.
2) We shard on the `__address__` label, which is a `Pod`s IP, which is not something we can deterministically implement in an e2e test, unfortunately.

test/framework/service.go
brancz: Whops, not sure how that got there.

test/framework/pod.go
brancz: importing k/k is a nightmare, I'll give a reference where this is from yes

pkg/prometheus/operator.go
brancz: Done

test/e2e/prometheus_test.go
lilic: For me github said its using:
> func (f *Framework) Poll(timeout, pollInterval time.Duration, pollFunc func() (bool, error)) error

https://github.com/prometheus-operator/prometheus-operator/blob/a6f0c9a10ec0cd91e1d9c1d9e8af30371cf9af82/test/framework/helpers.go#L140

lilic: It seems like github mistook it, we are indeed using Poll from "k8s.io/apimachinery/pkg/util/wait" and not the one from our helpers. 

Side note: Any reason for this? If we don't find our own helper useful and we just want to go with the apimachinery one might be nice to open issue to get rid of the custom one?

brancz: Looks like framework.Poll is actually unused. How about we remove it in a follow up?

lilic (approved):
lgtm

🎉

PR #3250, opened 2020-05-27
Add scrapeTimeout as global configurable parameter

carlosedp:
Add global configurable scrapeTimeout parameter to allow monitoring
targets on clusters consisting of slower hosts like Raspberry Pi and
many ARM boards used for labs.

Many times the 10s default timeout is not enough for targets like kube-controller or kube-scheduler so some targets get a "Context deadline exceeded" error.

Signed-off-by: Carlos de Paula <me@carlosedp.com>


carlosedp:
Works fine on my ARM cluster:

Jsonnet:

image

Config:

image

brancz:
lgtm on green

carlosedp:
Tests seem flaky. Fails seems not related to the change as I looked into the PR test runs and lots fail around `TestAllNS`, maybe it's timing-out. Any idea?

brancz:
hmm I kicked the jobs a couple of times, I think there is actually something not working

pgier:
It seems be consistently failing on `TestAllNS/y/PromGetAuthSecret`.  Maybe try increasing the scrape timeout for this test?

pkg/prometheus/promcfg.go
```diff
@@ -177,6 +177,11 @@ func (cg *configGenerator) generateConfig(
 		scrapeInterval = p.Spec.ScrapeInterval
 	}
 
+	scrapeTimeout := "10s"
```
lilic, line 180: Whats the reason for the 10seconds here, might have missed something?

carlosedp: It's the default timeout. Even changing to 30s the same test fails. I don't think it is related to the change.

carlosedp:
I've reverted the timeout to 10s (the default Prometheus scrape_timeout) and added a 60s timeout for the failing test.

carlosedp:
Now we're good to go. Better and cleaner implementation, also Travis is green!
Thanks for the patience!

PR #3272, opened 2020-06-08
pkg/alertmanager: mergepatch alertmanager containers

s-urbaniak:
Instead of just appending user defined containers to the alertmanager
pods, use k8sutil.MergePatchContainers so that the default containers
can be modified.

This is a backport of https://github.com/coreos/prometheus-operator/pull/3080

/cc @coreos/team-monitoring


brancz:
(fixed the PR title but the commit title still needs fixing, then lgtm)

PR #3276, opened 2020-06-11
pkg: Add EnforcedSampleLimit which enforces a global sample limit

lilic:
for each ServiceMonitor or/and PodMonitor. EnforcedSampleLimit is taken
in favour of any SampleLimit set per ServiceMonitor or/and PodMonitor
resource. This is meant to be used by admins to be able to enforce a
limit of samples/series for each target.

cc @coreos/team-monitoring


lilic:
As discussed we should also have a way for admins can enforce certain Service/Pod Monitors to be excluded from this limit, but this can be done in a follow up.

pkg/prometheus/promcfg_test.go
```diff
@@ -2704,3 +2704,35 @@ func TestHonorTimestamps(t *testing.T) {
...
+	}{
+		{
+			Enforced: 100,
+			User:     1000,
+			Expected: 100,
+		},
+		{
+			Enforced: 0,
+			User:     888,
+			Expected: 0,
```
brancz, line 2722: That's not what I would have expected. `0` enforced means nothing is enforced, no? In that case if a user chooses a stricter limit, I would expect that limit to be used.

pkg/prometheus/promcfg_test.go
```diff
@@ -2704,3 +2704,35 @@ func TestHonorTimestamps(t *testing.T) {
...
+		},
+		{
+			Enforced: 1,
+			User:     0,
+			Expected: 1,
```
brancz, line 2727: nice edge case!

pkg/prometheus/promcfg_test.go
```diff
@@ -2704,3 +2704,35 @@ func TestHonorTimestamps(t *testing.T) {
 		}
 	}
 }
+
+func TestGetSampleLimit(t *testing.T) {
+	tcs := []struct {
+		Enforced uint64
+		Expected uint64
+		User     uint64
+	}{
```
brancz, line 2713: I would expect one more case where the user specifies a lower limit than the enforced one

lilic: Will do but its sort of already covered by:
```
 			Enforced: 0,
			User:     888,
			Expected: 888,
```

lilic:
@brancz PTAL, thanks!

pkg/prometheus/promcfg_test.go
lilic: Yes! Good catch!

brancz:
Nice! lgtm

PR #3281, opened 2020-06-13
Support Prometheus vertical compaction

yeya24:

Add an `allowOverlappingBlocks` to enable vertical compaction


yeya24:
CI failed but seems not related to this pr...

pkg/prometheus/statefulset.go
```diff
@@ -431,6 +431,12 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		}
 	}
 
+	if p.Spec.AllowOverlappingBlocks {
+		promArgs = append(promArgs, "-storage.tsdb.allow-overlapping-blocks")
+	} else {
```
pgier, line 436: To maintain compatibility with old Prometheus versions (< 2.8), I'd prefer that we leave this field unset when the spec field is set to false.

yeya24: Good point!

yeya24:
Updated.

pkg/apis/monitoring/v1/types.go
```diff
@@ -326,6 +326,9 @@ type PrometheusSpec struct {
 	// the desired limit.
 	// Note that if SampleLimit is lower that value will be taken instead.
 	EnforcedSampleLimit *uint64 `json:"enforcedSampleLimit,omitempty"`
+	// Allow overlapping blocks in Prometheus, which in turn enables vertical compaction
```
lilic, line 329: Nit: this should start with `AllowOverlappingBlocks ...` same as others above.

yeya24:
> Thanks for the PR!
> 
> https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L213 This is an EXPERIMENTAL flag so we should mark it as such here as well. cc @brancz how were EXPERIMENTAL flags done in the past?

I am not sure but if it is necessary I can add some comments to mark this as experimental. But https://github.com/coreos/prometheus-operator/blob/master/pkg/apis/monitoring/v1/types.go#L140 is an experimental flag in Prometheus as well, we don't have any comments for it.

pkg/apis/monitoring/v1/types.go
yeya24: Done.

lilic:
> is an experimental flag in Prometheus as well, we don't have any comments for it.

Agreed we might have did it this way in the past, but doesn't mean we can't change it to be experimental.

@coreos/team-monitoring any thoughts on for or against having this flag be marked in our API as experimental.

yeya24:
> > is an experimental flag in Prometheus as well, we don't have any comments for it.
> 
> Agreed we might have did it this way in the past, but doesn't mean we can't change it to be experimental.
> 
> @coreos/team-monitoring any thoughts on for or against having this flag be marked in our API as experimental.

I will mark this as experimental. I agree vertical compaction is not as widely used as wal compression so I am not sure it is stable.

pkg/prometheus/statefulset.go
```diff
@@ -431,6 +431,10 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		}
 	}
 
+	if p.Spec.AllowOverlappingBlocks {
```
brancz, line 434: this should be gated with the version of prometheus that this was introduced in

brancz:
Agreed that it's good to mark this experimental.

PR #3294, opened 2020-06-24
refactor: decouple pod labels from selector labels

### LINKED ISSUE #3120: Hot-Deploy in Prometheus operator without downtime
Hello,

I have a question,

is there a way to hot-deploy the prometheus/alertmanager resources in Kubernetes without downtime?
I mean if I would like to change/add/delete some field of the Prometheus kind resource for example the image, version, retention, any other or some configuration of the additional-scrape-config, and when that change is performed then prometheus is updated (manually or automaticaly) without downtime for example some kind of rolling-update feature of kubernetes or how deployments in kubernetes works.

Thanks.

kissmikijr:
prometheus pods can not be rolled out without downtime when label's are changed

Fixes #3120

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -657,6 +657,10 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 
 	podAnnotations := map[string]string{}
 	podLabels := map[string]string{}
+	podSelectorLabels := map[string]string{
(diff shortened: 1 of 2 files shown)


kissmikijr:
@brancz Hey here is my try to fix the issue that we talked about, if you could check this out and review this I'd appreciate it if I've solved the problem!
thanks

brancz:
Wow, thank you so much for following up so quickly! This lgtm once CI goes green! :)

PR #3299, opened 2020-06-26
Ignore secrets of type `helm.sh/release.v1`

### LINKED ISSUE #3298: Prometheus Operator watching every secret leads to high memory usage
**What happened?**

After updating to Helm V3, we saw an increase in the Prometheus-Operator memory usage (enough to lead to it being OOMKilled) from Base64 decode functions and json parsing.

After investigating, we found the root cause of the problem: the operator is watching every secrets and Helm v3 created several thousands of them (and they are in the namespace of the resources instead of a namespace we can exclude such as kube-system).
The operator is therefore wasting time scanning thousands of secrets it won't be able to use.

Prometheus-operator should either watch for secret he might be interested in or filter out the secrets he knows he can't use.
Maybe introduce an option such as `SecretWatchListSelector` and `ConfigmapWatchListSelector` to allow the user to configure it.

Create several thousands secrets and observe the memory usage behavior of the operator. 

**Environment**

GKE - Kube 1.16.8

* Prometheus Operator version: quay.io/coreos/prometheus-operator:v0.38.1 (I don't think the version is relevant)

* Kubernetes version information: v1.16.8-gke.15

* Kubernetes cluster kind: GKE

* Manifests: N/A

* Prometheus Operator Logs: No suspicious logs

Impact of filtering out Helm V3 secrets:
```diff
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -330,7 +330,7 @@ func New(conf Config, logger log.Logger, r prometheus.Registerer) (*Operator, er
        c.secrInf = cache.NewSharedIndexInformer(
                c.metrics.NewInstrumentedListerWatcher(
                        listwatch.MultiNamespaceListerWatcher(c.logger, c.config.Namespaces.PrometheusAllowList, c.config.Namespaces.DenyList, func(namespace string) cache.ListerWatcher {
-                               return cache.NewListWatchFromClient(c.kclient.CoreV1().RESTClient(), "secrets", namespace, fields.Everything())
+                               return cache.NewListWatchFromClient(c.kclient.CoreV1().RESTClient(), "secrets", namespace, fields.OneTermNotEqualSelector("type", "helm.sh/release.v1"))
                        }),
                ),
                &v1.Secret{}, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
```

Relevant go_memstat metrics after the patch:
image

obiesmans:
Fixes #3298

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -330,7 +330,7 @@ func New(conf Config, logger log.Logger, r prometheus.Registerer) (*Operator, er
 	c.secrInf = cache.NewSharedIndexInformer(
 		c.metrics.NewInstrumentedListerWatcher(
 			listwatch.MultiNamespaceListerWatcher(c.logger, c.config.Namespaces.PrometheusAllowList, c.config.Namespaces.DenyList, func(namespace string) cache.ListerWatcher {
-				return cache.NewListWatchFromClient(c.kclient.CoreV1().RESTClient(), "secrets", namespace, fields.Everything())
+				return cache.NewListWatchFromClient(c.kclient.CoreV1().RESTClient(), "secrets", namespace, fields.OneTermNotEqualSelector("type", "helm.sh/release.v1"))
 			}),
 		),
 		&v1.Secret{}, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},


brancz:
Maybe having a configurable selector would be more adaptable?

lilic:
@brancz discussion is here https://github.com/coreos/prometheus-operator/issues/3298 there are two suggestions, do you mind commenting there? 

@obiesmans  would suggest we close this PR and once we resolve the issue you can open a new PR as per the agreed suggestion, does that sound good? Thanks!

obiesmans:
@lilic : Thanks for the suggestion. Closed in favor of #3355.

PR #3304, opened 2020-06-29
pkg/prometheus: Allow enforcing namespace label in Probe configs

brancz:
Quick follow up as described in https://github.com/coreos/prometheus-operator/pull/2832#pullrequestreview-428695780.

@lilic @pgier @simonpasquier @auhlig

### DIFF
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -451,6 +451,106 @@ scrape_configs:
     - promcon.io
     labels:
       static: label
+  relabel_configs: []
+alerting:
+  alert_relabel_configs:
+  - action: labeldrop
+    regex: prometheus_replica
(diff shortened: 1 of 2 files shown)


pkg/prometheus/promcfg_test.go
```diff
@@ -473,6 +477,7 @@ func TestProbeIngressSDConfigGeneration(t *testing.T) {
 				Namespace: "default",
 			},
 			Spec: monitoringv1.PrometheusSpec{
+				EnforcedNamespaceLabel: "namespace",
```
lilic, line 480: Could we rather leave this one not set, to make sure its not enforced by default when not set? :) Or do we have another test like this?

brancz: Yes, let me add another test

PR #3306, opened 2020-06-29
fix probe promcfg

yeya24:

Fixes prometheus config for Probe.

1. If only ProbeSelector is set, PodMonitorSelector and ServiceMonitorSelector not set, it will do nothing.
2. Relabelings are not added for Probe.

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -373,7 +373,7 @@ func (cg *configGenerator) generateConfig(
 	return yaml.Marshal(cfg)
 }
 
-// honorLabels determinates the value of honor_labels.
+// honorLabels determines the value of honor_labels.
 // if overrideHonorLabels is true and user tries to set the
 // value to true, we want to set honor_labels to false.
 func honorLabels(userHonorLabels, overrideHonorLabels bool) bool {
@@ -662,6 +662,8 @@ func (cg *configGenerator) generateProbeConfig(
 		{Key: "module", Value: []string{m.Spec.Module}},
 	}})
 
+	var relabelings []yaml.MapSlice
+
 	// Generate static_config section.
 	if m.Spec.Targets.StaticConfig != nil {
 		staticConfig := yaml.MapSlice{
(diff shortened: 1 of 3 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -747,22 +762,6 @@ func (cg *configGenerator) generateProbeConfig(
 			cfg = append(cfg, cg.generateK8SSDConfig(selectedNamespaces, apiserverConfig, basicAuthSecrets, kubernetesSDRoleIngress))
 		}
 
-		// Relabelings for prober.
```
brancz, line 750: why would this not be needed for ingress targets?

yeya24: Thanks for the review, I will add it back.

yeya24:
I am not sure about the current code, why static configs don't have relabeling configs? Is it a bug or expected?

This is one expected result in the unit test.

``` yaml
global:
  evaluation_interval: 30s
  scrape_interval: 30s
  external_labels:
    prometheus: default/test
rule_files: []
scrape_configs:
- job_name: default/testprobe1
  honor_timestamps: true
  metrics_path: /probe
  scheme: http
  params:
    module:
    - http_2xx
  static_configs:
  - targets:
    - prometheus.io
    - promcon.io
    labels:
      static: label
  relabel_configs: []
alerting:
  alert_relabel_configs:
  - action: labeldrop
    regex: prometheus_replica
  alertmanagers: []
```

brancz:
Static targets shouldn't have relabel configs, it doesn't make sense, they should just have labels, which can already be configured.

brancz:
Thanks! :)

yeya24:
> Static targets shouldn't have relabel configs, it doesn't make sense, they should just have labels, which can already be configured.

Sorry, I just saw your reply. Do you mean the relabel configs other than those for Blackbox exporter? Or including the Blackbox exporter relabel configs?

I add these relabel configs to static config as well. If this is wrong, I can open another pr to fix it.

```
		relabelings = append(relabelings, []yaml.MapSlice{
			{
				{Key: "source_labels", Value: []string{"__address__"}},
				{Key: "target_label", Value: "__param_target"},
			},
			{
				{Key: "source_labels", Value: []string{"__param_target"}},
				{Key: "target_label", Value: "instance"},
			},
			{
				{Key: "target_label", Value: "__address__"},
				{Key: "replacement", Value: m.Spec.ProberSpec.URL},
			},
		}...)
```

brancz:
This looks good now. But to make 100% sure, would you mind adding a full e2e test for static and ingress discovered targets? Just so we can be totally sure it works end to end.

yeya24:
> This looks good now. But to make 100% sure, would you mind adding a full e2e test for static and ingress discovered targets? Just so we can be totally sure it works end to end.

Yes, I can add E2E tests in next prs.

brancz:
Awesome! :tada:

PR #3307, opened 2020-06-30
set scrapeInterval and evaluationInterval to 1m

wiardvanrij:
Both scrapeInterval and evaluationInterval are on 30s. This commit changes this to 1m and updates the tests included with it.

Reason: 

- These defaults are not documented for the Prometheus Operator
- These defaults differ from official Prometheus documentation: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration


wiardvanrij:
I have to fix the unit-tests as they are failing at the moment.

PR #3308, opened 2020-06-30
Normalize default durations

wiardvanrij:
Both scrapeInterval and evaluationInterval are on 30s. This commit changes this to 1m and updates the tests included with it.

Reason:

- These defaults are not documented for the Prometheus Operator
- These defaults differ from official Prometheus documentation. See: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -174,12 +174,12 @@ func (cg *configGenerator) generateConfig(
 
 	cfg := yaml.MapSlice{}
 
-	scrapeInterval := "30s"
+	scrapeInterval := "1m"
 	if p.Spec.ScrapeInterval != "" {
 		scrapeInterval = p.Spec.ScrapeInterval
 	}
 
-	evaluationInterval := "30s"
+	evaluationInterval := "1m"
 	if p.Spec.EvaluationInterval != "" {
 		evaluationInterval = p.Spec.EvaluationInterval
 	}
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -67,8 +67,8 @@ func TestGlobalSettings(t *testing.T) {
 		{
 			Version: "v2.15.2",
 			Expected: `global:
-  evaluation_interval: 30s
-  scrape_interval: 30s
+  evaluation_interval: 1m
+  scrape_interval: 1m
   external_labels:
     prometheus: /
     prometheus_replica: $(POD_NAME)
@@ -83,10 +83,10 @@ alerting:
 		},
 		{
 			Version:            "v2.15.2",
-			EvaluationInterval: "60s",
+			EvaluationInterval: "1m",
 			Expected: `global:
-  evaluation_interval: 60s
-  scrape_interval: 30s
+  evaluation_interval: 1m
+  scrape_interval: 1m
   external_labels:
     prometheus: /
     prometheus_replica: $(POD_NAME)
@@ -121,8 +121,8 @@ alerting:
 			Version:       "v2.15.2",
 			ScrapeTimeout: "30s",
 			Expected: `global:
-  evaluation_interval: 30s
-  scrape_interval: 30s
+  evaluation_interval: 1m
+  scrape_interval: 1m
   external_labels:
     prometheus: /
     prometheus_replica: $(POD_NAME)
@@ -143,8 +143,8 @@ alerting:
 				"key2": "value2",
 			},
 			Expected: `global:
-  evaluation_interval: 30s
-  scrape_interval: 30s
+  evaluation_interval: 1m
+  scrape_interval: 1m
   external_labels:
     key1: value1
     key2: value2
@@ -163,8 +163,8 @@ alerting:
 			Version:      "v2.16.0",
 			QueryLogFile: "test.log",
 			Expected: `global:
-  evaluation_interval: 30s
-  scrape_interval: 30s
+  evaluation_interval: 1m
+  scrape_interval: 1m
   external_labels:
     prometheus: /
     prometheus_replica: $(POD_NAME)
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -101,10 +101,10 @@ alerting:
 		},
 		{
 			Version:        "v2.15.2",
-			ScrapeInterval: "60s",
+			ScrapeInterval: "1m",
 			Expected: `global:
-  evaluation_interval: 30s
-  scrape_interval: 60s
+  evaluation_interval: 1m
+  scrape_interval: 1m
   external_labels:
     prometheus: /
     prometheus_replica: $(POD_NAME)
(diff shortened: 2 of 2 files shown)


brancz:
Thank you!

Dominik-K:
@wiardvanrij @brancz This breaks the API which has these defaults implicitly set. No matter if  documented or not, IMO. 

Breaking e.g. for alert-rules which currently uses a [range-duration](https://prometheus.io/docs/prometheus/latest/querying/basics/#range-vector-selectors) `< 2m` (like in `http_requests_total[1m]`). They would simply NOT fire anymore (for `duration = 1m`) or fire quasi-randomly less (for `1m < duration < 2m`).

Why not instead documenting the current released defaults instead? This commit isn't part of a release, yet. So, a revert would be quite easy.

wiardvanrij:
@Dominik-K 

If the `scrape_interval` is ignored by default, which is already quite odd to do. Then I _assume_ people _assume_ the 1m default interval (https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration) is used. 

This commit removes some assumptions on `scrape_interval` by setting it to the de facto standard of Prometheus. 

Setting/using alert rules, I don't see this as breaking as it's per your own configuration. What I'm trying to say is that I would rather force people to check their scrape_interval's and think about it, rather than keeping it explicit and somewhat hidden.

I would agree that in a release, this would be noted as possibly breaking. In such extend that people actually are aware of this setting (and its default). Yet I also want to quote the first line on the readme: 

> Project status: beta Not all planned features are completed. The API, spec, status and other user facing objects may change, but in a backward compatible way.

So, eventually that was my reason to make this PR.

P.s. I'm in no way maintainer or anything, this is just merely my 2 cents and I'm happy with any solution ;)

Dominik-K:
@wiardvanrij 

Thanks for your quick and detailed answer.

> If the `scrape_interval` is ignored by default, which is already quite odd to do. 

Who ignores the `scrape_interval` by default? That may be my missing point. 

Actually, I stumbled upon this PR in figuring out what is the default `scrapeInterval` (in the `PrometheusSpec`) because I've seen I literally overlooked to set it at one project. As I checked it in the live-config (`global: scrape_interval: 30s`) and compared it with the current `master` in here, I got confused and got here.

> Then I _assume_ people _assume_ the 1m default interval (https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration) is used.

If it's somewhere ignored, I agree. But then, in the end, Prometheus uses defaults. Just re-checked it's `1m` for both ([besides a short exception in `v0.14.0`](https://github.com/prometheus/prometheus/pull/763)), that's right.


> Setting/using alert rules, I don't see this as breaking as it's per your own configuration.

Easier setting/maintaining "own configuration" is one main purpose of this project, IMO. So the API shouldn't break sth. like this or it should at least be backward-compatible, as you linked to the Readme, too 👍 


> What I'm trying to say is that I would rather force people to check their scrape_interval's and think about it, rather than keeping it explicit and somewhat hidden.

Then the better way would be to make both settings mandatory in the `PrometheusSpec`. That breaking API-change would be backward-compatible. In a sense that it wouldn't (silently) change implicit values, with worse effects (see alert-rules example). Just for completeness: If you _explicitly_ set `scrapeInterval`, you're not affected by both possible changes (diff. value or mandatory) anyway.


> I would agree that in a release, this would be noted as possibly breaking. In such extend that people actually are aware of this setting (and its default). 

Yes. But documenting "Default: `1m` (since version `0.41.0`; before: `30s`)" makes me wonder if the change really makes sense to justify a note like this.

> P.s. I'm in no way maintainer or anything, this is just merely my 2 cents and I'm happy with any solution ;)

Me too. This are my 2 cents, too. To avoid negative surprises or annoyances.

brancz:
My apologies, we should have indeed thought this through a bit more. How do you suggest we proceed @Dominik-K? Do you think reverting this is best and doing a patch release? Do you think making it more explicit in the changelog/release notes would be better?

Dominik-K:
@brancz No problem. I'm happy you're answering quickly 👍 

The commit https://github.com/coreos/prometheus-operator/commit/a7e60b2600afa222c10f29f81b098b0daafbf485 (from July, 1st) isn't released yet ([latest release `v0.40.0` from June, 18th](https://github.com/coreos/prometheus-operator/releases/tag/v0.40.0)).

So I suggest to revert this PR-commit and create an extra PR which either 
* documents the default-values for the intervals 
* OR make these fields obligatory.
The decision is up to you. I don't know your overall design on this enough.

@wiardvanrij Are you ok with this?

wiardvanrij:
Yes, I'm fine with a revert to come up with a solid solution that helps everyone. 

For the exact solution I'm uncertain. It feels documenting a default value, which is not the same as Prometheus itself makes it overcomplicated. The Prometheus documentation is really solid. I would really suggest honouring those specs.

brancz:
There are a couple of odd behaviors that we have collected over time, it's just how software grows. We were thinking of one day doing a release that would be "a bit more breaking", that would do changes like this. I would suggest for now we revert this, and put this on a "breaking release" issue. What do you think?

Dominik-K:
TL;DR: Revert and make the settings mandatory instead of changing defaults. @brancz 

@wiardvanrij 
> It feels documenting a default value, which is not the same as Prometheus itself makes it overcomplicated. The Prometheus documentation is really solid. I would really suggest honouring those specs.

At the first moment, I was uncertain, too. Because in general I fully support this argument.

In the search for a decisive argument pro/contra changing the default values: Who are the affected users we're talking about? Meaning who didn't explicitly set `scrapeInterval` or `evaluationInterval`?
I think of these:

1. Users who took a getting-started approach and stuck with it as it's working out-of-the-box. Both [`prometheus-operator`'s getting-started](https://github.com/coreos/prometheus-operator/pull/2870/files?file-filters%5B%5D=.yaml) as well as [`kube-prometheus`'s default manifest](https://github.com/coreos/kube-prometheus/blame/master/manifests/prometheus-prometheus.yaml).
2. Users who started from scratch `Prometheus`-config and considering the current default (`30s`) as ok and didn't explicitly set it. Perhaps because they put trust in the pledge of the Readme (you mentioned). (Yes, they should best know the best-practice to pinpoint default-values, as well :-). But the reality ...)

The others should have set the values explicitly, i.e. they're not affected. (If I missed a group, then it would be even more severe.)

So, I guess that many users of Prometheus-Operator fall into the two groups (we have 1 project with `kube-prometheus` in the first group). 

Another aspect would be to look at how users upgrade a project like this. Knowing that Prometheus has a very well and long track of not breaking things (esp. database; my impression), you may shed the light on the Prometheus-Operator as well :-) 

So, you may just increase the version number and see from error-messages if something is wrong. Without looking at release-notes or whatsoever before, for sure. 
But in this case (changing default values), there won't be any error message. Then you will/must find it out the hard way. The worst, with a long(er) MTTD because of non-working alerts.

Long story, short: I vote for **don't change the default values**.

So, either you could leave the values like before. Or you **make them mandatory**. This would bring both groups to think about these intervals again (or for the first time at all). And it would lead to better practice.

I could create PR for it after we've settled decision.

wiardvanrij:
I think you make a lot of sense so I agree :) Cheers.

As for a possible change, rather than making the setting mandatory, perhaps we could just add extra information. Information on start-up in case `scrapeInterval` is not defined. I.e. "Using default scrapeInterval: {value}, it's advised to set this yourself". Then also be more explicit in the docs / api spec regarding this value. 

I'm also fine with just only a doc/api 'info update'.

brancz:
What you're saying makes sense, but we cannot break things at API level, and making a field mandatory would be exactly that. In any case it seems that we agree on reverting this :)

PR #3310, opened 2020-07-02
Add E2E test for probe static targets

yeya24:

This pr adds a basic E2E test for Probe with static targets.

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/url"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -218,7 +218,7 @@ test/instrumented-sample-app/certs/cert.pem test/instrumented-sample-app/certs/k
 .PHONY: test-e2e
 test-e2e: KUBECONFIG?=$(HOME)/.kube/config
 test-e2e: test/instrumented-sample-app/certs/cert.pem test/instrumented-sample-app/certs/key.pem
-	go test -timeout 55m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG) --operator-image=$(REPO):$(TAG) -count=1
+	go test -timeout 55m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG) --operator-image=$(REPO):master -count=1
```
yeya24, line 221: I use master tag here since the `Probe` feature is not released yet, so let's just use this to test in CI first. I can change the tag when this is included in the next release.

test/e2e/prometheus_test.go
```diff
@@ -2284,6 +2284,60 @@ func testPromTLSConfigViaSecret(t *testing.T) {
...
+	}
+
+	if finalizerFn, err := testFramework.CreateServiceAndWaitUntilReady(framework.KubeClient, ns, svc); err != nil {
+		t.Fatal(errors.Wrap(err, "creating prometheus service failed"))
+	} else {
+		ctx.AddFinalizerFn(finalizerFn)
+	}
+
+	// Wait for Probe target
+	if err := framework.WaitForTargets(ns, svc.Name, 1); err != nil {
```
brancz, line 2336: I think for this test it would be useful to inspect this target a bit further and see that the configuration of the target is as expected.

Makefile
brancz: This shouldn't need to be necessary, if you follow these [instructions](https://github.com/coreos/prometheus-operator#running-end-to-end-tests-on-local-kind-cluster) the tag based approach should work just fine.

brancz:
Amazing work! Thank you so much!

yeya24:
@brancz Updated. Please take a look when you have time.

test/e2e/prometheus_test.go
yeya24: Yeah, you are right. Thanks!

yeya24:
Maybe this is still wrong.

This is what the static probe looks like. I guess the value of `instance` should be the Blackbox exporter address?
Screenshot from 2020-07-02 22-33-34

PR #3311, opened 2020-07-02
misc: remove v1beta1 crd remainings

dgrisonnet:
Some v1beta1 CRD files were inadvertently added alongside the probe. As we are only supporting v1 CRDs, this PR intends to remove these legacy files.


pgier:
LGTM

PR #3312, opened 2020-07-03
Fix probe ingress configs

yeya24:

Tried probe with ingress locally and found this problem. The point is that we should put relabeling configs for Blackbox exporter after relabeling configs for Ingress SD, otherwise `__address__` will be changed to proberURL, so it would go wrong here https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/promcfg.go#L788.

Verification.

Screenshot from 2020-07-02 23-13-20

probe

``` yaml
kind: Probe
metadata:
  name: ingress
  namespace: default
  labels:
    probe: test
spec:
  jobName: test
  prober:
    url: blackbox-prometheus-blackbox-exporter:9115
  targets:
    ingress:
      selector:
        matchLabels:
          probe: test
      namespaceSelector:
        matchNames:
        - default
  module: http_2xx
  interval: 15s
```

Ingress

``` yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test
  namespace: default
  labels:
    probe: test
spec:
  rules:
  - host: example.com
    http:
      paths:
      - path: /
        backend:
          serviceName: prometheus-self
          servicePort: web
```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -702,9 +702,7 @@ func (cg *configGenerator) generateProbeConfig(
 
 	// Generate kubernetes_sd_config section for ingress resources.
 	if m.Spec.Targets.StaticConfig == nil {
-		var (
-			labelKeys []string
-		)
+		labelKeys := make([]string, 0, len(m.Spec.Targets.Ingress.Selector.MatchLabels))
 
 		// Filter targets by ingresses selected by the monitor.
 		// Exact label matches.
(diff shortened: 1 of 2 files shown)


yeya24:
Though I know an E2E test for probe ingress is necessary as well. But it is not a small work, including creating ingress controller, setup RBAC, etc. So I don't want to add it in this PR.

brancz:
This needs fixing the test fixtures, otherwise this lgtm.

Thank you so much for trying all of this out so thoroughly and fixing these things. Your work is invaluable! :tada:

yeya24:
CI is green now.

brancz:
I don't know why Travis CI is not updating the Github status, but you are right. Merging! :tada:

PR #3315, opened 2020-07-03
pkg/prometheus: Ensure relabeling of container label in ServiceMonitors

brancz:
PodMonitors already default to relabeling namespace, pod and container
into the target labels. ServiceMonitors should do the same to allow easy
correlation between signals.

@coreos/team-monitoring


metalmatze:
Oh wow. Would have guessed that that's the default. :grin:

brancz:
I thought the same, which is what triggered this patch :)

PR #3318, opened 2020-07-03
Support job name in probe

yeya24:

Support jobName, honor_labels and honor_timestamps config in probe

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -632,10 +632,14 @@ func (cg *configGenerator) generateProbeConfig(
 	ignoreNamespaceSelectors bool,
 	enforcedNamespaceLabel string) yaml.MapSlice {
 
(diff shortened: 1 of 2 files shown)


brancz:
Do you have a concrete use for honoring labels and timestamps? It was discussed on the original PR, and then decided that it doesn't really make sense for probes, but if there is a legitimate use case I'd be more than happy to add it. Just don't want to add it for the sake of having it.

yeya24:
> Do you have a concrete use for honoring labels and timestamps? It was discussed on the original PR, and then decided that it doesn't really make sense for probes, but if there is a legitimate use case I'd be more than happy to add it. Just don't want to add it for the sake of having it.

Thanks, I check the original PR and I agree honor_labels and honor_timestamps are not necessary in Blackbox exporter case, you're right!

I update the pr to only include the job name update. Please take a look and see if this one is nice to have, thanks!

PR #3319, opened 2020-07-06
reload alertmanager on configmap/secret change

### LINKED ISSUE #3309: Alertmanager not reloaded on configMaps changes
**What happened?**

We have a configmap that stores our templates (e.g. for E-Mails) for the Alertmanager.
We specify an Alertmanager resource with the `configMaps` value set.
On startup the Alertmanager correctly uses the configured templates.

However, when we update the configmap (and only the configmap) the changes do not seem to be propagated to Alertmanager. While kubernetes automatically updates the mounted files based on the new version of the configmap; the Alertmanager does not pick up on this.

It seems that the Alertmanager is not reloaded on updates to configmaps listed in the `configMaps` field.

We would've expected Alertmanager to be automatically reloaded in order to take the changes in configmaps into account as this is what happens when updating the `configSecret` (the actual alertmanager configuration file).

**Environment**

    [`v0.39.0`](https://github.com/coreos/prometheus-operator/releases/tag/v0.39.0)
    

    `kubernetes v1.16.10`
    

* Kubernetes cluster kind:

    custom

A possible solution for this would be to configure [`configmap-reload`](https://github.com/jimmidyson/configmap-reload/tree/v0.3.0) differently. Currently it only mounts the alertmanager secret and listens to changes in those files. If we mount the entire `/etc/alertmanager` structure instead (including additional secrets (placed in `/etc/alertmanager/secrets`) and additional configmaps (placed in `/etc/alertmanager/configmaps`) and configure `configmap-reload` to use `/etc/alertmanager` as `-volume-dir` reloads should be triggered as expected.

0robustus1:
Via the configMaps and secrets fields we can add additional
files that should be mounted into the /etc/alertmanager directory.
These can then be used for templates (e.g. e-mail templates).

If those change we want to reload the alertmanager in order
to urge it to take the changed files into account.

closes #3309

----

I was thinking about how to potentially refine the `testAMReloadConfig` e2e test, but since the reload would not change the config that is exposed in via the API endpoint we could not check against anything. Ultimately the only component we could test against would be a metric, e.g. the Alertmanager config reload timestamp metric (`alertmanager_config_last_reload_success_timestamp_seconds`). I was not really certain whether this is something you would like to have in the test suite. Let me know whether i should aim to add a test for that.

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -435,6 +435,15 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 		},
 	}
 
+	reloadWatchDirs := []string{alertmanagerConfDir}
+	configReloaderVolumeMounts := []v1.VolumeMount{
+		{
+			Name:      "config-volume",
+			MountPath: alertmanagerConfDir,
+			ReadOnly:  true,
+		},
+	}
+
 	for _, s := range a.Spec.Secrets {
 		volumes = append(volumes, v1.Volume{
 			Name: k8sutil.SanitizeVolumeName("secret-" + s),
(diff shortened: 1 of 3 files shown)


brancz:
Could you add a test for this? Otherwise this lgtm! :)

0robustus1:
I added a couple of test cases to the `TestAMReloadConfig` test (one for configmaps & one for secrets).

brancz:
I would have been happy with a unit test that just checks the templating as well, but this is even better! Thank you! :tada:

PR #3323, opened 2020-07-07
pkg/namespace-labeler: fix whitespace

talal:
There should be an empty line between license header and package
declaration otherwise the license gets interpreted as package
documentation by godoc.


brancz:
Thank you!

PR #3328, opened 2020-07-09
Remove deprecated handling of nil {service,pod}Monitor

alvaroaleman:
the "deprecation warning" has been part of 2 releases now though, so I think it would be ok to finally remove this behavior entirely including this "warning"


pkg/prometheus/operator.go
```diff
@@ -1929,6 +1922,9 @@ func (c *Operator) selectPodMonitors(p *monitoringv1.Prometheus) (map[string]*mo
 	// Selectors might overlap. Deduplicate them along the keyFunc.
 	res := make(map[string]*monitoringv1.PodMonitor)
 
+	if p.Spec.PodMonitorSelector == nil {
```
alvaroaleman, line 1925: It would be nice to instead change the types so this becomes a literal instead of a pointer but that might break consumers. Do we care?

alvaroaleman:
hmm, locally the failing test succeeds (with kind):
```
$ go test ./test/e2e/ --kubeconfig=$KUBECONFIG -run TestAllNS/y/PromReloadConfig
ok  	github.com/coreos/prometheus-operator/test/e2e	133.600s
```

any idea?

pkg/prometheus/operator.go
brancz: Could you verify what would happen with an object currently has nil and we change the type to use the literal? If it defaults to the empty literal then I'm happy with changing this. As in if forward compatibility is guaranteed I'm totally ok with this. If we end up doing this, can we make sure to do the same with `probeSelector`?

brancz:
Re-kicked the job. It might be possible that that test used the deprecated behavior actually, so we would need to use something else to verify the test.

pkg/prometheus/operator.go
alvaroaleman: You mean if the json serialization doesn't have the field? It will default to the empty literal because it can not be nil anymore. Once we update the Prometheus, we will insert a `fieldName: {}` into it.

I was mostly thinking about possible downstream consumers of the api if there are any.

tux-o-matic:
Those Deprecation warnings in the doc are confusing, wouldn't mind seeing the doc updated.
And while on the topic, are `podMonitorSelector` and `probeSelector` really *experimental* features?

brancz:
PodMonitors are probably safe to remove the experimental, probes are only a recent feature so they still deserve the experimental wording.

Issue #3218, opened 2020-05-13
Errors in log when watching multiple namespaces 


pgier:
While watching more than one namespace for any resource (Prometheuses, PrometheusRules, Alertmanagers, etc.)  The custom multiListerWatcher Watch() function will occasionally run into an error where the `resourceVersion` string does not match the expected format.  This produces a log error and invokes some internal error handling code.
```
E0304 20:47:27.926789       1 reflector.go:280] github.com/coreos/prometheus-operator/pkg/prometheus/operator.go:467: Failed to watch *v1.PrometheusRule: expected resource version to have 33 parts to match the number of ListerWatchers
```
This is due to an interaction between the kubernetes client-go cache code and the prometheus-operator multiListerWatcher.  The `multiListerWatcher` expects that the resourceVersion string passed to the `Watch()` function, which is used to determine the list of recent object updates, will always be the same one returned from the custom `List()` function.  However, this is not the case, as the client-go cache will update the `resourceVersion` directly from one of the event objects to ensure we're only watching the latest changes.

The error message doesn't currently cause any serious issues, as the built in error handling successfully restarts the watch after the error is detected, however it would be better if we didn't get these errors.  Also it seems that our modifications to the `resourceVersion` goes against recommendations in the kubernetes documentation.

> [Resource versions must be treated as opaque by clients and passed unmodified back to the server.](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions)

We discussed this issue and some possible solutions with some of the team more familiar with the API server internals, and the recommended solution instead of using a custom `MultiNamespaceListerWatcher` we should use a separate `Informer` for each namespace.  This would avoid any modifications of the `resourceVersion` and would limit our interaction with the client_go libraries to the higher level and more commonly used API.

lilic:
Still valid.

lilic:
@s-urbaniak I believe we can close this now since the new listwatch was merged?

simonpasquier:
Yeah, it should be fixed by https://github.com/prometheus-operator/prometheus-operator/pull/3440.

KielChan:
I have encounter with same problem in kube-state-metrics.

PR #3340, opened 2020-07-17
listwatch: do not duplicate resource versions

lilic:
This fixes a known bug https://github.com/coreos/prometheus-operator/issues/3218 as well as an unknown until the 0.39 release where List and Watch were producing a large amount of errors most likely due to a combination of bumping client-go, using multi namespace watcher and removing the CRD management done by operator. This left the option of the multi list watcher to create a wrong set of resourceVersions because of duplicates, most likely due to large amount of resources filling the cache at startup. As all of these things combined is very rare, even hard to reproduce, which is why this was most likely not experienced by a lot of users, hence lack of bug reports. 

Note: I really want to move away from using a custom multi namespace watcher very soon so did not spend too much time on refactor. https://github.com/coreos/prometheus-operator/issues/3339

This was tested on both vanilla k8s 1.18.3 and more complex setup of OpenShift monitoring where we make heavy use of this feature.

cc @coreos/team-monitoring 

Thanks @mfojtik for the initial fix.

Fixes https://github.com/coreos/prometheus-operator/issues/3218

### DIFF
--- a/pkg/listwatch/listwatch.go
+++ b/pkg/listwatch/listwatch.go
@@ -169,14 +169,14 @@ func (mlw multiListerWatcher) List(options metav1.ListOptions) (runtime.Object, 
 // It returns a watch.Interface that combines the output from the
 // watch.Interface of every cache.ListerWatcher into a single result chan.
 func (mlw multiListerWatcher) Watch(options metav1.ListOptions) (watch.Interface, error) {
-	resourceVersions := make([]string, len(mlw))
+	var resourceVersions string
 	// Allow resource versions to be "".
 	if options.ResourceVersion != "" {
 		rvs := strings.Split(options.ResourceVersion, "/")
-		if len(rvs) != 1 {
-			return nil, fmt.Errorf("expected resource version to have 1 parts to match the number of ListerWatchers, got %d", len(rvs))
+		if len(rvs) > 1 {
+			return nil, fmt.Errorf("expected resource version to have 1 part, got %d", len(rvs))
 		}
-		resourceVersions = rvs
+		resourceVersions = options.ResourceVersion
 	}
 	return newMultiWatch(mlw, resourceVersions, options)
 }
@@ -190,9 +190,8 @@ type multiWatch struct {
 }
 
 // newMultiWatch returns a new multiWatch or an error if one of the underlying
-// Watch funcs errored. The length of []cache.ListerWatcher and []string must
-// match.
-func newMultiWatch(lws []cache.ListerWatcher, resourceVersions []string, options metav1.ListOptions) (*multiWatch, error) {
+// Watch funcs errored.
+func newMultiWatch(lws []cache.ListerWatcher, resourceVersions string, options metav1.ListOptions) (*multiWatch, error) {
 	var (
 		result   = make(chan watch.Event)
 		stopped  = make(chan struct{})
@@ -204,7 +203,7 @@ func newMultiWatch(lws []cache.ListerWatcher, resourceVersions []string, options
 
 	for _, lw := range lws {
 		o := options.DeepCopy()
-		o.ResourceVersion = resourceVersions[0]
+		o.ResourceVersion = resourceVersions
 		w, err := lw.Watch(*o)
 		if err != nil {
 			return nil, err
--- a/pkg/listwatch/listwatch.go
+++ b/pkg/listwatch/listwatch.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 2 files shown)


simonpasquier:
lgtm

pkg/listwatch/listwatch.go
```diff
@@ -170,8 +173,8 @@ func (mlw multiListerWatcher) Watch(options metav1.ListOptions) (watch.Interface
 	// Allow resource versions to be "".
 	if options.ResourceVersion != "" {
 		rvs := strings.Split(options.ResourceVersion, "/")
-		if len(rvs) != len(mlw) {
-			return nil, fmt.Errorf("expected resource version to have %d parts to match the number of ListerWatchers", len(mlw))
+		if len(rvs) != 1 {
+			return nil, fmt.Errorf("expected resource version to have 1 parts to match the number of ListerWatchers, got %d", len(rvs))
```
squat, line 177: I don't understand the logic here. The reason we combine multiple resource versions with a slash is to later be able to distribute the resources to multiple watchers. Yet here we are enforcing that there should only be one version. So why not simply not join versions in the first place rather than fail if we did join then?

pkg/listwatch/listwatch.go
```diff
@@ -199,9 +202,9 @@ func newMultiWatch(lws []cache.ListerWatcher, resourceVersions []string, options
 
 	wg.Add(len(lws))
 
-	for i, lw := range lws {
+	for _, lw := range lws {
 		o := options.DeepCopy()
-		o.ResourceVersion = resourceVersions[i]
+		o.ResourceVersion = resourceVersions[0]
```
squat, line 207: This will produce odd results if more than one listerwatcher is given since they'll now be watching from the wrong version. Why not enforce that lws has length == 1? Or that it is a single instance vs a slice?

pkg/listwatch/listwatch.go
lilic: Yes you are right but the new exception is that we always must have a single resource version (I realize this makes use of implemention details of resourceVersion right now, but until refactoring its reliable enough judging by our tests and manual testing) there is certainly room for refactoring but we would end up refactoring or removing the entire thing, the aim is to do the minimal to fix the bug yet not modify as this is aimed at 0.40 release. See https://github.com/coreos/prometheus-operator/issues/3218 for background of this specific one.

pkg/listwatch/listwatch.go
lilic: > Or that it is a single instance vs a slice?

Yes exactly we could simply not pass in a slice and instead a single instance and refactor the entire function, but as mentioned in the description I did not want to focus too much on refactoring the entire thing as this fixes a bug in 0.40 version, but in master as per the linked issue we want to move away from custom multi listwatcher as we had a few bugs with it and it has known issues as we know now as from the research Paul did #3218. Hope that clarifies it? :)

brancz: I'm perfectly happy with a short term fix, just the string slice should be a single string, and we have a single call site, so it's a 3 line change from this one :)

squat: Agreed, as mentioned, let's make this a single string vs a slice to avoid confusion and unexpected results/usages

lilic: Sure will make the change, but FYI it's not a 3 line change from this one ;)

brancz: My apologies, if it does turn out to be way to complicated I'm happy with not doing it, just wanted us to try.

lilic: No worries, just need to adjust all the unit tests and retest on vanilla k8s and OpenShift clusters and CI so might take a day or two longer depending on CI. :) WIP until then.

squat: just chiming in that my fear is that without this last change we are leaving the library in a particularly weird place where it doesn't attempt to do the things it is supposedly designed for (even if the original impl has bugs). I am worried that this will be confusing and could lead to future issues or misunderstandings

brancz: understandable, I'm not too worried since this is just on the release branch, and not into master, but leaving it up to Lili to judge the trade-offs, I'm happy with going either way

lilic:
We have one failure:
>     alertmanager_test.go:287: failed to wait for initialized alertmanager cluster: the server is currently unable to handle the request (get pods alertmanager-test-1): failed to get correct amount of peers, expected 3, got 1, addresses [{01EDNS7FGW81R8TNVCDWT0MPRF 172.17.0.5:9094}]

But I have checked and seems like we have had this failure or similar before a few times. Will restart the build.

lilic:
@brancz @squat PTAL I have addressed the comment, tested it on vanilla k8s and openshift and it works correctly.

s-urbaniak:
I believe the code here introduces cascading failures again. The way I read the code here is that we _might_ join resource versions in `List` [1] but we will _always_ err out in `Watch` whenever that join happened in a previous `List`.

[1] https://github.com/prometheus-operator/prometheus-operator/blob/0dca0f21ffff72a063db8855b5d515e15ab0dccb/pkg/listwatch/listwatch.go#L164

[2] https://github.com/prometheus-operator/prometheus-operator/blob/0dca0f21ffff72a063db8855b5d515e15ab0dccb/pkg/listwatch/listwatch.go#L176

s-urbaniak:
Also, I am trying to understand the reasoning for the fix. Do we assume here, that we will _always_ get only _one_ resource version from all underlying watchers the rationale being that API server always returns the latest resource version independent of the namespace being watched?

PR #3348, opened 2020-07-20
Fix Syntax error in alert manager network policy

austinbv:
There is a small syntax error in the api for a network policy. This should make it easier for people to copy and paste in NPs


brancz:
The file content is templated from `/example/networkpolicies/alertmanager.yaml`, could you update there as well?

PR #3355, opened 2020-07-23
add SecretListWatchSelector to reduce memory and CPU footprint

### LINKED ISSUE #3298: Prometheus Operator watching every secret leads to high memory usage

Sayrus:
Fixes #3298 

Supersedes #3299 

This PR allows us to set a field selector. This follows the discussion about not wanting an opinionated selector in the code. With this option we can both exclude:
```
--secret-field-selector='type!=helm.sh/release.v1'
```
and select:
```
--secret-field-selector='type=prometheus.io/whatever-we-want'
```

Because this is a field selector, we can chain the selectors using `,`.
The only restriction is that you can't use an OR operator here.

Moreover, the default value (empty string) is equivalent to `fields.Everything` so this is not a breaking change !

⚠️ Note, some errors may come from the WatchList if the Field Selector is valid but refused by the Kube API (for instance: field label <xxx> is not supported)
While they are properly reported, it seems that the Prometheus Operator is still considered "healthy".

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -185,6 +185,7 @@ func init() {
 	flagset.StringVar(&cfg.PromSelector, "prometheus-instance-selector", "", "Label selector to filter Prometheus Custom Resources to watch.")
 	flagset.StringVar(&cfg.AlertManagerSelector, "alertmanager-instance-selector", "", "Label selector to filter AlertManager Custom Resources to watch.")
 	flagset.StringVar(&cfg.ThanosRulerSelector, "thanos-ruler-instance-selector", "", "Label selector to filter ThanosRuler Custom Resources to watch.")
+	flagset.StringVar(&cfg.SecretListWatchSelector, "secret-selector", "", "Field selector to filter Secrets to watch")
 }
 
 func Main() int {
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -185,7 +185,7 @@ func init() {
 	flagset.StringVar(&cfg.PromSelector, "prometheus-instance-selector", "", "Label selector to filter Prometheus Custom Resources to watch.")
 	flagset.StringVar(&cfg.AlertManagerSelector, "alertmanager-instance-selector", "", "Label selector to filter AlertManager Custom Resources to watch.")
 	flagset.StringVar(&cfg.ThanosRulerSelector, "thanos-ruler-instance-selector", "", "Label selector to filter ThanosRuler Custom Resources to watch.")
-	flagset.StringVar(&cfg.SecretListWatchSelector, "secret-selector", "", "Field selector to filter Secrets to watch")
+	flagset.StringVar(&cfg.SecretListWatchSelector, "secret-field-selector", "", "Field selector to filter Secrets to watch")
 }
 
 func Main() int {
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -153,6 +153,7 @@ type Config struct {
 	PromSelector                  string
 	AlertManagerSelector          string
 	ThanosRulerSelector           string
+	SecretListWatchSelector       string
 }
 
 type Namespaces struct {
@@ -200,6 +201,11 @@ func New(conf Config, logger log.Logger, r prometheus.Registerer) (*Operator, er
 
 	if _, err := labels.Parse(conf.AlertManagerSelector); err != nil {
 		return nil, errors.Wrap(err, "can not parse alertmanager selector value")
+	}
+
+	secretListWatchSelector, err := fields.ParseSelector(conf.SecretListWatchSelector)
+	if err != nil {
+		return nil, errors.Wrap(err, "can not parse secrets selector value")
 	}
 
 	kubeletObjectName := ""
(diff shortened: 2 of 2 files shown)


brancz:
Awesome! Thank you!

matthyx:
> ⚠️ Note, some errors may come from the WatchList if the Field Selector is valid but refused by the Kube API (for instance: field label is not supported)

@brancz does it mean it's expected to see such errors in the logs:
```
E1102 14:10:47.388291       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.8/tools/cache/reflector.go:125: Failed to list *v1.Secret: field label not supported: "type
E1102 14:11:40.082751       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.8/tools/cache/reflector.go:125: Failed to list *v1.Secret: field label not supported: "type
```

I am trying to filter out secrets of type dockercfg with the setting:
`- '--secret-field-selector="type!=kubernetes.io/dockercfg"'`

matthyx:
> I am trying to filter out secrets of type dockercfg with the setting:
> `- '--secret-field-selector="type!=kubernetes.io/dockercfg"'`

Note on that issue... it has been solved by removing the quotes using scalar block chomping:
```
            - >-
              --secret-field-selector=type!=kubernetes.io/dockercfg
```

Sayrus:
> > I am trying to filter out secrets of type dockercfg with the setting:
> > `- '--secret-field-selector="type!=kubernetes.io/dockercfg"'`
> 
> Note on that issue... it has been solved by removing the quotes using scalar block chomping:
> 
> ```
>             - >-
>               --secret-field-selector=type!=kubernetes.io/dockercfg
> ```

Now you make me doubt my own implementation :fearful: 
Does ` - --secret-field-selector='type!=kubernetes.io/dockercfg'` work? (Not using double quotes and only single quoting the value.

Note that the error I was talking about is not related to what you had where it says `"type`. The parsing failed as you still have a double quote in your value. The error I talked about occurs when you try to filter using a field that is not usable in field selectors. Removing the quotes would not help as it would still be the same field.
You get the same error because the misquoted field is an unknown field, which means it also is an unsupported field label.

matthyx:
Single quote doesn't work as well...
`E1104 12:46:58.570454       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.8/tools/cache/reflector.go:125: Failed to list *v1.Secret: field label not supported: 'type`

matthyx:
Filtering with `type` works, so it's clearly a flag parsing issue...

Sayrus:
Then this means that your variables are already escaped, see the following:
```
$ go run test.go --var='test'
var: test
```
```
$ go run test.go "--var='test'"
var: 'test'
```

matthyx:
As a Kubernetes deployment... but it's not the only argument:
```
    - args:
        - '--kubelet-service=kube-system/kubelet'
        - '--logtostderr=true'
        - '--prometheus-instance-selector=prometheus=lo'
        - '--alertmanager-instance-selector=prometheus=lo'
        - >-
          --secret-field-selector=type!=kubernetes.io/dockercfg,type!=kubernetes.io/service-account-token
```

I think quotes are inherited by the yaml handling...

Sayrus:
Sounds like you should be able to use unquoted values as long as they are valid YAML. Thanks for documenting the issue and providing a workaround.

simonpasquier (commented):
Thanks! The same change needs to be applied to pkg/prometheus/agent/operator.go, and pkg/prometheus/server/operator.go.

PR #3358, opened 2020-07-24
Propagate Prometheus image version to statefulset

jbfavre:
Since `0.35.0` release, some fields have been deprecated:

- `spec.baseImage`
- `spec.tag`
- `spec.sha`

In code, pkg/prometheus/statefulset.go line 284 and later, operator build prometheusImagePath using deprecated fields and spec.Version. At line 287, if spec.Image is defined, operator will simply replace prometheusImagePath value with spec.Image value.

This skip the version, which is not propagated anymore to statefulset definition.
This patch intends to be able to use either version or tag information to set Prometheus version for statefulset.
This brings back the old behaviour while keeping the deprecation mechanism

### DIFF
--- a/pkg/prometheus/statefulset_test.go
+++ b/pkg/prometheus/statefulset_test.go
@@ -579,7 +579,7 @@ func TestTagAndShaAndVersion(t *testing.T) {
 		image := "my-reg/prometheus"
 		sset, err := makeStatefulSet(monitoringv1.Prometheus{
 			Spec: monitoringv1.PrometheusSpec{
-				Image:   &image,
+				Image: &image,
 			},
 		}, defaultTestConfig, nil, "")
 		if err != nil {
@@ -596,8 +596,8 @@ func TestTagAndShaAndVersion(t *testing.T) {
 		image := "my-reg/prometheus"
 		sset, err := makeStatefulSet(monitoringv1.Prometheus{
 			Spec: monitoringv1.PrometheusSpec{
-				SHA:     "7384a79f4b4991bf8269e7452390249b7c70bcdd10509c8c1c6c6e30e32fb324",
-				Image:   &image,
+				SHA:   "7384a79f4b4991bf8269e7452390249b7c70bcdd10509c8c1c6c6e30e32fb324",
+				Image: &image,
 			},
 		}, defaultTestConfig, nil, "")
 		if err != nil {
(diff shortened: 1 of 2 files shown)


brancz:
Could we add some tests for these scenarios?

jbfavre:
@brancz sure I can. What scenarios do you want to test ?
I'll also fix the `TestTagAndShaAndVersion` unit test

pkg/prometheus/statefulset_test.go
```diff
@@ -513,11 +533,120 @@ func TestTagAndShaAndVersion(t *testing.T) {
 		}
 
 		resultImage := sset.Spec.Template.Spec.Containers[0].Image
-		expected := "my-reg/prometheus:latest"
+		expected := "my-reg/prometheus:latest@sha256:7384a79f4b4991bf8269e7452390249b7c70bcdd10509c8c1c6c6e30e32fb324"
```
brancz, line 536: is this right? if I'm not mistaken I believe this would not work

jbfavre: Looks like it's working:
```
> docker pull debian
Using default tag: latest
latest: Pulling from library/debian
31dd5ebca5ef: Pull complete 
Digest: sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
Status: Downloaded newer image for debian:latest
docker.io/library/debian:latest

> docker pull debian:latest
latest: Pulling from library/debian
Digest: sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
Status: Image is up to date for debian:latest
docker.io/library/debian:latest

> docker pull debian@sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a: Pulling from library/debian
Digest: sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
Status: Image is up to date for debian@sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
docker.io/library/debian@sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a

> docker pull debian:latest@sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a: Pulling from library/debian
Digest: sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
Status: Image is up to date for debian@sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
docker.io/library/debian:latest@sha256:68f4e2259032a4e6f5035804e64438b52af8dd5889528b305b9059183ea4cd2a
```

PR #3361, opened 2020-07-24
test/framework: fix prometheus active targets polling

dgrisonnet:
In order to check if prometheus is able to scrape a target, we are
querying the /api/v1/targets endpoint and checking if the number of
active targets is equal to the one we expect. However, active targets
displayed by the API can sometimes be unhealthy. For example, when the
scraped target responds 401 Unauthorized, the target is still considered
active but its health will be set to "down". Because of that,
unauthorized scrapes will be considered valid in the tests.

Thus, in order to verify that prometheus is allowed to scrape a target,
we also need to check if it is healthy.

Example of active target being down:
```json
{
    "status": "success",
    "data": {
        "activeTargets": [
            {
                "..."
                "lastError": "server returned HTTP status 401 Unauthorized",
                "lastScrape": "2020-07-24T09:54:42.018550304Z",
                "lastScrapeDuration": 0.001038429,
                "health": "down"
            }
        ],
    }
}

```

### DIFF
--- a/test/framework/prometheus.go
+++ b/test/framework/prometheus.go
@@ -324,7 +324,8 @@ func promImage(version string) string {
 	return fmt.Sprintf("quay.io/prometheus/prometheus:%s", version)
 }
 
-func (f *Framework) WaitForTargets(ns, svcName string, amount int) error {
+// WaitForActiveTargets waits for a number of targets to be configured.
+func (f *Framework) WaitForActiveTargets(ns, svcName string, amount int) error {
 	var targets []*Target
 
 	if err := wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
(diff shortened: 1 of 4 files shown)


test/framework/prometheus.go
```diff
@@ -494,9 +538,18 @@ func (f *Framework) WaitForPrometheusFiringAlert(ns, svcName, alertName string)
 	return nil
 }
 
+type targetHealth string
+
+const (
+	healthGood targetHealth = "up"
```
brancz, line 544: we shouldn't rename these, the variable names should just be `healthUp` and `healthDown`

dgrisonnet: I mirrored what is exposed by Prometheus scrape package here: https://github.com/prometheus/prometheus/blob/master/scrape/target.go#L43-L44

brancz: ok makes sense, at least consistent in one way :)

PR #3362, opened 2020-07-24
test/e2e: fix scraping path in bearer-token test

dgrisonnet:
Fix the metrics scraping path of the bearer token e2e test to match the [endpoint](https://github.com/coreos/prometheus-operator/blob/master/test/instrumented-sample-app/main.go#L59-L68) of the sample app.


dgrisonnet:
Moved to #3361 as the framework fix highlights this issue.

PR #3369, opened 2020-07-27
docs: update compatibility warning

abursavich:
Update existing compatibility warning in docs to reflect the switch to apiextensions.k8s.io/v1 CRD in v0.39.0 which requires kubernetes v1.16 or newer (#3175, #3187).


brancz:
Thank you so much for this, I think this is very useful!

It appears a formatter is complaining https://travis-ci.org/github/coreos/prometheus-operator/jobs/712349042, otherwise this lgtm!

abursavich:
Reverted accidental whitespace change. Should be good now.

PR #3373, opened 2020-07-29
listwatch: do not duplicate resource versions

lilic:
This cherry-picks the https://github.com/coreos/prometheus-operator/pull/3340 into master.

cc @brancz @paulfantom


brancz:
lgtm

Issue #3354, opened 2020-07-23
"targetPort" do not work in "ServiceMonitor"


vincent-pli:
**What happened?**
Try to use `targetPort` in `serviceMonitor` to retrieval data.
Seems it's not work, so I prepare two SVC point to same pod, one has port name, another not.
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/name: helloword-1
  name: helloword-service-1
  namespace: helloword
spec:
  ports:
  - port: 10501
    protocol: TCP
    targetPort: 10501
  selector:
    app.kubernetes.io/name: helloword-pod
  type: ClusterIP
```
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/name: helloword-2
  name: helloword-service-2
  namespace: helloword
spec:
  ports:
  - name: metrics 
    port: 10501
    protocol: TCP
    targetPort: 10501
  selector:
    app.kubernetes.io/name: helloword-pod
  type: ClusterIP
```

Then create two `serviceMonitor` for each:
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    k8s-app: prometheus
  name: my-hpac-1
  namespace: openshift-monitoring
spec:
  endpoints:
  - interval: 10s
    targetPort: 10501
  namespaceSelector:
    matchNames:
    - helloword
  selector:
    matchLabels:
      app.kubernetes.io/name: helloword-1
```
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: my-hpac-2
  namespace: openshift-monitoring
  labels:
    k8s-app: prometheus
spec:
  endpoints:
  - interval: 10s
    port: metrics
  namespaceSelector:
    matchNames:
    - helloword
  selector:
    matchLabels:
      app.kubernetes.io/name: helloword-1
```

The second on (with `port`) works, but the first one(with `targetPort` not work)
I checked the status on GUI of Prometheus: status-service discovery
```
openshift-monitoring/my-hpac-1/0 (0/2 active targets)
```

So, any suggestion, thanks.

vincent-pli:
I figure out the issue, when use `targetPort` the related pod must specify a container port, or it will not work.
I'm new for Prometheus, where can I fine document for it, thanks.

brancz:
[...]

PR #3374, opened 2020-07-29
Clarify targetPort in endpoint

vincent-pli:
Fix issue: #3354


brancz:
API docs are generated, you need to also change this here https://github.com/coreos/prometheus-operator/blob/c9758cf8de3ef21973c03e98d8e4c01a2d4bf0a9/pkg/apis/monitoring/v1/types.go#L697

brancz:
I recommend to just run

```bash
make --always-make format generate
```

this will format and re-generate everything correctly, then CI should pass

vincent-pli:
@brancz 
Thanks for helping 😄

brancz:
Thanks for bearing with us! :) 

lgtm :tada:

PR #3377, opened 2020-07-30
🐛 Don't overwrite __param_target

coderanger:
It is already set above using the sd metadata, no need to overwrite it back to __address__.


brancz:
ah unit test test-fixtures need fixing as this is gone now

brancz:
Thank you!

PR #3392, opened 2020-08-06
pkg/operator/image.go: Adjust image path building

lilic:
Image can contain tag already, this checks if it does it just returns
the image. Otherwise sha/digest takes priority over tag and lastly
version is taken due to historic reasons.

cc @prometheus-operator/prometheus-operator-reviewers

### DIFF
--- a/pkg/operator/image.go
+++ b/pkg/operator/image.go
@@ -17,26 +17,38 @@ package operator
 import (
 	"fmt"
 	"strings"
+
+	dockerref "github.com/docker/distribution/reference"
 )
 
 // BuildImagePath builds a container image path based on
 // the given parameters.
-// baseImage and version are used by default.
-// If the tag is specified, we use the tag to identify the container image.
-// If the sha is specified, we use the sha to identify the container image,
-// as it has even stronger immutable guarantees to identify the image.
-func BuildImagePath(baseImage, version, tag, sha string) string {
-	image := baseImage
-	if version != "" {
-		image = fmt.Sprintf("%s:%s", baseImage, version)
+// If the image contains tag or digest then image will be returned.
+// Inspired by kubernetes code handling of image building.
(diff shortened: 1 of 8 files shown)


brancz:
Looks like we need to adapt some of the unit test expectations. What this PR does is how I expect image handling to work though, so fixing the unit test expectations is the right thing to do/

lilic:
Fixed the unit tests @brancz can you please have another look, thanks!

pkg/operator/image.go
```diff
@@ -17,26 +17,34 @@ package operator
...
+func BuildImagePath(image, version, tag, sha string) (string, error) {
+	// If the image contains no tag or digest, a default tag should be applied.
+	named, err := dockerref.ParseNormalizedNamed(image)
+	if err != nil {
+		return "", fmt.Errorf("couldn't parse image reference %q: %v", image, err)
 	}
-	if sha != "" {
-		image = fmt.Sprintf("%s@sha256:%s", baseImage, sha)
+	_, isTagged := named.(dockerref.Tagged)
+	_, isDigested := named.(dockerref.Digested)
```
s-urbaniak, line 33: nice :+1:

pkg/operator/image.go
```diff
@@ -17,26 +17,34 @@ package operator
...
+	if !isTagged && !isDigested {
```
s-urbaniak, line 34: nit: instead of having a big if block here, return early:

```
if isTagged || isDigested {
  return image, nil
}
```
and reindent the complex if-else logic:
```
  if sha != " { ...
  } else if version != "" {
 ..
  }
return image, nil
```

lilic:
@s-urbaniak addressed your comment, PTAL, thanks!! 🎉

s-urbaniak:
lgtm :tada:

PR #3395, opened 2020-08-07
Break the API types out into their own module

### LINKED ISSUE #3097: Moving APIs to separate package
**What did you do?**
I've attempted to use the Prometheus Golang types inside another operator and due to the fact that means I have to import all of `coreos/prometheus-operator` -- It results in some conflicts with my own dependencies.

It would be nice if the project could look into moving it's types out to a seperate repo so users that interact with the Golang Kubernetes APIs can reference them

Something similar to this: https://github.com/kubernetes/api

**Environment**
N/A

N/A

N/A

* Kubernetes cluster kind:
N/A

* Manifests:
N/A

* Prometheus Operator Logs:
N/A

I would be willing to do the work if this is something the team agrees with

matthiasr:
This allows others to import them without incurring all the dependencies
of the operator transitively, and avoid version conflicts with other
dependencies as much as possible.

Fixed #3097.


matthiasr:
Ah yes, thank you for the reminder. The problem is that `controller-gen` cannot deal with multiple modules very well:

```
/home/travis/gopath/src/github.com/prometheus-operator/prometheus-operator/tmp/bin/controller-gen object:headerFile=./.header \

	paths=./pkg/apis/monitoring/v1

Error: go [-e -json -compiled=true -test=false -export=false -deps=true -find=false -tags ignore_autogenerated -- ./pkg/apis/monitoring/v1]: exit status 1: go: directory pkg/apis/monitoring/v1 is outside main module
```

From manual testing, I think the solution is to exec controller-gen from inside the new module. It will take me some time to get to implementing that.

matthiasr:
After rebasing on top of #3432 the tests now passed!

s-urbaniak:
Nice, thank you! Small and tidy change :tada:

LGTM from my side, but I'd like to have another two eyes for verification, cc @prometheus-operator/prometheus-operator-reviewers

metalmatze (commented):
If it works I'm happy with it, haven't actually tried separating sub-packages into own packages. Cool! :+1:

carlpett:
I may be missing something here, but as an external consumer of these types, don't I still need to depend on the "main" prometheus-operator module in order to access the Kubernetes client types?

carlpett:
That's what I thought, but
```golang
import (
    "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1" // Separate module
    "github.com/coreos/prometheus-operator/pkg/client/versioned" // Not under pkg/apis?
)
```

If I want to eg have a watcher on a ServiceMonitor, I believe I need both those?

lilic:
We have this issue https://github.com/prometheus-operator/prometheus-operator/issues/3497 I believe it's related?

simonpasquier:
> If I want to eg have a watcher on a ServiceMonitor, I believe I need both those?

That's correct. This PR enables to consume the CRD structs without depending on the whole prometheus-operator code base but it didn't separate the client part.

carlpett:
Is this by design, or something to improve? Should I as a downstream consumer rather do autogeneration for the client bits?

s-urbaniak:
I think exporting the generated client should also be covered, not only the types, in my opinion we could make a follow-up PR exporting those bits too.

matthiasr:
Good catch, I didn't think about the client at all; we only use the type definitions to generate JSON that we apply separately.

PR #3396, opened 2020-08-07
WIP pkg/listwatch: err out with multiple resource versions in List

s-urbaniak:
https://github.com/prometheus-operator/prometheus-operator/pull/3340 added a workaround to err out if multiple resource versions are reported from api server.

However the preceeding List function allows multiple resource versions to slip through.
This causes potential cascading failures.

Note: as #3340 this is not a final fix but a try to cause additional cascading failures due to different resource versions being reported by API server.

/cc @lilic @brancz @paulfantom @metalmatze

### DIFF
--- a/pkg/listwatch/listwatch.go
+++ b/pkg/listwatch/listwatch.go
@@ -159,6 +159,11 @@ func (mlw multiListerWatcher) List(options metav1.ListOptions) (runtime.Object, 
 			resourceVersions.Insert(metaObj.GetResourceVersion())
 		}
 	}
+
+	if len(resourceVersions) > 1 {
+		return nil, fmt.Errorf("list: expected resource version to have 1 part, got %d", len(resourceVersions))
+	}
+
 	// Combine the resource versions so that the composite Watch method can
 	// distribute appropriate versions to each underlying Watch func.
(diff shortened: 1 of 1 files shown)


s-urbaniak:
Note: this PR doesn't fix the issue the same way https://github.com/prometheus-operator/prometheus-operator/pull/3340 didn't fix it.

It is just the continuation of the #3340 workaround for situations where distinct resourceVersions are being reported at startup time of prom-operator from api server.

#3340 err'ed out in this edge case in Watch only with the hope that eventually api server will reach consensus and report just one resourceVersion.

However, what happens is that List does not err out, passing multiple / separated resourceVersions to the underlying api machinery code, causing cascading failures like missed reconciliations and underlying ParseUint errors not reaching a stable state.

With adding explicit error handling the hope is that we introduce the initial intention: err out as early as possible with the hope that the system will retry the List operation converging against a stable state.

I will execute a couple of e2e runs in the openshift test infrastructure and see if we can observe the initial intended converging behavior.

lilic:
@s-urbaniak can you do a PR to merge this into both 0.41 release and into master as they are both affected by this bug, thanks!

s-urbaniak:
@lilic definitely that is the plan :+1: please give me a little time to investigate if this actually hot-fixes the issue.

PR #3402, opened 2020-08-11
[0.41]: pkg/operator/image.go: Adjust image path building

lilic:
This ports the fix that landed in master. cc @s-urbaniak @brancz PTAL.

Should we mention this in the release notes of 0.41 or how do we go forward when adding a patch like this?


s-urbaniak:
@lilic i believe, adding a not in the release notes for 0.41 is definitely worth it and thank you @paulfantom for the entry in https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.41.1 :+1:

PR #3409, opened 2020-08-12
WIP pkg/listwatch: remove multilistwatcher

s-urbaniak:
*Note*: this is an experiment, please do not merge but of course feel free to review, leave comments to assert if this a viable path forward at all.


brancz:
> please do not merge/review

can we comment at all? :D

s-urbaniak:
@brancz yes please! :-D After initial testing I am pretty sure I do not want to move forward with this and it also showed false assumptions in the original denylisting filter.

side note: this is literally just an experiment.

brancz:
Yeah I don't think we can do this, in particular because it broadens permissions needed by the Prometheus Operator which breaks setups, and we especially cannot do this in a patch release.

brancz:
Beyond that, the whole reason of having individual list watches was because we knew there were circumstances possible where a single object could be so malformed that it could prevent the entire prometheus operator from working. We haven't had this case in a while but that was part of the initial premise, and I would be hesitant to give up this characteristic because we want to fix another unrelated bug.

s-urbaniak:
closing out as we cannot pursue this path.

bparees:
> Beyond that, the whole reason of having individual list watches was because we knew there were circumstances possible where a single object could be so malformed that it could prevent the entire prometheus operator from working. 

not sure how that applies here.  If we're filtering the objects on namespace, then what sort of malformation could break us before we managed to filter out objects from namespaces we don't care about?

brancz:
The unmarshaling of the list request failing, if that fails 100% of prometheus operator doesn't work. This is what initially made us segregate these so we can have some sort of insurance for namespaces we control, but not fail across the board.

bparees:
> The unmarshaling of the list request failing, if that fails 100% of prometheus operator doesn't work. 

sure, but how does that happen?  someone managed to create the object in etcd, why can it now not be marshalled/unmarshalled w/o error?

there are many operators that watch all namespaces for CRDs, so if this is really a likely issue, there are a great number of components that would be exposed to it

brancz:
For something entirely new I agree with you, but for historic reasons this may be the case with the Prometheus Operator, when CRs did not necessarily get validated at all, we introduced this feature for literally this reason, so removing it in a patch release would be hostile as a project.

PR #3413, opened 2020-08-17
Fix validation logic for SecretOrConfigMap

matthiasr:
This was flagged by
[golangci-lint](https://staticcheck.io/docs/checks#SA4022). The check
was for the address of the pointer, not the value. Add a test (failing
on master) to verify this, and fix the validation logic.

Follow-up to #2716.

### DIFF
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -897,7 +897,7 @@ func (e *SecretOrConfigMapValidationError) Error() string {
 
 // Validate semantically validates the given TLSConfig.
 func (c *SecretOrConfigMap) Validate() error {
-	if &c.Secret != nil && &c.ConfigMap != nil {
+	if c.Secret != nil && c.ConfigMap != nil {
 		return &SecretOrConfigMapValidationError{"SecretOrConfigMap can not specify both Secret and ConfigMap"}
 	}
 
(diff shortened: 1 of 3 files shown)


matthiasr:
The main question I have is: how has this not become a problem so far? I feel like I am missing something about how all this works together; if I understood the bug correctly this should fail any and all TLS configs?

simonpasquier:
Thanks a lot!

> how has this not become a problem so far?

If I read the code correctly, it would only manifest if you've set `arbitraryFSAccessThroughSMs` to `true` in your Prometheus spec and define a non-nil TLSConfig. I'll do some additional testing.

matthiasr:
aha, that partially explains why exactly those two e2e tests are failing now …

s-urbaniak:
@matthiasr great catch thank you :+1:

matthiasr:
I believe the `denied-tls-secret` and `denied-tls-configmap` e2e tests are wrong – they are testing that local filesystem access is denied, but the configuration that they give does not perform local filesystem access. They passed because [validation always failed](https://github.com/prometheus-operator/prometheus-operator/blob/5c3520960e179458c0c98dae94d4cae65f71aa7f/pkg/prometheus/operator.go#L2067), but they should not pass.

s-urbaniak:
LGTM on green.

PR #3415, opened 2020-08-18
pkg: add prometheus_operator_reconcile_operations_total metric

simonpasquier:
We already have the `prometheus_operator_reconcile_errors_total` metric
to track the number of reconciliation attempts that failed but we miss
the number of attempts which makes it harder to alert on it. With this
change, we can compute the ratio of reconciliations that failed.


s-urbaniak:
very nice, LGTM on green, thank you!

PR #3417, opened 2020-08-18
test/e2e: remove duplicated subtests

simonpasquier:
After #3413, it appears that some of the sub-tests in
`testPromArbitraryFSAcc` aren't useful because the outcome doesn't
depend on the value of ArbitraryFSAccessThroughSMsConfig.Deny. In all
cases, the test should succeed.


s-urbaniak:
LGTM

PR #3421, opened 2020-08-19
Add metrics for selected/rejected resources and sync status

### LINKED ISSUE #3327: Add new metric when prometheus is stuck on "creating config failed"
**What is missing?**

There is no metric to alert when prometheus fail to create the new configuration before reload it.

**Context:**
We work on a shared kubernetes cluster with different teams and someone deleted, by mistake, one secret used to get the api credentials used by the service monitor. And because of that prometheus started to throw this error:

```
E0709 12:11:36.444821       1 operator.go:994] Sync "prometheus-operator" failed: creating config failed: could not generate basicAuth for servicemonitor my-app. secret username key "user" in secret "basic-auth" not found
```

And because of this error all the new service monitor we configure are not added to the configuration (maybe a related bug?)

**Why do we need it?**

We need to know when the create configuration procedure is failing to fix it as soon we notice to improve our reliability.

**Environment**

    `v0.38.1`

simonpasquier:
This change adds two metrics:
* `prometheus_operator_managed_resources`, the number of resources managed by the controller per state (either selected or rejected).
* `prometheus_operator_syncs`, number of sync operations per status and controller.
    
It also adds a new `PrometheusOperatorSyncFailed` alert that fires when `prometheus_operator_syncs{status="failed"}` is greater than 0.

Closes #3327

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -280,6 +280,7 @@ func (c *Operator) processNextWorkItem(ctx context.Context) bool {
 
 	c.metrics.ReconcileCounter().Inc()
 	err := c.sync(ctx, key.(string))
+	c.metrics.SetSyncStatus(key.(string), err == nil)
 	if err == nil {
 		c.queue.Forget(key)
 		return true
@@ -399,6 +400,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	aobj, err := c.alrtInfs.Get(key)
 
 	if apierrors.IsNotFound(err) {
+		c.metrics.ForgetObject(key)
 		// Dependent resources are cleaned up by K8s via OwnerReferences
 		return nil
 	}
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -16,16 +16,33 @@ package operator
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+
 	"github.com/prometheus/client_golang/prometheus"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
+)
+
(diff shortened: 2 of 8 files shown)


pkg/alertmanager/operator.go
```diff
@@ -135,7 +135,7 @@ func New(c prometheusoperator.Config, logger log.Logger, r prometheus.Registerer
 		&monitoringv1.Alertmanager{}, resyncPeriod, cache.Indexers{},
 	)
 	o.metrics.MustRegister(NewAlertmanagerCollector(o.alrtInf.GetStore()))
-	o.metrics.MustRegister(operator.NewStoreCollector("alertmanager", o.alrtInf.GetStore()))
+	o.metrics.MustRegister(operator.NewStoreCollector(monitoringv1.AlertmanagersKind, o.alrtInf.GetStore()))
```
s-urbaniak, line 138: Let's start thinking how this can incorporate multiple stores too (not relevant for this PR though) since we plan to move away from multilistwatcher to a native array of informers/stores.

Any ideas? I attempted something in my WIP PR but maybe you have some idea too?

s-urbaniak (commented):
Looks great so far!

simonpasquier:
My initial approach is too naive because we need to track sync operations per prometheus/alertmanager/thanos object somehow. Working on improving this :)

simonpasquier:
This is ready for review.

pkg/operator/operator.go
```diff
@@ -39,6 +56,15 @@ type Metrics struct {
 	// objects. It is split in the dimensions of Kubernetes objects and
 	// corresponding actions (add, delete, update).
 	triggerByCounter *prometheus.CounterVec
+
+	mtx       sync.RWMutex
```
s-urbaniak, line 60: do you mind to quickly document which fields this mutex is protecting? A simple `// protects all fields below` should do the trick.

s-urbaniak:
looks great to me, leaving the final review to another pair of eyes, cc @prometheus-operator/kube-prometheus-reviewers

midnightconman:
Can we add a note to the changelog about the metrics name and behavior change? It took me a while to find where this change was introduced.

simonpasquier:
@midnightconman we'll try to do better in the future. Hopefully we now have a process to add changelog entries to the PRs directly and generate the release changelog from them.

PR #3427, opened 2020-08-20
Fix shellcheck Make target (ignore vendored files)

grdryn:
Before this change, the following error occurs for the shellcheck
target (which is a prerequisite for other targets like format):

```
In ./scripts/generate/vendor/github.com/ksonnet/ksonnet-lib/scripts/install_jsonnet.sh line 7:
JSONNET_VERSION=$
^-------------^ SC2034: JSONNET_VERSION appears unused. Verify use (or export if used externally).

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- JSONNET_VERSION appears unused. V...
make: *** [Makefile:246: shellcheck] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,7 +242,7 @@ check-license:
 
 .PHONY: shellcheck
 shellcheck: $(SHELLCHECK_BINARY)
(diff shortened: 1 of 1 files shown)


simonpasquier:
Thanks!

PR #3429, opened 2020-08-21
Support Job Name in service and pod monitors

omegas27:
Signed-off-by: Quentin Bisson <quentin@giantswarm.io>

Add support for jobName in service and pod monitors


simonpasquier:
Thanks for the PR! Can you explain what is your use case and why it couldn't be solved by the `jobLabel` and/or `relabelings`?

omegas27:
The main reason is for customization as we need to use additional scrape configs along with service monitors and the generated job names do not align with our naming scheme so we would like to be able to customize it. 
I know the job name is only visible on the Prometheus UI but we would like to be able to customize them anyway

brancz:
I think I'm generally ok with having this field, but I would prefer if the "original" job name continues to be the generated job name from the podmonitor/servicemonitor so it's easy to reference where each scrape config came from. Basically that just means that instead of the original job name we would set it via a pre-configured relabeling rule.

omegas27:
@brancz so if I understand correctly, we would keep the generated name but instead of setting it as the job name, we set it as a label like createdBy if we override the job name?

brancz:
yeah instead of setting it directly, set it via relabeling a "static" value, the static value being what's passed to the Pod/ServiceMonitors.

omegas27:
From what I found in the prometheus code base, there is no way to relabel the job name. We can only redefine the job label which is not what is shown in the prometheus UI under the list of targets.

I actually did this change similar to how it is handled for the Probe CR.

Would defining a source_service_monitor and source_pod-monitor label with the generated name be enough to know which ServiceMonitor or PodMonitor created it like so:
`relabelings = append(relabelings, yaml.MapSlice{
		{Key: "action", Value: "keep"},
		{Key: "source_labels", Value: []string{"source_service_monitor"}},
		{Key: "regex", Value: generatedJobName},
	})
`

Or did I totally misunderstood what you meant?

brancz:
The snippet [here](https://github.com/prometheus-operator/prometheus-operator/blob/433060233e9fa2f1c274bb0c2d0ea3dd66f302b9/pkg/prometheus/promcfg.go#L1041-L1044) sets the job label, so the targets all have that job label.

omegas27:
I think we are not talking about the same field.
What I would like to be able to customize is the Job Name visible in the UI and in the scrape configs instead of the generated names:
image

This change cannot actually be done using the job label as the job label is for targets and not scrape config jobs

brancz:
Ah yes, that is precisely what I do not want to be configurable. The whole point is that they are uniquely identifiable and cannot collide the way they exist. I would like to keep this characteristic. Can you explain your use cases?

omegas27:
Sorry for the delay in my answer.

We provide a management cluster that takes care of the lifecycle of tenant clusters and we currently host a prometheus in the management cluster.

As we are facing more issues with our amount of data, we decided to shard our setup by creating one prometheus per tenant cluster, while all those prometheuses are hosted in the management cluster.

The main reason we need this is because we also monitor node components like the kubelet, cadvisor, the CRI, CNI and other daemon components. 
As the prometheus operator does not support NodeMonitors, we need to define those node monitors as additional scrape configs and define the name ourselves. For the sake of homogeneity, we would need to have them match the generated ones or be able to override the generated ones, explaining this PR.

brancz:
I think for the sake of having names homogenous I would prefer to push on https://github.com/prometheus-operator/prometheus-operator/issues/2787 rather. Which would allow you to do any scrape config (such as those that you specify via additionalScrapeConfigs).

PR #3432, opened 2020-08-21
Avoid CI timeouts in TestConfigGeneration

matthiasr:
This test generates the same configuration many times, for each
Prometheus version, to see if it is deterministic. As the compatibility
matrix grows, test times increase. Now, this sometimes fails in CI
because Travis kills jobs after 10 minutes of no output.

Run each version as a subtest, and run tests with `-v`, so that output
is produced after each version. This avoids the no-output timeout.

Parallelize testing for each Prometheus version.

When the tests are run with `-short` (as in `make test-unit`), only try
one hundred iterations. With the race detector on, as in that target, this takes
around 5 seconds. Without the race detector, short tests on this
package now run quick enough for fast iteration in an IDE.

Add an additional target and Travis job for running the long tests, but
without the race detector. This brings the run time for the full 1000
iterations per version to under a minute.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ jobs:
     script: cd cmd/po-rule-migration && go install
   - name: "Run unit tests"
     script: make test-unit
+  - name: "Run extended tests"
+    script: make test-long
   - name: "Run e2e tests (Alertmanager)"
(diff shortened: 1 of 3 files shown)


simonpasquier (approved):
Thanks!

Issue #3433, opened 2020-08-22
hope add terminationGracePeriodSeconds support


lurenjia528:
**What is missing?**
hope add `sts.spec.template.spec.terminationGracePeriodSeconds`  support

**Why do we need it?**
releate  to　`https://github.com/prometheus-operator/prometheus-operator/issues/3425` ,

if prometheus(sts) is running on node-1,when node-1 is down, the prometheus(sts) do not recreate on other ready node,
yeah,sts is design  so , https://github.com/kubernetes/kubernetes/issues/54368#issuecomment-339378164 ,  but when set terminationGracePeriodSeconds=0, sts can recreate on other ready node,then ,this enables fault migration 

so ,I hope add terminationGracePeriodSeconds in [prometheusSpec](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheusspec)

brancz:
I agree with @coderanger that duplicating is in fact the correct strategy here, I don't want to expose a field that enables an anti-pattern.

simonpasquier:
Irrespective of the use case mentioned by @lurenjia528 (immediate shutdown), there might be a need to customize the `TerminationGracePeriodSeconds` to a higher value than the default 600s?
We've got a [report](https://bugzilla.redhat.com/show_bug.cgi?id=1883749) about Prometheus containers requiring close to 10 minutes to flush the head to WAL on shutdown (not sure exactly why it took so long, maybe the node was overloaded). In one occasion, the container was presumably SIGKILLed and the data on disk was corrupted after that.

brancz:
@simonpasquier I think we should focus on the Prometheus improvements that are happening with checkpoints and snapshots on shutdown, so that won't be necessary.

simonpasquier:
#3673 is another report about the hard-coded termination grace period. This looks similar to #3278: some cloud providers enforce a maximum value for `terminationGracePeriodSeconds` preventing the roll-out of the prometheus statefulset.

diyavicky:
@simonpasquier So at this point there is no possible way to provide acceptable custom value for terminationGracePeriod?
Just trying to understand the current state of this issue.

Thanks!

diyavicky:
@simonpasquier @brancz Could you please advise if there is a possible way to mention custom terminationGracePeriod by any chance?

Please advise!

Thanks!

diyavicky:
@simonpasquier @brancz Really appreciate any update on the above query.

[...]

PR #3434, opened 2020-08-22
add Termination grace period seconds

lurenjia528:
add terminationGracePeriodSeconds to prometheusSpec and alertmanagerSpec

releate to 
https://github.com/prometheus-operator/prometheus-operator/issues/3433

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -366,8 +366,13 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 			MountPath: secretsDir + s,
 		})
 	}
-
-	terminationGracePeriod := int64(0)
+	
+	var terminationGracePeriod int64
+	if nil != a.Spec.TerminationGracePeriodSeconds {
+		terminationGracePeriod = *a.Spec.TerminationGracePeriodSeconds
+	} else {
+		terminationGracePeriod = int64(300)
+	}
 	finalLabels := config.Labels.Merge(podLabels)
 	return &appsv1.StatefulSetSpec{
 		ServiceName: governingServiceName,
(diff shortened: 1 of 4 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -266,7 +266,12 @@ func makeStatefulSetService(p *monitoringv1.Prometheus, config Config) *v1.Servi
 func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapNames []string) (*appsv1.StatefulSetSpec, error) {
 	// Prometheus may take quite long to shut down to checkpoint existing data.
 	// Allow up to 10 minutes for clean termination.
-	terminationGracePeriod := int64(600)
+	var terminationGracePeriod int64
+	if nil != p.Spec.TerminationGracePeriodSeconds {
+		terminationGracePeriod = *p.Spec.TerminationGracePeriodSeconds
+	} else {
+		terminationGracePeriod = int64(300)
```
simonpasquier, line 273: why changing the default?

pkg/alertmanager/statefulset.go
```diff
@@ -366,8 +366,13 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 			MountPath: secretsDir + s,
 		})
 	}
-
-	terminationGracePeriod := int64(0)
```
simonpasquier, line 370: you'd need to rebase your fork because the default value is currently [120s](https://github.com/prometheus-operator/prometheus-operator/blob/000315b9aaa20e228f38e74852741d383bedafe4/pkg/alertmanager/statefulset.go#L495).

pkg/prometheus/statefulset.go
lurenjia528: I modify the v0.23.0,not the master

simonpasquier: v0.23.0 is a very old release that isn't supported anymore. Please rebase on top of the `master` branch.

brancz:
As stated in https://github.com/prometheus-operator/prometheus-operator/issues/3433 I don't believe this is how this should be solved, and we should not enable an anti-pattern like that.

PR #3435, opened 2020-08-24
Added vSphere Storage integration

appdess:
Hi, I´ve added how to create a Storage Class on vSphere and adjust the chart so it can make use of persistent volumes on vSphere using the CSI-Provider.


coderanger:
This is the repo for the operator itself, not the Helm chart. That said, neither the operator nor chart care about what type of volume you use, just set the storage class appropriately.

PR #3436, opened 2020-08-24
pkg/alertmanager: Use lower value for --cluster.reconnect-timeout

hwoarang:
Alertmanager in cluster mode resolves the DNS name of each peer and
caches its IP address which uses on regular intervals to 'refresh'
the connection.

In high-dynamic environment like kubernetes, it's possible that
alertmanager pods come and go on frequent intervals. The default timeout
value of 6h is not suitable in that case as alertmanager will keep
trying to reconnect to a non-existing pod over and over until it gives
up and remove that peer from the member list. During this period of
time, the cluster is reported to be in a degraded state due to the
missing member.

As such, it's best to use a lower value which will allow the
alertmanager to remove the pod from the list of peers soon
after it disappears.

Related: https://github.com/prometheus/alertmanager/issues/2250

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -384,6 +384,12 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 				// below Alertmanager v0.15.0 high availability flags are prefixed with 'mesh' instead of 'cluster'
 				amArgs[i] = strings.Replace(amArgs[i], "--cluster.", "--mesh.", 1)
 			}
+		} else {
+			// reconnect-timeout was added in 0.15 (https://github.com/prometheus/alertmanager/pull/1384)
+			// Override default 6h value to allow AlertManager cluster to
+			// quickly remove a cluster member after its pod restarted or during a
+			// regular rolling update.
+			amArgs = append(amArgs, "--cluster.reconnect-timeout=5m")
 		}
 		if version.Minor < 13 {
 			for i := range amArgs {


pkg/alertmanager/statefulset.go
```diff
@@ -239,6 +239,7 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 		amArgs = append(amArgs, "--cluster.listen-address=")
 	} else {
 		amArgs = append(amArgs, "--cluster.listen-address=[$(POD_IP)]:9094")
+		amArgs = append(amArgs, "--cluster.reconnect-timeout=5m")
```
simonpasquier, line 242: we still support Alertmanager < v0.15 on a best-effort basis so it'd be better if this is added around L375.

hwoarang: @simonpasquier thank you for the review. I have updated the PR

simonpasquier:
> The default timeout value of 6h is not suitable in that case as alertmanager will keep
trying to reconnect to a non-existing pod over and over until it gives
up and goes through another DNS resolution process.

To be clear Alertmanager will continuously resolve the peer addresses irrespective of the `--cluster.reconnect-timeout` value. The flag defines how long Alertmanager will try to reconnect to an IP address it has been connected before. I do agree that in dynamic environments such as Kubernetes, it makes sense to lower the default value.

hwoarang:
@simonpasquier Thank you for the clarification. I can update the commit message if you want

simonpasquier:
Yes please, update the comment too. Thanks!

hwoarang:
> Yes please, update the comment too. Thanks!

@simonpasquier I have updated both. I hope it's better now

simonpasquier:
The CI fails because of Go formatting, please run `make go-fmt` and commit the changes.

pkg/alertmanager/statefulset.go
```diff
@@ -384,6 +384,12 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 				// below Alertmanager v0.15.0 high availability flags are prefixed with 'mesh' instead of 'cluster'
 				amArgs[i] = strings.Replace(amArgs[i], "--cluster.", "--mesh.", 1)
 			}
+		} else {
+			// reconnect-timeout was added in 0.15 (https://github.com/prometheus/alertmanager/pull/1384)
+			// Override default 6h value to allow AlertManager cluster to
+            // quickly remove a cluster member after its pod restarted or during a
+            // regular rolling update
```
simonpasquier, line 391: missing dot at the end

hwoarang: done

simonpasquier (approved):
LGTM

s-urbaniak:
LGTM, but potentially a follow-up question: Does it make sense to configure this setting? My impression is that it shouldn't as it is an implementation detail of how we reconcile alertmanager instances. We could add that rationale though at least as a comment, wdyt?

hwoarang:
> LGTM, but potentially a follow-up question: Does it make sense to configure this setting? My impression is that it shouldn't as it is an implementation detail of how we reconcile alertmanager instances. We could add that rationale though at least as a comment, wdyt?

I agree that configuring this option shouldn't be necessary. I think the existing comment explains the reason to change the default value. Do you think it needs more details?

s-urbaniak:
I think we're good, merging, thank you!

PR #3420, opened 2020-08-19
WIP: remove multilistwatcher and denylistfilter

s-urbaniak:
This is an initial attempt to remove the multilistwatcher and denylisting filter.
It is meant to assert if e2e tests pass.

Note: there is going to be a follow-up for master, 0.40 is not planned to be patched.

Fixes #3339, #3410

Issue #3410, opened 2020-08-13
--deny-namespaces skips *v1.Status events


s-urbaniak:
**What happened?**

During an unrelated experiment https://github.com/prometheus-operator/prometheus-operator/pull/3409 it occurred to me that the existing deny listing watch logic which filters namespaced objects only can miss Watch events, i.e. `*v1.Status` event object types.

https://github.com/prometheus-operator/prometheus-operator/blob/5c3520960e179458c0c98dae94d4cae65f71aa7f/pkg/listwatch/namespace_denylist.go#L148-L154

I am not fully aware of the functionality ramifications, but currently, if the user specifies a denylist of namespaces, potentially status Watch events will never be dispatched.

**Did you expect to see something different?**

I expect all events related to denylisted namespaces to be dispatched.

**How to reproduce it (as minimally and precisely as possible)**:

Set up prometheus-operator with `--deny-namespaces` and watch for log entries like:

```
msg="unexpected object type in event (*v1.Status): &Status{ListMeta:ListMeta{SelfLink:,ResourceVersion:,Continue:,RemainingItemCount:nil,},Status:Failure,Message:too old resource version: 30297 (42339),Reason:Expired,Details:nil,Code:410,}"
```

**Environment**

* Prometheus Operator version: 0.40.0

* Kubernetes version information: 1.19 (but proably earlier ones too)

* Kubernetes cluster kind: N/A

* Manifests: N/A

* Prometheus Operator Logs: See above

**Anything else we need to know?**:

s-urbaniak:
suggestion: let's refactor the denylist logic as a follow-up to the planned refactoring of multilistwatcher.

We learned that providing a custom `ListWatcher` interface always causes unexpected side-effects with the underlying client-go machinery as experienced in [1] and [2].

[1] https://github.com/prometheus-operator/prometheus-operator/issues/3218
[2] https://github.com/prometheus-operator/prometheus-operator/issues/3218

s-urbaniak:
Continued investigation showed that we can leverage `FieldSelectors` (see https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) in the style of `namespace!=foo,namespace!=bar,...` to implement the same functionality as the existing denylisting filter.

It also benefits from increased performance as the resultset is limited in the first place, without the need for in-place filtering.

s-urbaniak:
This is fixed in #3440, closing.

PR #3440, opened 2020-08-25
remove multilistwatcher and denylistfilter

### LINKED ISSUE #3339: Remove the need for custom multi namespace listwatcher
This has caused a few problems in the past, and made it difficult to debug due to its nature. 

TODO: write up plan on how to remove this, possibly using https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#labelselectorrequirement-v1-meta

See background https://github.com/coreos/prometheus-operator/issues/3218

s-urbaniak:
This is removing multilistwatch and denylist listerwatcher in favor of native informers.

It is still WIP and a follow-up of #3420 which addresses the 0.40 branch only.

TODO:
- [x] extend this to the other sub-operator control loops (thanos).
- [x] think about reducing `for ...` loops not to iterate over all informers. As discussed OOB with @sttts we want to implement a minimal `Lister` and a subset of `Informer` methods to implement multi-namespace concerns.
- [x] add tests for newly introduced packages
- [x] add an explicit permission e2e test asserting the invariant expressed in https://github.com/prometheus-operator/prometheus-operator/pull/3409
- [x] do a benchmark comparing a single informer using multilistwatcher vs. an array of informers used here
- [x] Since the underlying ListerWatchers are now hidden behind factories, think how we can hook `instrumentedListerWatcher` (if at all necessary), handled in https://github.com/prometheus-operator/prometheus-operator/pull/3421.

Fixes #3339, #3410


s-urbaniak:
An initial benchmark shows the following results. Based on an OpenShift deployment resembling a heavily used `--namespaces` based deployment and the following indicators:

- Number of namespaces set in `--namespaces`: 39
- Number of reconciled servicemonitors: 51
- Number of reconciled prometheusrules: 22
- Number of reconciled prometheuses: 1
- Number of reconciled alertmanagers: 1

We get the following cpu/memory figures:

1. With the current `multilistwatcher`:
image

2. With the list of informers used in this PR:
image

Verdict:
One can see that we have an increase of ~10MB for 39 namespaces. This is expected as we maintain a list of informers now, each one having its own cache instance.

CPU wise no significant difference was observed. The initial spike is most likely due to scrape interval irregularity. Both versions converge against ~1 millicore CPU usage with spikes of ~28 millicores, most likely when relists happen.

s-urbaniak:
ping @squat for initial reviews. as discussed, do you think this is heading in the right direction? if so, i'd add more tests, documentation.

simonpasquier (commented):
This looks great! :tada: 

> Since the underlying ListerWatchers are now hidden behind factories, think how we can hook instrumentedListerWatcher (if at all necessary).

It might still be useful to have some sort of client-side API metrics but maybe we should rather instrument the REST client with [`k8s.io/client-go/tools/metrics`](https://pkg.go.dev/k8s.io/client-go@v0.19.0/tools/metrics)?

Regarding `storeCollector`, I think we can remove it. It will be better handled by #3421 IMO.

pkg/alertmanager/operator.go
```diff
@@ -117,56 +115,57 @@ func New(c prometheusoperator.Config, logger log.Logger, r prometheus.Registerer
...
 	)
-	o.metrics.MustRegister(NewAlertmanagerCollector(o.alrtInf.GetStore()))
-	o.metrics.MustRegister(operator.NewStoreCollector("alertmanager", o.alrtInf.GetStore()))
-
-	o.ssetInf = cache.NewSharedIndexInformer(
-		o.metrics.NewInstrumentedListerWatcher(
-			listwatch.MultiNamespaceListerWatcher(o.logger, o.config.Namespaces.AlertmanagerAllowList, o.config.Namespaces.DenyList, func(namespace string) cache.ListerWatcher {
-				return cache.NewListWatchFromClient(o.kclient.AppsV1().RESTClient(), "statefulsets", namespace, fields.Everything())
-			}),
+
```
squat, line 128: tiny nit: normally we cuddle error checks with the source of the error

pkg/alertmanager/operator.go
```diff
@@ -117,56 +115,57 @@ func New(c prometheusoperator.Config, logger log.Logger, r prometheus.Registerer
...
+	}
+
+	o.ssetInfs, err = informers.NewInformersForResource(
+		informers.NewKubeInformerFactories(
+			o.config.Namespaces.AlertmanagerAllowList, o.config.Namespaces.DenyList,
+			o.kclient, resyncPeriod, nil,
 		),
-		&appsv1.StatefulSet{}, resyncPeriod, cache.Indexers{},
+		appsv1.SchemeGroupVersion.WithResource("statefulsets"),
 	)
```
squat, line 140: same

pkg/alertmanager/operator.go
```diff
@@ -117,56 +115,57 @@ func New(c prometheusoperator.Config, logger log.Logger, r prometheus.Registerer
...
-					},
-					WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-						options.LabelSelector = o.config.AlertManagerSelector
-						return o.mclient.MonitoringV1().Alertmanagers(namespace).Watch(context.TODO(), options)
-					},
-				}
-			}),
+	o.alrtInfs, err = informers.NewInformersForResource(
+		informers.NewMonitoringInformerFactories(
+			o.config.Namespaces.AlertmanagerAllowList, o.config.Namespaces.DenyList,
```
squat, line 120: nit: let's put every argument either on the same line or every argument on it's own line

pkg/alertmanager/operator.go
```diff
@@ -426,11 +428,14 @@ func (c *Operator) sync(key string) error {
...
-	if err != nil {
+	obj, err := c.ssetInfs.Get(alertmanagerKeyToStatefulSetKey(key))
+
+	if err != nil && !apierrors.IsNotFound(err) {
 		return errors.Wrap(err, "retrieving statefulset failed")
 	}
 
+	exists := !apierrors.IsNotFound(err)
+
 	if !exists {
```
squat, line 439: here we are negating a negation. It would be much clearer if we simply wrote:
```go
if apierrors.IsNotFound(err) {
}
```

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+
+package informers
+
+import (
+	"strings"
+	"time"
+
+	"github.com/prometheus-operator/prometheus-operator/pkg/listwatch"
+
+	"github.com/pkg/errors"
```
squat, line 23: nit, we can cuddle this with the packages below, no?

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+)
+
+type GenericInformer interface {
+	Informer() cache.SharedIndexInformer
+	Lister() cache.GenericLister
+}
+
+// InformerFactoriesForNamespaces is a simple way to combine several shared informers into a single struct with unified listing power
```
squat, line 40: Let's make this comment a full sentence and add a `.` at the end

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+)
+
+type GenericInformer interface {
```
squat, line 35: `GenericInformer` is a bit misleading IMO, as the canonical definition of what comprises and informer is in the client-go codebase. Consider maybe `InformLister` following the example from the golang io package of composed informers of e.g. `ReadWriter`.

s-urbaniak: nice, that's a much better and more idiomatic naming :+1:

pkg/alertmanager/operator.go
s-urbaniak: agreed, and i think i did this only due to cherry-pick friendliness, but the inline version should be as friendly.

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+	Lister() cache.GenericLister
+}
+
+// InformerFactoriesForNamespaces is a simple way to combine several shared informers into a single struct with unified listing power
+type InformerFactoriesForNamespaces interface {
+	ForResource(namespace string, resource schema.GroupVersionResource) (GenericInformer, error)
+	Namespaces() sets.String
+}
+
+type InformersForResource struct {
```
squat, line 46: There is some stutter here when using this type from another package: `informers.InformersForResource`

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+type GenericInformer interface {
+	Informer() cache.SharedIndexInformer
+	Lister() cache.GenericLister
+}
+
+// InformerFactoriesForNamespaces is a simple way to combine several shared informers into a single struct with unified listing power
+type InformerFactoriesForNamespaces interface {
```
squat, line 41: Ditto from below

pkg/informers/informers.go
squat: In general I think it is clearer to leave out the qualification of `simple` from these comments. It can be confusing to readers that may wonder "What is the not simple way?" "Is the difficult way actually better?"

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+	return nil
+}
+
+func (w *InformersForResource) Get(name string) (runtime.Object, error) {
+	var err error
+
+	for _, inf := range w.informers {
+		var ret runtime.Object
+		ret, err = inf.Lister().Get(name)
+
```
squat, line 130: let's cuddle the error setting and the error check

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+		if apierrors.IsNotFound(err) {
```
squat, line 131: I'm not sure this is correct. What if a name doesn't exist in ANY informer? Then we will continue and eventually return an error that says that the error was not found in one _particular_ namespace, i.e. the last informer that was checked, which may not be the namespace the caller actually cares about.

pkg/informers/kube.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+)
+
+func NewKubeInformerFactories(
```
squat, line 27: I think it's a little bit confusing that this has `Kube` in the name. I kind of think that all of this code is `kube`-related, since all of it is dealing with informers. Why not simply `NewInformerFactories`?

squat: hmm I see that this is because of the fact that this func takes a kubernetes.Interface as opposed to the monitoring client ://// I wish we could abstract this better to be less repetitive. If only we had generics :))

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	for ns, _ := range allowedNamespaces {
+		namespaces = append(namespaces, ns)
+	}
+
+	return tweaks, namespaces
+}
+
+func denyListTweak(options *metav1.ListOptions, namespaces map[string]struct{}) {
```
squat, line 168: Instead of copying this in two places, can we export it once and reuse it?

pkg/prometheus/operator.go
```diff
@@ -249,130 +249,112 @@ func New(conf Config, logger log.Logger, r prometheus.Registerer) (*Operator, er
...
-						return c.kclient.CoreV1().ConfigMaps(namespace).List(context.TODO(), options)
-					},
-					WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-						options.LabelSelector = labelPrometheusName
-						return c.kclient.CoreV1().ConfigMaps(namespace).Watch(context.TODO(), options)
-					},
-				}
-			}),
+
+	if err != nil {
```
squat, line 313: Let's be consistent with the cuddling of error checks

pkg/prometheus/operator.go
```diff
@@ -1000,7 +1000,9 @@ func (c *Operator) enqueueForNamespace(store cache.Store, nsName string) {
 	}
 	ns := nsObject.(*v1.Namespace)
 
-	err = cache.ListAll(c.promInf.GetStore(), labels.Everything(), func(obj interface{}) {
+	objs, err := c.promInfs.List(labels.Everything())
```
squat, line 1003: we should check the return error before looping over the returned values

pkg/thanos/operator.go
```diff
@@ -569,16 +577,17 @@ func (o *Operator) processNextWorkItem() bool {
...
 
 func (o *Operator) sync(key string) error {
-	obj, exists, err := o.thanosRulerInf.GetIndexer().GetByKey(key)
-	if err != nil {
-		return err
-	}
-	if !exists {
+	trobj, err := o.thanosRulerInfs.Get(key)
+
+	if apierrors.IsNotFound(err) {
```
squat, line 582: same, let's be consistent here

squat (dismissed):
Overall this looks like a good direction. There are a few potential edge cases and some nits that could use your attention

pkg/informers/informers.go
s-urbaniak: `Get` is invoked with `[ns]/[name]`. client-go does not inject the namespace of the underlying informer but keeps the requested namespace. I coded a quick example, where `ns1`, `ns2`, `ns3` all contain a servicemonitor called `prometheus-example-monitor`:

When we initialize the informers with `ns1` and `ns2` only:
```
	infs, err := informers.NewInformersForResource(
		informers.NewMonitoringInformerFactories(
			map[string]struct{}{
				"ns1": {},
				"ns2": {},
			}, nil,
			mclient, 5*time.Minute, nil,
		),
		monitoringv1.SchemeGroupVersion.WithResource(monitoringv1.ServiceMonitorName),
	)
```

And look for `ns3/prometheus-example-monitor` 
```
	ret, err := infs.Get("ns3/prometheus-example-monitor")
```
the following error is being returned by client-go:
```
    informers_test.go:63: servicemonitors.monitoring.coreos.com "ns3/prometheus-example-monitor" not found
```

You are right that the error is returned by the `"ns2"` informer, but the error message is the same as for the `"ns1"` informer.

s-urbaniak: However, we could construct our own explicit apierror to avoid confusion, wdyt?

pkg/informers/informers.go
s-urbaniak: ha, great catch! let's call it `informers.ForResource`!

pkg/informers/kube.go
s-urbaniak: yes this is unfortunate indeed :-/

I tried to factor as much as possible into the `ForResource` informer, but some duplication stayed. Generics would've helped (or code generation).

pkg/informers/informers.go
s-urbaniak: It's not a copy, one uses `metadata.namespace!=...` the other uses `metadata.name!=..`. I will rename them to avoid confusion.

pkg/prometheus/operator.go
s-urbaniak: agreed, cherry-pick friendliness as an excuse here, but i think git will be smart enough. let me pull the err check above.

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+	if listwatch.IsAllNamespaces(allowedNamespaces) {
+		namespaces = append(namespaces, v1.NamespaceAll)
+
+		return func(options *v1.ListOptions) {
+			tweaks(options)
+			denyListTweak(options, deniedNamespaces)
+		}, namespaces
+	}
+
+	for ns, _ := range allowedNamespaces {
```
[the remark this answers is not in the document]
s-urbaniak: fixed

pkg/informers/informers.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func denyListTweak(options *metav1.ListOptions, namespaces map[string]struct{}) {
+	if len(namespaces) == 0 {
+		return
+	}
+
+	var denied []string
+
+	for ns, _ := range namespaces {
```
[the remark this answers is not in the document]
s-urbaniak: fixed

pkg/listwatch/listwatch.go
```diff
@@ -284,3 +291,17 @@ func IdenticalNamespaces(a, b map[string]struct{}) bool {
...
 }
+
+func denyListTweak(options *metav1.ListOptions, namespaces map[string]struct{}) {
+	if len(namespaces) == 0 {
+		return
+	}
+
+	var denied []string
+
+	for ns, _ := range namespaces {
```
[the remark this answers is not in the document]
s-urbaniak: fixed

pkg/alertmanager/operator.go
s-urbaniak: fixed

pkg/alertmanager/operator.go
s-urbaniak: fixed

pkg/alertmanager/operator.go
s-urbaniak: fixed

pkg/alertmanager/operator.go
s-urbaniak: fixed

pkg/informers/informers.go
s-urbaniak: fixed

pkg/informers/informers.go
s-urbaniak: fixed

pkg/informers/informers.go
s-urbaniak: fixed

pkg/informers/informers.go
s-urbaniak: fixed

pkg/informers/informers.go
s-urbaniak: fixed

pkg/prometheus/operator.go
s-urbaniak: fixed

pkg/informers/informers.go
s-urbaniak: fixed

pkg/informers/informers.go
s-urbaniak: fixed

pkg/prometheus/operator.go
s-urbaniak: fixed

pkg/thanos/operator.go
s-urbaniak: fixed

s-urbaniak:
All green, time to add some documentation and tests.

s-urbaniak:
@squat @simonpasquier removing the WIP here for a final pass.

pkg/informers/informers.go
```diff
@@ -0,0 +1,204 @@
...
+	Lister() cache.GenericLister
+}
+
+// FactoriesForNamespaces is a way to combine several shared informers into a single struct with unified listing power.
+type FactoriesForNamespaces interface {
+	ForResource(namespace string, resource schema.GroupVersionResource) (InformLister, error)
+	Namespaces() sets.String
+}
+
+// ForResource contains a list informers for a concrete resource type,
```
simonpasquier, line 46:
```suggestion
// ForResource contains a slice of InformLister for a concrete resource type,
```

pkg/informers/informers.go
```diff
@@ -0,0 +1,204 @@
...
+// It takes a namespace aware informer factory, wrapped in a FactoriesForNamespaces interface
+// that is able to instantiate an informer for a given namespace.
+func NewInformersForResource(ifs FactoriesForNamespaces, resource schema.GroupVersionResource) (*ForResource, error) {
+	namespaces := ifs.Namespaces().List()
+	var informers []InformLister
+
+	for _, ns := range namespaces {
+		informer, err := ifs.ForResource(ns, resource)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error getting informer for resource %v", resource)
```
simonpasquier, line 65:
```suggestion
			return nil, errors.Wrapf(err, "error getting informer in namespace %q for resource %v", ns, resource)
```

pkg/informers/informers.go
```diff
@@ -0,0 +1,204 @@
...
+			return nil, err
+		}
+
+		return ret, nil
+	}
+
+	return nil, err
+}
+
+// newInformerOptions constructs a list option tweak function and a list of namespaces
```
simonpasquier, line 156:
```suggestion
// newInformerOptions returns a list option tweak function and a list of namespaces
```

pkg/informers/informers.go
```diff
@@ -0,0 +1,204 @@
...
+
+	var namespaces []string
+
+	if listwatch.IsAllNamespaces(allowedNamespaces) {
+		namespaces = append(namespaces, v1.NamespaceAll)
+
+		return func(options *v1.ListOptions) {
+			tweaks(options)
+			denyNamespacesTweak(options, deniedNamespaces)
+		}, namespaces
```
simonpasquier, line 176:
```suggestion
		return func(options *v1.ListOptions) {
			tweaks(options)
			denyNamespacesTweak(options, deniedNamespaces)
		}, []string{v1.NamespaceAll}
```

pkg/informers/informers.go
```diff
@@ -0,0 +1,204 @@
...
+		return ret, nil
+	}
+
+	return nil, err
+}
+
+// newInformerOptions constructs a list option tweak function and a list of namespaces
+// based on the given allowed and denied namespaces.
+//
+// If allowedNamespaces contains one entry containing k8s.io/apimachinery/pkg/apis/meta/v1.NamespaceAll
```
simonpasquier, line 159:
```suggestion
// If allowedNamespaces contains one only entry equal to k8s.io/apimachinery/pkg/apis/meta/v1.NamespaceAll
```

pkg/informers/informers.go
```diff
@@ -0,0 +1,204 @@
...
+
+	for ns := range allowedNamespaces {
+		namespaces = append(namespaces, ns)
+	}
+
+	return tweaks, namespaces
+}
+
+// denyNamespacesTweak is a simple method modifying the given list options
+// by adding a field selector not matching the given namespaces.
```
simonpasquier, line 187:
```suggestion
// denyNamespacesTweak modifies the given list options
// by adding a field selector not matching the given namespaces.
```

pkg/informers/informers_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+			expectedNamespaces: []string{
+				v1.NamespaceAll,
+			},
+			expectedOptions: v1.ListOptions{
+				FieldSelector: "metadata.namespace!=denied1,metadata.namespace!=denied2",
+			},
+		},
+		{
+			name: "denied namespaces with tweak",
```
simonpasquier, line 200: I don't see that the `tweaks` field is set.

pkg/informers/informers_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+			expectedNamespaces: []string{
+				v1.NamespaceAll,
+			},
+			expectedOptions: v1.ListOptions{
+				FieldSelector: "metadata.namespace!=denied1,metadata.namespace!=denied2",
+			},
+		},
+		{
+			name: "all allowed namespaces no denied namespaces",
```
simonpasquier, line 217: `name` doesn't seem to match with the test inputs.

pkg/listwatch/listwatch.go
```diff
@@ -66,15 +62,22 @@ func NewUnprivilegedNamespaceListWatchFromClient(l log.Logger, c cache.Getter, a
...
-			cache.NewFilteredListWatchFromClient(c, "namespaces", metav1.NamespaceAll, optionsModifier),
-		)
+		tweak := func(options *metav1.ListOptions) {
+			if optionsModifier != nil {
+				optionsModifier(options)
+			}
+
+			fieldSelector := options.FieldSelector
+			denyNamesTweak(options, deniedNamespaces)
+			options.FieldSelector = strings.Join([]string{options.FieldSelector, fieldSelector}, ",")
```
simonpasquier, line 75: couldn't this be done in `denyNamesTweak()`?

test/framework/framework.go
```diff
@@ -187,8 +187,20 @@ func (f *Framework) CreatePrometheusOperator(ns, opImage string, namespaceAllowl
 		return nil, errors.Wrap(err, "failed to update prometheus cluster role")
 	}
 
-	if _, err := CreateClusterRoleBinding(f.KubeClient, ns, "../../example/rbac/prometheus-operator/prometheus-operator-cluster-role-binding.yaml"); err != nil && !apierrors.IsAlreadyExists(err) {
-		return nil, errors.Wrap(err, "failed to create prometheus cluster role binding")
+	if createClusterRoleBindings {
```
simonpasquier, line 190: neat!

simonpasquier (approved):
LGTM. This is awesome! I've got only a few minor remarks. Feel free to ignore the cosmetic ones :)

pkg/informers/informers_test.go
s-urbaniak: fixed

pkg/informers/informers_test.go
s-urbaniak: fixed

pkg/listwatch/listwatch.go
s-urbaniak: indeed, I unified the methods

test/framework/framework.go
s-urbaniak: the amount of parameters for `CreatePrometheusOperator` starts worrying me though :thinking:

s-urbaniak:
@squat PTAL, looking good so far? :-)

dgrisonnet:
> Since the underlying ListerWatchers are now hidden behind factories, think how we can hook instrumentedListerWatcher (if at all necessary).

I looked into that when I was writing kube-events-exporter. The only way I found to hook the instrumentedListerWatcher was to pre-register each resource informer with [InformerFor](https://godoc.org/k8s.io/client-go/informers/internalinterfaces#SharedInformerFactory) and hook the lister watcher in the NewInformerFunc.
From what I understood this means that you need to rewrite every NewFiltered**X**Informer functions that you want to instrument and register all the custom constructors to your factory.

s-urbaniak:
I think I'll tackle the metrics part independent of this PR with simon's suggestion.

simonpasquier (approved):
LGTM

s-urbaniak:
Rebased to pull in all other changes, also pinging other maintainers @prometheus-operator/kube-prometheus-reviewers for a final pass.

s-urbaniak:
Whopsie, the rebase was not correct, fixing.

test/e2e/denylist_test.go
```diff
@@ -20,8 +20,8 @@ import (
 
 	"github.com/gogo/protobuf/proto"
 
-	testFramework "github.com/prometheus-operator/prometheus-operator/test/framework"
 	"github.com/pkg/errors"
+	testFramework "github.com/prometheus-operator/prometheus-operator/test/framework"
```
lilic, line 24: In this test we seem to only test ServiceMonitor and Prometheus custom resources, would be good to also add ThanosRuler test, what do you think? I believe the test was written before the addition of that CR, so would be nice since we are changing it to test that resource as well. Does that make sense?

lilic (commented):
Looking good!

pkg/informers/monitoring.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	informers "github.com/prometheus-operator/prometheus-operator/pkg/client/informers/externalversions"
+	monitoring "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func NewMonitoringInformerFactories(
```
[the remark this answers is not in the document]
s-urbaniak: definitely! doh, i overlooked.

test/framework/ressources/prometheus-operator-role-binding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
s-urbaniak: Currently we test prometheus-operator with having a cluster role binding against reconciling its CRDs.

This adds an e2e test which asserts whether reconciliation still works if a) prometheus-operator is started with `--namespace` and having permissions for those namespaces only.

As of today we have no such test, this introduces it.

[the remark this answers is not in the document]
s-urbaniak: To be more precise: it changes the existing allow list e2e test to use _role_ bindings only. The idea is to test against regressions in the future to prevent necessitating wider permissions than necessary.

test/e2e/denylist_test.go
s-urbaniak: definitely, let me add that e2e test.

test/framework/ressources/prometheus-operator-role-binding.yaml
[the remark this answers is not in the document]
lilic: Great! Thanks for explaining!

test/e2e/denylist_test.go
```diff
@@ -38,7 +38,7 @@ func testDenyPrometheus(t *testing.T) {
 
 	ctx.SetupPrometheusRBAC(t, operatorNamespace, framework.KubeClient)
 
-	_, err := framework.CreatePrometheusOperator(operatorNamespace, *opImage, nil, deniedNamespaces, nil, nil, false)
```
lilic, line 41: While testing I noticed that deleting a ServiceMonitor does not actually update the targets correctly. I would suggest we add a test for this, seems like a regression as before it reconciled that correctly.

This might be the case for other resources, did not test, but might we work looking at other CRs.

s-urbaniak: Interesting, let me add a test case for this.

s-urbaniak: Works for me when testing manually, but adding a test is definitely a good idea. cc'ing @simonpasquier as i believe lili saw the latency by configmap-reloader which is currently being refactored.

s-urbaniak: Added tests.

s-urbaniak:
The e2e test timed out, restarted.

pkg/informers/informers.go
squat: Ack; let's keep this in mind but this is not blocking. At the heart of this, because we are doing a lot of magic, we may need to build our own errors too in order to avoid a leaky abstraction. E.g. say that next week client-go changes the error message to be "resource x/y could not be found in namespace z"; this would be confusing to users asking about resource "y" in namespace "x".

pkg/informers/informers.go
```diff
@@ -0,0 +1,179 @@
...
+
+// FactoriesForNamespaces is a way to combine several shared informers into a single struct with unified listing power.
+type FactoriesForNamespaces interface {
+	ForResource(namespace string, resource schema.GroupVersionResource) (InformLister, error)
+	Namespaces() sets.String
+}
+
+// ForResource contains a slice of InformLister for a concrete resource type,
+// one per namespace.
+type ForResource struct {
```
squat, line 45: Why are we creating a struct with an field that is a slice of InformListers rather than directly aliasing the `[]InformLister` type, e.g.

```golang
type ForResource []InformLister
```

If we were adding and removing elements from the slice on-the-fly, then I agree we would want to use a struct in order to keep a consistent reference for the user, but it looks like this only occurs at creation-time.

squat (commented):
Overall looks great to me. One outstanding issue, but it's not blocking

pkg/informers/informers.go
s-urbaniak: I have been thinking about this too but since this is not an algorithmic data structure but rather an infrastructural one I left it as a struct. In fact, as a follow-up I want to leave some logging debug traces where we want a logger in the struct too.

s-urbaniak:
Merging as we seem have to reached consensus. Thank you everybody for the great reviews! :tada:

Issue #2927, opened 2019-12-27
Add AlertRoute for merging alertmanager configs for different teams


sev3ryn:
**What is missing?**
I believe, ability for each team/project to configure their alertmanager routes, grouping options and own receivers is a missing part in Prometheus operator. 

**Why do we need it?**
K8s CRD( let's name it AlertRoute) together with ServiceMonitor and PrometheusRule will give teams full control over their monitoring and alerting.

Is it something you will consider merging if I contribute a PR for it?

brancz:
I think this is a valid use case and one that we would want. Merging the Alertmanager routing tree from different configs is tricky but I think it can be done. If you want to write up a design for it that would be fantastic! Then we can discuss the design and ultimately we/you can get to implementing this! :)

spencergilbert:
I know that Rancher does some level of merging of Alertmanager configs for it's monitoring solution, it could be a useful reference

brancz:
Do you have a link to that?

spencergilbert:
I think most of that logic lives [here](https://github.com/rancher/rancher/blob/master/pkg/controllers/user/alert/config/config.go)

brancz:
I can't find config merging there (doesn't mean it doesn't exist but I couldn't find it). What you linked looks like a copy of the `config.go` file from the Alertmanager project.

spencergilbert:
Sorry, let me keep digging - I haven't looked at the internals much, just the user end

sev3ryn:
I'm new to writing k8s CRD - please correct me if I'm saying something that is absolutely wrong:

the way I see design of it - we need 2 new k8s CRDs:
1. AlertReceiver - same config as any of receivers https://prometheus.io/docs/alerting/configuration/#receiver but instead of fields with secrets you have to specify the k8s secret. Also there will be additional field templates: in which you specify list of configmaps to use as templates
2. AlertRoute - same as https://prometheus.io/docs/alerting/configuration/#route . All created AlertRoutes will be treated as 1st level routes (after default one) and there will be option to specify subroutes in it.

all other configs - global settings,  inhibit_rules will stay as now in Alertmanager CRD - just  to store it in the end in configmap(same as prometheus config) and not secret. Default route should be created in AlertRoute CRD but with some special name.

brancz:
[...]

Issue #2766, opened 2019-09-18
Add Alertmanager config options to Alertmanager crd


benjaminhuo:
**What is missing?**
Currently Alertmanager crd spec doesn't have config options like prometheus crd

Just wondering if it's reasonable to add alert manager config to alertmanager crd spec? 

And then regenerate the alertmanager secrets and then call alertmanager's reload interface?

**Why do we need it?**
Currently we've to edit the alertmanager secret manually and then reload

brancz:
The reload should happen automatically. The reason we haven't added the Alertmanager configuration itself to the CRD is because we see similar collaboration potential with Alertmanager route configurations and receivers as there is with ServiceMonitors. As in each namespace should be able to specify these for themselves.

benjaminhuo:
Thanks for the reply :)
Yes, the reload occurs whenever the secret is updated
You mean we should consider abstracting route and receivers as servicemonitor does?

brancz:
Yes exactly. It'll be some time, but we already have some of this work planned.

benjaminhuo:
Great! The route is kind of recursive, it's a bit complex. Break the big alert config into smaller parts is a good idea 👍

midnightconman:
I would like to see this happen and would be willing to help out... would we first want to see schema, before we start working on code?

LiliC:
Yes defining a schema in advance would be good, feel free to write up initial outline so we have something to start with. Thanks!

cc @simonpasquier

brancz:
There are multiple discussions happening around this, I think we need to start deduplicating (https://github.com/coreos/prometheus-operator/issues/2927).

simonpasquier:
Hello! I've started a design document:
https://docs.google.com/document/d/1aVVttvocop8zNezwrNFbHh_HKRW_fhJUZqhmrr8fGK0/edit?usp=sharing

Feel free to comment there...

benjaminhuo:
@simonpasquier That's great! Will take a look at this!

simonpasquier:
Closed by #3451

Issue #1528, opened 2018-06-26
(Cross-)namespaced Alertmanager configuration


brancz:
Similar to how Prometheus can be configured using the ServiceMonitor and PrometheusRule objects, it would be nice if users could configure Alertmanager routes and receivers within their own namespace, for their respective alerts.

This is not as easy as it sounds, as typically when an Alertmanager route matches it does not continue to match any further routes, unless the `continue: true` field is set. This means that if we were to give people the control over these they could be either misconfigured and other teams don't receive alerts, or worse in a multi tenancy environment it could cause alerts that should be sent not to be sent.

The question is whether we should just enable users to configure everything like in the upstream configuration or a subset, or an entirely different abstraction. Ideally I'd like to not restrict users in any way, but I currently don't see a way without doing so, without also violating the requirements above.

There has been some discussion around this in #1498.

@mxinden @Quentin-M

mxinden:
I can see two ways here:

1. Inject the `continue: true` into each route configuration.
2. Create one top level route per Kubernetes namespace and have each alert-route-crd be a sub-route to its namespaced route.

nukepuppy:
still a needed feature honestly

s-urbaniak:
Just as a heads-up that this feature is probably going to be bumped in priority.
/cc @simonpasquier @pgier

simonpasquier:
Hello! I've started a design document:
https://docs.google.com/document/d/1aVVttvocop8zNezwrNFbHh_HKRW_fhJUZqhmrr8fGK0/edit?usp=sharing

Feel free to comment there...

brancz:
Design is done, implementation still outstanding.

simonpasquier:
Closed by #3451

PR #3451, opened 2020-08-27
Add AlertmanagerConfig CRD

### LINKED ISSUE #3261: [Proposal] AlertmanagerRule CRD
**What is missing?**
Alertmanager is only configurable in one way, normally via deployment and creation of its secret. This means that the best alternative is to inject additional changes via jobs running in the cluster reading configuration files or firing up alert managers to support subsections of users as effective as possible. Neither of these solutions are finite in their approach or scale well.

**Why do we need it?**
This limitation of configuration means that operators must include all possibilities for routes and receivers is locked within it and hidden from users within the cluster. Exposing just part of the alertmanager config would vastly improve the usability of this tool and reduce the operational cost for adding new teams or alert pathways. I suggest the following

```
kind: AlertmanagerRule
metadata:
  name: <name>
  namespace: <namespace>
spec:
  routes: # One or more notification routes
    - <routes> # following prometheus standard route definition
  receivers: # A list of notification receivers.
    - name: <name> # The receiver name as defined in the <receiver> spec
      secretRef: <secret name> # kubernetes secret name as found within the namespace which contains one or more receiver config blocks as found in the <receiver> spec (eg <email_config>)
```
* The alertmanager defaults and general config blocks would be configured via the same process as it is now. These would only provide additional routes and receivers  
* The operator would discover AlertmanagerRule resources within the cluster and merge them with the defaults defined within the existing config
* All routes added would be non blocking, with continue forced to true. This can be not great, but there is no better work around to make sure someone does not clobber all down level alerts unless we can use multiple root routes instead (documentation is unclear and ive not tested multiple root routes)

**Environment**

* Prometheus Operator version: All

grdryn:
This builds upon @brancz's work in #3265 and will hopefully eventually implement the core of the proposal at https://docs.google.com/document/d/1aVVttvocop8zNezwrNFbHh_HKRW_fhJUZqhmrr8fGK0/edit?usp=sharing

Closes #3261, #2766, #2927, #1528.


s-urbaniak:
@grdryn Is this ready for final review? The build failing and I think we should add some more e2e tests here too. Let us know if you need further assistance :)

grdryn:
@s-urbaniak no, this is not ready for review yet. There's still a lot to do! :slightly_smiling_face: I think GH added reviewers automatically.

I will indeed be in touch if/when I need help, thanks! :slightly_smiling_face:

pkg/alertmanager/operator.go
```diff
@@ -469,6 +637,336 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
+func (c *Operator) provisionAlertmanagerConfiguration(ctx context.Context, am *monitoringv1.Alertmanager) error {
+	// If no AlertmanagerConfig selectors are configured, the user wants to
+	// manage configuration themselves.
+	if am.Spec.AlertmanagerConfigSelector == nil {
```
grdryn, line 643: TODO: in this case we'll still need to copy the base secret (either that or choose to mount the base secret to the pods).

pkg/alertmanager/operator.go
```diff
@@ -469,6 +637,336 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		level.Debug(c.logger).Log("msg", ", no AlertmanagerConfig selector specified, leaving configuration unmanaged", "alertmanager", am.Name, "namespace", am.Namespace)
+		return nil
+	}
+
+	secretName := defaultConfigSecretName(am.Name)
+	if am.Spec.ConfigSecret != "" {
```
grdryn, line 649: What should we use as a base if this is not set (it's currently optional)?

pkg/alertmanager/operator.go
```diff
@@ -469,6 +637,336 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	err := cache.ListAll(c.nsAlrtInf.GetStore(), selector, func(obj interface{}) {
+		ns = append(ns, obj.(*v1.Namespace).Name)
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list namespaces")
+	}
+	return ns, nil
+}
+
+func (c *Operator) createOrUpdateTLSAssetSecret(am *monitoringv1.Alertmanager) error {
```
grdryn, line 793: I copied these functions for handling the tls assets from the prometheus controller, and might have tweaked slightly to fit.

pkg/alertmanager/operator.go
```diff
@@ -469,6 +637,336 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	}
+
+	for k, v := range amcAssets {
+		assets[k] = v
+	}
+
+	return assets, nil
+}
+
+func (c *Operator) loadTLSAssetsFromTLSConfigList(ns string, tlsConfigs []*monitoringv1.TLSConfig) (map[string]TLSAsset, error) {
```
grdryn, line 886: One thing that I'm not sure about -- the `monitoringv1.TLSConfig` type has additional fields that aren't handled here: `CAFile, CertFile, KeyFile`. I don't think they're read by the prometheus controller either. Is there something we should be doing with those?

pkg/apis/monitoring/v1/alertmanager_config_types.go
```diff
@@ -0,0 +1,117 @@
...
+	Continue       bool      `json:"continue,omitempty"`
+	Routes         []Route   `json:"routes,omitempty"`
+}
+
+type Receiver struct {
+	Name             string            `json:"name"`
+	PagerDutyConfigs []PagerDutyConfig `json:"pagerDutyConfigs,omitempty"`
+}
+
+type PagerDutyConfig struct {
```
grdryn, line 72: TODO: Add Class/Group/Component fields here.

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,499 @@
...
+	out := &pagerdutyConfig{}
+
+	if in.SendResolved != nil {
+		out.VSendResolved = *in.SendResolved
+	}
+
+	if in.RoutingKey != nil {
+		routingKey, err := getCredFromSecret(sClient, *in.RoutingKey, "routing key", crKey.Namespace+"/"+in.RoutingKey.Name, cg.nsSecretCache)
+		if err != nil {
+			level.Warn(cg.logger).Log("msg", "failed to get routing key from secret for pagerduty, ignoring it and continuing",
```
grdryn, line 323: Here, and in some other places in this file, if there's an optional field that specifieds a SecretKeySelector or ConfigMapKeySelector that causes an error when we try to get the value, I've just added a warning log and continued. I'm not sure if that's the best approach or not. Suggestions?

grdryn (commented):
@simonpasquier @s-urbaniak @brancz This currently "works" for the CRD type that exists so far (i.e. only the PagerDuty receiver type is implemented, but others should be easy to add after). Probably some tests need to be fixed up & some more added, I'll look in that direction next.

Could you please take a look when you have some time, to see if what I've done so far is generally in the right direction, or if there are things you'd like changed? It's very much inspired by how the prometheus controller handles ServiceMonitor (etc) CRs, and in many cases blatently plagiarised. At least that should make it somewhat familiar! :slightly_smiling_face:

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,499 @@
...
+	// Enforce continue to be true for main Route in a CR
+	cont := in.Continue
+	if firstLevelRoute {
+		cont = true
+	}
+
+	match := map[string]string{}
+	matchRE := map[string]string{}
+	if firstLevelRoute {
+		match["namespace"] = crKey.Namespace
```
grdryn, line 254: I'm not 100% sure about this -- should it replace _all_ existing matchers with a single namespace matcher like this, or should it just replace existing _namespace_ matchers with this one?

grdryn:
Existing tests appear to work now! :tada: However, the prometheus e2e tests are failing in travis with:

> The job exceeded the maximum time limit for jobs, and has been terminated.

Is that a known issue, or is it caused by the changes here? I don't think these changes should affect those prometheus controller e2e tests that much :man_shrugging:

pkg/alertmanager/operator.go
brancz: If I understand the question correctly, then the current default secret name.

brancz: Ah my bad, I get it now. If this is not set, and the "default secret" doesn't exist, then we handle that gracefully and still generate the configuration and mount the generated file. The change will essentially be that Alertmanager pods will be scheduled even if users don't configure routes (previously the base secret had to be present).

grdryn: @brancz thanks for replying! :) 

Currently (AIUI), if the field is not set, we assume that there's a secret there with the default name. Since we don't read the secret (just mount it as a volume, let k8s handle if it's missing or not).

Now, given that we're making a new Secret, if there's no Secret we generate a skeleton base, so the config can be generated from the AlertmanagerConfig CRs? I'll try that. I think that means that some error conditions here can go away: we don't error if there's no secret, and we could also not error if there _is_ a config secret, but it doesn't contain a base route (we could just add one with a null receiver (and add that receiver if it's missing)? If a base secret does get added (or the missing route gets added to the existing one), then we'll be regenerating the mounted secret anyway, so everything should work.

Does that sound reasonable to you @simonpasquier?

grdryn: I've changed it to something along the lines of what I was thinking about, let me know what you think!

I wonder if a validating webhook for `Alertmanager` resources would be a good idea? i.e. reject submission of a CR if the `configSecret` is set to a missing secret, or to a secret that doesn't have the expected field, or if that field can't be parsed to a valid Alertmanager config. That way people would know immediately that something is wrong, rather than the operator either failing or doing something that they didn't intend.

example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
```diff
@@ -0,0 +1,402 @@
...
+spec:
+  group: monitoring.coreos.com
+  names:
+    kind: AlertmanagerConfig
+    listKind: AlertmanagerConfigList
+    plural: alertmanagerconfigs
+    singular: alertmanagerconfig
+  scope: Namespaced
+  versions:
+  - name: v1
```
lilic, line 19: We don't want to promote this to v1 directly, lets instead change to v1beta1 version. cc @brancz @s-urbaniak we discussed this in a separate issue.

grdryn: @lilic Thanks for the feedback, I'm definitely up for changing it. :+1: 

Has there ever been a case previously in this operator where there was more than one version in the group? The reason I'm asking is that I'm having a little trouble figuring out how to get `client-gen` to generate `pkg/client/versioned/clientset.go` to contain both `v1` and `v1beta1`. It seems like it only accepts one version per group as input: https://github.com/kubernetes/code-generator/blob/master/cmd/client-gen/args/args.go#L78

grdryn: @s-urbaniak You may have suggestions here, as it's related to #2163

lilic: Yes, I was thinking the same when I wrote the comment, honestly I don't think so. I believe because of the history all were promoted at once. @dgrisonnet touched some code generation in the past maybe he has more insight into it. Otherwise we can also build it twice maybe?

grdryn: @lilic I've trying running the `client-gen` command twice, with different `--input` values, if that's what you mean? Unfortunately that leads to just the version specified in the second invocation being used.

My thinking is that either we'd need to generate a different clientset for each version, something like `pkg/client/v1/versioned/` and `pkg/client/v1beta1/versioned/` side-by-side (I don't know what the knock-on effects of that are); or else do some post-processing (I don't know how much work that would be).

Another alternative might be to use a different group for v1beta1 (something like `monitoring/v1` but `prometheusoperator/v1beta1`? :thinking:

dgrisonnet: @grdryn you should be able to do so using client-gen as it is the tool used to generate kubernetes' clientset.

I found an example in the apiserver-sample project. They are generating the [clientset](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/sample-apiserver/pkg/generated/clientset/versioned) for the wardle group v1alpha  and v1beta1 using the following [update-codegen.sh](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/sample-apiserver/hack/update-codegen.sh) script. By digging into that script, you might be able to understand how to pass the input arguments correctly.

However, I think that we would be better off moving to code-generator's [generate-groups.sh](https://github.com/kubernetes/code-generator/blob/master/generate-groups.sh) script that is used in official kubernetes projects instead of maintaining our current custom generation scripts.

grdryn: Thanks @dgrisonnet that will definitely help. I'm starting to think that it actually works the way we need it to, even though it seems to say it doesn't. I'll hopefully get further with this today.

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,514 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Customization of Config type from alertmanager repo:
+// https://github.com/prometheus/alertmanager/blob/master/config/config.go
+//
+// Custom global type to get around obfuscation of secret values when
+// marshalling. See the following issue for details:
+// https://github.com/prometheus/alertmanager/issues/1985
```
simonpasquier, line 40: I'm not sure this is how we should do it. For Prometheus config, we build the content "manually" using `yaml.MapSlice` and depend on any Prometheus internal data structure.

pkg/alertmanager/amcfg.go
simonpasquier: we need to override any existing `namespace` matchers.

pkg/alertmanager/amcfg.go
simonpasquier: We've improved things lately on the service monitor side: the reconciliation loop pulls all data from secrets & configmaps before passing it to the code responsible for generating the configuration.
See https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/prometheus/asset_store.go and https://github.com/prometheus-operator/prometheus-operator/blob/5ab8453b136d58500e8eac9ee5dde149dfdba658/pkg/prometheus/operator.go#L1234

pkg/alertmanager/operator.go
simonpasquier: agree with your approach Gerard. And a validating webhook would be a great addition!

pkg/alertmanager/operator.go
simonpasquier: Check #3445

pkg/alertmanager/operator.go
simonpasquier: `CAFile`, `CertFile` and `KeyFile` aren't things that we want to support for new objects because it isn't flexible enough for users compared to key selectors (see [comment](https://github.com/prometheus-operator/prometheus-operator/pull/3370#discussion_r461665236)).

simonpasquier (commented):
Thanks a lot for doing this! I've made a first limited review. I'll add more later.

pkg/alertmanager/amcfg.go
grdryn: > I'm not sure this is how we should do it.

You don't sound fully convinced by either approach yet, so I'll try to sway you towards this one. :slightly_smiling_face: If I can't make a good case, and there's consensus that using `yaml.MapSlice` is better, then I can try to change it. :+1: 

Is doing everything with `yaml.MapSlice` considered better? Is it more efficient? To me it sounds less nice than unmarshalling the yaml into a type -- less readable and (maybe?) less type safe (`yaml.MapItem` just holding `interface{}` values, etc).

I'm also not sure that this is really depending on an internal alertmanager data structure, since the data structure from the alertmanager project is unfortunately not usable. Instead, this is the prometheus-operator internal type representing the alertmanager config, rather than the alertmanager internal type representing the alertmanager config! :wink:

It's unfortunate that there isn't a type from the Alertmanager project that we can rely on. If there was, and the configuration structure changed in a backwards-incompatible way, then we'd learn at compile-time here, and make any  required changes before releasing a new version of the prometheus-operator that supports that version of Alertmanager.

I manually copied and modified these types from the alertmanager project. That probably could be better...I wonder if it would be good to generate them from the upstream internal types (it's all just a representation of the config yaml that both projects need to understand in the same way)?

Also, these types are _properly_ internal -- we _know_ that nobody will depend on them, unlike in the alertmanager project, where they're _implicitly_ internal, but still exposed, so people could incidentally depend on them. Since they're explicitly an internal implementation detail, we could potentially keep them for now, and if there's a good case for moving to the `yaml.MapSlice` approach in future, we can do that without fear. :slightly_smiling_face:

pkg/alertmanager/operator.go
grdryn: Got it :+1: For now I'll stop using that `monitoringv1.TLSConfig` and make an equivalent to the `SafeTLSConfig` type and maybe once #3370 is merged we can switch to that.

pkg/alertmanager/amcfg.go
brancz: I think having the types is the right way to go, ideally, they will be automatically sync-able in the future. The `yaml.MapSlice` I believe is a major source of allocations and cpu time in the Prometheus config generation, I'd like to one day be able to replace it with a similar mechanism I think (maybe having it typed here is the right way to experiment whether it's the right thing to do).

pkg/alertmanager/operator.go
grdryn: Done.

pkg/alertmanager/operator.go
grdryn: Done.

pkg/alertmanager/amcfg.go
grdryn: Done.

pkg/alertmanager/amcfg.go
grdryn: Done.

example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
grdryn: Done (v1beta1).

pkg/alertmanager/operator.go
grdryn: I started looking at adding to the existing admission webhook stuff, and how to add more to it, but then I stopped because it doesn't seem to be used by default. I've got some wip changes to it locally, but I think those might be better as a separate change rather than adding more to this one. Let me know if you think it should be in this PR and I can finish it and add here. :+1:

grdryn:
I'm going to remove the WIP prefix for this PR now, as I think it's close to a state that could be merged. Only the PagerDuty receiver has been added so far. I can add the others here, or I can do them in follow-up changes if you'd prefer.

lilic:
> Only the PagerDuty receiver has been added so far. I can add the others here, or I can do them in follow-up changes if you'd prefer.

I am fine in a follow-up as long as you can open the issues for maybe each one would be a good one for outside contributors to take, but up to @simonpasquier and @brancz on that one.

pkg/alertmanager/asset_store.go
```diff
@@ -0,0 +1,267 @@
...
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
```
dgrisonnet, line 29: Do we really need to duplicate `pkg/prometheus/asset_store` here?

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,509 @@
...
+	TargetMatch   map[string]string `yaml:"target_match,omitempty" json:"target_match,omitempty"`
+	TargetMatchRE map[string]string `yaml:"target_match_re,omitempty" json:"target_match_re,omitempty"`
+	SourceMatch   map[string]string `yaml:"source_match,omitempty" json:"source_match,omitempty"`
+	SourceMatchRE map[string]string `yaml:"source_match_re,omitempty" json:"source_match_re,omitempty"`
+	Equal         []string          `yaml:"equal,omitempty" json:"equal,omitempty"`
+}
+
+type receiver struct {
+	Name string `yaml:"name" json:"name"`
+	// EmailConfigs     []*EmailConfig     `yaml:"email_configs,omitempty" json:"email_configs,omitempty"`
```
lilic, line 100: Lets leave a TODO here why they are commented out.

pkg/alertmanager/operator.go
```diff
@@ -183,13 +281,150 @@ func (c *Operator) addHandlers() {
...
+	o, ok := c.getObject(obj)
+	if ok {
+		level.Debug(c.logger).Log("msg", "AlertmanagerConfig delete")
+		c.metrics.TriggerByCounter(monitoringv1beta1.AlertmanagerConfigKind, "delete").Inc()
+
+		c.enqueueForMonitorNamespace(o.GetNamespace())
+	}
+}
+
+// TODO: Do we need to enque secrets just for the namespace or in general?
```
lilic, line 335: Is this TODO still valid before we merge the PR?

pkg/apis/monitoring/v1beta1/doc.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2017 The prometheus-operator Authors
```
lilic, line 1: This is a new file yet it says 2017, I assume its generated so I guess we should fix some generation there, do you mind opening an issue if this is the case, thanks!

lilic (commented):
Great work on the PR! Thanks so much! Couple of comments, leaving up to Frederic and Simon to have a final look, no blockers on my side.

test/e2e/alertmanager_test.go
```diff
@@ -715,3 +716,101 @@ inhibit_rules:
...
+				}},
+			}},
+		},
+	}
+
+	if _, err := framework.MonClientV1beta1.AlertmanagerConfigs(ns).Create(context.TODO(), configCR, metav1.CreateOptions{}); err != nil {
+		t.Fatal(err)
+	}
+	// Wait for the change above to take effect.
+	time.Sleep(30 * time.Second)
```
dgrisonnet, line 776: sounds flaky, can we poll for the config to be created instead?

dgrisonnet (commented):
Awesome! Just a few comments, but it looks really good.

brancz:
I won't get to doing a deep dive on this soon, so it's better for @simonpasquier to do so. Two general comments:

* Yes I think starting with just one provider is fine, the rest is pretty mechanical and can be done in follow up PRs.
* Let's make sure we start with a v1alpha1 API :)

grdryn:
> > Only the PagerDuty receiver has been added so far. I can add the others here, or I can do them in follow-up changes if you'd prefer.
> 
> I am fine in a follow-up as long as you can open the issues for maybe each one would be a good one for outside contributors to take, but up to @simonpasquier and @brancz on that one.

Good idea. I've created the following issues for these now: #3533, #3534, #3535, #3536, #3537, #3538, #3539.

I may end up doing one or two of them that I know my org will use, but it should be possible for anyone to take any of them on once this is merged, I think! :slightly_smiling_face:

pkg/alertmanager/amcfg.go
grdryn: I think the created issues are better than a TODO, so I'll just remove these commented out lines here instead.

pkg/alertmanager/operator.go
grdryn: I plagiarized this TODO when I plagiarised this function from here :see_no_evil: https://github.com/prometheus-operator/prometheus-operator/blob/9a0823d7d1c2a1034a3ac7a05118cc02c1d0df52/pkg/prometheus/operator.go#L891

I could be wrong, but here's how I see it:

* We should enqueue reconciliation for an `Alertmanager` if any of the `Secrets` referenced in its spec are deleted (it has a `secrets` field which is a list, `configSecret` field, `containers` or `initContainers` fields may reference a Secret to mount, etc). These all have to be in the same namespace as the Alertmanager. This function, enqueuing all Alertmanagers in the namespace, covers these cases, I believe.

* We _may_ want to also enqueue reconciliation for an `Alertmanager`, if there's an AlertmanagerConfig selected by it that references a Secret that's deleted (upon reconciliation, if the Secret is missing for an AlertmanagerConfig, then that AlertmanagerConfig will be excluded when building the config file). If that's what we want, then I think it would _either_ mean enqueuing all Alertmanagers in _all_ namespaces, for any Secret changes in any AlertmanagerConfig namespaces; or determining if an Alertmanager should be reconciled by checking if the Secret is referenced in either the Alertmanager spec, or in any AlertmanagerConfigs that it selects.

I'm starting to confuse myself. In any case, if we want the second option in addition to the first, I'm not quite sure how to achieve that here. :thinking:

grdryn:
> Let's make sure we start with a v1alpha1 API :)

@brancz @lilic before I change this again, are you sure it's v1alpha1 that you want? (v1beta1 was suggested here a few days ago, and that's what it currently is in this change set)

brancz:
My apologies, I wasn't following that closely, I didn't catch that. I don't feel too strongly, but it just seems that if we want any type of feedback where we can still change the API then an alpha version seems appropriate.

lilic:
Yes, I think let's follow the correct order, sorry about that if it was me that suggested it initially. Let's go with `v1alpha1`! Thank you!

omegas27:
I would be up to do the opsgenie and the slack receivers

pkg/apis/monitoring/v1beta1/doc.go
grdryn: This file was actually manually created, I just copied it from the v1 folder I think, so I'll update it manually.

For generated files, they will have 2018 as the copyright year, so I've created #3546 for those just now.

pkg/alertmanager/asset_store.go
grdryn: For this I was using the [_Rule of Three_ for refactoring](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), probably because I didn't want to have to think about where best to move it to. If you want me to de-duplicate, just let me know where in the project I should move the existing asset_store code to. :slightly_smiling_face: 

Alternatively, the de-duplication could be done in a separate change set (this duplication doesn't really expose any more functions or types publicly from the project, but de-duplication would require some things to be exposed).

pkg/alertmanager/amcfg_test.go
```diff
@@ -0,0 +1,255 @@
...
+- name: mynamespace-myamc-test
+  pagerduty_configs:
+  - send_resolved: false
+    routing_key: 1234abc
+templates: []
+`,
+		},
+	}
+
+	for _, tc := range testCases {
```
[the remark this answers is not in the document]
grdryn: I'm hesitant to do that in this case. Currently it takes 0.030s to run the tests in this file on my machine (consistently over 5 runs), and following the instructions on the Go wiki for running table driven tests in parallel ([here](https://github.com/golang/go/wiki/TableDrivenTests#parallel-testing)), it takes between 0.030s and 0.032s.

Since the tests are already fast (and parallelization may seemingly slow them down in some cases), I think parallelizing them may be just adding complexity in this case. WDYT?

pkg/alertmanager/asset_store_test.go
```diff
@@ -0,0 +1,612 @@
...
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+)
+
+func TestAddBearerToken(t *testing.T) {
```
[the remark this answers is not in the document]
grdryn: As above, these unit tests are already fast. Additionally in this case, they're a copy of the existing tests for the asset store in the prometheus package.

grdryn:
> @grdryn I would be up to do the opsgenie and the slack receivers

@omegas27 sounds good to me! :slightly_smiling_face: Feel free to comment on one or both of the issues created for those indicating that you'd like to do them, in case anyone else thinks of doing them otherwise. Also, I wouldn't bother starting until this pull request gets merged, in case there are any structural changes needed here that would affect it.

When you do get around to it, if you need any help, feel free to contact me! :+1:

pkg/alertmanager/asset_store.go
dgrisonnet: You were right with doing so, but I can definitely see a third duplication occurring overtime.
Since refactoring would require to rewrite the original code partially, I think we can keep it as is here and do the refactor as part of another effort.

From the existing architecture, I'd say that the `k8sutil` package is the most neutral place to host this code.

cc @simonpasquier as you recently wrote this package and might have some thoughts on that.

lilic (approved):
Thanks for all the hard work on this one! 🎉 

After rebase and some commits squashed it lgtm 👍 We can always fix things in a follow up.

@prometheus-operator/prometheus-operator-reviewers letting you have a final look before we merge.

pkg/alertmanager/operator.go
```diff
@@ -131,6 +167,38 @@ func New(ctx context.Context, c prometheusoperator.Config, logger log.Logger, r
 		return nil, errors.Wrap(err, "error creating alertmanager informers")
 	}
 
+	o.alrtCfgInfs, err = informers.NewInformersForResource(
+		informers.NewMonitoringInformerFactories(
+			o.config.Namespaces.AlertmanagerAllowList,
```
simonpasquier, line 172: Prometheus operator would only watch for `AlertmanagerConfig` resources that live in the same namespace(s) than `Alertmanager` resources. If we want to support a different mode, we'd need to introduce an extra parameter/flag (e.g. `--alertmanagerconfig-instance-namespaces=...`). But this isn't going to work if you want to reconcile `Alertmanager` resources from one namespace with `AlertmanagerConfig` from all namespaces.

@s-urbaniak @brancz thoughts?

pkg/alertmanager/operator.go
```diff
@@ -131,6 +167,38 @@ func New(ctx context.Context, c prometheusoperator.Config, logger log.Logger, r
...
+		),
+		monitoringv1alpha1.SchemeGroupVersion.WithResource(monitoringv1alpha1.AlertmanagerConfigName),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating alertmanager informers")
+	}
+
+	o.secrInfs, err = informers.NewInformersForResource(
+		informers.NewKubeInformerFactories(
+			o.config.Namespaces.AlertmanagerAllowList,
```
simonpasquier, line 188: See comment about `alrtCfgInfs`

pkg/alertmanager/operator.go
```diff
@@ -475,6 +726,281 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	if secret != nil && secret.Data["alertmanager.yaml"] != nil {
+		rawBaseConfig = secret.Data["alertmanager.yaml"]
+	} else {
+		level.Warn(c.logger).Log("msg", "either base config secret not found, or alertmanager.yaml field not specified",
+			"secret", secretName, "alertmanager", am.Name, "namespace", am.Namespace)
+
+		fallbackConfig := "route:\n  receiver: 'null'\n" +
+			"receivers:\n- name: 'null'"
+		rawBaseConfig = []byte(fallbackConfig)
+	}
```
simonpasquier, line 751: I'm wondering if we should have `Info` rather than `Warn`?

```suggestion
	rawBaseConfig := []byte(`route:
	receiver: 'null'
receivers:
- name: 'null'`)
	if secret != nil && secret.Data["alertmanager.yaml"] != nil {
		rawBaseConfig = secret.Data["alertmanager.yaml"]
	} else {
		level.Info(c.logger).Log("msg", "either base config secret not found, or alertmanager.yaml field not specified",
			"secret", secretName, "alertmanager", am.Name, "namespace", am.Namespace)
	}
```

pkg/alertmanager/operator.go
```diff
@@ -475,6 +726,281 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+
+	baseConfig, err := loadCfg(string(rawBaseConfig))
+	if err != nil {
+		return errors.Wrap(err, "base config from Secret could not be parsed")
+	}
+
+	// If the base route is missing from the secret, then alertmanager will
+	// fail to load the config (and we'll fail to add subroutes)
+	foundRoute := baseConfig.Route != nil
```
simonpasquier, line 760: Alertmanager config loader should prevent that.

pkg/alertmanager/operator.go
```diff
@@ -475,6 +726,281 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+
+	err = c.createOrUpdateGeneratedConfigSecret(am, conf)
+	if err != nil {
+		return errors.Wrap(err, "create or update generated config secret failed")
+	}
+
+	return nil
+}
+
+func (c *Operator) createOrUpdateGeneratedConfigSecret(am *monitoringv1.Alertmanager, conf []byte) error {
```
simonpasquier, line 799:
```suggestion
func (c *Operator) createOrUpdateGeneratedConfigSecret(ctx context.Context, am *monitoringv1.Alertmanager, conf []byte) error {
```

pkg/alertmanager/operator.go
```diff
@@ -475,6 +726,281 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+			}
+		})
+	}
+
+	res := make(map[string]*monitoringv1alpha1.AlertmanagerConfig, len(amConfigs))
+	for namespaceAndName, amc := range amConfigs {
+		var err error
+
+		for i, receiver := range amc.Spec.Receivers {
+			amcKey := fmt.Sprintf("serviceMonitor/%s/%s/%d", amc.GetNamespace(), amc.GetName(), i)
```
simonpasquier, line 887:
```suggestion
			amcKey := fmt.Sprintf("alertmanagerConfig/%s/%s/%d", amc.GetNamespace(), amc.GetName(), i)
```

pkg/alertmanager/operator.go
grdryn: Good point, if it was nil we'd never get to this point. I've updated it now.

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,502 @@
...
+		pagerdutyConfigs = make([]*pagerdutyConfig, l)
+		for i := range in.PagerDutyConfigs {
+			pagerdutyConfigs[i] = cg.convertPagerdutyConfig(ctx, in.PagerDutyConfigs[i], crKey)
+		}
+	}
+
+	return &receiver{
+		Name:             prefixReceiverName(in.Name, crKey),
+		PagerdutyConfigs: pagerdutyConfigs,
+	}
```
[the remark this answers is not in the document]
grdryn: I've left this one as-is for now -- as we add additional receivers, I don't think we'd want to return if there aren't any PagerDutyConfigs or EmailConfigs, we'd just look at the subsequent receiver types. I think it will become clearer as we add the other receiver types, and we can update it then as needed.

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,502 @@
...
+	out.Details = details
+
+	if in.HTTPConfig != nil {
+		out.HTTPConfig = cg.convertHTTPConfig(ctx, *in.HTTPConfig, crKey)
+	}
+
+	return out
+}
+
+func convertInhibitRule(in *monitoringv1alpha1.InhibitRule) *inhibitRule {
```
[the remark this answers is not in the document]
grdryn: Oops, I missed that. Fixed, thanks.

pkg/alertmanager/amcfg.go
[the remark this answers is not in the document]
simonpasquier: :+1:

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,508 @@
...
+	if !strings.HasPrefix(r.Receiver, prefix) {
+		r.Receiver = prefix + r.Receiver
+	}
+
+	if r.Routes != nil {
+		for i := range r.Routes {
+			prefixRouteReceivers(&r.Routes[i], prefix)
+		}
+	}
+}
```
simonpasquier, line 177: It doesn't seem to be used anywhere.

pkg/alertmanager/operator.go
s-urbaniak: Taking precedence of the thanos ruler sub-operator code: The thanos ruler operator loop watches `PrometheusRule` objects in `o.config.Namespaces.AllowList`, not `o.config.Namespaces.ThanosRulerAllowList`., because `--thanos-ruler-instance-namespaces` is supposed to restrict namespaces that _deploy_ a whole thanos ruler.

I think the same rule applies here. `--alertmanager-instance-namespaces`, which translates to `o.config.Namespaces.AlertmanagerAllowList` should restrict namespaces that _deploy_ a whole alertmanager.

`AlertmanagerConfig` resources on the other hand only control a part of alertmanager configuration, not a whole deployment. This would be in line with the logic elsewhere. Users could restrict watching `AlertmanagerConfig` resources using `--namespaces` and `--deny-namespaces` and have finer grained control of namespaces which deploy an alertmanager instance using `--alertmanager-instance-namespaces`.

In the future we can add finer grained CLI options for concrete resources (not only `AlertmanagerConfig`) as a whole new set of settings.

TLDR: I suggest referencing `o.config.Namespaces.AllowList` here.

also cc'ing @prometheus-operator/prometheus-operator-reviewers for another set of opinions.

simonpasquier: :+1: to what @s-urbaniak outlined

grdryn: Done, thanks for the guidance! :slightly_smiling_face:

pkg/alertmanager/operator.go
```diff
@@ -176,13 +274,150 @@ func (c *Operator) addHandlers() {
...
+		return
+	}
+
+	for _, obj := range objs {
+		// Check for Alertmanager instances in the namespace.
+		am := obj.(*monitoringv1.Alertmanager)
+		if am.Namespace == nsName {
+			c.enqueue(am)
+			return
+		}
```
simonpasquier, line 401: we don't need this. Handlers on `Alertmanager` objects call `c.enqueue()` directly.

simonpasquier (commented):
A few comments for the Alertmanager controller. Otherwise this is looking really good. Given the amount of the code changes + the fact that the resource is `v1alpha1`, :+1: for merging this after addressing my last remarks.

pkg/alertmanager/operator.go
```diff
@@ -176,13 +274,150 @@ func (c *Operator) addHandlers() {
...
+	o, ok := c.getObject(obj)
+	if ok {
+		level.Debug(c.logger).Log("msg", "Secret added")
+		c.metrics.TriggerByCounter("Secret", "add").Inc()
+
+		c.enqueueForMonitorNamespace(o.GetNamespace())
+	}
+}
+
+func (c *Operator) enqueueForMonitorNamespace(nsName string) {
```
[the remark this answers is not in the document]
grdryn: Sure, I've done that now. I have to admit to being less familiar with this part of the changes, I think it comes from branczs earlier WIP.

pkg/alertmanager/operator.go
grdryn: Done, and again, not so familiar with this part. I think the other sub-operators (prometheus, thanos) may have the same thing. At the risk of pulling on another thread and making this PR longer, I'll leave those other ones for now since they're not related to this change.

pkg/alertmanager/operator.go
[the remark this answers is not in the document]
simonpasquier: I completely understand, you've done an amazing work knowing that you didn't know extensively the code base before :)

pkg/alertmanager/amcfg.go
```diff
@@ -0,0 +1,494 @@
...
+	cont := in.Continue
+	if firstLevelRoute {
+		cont = true
+	}
+
+	match := map[string]string{}
+	matchRE := map[string]string{}
+
+	for _, matcher := range in.Matchers {
+		if *matcher.Regex {
```
simonpasquier, line 244: need to check for `nil` or probably better to change `Regex` from `*bool` to `bool`.

pkg/alertmanager/operator.go
```diff
@@ -176,13 +274,142 @@ func (c *Operator) addHandlers() {
...
+		level.Error(c.logger).Log(
+			"msg", "listing all Alertmanager instances from cache failed",
+			"err", err,
+		)
+		return
+	}
+
+	for _, obj := range objs {
+		// Check for Alertmanager instances in the namespace.
+		am := obj.(*monitoringv1.Alertmanager)
```
simonpasquier, line 393: we still need to enqueue if the object is in the alertmanager's namespace and `alertmanagerConfigNamespaceSelector` is `nil`.

```suggestion
		am := obj.(*monitoringv1.Alertmanager)
		if am.Namespace == nsName {
			c.enqueue(am)
			return
		}
```

pkg/alertmanager/operator.go
simonpasquier: Sorry my remark was wrong... We need to enqueue an event here.

pkg/alertmanager/amcfg.go
grdryn: Changed to `bool`. The types in `alertmanager_config_types.go` are another thing that existed before I started working on this, so I didn't think about them too deeply. If there are any others you'd like me to change, let me know!

simonpasquier:
:+1: for me. Lets give other @prometheus-operator/prometheus-operator-reviewers a chance to review it :)

lilic (approved):
lgtm 👍 

Thanks again!!

lilic:
The only thing I would ask if the commits can be squashed, just to make it easier in case we need to revert something. 25 seems a bit too much even for this much changes. Do you mind doing that @grdryn and then we can merge 🎉

simonpasquier:
We could also `Squash and merge` with the GitHub UI.

lilic:
Personally, I like to have generated (API in this example) and actual code separated, but not a blocker for me, happy to do it via github UI as well.

simonpasquier:
Leaving the final call to @grdryn if he's willing to squash the commits and split the generated code in a separate commit :)

grdryn:
I've squashed to 3 commits: brancz's original work, then the rest is the split between manual code changes and generated files. Is that the split you're looking for?

lilic:
Thanks sounds good!

lilic:
Thanks so much again for the amazing work and patience! 🎉

moshloop:
Thank you to everyone involved in getting this merged 🎉🙌

keren-benzion:
Hi, 
I think this can come in handy in my organisation. Do you have maybe a document which explains how to implement this? to make it easier for me to deploy it?

Issue #3381, opened 2020-08-03
Secure Prometheus Config Reload


scott2449:
**What is missing?**
The current reload works via /-/reload endpoint which requires web.enable-lifecycle which also allows prometheus to be killed with /-/quit. For those of us exposing the Prometheus endpoint this is very problematic. Fortunately there is much better way to do this! https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/ (since 1.10). Once this is enabled you can send a `kill -1 1` (sighup process 1) from the sidecar instead.

**Why do we need it?**
Much lighter weight, more secure, and slightly faster. No need for web.enable-lifecycle at all!

**Environment**
All

* Prometheus Operator version:
N/A

**Anything else we need to know?**:
Nope

scott2449:
When I say current I am specifically talking about https://github.com/coreos/prometheus-operator/blob/master/cmd/prometheus-config-reloader/main.go

brancz:
Yeah I think this is a great one, we were waiting for the feature to be ubiquitous enough so that we could make use of it for this. I think there should be an old issue somewhere about this as well. If you would like to fix this, please go ahead! :)

scott2449:
Will do! We are going to take a crack at this on our project. If it works well we'll make sure it's nice and tidy and submit a PR 👌

scott2449:
Ooof at first I thought this was limited to the **configmap** reload... but the core reload is doing the same thing. IMO there should be 1 reloader to do both and neither should use the http endpoint. The unfortunately would require touching the operator helm, the operator itself, and the base helm. I pretty sure we are not going to get time for that... we also may need to abandon the operator completely and continue to manage ourselves.

brancz:
The fact that there are two reloaders is a bit of tech debt, there is no reason why they can't be one. You're of course free to choose tools as you wish, I highly doubt that rolling everything from scratch and not making use of the 4+ years of this project has in experience of operating, configuring and running Prometheus is less work than changing the reloader a bit. Again you're of course free to do what you think is best for your situation.

coderanger:
@scott2449 If you are exposing things through an Ingress, you can map the whole `/-/` path prefix to a different service, presumably one that does nothing but returns a 404 for all requests or similar.

scott2449:
[...]

PR #3457, opened 2020-09-01
*: use a single reloader for Prometheus

simonpasquier:
This change gets rid of the additional reloader that was deployed in the
Prometheus pod to monitor rules 

It also changes the Alertmanager and Thanos ruler statefulsets to use
the container image from the `--prometheus-config-reloader` flag instead
of `--config-reloader-image`. We pick `prometheus-config-reloader`
because it supports gzip'ed configuration file and is "smarter" about
when to reload the configuration (e.g. it checks that the content on
disk has effectively changed before reloading).

Since the Prometheus config reloader doesn't watch only rule directories
anymore (e.g. for Alertmanager it watches mounted secrets), it gets a
new `--watched-dir` flag.

Related to #3381.


brancz:
Very nice! I've been looking forward to getting back to this state!

simonpasquier:
While doing some tests, I've discovered that the Thanos reloader didn't detect changes to the watched directories in a timely fashion: it relies only on periodic resyncs instead of inotify events. For the time being, this PR uses my Thanos fork as a way to validate the Thanos reloader improvements but I'll submit these changes upstream of course.

simonpasquier:
https://github.com/thanos-io/thanos/pull/3136 got merged, I'll update this PR to depend on it.

simonpasquier:
This is ready for review :)

bundle.yaml
```diff
@@ -669,9 +669,14 @@ spec:
                   The secret is mounted into /etc/alertmanager/config.
                 type: string
               containers:
-                description: Containers allows injecting additional containers. This
+                description: 'Containers allows injecting additional containers. This
```
lilic, line 672: Curious but how come this wtih `'`?

cmd/operator/main.go
```diff
@@ -166,10 +167,10 @@ func init() {
 	// Prometheus Operator image, tagged with the same semver version. Default to
 	// the Prometheus Operator version if no Prometheus config reloader image is
 	// specified.
-	flagset.StringVar(&cfg.PrometheusConfigReloaderImage, "prometheus-config-reloader", operator.DefaultPrometheusConfigReloaderImage, "Prometheus config reloader image")
-	flagset.StringVar(&cfg.ConfigReloaderImage, "config-reloader-image", operator.DefaultConfigMapReloaderImage, "Reload Image")
-	flagset.StringVar(&cfg.ConfigReloaderCPU, "config-reloader-cpu", "100m", "Config Reloader CPU. Value \"0\" disables it and causes no limit to be configured.")
-	flagset.StringVar(&cfg.ConfigReloaderMemory, "config-reloader-memory", "25Mi", "Config Reloader Memory. Value \"0\" disables it and causes no limit to be configured.")
+	flagset.StringVar(&cfg.ReloaderConfig.Image, "prometheus-config-reloader", operator.DefaultPrometheusConfigReloaderImage, "Prometheus config reloader image")
+	flagset.StringVar(&deprecatedConfigReloaderImage, "config-reloader-image", "", "Reload image (deprecated)")
```
paulfantom, line 171: Can we add a comment to note when we want to remove this option?

cmd/operator/main.go
```diff
@@ -218,6 +219,14 @@ func Main() int {
 
 	logger.Log("msg", fmt.Sprintf("Starting Prometheus Operator version '%v'.", version.Version))
 
+	if deprecatedConfigReloaderImage != "" {
+		level.Warn(logger).Log(
+			"msg", "'--config-reloader-image' flag is ignored, only '--prometheus-config-reloader' is used",
```
lilic, line 224: If we are deprecating this right now, why are we ignoring it, should there not be a grace period of one release?

paulfantom:
I build images out of this PR and I'll test it after https://github.com/thaum-xyz/ankhmorpork/pull/42 is merged.

cmd/operator/main.go
paulfantom: I don't see anything wrong here since features provided by both config reloaders are now integrated into the one which is included in this repo. With this, we are not deprecating any features.

s-urbaniak: Interesting question though, I also lean towards the fact that we are not breaking the API (cli options) but change the functionality underneath.

simonpasquier: What @paulfantom said: we keep the CLI flag to not break existing deployments.
One caveat though: you won't be able to mix this new version of `prometheus-operator` with a previous version of `prometheus-config-reloader` because the former expects new CLI flags from the latter (and also the older `prometheus-config-reloader` versions have [bugs](https://github.com/prometheus-operator/prometheus-operator/pull/3457#issuecomment-688689962) when watching directories).

cmd/operator/main.go
simonpasquier: I've picked v0.45.0 but happy to hear otherwise.

bundle.yaml
simonpasquier: hmm not sure :)

simonpasquier: Maybe because I've updated the description and it now contains backticks?

cmd/operator/main.go
paulfantom: SGTM

paulfantom:
Results after running it for few hours:

1) CPU usage is 2-3x higher than what we had with previous reloaders.
Screenshot_20200915_185214

2) Rules are not reloaded. I get following error message in config-reloader logs:
```
level=error ts=2020-09-15T16:56:39.920275292Z caller=runutil.go:98 msg="function failed. Retrying in next tick" err="trigger reload: received non-200 response: 500 Internal Server Error; have you set `--web.enable-lifecycle` Prometheus flag?"
```
But `web.enable-lifecycle` is set to `true`.

Additionally, I am getting frequent CPUThrottlingHigh alert for new reloader after using it.

brancz:
Let's have a look but I also wouldn't worry too much when the scale is less than 1 millicore.

paulfantom (approved):
It seems that my issues were unrelated and fixed by https://github.com/prometheus-operator/prometheus-operator/pull/3515. It is strange that `PrometheusBadConfig` alert didn't fire though :thinking:  Especially that I had:
```
vel=error ts=2020-09-18T11:35:26.218Z caller=manager.go:930 component="rule manager" msg="loading groups failed" err="/etc/prometheus/rules/prometheus-k8s-rulefiles-0/monitoring-prometheus-operator-rules.yaml: 35:11: group \"prometheus-operator\", rule 3, \"PrometheusOperatorReconcileErrors\": could not parse expression: 2:1: parse error: unclosed left parenthesis"
level=error ts=2020-09-18T11:35:26.218Z caller=main.go:870 msg="Failed to apply configuration" err="error loading rules, previous rule set restored"
```
in prometheus logs.

Regarding this PR all looks good.

cmd/prometheus-config-reloader/main.go
```diff
@@ -17,21 +17,32 @@ package main
...
 const (
-	logFormatLogfmt                     = "logfmt"
-	logFormatJson                       = "json"
+	logFormatLogfmt = "logfmt"
+	logFormatJson   = "json"
+
+	logLevelDebug = "debug"
+	logLevelInfo  = "info"
+	logLevelWarn  = "warn"
+	logLevelError = "error"
```
lilic, line 44: We have none as well in prom-operator, is there a reason why we don't use it here as well? For consistency reasons is why I am mainly asking.

cmd/prometheus-config-reloader/main.go
```diff
@@ -72,24 +99,55 @@ func main() {
...
+		rel := reloader.New(
+			logger,
+			r,
+			&reloader.Options{
+				ReloadURL:     *reloadURL,
+				CfgFile:       *cfgFile,
+				CfgOutputFile: *cfgSubstFile,
+				WatchedDirs:   *watchedDir,
+				DelayInterval: 1 * time.Second,
+				WatchInterval: 3 * time.Minute,
```
lilic, line 147: Can you explain a bit the numbers here you used as the defaults. And is there a way to override them I couldn't see it yet? Thanks!

bundle.yaml
lilic: No clue, the only thing I am concerned and why I asked is because if its still valid as we don't have it for other descriptions. But I guess e2e tests passed so that confirms it?

pkg/operator/config.go
```diff
@@ -0,0 +1,97 @@
...
+			m[sp[0]] = sp[1]
+		}
+	}
+	(*labels).LabelsMap = m
+	(*labels).LabelsString = value
+	return nil
+}
+
+type Namespaces struct {
+	// allow list/deny list for common custom resources
```
lilic, line 93: Nit feel free to ignore: 

`// AllowList, DenyList for common custom resources`

pkg/operator/statefulset.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"fmt"
+	"net/url"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const configReloaderPort = 8080
```
lilic, line 25: Is this something we ever imagine users would want to modify maybe in the future via a flag?

lilic (commented):
Great work on the PR, a couple of questions.

cmd/prometheus-config-reloader/main.go
```diff
@@ -98,6 +156,15 @@ func main() {
 		})
 	}
 
+	if *listenAddress != "" {
+		g.Add(func() error {
+			level.Info(logger).Log("msg", "Starting web server for metrics", "listen", *listenAddress)
+			http.Handle("/metrics", promhttp.HandlerFor(r, promhttp.HandlerOpts{Registry: r}))
+			return http.ListenAndServe(*listenAddress, nil)
```
dgrisonnet, line 163: Is there a plan to monitor these metrics in the operator bundle/examples?

dgrisonnet (commented):
You mentioned some changes to the watch mechanism of the config reloader. Is there any chance we could restrict the operator ClusterRole with these?

cmd/prometheus-config-reloader/main.go
simonpasquier: I wasn't aware of the `none` level (Prometheus doesn't have it) but I agree that it's better to be consistent with what we have already.

cmd/prometheus-config-reloader/main.go
simonpasquier: The `WatchInterval` value is what was hardcoded previously in the Thanos code. I've chosen the `DelayInterval` value arbitrarily but open to make both values configurable.

lilic: Happy to have it in a followup or if we run into problems or per user requests, was just curious of the values. Up to you!

pkg/operator/statefulset.go
simonpasquier: I'm not sure, we don't do it for the other containers AFAICT.

simonpasquier:
> You mentioned some changes to the watch mechanism of the config reloader. Is there any chance we could restrict the operator ClusterRole with these?

The watch mechanism here is the config reloader watching the configuration file and directories. It doesn't have anything to do with Kubernetes API watches if this is what you're referring to.

cmd/prometheus-config-reloader/main.go
s-urbaniak: suggestion: maybe let's also add comments rationalizing the chosen numbers

simonpasquier: I've added a comment for each `const` identifier.

cmd/prometheus-config-reloader/main.go
simonpasquier: These metrics are exposed by the config reloader so the service monitor and alerts should probably live elsewhere than this repository?

pkg/operator/statefulset.go
```diff
@@ -0,0 +1,102 @@
...
+	)
+
+	var listenAddress string
+	if listenLocal {
+		listenAddress = localHost
+	} else {
+		ports = append(
+			ports,
+			v1.ContainerPort{
+				Name:          "web",
```
paulfantom, line 49: nit: Can this port be named differently? We are using the same name for prometheus pod and this will cause pod to have two ports which are named the same. It might be a bit confusing, especially when it comes to writing a ServiceMonitor.

cmd/prometheus-config-reloader/main.go
paulfantom: Seems like a follow-up would be in kube-prometheus as we will need to adjust 2 services and 2 serviceMonitors.
It would be best to include alerts for this in prometheus-operator mixin. But this would be best to do in a follow-up too.

pkg/operator/statefulset.go
simonpasquier: Good point!

lilic (approved):
lgtm

Do you mind just squashing the commits, thanks! 🎉 🚢

simonpasquier:
Commits squashed.

paulfantom (approved):
LGTM. Tests on my homelab didn't show any new problems.

s-urbaniak:
LGTM, thank you! :tada:

PR #3477, opened 2020-09-08
extract terminationgraceperiod to parameter

lurenjia528:
make terminationgraceperiod  Configurable


brancz:
Can you explain the use case here? There should not be a reason to explicitly set this.

lurenjia528:
> Can you explain the use case here? There should not be a reason to explicitly set this.

it is releate to 
https://github.com/prometheus-operator/prometheus-operator/pull/3434

https://github.com/prometheus-operator/prometheus-operator/issues/3433

brancz:
As mentioned in https://github.com/prometheus-operator/prometheus-operator/issues/3433, I don't think this is the right strategy to solve your problem.

Issue #3476, opened 2020-09-07
Probe with same job name not merged in prom config


nerzhul:
**What happened?**
Hello, we are trying your new very nice probe support. It works fine except if we have 2 probes with the same job name. That generate an invalid prometheus configuration with 2 jobs with the same name:

```
level=error ts=2020-09-07T14:44:54.512Z caller=main.go:614 msg="Error reloading config" err="couldn't load configuration (--config.file=\"/etc/prometheus/config_out/prometheus.env.yaml\"): parsing YAML file /etc/prometheus/config_out/prometheus.env.yaml: found multiple scrape configs with job name \"kube-prometheus-probe\""
```

**Did you expect to see some different?**
We would prefer to have a single job with both probes merged across namespaces

**How to reproduce it (as minimally and precisely as possible)**:
Create 2 probes on 2 namespaces with name job => boom

**Environment**
Kubernetes 1.18.8 with Debian 10.5 and containerd 1.4.0

* Prometheus Operator version: 0.41.1

    `Insert image tag or Git SHA here`
    

* Kubernetes version information: 1.18.8

    `kubectl version`
    

* Kubernetes cluster kind: managed by our internal SRE team

    insert how you created your cluster: kops, bootkube, tectonic-installer, etc.

* Manifests: 

```
insert manifests relevant to the issue
```

* Prometheus Operator Logs:

No specific log here

```
Insert Prometheus Operator logs relevant to the issue here
```

* Prometheus Logs:

```
Insert Prometheus logs relevant to the issue here
```

```
level=error ts=2020-09-07T14:44:54.512Z caller=main.go:614 msg="Error reloading config" err="couldn't load configuration (--config.file=\"/etc/prometheus/config_out/prometheus.env.yaml\"): parsing YAML file /etc/prometheus/config_out/prometheus.env.yaml: found multiple scrape configs with job name \"kube-prometheus-probe\""
```

**Anything else we need to know?**: N/A

brancz:
I see yes, this is indeed a bug.

Here: https://github.com/prometheus-operator/prometheus-operator/blob/289ee029ef537c38090cb9a0249e7062a3173936/pkg/prometheus/promcfg.go#L638

[...]

PR #3481, opened 2020-09-09
pkg/prometheus/promcfg.go: Generate more unique probe job name

lilic:
Closes https://github.com/prometheus-operator/prometheus-operator/issues/3476

I have not tested this as I am waiting for examples from @nerzhul. Hence WIP until then, just opening PR in case @nerzhul wants to test it out.

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -629,9 +629,6 @@ func (cg *configGenerator) generateProbeConfig(
 	enforcedNamespaceLabel string) yaml.MapSlice {
 
 	jobName := fmt.Sprintf("%s/%s", m.Namespace, m.Name)
-	if m.Spec.JobName != "" {
-		jobName = m.Spec.JobName
-	}
 	cfg := yaml.MapSlice{
 		{
 			Key:   "job_name",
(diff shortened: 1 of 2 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -275,10 +275,10 @@ func (cg *configGenerator) generateConfig(
 		}
 	}
 
-	for _, identifier := range probeIdentifiers {
+	for i, identifier := range probeIdentifiers {
 		scrapeConfigs = append(scrapeConfigs,
 			cg.generateProbeConfig(
-				version,
+				version, i,
```
brancz, line 281: probes can't actually produce multiple scrape configs, so it's sufficient to just use the identifier without an integer here

pkg/prometheus/promcfg.go
```diff
@@ -628,7 +629,7 @@ func (cg *configGenerator) generateProbeConfig(
 	ignoreNamespaceSelectors bool,
 	enforcedNamespaceLabel string) yaml.MapSlice {
 
-	jobName := fmt.Sprintf("%s/%s", m.Namespace, m.Name)
+	jobName := fmt.Sprintf("%s/%s/%d", m.Namespace, m.Name, i)
 	if m.Spec.JobName != "" {
 		jobName = m.Spec.JobName
```
brancz, line 634: this is the actual problem, we need to not set the job name directly, we need to instead relabel the `job` label in relabel configs

brancz (approved):
nice!

PR #3490, opened 2020-09-11
add_hostpath

linclaus:
Add a way to mount hostPath for Prometheus and Alertmanager


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/prometheus/busybox:latest
+FROM golang:1.13
```
coderanger, line 1: Why did you change this?

pkg/prometheus/statefulset.go
```diff
@@ -844,6 +852,8 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *Config, ruleConfigMapName
 		prometheusConfigReloaderResources.Limits[v1.ResourceMemory] = resource.MustParse(c.ConfigReloaderMemory)
 	}
 
+	privileged := true
```
coderanger, line 855: Uhhhhhhh, this is a very very bad thing.

coderanger (commented):
I think you may have accidentally included a lot more than you meant to. Are you sure you can't do this via the existing `containers` deep merge override?

linclaus:
> 
> 
> I think you may have accidentally included a lot more than you meant to. Are you sure you can't do this via the existing `containers` deep merge override?

linclaus:
> 
> 
> I think you may have accidentally included a lot more than you meant to. Are you sure you can't do this via the existing `containers` deep merge override?

There is something wrong with my local branch, and I will redo it after my correction.

PR #3491, opened 2020-09-11
add a way to mount hostpath for Prometheus and Alertmanager

linclaus:


brancz:
Can you elaborate on the use case here? This is generally an anti pattern, that I would not like to allow.

PR #3509, opened 2020-09-15
po-docgen: fix doc generation for inlined fields

dgrisonnet:
When po-docgen was generating the API documentation, it was
intentionally skipping inlined fields. Thus, they were absent from the
documentation instead of being displayed inline.
The assumption to skip inlined fields was correct for foreign types as
we most likely don't want their content to be part of the API
documentation. However, we want the content of internal types to be
inlined correctly.

This PR originates from the following discussion: https://github.com/prometheus-operator/prometheus-operator/pull/3370#discussion_r486931129

### DIFF
--- a/cmd/po-docgen/api.go
+++ b/cmd/po-docgen/api.go
@@ -51,6 +51,7 @@ var (
 	}
 
 	selfLinks = map[string]string{}
+	typesDoc  = map[string]KubeTypes{}
 )
 
 func toSectionLink(name string) string {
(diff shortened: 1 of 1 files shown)


dgrisonnet:
On top of my other PR, we get the following document: [api.md](https://github.com/prometheus-operator/prometheus-operator/compare/master...dgrisonnet:inlined-field-doc-example#diff-b1a95f484fe1197853692d47abbfe919)

paulfantom (approved):
lgtm

lilic (approved):
lgtm

Just one question, we had nothing inlined that was internal only external parts hence no changes to the api.md file, correct?

dgrisonnet:
> Just one question, we had nothing inlined that was internal only external parts hence no changes to the api.md file, correct?

Correct, we only had external metav1.TypeMeta fields inlined.

PR #3513, opened 2020-09-17
Add service label overload

omegas27:
Allow setting labels and annotations to alertmanager, prometheus and thanos ruler services to be able to scrape them using a servicemonitor


pkg/alertmanager/statefulset.go
```diff
@@ -162,12 +162,14 @@ func makeStatefulSetService(p *monitoringv1.Alertmanager, config Config) *v1.Ser
 		p.Spec.PortName = defaultPortName
 	}
 
+	labels := config.Labels.Merge(p.Spec.ServiceMetadata.Labels)
+	labels["operated-alertmanager"] = "true"
+
 	svc := &v1.Service{
```
lilic, line 168: This service is global per namespace, not per Alertmanager/Prometheus/ThanosRuler instance. So this approach won't work. But maybe I missing out something, can you describe your use case here?

omegas27: This service used by the statefulset seems to be created in the namespace where the Alertmanager/Prometheus/ThanosRuler are created yes.
```go
// Create governing service if it doesn't exist.
	svcClient := c.kclient.CoreV1().Services(p.Namespace)
	if err := k8sutil.CreateOrUpdateService(svcClient, makeStatefulSetService(p, c.config)); err != nil {
		return errors.Wrap(err, "synchronizing governing service failed")
	}
```

Our use case is that we have a management cluster and multiple tenant cluster managed by it and we have a prometheus server per tenant cluster.

Hence, by being able to add custom annotations like prometheus.io/scrape for example and labels to this service, we could define a unique ServiceMonitor to scrape every services in the management cluster and not a specific PodMonitor for the Alertmanager/Prometheus/ThanosRuler instances behind the services.

lilic: >This service used by the statefulset seems to be created in the namespace where the Alertmanager/Prometheus/ThanosRuler are created yes.

Yes but if you create two of them, they can conflict and would just override each other. Do you mind instead opening an issue with feature request detailing what you want to achieve, this approach will not work. Hope that makes sense, but thank you for all the contributions! 🎉 :)

omegas27: Sure, I will do that and close this PR for now then :) Thank you for your insightful reviews :D

PR #3523, opened 2020-09-21
rules: allow additional custom rule file paths

FUSAKLA:
Hi, I'm just adopting Prometheus operator after a long time provisioning Prometheus myself.
I'm trying to migrate the current setup as is but to use mainly the ServiceMonitors. 

Currently we have our own sidecar which does sync of git repo, linting and testing of rules and mounting them to the Prometheus container. I'm able to do all of this with the Operator except for specifying the custom path which the Prometheus should load in addition to the mounted configmaps paths.

So this allows user to configure custom rule file paths.

Signed-off-by: Martin Chodur <m.chodur@seznam.cz>


s-urbaniak:
I see this could help in migrating pure Prometheus setups to Prometheus Operator. I am just afraid that this circumvents many abstractions we built over time.

Let's see what the other maintainers think, cc @prometheus-operator/prometheus-operator-reviewers

FUSAKLA:
TBH I'm not completely sure if it will be just temporary solution. 
We are quite happy with having all the rules in one place adding possibility to test them if they do not collide anyhow, generic rules unit tests affecting multiple different rules across the system etc. The sidecar pulling works just fine so maybe it won't be just the migration path.

But I do get the concern. On the other hand there is already multiple "additional" ways to inject custom configuration, scrape targets, relabeling etc.

metalmatze:
I agree with @s-urbaniak on this. Probably migrating the ConfigMaps one after another should be the way forward. :man_shrugging:

FUSAKLA:
I'm not sure if I get it right that you are against this change or not? 
Or is there anything I can do additionally to make this more acceptable from the project perspective?

If not, what would you suggest, modifying the configmap generated by the operator myself?

FUSAKLA:
@metalmatze @s-urbaniak friendly bump, any update on this? I'm still blocked by it unfortunately :/

FUSAKLA:
:arrow_up:  rebased on current master

s-urbaniak:
As mentioned above I would like to have additional opinions from the other maintainers as I am not convinced we should allow circumventing all abstractions, cc'ing @prometheus-operator/kube-prometheus-reviewers

nrchakradhar:
Do you have any idea on a tool that we can write to migrate existing deployments to the operator? This would probably be the reverse of what operator does by generating prometheus.yaml from CRs. I can work along with you if you or others have some ideas/thoughts.

lilic:
Hello! What prevents you from migrating to the Custom Resources for the rules? Should be pretty straightforward to do that with a script if nothing else? If not can you explain why not, this PR lacks a few more info and use cases before it can be accepted.

I would suggest next time to open issue first where we can discuss if we want the feature, just to avoid work for you in case we say no. :)

PR #3525, opened 2020-09-22
Allow override of the prometheus web page title

omegas27:
Add the web.page-title flag to the prometheus CRD so that we can override the Prometheus title in the UI to be able to identify with a quick glance in which prometheus we are when we have multiple scraping different targets


JosephSalisbury:
**yes**

brancz:
Awesome! :tada: 

lgtm

Issue #3411, opened 2020-08-13
Service Sync - spec.ipFamily: Invalid value: "null": field is immutable


n1kofr:
**What happened?**

After deploying Prometheus Operator, the operated services are properly created, but the sync is failing due to the error: 

[spec.ipFamily: Invalid value: "null": field is immutable, spec.ipFamily: Required value]

**Did you expect to see something different?**

Prometheus Operator should be updated to manage the "spec.ipFamily"

**How to reproduce it (as minimally and precisely as possible)**:

Install Prometheus Operator with Prometheus Operator Chart on Kubernetes 1.18, then the warning is displayed regularly in the logs.

**Environment**

* Prometheus Operator version:

quay.io/coreos/prometheus-operator:v0.38.1

* Kubernetes version information:

Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:58:53Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

* Kubernetes cluster kind:

Created cluster using RKE.

* Manifests:

```
insert manifests relevant to the issue
```

* Prometheus Operator Logs:

```
ts=2020-08-13T12:19:18.818788468Z caller=main.go:209 msg="Starting Prometheus Operator version '0.38.1'."
ts=2020-08-13T12:19:18.826679805Z caller=main.go:99 msg="Staring insecure server on :8080"
level=info ts=2020-08-13T12:19:18.835576266Z caller=operator.go:452 component=prometheusoperator msg="connection established" cluster-version=v1.18.6
level=info ts=2020-08-13T12:19:18.835604574Z caller=operator.go:294 component=thanosoperator msg="connection established" cluster-version=v1.18.6
level=info ts=2020-08-13T12:19:18.835585321Z caller=operator.go:214 component=alertmanageroperator msg="connection established" cluster-version=v1.18.6
[...]

PR #3526, opened 2020-09-22
Update CreateOrUpdateService in k8sutil

KadenLNelson:
When updating the service, set the `IPFamily` attribute of the service spec.

fixes https://github.com/prometheus-operator/prometheus-operator/issues/3411

---

Along with https://github.com/n1kofr, I am also experiencing the issue where prometheus operator is failing to sync/update governing services. The root cause I believe is within these lines of code https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/k8sutil/k8sutil.go#L108-L129. When prometheus-operator is updating the service, the svc object in the CreateOrUpdateService function does not have the IPFamily attribute set in its object spec, thus it's effectively nil which explains the following error;

```log
E0921 17:20:17.185103       1 operator.go:322] Sync "system/prometheus-operator-kube-p-alertmanager" failed: synchronizing governing service failed: updating service object failed: Service "alertmanager-operated" is invalid: [spec.ipFamily: Invalid value: "null": field is immutable, spec.ipFamily: Required value]
```

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -118,6 +118,7 @@ func CreateOrUpdateService(ctx context.Context, sclient clientv1.ServiceInterfac
 		}
 	} else {
 		svc.ResourceVersion = service.ResourceVersion
+		svc.Spec.IPFamily = service.Spec.IPFamily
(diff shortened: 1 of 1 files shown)


brancz:
Any reason why this is just a draft? :)

PR #3531, opened 2020-09-23
set pkg/apis/monitoring tag to v0.42.1

TheoBrigitte:
Towards: https://github.com/prometheus-operator/prometheus-operator/issues/3530

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -31,17 +31,26 @@ Maintaining the release branches for older minor releases happens on a best effo
 
 ## Prepare your release
 
+Bump the version in the `VERSION` file in the root of the repository.
+
+While on the master branch tag `pkg/apis/monitoring` package:
+
+```bash
+$ tag=$(< VERSION) && git tag -s "pkg/apis/monitoring/v${tag}" -m "pkg/apis/monitoring/v${tag}"
+$ git push origin "pkg/apis/monitoring/v${tag}"
+```
+
 For a patch release, work in the branch of the minor release you want to patch.
 
 For a new major or minor release, create the corresponding release branch based on the master branch.
 
-Bump the version in the `VERSION` file in the root of the repository. Once that's done, a number of files have to be re-generated, this is automated with the following make target:
+Once that's done, a number of files have to be re-generated, this is automated with the following make target:
 
 ```bash
 $ make clean generate
 ```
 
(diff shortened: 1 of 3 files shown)


lilic:
Thanks for the PR, I actually have not yet finished the 0.42.1 release by bringing it into master, but this is a good start! Thanks so much! Do you mind just updating the release notes here with the above steps, its not needed just would be nice. If not I can do it in another PR as well! https://github.com/prometheus-operator/prometheus-operator/blob/master/RELEASE.md

TheoBrigitte:
You're welcome. Yes sure I can do that, you mean updating the CHANGELOG.md right ?

lilic:
Sorry was not clear enough, the CHANGELOG I can update in a follow up PR, I meant adding the change you just did in go.mod file to the RELEASE.md process so we don't forget this step in the future, hope that makes more sense?

TheoBrigitte:
Ha ok understood :+1:

TheoBrigitte:
Wait we can't actually do this, because the tag is not yet there when creating the release. So `go build` will fail since it won't be able to find the new tag.

TheoBrigitte:
`pkg/apis/monitoring` will always be 1 version behind in go.mod then. Feels weird to me. Otherwise this has to happen after the release was created.

simonpasquier:
Yes we need to tag `github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring` before cutting the prometheus-operator release.

TheoBrigitte:
This should do it:

* from `master` set the new tag on `pkg/apis/monitoring`
* move to release branch and prepare the release including getting the new `pkg/apis/monitoring` version
* do the release

Tag from `pkg/apis/monitoring` will always be behind, but I do not see another way of achieving this.

RELEASE.md
```diff
@@ -31,16 +31,31 @@ Maintaining the release branches for older minor releases happens on a best effo
...
+A number of files have to be re-generated, this is automated with the following make target:
 
 ```bash
 $ make clean generate
 ```
 
+Bump the version for `pkg/apis/monitoring` package in go.mod:
+
+```bash
+$ go get -u github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring
```
simonpasquier, line 56:
```suggestion
$ go get "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring@v${VERSION}"
```

TheoBrigitte:
Are we good here?

brancz:
Thank you!

PR #3545, opened 2020-09-25
Don't block when namespace aren't found

### LINKED ISSUE #3347: Syncing prometheus fails when a ns error occurs but none of other resources have this problem
**What happened?**

Passed two namespaces that do not exist and two that do exist via the prometheus operator flag `--namespaces`. Seen the expected errors of:
> E0720 10:40:22.043151   18190 reflector.go:178] github.com/coreos/prometheus-operator/pkg/prometheus/operator.go:484: Failed to list *v1.Namespace: namespaces "blah" not found

All resources were up and running and syncing for example:
> level=info ts=2020-07-20T08:40:33.573732Z caller=operator.go:427 component=alertmanageroperator msg="sync alertmanager" key=monitoring/main

and it correctly picked up the Alertmanager CR instance.

But it failed to the same for prometheus CR instance and create the needed resources, and in the logs it seems like the syncing never actually happened.

When only the existing namespaces are created this error does not happen, I can see the sync of prometheus in the logs and prometheus resources are created correctly:
> level=info ts=2020-07-20T08:37:02.519745Z caller=operator.go:171 component=alertmanageroperator msg="successfully synced all caches"
level=info ts=2020-07-20T08:37:02.519754Z caller=operator.go:407 component=prometheusoperator msg="successfully synced all caches"
level=info ts=2020-07-20T08:37:02.519874Z caller=operator.go:427 component=alertmanageroperator msg="sync alertmanager" key=monitoring/main
level=info ts=2020-07-20T08:37:02.519917Z caller=operator.go:1122 component=prometheusoperator msg="sync prometheus" key=monitoring/k8s
level=info ts=2020-07-20T08:37:02.555934Z caller=operator.go:427 component=alertmanageroperator msg="sync alertmanager" key=monitoring/main
level=info ts=2020-07-20T08:37:02.620839Z caller=operator.go:1122 component=prometheusoperator msg="sync prometheus" key=monitoring/k8s

I expected that prometheus operator ignores the errors and tolerates those and keeps reconciles despite any non related errors like namespaces in the above example.

Start prometheus operator with three non existing namespaces, example:
`--namespaces=default,ns2,blah,ns1,monitoring`

**Environment**

Tried with 0.40 and master

    `kubectl version`
    

* Kubernetes cluster kind:

1.18.3 vanilla minikube cluster.

simonpasquier:
The namespaces listerwatcher shouldn't return an error when some of the monitored namespaces don't exist. Otherwise the operator will block while waiting for the informers to sync their cache and it will never be able to reconcile resources.
    
Closes #3347

### DIFF
--- a/example/mixin/alerts.yaml
+++ b/example/mixin/alerts.yaml
@@ -41,3 +41,13 @@ groups:
     for: 10m
     labels:
       severity: warning
+  - alert: PrometheusOperatorNotReady
+    annotations:
+      description: Prometheus operator in {{ $labels.namespace }} namespace isn't
+        ready to reconcile {{ $labels.controller }} resources.
+      summary: Prometheus operator not ready
+    expr: |
+      min by(namespace, controller) (max_over_time(prometheus_operator_ready{job="prometheus-operator"}[5m]) == 0)
+    for: 5m
+    labels:
+      severity: warning
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -44,6 +44,7 @@ type Metrics struct {
 	// objects. It is split in the dimensions of Kubernetes objects and
 	// corresponding actions (add, delete, update).
 	triggerByCounter *prometheus.CounterVec
+	ready            prometheus.Gauge
 }
 
 // NewMetrics initializes operator metrics and registers them with the given registerer.
@@ -85,6 +86,10 @@ func NewMetrics(name string, r prometheus.Registerer) *Metrics {
 			Name: "prometheus_operator_watch_operations_failed_total",
 			Help: "Total number of watch operations that failed",
 		}),
+		ready: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "prometheus_operator_ready",
+			Help: "1 when the controller is ready to reconcile resources, 0 otherwise",
+		}),
 	}
 	m.reg.MustRegister(
 		m.reconcileCounter,
@@ -95,6 +100,7 @@ func NewMetrics(name string, r prometheus.Registerer) *Metrics {
 		m.listFailedCounter,
 		m.watchCounter,
 		m.watchFailedCounter,
+		m.ready,
 	)
 	return &m
 }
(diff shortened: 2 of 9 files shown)


pkg/operator/operator.go
```diff
@@ -201,3 +206,35 @@ func SanitizeSTS(sts *appsv1.StatefulSet) {
 		sts.Spec.VolumeClaimTemplates[i].Kind = ""
 	}
 }
+
+// WaitForCacheSync synchronizes the informer's cache and will log a warning
+// every minute if the operation hasn't completed yet.
+func WaitForCacheSync(ctx context.Context, logger log.Logger, inf cache.SharedIndexInformer) bool {
```
s-urbaniak, line 212: Do you mind to elaborate why we need to do this in a retry-fashion?

simonpasquier: I've updated the comment

simonpasquier:
I've also added a metric `prometheus_operator_ready` tracking whether a given controller is ready to reconcile resources or not.

s-urbaniak:
LGTM, but leaving another pair of eyes to @prometheus-operator/kube-prometheus-reviewers

example/mixin/alerts.yaml
```diff
@@ -41,3 +41,13 @@ groups:
     for: 10m
     labels:
       severity: warning
+  - alert: PrometheusOperatorNotReady
+    annotations:
+      description: Prometheus operator in {{ $labels.namespace }} namespace isn't
+        ready to reconcile {{ $labels.controller }} resources.
+      summary: Prometheus operator not ready
+    expr: |
+      min by(namespace, controller) (min_over_time(prometheus_operator_ready{job="prometheus-operator"}[5m]) == 0)
```
lilic, line 50: We can tweak this later, but maybe 5 minutes might be a bit too short?

lilic (approved):
Great work! 🎉 👍 

We can tweak alert when we deploy into our clusters or on user feedback, not blocking at all.

example/mixin/alerts.yaml
simonpasquier: After more thinking maybe it should be:

```
min by(namespace, controller) (max_over_time(prometheus_operator_ready{job="prometheus-operator"}[5m]) == 0)
```

PR #3554, opened 2020-09-30
cmd/prometheus-config-reloader: enable pprof endpoints

simonpasquier:


s-urbaniak:
lgtm :+1:

lilic (approved):
🎉 lgtm

simonpasquier:
Given that the config reloader is (should be) low on resources, I'm not sure it's worth adding the overhead of pprof endpoints.
If anyone wants to investigate memory/CPU usage, it's probably better to manually profile the binary or add benchmarks to `github.com/thanos-io/thanos/pkg/reloader`.

Issue #3540, opened 2020-09-24
Flaky e2e test: E2E tests (alertmanager) 


lilic:
We seem to have had a flaky test for some time, here are some examples I did not have a chance to look into it yet, but it does finish after some retries.

https://github.com/prometheus-operator/prometheus-operator/runs/1155429560?check_suite_focus=true 
>         --- FAIL: TestAllNS/x/AMClusterAfterRollingUpdate (81.08s)
https://github.com/prometheus-operator/prometheus-operator/runs/1153860258?check_suite_focus=true
>         --- FAIL: TestAllNS/x/AMZeroDowntimeRollingDeployment (184.50s)

Seems like there are a couple of flakes, not just one. Might be as easy as retrying or as difficult as rewriting tests.

simonpasquier:
Another flake here: https://pipelines.actions.githubusercontent.com/vPMnvJjkGBteTBBeHIuBuscIAxYyVzD3ZDtsis3lp8F0MLY5xH/_apis/pipelines/1/runs/129/signedlogcontent/16?urlExpires=2020-10-01T11%3A52%3A28.2712240Z&urlSigningMethod=HMACV1&urlSignature=tceE7Uj9qlkXbu3QFekc5Bjz2TuYbCSlH6hUQbNijB0%3D

BTW the e2e tests still use the [Alertmanager API v1](https://github.com/prometheus-operator/prometheus-operator/blob/434f9f7e0b75cd1bcb444cf4532a53112074af4e/test/framework/alertmanager.go#L257) and this should be changed.

lilic:
@simonpasquier 
> BTW the e2e tests still use the Alertmanager API v1 and this should be changed.

Would be a good first issue maybe, do you mind opening it?

simonpasquier:
Instead of returning the error from `GetAlertmanagerStatus()`, we should save it in `pollErr` and return no error.

https://github.com/prometheus-operator/prometheus-operator/blob/434f9f7e0b75cd1bcb444cf4532a53112074af4e/test/framework/alertmanager.go#L218-L221

arghya88:
I will work on this.

PR #3558, opened 2020-10-01
Update Alertmanager api to v2 in E2E test

### LINKED ISSUE #3557: End-to-end tests should use the Alertmanager API v2
the e2e tests still use the [Alertmanager API v1](https://github.com/prometheus-operator/prometheus-operator/blob/434f9f7e0b75cd1bcb444cf4532a53112074af4e/test/framework/alertmanager.go#L257) and this should be changed to the API v2 instead.

Discovered while looking at https://github.com/prometheus-operator/prometheus-operator/issues/3540.

arghya88:
Fixes #3557

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -31,10 +31,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/prometheus-operator/prometheus-operator/pkg/alertmanager"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	testFramework "github.com/prometheus-operator/prometheus-operator/test/framework"
-	"github.com/golang/protobuf/proto"
 )
 
 func testAMCreateDeleteCluster(t *testing.T) {
@@ -327,8 +327,8 @@ func testAMClusterGossipSilences(t *testing.T) {
 				return false, nil
 			}
 
-			if silences[0].ID != silId {
-				return false, errors.Errorf("expected silence id on alertmanager %v to match id of created silence '%v' but got %v", i, silId, silences[0].ID)
+			if *silences[0].ID != silId {
+				return false, errors.Errorf("expected silence id on alertmanager %v to match id of created silence '%v' but got %v", i, silId, *silences[0].ID)
 			}
 			return true, nil
 		})
@@ -636,7 +636,6 @@ inhibit_rules:
 
 		go func() {
 			ticker := time.NewTicker(100 * time.Millisecond)
-			start := time.Now()
 			failures := 0
 			for {
 				select {
@@ -644,7 +643,6 @@ inhibit_rules:
 					err := framework.SendAlertToAlertmanager(
 						ns,
 						"alertmanager-rolling-deploy-"+strconv.Itoa(replica),
-						start,
 					)
 					if err != nil {
 						failures++
--- a/test/framework/alertmanager.go
+++ b/test/framework/alertmanager.go
@@ -29,9 +29,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
 
+	"github.com/pkg/errors"
 	"github.com/prometheus-operator/prometheus-operator/pkg/alertmanager"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
-	"github.com/pkg/errors"
+	"github.com/prometheus/alertmanager/api/v2/client/silence"
+	"github.com/prometheus/alertmanager/api/v2/models"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/textparse"
 )
@@ -221,28 +223,31 @@ func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers 
 		}
 
 		isAlertmanagerInClusterMode := amountPeers > 1 || forceEnableClusterMode
-		if !isAlertmanagerInClusterMode && amStatus.Status == "success" {
+		if !isAlertmanagerInClusterMode {
 			return true, nil
 		}
 
-		if amStatus.Data.ClusterStatus == nil {
+		if amStatus.Cluster == nil {
 			pollError = fmt.Errorf("do not have a cluster status")
 			return false, nil
 		}
 
-		if len(amStatus.Data.ClusterStatus.Peers) == amountPeers {
-			return true, nil
-
+		if *amStatus.Cluster.Status != "ready" {
+			pollError = fmt.Errorf("failed to get cluser status, expected ready, got %s", *amStatus.Cluster.Status)
+			return false, nil
 		}
 
-		pollError = fmt.Errorf(
-			"failed to get correct amount of peers, expected %d, got %d, addresses %v",
-			amountPeers,
-			len(amStatus.Data.ClusterStatus.Peers),
-			amStatus.Data.ClusterStatus.Peers,
-		)
+		if len(amStatus.Cluster.Peers) != amountPeers {
 
-		return false, nil
+			var addrs = make([]string, len(amStatus.Cluster.Peers))
+			for i := range amStatus.Cluster.Peers {
+				addrs[i] = *amStatus.Cluster.Peers[i].Name
+			}
+			pollError = fmt.Errorf("failed to get correct amount of peers, expected %d, got %d, addresses %v", amountPeers, len(amStatus.Cluster.Peers), addrs)
+			return false, nil
+
+		}
+		return true, nil
 	})
 
 	if err != nil {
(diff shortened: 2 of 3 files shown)


simonpasquier:
Thanks for this! You need also to adapt the structs that parse the API responses because v1 and v2 payloads are different: 

https://github.com/prometheus-operator/prometheus-operator/blob/434f9f7e0b75cd1bcb444cf4532a53112074af4e/test/framework/alertmanager.go#L444-L484

Because the repository already depends [indirectly](https://github.com/prometheus-operator/prometheus-operator/blob/434f9f7e0b75cd1bcb444cf4532a53112074af4e/go.sum#L859-L864) on `github.com/prometheus/alertmanager`, we can import `github.com/prometheus/alertmanager/api/v2/models` to use [`AlertmanagerStatus`](https://pkg.go.dev/github.com/prometheus/alertmanager@v0.21.0/api/v2/models#AlertmanagerStatus), [`PostableSilence`](https://pkg.go.dev/github.com/prometheus/alertmanager@v0.21.0/api/v2/models#PostableSilence), [`GettableSilences`](https://pkg.go.dev/github.com/prometheus/alertmanager@v0.21.0/api/v2/models#GettableSilences) and [`PostableAlerts`](https://pkg.go.dev/github.com/prometheus/alertmanager@v0.21.0/api/v2/models#PostableAlerts).

test/framework/alertmanager.go
```diff
@@ -292,15 +293,7 @@ func (f *Framework) CreateSilence(ns, n string) (string, error) {
...
-			"expected Alertmanager to return 'success', but got '%v' instead",
-			createSilenceResponse.Status,
-		)
-	}
-
-	return createSilenceResponse.Data.SilenceID, nil
+	return createSilenceResponse.ID, nil
 }
 
 // alert represents an alert that can be posted to the /api/v1/alerts endpoint
```
simonpasquier, line 299: we can use `models.PostableAlert` and remove the internal `alert` struct.

test/framework/alertmanager.go
```diff
@@ -336,8 +329,8 @@ func (f *Framework) SendAlertToAlertmanager(ns, n string, start time.Time) error
 		return err
 	}
 
-	var postAlertResp amAPIPostAlertResp
-	request := ProxyPostPod(f.KubeClient, ns, n, "api/v1/alerts", string(b))
+	var postAlertResp models.PostableAlerts
```
simonpasquier, line 332: We don't need (anymore?) to unmarshal the response into `postAlertResp` because the server replies with an empty body and `200 OK` in case of success.

test/framework/alertmanager.go
```diff
@@ -347,47 +340,43 @@ func (f *Framework) SendAlertToAlertmanager(ns, n string, start time.Time) error
...
 	}
 
-	if getSilencesResponse.Status != "success" {
-		return getSilencesResponse.Data, errors.Errorf(
-			"expected Alertmanager to return 'success', but got '%v' instead",
-			getSilencesResponse.Status,
+	if *getSilencesResponse[0].Status.State != "active" {
+		return getSilencesResponse, errors.Errorf(
+			"expected Alertmanager to return 'active', but got '%v' instead",
+			getSilencesResponse,
```
simonpasquier, line 362: This isn't needed, the caller will check that the IDs matche.

```suggestion
```

simonpasquier (commented):
I've made a few remarks but it's looking great!

test/framework/alertmanager.go
```diff
@@ -221,25 +222,25 @@ func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers
...
 		}
 
-		if amStatus.Data.ClusterStatus == nil {
+		if amStatus.Cluster == nil {
 			pollError = fmt.Errorf("do not have a cluster status")
 			return false, nil
 		}
 
-		if len(amStatus.Data.ClusterStatus.Peers) == amountPeers {
+		if *amStatus.Cluster.Status == "ready" && len(amStatus.Cluster.Peers) == amountPeers {
```
simonpasquier, line 234: can you split the conditions in 2 and set a proper message for `pollError`?

test/framework/alertmanager.go
```diff
@@ -292,102 +293,61 @@ func (f *Framework) CreateSilence(ns, n string) (string, error) {
...
 
 // SendAlertToAlertmanager sends an alert to the alertmanager in the given
 // namespace (ns) with the given name (n).
-func (f *Framework) SendAlertToAlertmanager(ns, n string, start time.Time) error {
-	alerts := []*alert{&alert{
-		Labels: map[string]string{
-			"alertname": "ExampleAlert", "prometheus": "my-prometheus",
+func (f *Framework) SendAlertToAlertmanager(ns, n string) error {
+
+	alerts := models.PostableAlert{
```
simonpasquier, line 303:
```suggestion
	alerts := models.PostableAlerts{
```

arghya88:
@simonpasquier Have made the suggested changes. Any idea why the tests are failing? Is it because of flake?

test/framework/alertmanager.go
```diff
@@ -221,25 +222,31 @@ func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers
...
 
-		if len(amStatus.Data.ClusterStatus.Peers) == amountPeers {
-			return true, nil
+		if *amStatus.Cluster.Status != "ready" {
+			pollError = fmt.Errorf("failed to get cluser status, expected ready, got %s", *amStatus.Cluster.Status)
+			return false, nil
+		}
+
+		if len(amStatus.Cluster.Peers) != amountPeers {
+			pollError = fmt.Errorf("failed to get correct amount of peers, expected %d, got %d", amountPeers, len(amStatus.Cluster.Peers))
```
simonpasquier, line 240:
```suggestion
			var addrs := make([]string, len(amStatus.Cluster.Peers))
			for i := range amStatus.Cluster.Peers {
				addrs[i] = *amStatus.Cluster.Peers[i].Name
			}
			pollError = fmt.Errorf("failed to get correct amount of peers, expected %d, got %d, addresses %v", amountPeers, len(amStatus.Cluster.Peers), addrs)
```

test/framework/alertmanager.go
```diff
@@ -221,25 +222,31 @@ func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers
...
+			return false, nil
 
 		}
 
 		pollError = fmt.Errorf(
```
simonpasquier, line 245: you need to remove this line and have `return true, nil` as the last statement because when we reach this point, all checks have passed.

simonpasquier (commented):
> Any idea why the tests are failing?

You need to update slightly `WaitForAlertmanagerInitialized()` and it should be ok.

test/e2e/alertmanager_test.go
```diff
@@ -327,7 +327,7 @@ func testAMClusterGossipSilences(t *testing.T) {
 				return false, nil
 			}
 
-			if silences[0].ID != silId {
+			if *silences[0].ID != silId {
 				return false, errors.Errorf("expected silence id on alertmanager %v to match id of created silence '%v' but got %v", i, silId, silences[0].ID)
```
simonpasquier, line 331:
```suggestion
				return false, errors.Errorf("expected silence id on alertmanager %v to match id of created silence '%v' but got %v", i, silId, *silences[0].ID)
```

test/framework/alertmanager.go
```diff
@@ -277,11 +281,11 @@ func (f *Framework) GetAlertmanagerMetrics(ns, n string) (textparse.Parser, erro
 }
 
 func (f *Framework) CreateSilence(ns, n string) (string, error) {
-	var createSilenceResponse amAPICreateSilResp
+	var createSilenceResponse models.PostableSilence
```
simonpasquier, line 284: Ah, my bad. The returned payload isn't a `models.PostableSilence` struct. We need [`PostSilencesOKBody`](https://github.com/prometheus/alertmanager/blob/4c6c03ebfe21009c546e4d1e9b92c371d67c021d/api/v2/client/silence/post_silences_responses.go#L162-L166) here.

test/framework/alertmanager.go
```diff
@@ -277,11 +281,11 @@ func (f *Framework) GetAlertmanagerMetrics(ns, n string) (textparse.Parser, erro
...
 
 	request := ProxyPostPod(
 		f.KubeClient, ns, n,
-		"/api/v1/silences",
+		"/api/v2/silences",
 		`{"id":"","createdBy":"Max Mustermann","comment":"1234","startsAt":"2030-04-09T09:16:15.114Z","endsAt":"2031-04-09T11:16:15.114Z","matchers":[{"name":"test","value":"123","isRegex":false}]}`,
```
simonpasquier, line 289: (nit) we don't need `id` here

```suggestion
		`{"createdBy":"Max Mustermann","comment":"1234","startsAt":"2030-04-09T09:16:15.114Z","endsAt":"2031-04-09T11:16:15.114Z","matchers":[{"name":"test","value":"123","isRegex":false}]}`,
```

test/framework/alertmanager.go
arghya88: So we need to create a copy of the struct right or there is a way to import and use?

test/framework/alertmanager.go
```diff
@@ -441,44 +405,8 @@ func (f *Framework) WaitForAlertmanagerConfigToBeReloaded(ns, amName string, pre
...
-type amAPIStatusData struct {
-	ClusterStatus *clusterStatus `json:"clusterStatus,omitempty"`
-	ConfigYAML    string         `json:"configYAML"`
-}
-
-type clusterPeer struct {
-	Name    string `json:"name"`
-	Address string `json:"address"`
-}
+type postSilencesOKBody struct {
```
simonpasquier, line 408: It's ok to import the struct from `github.com/prometheus/alertmanager/api/v2/client/silence` rather than duplicating it since we already depend on the `github.com/prometheus/alertmanager` module anyway.

simonpasquier (approved):
:+1:

simonpasquier:
Thanks for your contribution @arghya88! If you're up for another one, we still have to fix [#3540](https://github.com/prometheus-operator/prometheus-operator/issues/3540#issuecomment-702140486) :smiley:

PR #3560, opened 2020-10-03
update thanos example rbac

kiddo3:
Hi,
trying to making work the example, I found the rbac for the thanos example was out of date. I used the same permissions specifieds in: [https://github.com/kiddo3/prometheus-operator/blob/master/Documentation/rbac.md](https://github.com/kiddo3/prometheus-operator/blob/master/Documentation/rbac.md)


brancz:
Thanks!

PR #3565, opened 2020-10-05
Fix flaky e2e Alertmanager test 

### LINKED ISSUE #3540: Flaky e2e test: E2E tests (alertmanager)

arghya88:
Fixes #3540

### DIFF
--- a/test/framework/alertmanager.go
+++ b/test/framework/alertmanager.go
@@ -217,9 +217,11 @@ func amImage(version string) string {
 func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers int, forceEnableClusterMode bool) error {
 	var pollError error
 	err := wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
+
 		amStatus, err := f.GetAlertmanagerStatus(ns, name)
 		if err != nil {
-			return false, err
+			pollError = fmt.Errorf("failed to query Alertmanager: %s", err)
+			return false, nil
 		}
 
 		isAlertmanagerInClusterMode := amountPeers > 1 || forceEnableClusterMode
@@ -261,6 +263,7 @@ func (f *Framework) GetAlertmanagerStatus(ns, n string) (models.AlertmanagerStat
 	var amStatus models.AlertmanagerStatus
 	request := ProxyGetPod(f.KubeClient, ns, n, "/api/v2/status")
 	resp, err := request.DoRaw(context.TODO())
+
 	if err != nil {
 		return amStatus, err
 	}
@@ -268,7 +271,6 @@ func (f *Framework) GetAlertmanagerStatus(ns, n string) (models.AlertmanagerStat
 	if err := json.Unmarshal(resp, &amStatus); err != nil {
 		return amStatus, err
 	}
-
 	return amStatus, nil
 }
 
@@ -282,7 +284,6 @@ func (f *Framework) GetAlertmanagerMetrics(ns, n string) (textparse.Parser, erro
 }
 
 func (f *Framework) CreateSilence(ns, n string) (string, error) {
-
 	var createSilenceResponse silence.PostSilencesOKBody
 
 	request := ProxyPostPod(
@@ -304,7 +305,6 @@ func (f *Framework) CreateSilence(ns, n string) (string, error) {
 // SendAlertToAlertmanager sends an alert to the alertmanager in the given
 // namespace (ns) with the given name (n).
 func (f *Framework) SendAlertToAlertmanager(ns, n string) error {
-
 	alerts := models.PostableAlerts{{
 		Alert: models.Alert{
 			GeneratorURL: "http://prometheus-test-0:9090/graph?g0.expr=vector%281%29\u0026g0.tab=1",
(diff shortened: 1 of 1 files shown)


test/framework/alertmanager.go
```diff
@@ -217,10 +217,7 @@ func amImage(version string) string {
 func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers int, forceEnableClusterMode bool) error {
 	var pollError error
 	err := wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
-		amStatus, err := f.GetAlertmanagerStatus(ns, name)
-		if err != nil {
-			return false, err
-		}
+		amStatus := f.GetAlertmanagerStatus(ns, name)
```
simonpasquier, line 220: As written in the issue, we should return false to retry later and the returned error needs to be saved in case the wait loop fails eventually. 

```suggestion
		amStatus, err := f.GetAlertmanagerStatus(ns, name)
		if err != nil {
			pollError := fmt.Errorf("failed to query Alertmanager: %s", err)
			return false, nil
		}
```

test/framework/alertmanager.go
```diff
@@ -257,19 +254,22 @@ func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers
 	return nil
 }
 
-func (f *Framework) GetAlertmanagerStatus(ns, n string) (models.AlertmanagerStatus, error) {
+func (f *Framework) GetAlertmanagerStatus(ns, n string) models.AlertmanagerStatus {
```
simonpasquier, line 257: `GetAlertmanagerStatus()` shouldn't swallow the error.

test/framework/alertmanager.go
```diff
@@ -348,10 +348,7 @@ func (f *Framework) GetSilences(ns, n string) (models.GettableSilences, error) {
 // string.
 func (f *Framework) WaitForAlertmanagerConfigToContainString(ns, amName, expectedString string) error {
 	err := wait.Poll(10*time.Second, time.Minute*5, func() (bool, error) {
-		amStatus, err := f.GetAlertmanagerStatus(ns, "alertmanager-"+amName+"-0")
-		if err != nil {
-			return false, err
-		}
+		amStatus := f.GetAlertmanagerStatus(ns, "alertmanager-"+amName+"-0")
```
simonpasquier, line 351: same here

test/framework/alertmanager.go
```diff
@@ -258,17 +261,22 @@ func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers
...
 
 func (f *Framework) GetAlertmanagerStatus(ns, n string) (models.AlertmanagerStatus, error) {
+
 	var amStatus models.AlertmanagerStatus
 	request := ProxyGetPod(f.KubeClient, ns, n, "/api/v2/status")
 	resp, err := request.DoRaw(context.TODO())
+
 	if err != nil {
+
+		fmt.Errorf("failed to get alertmanager status response: %v", err)
```
simonpasquier, line 271:
```suggestion
```

test/framework/alertmanager.go
```diff
@@ -258,17 +261,22 @@ func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers
...
 		return amStatus, err
 	}
 
 	if err := json.Unmarshal(resp, &amStatus); err != nil {
+
+		fmt.Errorf("failed to unmarshal alertmanager status response: %v", err)
```
simonpasquier, line 277:
```suggestion
```

test/framework/alertmanager.go
```diff
@@ -347,10 +355,14 @@ func (f *Framework) GetSilences(ns, n string) (models.GettableSilences, error) {
 // configuration via the Alertmanager's API and checks if it contains the given
 // string.
 func (f *Framework) WaitForAlertmanagerConfigToContainString(ns, amName, expectedString string) error {
+
+	var pollError error
```
simonpasquier, line 359: you'd need to modify L376 to report `pollErr` there (see L254 for example). Also it would make sense to set  `pollErr` when the returned config doesn't match with the expected value.

arghya88: Done...PTAL

test/framework/alertmanager.go
```diff
@@ -215,11 +215,14 @@ func amImage(version string) string {
 }
 
 func (f *Framework) WaitForAlertmanagerInitialized(ns, name string, amountPeers int, forceEnableClusterMode bool) error {
+
```
lilic, line 218: Nitpick: but we tend to not do a new line after function decleration, so this is not really needed.

arghya88: Let me fix it quickly

simonpasquier (commented):
All good for me. It can be merged on green.

lilic:
Thanks! 🎉

PR #3567, opened 2020-10-05
pkg/prometheus/statefulset: Exec requests in local probes

wking:
Reduce the number of processes in play by using [`exec`][1] to transition to the `curl`/`wget` command without creating a new process.  Besides increasing efficiency, this may reduce the chance of [leaking zombie processes][2], or at least reduce the number of processes leaked when an exec-probe zombificaiton occurs.

The `ListenLocal` stuff landed in ca15e870da (#1051), the conditional get/exec probe logic landed in 8011f6364c (#2763), and the conditional `wget`/`curl` logic landed in 5eb9cb53db (#2763).

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#exec
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1878772

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -556,7 +556,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, ruleConf
 		})
 	}
 
-	const localProbe = `if [ -x "$(command -v curl)" ]; then curl %s; elif [ -x "$(command -v wget)" ]; then wget -q -O /dev/null %s; else exit 1; fi`
+	const localProbe = `if [ -x "$(command -v curl)" ]; then exec curl %s; elif [ -x "$(command -v wget)" ]; then exec wget -q -O /dev/null %s; else exit 1; fi`
 
 	var readinessProbeHandler v1.Handler
 	if (version.Major == 1 && version.Minor >= 8) || version.Major == 2 {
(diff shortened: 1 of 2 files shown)


coderanger:
Zombie process leaks aren't a random thing but this will save some small bit of memory overhead so seems like a good thing :)

wking:
> Zombie process leaks aren't a random thing...

For a script like this without `trap` signal handling, I expect child processes might become zombies if the parent shell process is killed while the wget/curl child is running.  But obviously you won't get that child wget/curl zombie if the parent shell process is killed before or after the wget/curl child is running.  And now, with this PR, there will no longer be a child process to worry about.  [The bug][1] also had some zombie `sh` processes, so there must be something else going on, but I expect this change to remove the zombies which are `curl` processes in the reported bug.  And yeah, saving some overhead is nice too :).

I've pushed fdb2b3f3a -> 43d594d1d to update `TestListenLocal` with the new expected commands, hopefully this run will be green :crossed_fingers: 

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1878772

coderanger:
The root of the zombie problem with exec probes is a container command that doesn't react to SIGCHLD and/or doesn't call `wait(-1)` in a loop somewhere. We should check if Prometheus has such a handler, but if this isn't already a huge problem then presumably it does.

simonpasquier (commented):
:+1: thanks for this!

s-urbaniak:
nice, thank you!

Issue #3546, opened 2020-09-25
Generated copyright headers always use 2018 as copyright year


grdryn:
This came up in a code review with @lilic here: https://github.com/prometheus-operator/prometheus-operator/pull/3451#discussion_r493294089.

See [here](https://github.com/prometheus-operator/prometheus-operator/blob/master/.header#L1) for where the value comes from.

This means that when we run generation commands, the year is always set to 2018, even for newly-generated files.

Should we update this to 2020, meaning that all generated files will have 2020 in the copyright header? What about manually-created files?

dgrisonnet:
I'd say we should follow what Kubernetes is doing, aka removing the YEAR for the generated go files: https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.generatego.txt

As for the files created manually, we should continue with the usual mechanism to set the YEAR according to the one the file was created at.

PR #3569, opened 2020-10-07
ISSUE 3546 remove year from .header file

rothsa:
Removing the year based on the suggestion here: https://github.com/prometheus-operator/prometheus-operator/issues/3546#issuecomment-698958051

Would you also appreciate if we update the existing files based on the year the file was created (from git blame)?

cc @theatomicoption


brancz:
Thanks! If you want to try and trace the years back, that would be awesome!

PR #3577, opened 2020-10-08
pkg/prometheus: ensure minimal version for target limits

simonpasquier:
The target_limit configuration field is only supported since Prometheus
v2.21.0. In the initial commit implementing target limits for the
Prometheus resource, we forgot to check which Prometheus version is
being deployed.


brancz (approved):

brancz:
lgtm on green

PR #3578, opened 2020-10-08
Discard invalid TLS configurations

### LINKED ISSUE #3550: The Prometheus operator should discard invalid TLS configurations
**What happened?**

I created a service monitor with a `cert` field but no `keySecret`. The operator generated the new configuration but Prometheus failed to reload because it expects that both cert and key are [defined](https://github.com/prometheus/common/blob/317b7b125e8fddda956d0c9574e5f03f438ed5bc/config/http_config.go#L344-L354).

Prometheus logs:

```
level=info ts=2020-09-29T13:28:36.663Z caller=main.go:833 msg="Completed loading of configuration file" filename=/etc/prometheus/config_out/prometheus.env.yaml
level=error ts=2020-09-29T13:28:38.116Z caller=manager.go:188 component="scrape manager" msg="error creating new scrape pool" err="error creating HTTP client: client cert file \"/etc/prometheus/certs/secret_default_tls-materials_cert\" specified without client key file" scrape_pool=default/example-app/0
```

Note that if the Prometheus container crashes, it will fail to start again.

Finally the issue isn't limited to the cert/key case described here. The same thing happens if one provides an invalid certificate, key or CA.

The operator should have rejected the invalid TLS configuration.

See manifests below.

**Environment**

master

N/A

* Kubernetes cluster kind:

N/A

* Manifests:

```
apiVersion: v1
data:
  cert: Zm9vYmFy
kind: Secret
metadata:
  name: tls-materials
type: Opaque
---
kind: ServiceMonitor
metadata:
  name: example-app
spec:
  endpoints:
  - port: web
    tlsConfig:
      cert:
        secret:
          key: cert
          name: tls-materials
  selector:
    matchLabels:
      app: example-app
```

* Prometheus Operator Logs:

N/A

dgrisonnet:
TLS misconfiguration makes the prometheus container crash. To avoid
that, we should discard the invalid configurations right away.
The possible invalid configurations are the following ones:
- cert without key
- key without cert
- cert with wrong key
- ~~ca with unsigned cert~~

There is a limitation with file-based assets for the later case as we can't access their data.

Fixes #3550


pkg/prometheus/asset_store.go
```diff
@@ -109,59 +112,72 @@ func assetKeyFunc(obj interface{}) (string, error) {
...
-			ca  string
-			err error
-		)
-
-		switch {
-		case tlsConfig.CA.Secret != nil:
-			ca, err = a.getSecretKey(ctx, ns, *tlsConfig.CA.Secret)
+	err := tlsConfig.Validate()
+	if err != nil {
+		return err
```
s-urbaniak, line 117: nit: let's wrap the error, i think we are using the `pkg/errors` package here, so we can use something like `return errors.Wrap(err, "TLS configuration validation failed")`.

pkg/apis/monitoring/v1/types.go
```diff
@@ -1010,6 +1016,12 @@ func (c *TLSConfig) Validate() error {
 		return &TLSConfigValidationError{"tls config can not both specify KeyFile and KeySecret"}
 	}
 
+	if c.CertFile != "" && (c.KeyFile == "" || c.KeySecret == nil) {
+		return &TLSConfigValidationError{"client cert specified without client key"}
+	} else if c.KeyFile != "" && (c.CertFile == "" || c.Cert == (SecretOrConfigMap{})) {
```
s-urbaniak, line 1021: stylistic nit: the `else if` is not necessary. it is more readable if you simply add a newline with a new `if` statement.

s-urbaniak: concretely:
```
	if c.Cert != (SecretOrConfigMap{}) && c.KeySecret == nil {
		return &TLSConfigValidationError{"client cert specified without client key"}
	}

	if c.KeySecret != nil && c.Cert == (SecretOrConfigMap{}) {
		return &TLSConfigValidationError{"client key specified without client cert"}
	}

	return nil
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -970,6 +970,12 @@ func (c *SafeTLSConfig) Validate() error {
 		}
 	}
 
+	if c.Cert != (SecretOrConfigMap{}) && c.KeySecret == nil {
+		return &TLSConfigValidationError{"client cert specified without client key"}
+	} else if c.KeySecret != nil && c.Cert == (SecretOrConfigMap{}) {
+		return &TLSConfigValidationError{"client key specified without client cert"}
+	}
```
s-urbaniak, line 977: see below

pkg/apis/monitoring/v1/types.go
```diff
@@ -970,6 +970,12 @@ func (c *SafeTLSConfig) Validate() error {
 		}
 	}
 
+	if c.Cert != (SecretOrConfigMap{}) && c.KeySecret == nil {
```
s-urbaniak, line 973: while we are here: does it make sense to add a unit test for `*SafeTLSConfig.Validate` here in this package? I think we are missing one.

pkg/prometheus/asset_store.go
```diff
@@ -109,59 +112,72 @@ func assetKeyFunc(obj interface{}) (string, error) {
...
-	if tlsConfig.CA != (monitoringv1.SecretOrConfigMap{}) {
-		var (
-			ca  string
-			err error
-		)
-
-		switch {
-		case tlsConfig.CA.Secret != nil:
-			ca, err = a.getSecretKey(ctx, ns, *tlsConfig.CA.Secret)
+	err := tlsConfig.Validate()
```
s-urbaniak, line 115: this method looks much cleaner now, nice! :tada:

s-urbaniak:
so far great stuff, leaving another set of eyes to @prometheus-operator/prometheus-operator-reviewers as this touches TLS verification code.

pkg/prometheus/asset_store.go
```diff
@@ -261,3 +287,30 @@ func (a *assetStore) getSecretKey(ctx context.Context, namespace string, sel v1.
...
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse certificate")
+	}
+
+	opts := x509.VerifyOptions{
+		Roots: roots,
+	}
+
+	if _, err := cert.Verify(opts); err != nil {
```
simonpasquier, line 311: I'm not sure we should go that deep in the verification, especially since the user may ask to skip the TLS verification in the pod/service monitor definition. I would only check that the CA PEM is valid.

pkg/prometheus/operator.go
```diff
@@ -1630,15 +1630,18 @@ func (c *Operator) selectServiceMonitors(ctx context.Context, p *monitoringv1.Pr
 			smKey := fmt.Sprintf("serviceMonitor/%s/%s/%d", sm.GetNamespace(), sm.GetName(), i)
 
 			if err = store.addBearerToken(ctx, sm.GetNamespace(), endpoint.BearerTokenSecret, smKey); err != nil {
+				level.Error(c.logger).Log("error", err, "namespace", sm.GetNamespace(), "servicemonitor", sm.GetName())
```
simonpasquier, line 1633: I wouldn't log anything here since we're already logging a warning after we break from the for loop.

pkg/prometheus/operator.go
```diff
@@ -1630,15 +1630,18 @@ func (c *Operator) selectServiceMonitors(ctx context.Context, p *monitoringv1.Pr
...
 
 			if err = store.addBasicAuth(ctx, sm.GetNamespace(), endpoint.BasicAuth, smKey); err != nil {
+				level.Error(c.logger).Log("error", err, "namespace", sm.GetNamespace(), "servicemonitor", sm.GetName())
 				break
 			}
 
 			if endpoint.TLSConfig != nil {
-				if err = store.addSafeTLSConfig(ctx, sm.GetNamespace(), endpoint.TLSConfig.SafeTLSConfig); err != nil {
+				if err = store.addTLSConfig(ctx, sm.GetNamespace(), endpoint.TLSConfig); err != nil {
+					level.Error(c.logger).Log("error", err, "namespace", sm.GetNamespace(), "servicemonitor", sm.GetName())
```
simonpasquier, line 1644: ditto

pkg/prometheus/operator.go
```diff
@@ -1718,15 +1721,18 @@ func (c *Operator) selectPodMonitors(ctx context.Context, p *monitoringv1.Promet
 			pmKey := fmt.Sprintf("podMonitor/%s/%s/%d", pm.GetNamespace(), pm.GetName(), i)
 
 			if err = store.addBearerToken(ctx, pm.GetNamespace(), endpoint.BearerTokenSecret, pmKey); err != nil {
+				level.Error(c.logger).Log("error", err, "namespace", pm.GetNamespace(), "podmonitor", pm.GetName())
```
simonpasquier, line 1724: ditto

pkg/prometheus/operator.go
```diff
@@ -1718,15 +1721,18 @@ func (c *Operator) selectPodMonitors(ctx context.Context, p *monitoringv1.Promet
...
 				break
 			}
 
 			if err = store.addBasicAuth(ctx, pm.GetNamespace(), endpoint.BasicAuth, pmKey); err != nil {
+				level.Error(c.logger).Log("error", err, "namespace", pm.GetNamespace(), "podmonitor", pm.GetName())
```
simonpasquier, line 1729: ditto

pkg/prometheus/operator.go
```diff
@@ -1718,15 +1721,18 @@ func (c *Operator) selectPodMonitors(ctx context.Context, p *monitoringv1.Promet
...
 				break
 			}
 
 			if endpoint.TLSConfig != nil {
 				if err = store.addSafeTLSConfig(ctx, pm.GetNamespace(), endpoint.TLSConfig.SafeTLSConfig); err != nil {
+					level.Error(c.logger).Log("error", err, "namespace", pm.GetNamespace(), "podmonitor", pm.GetName())
```
simonpasquier, line 1735: ditto

simonpasquier (commented):
Nice!

pkg/prometheus/asset_store.go
dgrisonnet: I had forgotten to check if the TLS verification was skipped before calling this function, it should be good now.

If we don't discard unsigned certificates right away, the invalid configuration will be hand over to Prometheus which might discard it at some point. To me, this ends up being the same issue as with the invalid keys/certificates.

simonpasquier: Hmm, we shouldn't verify the client certificate/key against the CA: client/key are used for client authentication (e.g. let the server authenticate Prometheus) while the CA is used by Prometheus to verify the certificate presented by the server.

dgrisonnet: Makes sense, thank you for the explanation.

pkg/apis/monitoring/v1/types.go
```diff
@@ -998,19 +1006,40 @@ func (e *TLSConfigValidationError) Error() string {
...
+			return &TLSConfigValidationError{"tls config CA is invalid"}
+		}
 	}
 
 	if c.Cert != (SecretOrConfigMap{}) && c.CertFile != "" {
-		return &TLSConfigValidationError{"tls config can not both specify CertFile and Cert"}
+		if c.CertFile != "" {
+			return &TLSConfigValidationError{"tls config can not both specify CertFile and Cert"}
+		}
+		if err := c.CA.Validate(); err != nil {
```
simonpasquier, line 1022:
```suggestion
		if err := c.Cert.Validate(); err != nil {
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -998,19 +1006,40 @@ func (e *TLSConfigValidationError) Error() string {
...
+	if c.CA != (SecretOrConfigMap{}) {
+		if c.CAFile != "" {
+			return &TLSConfigValidationError{"tls config can not both specify CAFile and CA"}
+		}
+		if err := c.CA.Validate(); err != nil {
+			return &TLSConfigValidationError{"tls config CA is invalid"}
+		}
 	}
 
 	if c.Cert != (SecretOrConfigMap{}) && c.CertFile != "" {
```
simonpasquier, line 1018:
```suggestion
	if c.Cert != (SecretOrConfigMap{}) {
```

pkg/prometheus/asset_store_test.go
```diff
@@ -26,6 +26,45 @@ import (
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 )
 
+var (
+	caPEM = `-----BEGIN CERTIFICATE-----
```
simonpasquier, line 30: It seems that this certificate is only valid until `Oct  8 12:52:00 2021 GMT`. Can you extend its lifetime to something very far in the future?

pkg/prometheus/asset_store_test.go
```diff
@@ -26,6 +26,45 @@ import (
...
+HwYDVQQKDBhJbnRlcm5ldCBXaWRnaXRzIFB0eSBMdGQwXDANBgkqhkiG9w0BAQEF
+AANLADBIAkEA6Ms44U9YF87b8H4Xeg0WG9HdjG+8YmVtn7ADI07JuUnRiC+s/r0D
+6Y9YoUZS4SYCF9EL1KScuqI886F1pF29rwIDAQABo1MwUTAdBgNVHQ4EFgQU8Aj6
+J+hDw0JYZCCpy354WcczbD8wHwYDVR0jBBgwFoAU8Aj6J+hDw0JYZCCpy354Wccz
+bD8wDwYDVR0TAQH/BAUwAwEB/zANBgkqhkiG9w0BAQsFAANBALC/hvicZRBdwzNY
+P00PA2vNN40XYRLB+MhuN+IyaOU3sR9ckLJqDSLtlXlO0BM8985MGOgMuRHTtOAv
+mB8zmAA=
+-----END CERTIFICATE-----`
+
+	certPEM = `-----BEGIN CERTIFICATE-----
```
simonpasquier, line 44: ditto

pkg/prometheus/asset_store_test.go
dgrisonnet: Thank you for checking, I've added a hundred years to it.

simonpasquier (approved):
LGTM. Lets have another pair of :eyes: on it...

s-urbaniak:
This looks very clean to me, LGTM too :+1: 
Thank you @dgrisonnet !

PR #3592, opened 2020-10-15
Add more descriptive binary version information

mszostok:
**Description**

This pull request adds more descriptive version information. 

Proposed changes:
- Handling version in the same way in all CLI binaries also in the `prometheus-config-reloader` that uses not idiomatic kingpin framework
- Prometheus Operator has exactly the same behavior as Prometheus:
    - Print in logs msg with Info and build context
    ```
    level=info ts=2020-10-14T21:00:58.8016435Z caller=main.go:227 msg="Starting Prometheus Operator" version="(version=0.42.2, branch=master, revision=4f2cce69dccf1580068ccf4c7fbd1a8ad83257ea)"
    level=info ts=2020-10-14T21:00:58.8017066Z caller=main.go:228 build_context="(go=go1.14.9, user=mszostok, date=20201014-19:21:54)"
    ```
    - Dump the same build_info metrics
- Split version functions into “domain” files

The implemented functionality is fully compatible with Prometheus but I’ve added new features that allow users to:
- Print full version with all details
  ```
  $ ./prometheus-config-reloader --version
  prometheus-config-reloader, version 0.42.2 (branch: refs/tags/v0.42.2, revision: 4f2cce69dccf1580068ccf4c7fbd1a8ad83257ea)
    build user:       mszostok
    build date:       20201014-19:20:25
    go version:       go1.14.9
  ```
- Print short version (just the version number that can be used on CI for version validation)
  ```
  $ ./prometheus-config-reloader --short-version
  0.42.2 
  ```

**TODO**
- Add documentation about versions
- Change the pull request template to include sth like?
  ```
  * Binary version:
  
	  insert output of `<binary> --version` here
  ```

**Testing**

I configured the GitHub Action pipeline on my repo and here are the docker images that you can download and check:

  #### Master:
  GitHub  Action docker build: https://github.com/mszostok/prometheus-operator/runs/1256326107?check_suite_focus=true#step:4:8

  ```
  docker run quay.io/mszostok/prometheus-config-reloader:master --short-version
  docker run quay.io/mszostok/prometheus-config-reloader:master --version

  docker run quay.io/mszostok/prometheus-operator:master --short-version
  docker run quay.io/mszostok/prometheus-operator:master --version
  ```

  #### Release:

  GitHub Action docker build: https://github.com/mszostok/prometheus-operator/runs/1256108332?check_suite_focus=true#step:5:12

  ```
  docker run quay.io/mszostok/prometheus-config-reloader:v0.42.2 --short-version
  docker run quay.io/mszostok/prometheus-config-reloader:v0.42.2 --version

  docker run quay.io/mszostok/prometheus-operator:v0.42.2 --short-version
  docker run quay.io/mszostok/prometheus-operator:v0.42.2 --version
  ```

Fixes https://github.com/prometheus-operator/prometheus-operator/issues/3527


cmd/prometheus-config-reloader/main.go
```diff
@@ -103,11 +103,18 @@ func main() {
 	reloadURL := app.Flag("reload-url", "reload URL to trigger Prometheus reload on").
 		Default("http://127.0.0.1:9090/-/reload").URL()
 
+	version.RegisterIntoKingpinFlags(app)
```
simonpasquier, line 106: could you also update the starting log at L146?

simonpasquier (commented):
Very nice! I don't feel strongly that we duplicate code that already exists in [`github.com/prometheus/common/version`](https://pkg.go.dev/github.com/prometheus/common@v0.14.0/version). Is it only because of the new `--short-version` option?

mszostok:
Hi @simonpasquier 
Thanks for, so quick review!

I will just share my thoughts on why I copied some part of the `github.com/prometheus/common/version`:
- IMO the software version management is more internal, so vendoring external source for doing that was for me somehow not a way to go. For Prometheus, I can understand that because all those stuff are under a single GitHub organization, so having a common library is reasonable.
- In the current code I grep with `github.com/prometheus/common/version` and didn't find any imports, so I didn't want to add overhead with maintaining another dependency, as sometimes owning sth is easier instead of watch for changes.
- It's not possible to add new fields, and I wanted to propose to add new fields about compiler and platform, see:
https://github.com/mszostok/codeowners-validator/blob/master/pkg/version/version.go#L40-L41
  But they are not, so important but gives a more debug info

but if going with the Prometheus common lib is a way to go that's even better. I refactored the proposed solution to vendor the` github.com/prometheus/common/version` and only extend that with CLI helpers and code coverage, so even if someone will bump the `github.com/prometheus/common/version` dependency, unit-tests will ensure that the contract was not broken and if some adjustments are needed or not.

Now you can check which version is preferred by switching the commits, before merging I can squash them to a single one :)

I also rebuild the master with the newest changes: https://github.com/mszostok/prometheus-operator/runs/1260477686?check_suite_focus=true#step:4:8 (all green and versioning is still working)

cmd/prometheus-config-reloader/main.go
mszostok: fixed :)

simonpasquier:
Thanks a lot for offering the 2 options! Personally I feel that reusing `github.com/prometheus/common/version` is better from a maintenance point of view (the maintainers of this repository have a good understanding of what happens in Prometheus upstream so if the package breaks its interface, we should be able to adapt). But lets see what other @prometheus-operator/prometheus-operator-reviewers think.

brancz:
Agreed with @simonpasquier

simonpasquier:
Thanks!

PR #3598, opened 2020-10-16
Support for Pod Topology Spread Constraints for Prometheus, Alertmanager, and ThanosRuler CRDs

### LINKED ISSUE #3595: Support for Pod Topology Spread Constraints for Prometheus and Alertmanager CRDs
**What is missing?**
The Prometheus and Alertmanager CRDs support Pod affinities and anti-affinities as well as Node affinities, but not `topologySpreadConstraints`, which is a new feature (see https://kubernetes.io/blog/2020/05/introducing-podtopologyspread/ and https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/)

**Why do we need it?**
Having an even Pod spread across zones/regions for the StatefulSets generated by the Prometheus and Alertmanager CRDs would be useful for HA setups and is more flexible than using something like Pod anti-affinities with `topology.kubernetes.io/zone` for the `topologyKey`.

AshleyDumaine:
This PR resolves #3595 

I forgot to mention ThanosRulers in the issue, but since that CRD also supports Pod and Node affinities, I figured it would make sense to add topologySpreadConstraints to the resulting StatefulSet as well.


brancz (approved):
Awesome, lgtm! Thank you so much for taking care of this!

PR #3600, opened 2020-10-19
Documentation: Aligned serviceMonitor usage inline with getting-started

nrchakradhar:
Aligned serviceMonitor usage inline with getting-started


paulfantom (approved):
lgtm

lilic:
Thanks for your contribution!! I just edited the title to be more in line with your description, this makes it easier for us when we write up the CHANGELOG. :)

PR #3604, opened 2020-10-20
pkg/alertmanager: skip invalid PagerDuty configs

simonpasquier:
Instead of silently discarding invalid PagerDuty configurations when
generating the whole Alertmanager configuration, we reject those upfront
so it can be detected with the `prometheus_operator_managed_resources`
metric.

cc @grdryn


paulfantom (approved):
lgtm

PR #3605, opened 2020-10-20
Update Prometheus and Thanos versions

simonpasquier:


brancz (approved):

simonpasquier:
For some unexpected reason, this breaks the remote write e2e tests. I'll need to dig further...

s-urbaniak (approved):
lgtm, thank you!

s-urbaniak:
@simonpasquier out of curiosity: what was the rationale for removing openssl in favor of Go, not that I suggest to go back but I am literally curious :-)

simonpasquier:
> what was the rationale for removing openssl in favor of Go, not that I suggest to go back but I am literally curious :-)

1) I couldn't find the exact combination of `openssl` CLI flags that would generate the certificates with all required fields.
2) it should be agnostic of the platform (e.g. Linux vs Mac).

We could even take it one step further: remove these certificates from the repository and generate one-time certificates before running the tests.

PR #3616, opened 2020-10-21
Move duplicated asset code to the assets package

dgrisonnet:
The asset_store code was duplicated in the prometheus and alertmanager
package. Some recent changes added TLS validation to the Prometheus
asset store only. To avoid diverging anymore and allow the alertmanager
code to benefit from the recent additions a refactor was necessary.


pkg/alertmanager/amcfg.go
```diff
@@ -166,12 +167,12 @@ func (c alertmanagerConfig) String() string {
 type configGenerator struct {
 	logger  log.Logger
 	kclient kubernetes.Interface
-	store   *assetStore
+	store   *assets.AssetStore
```
simonpasquier, line 170: since `AssetStore` lives in its own package, it can be renamed `Store`.

pkg/alertmanager/amcfg.go
```diff
@@ -166,12 +167,12 @@ func (c alertmanagerConfig) String() string {
...
 }
 
-func newConfigGenerator(logger log.Logger, kclient kubernetes.Interface, store *assetStore) *configGenerator {
+func newConfigGenerator(logger log.Logger, kclient kubernetes.Interface, store *assets.AssetStore) *configGenerator {
```
simonpasquier, line 173: We can simplify and require a non `nil` store here. `kclient` isn't used anywhere I think so it can be removed from the `configGenerator` struct.

```suggestion
func newConfigGenerator(logger log.Logger, store *assets.AssetStore) *configGenerator {
```

simonpasquier:
This needs a rebase again.

dgrisonnet:
@simonpasquier done

simonpasquier (approved):
:+1:

PR #3617, opened 2020-10-22
Include EnforcedSampleLimit as a metric  as mentioned in (#3400)

### LINKED ISSUE #3400: Expose EnforcedSampleLimit as metric
EnforcedSampleLimit defines global limit on number of scraped samples and overrides any existing limits set for PodMonitor and ServiceMonitor.

**Why do we need it?**
Having this value in a metric allows us to create an alert to warn the admin that they are getting close to the limit, something like KubePersistentVolumeFillingUp alert. This means admin does not need to create this alert but it comes out of the box with the Prometheus operator mixin for example. This would be a warning level alert, for critical one there is already TargetDown alerts.

nrchakradhar:
Included EnforcedSampleLimit also as a metric for Prometheus when specified.

Closes #3400

### DIFF
--- a/pkg/prometheus/collector.go
+++ b/pkg/prometheus/collector.go
@@ -30,6 +30,16 @@ var (
 		}, nil,
 	)
 )
+var (
+	descPrometheusEnforcedSampleLimit = prometheus.NewDesc(
+		"prometheus_enforced_sample_limit",
+		"Global limit on number of scraped samples. Overrides any existing limits set for PodMonitor and ServiceMonitor.",
+		[]string{
+			"namespace",
+			"name",
+		}, nil,
+	)
+)
 
 type prometheusCollector struct {
 	stores []cache.Store
@@ -46,6 +56,7 @@ func NewPrometheusCollectorForStores(s ...cache.Store) *prometheusCollector {
 // Describe implements the prometheus.Collector interface.
 func (c *prometheusCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descPrometheusSpecReplicas
+	ch <- descPrometheusEnforcedSampleLimit
 }
 
 // Collect implements the prometheus.Collector interface.
@@ -63,4 +74,8 @@ func (c *prometheusCollector) collectPrometheus(ch chan<- prometheus.Metric, p *
 		replicas = float64(*p.Spec.Replicas)
 	}
 	ch <- prometheus.MustNewConstMetric(descPrometheusSpecReplicas, prometheus.GaugeValue, replicas, p.Namespace, p.Name)
+	// Include EnforcedSampleLimit in metrics if set in PrometheusSpec
+	if p.Spec.EnforcedSampleLimit != nil {
+		ch <- prometheus.MustNewConstMetric(descPrometheusEnforcedSampleLimit, prometheus.GaugeValue, float64(*p.Spec.EnforcedSampleLimit), p.Namespace, p.Name)
+	}
 }
--- a/pkg/prometheus/collector.go
+++ b/pkg/prometheus/collector.go
@@ -30,7 +30,7 @@ var (
 		}, nil,
 	)
 	descPrometheusEnforcedSampleLimit = prometheus.NewDesc(
-		"prometheus_operator_enforced_sample_limit",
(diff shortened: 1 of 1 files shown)


pkg/prometheus/collector.go
```diff
@@ -30,6 +30,16 @@ var (
 		}, nil,
 	)
 )
+var (
+	descPrometheusEnforcedSampleLimit = prometheus.NewDesc(
+		"prometheus_enforced_sample_limit",
```
simonpasquier, line 35:
```suggestion
		"prometheus_operator_enforced_sample_limit",
```

pkg/prometheus/collector.go
```diff
@@ -30,6 +30,16 @@ var (
...
+		"Global limit on number of scraped samples. Overrides any existing limits set for PodMonitor and ServiceMonitor.",
```
simonpasquier, line 36:
```suggestion
		"Global limit on the number of scraped samples per scrape target.",
```

pkg/prometheus/collector.go
```diff
@@ -63,4 +74,8 @@ func (c *prometheusCollector) collectPrometheus(ch chan<- prometheus.Metric, p *
 		replicas = float64(*p.Spec.Replicas)
 	}
 	ch <- prometheus.MustNewConstMetric(descPrometheusSpecReplicas, prometheus.GaugeValue, replicas, p.Namespace, p.Name)
+	// Include EnforcedSampleLimit in metrics if set in PrometheusSpec
```
simonpasquier, line 77:
```suggestion
	// Include EnforcedSampleLimit in metrics if set in PrometheusSpec.
```

pkg/prometheus/collector.go
```diff
@@ -30,6 +30,16 @@ var (
 		}, nil,
 	)
 )
+var (
```
simonpasquier, line 33: you can combine both declarations:

```go
var (
	descPrometheusSpecReplicas = prometheus.NewDesc(
            ...
        )
	descPrometheusEnforcedSampleLimit = prometheus.NewDesc(
            ...
        )
)
```

simonpasquier (commented):
Nice!

pkg/prometheus/collector.go
```diff
@@ -29,6 +29,14 @@ var (
 			"name",
 		}, nil,
 	)
+	descPrometheusEnforcedSampleLimit = prometheus.NewDesc(
+		"prometheus_operator_enforced_sample_limit",
+		"Global limit on the number of scraped samples per scrape target.",
+		[]string{
+			"namespace",
+			"name",
```
lilic, line 37: I believe this is Prometheus name, wonder if this is descriptive enough of a label name? @nrchakradhar @simonpasquier wdyt?

pkg/prometheus/collector.go
```diff
@@ -63,4 +72,8 @@ func (c *prometheusCollector) collectPrometheus(ch chan<- prometheus.Metric, p *
 		replicas = float64(*p.Spec.Replicas)
 	}
 	ch <- prometheus.MustNewConstMetric(descPrometheusSpecReplicas, prometheus.GaugeValue, replicas, p.Namespace, p.Name)
+	// Include EnforcedSampleLimit in metrics if set in PrometheusSpec.
```
lilic, line 75:
```suggestion
	// Include EnforcedSampleLimit in metrics if set in Prometheus object.
```

lilic (commented):
Nice thanks for the contribution!! Sorry for the delay in response on the issue was OOF.

pkg/prometheus/collector.go
simonpasquier: This is consistent with the labels of `prometheus_operator_spec_replicas` above. Maybe the metric name should be `prometheus_operator_spec_enforced_sample_limit` after all?

lilic: I think `prometheus_operator_prometheus_enforced_sample_limit ` maybe?

nrchakradhar: This name looks more reasonable. Changed it.

lilic (commented):
lgtm thanks!! 🎉 

letting @simonpasquier have another look.

simonpasquier (approved):

Issue #3535, opened 2020-09-23
Add Slack Receiver type to AlertmanagerConfig CRD


grdryn:
**What is missing?**

In #3451 we added the AlertmanagerConfig CRD, with just the PagerDuty receiver type, meaning that alert notifications can only be routed to PagerDuty. Alertmanager supports several other Receiver types, with this being one (other issues will be created for the others).

**Why do we need it?**

We should be able to define all of the Alertmanager receivers in AlertmanagerConfig CRs.

**Anything else we need to know?**:

The proposal for AlertmanagerConfig CRD is [here](https://docs.google.com/document/d/1aVVttvocop8zNezwrNFbHh_HKRW_fhJUZqhmrr8fGK0/edit#heading=h.5ua2yyj2xm4h), and it includes some information on this receiver type.

The Alertmanager config docs [here](https://prometheus.io/docs/alerting/latest/configuration/) show how this receiver is expected to be in the resulting Alertmanager config.

omegas27:
I am up for implementing this one when https://github.com/prometheus-operator/prometheus-operator/pull/3451 is merged

PR #3618, opened 2020-10-23
Add Slack Receiver type to AlertmanagerConfig CRD

omegas27:
Closes https://github.com/prometheus-operator/prometheus-operator/issues/3535


pkg/alertmanager/amcfg.go
```diff
@@ -390,6 +452,152 @@ func (cg *configGenerator) convertWebhookConfig(ctx context.Context, in monitori
...
+
+	if in.LinkNames != nil {
+		out.LinkNames = *in.LinkNames
+	}
+
+	out.MrkdwnIn = in.MrkdwnIn
+
+	var actions []slackAction
+	if len(in.Actions) > 0 {
+		actions = make([]slackAction, 0)
```
grdryn, line 538: Suggestion for these kinds of loops: I think this will save some computational effort, given that you know wht the size of the slice will be upfront. This might mean that you'd need to say e.g. `actions[i] = action`, rather than `append(actions, action)` later though.

```suggestion
	if l := len(in.Actions); l > 0 {
		actions = make([]slackAction, l)
```

grdryn (commented):
Hey @omegas27 thanks for doing this one! I've left a couple of inline suggestions. It also seems that the alertmanager e2e test is failing here -- let me know if you need any help getting that to work again.

pkg/alertmanager/amcfg.go
omegas27: Thanks :)

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -98,6 +99,118 @@ type PagerDutyConfig struct {
...
+	Style        string                  `json:"style,omitempty"`
+	Name         string                  `json:"name,omitempty"`
+	Value        string                  `json:"value,omitempty"`
+	ConfirmField *SlackConfirmationField `json:"confirm,omitempty"`
+}
+
+// Validate ensures SlackAction is valid
+func (sa *SlackAction) Validate() error {
+	if sa.Type == "" {
+		return errors.New("missing type in Slack action configuration")
```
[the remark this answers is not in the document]
omegas27: Thanks a lot for your review :)
Is there a way to test the operator locally so I could debug the current testing issue otherwise, I could use some help to unlock the e2e situation :D

grdryn:
@omegas27 Yes, you can run the e2e tests locally. There might be easier ways, but here's one:

## Running the tests

1. Install and run minikube
2. Build the operator image for your version & make it available somewhere (e.g. quay.io)
3. Run `make test-e2e` with the `REPO` and `TAG` environment variables set for your image. E.g.:

```
REPO=quay.io/omeas27/prometheus-operator TAG=add-slack-receiver-type make test-e2e
```

You can narrow the set of tests that will be run by using additional environment variables. E.g. the above would change to:

```
EXCLUDE_PROMETHEUS_TESTS=true EXCLUDE_THANOS_TESTS=true REPO=quay.io/omeas27/prometheus-operator TAG=add-slack-receiver-type make test-e2e
```

## Using a debugger with the tests

You'll notice that what `make test-e2e` does (after some pre-requisite cert stuff) is run `go test` with a bunch of arguments. You can do what you wish with that for testing, such as by changing the command to run the tests with a debugger if you wish, or narrowing further down to specific tests, or whatever. I usually use `dlv` for debugging (with the gdlv frontend), so I'd change the command to something like this, and run it in one terminal:

```diff
- go test -timeout 55m -v ./test/e2e/  --kubeconfig=/var/home/gryan/.kube/config --operator-image=quay.io/grdryn/prometheus-operator:mytag -count=1
+ dlv test --headless --listen [::]:2345 -- -timeout 55m -v ./test/e2e/ --kubeconfig=/var/home/gryan/.kube/config --operator-image=quay.io/grdryn/prometheus-operator:mytag -count=1
```

Then connect to it from gdlv, by running the following in another terminal:
```
gdlv connect [::]:2345
```

omegas27:
@grdryn Thank you that really helps :D

pkg/alertmanager/operator.go
```diff
@@ -927,6 +927,24 @@ func checkAlertmanagerConfig(ctx context.Context, amc *monitoringv1alpha1.Alertm
 			}
 		}
 
+		for j, sConfig := range receiver.SlackConfigs {
```
nrchakradhar, line 930: A request: 
Can we have this function checkAlertmanagerConfig be refactored a bit to call two funcs checkAlertManagerRecievers and checkAlertManagerRoutes?

In checkAlertManagerReceivers, each inner for loop can be a separate smaller func.

Feel free to ignore this comment if its not suitable.

lilic:
Hey @omegas27 can you please rebase, we can then have another round of reviews, thanks! :)

omegas27:
@lilic The rebase is done :)

pkg/alertmanager/operator.go
omegas27: I think splitting the check for the receivers makes sense but in my opinion, this should be done in another PR dedicated to that change. I can work on it next week

omegas27: @nrchakradhar I split the loop in smaller functions as requested :)

grdryn:
@omegas27 thanks for incorporating my suggestions. I took another look over this now, and it looks good to me! :+1:

lilic (approved):
Thanks!

lgtm

PR #3621, opened 2020-10-25
Remove pkg/errors dep from apis/monitoring module

grdryn:
I think this dependency is usually not needed any longer, given that the standard library has the likes of errors.New, and fmt.Errorf.

My understanding is that apis/monitoring is a standalong module now to make it easier for other projects to import it, with a minimal set of transitive dependecies to worry about. If I'm correct, then github.com/pkg/errors is an easy one to keep out.

I noticed this while catching up with the recent changes around AlertManagerConfigs, so it was only recently introduced in #3593.

### DIFF
--- a/pkg/apis/monitoring/go.mod
+++ b/pkg/apis/monitoring/go.mod
@@ -3,7 +3,6 @@ module github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring
 go 1.14
 
 require (
-	github.com/pkg/errors v0.9.1
 	k8s.io/api v0.18.3
 	k8s.io/apimachinery v0.18.3
 )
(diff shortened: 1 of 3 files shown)


coderanger:
This does mean not using Wrap/Wrapf since those are not in the stdlib.

grdryn:
@coderanger I think that case is covered by `fmt.Errorf` with the (relatively new) `%w` verb in the format specifier, is it not? https://golang.org/pkg/fmt/#Errorf

In any case, it seems like those Wrap/Wrapf functions aren't used in the `github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring` module, so I still think this is a good candidate to change. If you wanted to drop that dependency for the main `github.com/prometheus-operator/prometheus-operator` module, then making sure the stdlib covered all the cases that are used (I think it would), would definitely be something to consider at that point :+1:

s-urbaniak:
@grdryn thank you a lot for the contribution! This lgtm from my side, plus an after-thought: Does it make sense to sweep through the operator code base too and remove pkg/errors for good everywhere? If the @prometheus-operator/prometheus-operator-reviewers agree we can create a `help-wanted` issue to address this.

grdryn:
bq. Does it make sense to sweep through the operator code base too and remove pkg/errors for good everywhere?

It's definitely something that could be considered. :slightly_smiling_face:  I'm fairly sure that module isn't needed any longer with current versions of Go, and the project README states that it's in "maintenance mode". A couple of things to note about removing it from the rest of this project (lessons learned from doing it elsewhere):

1. There's a good chance that on an active/popular project like this, for a period of time after making the change, there will likely be active WIP/pull requests that will still refer to the old depenency & cause it to be pulled back in
2. You may have other depenencies that still pull it in as a transitive dependency, so it may still show up in your dependency list

I do think it's a good thing to do though -- the stdlib version may get future improvements, whereas it sounds like the old module will not really.

Issue #3517, opened 2020-09-19
prometheus-operator crash on a Probe 


kehao95:
**What happened?**
I didn't fully understand the [API of Probe](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#probe) especially the difference between `prober` and `targets`  so I decided to try it out and I composed an (may be invalid) probe.yaml and applied. 
prometheus-operator crashed when I applied the `Probe.yaml`

**Did you expect to see something different?**

Some error output telling that was an invalid `probe.spec`
Or at least don't crash prometheus-operator.

**How to reproduce it (as minimally and precisely as possible)**:

```yaml
# kubectl apply -f test-probe.yaml
apiVersion: monitoring.coreos.com/v1
kind: Probe
metadata: 
  namespace: monitoring
  name: test-probe
spec: 
  jobName: test-prob-name
  prober:
    url: https://www.example.com # was using my app url
    scheme: http
    path: /v1/health
  interval: 15s
```
delete the Probe then the prometheus-operator can run again

**Environment**

* Prometheus Operator version:

    I've tried `v0.41.1` and `v0.42.0` so far. They all crash on this.

* Kubernetes version information:

    ```
    Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0",  GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"darwin/amd64"}
    Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.13-gke.401", GitCommit:"eb94c181eea5290e9da1238db02cfef263542f5f", GitTreeState:"clean", BuildDate:"2020-09-09T00:57:35Z", GoVersion:"go1.13.9b4", Compiler:"gc", Platform:"linux/amd64"}
    ```

* Kubernetes cluster kind:

    GKE `v1.16.13-gke.401`

* Manifests:

   see above `test-probe.yaml`

* Prometheus Operator Logs:

```
 prometheus-operator level=info ts=2020-09-19T03:59:33.234581106Z caller=operator.go:1203 component=prometheusoperator msg="sync prometheus" key=monitoring/k8s
 prometheus-operator panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #3637, opened 2020-10-31
add sanity check for empty probe targets

sschne:
This PR adds a sanity check for Probes, so that empty targets will not end up in a nil pointer dereference.

Closes: #3517

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1743,7 +1743,7 @@ func (c *Operator) selectPodMonitors(ctx context.Context, p *monitoringv1.Promet
 func (c *Operator) selectProbes(p *monitoringv1.Prometheus) (map[string]*monitoringv1.Probe, error) {
 	namespaces := []string{}
 	// Selectors might overlap. Deduplicate them along the keyFunc.
-	res := make(map[string]*monitoringv1.Probe)
+	probes := make(map[string]*monitoringv1.Probe)
 
 	bMonSelector, err := metav1.LabelSelectorAsSelector(p.Spec.ProbeSelector)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -740,10 +743,11 @@ func (cg *configGenerator) generateProbeConfig(
 		}...)
 
 		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: enforceNamespaceLabel(relabelings, m.Namespace, enforcedNamespaceLabel)})
-	}
-
-	// Generate kubernetes_sd_config section for ingress resources.
-	if m.Spec.Targets.StaticConfig == nil {
+	} else {
```
s-urbaniak, line 746: I think there is nothing happening after the `if` block. So how about simply returning `return cfg, nil` after line 745 which alles unindenting the whole `else` block?

simonpasquier (commented):
What about implementing the check in [`*Operator.selectProbes()`](https://github.com/prometheus-operator/prometheus-operator/blob/7daed4b44b1a198972f3d7ea3c904895fb536861/pkg/prometheus/operator.go#L1791) so an invalid Probe definition doesn't block the reconciliation loop?

pkg/prometheus/promcfg.go
sschne: Thanks, changed it accordingly

sschne:
> What about implementing the check in [`*Operator.selectProbes()`](https://github.com/prometheus-operator/prometheus-operator/blob/7daed4b44b1a198972f3d7ea3c904895fb536861/pkg/prometheus/operator.go#L1791) so an invalid Probe definition doesn't block the reconciliation loop?

sschne:
> What about implementing the check in [`*Operator.selectProbes()`](https://github.com/prometheus-operator/prometheus-operator/blob/7daed4b44b1a198972f3d7ea3c904895fb536861/pkg/prometheus/operator.go#L1791) so an invalid Probe definition doesn't block the reconciliation loop?

I checked and i think this is not the correct function to validate the probes content, as it is only fetching probes by its labelSelector and returns them. As the generateProbeConfig() function actually looks into the content of the Probe, this might be the better function.

simonpasquier:
@sschne the problem with the current approach is that an invalid probe definition would prevent the operator from reconciling the whole Prometheus resource. See what is done in [`Operator.selectPodMonitors()`](https://github.com/prometheus-operator/prometheus-operator/blob/30b03e839a460ef230256873c10ad56f9a5abd37/pkg/prometheus/operator.go#L1703) where we discard invalid pod monitors.

We need also to clarify whether a probe with empty static config and empty ingress definition should be dropped. From my POV, it is the right thing to do but pinging @lilic for confirmation. It should also be documented in the [API](https://github.com/prometheus-operator/prometheus-operator/blob/30b03e839a460ef230256873c10ad56f9a5abd37/pkg/apis/monitoring/v1/types.go#L873-L881).

lilic:
Agreed with @simonpasquier approach here, it should be non-blocking and discarding, and logging it would be the right approach here.

@sschne Thanks for the PR let us know if this makes sense and if you need any more info!

s-urbaniak:
also from my side agreed on the proposed behavior by @lilic and @simonpasquier :+1:

sschne:
@simonpasquier i've added the check in Operator.selectProbes() function in the style of the Operator.selectPodMonitors() function. Also I left the else block change in the generateProbeConfig() in as s-urbaniak advised.
Sadly, adding code coverage is quite hard here. I hope this is sufficient as is.

pkg/prometheus/promcfg.go
```diff
@@ -740,112 +740,110 @@ func (cg *configGenerator) generateProbeConfig(
 		}...)
 
 		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: enforceNamespaceLabel(relabelings, m.Namespace, enforcedNamespaceLabel)})
+		return cfg
```
simonpasquier, line 743: It would mean that if `StaticConfig` and `Ingress` are defined at the same time, only the former would be used. I'm not sure this is what we wanted initially. @lilic?

sschne: From what i can see, i didn't change the behavior, as the two if statements for checking the content of m.Spec.Targets.StaticConfig are complementary. If this is no intentional behavior, i can revert this and we can file an issue for it

simonpasquier: @sschne totally agree. I was pinging @lilic for clarifications so we don't forget :)

sschne: done

paulfantom (approved):

simonpasquier (approved):
Thanks!

PR #3647, opened 2020-11-04
pkg/alertmanager: copy user data to the generated config secret

### LINKED ISSUE #3638: AlertmanagerSpec configSecret only mounts `alertmanager.yaml`
**What happened?**

Utilized `configSecret` field of the `AlertmanagerSpec` to mount 2 files to `/etc/alertmanager/config`, however only the key named `alertmanager.yaml` was mounted.

I expected both files to be mounted (in this case it was `.tmpl` file for notification templates)

Create a secret with two file keys:

```
apiVersion: v1
data:
  alertmanager.yaml: <some base64>
  my.tmpl: <some base64>
kind: Secret
metadata:
  annotations:
```

**Environment**

    `prometheus-operator:v0.43.0`
    

    `v1.18.1`
    

* Kubernetes cluster kind:

    kops

* Manifests:

```
kind: Alertmanager
metadata:
  labels:
    alertmanager: main
  name: main
  namespace: monitoring
spec:
  configSecret: alertmanager-config-8hcb55hchg < secret with two files
  externalUrl: *********
  image: quay.io/prometheus/alertmanager
  nodeSelector:
    kubernetes.io/os: linux
  replicas: 3
  securityContext:
    fsGroup: 2000
    runAsNonRoot: true
    runAsUser: 1000
  serviceAccountName: alertmanager-main
  version: v0.21.0
```

* Prometheus Operator Logs:

```
```

My guess is the code is explicitly looking for the `alertmanager.yaml` key and nothing else, which is fine. What I was trying to do was use `kustomize` to `configMapGenerator/secretGenerator` to take my alertmanager templates and not only mount them to a known path, but let `kustomize` use a sha on the name so alertmanager also reloads when a new one is set. Lastly, this does not work because in your alertmanager config, you can't wildcard a directory in the `templates` path (example below)

```
global:
  resolve_timeout: 5m
  slack_api_url: https://SLACK_TOKEN
templates:
- '/etc/alertmanager/configmaps/alertmanager-templates-SOMESHA?/my.tmpl'
```

simonpasquier:
The user-provided secret containing the base Alertmanager configuration
may also include additional data (such as template files) that should be
copied to the final secret so the Alertmanager container can get access
to.

Closes #3638

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -90,6 +90,7 @@ type Config struct {
 	Namespaces                   operator.Namespaces
 	Labels                       operator.Labels
 	AlertManagerSelector         string
+	SecretListWatchSelector      string
 }
 
 // New creates a new controller.
@@ -109,11 +110,6 @@ func New(ctx context.Context, c operator.Config, logger log.Logger, r prometheus
 		return nil, errors.Wrap(err, "instantiating monitoring client failed")
 	}
 
-	secretListWatchSelector, err := fields.ParseSelector(c.SecretListWatchSelector)
-	if err != nil {
-		return nil, errors.Wrap(err, "can not parse secrets selector value")
-	}
-
 	o := &Operator{
 		kclient: client,
 		mclient: mclient,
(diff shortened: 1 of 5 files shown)


pkg/alertmanager/operator.go
```diff
@@ -653,7 +664,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 
 	assetStore := assets.NewStore(c.kclient.CoreV1(), c.kclient.CoreV1())
 
-	if err := c.provisionAlertmanagerConfiguration(context.TODO(), am, assetStore); err != nil {
+	if err := c.provisionAlertmanagerConfiguration(ctx, am, assetStore); err != nil {
```
s-urbaniak, line 667: great catch :+1:

s-urbaniak:
LGTM

PR #3653, opened 2020-11-06
Fix typos and add codespell pipeline

invidian:


lilic:
Hey, thanks for the PR, for some reason checking the files changed page crashes chrome, so can't really review 👍 

cc @prometheus-operator/prometheus-operator-reviewers this also adds a new test for spell check, is this something we want to add as part of new step or as part of existing one?

test/e2e/prometheus_test.go
```diff
@@ -2336,7 +2336,7 @@ func testThanos(t *testing.T) {
 		proxyGet := framework.KubeClient.CoreV1().Services(ns).ProxyGet
 		request := proxyGet("http", qrySvc.Name, "http-query", "/api/v1/query", map[string]string{
 			"query": "prometheus_build_info",
-			"dedup": "false",
+			"dedupe": "false",
```
simonpasquier, line 2339: this breaks the e2e test

invidian: Oops. I'll skip `dedup` word then.

invidian: Done. Thanks for pointing out!

simonpasquier:
I'm on the fence when it comes to enforcement by the CI. Because `codespell` doesn't parse the Go code, it can suggest modifications that are invalid such as in the [initial revision](https://github.com/prometheus-operator/prometheus-operator/pull/3653#pullrequestreview-525201143). Luckily our tests caught it but it could have been easily missed in code reviews. Also maintaining the list of exceptions feels tedious IMHO.
I'd be more comfortable if the check only applies to comments and docs but I'm not sure `codespell` can get to this level of granularity.

invidian:
>I'm on the fence when it comes to enforcement by the CI. Because codespell doesn't parse the Go code, it can suggest modifications that are invalid such as in the initial revision. Luckily our tests caught it but it could have been easily missed in code reviews. Also maintaining the list of exceptions feels tedious IMHO.
I'd be more comfortable if the check only applies to comments and docs but I'm not sure codespell can get to this level of granularity.

Well, typos also happen in names used in the code, including error messages, so I think it's valuable to run in across entire code base. Of course this is something to take with a pinch of salt, the same as code linters. In some cases it might be wrong, but overall it adds big value. But exceptions can be whitelisted or alternative names can be used in code. This might actually also improve readability, if you initially choose a name, which may seem like a typo to someone else.

I don't insist on adding a CI check for it, I can remove this commit. Then I may revisit after some time with new fixes if they affect projects I'm working on (e.g. [Lokomotive](https://github.com/kinvolk/lokomotive), where we ship Helm charts for Prometheus Operator, which include CRDs).

simonpasquier:
@invidian don't get me wrong: I very much appreciate the PR :bowing_man: and only questions the CI addition.
Lets see what other maintainers think of it :)

lilic:
Yes agreed with @simonpasquier on this, I am not a huge fan of adding this to CI. 

So I would suggest we can just keep the fixing of the spelling and if you don't mind opening a new issue to add some form of spell check for this in the future, this way we can avoid discussing on this PR and rather discuss on the isse, sounds good? :) 

Thanks again!

invidian:
Created https://github.com/prometheus-operator/prometheus-operator/issues/3654 then, will remove the commit now.

simonpasquier:
:+1: thanks a lot!

lilic (approved):
Thanks!

lgtm

Issue #3659, opened 2020-11-09
Default config-reloader-memory (25Mi) may be too small for the new unifed prometheus config reloader


desaintmartin:
**What happened?**

Lots of OOMKills on config-reloader container from prometheus Pod since upgrading to prometheus-operator 0.43.0.

**Did you expect to see something different?**

Not having OOMKills

**How to reproduce it (as minimally and precisely as possible)**:

- Deploy prometheus operator with default values
- Observe config-reloader being killed every few hours

**Environment**

Using kube-prometheus-stack Helm Chart, which uses default values for `--config-reloader-memory`

* Prometheus Operator version:

    `0.43.x`
    

* Kubernetes version information:

    `1.18.x`

simonpasquier:
Out of curiosity, what's the size of your prometheus config? Also can you share a graph of memory usage of the container?

simonpasquier:
That being said, it seems a valid issue given that previously we had 2 reloader containers, each capped at 25Mi by default :)

desaintmartin:
Hm yes I haven't mentioned that, I don't know if 50Mi is the right "default" value but that looked good first approach. Probably something in the middle would be right for ~90% of Prometheus deployments.

Prometheus config is ~1200 lines long and rulefiles is ~10000 lines long. Quite big.

simonpasquier:
How does the memory graph look like if you set the limit to 50Mi? Is it steadily increasing, flat, sawtooth?

desaintmartin:
Hm, it looks... weird.
This is for the last 3 days. Looks stable at 39MB
<img width="806" alt="Screenshot 2020-11-12 at 15 40 42" src="https://user-images.githubusercontent.com/38929/98953870-8153a300-24fd-11eb-90ff-46a69ca38883.png">

PR #3660, opened 2020-11-09
operator cmd: set --config-reloader-memory to 50Mi instead of 25Mi.

desaintmartin:
New unified reloader takes more memory than the two previous ones.

Solves https://github.com/prometheus-operator/prometheus-operator/issues/3659


brancz:
Thanks!

PR #3663, opened 2020-11-10
Add settings for alertmanager clusters

### LINKED ISSUE #38: Add PodMonitor

hacktastic:
alertmanager: adds settings for alertmanager statefulsets

This PR exposes three new alertmanager flags:
- --cluster.gossip-interval
- --cluster.pushpull-interval
- --cluster.peer-timeout

Each flag has a corresponding field on AlertmanagerSpec:
- clusterGossipInterval
- clusterPushpullInterval
- clusterPeerTimeout


pkg/apis/monitoring/v1/types.go
```diff
@@ -1283,6 +1283,12 @@ type AlertmanagerSpec struct {
 	// Needs to be provided for non RFC1918 [1] (public) addresses.
 	// [1] RFC1918: https://tools.ietf.org/html/rfc1918
 	ClusterAdvertiseAddress string `json:"clusterAdvertiseAddress,omitempty"`
+	// Interval between gossip attempts
+	ClusterGossipInterval string `json:"clusterGossipInterval,omitempty"`
+	// Interval between pushpull attempts
```
simonpasquier, line 1288:
```suggestion
	// Interval between pushpull attempts.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1283,6 +1283,12 @@ type AlertmanagerSpec struct {
 	// Needs to be provided for non RFC1918 [1] (public) addresses.
 	// [1] RFC1918: https://tools.ietf.org/html/rfc1918
 	ClusterAdvertiseAddress string `json:"clusterAdvertiseAddress,omitempty"`
+	// Interval between gossip attempts
```
simonpasquier, line 1286:
```suggestion
	// Interval between gossip attempts.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1283,6 +1283,12 @@ type AlertmanagerSpec struct {
...
+	ClusterGossipInterval string `json:"clusterGossipInterval,omitempty"`
+	// Interval between pushpull attempts
+	ClusterPushpullInterval string `json:"clusterPushpullInterval,omitempty"`
+	// Timeout for cluster peering
```
simonpasquier, line 1290:
```suggestion
	// Timeout for cluster peering.
```

simonpasquier (commented):
Thanks! The linked issue doesn't look correct :)

hacktastic:
@simonpasquier would you like me to squash and re-push?

simonpasquier (approved):
lgtm

PR #3686, opened 2020-11-16
promcfg: Change label regex to match empty strings

Zebradil:
This fixes matching labels with empty values when using Exists/NotExists operators.

I faced this issue while using the following bits of configuration:

ServiceMonitor:
```yaml
kind: ServiceMonitor
metadata:
  name: generic
  namespace: myns
spec:
  endpoints:
  - port: metrics
  selector:
    matchExpressions:
    - key: scrape-metrics
      operator: Exists
```

Service
```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/component: storage
    app.kubernetes.io/name: fluentd
    scrape-metrics: ""
  name: fluentd
  namespace: myns
spec:
  clusterIP: None
  ports:
  - name: metrics
    port: 24231
    targetPort: metrics
  selector:
    app.kubernetes.io/component: storage
    app.kubernetes.io/name: fluentd
```

The following selector should have matched `scrape-metris: ""` label, but it didn't:
```yaml
  selector:
    matchExpressions:
    - key: scrape-metrics
      operator: Exists
```

This is the corresponding bit from prometheus config:
```yaml
  relabel_configs:
  - action: keep
    source_labels:
    - __meta_kubernetes_service_label_scrape_metrics
    regex: .+
```

If this change makes sense, I'll update testst accordingly.

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -534,14 +534,14 @@ func (cg *configGenerator) generatePodMonitorConfig(
 		case metav1.LabelSelectorOpExists:
 			relabelings = append(relabelings, yaml.MapSlice{
 				{Key: "action", Value: "keep"},
-				{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_label_" + sanitizeLabelName(exp.Key)}},
-				{Key: "regex", Value: ".*"},
+				{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_labelpresent_" + sanitizeLabelName(exp.Key)}},
+				{Key: "value", Value: "true"},
 			})
 		case metav1.LabelSelectorOpDoesNotExist:
 			relabelings = append(relabelings, yaml.MapSlice{
 				{Key: "action", Value: "drop"},
-				{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_label_" + sanitizeLabelName(exp.Key)}},
-				{Key: "regex", Value: ".*"},
+				{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_labelpresent_" + sanitizeLabelName(exp.Key)}},
+				{Key: "value", Value: "true"},
 			})
 		}
 	}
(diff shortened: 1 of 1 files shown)


simonpasquier:
Hmm because of how label matching works, it would also match services that don't have the said label. We need to leverage `__meta_kubernetes_<object>_labelpresent_<label>` (this is available since [v2.9.0](https://github.com/prometheus/prometheus/commit/dafd1632a236e7f2510a9acfac16f0a72a84a1d7)).

Zebradil:
Thanks for the feedback!
I'll update this PR accordingly. Moving it into WIP for now.

pkg/prometheus/promcfg.go
```diff
@@ -534,14 +534,14 @@ func (cg *configGenerator) generatePodMonitorConfig(
 		case metav1.LabelSelectorOpExists:
 			relabelings = append(relabelings, yaml.MapSlice{
 				{Key: "action", Value: "keep"},
-				{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_label_" + sanitizeLabelName(exp.Key)}},
-				{Key: "regex", Value: ".+"},
+				{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_labelpresent_" + sanitizeLabelName(exp.Key)}},
```
brancz, line 537: With this change, any actual values of labels would be lost. I think we need a combined `regex` something along the lines of:

```
{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_labelpresent_" + sanitizeLabelName(exp.Key), "__meta_kubernetes_pod_label_" + sanitizeLabelName(exp.Key)}},
{Key: "regex", Value: "true;.+"},
```

Zebradil: Thanks! I've updated the code accordingly.

brancz: I was slightly off the first time around actually, sorry for not realizing this the first time around. It should actually be

{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_labelpresent_" + sanitizeLabelName(exp.Key)}},
{Key: "regex", Value: "true"},

Zebradil: No worries, I didn't bother myself to check documentation on this topic, so thanks for taking care of it. I'll change the code.

brancz (approved):
This lgtm! I think you can move this from being a draft to the real PR.

cc @simonpasquier could you give this another review as well?

simonpasquier:
lgtm too :+1:

Issue #3122, opened 2020-04-04
kubernetes 1.18, kubectl apply --validate=yes can't install CRD


Slach:
**What happened?**
kubectl apply --validate=yes can't install CRD from github URL

I try to run following command
```
kubectl --namespace=prometheus apply --validate=true -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagers.yaml
```
and got exit code 1
and following error
```
The CustomResourceDefinition "alertmanagers.monitoring.coreos.com" is invalid:
* spec.validation.openAPIV3Schema.properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```

**Did you expect to see something different?**
yes, on kubectl 1.17.x and kubernetes 1.17, I didn't get any errors

**How to reproduce it (as minimally and precisely as possible)**:
- install k8s 1.18 from zero
- try install latest CRD's with validation

**Environment**
clear installation 
minikube 1.9.1 + kubernetes 1.18

* Kubernetes version information:

    `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:50:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:

    minikube

donbowman:
This is due to https://github.com/kubernetes/kubernetes/pull/84920

Slach:
@donbowman could you suggest how CRD should be changed to pass this validation?

lilic:
Can you try to do a kubectl create instead?

Slach:
[...]

Issue #3022, opened 2020-02-11
CRD manifests are now generated with status field


omerlh:
**What happened?**
I update my jsonnet library to the latest version and noticed that now the CRDs contains a new field, `status`. This is problematic when using GitOps tools like flux/argo - as this field is updated by k8s and therefore the sync from Git to Kubernetes fails.

**Did you expect to see some different?**
No

**How to reproduce it (as minimally and precisely as possible)**:
Generate manifests using the latest jsonnet library:

![image](https://user-images.githubusercontent.com/6730584/74249931-93947c80-4cf2-11ea-82aa-ec58e880be4f.png)

omerlh:
A workaround for the issue is by doing the following jsonnet magic (if using `kube-prometheus`):
```jsonnet
    local po = super.prometheusOperator,
    prometheusOperator:: 
    {
      
      [name]: std.mergePatch(po[name], {
          status: null,
        })
      for name in std.objectFields(po)
    },
```

brancz:
@pgier can we make sure we remove this entirely? We’re not setting any status so let’s remove the connection until we do.

lilic:
@omerlh you are also welcome to remove this by doing a PR! 😊

PR #3694, opened 2020-11-19
Bump controller-tools to v0.4.1

dgrisonnet:
This bump includes validation improvements for CRDs and allows us to remove a dirty hack in the generation script.

As this bump was blocked until now, I'll give more context about it. Originally, a controller-tools bump was done a few months ago from v0.2.4 to v0.2.8 but it was reverted because the generated CRDs were invalid: #3122
To upgrade this package again, we needed to use v1 CRDs and define a default for the property causing the validation error: #3137 
The CRD upgrade was done a few releases ago in prometheus-operator but we were still waiting for the default property to be added in controller-tools. As it was added to v0.4.1, we can now safely resume the upgrade.

On a side note, this bump was also meant to resolve the few issues we've seen in prometheus-operator and kube-prometheus about the status field generated at the end of the CRD manifests: #3022. A fix for that was added in controller-tools but later reverted because it was not working as intended. Thus, we might need to define a new strategy to tackle these issues.


simonpasquier:
lgtm

Issue #3690, opened 2020-11-18
Inconsistent naming between base Alertmanager and AlertmanagerConfig CRD receiver keys


longtomjr:
The capitalization for the `AlertmanagerConfig` CRD receiver keys differ from how the words are broken up in the base Alertmanager config
https://github.com/prometheus-operator/prometheus-operator/blob/131dad67dd1537387b751f9829ba04dbc3675a00/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L78-L84
vs
```
# Configurations for several notification integrations.
email_configs:
  [ - <email_config>, ... ]
pagerduty_configs:
  [ - <pagerduty_config>, ... ]
pushover_configs:
  [ - <pushover_config>, ... ]
slack_configs:
  [ - <slack_config>, ... ]
opsgenie_configs:
  [ - <opsgenie_config>, ... ]
webhook_configs:
  [ - <webhook_config>, ... ]
victorops_configs:
  [ - <victorops_config>, ... ]
wechat_configs:
  [ - <wechat_config>, ... ]
```
https://prometheus.io/docs/alerting/latest/configuration/#receiver

I can see this going both ways, since the capitalization is consistent with the company branding `WeChat` and `PagerDuty`. I think it makes sense to use the camel casing for separating words (and ignoring the case in branding), in the same way Alertmanager use the snake casing without the branding part. 

I think having this be consistent with Alertmanager makes sense, especially since the word separation is unambiguous.

lilic:
Thanks for the issue. At least we can choose to go one way and agreed, being consistent with alertmanager is the right way, as we do this with other config fields.

From what I can gather we need the change the following two, or did I miss any:
- weChatConfigs -> wechatConfigs
- pagerDutyConfigs -> pagerdutyConfigs

cc @simonpasquier and @grdryn for further thoughts on this.

simonpasquier:
I'm fine with it, consistency is definitely good. I only see the 2 fields @lilic mentioned that would need a change.

grdryn:
Sounds good to me. :+1:

lilic:
@longtomjr are you interested in doing the changes and opening a PR? No worries if not, we can add a help-wanted in case anyone wants to pick up as its a good first issue.

PR #3696, opened 2020-11-19
Fix inconsistent naming in alertmgr crd

arghya88:
Signed-off-by: Arghya Sadhu <arghya88@gmail.com>

Fixes #3690

PR #3702, opened 2020-11-20
Refactor AlertmanagerConfig types and keys for consistency with Alertmanager

longtomjr:
Closes: #3690

I tried to keep all the documentation that talks about the 3rd party services in line with the brand formatting of the services, but all the variables, fields and types are consistent with Alertmanager.


simonpasquier:
I've merged #3696 which had a narrower scope but feel free to rebase to get the internal structs consistent.

PR #3703, opened 2020-11-20
*: fix child routes support in AlertmanagerConfig

### LINKED ISSUE #3688: [alertmanagerconfig] routes field gets empty when applied to cluster
**What happened?**

When I apply the following configuration via helm the `routes:` field is suddenly empty on the cluster:

```yaml

kind: AlertmanagerConfig
metadata:
  name: foo
  labels:
    release: prometheus-stack
  namespace: foo
spec:
  receivers:
    - name: foo-msteams
      webhookConfigs:
        - url: "http://prometheus-msteams:2000/foo"
          sendResolved: true
  route:
    groupBy: [...]
    groupWait: 30s
    groupInterval: 5m
    repeatInterval: 12h
    receiver: foo-msteams
    routes:
    - receiver: foo-msteams
      matchers:
      - name: customer
        value: foo
```

The applied version looks like this: `kubectl describe alertmanager`:
```
Name:         foo
Namespace:    foo
Labels:       app.kubernetes.io/managed-by=Helm
              release=prometheus-stack
Annotations:  meta.helm.sh/release-name: hfoo
              meta.helm.sh/release-namespace: foo
API Version:  monitoring.coreos.com/v1alpha1
Kind:         AlertmanagerConfig
Metadata:
  Creation Timestamp:  2020-11-17T13:52:03Z
  Generation:          3
  Resource Version:    94280348
  Self Link:           /apis/monitoring.coreos.com/v1alpha1/namespaces/hs-schraeder/alertmanagerconfigs/hs-schraeder
  UID:                 2f18718e-e0e8-4f24-a101-1ae4d619d1e0
Spec:
  Receivers:
    Name:  foo-msteams
    Webhook Configs:
      Send Resolved:  true
      URL:            http://prometheus-msteams:2000/foo
  Route:
    Group By:
      ...
    Group Interval:   5m
    Group Wait:       30s
    Receiver:         foo-msteams
    Repeat Interval:  12h
    Routes:
Events:  <none>
```

This leads to this error message here:
```
component=alertmanageroperator msg="skipping alertmanagerconfig" error="receiver \"\" not found" alertmanagerconfig=foo/foo namespace=monitoring alertmanager=prometheus-stack-kube-prom-alertmanager
```

I expect that the routes section gets deployed, too.

Copy my YAML and remove the helm go templates...

**Environment**

    `quay.io/prometheus-operator/prometheus-operator:v0.43.`
    

    `Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"archive", BuildDate:"2020-11-15T17:26:49Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.6", GitCommit:"d32e40e20d167e103faf894261614c5b45c44198", GitTreeState:"clean", BuildDate:"2020-05-20T13:08:34Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}`
    

* Kubernetes cluster kind:

    insert how you created your cluster: Rancher

simonpasquier:
The `route` field of the AlertmanagerConfig CRD introduces a circular
reference which doesn't play well with the Kubernetes API. The end
result is that any sub-route from an AlertmanagerConfig object doesn't
show up in the generated configuration.

To get rid of the recursion, child routes are modeled as raw JSON
fields. The downside is that the API will only check that sub-routes
contain valid JSON data. The operator will ensure that the routes
definition is semantically correct.

Closes #3688


simonpasquier:
cc @grdryn since you did the initial implementation :)

grdryn:
Interesting! I didn't realise this would cause an issue :thinking:

When working on this, I had created https://github.com/kubernetes-sigs/controller-tools/issues/477 when the CRD didn't get generated properly, and https://github.com/kubernetes-sigs/controller-tools/issues/489 is the remaining follow-up issue. It might not be useful here, but linking just in case :)

simonpasquier:
Thanks @grdryn for the additional links. Indeed I came to the same conclusion that recursive types aren't supported currently and it doesn't seem an easy feature to add (I think https://github.com/kubernetes/kubernetes/issues/62872 is the tracking issue). 
Going down the rabbit hole, I found https://github.com/kubernetes-sigs/controller-tools/issues/294 and https://github.com/kubernetes-sigs/controller-tools/pull/126. The latter mentions a few structs to circumvent the problem, in particular `apiextensionsv1.JSON` that I've used here.

pkg/alertmanager/amcfg.go
```diff
@@ -138,20 +137,26 @@ func convertRoute(in *monitoringv1alpha1.Route, crKey types.NamespacedName, firs
...
 	if len(in.Routes) > 0 {
-		routes := make([]*route, len(in.Routes))
-		for i := range in.Routes {
-			routes[i] = convertRoute(&in.Routes[i], crKey, false)
+		routes = make([]*route, len(in.Routes))
+		children, err := in.ChildRoutes()
+		if err != nil {
+			// The controller has already checked that Routes are valid. If an
+			// error is returned here, it's a bug.
+			panic(err)
```
lilic, line 156: Can you clarify why we panic when we actually say above that error is returned here?

simonpasquier: In the comment, I meant to say that we shouldn't get any error here because child routes have already been checked in `checkAlertmanagerRoutes()` which gets called before generating the Alertmanager configuration.

https://github.com/prometheus-operator/prometheus-operator/pull/3703/files#diff-74002036bea7ef8f78c7623abba8d63a3e2578be73fb2dc8bcca154df57ce362R1087-R1103

simonpasquier:
Following @paulfantom comment, I've created https://github.com/prometheus-operator/prometheus-operator/issues/3710 to implement admission webhook for `AlertmanagerConfig`. Despite this being important with this PR moving inner routes to raw JSON payload, it would also be useful for other fields.

pkg/alertmanager/amcfg.go
lilic: Makes sense now, thanks!

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -102,8 +105,29 @@ type Route struct {
 	// Boolean indicating whether an alert should continue matching subsequent
 	// sibling nodes. It will always be overridden to true for the first-level
 	// route by the Prometheus operator.
-	Continue bool    `json:"continue,omitempty"`
-	Routes   []Route `json:"routes,omitempty"`
+	Continue bool `json:"continue,omitempty"`
+	// Child routes.
+	Routes []apiextensionsv1.JSON `json:"routes,omitempty"`
+	// Note: this comment applies to the field definition below but appears
```
lilic, line 111: > Note: this comment applies to the field definition below but appears below otherwise it gets included in the generated manifest.

Not sure I understand the two below things, maybe we can clarify this a bit?

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -102,8 +105,29 @@ type Route struct {
...
+	// below otherwise it gets included in the generated manifest.
+	// CRD schema doesn't support self referential types for now (see
+	// https://github.com/kubernetes/kubernetes/issues/62872).  We have to use
```
lilic, line 114:
```suggestion
	// https://github.com/kubernetes/kubernetes/issues/62872). We have to use
```

lilic (commented):
Couple of comments, otherwise lgtm, great stuff!

I do have one question regarding what happens for folks who already make use of this, I understand we are alpha API and we can break this, but wondering if a warning is needed in our CHANGELOG?

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
simonpasquier: you're right, it didn't read correctly :)

simonpasquier:
> I do have one question regarding what happens for folks who already make use of this, I understand we are alpha API and we can break this, but wondering if a warning is needed in our CHANGELOG?

For now, child routes in `AlertmanagerConfig` don't appear in the generated config so at best, it isn't doing what people expect it to do. Worst case, someone provisioned child routes that aren't correct and in this case, the operator would reject the `AlertmanagerConfig` object.
But I agree that it should be mentioned as `[CHANGE]` in the changelog.

lilic (approved):
lgtm on green 🎉

brancz:
Any reason we haven't merged this yet? :)

simonpasquier:
No reason except ETOOMANYMEETINGS and flaky quay.io :)

PR #3713, opened 2020-11-27
pkg/alertmanager: change 'send_resolved' fields to *bool

simonpasquier:
Depending on the integration, the default value of 'send_resolved' can
be either 'true' or 'false'. This means that we need to represent it as
a pointer to bool in the Alertmanager configuration struct so the field
isn't rendered if nil.


brancz:
Nice! lgtm :+1: :tada:

PR #3716, opened 2020-11-30
Bump Prometheus and Thanos versions

dgrisonnet:
Update the compatibility matrix with the latest releases of Prometheus and update the default Thanos version to the latest release.


simonpasquier (approved):
:+1:

PR #3728, opened 2020-12-07
fix #3726 - omit empty responder from list

### LINKED ISSUE #3726: AlertmanagerConfig - Opsgenie Responder
**What happened?**

Hi I'm switching towards the new AlertmanagerConfig CRD, but when I try to configure Opsgenie responders I get the following error: `OpsGenieConfig responder {   } has to have at least one of id, username or name specified"`.

I tried it with different values for responders:

```yaml
responders:
-  name: "{{ if .CommonLabels.responder }}{{ .CommonLabels.responder }}{{ else }}sre{{ end }}"
    type: team
```
and
``` yaml
responders:
-  name: sre
    type: team
```

Both with the same result.

That the Opsgenie responder will be configured within Alertmanager.

Providing a `opsgenieConfigs.responders` object within the CRD.

**Environment**

    Release 0.44 deployed via kube-prometheus helm chart. CRDs are updated to the latest 0.44 release.

    1.17

* Kubernetes cluster kind:

    GKE

* Prometheus Operator Logs:

Operator logs are fine, this the full Alertmanager line.
```
level=error ts=2020-12-07T15:35:55.895Z caller=coordinator.go:124 component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/config/alertmanager.yaml err="OpsGenieConfig responder {   } has to have at least one of id, username or name specified"
```

jaredstehler:
Fix for #3726 - original code initializes a list of Responder with an empty first object, which results in the config rendering with an invalid responder.

```release-note:bug
Omit invalid (empty) responder from opsgenie config
```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -589,7 +589,7 @@ func (cg *configGenerator) convertOpsgenieConfig(ctx context.Context, in monitor
 
 	var responders []opsgenieResponder
 	if l := len(in.Responders); l > 0 {
-		responders = make([]opsgenieResponder, l)
+		responders = []opsgenieResponder{}
 		for _, r := range in.Responders {
 			var responder opsgenieResponder = opsgenieResponder{
 				ID:       r.ID,
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -589,7 +589,7 @@ func (cg *configGenerator) convertOpsgenieConfig(ctx context.Context, in monitor
 
 	var responders []opsgenieResponder
 	if l := len(in.Responders); l > 0 {
-		responders = []opsgenieResponder{}
+		responders = make([]opsgenieResponder, 0, l)
 		for _, r := range in.Responders {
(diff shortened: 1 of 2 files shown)


pkg/alertmanager/amcfg.go
```diff
@@ -589,7 +589,7 @@ func (cg *configGenerator) convertOpsgenieConfig(ctx context.Context, in monitor
 
 	var responders []opsgenieResponder
 	if l := len(in.Responders); l > 0 {
-		responders = make([]opsgenieResponder, l)
+		responders = []opsgenieResponder{}
```
simonpasquier, line 592: I think the original intent was:

```suggestion
		responders = make([]opsgenieResponder, 0, l)
```

simonpasquier (commented):
Thanks! Could you base the PR against the `release-0.44` branch so we can have the fix in `v0.44.1`?

jaredstehler:
> Thanks! Could you base the PR against the `release-0.44` branch so we can have the fix in `v0.44.1`?

done!

simonpasquier (approved):

simonpasquier:
Thanks!

PR #3730, opened 2020-12-08
log error when reloader watching file failed

skyever:
log the error msg if reloader watcher failed or we don't what it faield.

```release-note:REPLACEME

```


cmd/prometheus-config-reloader/main.go
```diff
@@ -182,7 +182,8 @@ func main() {
 			level.Info(logger).Log("msg", "Starting web server for metrics", "listen", *listenAddress)
 			http.Handle("/metrics", promhttp.HandlerFor(r, promhttp.HandlerOpts{Registry: r}))
 			return http.ListenAndServe(*listenAddress, nil)
-		}, func(error) {
+		}, func(err error) {
+			level.Error(logger).Log("Error", err)
```
simonpasquier, line 186: :+1:, just one small nit
 
```suggestion
			level.Error(logger).Log("err", err)
```

simonpasquier (approved):
LGTM

PR #3742, opened 2020-12-15
*: add validation to AlertmanagerConfig resources

simonpasquier:
This change enhances the OpenAPI spec of the Alertmanager CRD so the Kubernetes API can validate more fields and provide more valuable feedback to clients.

**Release Note**

```release-note:change
Add schema validations to AlertmanagerConfig CRD.
```


s-urbaniak:
lgtm :+1:

bundle.yaml
```diff
@@ -49,6 +49,7 @@ spec:
                         properties:
                           name:
                             description: Label to match.
+                            minLength: 1
```
lilic, line 52: Sorry, not too familiar with how validations work in practice. What is minLenght here, number of chars or number of strings? Thanks!

pkg/alertmanager/operator.go
```diff
@@ -1074,10 +1074,10 @@ func checkWechatConfigs(ctx context.Context, configs []monitoringv1alpha1.WeChat
 	for i, config := range configs {
 		wechatConfigKey := fmt.Sprintf("%s/wechat/%d", key, i)
 
-		if config.APIURL != nil {
-			_, err := url.Parse(*config.APIURL)
+		if len(config.APIURL) > 0 {
+			_, err := url.Parse(config.APIURL)
```
lilic, line 1078: I guess this is a validation we cannot do via the CRD itself? If its a valid URL? (Sorry if I missed it somewhere)

lilic (commented):
Nice work!

How would migration work here, if you have an existing configuration that makes these new changes invalid?

bundle.yaml
simonpasquier: this the minimal string length, e.g. you can't have an empty label name.

pkg/alertmanager/operator.go
simonpasquier: AFAICT yes, you can't really do complex validation like in the API. Also the client is free to skip API validations (`kubectl apply --validate=false -f ...`) so we still need to check it in the controller's code.

simonpasquier:
> How would migration work here, if you have an existing configuration that makes these new changes invalid?

The resources should continue to exist but they might be rejected by the controller when reconciling the Alertmanager configuration. It should be called out in the changelog.

simonpasquier:
@lilic is it ok for you now?

lilic:
lgtm sorry missed it in the noise of notifications!

PR #3757, opened 2020-12-22
Fix: Plural for Prometheus is Prometheis

cristiklein:
After extensive research, it has been determined that the correct plural
for 'Prometheus' is 'Prometheis'. https://prometheus.io/docs/introduction/faq/#what-is-the-plural-of-prometheus

Let us use the correct plural, to avoid angering the Titan god of fire.

Since this is a breaking API change, VERSION was bumped accordingly. All
users of prometheus-operator will have to update their Charts. This is a small
sacrifice given the greater danger: One never knows, but the last thing
operators need is to have their holidays ruined by the wrath of Prometheus.

Happy Holidays everyone!

```release-note:bug
The correct plural is now used for the Titan god of fire. Please update your usage of this Chart.
```


simonpasquier:
I'm afraid that we fear more the anger of our users than the potential curse of mythological creatures :)

coderanger:
I think Prometheus would be a pretty understanding guy, probably just happy that his whole "humanity" experiment turned out okay and not hung up about names :)

lilic:
Thanks for the PR! I think this might be a good candidate for when we switch to next version of the Prometheus custom resource, but for now like Simon mentioned this would cause breaking changes for our users. For now I don't think we have a need to switch to next major version so I would suggest we close the PR and instead open an issue for this @cristiklein so we don't forget when we do make breaking changes. :)

brancz:
It's not actually all that clear :D 

https://www.youtube.com/watch?v=B_CDeYrqxjQ

PR #3762, opened 2020-12-26
Optimize thanos sidecar code logic and add `TracingConfigFile` field.

### LINKED ISSUE #3763: Optimize thanos sidecar code logic and add `TracingConfigFile` field
**Why do we need it?**
1. Optimize`ObjectStorageConfig` and `ObjectStorageConfigFile` field processing logic , which match comment logic `When used alongside with ObjectStorageConfig, ObjectStorageConfigFile takes precedence`.
2. Add `TracingConfigFile` field support `-tracing.config-file` args.
3. When disable prometheus compact,we can set the `--storage.tsdb.min-block-duration` and `--storage.tsdb.max-block-duration `parameters together.

**Environment**
master branch.

mikechengwei:
Fix #3763
1. Optimize`ObjectStorageConfig` and `ObjectStorageConfigFile` field processing logic , which match comment logic `When used alongside with ObjectStorageConfig, ObjectStorageConfigFile takes precedence`.
2. Add `TracingConfigFile` field support `-tracing.config-file` args.
3. When disable prometheus compact,we can set the `--storage.tsdb.min-block-duration` and `--storage.tsdb.max-block-duration `parameters together.

- release-note:enhancement
```release-note:REPLACEME
Add `TracingConfigFile` field into thanos configuration.
```


pkg/prometheus/statefulset.go
```diff
@@ -750,7 +750,9 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 		}
 
 		if p.Spec.Thanos.ObjectStorageConfig != nil || p.Spec.Thanos.ObjectStorageConfigFile != nil {
-			if p.Spec.Thanos.ObjectStorageConfig != nil {
+			if p.Spec.Thanos.ObjectStorageConfigFile != nil {
+				container.Args = append(container.Args, "--objstore.config-file="+*p.Spec.Thanos.ObjectStorageConfigFile)
+			} else {
```
s-urbaniak, line 755: this feels like a duplication of logic that is already present in Thanos itself: https://github.com/thanos-io/thanos/blob/dd670a3e8cd2d99bd86169bdaf6bf304ab5ea2f9/pkg/extflag/pathorcontent.go#L50-L73

As far as i understand we don't really need the if-else switch.

pkg/prometheus/statefulset.go
```diff
@@ -779,14 +776,18 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
...
-			container.Env = append(container.Env, v1.EnvVar{
-				Name: "TRACING_CONFIG",
-				ValueFrom: &v1.EnvVarSource{
-					SecretKeyRef: p.Spec.Thanos.TracingConfig,
-				},
-			})
+		if p.Spec.Thanos.TracingConfig != nil || p.Spec.Thanos.TracingConfigFile != nil {
+			if p.Spec.Thanos.TracingConfigFile != nil {
+				container.Args = append(container.Args, "--tracing.config-file="+*p.Spec.Thanos.TracingConfigFile)
+			} else {
```
s-urbaniak, line 782: see above regarding the if-else construct.

pkg/prometheus/statefulset.go
```diff
@@ -807,6 +808,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 	}
 	if disableCompaction {
 		promArgs = append(promArgs, "--storage.tsdb.max-block-duration=2h")
+		promArgs = append(promArgs, "--storage.tsdb.min-block-duration=2h")
```
s-urbaniak, line 811: this looks correct to me but leaving @brancz and/or @simonpasquier to have another look.

s-urbaniak:
thank you @mikechengwei for the contribution! this generally looks good to me, i just have a nit regardin the if-else construct and a follow-up question for the prometheus settings.

pkg/prometheus/statefulset.go
mikechengwei: Ok.

pkg/prometheus/statefulset.go
mikechengwei: Ok.

pkg/prometheus/statefulset.go
simonpasquier: From what I can tell, Thanos won't start if both flags are defined and not empty:
https://github.com/thanos-io/thanos/blob/dd670a3e8cd2d99bd86169bdaf6bf304ab5ea2f9/pkg/extflag/pathorcontent.go#L53-L55

But the [help text](https://github.com/thanos-io/thanos/blob/dd670a3e8cd2d99bd86169bdaf6bf304ab5ea2f9/pkg/extflag/pathorcontent.go#L36) for the `*-file` flag is a bit misleading: it says `Alternative to '...' flag (lower priority). ...`. It should either say that both flags are exclusive or implement the priority logic.

pkg/prometheus/statefulset.go
simonpasquier: The default value for `--storage.tsdb.min-block-duration` is already [2h](https://github.com/prometheus/prometheus/blob/063154eab720d8c3d495bd78312c0df090d0bf23/cmd/prometheus/main.go#L197-L198) and it shouldn't change for 2.x versions but it doesn't hurt to set it explicitly.

simonpasquier (commented):
Process-wise, I'd prefer to have a different PR for each change (e.g. object storage config, tracing config and block duration).

pkg/prometheus/statefulset.go
mikechengwei: I think we can add if-else to keep Semantic correctness.

mikechengwei:
> Process-wise, I'd prefer to have a different PR for each change (e.g. object storage config, tracing config and block duration).

Do I need to split the PR and submit it in turn this time?

blaquezzz (approved):

pkg/apis/monitoring/v1/types.go
```diff
@@ -537,6 +537,9 @@ type ThanosSpec struct {
 	ListenLocal bool `json:"listenLocal,omitempty"`
 	// TracingConfig configures tracing in Thanos. This is an experimental feature, it may change in any upcoming release in a breaking way.
 	TracingConfig *v1.SecretKeySelector `json:"tracingConfig,omitempty"`
+	// TracingConfig specifies the path of the tracing configuration file.
+	// When used alongside with TracingConfig, TracingConfigFile takes precedence.
+	TracingConfigFile *string `json:"tracingConfigFile,omitempty"`
```
simonpasquier, line 542: I'd rather have a `string` here than a `*string`.

simonpasquier (commented):
I have one small request but fine if it got merged as-is. Leaving final call to @lilic as the next release shepherd :)

lilic:
Thanks for the ping Simon, I plan on doing the release tomorrow morning. But agreed on Simons comment lets address that first.

pkg/apis/monitoring/v1/types.go
mikechengwei: done

mikechengwei:
@simonpasquier  I have modified，PTAL.

simonpasquier (approved):
lgtm

mikechengwei:
@lilic Please help me merge this pr.

mikechengwei:
@simonpasquier Can you help me merge this pr?

lilic (approved):
Thanks, lgtm

PR #3771, opened 2021-01-05
Documentation: Improve retention size description

lentzi90:
This updates the description of Prometheus `retentionSize` to specify what units are supported.
I found the supported units described [here](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L214).

Note: The documentation change in itself is minimal, but I'm unsure how to handle the fact that it impacts the CRDs (since the description is also in there).

- release-note:none


simonpasquier (commented):
Thanks! LGTM, the fact that the CRD description has changed isn't an issue.

PR #3779, opened 2021-01-06
Docs: Remove redundant step for add. scrape config

valentin-krasontovitsch:
Two steps in the instructions have the same result, making one of them redundant.

We remove the more "complicated" one (purely subjective and a feeling).

This is a minor fix of the documentation on additional scrape configs.

```release-note:none

```


metalmatze:
Seems good. Probably an oversight while rewriting the docs.

simonpasquier (approved):
Thanks!

PR #3785, opened 2021-01-12
*: make `route` field optional in `AlertmanagerConfig`

simonpasquier:
This is a follow-up of #3742. It is valid to define an
`AlertmanagerConfig` resource with only `inhibitRules` and we shouldn't
enforce that the `route` field is not empty.

**Release Note**

_no release note update because #3742 hasn't been released yet._

```release-note:none

```


lilic (approved):
Thanks, good catch!!

PR #3788, opened 2021-01-13
Bump deps and project default versions

lilic:
This is in preparation of the release that should happen today. 

Seems like the `IPFamily` field was [removed](https://kubernetes.io/docs/setup/release/notes/#urgent-upgrade-notes) from the Service object. 

```release-note:REPLACEME

```


pkg/k8sutil/k8sutil.go
```diff
@@ -118,7 +118,7 @@ func CreateOrUpdateService(ctx context.Context, sclient clientv1.ServiceInterfac
 		}
 	} else {
 		svc.ResourceVersion = service.ResourceVersion
-		svc.Spec.IPFamily = service.Spec.IPFamily
+		svc.Spec.IPFamilies = service.Spec.IPFamilies
```
lilic, line 121: Have not tested this yet, lets see if tests complain, but I suspect it should be as easy as this 🤞

go.mod
```diff
@@ -22,21 +22,21 @@ require (
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.44.1
 	github.com/prometheus-operator/prometheus-operator/pkg/client v0.0.0-00010101000000-000000000000
 	github.com/prometheus/alertmanager v0.21.0
-	github.com/prometheus/client_golang v1.8.0
-	github.com/prometheus/common v0.14.0
+	github.com/prometheus/client_golang v1.9.0
+	github.com/prometheus/common v0.15.0
 	github.com/prometheus/prometheus v1.8.2-0.20201015110737-0a7fdd3b7696
```
simonpasquier, line 27: do we want to bump prometheus and thanos versions here too?

simonpasquier (approved):

go.mod
lilic: Can do!

lilic: Done, any other ones. I didn't go through the minor ones as we have a lot, but if you can think of any other ones happy to bump!

metalmatze (approved):
Nice!

PR #3810, opened 2021-01-22
Fix: Preserve user added labels and annotations on Services, Endpoints and StatefulSets

Szymongib:
Fixes: #2018
This PR adds functionality to preserve any labels and annotations manually added by user on Services, Endpoints and Stateful Sets managed by the Operator. Labels and Annotations are preserved as long as the do not conflict with those added by the Operator.

Questions:
- I did not introduce it for Secrets yet as I am not sure if there is a use case for that, but I will add it for consistency sake if requested.
- Do we want e2e test for this functionality?

```release-note:bug
Preserve user added labels and annotations on Services, Endpoints and StatefulSets
```

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -175,6 +175,23 @@ func UpdateStatefulSet(ctx context.Context, sstClient clientappsv1.StatefulSetIn
 	return nil
 }
 
+// UpdateSecret merges metadata of existing Secret with new one and updates it.
+func UpdateSecret(ctx context.Context, secretClient clientv1.SecretInterface, secret *v1.Secret) error {
+	existingSecret, err := secretClient.Get(ctx, secret.Name, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "getting secret object failed")
+	}
+
+	mergeMetadata(&secret.ObjectMeta, existingSecret.ObjectMeta)
+
+	_, err = secretClient.Update(ctx, secret, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // GetMinorVersion returns the minor version as an integer
 func GetMinorVersion(dclient discovery.DiscoveryInterface) (int, error) {
 	v, err := dclient.ServerVersion()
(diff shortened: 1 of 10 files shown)


simonpasquier:
Thanks a lot for this!

> I did not introduce it for Secrets yet as I am not sure if there is a use case for that, but I will add it for consistency sake if requested.

I'd say that anything managed by the operator should be covered.

> Do we want e2e test for this functionality?

I think it is important enough that an e2e test would be appreciated.

Can you also do it for Alertmanager and Thanos ruler too?

pkg/k8sutil/k8sutil.go
```diff
@@ -189,3 +230,15 @@ func mergeOwnerReferences(old []metav1.OwnerReference, new []metav1.OwnerReferen
...
 }
+
+func mergeMetadata(new *metav1.ObjectMeta, old metav1.ObjectMeta) {
+	new.SetLabels(mergeMaps(new.Labels, old.Labels))
+	new.SetAnnotations(mergeMaps(new.Annotations, old.Annotations))
+}
+
+func mergeMaps(new map[string]string, old map[string]string) map[string]string {
+	for k, v := range new {
+		old[k] = v
```
coderanger, line 241: This function probably needs a check for `if old == nil`?

Szymongib: Yes, for sure. Thanks!

test/e2e/prometheus_test.go
```diff
@@ -1810,6 +1812,188 @@ func testPromOnlyUpdatedOnRelevantChanges(t *testing.T) {
...
+
+		_, err = rConf.update(res)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// Set label on CR to trigger reconcile
+	prometheusCRD.SetLabels(mergeMap(prometheusCRD.GetLabels(), map[string]string{"test": "reconcile"}))
+	prometheusCRD, err = framework.UpdatePrometheusAndWaitUntilReady(ns, prometheusCRD)
```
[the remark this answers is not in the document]
Szymongib: Yea, that makes sense. I though of doing it like that initially, but then forgot that `UpdatePrometheusAndWaitUntilReady` waits only for pods to be ready 🤦

simonpasquier (approved):
LGTM

simonpasquier:
Thanks for the PR!

Issue #3819, opened 2021-02-02
Generate receivers from alertmanagerconfig even if routes are empty


abhilashjoseph:
**What happened?**
Alert manager config file generation ignores receiver section generation when routes are not present. With a global alert manager config setup, generating a receiver alone might be required to merge with the alertmanager configuration from the values file directly in the helm chart.

**Did you expect to see something different?**
Receiver configuration is generated irrespective of whether routes are present or not in alertmanagerconfig custom resource

**How to reproduce it (as minimally and precisely as possible)**:
Create an alertmanagerconfig resource without any routes, the generated alertmanagerconfig.yaml will not have the receiver defined at all.

**Environment**

* Prometheus Operator version:
0.45.0

* Kubernetes version information:
1.16

* Kubernetes cluster kind:
aks

* Manifests:

```
insert manifests relevant to the issue
```

* Prometheus Operator Logs:

```
insert Prometheus Operator logs relevant to the issue here
```

**Anything else we need to know?**:

Issue #3737, opened 2020-12-10
Make AlertManagerConfig namespace label matching optional / configurable


bshifter:
It is a great thing that we have a validated alertmanager config with possible runtime reconfiguration. It’s much safer than the Helm chart values and easy to manage from 3rd parties.

However there are use-cases where the namespace restriction is not handy:

1. If an alert doesn't have a namespace label. There are already a couple of alerts (maybe from aggregated metrics) that have no namespace labels. In this case none of the alertmanager config will apply to them.
2. Global Kubernetes alerts, where namespace can be anything including future namespaces.
3. External alerts (outside of Kubernetes cluster) where namespace is not applicable at all.

Proposed solution
- Possible attribute to ignore namespace labels
- A singleton global configuration (maybe cluster scoped?)

We are happy to help with the implementation if we have a consensus about the issue.

brancz:
> A singleton global configuration (maybe cluster scoped?)

We haven't done a great job at documenting this, but the secret where we previously configured the entire Alertmanager configuration, still works, and can be used to configure exactly this. I do think though that this is a sub-par experience. I would suggest that we embed an AlertmanagerConfig in the Alertmanager object, where no namespace enforcement is applied. What do you think?

bshifter:
Thank you for the hint, let me take a closer look on the details!

tarokkk:
An AlertmanagerConfig under the Alertmanager object seems reasonable. I really like that we don't have to use raw secret so the input is validated against the Alertmanager config schema ;) Thank you for the quick answer

kragniz:
>I would suggest that we embed an AlertmanagerConfig in the Alertmanager object, where no namespace enforcement is applied. What do you think?

That would be great.

buckner:
Having the choice to disable the namespace label matching would be immensely helpful.

Would embedding an AlertmanagerConfig in the Alertmanager object result in rolling the alertmanager pod(s) on change? An attribute that disables namespace label matching would allow me to deploy a single global AlertmanagerConfig and take advantage of the config-reloader sidecar.

simonpasquier:
I agree that it's a valid request. The primary use case for AlertmanagerConfig was to offer self-service configuration of Alertmanager. That's why the initial implementation didn't change anything wrt the global Alertmanager configuration.

Instead of embedding, maybe we could have a reference to the `AlertmanagerConfig` resource in the same namespace? Especially since AlertmanagerConfig is v1alpha1.

```
[...]

PR #3820, opened 2021-02-02
Generate receivers from alertmanagerconfig even if routes are empty

abhilashjoseph:
This PR is to fix the issue logged here https://github.com/prometheus-operator/prometheus-operator/issues/3819. The Alertmanager config generation ignores the receiver configuration when no routes are present in the AlertManagerConfig definition. This changes the order of generation of routes (to avoid the loop skip, in case route is absent)

```release-note:
[BUGFIX] Generate receivers from alertmanagerconfig even if routes are empty. #3819
```


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.45.1 / 2021-02-01
+
+* [BUGFIX] Generate receivers from alertmanagerconfig even if routes are empty. #3819
```
s-urbaniak, line 3: nit: this is not necessary as we generate the changelog from the PR description.

VERSION
```diff
@@ -1 +1 @@
-0.45.0
+0.45.1
```
s-urbaniak, line 1: see above, the version will be adjusted once we cut a release.

pkg/alertmanager/amcfg.go
```diff
@@ -96,20 +96,20 @@ func (cg *configGenerator) generateConfig(
...
 		for _, receiver := range amConfigs[amConfigIdentifier].Spec.Receivers {
 			receivers, err := cg.convertReceiver(ctx, &receiver, crKey)
 			if err != nil {
 				return nil, errors.Wrapf(err, "AlertmanagerConfig %s", crKey.String())
 			}
 			baseConfig.Receivers = append(baseConfig.Receivers, receivers)
 		}
+
+		// Skip early if there's no route definition.
+		if amConfigs[amConfigIdentifier].Spec.Route == nil {
```
s-urbaniak, line 108: nit: maybe it is clearer if we simply do:
```
if amConfigs[amConfigIdentifier].Spec.Route != nil {
  subRoutes = append(subRoutes, convertRoute(amConfigs[amConfigIdentifier].Spec.Route, crKey, true))
{
```

s-urbaniak (changes requested):
just a few nits and cc @simonpasquier to have another pair of eyes.

s-urbaniak:
@abhilashjoseph thank you for the contribution! :tada: I just modified the PR description to include the change.

s-urbaniak:
another review comment: do you mind to add some unit tests in `amcfg_test.go` as well? thank you!

simonpasquier (commented):
IIUC you want to declare an `AlertmanagerConfig` resource like this:

```yaml
kind: AlertmanagerConfig
metadata:
  name: example
  namespace: default
spec:
  receivers:
  - name: 'webhook'
    webhookConfigs:
    - url: http://example.com/
```
And then reference it in the global Alertmanager config secret:

```
route:
  group_by: ['job']
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 12h
  receiver: default-example-webhook
```

Is this correct? IMO it would be a bit fragile and difficult to maintain because it relies on how the operator generates receiver names. Maybe you're interested by #3737 which aims at leveraging `AlertmanagerConfig` for the global Alertmanager configuration.

CHANGELOG.md
abhilashjoseph: Thank you @s-urbaniak , will remove this

abhilashjoseph:
> IIUC you want to declare an `AlertmanagerConfig` resource like this:
> 
> ```yaml
> apiVersion: monitoring.coreos.com/v1alpha1
> kind: AlertmanagerConfig
> metadata:
>   name: example
>   namespace: default
> spec:
>   receivers:
>   - name: 'webhook'
>     webhookConfigs:
>     - url: http://example.com/
> ```
> 
> And then reference it in the global Alertmanager config secret:
> 
> ```
> route:
>   group_by: ['job']
>   group_wait: 30s
>   group_interval: 5m
>   repeat_interval: 12h
>   receiver: default-example-webhook
> ```
> 
> Is this correct? IMO it would be a bit fragile and difficult to maintain because it relies on how the operator generates receiver names. Maybe you're interested by #3737 which aims at leveraging `AlertmanagerConfig` for the global Alertmanager configuration.

Thanks @simonpasquier, yes that issue is exactly why I ended up in making this change. Currently the mandatory namespace match at first level route, is forcing us to create the alertmanager config at all namespaces and also some alerts don't have the namespace label. The change to get the namespace label configurable would also be more involved because, it needs to deal with duplicate alert manager configs and overlap of alerts in the case where you have global alertmanagerconfig vs alertmanagerconfigs in each ns

VERSION
abhilashjoseph: @s-urbaniak should the go.mod change to the patch version be rolled back as well?

abhilashjoseph:
> another review comment: do you mind to add some unit tests in `amcfg_test.go` as well? thank you!

Added unit test!

abhilashjoseph:
@simonpasquier, @s-urbaniak, please let me know if I should keep this request open, This could be a very short term unblocker for the issue that many of us are facing, while we work on the global alertmanager config changes?

PhilipGough:
@abhilashjoseph - can this be closed now since https://github.com/prometheus-operator/prometheus-operator/commit/d41eb193e5bb250e1a1b0c60a120e4e78ab59765 is merged?

simonpasquier:
closed by #4220

PR #3821, opened 2021-02-02
Global alertmanagerconfig

abhilashjoseph:
This PR is to address the feature request outlined here https://github.com/prometheus-operator/prometheus-operator/issues/3737. This basically adds an additional parameter to enable/disable namespace label match to the first level configuration.

```release-note:
[FEATURE] Make AlertManagerConfig namespace label matching optional / configurable #3737
```


abhilashjoseph:
hello, @s-urbaniak, @simonpasquier, Could I please get a review on this, this is related to the alert manager config optional namespace label match feature. Thank you!

gmedders:
This has worked great for us, thank you @abhilashjoseph!

simonpasquier (commented):
I see where you're coming from but IMO it can't be a property of the `AlertmanagerConfig` resource. One use case we had in mind initially for the CRD was the following: an admin provisions a "shared" `Alertmanager` resource and individual teams are responsible for provisioning their Alertmanager config in their respective namespaces. In which case you want to avoid a team owning namespace A having access to alerts from namespace B (which they don't own).

As I [wrote](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-749570549) in #3737, I'd be ok having a reference to a global `AlertmanagerConfig` resource in `Alertmanager`.

Happy to hear opinions from @prometheus-operator/prometheus-operator-reviewers

abhilashjoseph:
> I see where you're coming from but IMO it can't be a property of the `AlertmanagerConfig` resource. One use case we had in mind initially for the CRD was the following: an admin provisions a "shared" `Alertmanager` resource and individual teams are responsible for provisioning their Alertmanager config in their respective namespaces. In which case you want to avoid a team owning namespace A having access to alerts from namespace B (which they don't own).
> 
> As I [wrote](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-749570549) in #3737, I'd be ok having a reference to a global `AlertmanagerConfig` resource in `Alertmanager`.
> 
> Happy to hear opinions from @prometheus-operator/prometheus-operator-reviewers

Thanks @simonpasquier, I understand the overall goal and future of getting the global configuration and namespaced alerts (owned by different namespace owners), I'm ok to move this config up a level, so that this flag belongs to the spec level (avoid having it at the Route level), since this would help in unblocking the global usage of alert manager and then the parallel, more elaborate use case of having namespaced and global configuration working together and trickling the global configs to namespaces not having overridden configs, to be worked upon.

abhilashjoseph:
Also, if these changes wouldn't adhere to the overall design goals, getting this PR https://github.com/prometheus-operator/prometheus-operator/pull/3820 out, which doesn't interfere with any of the broader goals, would help in unblocking the global usage. thank you!!

bhvishal9:
This would help us immensely, we have lot of namespaces and want to remove the `match` field from default route.

bhvishal9:
Worked perfectly for us too, thanks @abhilashjoseph.

simonpasquier:
@abhilashjoseph would you be able to work on including a "global" AlertmanagerConfig field to the Alertmanager resource as it's been discussed in #3737?

I think it is a more generic solution that would address the same needs. @brancz was also [leaning](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-742560975) towards this solution IIUC.

brancz:
I think there are two things missing today.

1) Roughly what is implemented here, although I think I would prefer if the naming was closer to the `enforcedNamespaceLabel` naming in the Prometheus resource, which has a similar purpose.

2) Either we need an exclusion list like in the Prometheus resource or we directly embed a non enforced config as part of the Alertmanager resource. I think I actually like the ergonomics of the later better but I would understand the argument that many configs may be necessary. (I think I might even be ok with having all of these possibilities)

simonpasquier:
> Either we need an exclusion list like in the Prometheus resource or we directly embed a non enforced config as part of the Alertmanager resource. I think I actually like the ergonomics of the later better but I would understand the argument that many configs may be necessary. (I think I might even be ok with having all of these possibilities)

After reflecting more on the multiple unscoped AlertmanagerConfig use cases, I can understand how it can be useful. If you send alerts from multiple Prometheus instances (or clusters) to a central Alertmanager cluster, you will probably route alerts on labels other than namespace (environment for instance). However maintaining the routing tree in a single `AlertmanagerConfig` resource might become tedious as it grows.

Having said that, I think that embedding/referencing a global AlertmanagerConfig resource in the Alertmanager resource is already a good first step forward.

abhilashjoseph:
> @abhilashjoseph would you be able to work on including a "global" AlertmanagerConfig field to the Alertmanager resource as it's been discussed in #3737?
> 
> I think it is a more generic solution that would address the same needs. @brancz was also [leaning](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-742560975) towards this solution IIUC.

@simonpasquier , I was starting to take a look at it, and figured out that the alertmanager apis are in version monitoring v1alpha1, but referencing those in the base alertmanagerconfig would result in a cyclic dependency (as some of the objects from v1 are referenced in v1alpha1, Is promoting the alertmanagerconfig apis to v1 the recommended solution here?

simonpasquier:
> Is promoting the alertmanagerconfig apis to v1 the recommended solution here?

I don't think we're ready for this.

IIUC you've looked at embedding the AlertmanagerConfig spec into the Alertmanager CRD? What if Alertmanager would reference the AlertmanagerConfig by name (as described [here](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-749570549))?

```
kind: Alertmanager
metadata:
  name: example
spec:
  alertmanagerConfig: # if defined it takes precedence over configSecret
    name: example # name of the AlertmanagerConfig resource in the same namespace used as the root config
    global: {...} # hash of global settings
    templates: [...] # list of template paths
  ...
```

ThiruNandagopal:
@s-urbaniak, @simonpasquier, can we please get this PR merged soon, Thank you.

TechIsCool:
Curious where everyone lands on this. I have been running on a patched version that implements this change since AlertManager went GA. When I originally started working on it I had the same concerns about global access etc. What I thought might be a solution is a feature flag passed into the prometheus-operator `--feature-alert-manager-config-global`.

Also do we have a path forward for this PR?

ThiruNandagopal:
@s-urbaniak, @simonpasquier, can we please get this PR merged soon ? Can you also please provide an ETA ? Thank you.

ThiruNandagopal:
@s-urbaniak, @simonpasquier can we please get this PR merged soon. Thank you.

jjo:
> @s-urbaniak, @simonpasquier can we please get this PR merged soon. Thank you.

Seconding, without being able to omit default per-namespace routing, prometheus-operator is not really usable for cross-namespace teams' workloads (without a lot of CRs duplication).

sudeepgupta90:
Like several others before me, may we please have this PR merged soon? This is a really cool feature I would like to use for my deployments! :)

simonpasquier:
Hey! Sorry for the late reply but this PR doesn't match with what the maintainers have agreed upon in https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-797577859. In practice, we suggest to extend the Alertmanager spec to embed a global AlertmanagerConfig field + other fields for the global and templates parameters which aren't covered by AlertmanagerConfig.
@abhilashjoseph can you modify your PR in this direction?

ThiruNandagopal:
@abhilashjoseph can you please acknowledge.

r0bj:
There is another PR addressing this issue: https://github.com/prometheus-operator/prometheus-operator/pull/4034

sunlintong:
> Hey! Sorry for the late reply but this PR doesn't match with what the maintainers have agreed upon in [#3737 (comment)](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-797577859). In practice, we suggest to extend the Alertmanager spec to embed a global AlertmanagerConfig field + other fields for the global and templates parameters which aren't covered by AlertmanagerConfig.
> @abhilashjoseph can you modify your PR in this direction?
Do you think this is the appropriate way ? If so,  I can open a PR to implement this.

sunlintong:
> Hey! Sorry for the late reply but this PR doesn't match with what the maintainers have agreed upon in [#3737 (comment)](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-797577859). In practice, we suggest to extend the Alertmanager spec to embed a global AlertmanagerConfig field + other fields for the global and templates parameters which aren't covered by AlertmanagerConfig.
> @abhilashjoseph can you modify your PR in this direction?

Do you think this is the appropriate way ? If so, I can open a PR to implement this.

sunlintong:
please see at #4220

sunlintong:
@bergerx Do you mean you are more inclined to use `enforcedNamespaceLabel` in Alertmanager CRD? If so, I'm very pleasure to implement. As I sail at https://github.com/prometheus-operator/prometheus-operator/pull/4220#issuecomment-929928997 . 
```yaml
kind: Alertmanager
metadata:
  name: myalertmanager
spec:
  ...
  enforcedNamespaceLabel: namespace
  alertmanagerConfigsExcludedFromEnforce:
    - configNamespace: ns1
      configName: config1
    - configNamespace: ns2
      configName: config2
```
This way is beautiful and flexible. Maybe I can open a new PR to do this.

bergerx:
I'm personally more inclined to have a `allowGlobalConfigs` in the `Alertmanager` CRD, and then a `globalConfig` field in the `AlertmanagerConfig` CRD. Which would allow us to mix and match both ways in the `AlertmanagerConfig`. Current solutions forces all `AlertmanagerConfig`s to follow either.

But many contributors here have their own opinions and requirements on this, I just feel like the maintainers already have an agreement (https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-797577859) and other methods will likely be silently rejected by no attendance from maintainers. They already seem to be either not really interested in this feature or they don't really have time/brain cycles left for this or have a good enough understanding of the use cases of the consumers.

And the way that I can see out of this current mess (#3820, #3821, #3908, #4033, #4034, #4220) is to aim to implement something that already has the blessing from maintainers. Having some progress here would be more important than doing it all right. And as people have different use cases this may be a really hairy issue to sort out all the fine details.

Alternatively, we can try to fire a new discussion around this in the #prometheus-operator-dev Slack channel or in the upcoming meetings. But I feel like this has already been a hot topic in various places and seems to be exhausting both users and maintainers, having these to reach an already agreed state would already be a win.

sunlintong:
> And many contributors here have their own opinions and requirements on this, I just feel like the maintainers already have an agreement (#3737 (comment)) and other methods will likely be silently rejected by no attendance from maintainers. They already seem to be either not really interested in this feature or they don't really have time/brain cycles left for this or do not have confidence in their understanding of the possible use cases.

I can't agree more. For me, any way is OK. So I implement as https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-797577859. But maintainers seems don't have many time to focus on this question recently. Maybe what we can do now is waiting.

simonpasquier:
Really sorry about the lag on this one but this requires some focus time from the maintainers that nobody had unfortunately. i'll get to it in the coming days and have a proper review of #4220.

achetronic:
> Really sorry about the lag on this one but this requires some focus time from the maintainers that nobody had unfortunately. i'll get to it in the coming days and have a proper review of #4220.

Me and my team would like this code to be merged, sincerely
Thank you

simonpasquier:
#4220 has been merged, it uses a different approach but users can now define their main Alertmanager configuration with the `AlertmanagerConfig` resource.

trexx:
@simonpasquier  That pull request 404's. What's happened to it?

jojotong:
> @simonpasquier That pull request 404's. What's happened to it?

Maybe my account meets some problems. My PR and discussions has been hidden by github, and this is my another account.

bergerx:
Weird that #4220 got missing.

For those who are curious, this is the commit with the merged implementation: https://github.com/prometheus-operator/prometheus-operator/commit/d41eb193e

PR #3822, opened 2021-02-03
feat: configurable thanos sidecar env vars

johanneswuerbach:
With the [jaeger agent running daemonset](https://www.jaegertracing.io/docs/1.21/operator/#installing-the-agent-as-daemonset) it is required to configure the `JAGER_AGENT_HOST` dynamically from the `status.hostIP`.

This PR allows to define custom environment variables for the thanos sidecar container so a `JAGER_AGENT_HOST` can be passed using:

```
env:
  - name: JAEGER_AGENT_HOST
     valueFrom:
       fieldRef:
          fieldPath: status.hostIP
```

```release-note:feature
Allow to provide environment variables to the thanos sidecar.
```


coderanger:
Given the niche nature of this, any reason to not use the `containers` deep merge?

johanneswuerbach:
I had considered it, but this comment seems to recommend against doing that https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/apis/monitoring/v1/types.go#L258

While this feature is fairly niche, I would assume exposing the ability to configure environment variables might also allow other use case so it might be a good compromise?

coderanger:
I mean setting arbitrary env vars could also lead to all kinds of unsupported configurations for the same reason. You break it, you buy it, either way :)

johanneswuerbach:
If that is the recommend path, I'm happily closing this :-)

Issue #3825, opened 2021-02-04
Allow independent setting of requests/limits for CPU/memory in configmap-reloader


JJJJJones:
**What is missing?**

Independent setting of requests/limits for [CPU](https://github.com/prometheus-operator/prometheus-operator/blob/master/cmd/operator/main.go#L173)/[memory](https://github.com/prometheus-operator/prometheus-operator/blob/master/cmd/operator/main.go#L174)

**Why do we need it?**

Not everyone who removes CPU limits (to get rid of annoying CPUThrottlingHigh alerts or ensure that sidecars aren't unnecessarily throttling) wants to throw out CPU requests because they're important in scheduling. Requests and limits aren't always set equivalent. 

**Environment**

* Prometheus Operator version: 0.43.0

**Anything else we need to know?**:

I'm writing up a PR to address this.

**Related links**

Config Reloader added CPU/memory setting with [0.29.0](https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.29.0). [PR](https://github.com/prometheus-operator/prometheus-operator/pull/2403/files)
[Similar](https://github.com/prometheus-operator/prometheus-operator/pull/2144) [headaches](https://github.com/prometheus-operator/prometheus-operator/issues/2171)  
[PR that made the flag also set requests](https://github.com/prometheus-operator/prometheus-operator/pull/2958)
[Work on reducing the CPUThrottlingHigh annoyance with inhibition](https://github.com/prometheus-operator/kube-prometheus/issues/861)

simonpasquier:
We now have dedicated CLI arguments for limits and requests.

PR #3826, opened 2021-02-04
Add flags for CPU/Memory requests/limits

JJJJJones:
Config Reloader's CPU requests/CPU limits and Memory requests/Memory limits are set whatever is given to their respective flags, this allows them to be independently set (EG CPU requests of 100m but no CPU limits) #3825

```release-note:REPLACEME
[FEATURE] Allow Config Reloader CPU requests/limits and memory requests/limits to be set independently #3825 
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -172,11 +172,11 @@ func init() {
 	flagset.StringVar(&cfg.ReloaderConfig.Image, "prometheus-config-reloader", operator.DefaultPrometheusConfigReloaderImage, "Prometheus config reloader image")
 	// TODO(JJJJJones): remove the '--config-reloader-cpu' flag before releasing v0.49.
 	flagset.StringVar(&cfg.ReloaderConfig.CPU, "config-reloader-cpu", "100m", "Config Reloader CPU request & limit. Value \"0\" disables it and causes no request/limit to be configured. Deprecated, it will be removed in v0.49.0.")
-	flagset.StringVar(&cfg.ReloaderConfig.CPURequest, "config-reloader-cpu-req", "", "Config Reloader CPU request. Value \"0\" disables it and causes no request to be configured.")
+	flagset.StringVar(&cfg.ReloaderConfig.CPURequest, "config-reloader-cpu-request", "", "Config Reloader CPU request. Value \"0\" disables it and causes no request to be configured.")
 	flagset.StringVar(&cfg.ReloaderConfig.CPULimit, "confiig-reloader-cpu-limit", "", "Config Reloader CPU limit. Value \"0\" disables it and causes no limit to be configured.")
 	// TODO(JJJJJones): remove the '--config-reloader-memory' flag before releasing v0.49.
 	flagset.StringVar(&cfg.ReloaderConfig.Memory, "config-reloader-memory", "50Mi", "Config Reloader Memory request & limit. Value \"0\" disables it and causes no request/limit to be configured. Deprecated, it will be removed in v0.49.0.")
-	flagset.StringVar(&cfg.ReloaderConfig.MemoryRequest, "config-reloader-memory-req", "", "Config Reloader Memory request. Value \"0\" disables it and causes no request to be configured.")
+	flagset.StringVar(&cfg.ReloaderConfig.MemoryRequest, "config-reloader-memory-request", "", "Config Reloader Memory request. Value \"0\" disables it and causes no request to be configured.")
 	flagset.StringVar(&cfg.ReloaderConfig.MemoryLimit, "config-reloader-memory-limit", "", "Config Reloader Memory limit. Value \"0\" disables it and causes no limit to be configured.")
 	flagset.StringVar(&cfg.AlertmanagerDefaultBaseImage, "alertmanager-default-base-image", operator.DefaultAlertmanagerBaseImage, "Alertmanager default base image (path without tag/version)")
 	flagset.StringVar(&cfg.PrometheusDefaultBaseImage, "prometheus-default-base-image", operator.DefaultPrometheusBaseImage, "Prometheus default base image (path without tag/version)")
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -252,6 +252,14 @@ func Main() int {
 	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
 	logger = log.With(logger, "caller", log.DefaultCaller)
 
+	// logger setup happens after flag parsing!
+	if cfg.ReloaderConfig.CPULimit != "100m" {
+		level.Warn(logger).Log("The --config-reloader-cpu flag has been deprecated! Please use --config-reloader-cpu-limit and --config-reloader-cpu-request according to your use-case.")
+	}
+	if cfg.ReloaderConfig.MemoryLimit != "50Mi" {
+		level.Warn(logger).Log("The --config-reloader-memory flag has been deprecated! Please use --config-reloader-memory-limit and --config-reloader-memory-request according to your use-case.")
+	}
+
 	// Above level 6, the k8s client would log bearer tokens in clear-text.
 	klog.ClampLevel(6)
 	klog.SetLogger(log.With(logger, "component", "k8s_client_runtime"))
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -170,10 +170,12 @@ func init() {
 	// the Prometheus Operator version if no Prometheus config reloader image is
 	// specified.
 	flagset.StringVar(&cfg.ReloaderConfig.Image, "prometheus-config-reloader", operator.DefaultPrometheusConfigReloaderImage, "Prometheus config reloader image")
-	flagset.StringVar(&cfg.ReloaderConfig.CPU, "config-reloader-cpu", "100m", "Config Reloader CPU request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
+	// TODO(JJJJJones): remove the '--config-reloader-cpu' flag before releasing v0.49.
+	flagset.StringVar(&cfg.ReloaderConfig.CPU, "config-reloader-cpu", "100m", "Config Reloader CPU request & limit. Value \"0\" disables it and causes no request/limit to be configured. Deprecated, it will be removed in v0.49.0.")
 	flagset.StringVar(&cfg.ReloaderConfig.CPURequest, "config-reloader-cpu-req", "", "Config Reloader CPU request. Value \"0\" disables it and causes no request to be configured.")
 	flagset.StringVar(&cfg.ReloaderConfig.CPULimit, "confiig-reloader-cpu-limit", "", "Config Reloader CPU limit. Value \"0\" disables it and causes no limit to be configured.")
-	flagset.StringVar(&cfg.ReloaderConfig.Memory, "config-reloader-memory", "50Mi", "Config Reloader Memory request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
+	// TODO(JJJJJones): remove the '--config-reloader-memory' flag before releasing v0.49.
+	flagset.StringVar(&cfg.ReloaderConfig.Memory, "config-reloader-memory", "50Mi", "Config Reloader Memory request & limit. Value \"0\" disables it and causes no request/limit to be configured. Deprecated, it will be removed in v0.49.0.")
 	flagset.StringVar(&cfg.ReloaderConfig.MemoryRequest, "config-reloader-memory-req", "", "Config Reloader Memory request. Value \"0\" disables it and causes no request to be configured.")
 	flagset.StringVar(&cfg.ReloaderConfig.MemoryLimit, "config-reloader-memory-limit", "", "Config Reloader Memory limit. Value \"0\" disables it and causes no limit to be configured.")
 	flagset.StringVar(&cfg.AlertmanagerDefaultBaseImage, "alertmanager-default-base-image", operator.DefaultAlertmanagerBaseImage, "Alertmanager default base image (path without tag/version)")
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -47,6 +47,7 @@ import (
 	"github.com/prometheus/common/version"
 	"golang.org/x/sync/errgroup"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	klog "k8s.io/klog"
 	klogv2 "k8s.io/klog/v2"
 )
@@ -67,6 +68,11 @@ const (
 
 const (
 	defaultOperatorTLSDir = "/etc/tls/private"
+)
+
+const (
+	defaultReloaderCPU    = "100m"
+	defaultReloaderMemory = "50Mi"
 )
 
 var (
@@ -136,7 +142,8 @@ var (
 		logFormatLogfmt,
 		logFormatJson,
 	}
-	cfg = operator.Config{}
+	cfg             = operator.Config{}
+	rcCPU, rcMemory string
 
 	rawTLSCipherSuites string
 	serverTLS          bool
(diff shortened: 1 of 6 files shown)


JJJJJones:
This was made with the assistance of @rmartinez3.

cmd/operator/main.go
```diff
@@ -171,7 +171,11 @@ func init() {
 	// specified.
 	flagset.StringVar(&cfg.ReloaderConfig.Image, "prometheus-config-reloader", operator.DefaultPrometheusConfigReloaderImage, "Prometheus config reloader image")
 	flagset.StringVar(&cfg.ReloaderConfig.CPU, "config-reloader-cpu", "100m", "Config Reloader CPU request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
+	flagset.StringVar(&cfg.ReloaderConfig.CPURequest, "config-reloader-cpu-req", "", "Config Reloader CPU request. Value \"0\" disables it and causes no request to be configured.")
+	flagset.StringVar(&cfg.ReloaderConfig.CPULimit, "confiig-reloader-cpu-limit", "", "Config Reloader CPU limit. Value \"0\" disables it and causes no limit to be configured.")
 	flagset.StringVar(&cfg.ReloaderConfig.Memory, "config-reloader-memory", "50Mi", "Config Reloader Memory request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
+	flagset.StringVar(&cfg.ReloaderConfig.MemoryRequest, "config-reloader-memory-req", "", "Config Reloader Memory request. Value \"0\" disables it and causes no request to be configured.")
```
lilic, line 177: Lets say you use both, we should probably at least write that config-reloader-memory supersedes this flag, correct?

pkg/operator/statefulset.go
```diff
@@ -75,13 +75,38 @@ func CreateConfigReloader(
 		Limits:   v1.ResourceList{},
 		Requests: v1.ResourceList{},
 	}
-	if config.CPU != "0" {
-		resources.Limits[v1.ResourceCPU] = resource.MustParse(config.CPU)
-		resources.Requests[v1.ResourceCPU] = resource.MustParse(config.CPU)
+
+	// Prioritize specific flags over generic ones
```
lilic, line 79: Lets add some tests for these use cases, unit tests is fine maybe we can separate it out to its standalone function, thanks!

pkg/operator/statefulset.go
```diff
@@ -75,13 +75,38 @@ func CreateConfigReloader(
...
+	if config.Memory != "" {
+		if config.MemoryRequest == "" {
+			config.MemoryRequest = config.Memory
+		}
+		if config.MemoryLimit == "" {
+			config.MemoryLimit = config.Memory
+		}
+	}
+
+	if config.CPURequest != "0" {
```
lilic, line 99: Should this not go under the `if config.CPU != ""` case? same for memory below?

lilic (commented):
Thanks for the PR!

@simonpasquier please have a look and what you think of this proposal, thanks!

pkg/operator/config.go
```diff
@@ -45,9 +45,13 @@ type Config struct {
 }
 
 type ReloaderConfig struct {
-	CPU    string
-	Memory string
-	Image  string
+	CPU           string
```
simonpasquier, line 48: I would get rid of the `CPU` field (same for `Memory`) and move the logic of choosing between the global value and request/limit values in main().

cmd/operator/main.go
JJJJJones: I forgot to leave a comment that I was unsure what the desired comment would be. Perhaps "Setting this flag invalidates flag `<relevant_flag>`"?

simonpasquier (commented):
IMO we should support only one way for setting limits/requests eventually so I would recommend to add a deprecation note on the `--config-reloader-cpu` and `--config-reloader-memory` flags.

cc @prometheus-operator/prometheus-operator-reviewers for more input.

pkg/operator/statefulset.go
JJJJJones: Currently, no memory/CPU limits are achieved in the code by not specifying a value for resources.Limits[v1.ResourceMemory] and resources.Limits[v1.ResourceCPU] respectively (as triggered by config.CPU and config.Memory being "0"). So this follows that by not specifying a value for the specific CPU/Memory request/limit if the value in the relevant config is "0".

JJJJJones:
> IMO we should support only one way for setting limits/requests eventually so I would recommend to add a deprecation note on the `--config-reloader-cpu` and `--config-reloader-memory` flags.

@simonpasquier I'm not opposed to that at all and I think that's not the worst possible idea just to avoid potential confusion (as an added benefit it simplifies the code in `pkg/operator/statefulset.go`) but I wasn't sure what was desired and didn't want to start breaking things with behavioral changes so I left those flags in and just wrote code to handle usage of both.

cmd/operator/main.go
```diff
@@ -171,7 +171,11 @@ func init() {
 	// specified.
 	flagset.StringVar(&cfg.ReloaderConfig.Image, "prometheus-config-reloader", operator.DefaultPrometheusConfigReloaderImage, "Prometheus config reloader image")
 	flagset.StringVar(&cfg.ReloaderConfig.CPU, "config-reloader-cpu", "100m", "Config Reloader CPU request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
+	flagset.StringVar(&cfg.ReloaderConfig.CPURequest, "config-reloader-cpu-req", "", "Config Reloader CPU request. Value \"0\" disables it and causes no request to be configured.")
```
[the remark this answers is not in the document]
JJJJJones: We would also want to change the memory request flag to be "config-reloader-memory-request".

I chose to go with req instead of request just to keep the flag length down. Dunno how much concern there is around that. If none, then i'll change the flags to "config-reloader-cpu-request" and "config-reloader-memory-request".

simonpasquier:
@JJJJJones sure no problem, that's the whole purpose of having code reviews :)
We usually support both flags a couple of releases so users can update their manifests and then we remove the deprecated flags (see #3773 and #3457 for instance).

Lets wait for other opinions though.

cmd/operator/main.go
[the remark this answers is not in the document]
lilic: There is no concern, request is better for both agreed.

lilic:
Agreed with Simon, I would vote for deprecation as well, but maybe support for at least two releases. @simonpasquier do you remember what we did with previous flags?

simonpasquier:
yes we said that the deprecated flags would be removed after 2 releases (e.g. deprecation announced in 0.43 and removed in 0.45)

lilic:
Sounds good to me, @JJJJJones do you mind applying the suggestion of marking current flags as deprecated in the flag help text. Unless @prometheus-operator/kube-prometheus-reviewers have any objections.

JJJJJones:
Want me to go ahead and draw up the PR for when those flags flags are removed as well?

lilic:
Opening an issue after we merge this PR and assigning to youself sounds perfect as if we make some changes rebasing might be harder.

s-urbaniak:
Agreed regarding deprecation notice.

pkg/operator/config.go
JJJJJones: I'll make a temporary variable to replace the now removed `cfg.ConfigReloader.CPU` and `cfg.ConfigRelaoder.Memory`.

JJJJJones: A number of tests break with the removal of ReloaderConfig.CPU and ReloaderConfig.Memory. Correcting...

JJJJJones: Fixed.

cmd/operator/main.go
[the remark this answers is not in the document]
JJJJJones: Sounds good, changed both

JJJJJones:
@lilic @s-urbaniak @simonpasquier 

I've added the deprecation notices, updated the flags to request, removed CPU and Memory from `operator.ReloaderConfig`, and moved some of the logic out of `pkg/operator/statefulset.go`. The existing tests have been updated for correctness with the removal of CPU and Memory from `operator.ReloaderConfig`.

lilic:
Nice work and progress, just ping when its ready for final review, thank you! 🎉

JJJJJones:
@lilic the tests have been written up. Let me know if you need anything else.

pkg/alertmanager/statefulset_test.go
```diff
@@ -33,9 +33,11 @@ import (
...
 		ReloaderConfig: operator.ReloaderConfig{
-			Image:  "quay.io/prometheus-operator/prometheus-config-reloader:latest",
-			CPU:    "100m",
-			Memory: "25Mi",
+			Image:         "quay.io/prometheus-operator/prometheus-config-reloader:latest",
+			CPURequest:    "100m",
+			CPULimit:      "100m",
+			MemoryRequest: "50Mi",
+			MemoryLimit:   "50Mi",
 		},
```
JJJJJones, line 41: All tests that specify resources utilize "50Mi" for memory so I've bumped the defaultTestConfig to match.

pkg/thanos/statefulset_test.go
```diff
@@ -64,9 +67,7 @@ func TestStatefulSetLabelingAndAnnotations(t *testing.T) {
 			Labels:      labels,
 			Annotations: annotations,
 		},
-		Spec: monitoringv1.ThanosRulerSpec{
-			QueryEndpoints: emptyQueryEndpoints,
-		},
+		Spec: monitoringv1.ThanosRulerSpec{QueryEndpoints: emptyQueryEndpoints},
```
JJJJJones, line 70: The test files are already aggressively long, may as well stay consistent and opt for one liners where possible.

pkg/prometheus/statefulset_test.go
```diff
@@ -1160,53 +1273,187 @@ func TestSidecarsNoCPULimits(t *testing.T) {
 	}
 }
 
-func TestReplicasConfigurationWithSharding(t *testing.T) {
+func TestSidecarsNoCPUResources(t *testing.T) {
 	testConfig := &operator.Config{
 		ReloaderConfig: operator.ReloaderConfig{
-			CPU:    "0",
-			Memory: "50Mi",
-			Image:  "quay.io/coreos/prometheus-config-reloader:latest",
```
JJJJJones, line 1168: This was the only place with the usage of prometheus-config-reloader from coreos and not prometheus-operator. I assume this was a mistake?

JJJJJones:
@lilic I realize my last message was a little unclear. I'd say the MR is ready for final reviews.

pkg/prometheus/statefulset_test.go
simonpasquier: good catch!

cmd/operator/main.go
```diff
@@ -189,6 +197,26 @@ func init() {
...
+	// Prioritize specific flags over generic ones
+	// If config.CPU has been set but not config.CPURequest/config.CPULimit, default "" to config.CPU
+	// If config.Memory has been set but not config.MemoryRequest or config.MemoryLimit, default "" to config.Memory
+	if rcCPU != "" {
+		if cfg.ReloaderConfig.CPURequest == "" {
+			cfg.ReloaderConfig.CPURequest = rcCPU
+		}
+		if cfg.ReloaderConfig.CPULimit == "" {
+			cfg.ReloaderConfig.CPULimit = rcCPU
+		}
```
simonpasquier, line 210: I would log a deprecation warning if rcCPU is different than `100m` so folks using the flag are potentially aware of the deprecation.

cmd/operator/main.go
```diff
@@ -189,6 +197,26 @@ func init() {
...
+	}
+	if rcMemory != "" {
+		if cfg.ReloaderConfig.MemoryRequest == "" {
+			cfg.ReloaderConfig.MemoryRequest = rcMemory
+		}
+		if cfg.ReloaderConfig.MemoryLimit == "" {
+			cfg.ReloaderConfig.MemoryLimit = rcMemory
+		}
```
simonpasquier, line 218: same here

pkg/operator/statefulset.go
```diff
@@ -75,13 +75,18 @@ func CreateConfigReloader(
 		Limits:   v1.ResourceList{},
 		Requests: v1.ResourceList{},
 	}
-	if config.CPU != "0" {
-		resources.Limits[v1.ResourceCPU] = resource.MustParse(config.CPU)
-		resources.Requests[v1.ResourceCPU] = resource.MustParse(config.CPU)
+
+	if config.CPURequest != "0" {
+		resources.Requests[v1.ResourceCPU] = resource.MustParse(config.CPURequest)
```
simonpasquier, line 80: Side note for a following PR: we should validate the values earlier in the process rather than panic'ing so late :)

simonpasquier (commented):
lgtm, just one ask to add a deprecation warning.

pkg/operator/statefulset.go
JJJJJones: I'm not sure what you mean by panicking. There is no zero/nil/null value for resources. The only way to for a container to have no CPU/memory requests/limits is to not set them. These checks are how we accomplish that.

cmd/operator/main.go
JJJJJones: I can't address those then and there with logger so I baked the warning check into Main(). Let me know if there's a preferred alternative solution.

pkg/operator/statefulset.go
simonpasquier: I mean that [`resource.MustParse()`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#MustParse) will panic if the provided value isn't a valid quantity.

simonpasquier (approved):
lgtm, thanks! I'll let @lilic and @s-urbaniak  a chance to chime in before we merge it.

pkg/operator/statefulset.go
JJJJJones: That makes sense. Want me to add to init() some code to do the check?

It would look like this

```go
	// ...
	flagset.StringVar(&cfg.PromSelector, "prometheus-instance-selector", "", "Label selector to filter Prometheus Custom Resources to watch.")
	flagset.StringVar(&cfg.AlertManagerSelector, "alertmanager-instance-selector", "", "Label selector to filter AlertManager Custom Resources to watch.")
	flagset.StringVar(&cfg.ThanosRulerSelector, "thanos-ruler-instance-selector", "", "Label selector to filter ThanosRuler Custom Resources to watch.")
	flagset.StringVar(&cfg.SecretListWatchSelector, "secret-field-selector", "", "Field selector to filter Secrets to watch")

	// Prioritize specific flags over generic ones
	// If config.CPU has been set but not config.CPURequest/config.CPULimit, default "" to config.CPU
	// If config.Memory has been set but not config.MemoryRequest or config.MemoryLimit, default "" to config.Memory
	if rcCPU != "" {
		if cfg.ReloaderConfig.CPURequest == "" {
			cfg.ReloaderConfig.CPURequest = rcCPU
		}
		if cfg.ReloaderConfig.CPULimit == "" {
			cfg.ReloaderConfig.CPULimit = rcCPU
		}
	}
	if rcMemory != "" {
		if cfg.ReloaderConfig.MemoryRequest == "" {
			cfg.ReloaderConfig.MemoryRequest = rcMemory
		}
		if cfg.ReloaderConfig.MemoryLimit == "" {
			cfg.ReloaderConfig.MemoryLimit = rcMemory
		}
	}
	resource.MustParse(cfg.ReloaderConfig.CPULimit)
	resource.MustParse(cfg.ReloaderConfig.CPURequest)
	resource.MustParse(cfg.ReloaderConfig.MemoryLimit)
	resource.MustParse(cfg.ReloaderConfig.MemoryRequest)
}
```

simonpasquier: It would be great! Instead of MustParse(), you could call ParseQuantity(), log the error and return from Main().

JJJJJones: I'll get the change committed.

JJJJJones: Change has been committed.

s-urbaniak:
lgtm as well, thank you! leaving the final pair of eyes to @lilic

cmd/operator/main.go
```diff
@@ -170,8 +177,14 @@ func init() {
...
-	flagset.StringVar(&cfg.ReloaderConfig.Memory, "config-reloader-memory", "50Mi", "Config Reloader Memory request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
+	// TODO(JJJJJones): remove the '--config-reloader-cpu' flag before releasing v0.48.
+	flagset.StringVar(&rcCPU, "config-reloader-cpu", defaultReloaderCPU, "Config Reloader CPU request & limit. Value \"0\" disables it and causes no request/limit to be configured. Deprecated, it will be removed in v0.48.0.")
+	flagset.StringVar(&cfg.ReloaderConfig.CPURequest, "config-reloader-cpu-request", "", "Config Reloader CPU request. Value \"0\" disables it and causes no request to be configured. Flag overrides `--config-reloader-cpu` value for the CPU request")
+	flagset.StringVar(&cfg.ReloaderConfig.CPULimit, "config-reloader-cpu-limit", "", "Config Reloader CPU limit. Value \"0\" disables it and causes no limit to be configured. Flag overrides `--config-reloader-cpu` for the CPU limit")
+	// TODO(JJJJJones): remove the '--config-reloader-memory' flag before releasing v0.48.
```
lilic, line 184: Same question as above :)

lilic (approved):
Mainly one question that can be done as a follow up! Thank you!!

cmd/operator/main.go
```diff
@@ -170,8 +177,14 @@ func init() {
 	// the Prometheus Operator version if no Prometheus config reloader image is
 	// specified.
 	flagset.StringVar(&cfg.ReloaderConfig.Image, "prometheus-config-reloader", operator.DefaultPrometheusConfigReloaderImage, "Prometheus config reloader image")
-	flagset.StringVar(&cfg.ReloaderConfig.CPU, "config-reloader-cpu", "100m", "Config Reloader CPU request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
-	flagset.StringVar(&cfg.ReloaderConfig.Memory, "config-reloader-memory", "50Mi", "Config Reloader Memory request & limit. Value \"0\" disables it and causes no request/limit to be configured.")
+	// TODO(JJJJJones): remove the '--config-reloader-cpu' flag before releasing v0.48.
```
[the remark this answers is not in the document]
JJJJJones: I'll make one.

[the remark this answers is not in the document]
JJJJJones: #3883

Issue #3829, opened 2021-02-09
[Alertmanager] global: smtp_require_tls configuration flag is not propagated to generated Alertmanager config secret


mdostal-hci:
**What happened?**
I have created a secret "alertmanager-test" with alertmanager.yaml contents:
```
global:
  smtp_require_tls: false
  resolve_timeout: 7m
  smtp_from: alertmanager@sampledomain.com
  smtp_hello: localhost
  smtp_smarthost: localhost:25

route:
  receiver: "null"
receivers:
- name: "null"
```

and appropriate alertmanager config object with reference to "configSecret: alertmanager-test". It was parsed by prometheus-operator to "alertmanager-generated" secret, but generated secret was missing "global: smtp_require_tls" configuration option.

**Did you expect to see something different?**
global:smtp_require_tls option should be present in generated secret.

**How to reproduce it (as minimally and precisely as possible)**:
See above, I will prepare PR with failing test.

**Environment**

* Prometheus Operator version:
v0.45.0

* Kubernetes version information:
version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.6", GitCommit:"d32e40e20d167e103faf894261614c5b45c44198", GitTreeState:"clean", BuildDate:"2020-05-20T13:08:34Z", GoVersion:"gohttps://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack1.13.9", Compiler:"gc", Platform:"linux/amd64"}

* Kubernetes cluster kind:
RKE

* Relevant manifests
```
apiVersion: v1
items:
- apiVersion: monitoring.coreos.com/v1
  kind: Alertmanager
  metadata:
    annotations:
      meta.helm.sh/release-name: eit
      meta.helm.sh/release-namespace: monitoring
    labels:
      app: kube-prometheus-stack-alertmanager
      app.kubernetes.io/managed-by: Helm
      chart: kube-prometheus-stack-13.2.1
      heritage: Helm
      release: eit
    name: kube-prometheus-stack-alertmanager
    namespace: monitoring
  spec:
    alertmanagerConfigNamespaceSelector: {}
    alertmanagerConfigSelector: {}
    configSecret: alertmanager-test
    externalUrl: http://alertmanager.somedomain.com/
    image: quay.io/prometheus-alertmanager:v0.21.0
    listenLocal: false
    logFormat: logfmt
    logLevel: info
    paused: false
    portName: web
    replicas: 1
    resources:
      limits:
[...]

PR #3830, opened 2021-02-09
Added test for smtp_require_tls global flag

mdostal-hci:
Added failing test for alertmanager configuration with "global: smtp_require_tls" flag. See #3829 for more details.

release-note:none

### DIFF
--- a/pkg/alertmanager/amcfg_test.go
+++ b/pkg/alertmanager/amcfg_test.go
@@ -61,6 +61,27 @@ templates: []
 `,
 		},
 		{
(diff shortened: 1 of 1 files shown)


simonpasquier:
Thanks for the PR and adding a test! To fix it, I assume that we need to turn this [bool](https://github.com/prometheus-operator/prometheus-operator/blob/f689892994656dda6de2aaf239a20b008b87db79/pkg/alertmanager/types.go#L53) into a *bool.

Issue #3850, opened 2021-02-18
Add remoteWrite custom HTTP headers


KrotanHill:
**What is missing?**
The possibility to add the remoteWrite custom HTTP headers introduced in Prometheus v2.25.0 (see [Changelog](https://github.com/prometheus/prometheus/releases/tag/v2.25.0) or [Updated Configuration](https://prometheus.io/docs/prometheus/2.25/configuration/configuration/#remote_write)). New Config Option:
```
# Custom HTTP headers to be sent along with each remote write request.
# Be aware that headers that are set by Prometheus itself can't be overwritten.
headers:
  [ <string>: <string> ... ]
```

**Why do we need it?**
To support the new Prometheus config option.

**Anything else we need to know?**:
I implemented the config extension and will create a Pull request for it.

PR #3851, opened 2021-02-19
Add remoteWrite custom HTTP headers

KrotanHill:
Adds the possibility to configure the custom HTTP headers introduced in Prometheus v2.25.0
Implements #3850

I tested this on a cluster sending a tenantID to [cortex](https://github.com/cortexproject/cortex) and it worked.

I tried to work as written in the documents. Pleas tell me, if I made something wrong.

```release-note:enhancement
Add custom HTTP headers in remoteWrite-config
```


pkg/prometheus/promcfg.go
```diff
@@ -1401,6 +1401,10 @@ func (cg *configGenerator) generateRemoteWriteConfig(version semver.Version, p *
 			{Key: "remote_timeout", Value: spec.RemoteTimeout},
 		}
 
+		if len(spec.Headers) > 0 {
```
brancz, line 1404: We need to protect this if additionally with the version being >=2.25.0

brancz (commented):
Just one small thing on compatibility, otherwise lgtm!

pkg/prometheus/promcfg.go
KrotanHill: Makes sense. Code is updated.

brancz:
Thank you!

PR #3855, opened 2021-02-19
Remove deprecation warnings for pod and service monitor selectors

### LINKED ISSUE #3854: Fields incorrectly marked as deprecated
**What happened?**

`podMonitorSelector` and `serviceMonitorSelector` are being marked as deprecated entirely by tools because of their comments. We should change this comment to just write out the behavior instead of using the typical deprecated format.

These fields are not deprecated entirely, just a certain behavior, so only that should be called out and not the entire field be deprecated.

Use the Jetbrains Kubernetes plugin (in GoLand) to write/edit a Prometheus resource

miguelvr:
closes #3854 

```release-note:bug
Removed deprecation warning on podMonitorSelector and serviceMonitorSelector, in PrometheusSpec, which could be a source of confusion to users.
```


pkg/apis/monitoring/v1/types.go
```diff
@@ -86,15 +86,15 @@ type PrometheusList struct {
 type PrometheusSpec struct {
 	// PodMetadata configures Labels and Annotations which are propagated to the prometheus pods.
 	PodMetadata *EmbeddedObjectMetadata `json:"podMetadata,omitempty"`
-	// ServiceMonitors to be selected for target discovery. *Deprecated:* if
+	// ServiceMonitors to be selected for target discovery. *Warning:* if
```
brancz, line 89: Hmm, I'm not sure this is exactly what we want to get across. The behavior of leaving both values unset allowing the entire config to be managed is what is indeed deprecated. It's just that not the entire field is deprecated, just that one behavior.

miguelvr: should that be done in the validating webhook instead, then?

brancz: Well, it's only deprecated, it still works as we speak. When we remove the behavior entirely we'll remove this note.

lastthyme (approved):

uhthomas:
Hi!

It would be great if we could get some movement on this PR.

Reading the [PrometheusSpec docs](https://prometheus-operator.dev/docs/operator/api/#prometheusspec) was very confusing as guides use both `podMonitorSelector` and `serviceMonitorSelector` but the API docs declare them both as deprecated.

I had to search the GitHub issues specifically to find out that they are in fact _not_ deprecated.

Thank you

nabokihms:
Plus, one to this issue. It is still valid. How can we move this forward?

pkg/apis/monitoring/v1/types.go
rexagod: How about `s/Warning/NOTE`? This should suffice the use case we're aiming to target, i.e., alerting the reader in some manner while avoiding false positives by the different tools that parse this information.

Issue #3835, opened 2021-02-12
[AlertManager] EmailConfig.headers key lost


Shimokuni:
**What happened?**

`EmailConfig.headers` key will be lost even if inserted any value.

I think the cause is that `strings.Title(key)` always returns empty string. Parhaps `strings.Title(d.Key)`？

https://github.com/prometheus-operator/prometheus-operator/blob/v0.44.0/pkg/alertmanager/amcfg.go#L727-L729

So, in addtion, if multiple `EmailConfig.headers` are defined, the job of reloading AlertManagerConfig will fail with error message: `duplicate header \"\" in email config`.

**Did you expect to see something different?**

* The job of reloading AlertManagerConfig succeeeds.
* `EmailConfig.headers` keys are defined

**How to reproduce it (as minimally and precisely as possible)**:

apply the manifest below.

**Environment**

* Prometheus Operator version:

    `0.44.0`
    

* Kubernetes version information:

    `Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}`
    

* Kubernetes cluster kind:

    on-premise cluster created by kubeadm

* Manifests:

```yaml
apiVersion: monitoring.coreos.com/v1alpha1
kind: AlertmanagerConfig
metadata:
  name: test
  labels:
    alertmanager: test
spec:
  route:
    groupWait: 15s
    groupInterval: 5m
    repeatInterval: 3h
    receiver: test
  receivers:
  - name: test
    emailConfigs:
    - to: reciever@example.com
      smarthost: smtp.example.com:25
      from: sender@example.com
      requireTLS: false
      sendResolved: true
      headers:
      - key: Subject
        value: Subject
      - key: Comment
        value: Comment
  inhibitRules:
  - sourceMatch:
    - name: severity
      value: emergency
      regex: false
    targetMatch:
[...]

PR #3856, opened 2021-02-20
fix the loss of EmailConfig.headers key

Shimokuni:
closes: https://github.com/prometheus-operator/prometheus-operator/issues/3835

```release-note:bug
fix the loss of EmailConfig.headers key
```


simonpasquier (approved):
Thanks!

PR #3865, opened 2021-02-23
Add Giant Swarm as an adopter

QuentinBisson:
Adds Giant Swarm to the Adopters list.

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -40,6 +40,20 @@ Details (optional):
 - 25000 samples/s
 - 1.2M active series
 
+## Giant Swarm
+
+https://www.giantswarm.io/
+
+Environments: AWS, Azure, Bare Metal
+
+Uses [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus): Yes (with additional tight Giant Swarm integrations)
+
+Details:
+- One prometheus operator per management cluster and one prometheus instance per workload cluster
+- Customers can also install kube-prometheus for their workload using our App Platform
+- 760000 samples/s
+- 35M active series
+
 ## <Insert Company/Organization Name>
 
 https://our-link.com/


ADOPTERS.md
```diff
@@ -40,6 +40,20 @@ Details (optional):
 - 25000 samples/s
 - 1.2M active series
 
+## Giant Swarm
+
+https://www.giantswarm.io/
+
+Environments: AWS, Azure, Bare Metal
+
+Uses [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus): Yes (with additional tight Giant Swarm integrations)
```
lilic, line 49: Nice, we are looking forward if that is something you can bring into kube-prometheus or needs to be on Giant Swarm level?

ADOPTERS.md
```diff
@@ -40,6 +40,20 @@ Details (optional):
...
+
+Details:
+- One prometheus operator per management cluster and one prometheus instance per workload cluster
+- Customers can also install kube-prometheus for their workload using our App Platform
+- 760000 samples/s
```
lilic, line 54: Very cool!! 🎉

lilic (approved):
Thank you!! 🎉

ADOPTERS.md
QuentinBisson: Our fork has minimal changes and are mostly on the Giant Swarm level. We tend to provide things we need/want upstream directly. 
For example, we disable app armor by default in the values file, add some files for our UI and we also retag the upstream images to be able to use them in China (we push them to a registry there to avoid timeouts during pulls). 

With the recent PR to be able to change the default image used by the prometheus operator, this part will be easier to change :)

You can find our version here https://github.com/giantswarm/prometheus-operator-app

PR #3879, opened 2021-02-25
Reconcile resources on namespace updates 

### LINKED ISSUE #3847: serviceMonitorNamespaceSelector not working correctly
**What happened?**

I need to activate and deactivate metric recollection at the namespace level.
I am using (trying) a label on every namespace I want to monitor with a specific Prometheus instance. 
Also, I need to monitor just a subset of ServiceMonitor (those with the `example.com.app: app-1` label).

I am using the following Prometheus definition that is supposed to implement the usecase. 

```yaml
---
kind: Prometheus
metadata:
  labels:
    example.com.app: app-1
    example.com.component: prometheus
  name: app-1
  namespace: monitoring-app-1
spec:
  podMetadata:
    labels:
      example.com.app: app-1
      example.com.component: prometheus
    matchLabels:
      monitoring: "true"
  serviceMonitorSelector:
    matchLabels:
      example.com.app: app-1
```

So, I can have multiple namespaces, some of them with the `monitoring: "true"` label. 
On the other side, I can have multiple ServiceMonitors with the `example.com.app: app-1` label in multiple namespaces (with or without the `"monitoring: true"` label).

I want to monitor only the ServiceMonitors with the label `example.com.app: app-1` inside namespaces having the `monitoring: true` label. In addition, I can activate/deactivate the monitoring of the ServiceMonitor changing the namespace label from `true` to `false` and the other way round.

**What currently happens is that changing the `namespace` label does not trigger any reconfiguration on the Prometheus instance.** So I can not activate/deactivate the metrics.

I expect to have the Prometheus reconfigured if a namespace label changed as it is part of the `serviceMonitorNamespaceSelector` Prometheus definition.

I tried to simplify the setup as much as possible:

```bash
$ kind create cluster --name prometheus-operator-issue
# Deploy the prometheus-operator
$ kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.45.0/bundle.yaml
# Deploy the usecase, review the code first, security matters 
$ kubectl apply -f https://gist.githubusercontent.com/angelbarrera92/9619ff72607920003fdc7153e8b30a4f/raw/16e06f9e4d1f57ebbd05126a6f23d5df4dfb0b01/prometheus-operator-namespace-issue.yaml
```

Then open a port-forward to the Prometheus UI:

```bash
$ kubectl port-forward -n monitoring-app-1 prometheus-app-1-0 9090:9090
```

And access it with a browser: http://localhost:9090/targets

Captura de pantalla 2021-02-18 a las 16 38 31

Then, try to deactivate the monitoring on the `monitoring-app-1` namespace by changing its label:

```bash
$ kubectl label ns monitoring-app-1 --overwrite monitoring=false
namespace/monitoring-app-1 labeled
```

Check the UI and wait a few minutes, nothing happens.

Captura de pantalla 2021-02-18 a las 16 40 37

**!!BUT!!** if you add a label (or any other change) to the ServiceMonitor, it will trigger a reconfiguration of the Prometheus server disabling the ServiceMonitor (as it was inside the namespace we changed the label above)

```bash
$ kubectl -n monitoring-app-1 label ServiceMonitor prometheus --overwrite update=1
servicemonitor.monitoring.coreos.com/prometheus labeled
```

After a few seconds/minutes, the ServiceMonitor disappear as it was inside the *deactivated* namespace:

Captura de pantalla 2021-02-18 a las 16 43 24

**Environment**

    `v0.45.0`

    `Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:12:48Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.1", GitCommit:"206bcadf021e76c27513500ca24182692aabd17e", GitTreeState:"clean", BuildDate:"2020-09-14T07:30:52Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}`

* Kubernetes cluster kind:

    `kind`

* Manifests:

- https://gist.githubusercontent.com/angelbarrera92/9619ff72607920003fdc7153e8b30a4f/raw/16e06f9e4d1f57ebbd05126a6f23d5df4dfb0b01/prometheus-operator-namespace-issue.yaml

* Prometheus Operator Logs:

Nothing really relevant happens.

It smells like the operator does not listen for `namespace` events.

simonpasquier:
Say that you have a Prometheus spec with
`serviceMonitorNamespaceSelector: {monitored: "true"}` and
`serviceMonitorSelector: {}` as well as a service monitor in namespace
"foo" which has the `monitored="true"` label. Then the operator wouldn't
remove the service monitor's targets when the `monitored="true"` label
is removed from namespace "foo" as one would expect. If namespace "foo"
doesn't have the selecting label initially, adding it later wouldn't
trigger the reconciliation either.

The same issue existed for pod monitors, probes, rules (Prometheus and
Thanos ruler) and alertmanager configurations.

This change adds logic to deal with namespace updates and trigger the
reconciliation if the namespace's labels have changed in a way that may
return different resources than previously. However it works only when
the operator uses a privileged lister/watcher on namespaces (e.g.  when
it isn't restricted to a set of allowed namespaces). In the other case,
the unprivileged lister/watcher doesn't propagate namespace updates as
expected, leaving the operator out-of-sync. This will be tackled in a
following change.

Relates #3847 

```release-note:bug
Reconcile resources on namespace updates when using privileged lister/watcher
```

### DIFF
--- a/pkg/k8sutil/labels.go
+++ b/pkg/k8sutil/labels.go
@@ -1,0 +1,44 @@
+// Copyright 2021 The prometheus-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8sutil
+
+import (
+	"reflect"
+
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// LabelSelectionHasChanged returns true if the selector doesn't yield the same results
+// for the old and current labels.
+func LabelSelectionHasChanged(old, current map[string]string, selector *metav1.LabelSelector) (bool, error) {
+	// If the labels haven't changed, the selector won't return different results.
+	if reflect.DeepEqual(old, current) {
+		return false, nil
+	}
+
+	sel, err := metav1.LabelSelectorAsSelector(selector)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to convert selector %q", selector.String())
+	}
+
+	// The selector doesn't restrict the selection thus old and current labels always match.
+	if sel.Empty() {
+		return false, nil
+	}
+
+	return sel.Matches(labels.Set(old)) != sel.Matches(labels.Set(current)), nil
+}
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -308,6 +308,15 @@ func (c *Operator) addHandlers() {
 		DeleteFunc: c.handleStatefulSetDelete,
 		UpdateFunc: c.handleStatefulSetUpdate,
 	})
+
+	// The controller needs to watch the namespaces in which the
+	// alertmanagerconfigs live because a label change on a namespace may
+	// trigger a configuration change.
+	// It doesn't need to watch on addition/deletion though because it's
+	// already covered by the event handlers on alertmanagerconfigs.
+	c.nsAlrtCfgInf.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: c.handleNamespaceUpdate,
+	})
 }
 
 func (c *Operator) handleAlertmanagerConfigAdd(obj interface{}) {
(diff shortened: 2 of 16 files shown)


simonpasquier:
> the unprivileged lister/watcher doesn't propagate namespace updates as
expected, leaving the operator out-of-sync. This will be tackled in a
following change.

to be more precise,

simonpasquier:
> the unprivileged lister/watcher doesn't propagate namespace updates as
expected, leaving the operator out-of-sync. This will be tackled in a
following change.

To be more precise, the operator will use an unprivileged lister/watcher when the namespaces to be watched is restricted by an allow-list.

https://github.com/prometheus-operator/prometheus-operator/blob/4826c904d86c93cffaed293c5ce85ad6b2051995/pkg/prometheus/operator.go#L284-L305

The implementation of the unprivileged lister/watcher uses a custom List() function which gets every individual namespace and collates them in a list. The Watch() function is a no-op because it isn't possible to watch individual items (https://github.com/kubernetes/kubernetes/issues/14289#issuecomment-146952401).

https://github.com/prometheus-operator/prometheus-operator/blob/4826c904d86c93cffaed293c5ce85ad6b2051995/pkg/listwatch/listwatch.go#L77-L105

The assumption has been that by lowering the sync period of the lister/watcher to 15 seconds, the informer would refresh the list of namespaces at this interval. Unfortunately it isn't happening because resync != relist. IIUC relist happens only when the watch function returns an error which is never the case here.

To fix it, we could:
* Check whether the service account has list/watch permissions on the namespace.
* If not, implement a poll-based lister/watcher that would relist allowed namespaces periodically and emit appropriate watch events (watch.ProxyWatcher might be useful).
* Otherwise use a normal lister/watcher with field selectors to restrict the list of watched namespaces (the idea being that it would more efficient that the previous option).

pkg/k8sutil/labels.go
```diff
@@ -0,0 +1,44 @@
...
+	"reflect"
+
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// LabelSelectionHasChanged returns true if the selector doesn't yield the same results
+// for the old and current labels.
+func LabelSelectionHasChanged(old, current map[string]string, selector *metav1.LabelSelector) (bool, error) {
```
s-urbaniak, line 27: suggestion (while I was reading it): let's call `selector` `namespaceSelector`, so the context is more clear to the reader.

s-urbaniak: otoh the method is quite ubiquitos, this is fine then.

s-urbaniak:
> Otherwise use a normal lister/watcher with field selectors to restrict the list of watched namespaces (the idea being that it would more efficient that the previous option).

this seems to be the best option, however I think field selectors work on client side only, so the lister/watcher would need to be privileged.

Anyways, this is a great initial fix @simonpasquier ! We should create a follow-up ticket to track fixing the priviledged lister/watcher.

lgtm from my side :+1:

pkg/k8sutil/labels.go
simonpasquier: this function is generic and doesn't care to which resource the labels apply (e.g. it can be namespace, pod, ...).

simonpasquier:
> We should create a follow-up ticket to track fixing the priviledged lister/watcher.

The original issue is still opened so we don't forget about fixing the **unprivileged** lister/watcher case.

Issue #3892, opened 2021-03-04
ServiceMonitor endpoint tls config missing in "tls-assets" when read from ConfigMap or Secret


Riohyc:
**What happened?**
When ServiceMonitors in different Namespace all have endpoints use tlsConfig from Secret or ConfigMap. Only one Namespace's ServiceMonitor can scrap targets by tls cert data while others will get scrap error of missing tls data like this:
```
Get "https://192.168.2.222:2379/metrics": unable to load specified CA cert /etc/prometheus/certs/nodeless-worker-2_cert-etcd-client_ca.crt: open /etc/prometheus/certs/nodeless-worker-2_cert-etcd-client_ca.crt: no such file or directory
```

**Did you expect to see something different?**
All ServiceMonitor and Targets should not get error because of missing tls data file.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create ServiceMonitors in different Namespace which all have endpoints use tlsConfig from Secret or ConfigMap.
2. Make sure ServiceMonitors can be successfully selected by your Prometheus instance.
3. You will find scrap error in prometheus web UI for all Targets expect one namespace.
4. Check the Secret with name `{Prometheus Instance Namesapce}/prometheus-{Prometheus Instance Name}-tls-assets` which is mounted by Prometheus Pod as tls cert data, you will find ONLY ONE Namespace's tls data is stored in it.

**Environment**
Linux, Debain10
* Prometheus Operator version:
v0.41.1

* Kubernetes version information:
1.15.3
    `kubectl version`
    

* Kubernetes cluster kind:
kubeadm

* Manifests:
None

* Prometheus Operator Logs:
None

**Anything else we need to know?**:
Only happend in `v0.41.1` or below because since `v0.42` operator use `store` to load tls assets with different logic.

Riohyc:
Bug is cause by code below:
https://github.com/prometheus-operator/prometheus-operator/blob/312d675008306b13c24d241bf4f0a882dbfa90d8/pkg/prometheus/operator.go#L1684-L1724

`smAssets` is always overrided by last ServiceMonitor in loop.

PR #3893, opened 2021-03-04
Fix ServiceMonitor endpoint tls config misconfiguration when data is read from Secret or ConfigMap in different Namespace

Riohyc:
Fix: #3892


dgrisonnet:
This bug should be fixed in prometheus-operator v0.42.0, please consider upgrading.

Also, considering that we are already at version v0.46 of prometheus-operator, I don't think it is reasonable to backport fixes to v0.41.

Riohyc:
> This bug should be fixed in prometheus-operator v0.42.0, please consider upgrading.
> 
> Also, considering that we are already at version v0.46 of prometheus-operator, I don't think it is reasonable to backport fixes to v0.41.

Thanks, upgrade to 0.45.

PR #3894, opened 2021-03-04
add rulefmt check to po-lint

mnacharov:
Fixes: #3147

Check prometheus rules in CRD PrometheusRules with promtool lib


brancz:
Super cool! Thank you! :+1: :tada:

PR #3902, opened 2021-03-08
Fix incomplete change in #3834

mrueg:
Unfortunately the change I made was incomplete and caused a inconsistency. It should be managed-by and not part-of for endpoints as well.
Ref: https://github.com/prometheus-operator/prometheus-operator/pull/3834


simonpasquier (approved):
lgtm. Leaving a final say to @paulfantom

metalmatze (approved):

metalmatze:
Seems like the docs need to be newly generated again. 
```
make format generate
```

mrueg:
> Seems like the docs need to be newly generated again.
> 
> ```
> make format generate
> ```

I forgot to `go fmt` it. should be working now.

PR #3904, opened 2021-03-09
alertmanager: add clusterListenLocal flag for config cluster.listen-address

### LINKED ISSUE #3857: Add support to change the `--cluster.listen-address` Alertmanage flage
**What is missing?**
A way to change the `--cluster.listen-address` AlertManager flag. The value is hard set to the pod IPs when there multiple replicas are deployed.

**Why do we need it?**
The current setup doesn't work well with Istio. Since all communication happens on the localhost network (127.0.0.1). This results in a `connect reset by peer` error from each peers.

**Environment**

    `0.45.0`
    

As a temporary fix, I have paused the `monitoring.coreos.com/v1.AlertManager` object, then edit the statefulset and update the flag to  `--cluster.listen-address=0.0.0.0:9094`.

rgynn:
alertmanager: add clusterListenLocal to enable flag cluster.listen-address be used with pod loopback address

enables the use of alertmanager in servicemesh like istio

Fixes #3857

```release-note:REPLACEME

```


rgynn:
Sorry for the messy PR, my github skills have gone rusty lately.

**Summary:**

Added a flag called clusterListenLocal that changes the `--cluster.listen-address` to 127.0.0.1:9094 for alertmanager.

Thought first that i could hijack the listenLocal flag the alertmanager spec already has (which only sets the web listener to loopback).
But changed my mind as to not brake any current setup behaviors. 
I do not know why you made the distinction as to why only the web listener was bound to loopback.

I also updated the `stateful_test.go` file with a table of testcases as to make sure that the old behavior was not broken.
And it sort of helps in getting a better overview of the matrix of flags in the alertmanager spec affecting one single cli flag.

**Why?**

Just like OP said previously, this enables us to bootstrap alertmanager easier in a service mesh like istio.

Let me know if I can help in any other way!

Best regards

pkg/alertmanager/statefulset.go
```diff
@@ -234,9 +234,17 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config) (*appsv1.S
 	}
 
 	if *a.Spec.Replicas == 1 && !a.Spec.ForceEnableClusterMode {
-		amArgs = append(amArgs, "--cluster.listen-address=")
+		if a.Spec.ClusterListenLocal {
+			amArgs = append(amArgs, "--cluster.listen-address=127.0.0.1:9094")
+		} else {
+			amArgs = append(amArgs, "--cluster.listen-address=")
+		}
```
simonpasquier, line 241: In this case, we don't want to listen on gossip traffic at all.

```suggestion
		amArgs = append(amArgs, "--cluster.listen-address=")
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1304,6 +1304,8 @@ type AlertmanagerSpec struct {
 	PriorityClassName string `json:"priorityClassName,omitempty"`
 	// AdditionalPeers allows injecting a set of additional Alertmanagers to peer with to form a highly available cluster.
 	AdditionalPeers []string `json:"additionalPeers,omitempty"`
+	// ClusterListenLocal makes the Alertmanager server listen on loopback for gossip communication
```
simonpasquier, line 1307:
```suggestion
	// ClusterListenLocal makes the Alertmanager server listen on loopback for cluster communication
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1304,6 +1304,8 @@ type AlertmanagerSpec struct {
 	PriorityClassName string `json:"priorityClassName,omitempty"`
 	// AdditionalPeers allows injecting a set of additional Alertmanagers to peer with to form a highly available cluster.
 	AdditionalPeers []string `json:"additionalPeers,omitempty"`
+	// ClusterListenLocal makes the Alertmanager server listen on loopback for cluster communication
```
simonpasquier, line 1307:
```suggestion
	// ClusterListenLocal makes the Alertmanager server listen on loopback for cluster communication.
```

pkg/alertmanager/statefulset_test.go
```diff
@@ -918,81 +918,96 @@ func TestTerminationPolicy(t *testing.T) {
...
+			ClusterListenLocal:     false,
+			ExpectedResult:         "--cluster.listen-address=[$(POD_IP)]:9094",
+		},
+		{
+			Name:                   "ClusterListenAddressForSingleReplicaWithClusterListenLocal",
+			Replicas:               1,
+			ForceEnableClusterMode: false,
+			ClusterListenLocal:     true,
+			ExpectedResult:         "--cluster.listen-address=",
+		},
```
simonpasquier, line 952: For completeness, don't we miss an additional test case with replicas=1, ForceEnableClusterMode=true and ClusterListenLocal=true?

rgynn: Added testcase now in the latest commit.
https://github.com/rgynn/prometheus-operator/blob/2686a1cfc4c51b16db697ee25ab27afbf140b56e/pkg/alertmanager/statefulset_test.go#L973-L979

simonpasquier:
So we have 2 approaches to solve the same use case apparently: this PR and #4038. @rgynn and @desaintmartin can you evaluate the advantages/drawbacks of each solution? I'm afraid that I don't grasp all the details about what Istio can support or not...

howardjohn:
This is not needed in Istio 1.10+, and will actually break in Istio 1.10. See https://istio.io/latest/blog/2021/upcoming-networking-changes/ for more details.

desaintmartin:
Exactly, this works in istio 1.9 but breaks in istio >= 1.10.

dmitriyr333:
FWIW, using **Istio v1.10.3** I was able to route traffic from Prom to Alertmanager with [Sidecar](https://istio.io/latest/docs/reference/config/networking/sidecar/#IstioIngressListener) resource. 
Note that I have set `listenLocal: false`
```
apiVersion: networking.istio.io/v1beta1
kind: Sidecar
metadata:
  namespace: prom-system
  name: prometheus
spec:
  workloadSelector:
    labels:
      app: prometheus
  egress:
  - port:
      number: 9093
      protocol: TCP
      name: web
    captureMode: NONE
    bind: 0.0.0.0
    hosts:
    - "*/prometheus-operator-kube-p-alertmanager.prom-system.svc.cluster.local"
```
Not sure if that is a good workaround

howardjohn:
@dmitriyr333 I am confused why you need the sidecar on 1.10 given you have `listenLocal: false`?

Happy to  look into it on my own cluster if you let me know how to reproduce (I am familiar with Istio but not alertmanager)

howardjohn:
Oh interesting. That is totally different from the issue discussed I think. The problem is the port is like:
```
  - name: web
    port: 9093
    protocol: TCP
    targetPort: web
```
. Your sidecar fixes it by declaring the protocol TCP

dmitriyr333:
@howardjohn you might be correct that I'm barking up the wrong issue here
but just for context:
- I'm using most of the default options from [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) chart
- without Istio sidecars enabled, Prometheus has no problems communicating to Alertmanager on the default port + Protocol
- with Istio sidecars added, I'm getting 404s
- Istio Sidecar resource for Prom egress helped me solved this

howardjohn:
I totally agree on that, just that it's not caused by the listen address
but rather HTTP vs TCP routing semantics (the former requires a host header
in Istio)

howardjohn:
@dmitriyr333 we are going a bit off topic here, but I am prety sure we can make that work out of the box, opened https://github.com/istio/istio/issues/34679 to track this

dmitriyr333:
> @dmitriyr333 we are going a bit off topic here, but I am prety sure we can make that work out of the box, opened [istio/istio#34679](https://github.com/istio/istio/issues/34679) to track this

Thank you!

PR #3906, opened 2021-03-10
pkg/prometheus: remove dead code

jan--f:
Removes setup code for the readiness handler that gets immediately
reassigned. The setup targets prometheus' /-/healthy/ endpoint and would
be a bad readiness probe anyway.

Signed-off-by: Jan Fajerski <jfajersk@redhat.com>

Really just a minor cleanup.

```release-note:REPLACEME

```


simonpasquier (approved):
thanks!

PR #3907, opened 2021-03-11
Update Veepee on ADOPTER.md

nerzhul:
Previously we only put one Kubernetes cluster.
Now we publish stats for all our clusters

```release-note:REPLACEME

```


lilic (approved):
Wow impressive, thanks for the update!

Issue #3908, opened 2021-03-11
AlertmanagerConfig routes combined into a central Alertmanager do not respect `continue: false` under `routes`


jandersen-plaid:
**What happened?**

We are currently templating out alerts for different teams to take on alerts from different namespaces. As part of this, we have placed an AlertmanagerConfig in each namespace that represents routing instructions for the team (as well as a central secret that contains more generalized or cross namespace routing rules).

While defining one of these rules, we set `continue: false` on an alert with one route, expecting no more rules to be processed after this when it gets generated in the central alertmanager. Since the generation adds the route as a second level child route, we assumed this would be fine. Instead, the alertmanager config generation overrode `continue: false` to be `true` (see [this line for the specific snippet that enables it and some comments](https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/alertmanager/amcfg.go#L104)) instead of respecting the explicitly set option on the AlertmanagerConfig config.

**Did you expect to see something different?**

We expected to see the `continue: false` make it to the central configuration and be respected during generation instead of being overriden. As per the [Alertmanager documentation](https://prometheus.io/docs/alerting/latest/configuration/#route), there is no need to enforce top-level nodes (though this was a secondary node) to have `continue: true`. What seems to be the only validation enforcement on `continue` is that [it cannot be in the top route](https://github.com/prometheus/alertmanager/blob/master/config/config.go#L185)

**How to reproduce it (as minimally and precisely as possible)**:

Create a base config, with a default route
```YAML
route:
  group_by:
  - alertname
  - cluster
  group_wait: 2m
  group_interval: 10m
  repeat_interval: 3h
  receiver: some-pager
  routes:
...
```

Create an AlertmanagerConfig with one route
```YAML
route:
  continue: false
  groupBy:
  - alertname
  matchers:
  - name: alertname
    value: "mean_alert"
  receiver: slack
```

Deploy both, when the file is finally generated in alertmanager, the end results will be:
```
route:
  group_by:
  - alertname
  - cluster
  receiver: some-pager
  routes:
  - receiver: slack
    continue: true
    group_by:
    - alertname
    match:
    - alertname: "mean_alert"
      namespace: "deployed-namespace"
...
```
**Environment**

* Prometheus Operator version: v0.44.0
* Kubernetes version information: v1.17.17
* Kubernetes cluster kind: kops

* Prometheus Operator Logs: No relevant logs

**Anything else we need to know?**:

[...]

PR #3909, opened 2021-03-12
Do not default to continue being true within second level routes of generated AlertmanagerConfig

jandersen-plaid:
This removes the default of `continue: true` within the `routes` for generated alertmanager configurations.

Fixes: https://github.com/prometheus-operator/prometheus-operator/issues/3908

```release-note:
[BUGFIX] Generated Alertmanager config nodes do not default to continue: true if they are at the top level of an AlertmanagerConfig 

```


lilic:
Hey, thank you for the PR. But it seems like the commits do not match your user, any reason for this? :)

jandersen-plaid:
Ah! Sorry, looks like my local git configuration is a bit split-brained! I am closing this out, but the difference in user is due to two different users being mixed over :(

PR #3913, opened 2021-03-13
Ensure that objects have unique prefixes to prevent name collision

havulv:
Adds `serviceMonitor`, `probe`, and `podMonitor` prefixes to the `job_name` of relevant configurations so that the generated configurations do not suffer from `job_name` collision.

Fixes #3853

**Release Note Template (will be copied)**

```release-note
Prefix job_name with object type to prevent name collision in generated config
```

PR #3914, opened 2021-03-14
Validate the generated prometheus config before returning it for secret creation

### LINKED ISSUE #3853: Scrape-config jobs with same name occur crashloop in prometheus
**What happened?**
Prometheus created scrape config jobs with same name after creation of a Servicemonitor and Podmonitor with the same name in the same namespace. After trying to reload /etc/prometheus/config_out/prometheus.env.yaml Prometheus crashed and keeps in status CrashLoopBackOff. 

Prometheus should stop reloading in case the /etc/prometheus/config_out/prometheus.env.yaml file is not valid and should not crash. Additionally Prometheus should prevent updating the configuration file that it becomes invalid.

- create Prometheus with kube-prometheus-stack helm chart 
- create a ServiceMonitor and a PodMonitor with the same name in the same namespace 

**Environment**

0.45.0 (deployed by kube-prometheus-stack helm chart)

kubernetes 1.17.* 

* Kubernetes cluster kind:
Amazon EKS service

* Manifests:

* Prometheus Operator Logs:

* Prometheus Logs:
level=error ts=2021-02-18T14:20:44.401Z caller=main.go:293 msg="Error loading config (--config.file=/etc/prometheus/config_out/prometheus.env.yaml)" err="parsing YAML file /etc/prometheus/config_out/prometheus.env.yaml: found multiple scrape configs with job name \"default/testmonitor/0\""

Additional scrape configs set:
```
- job_name: 'kubernetes-service-endpoints'
  kubernetes_sd_configs:
    - role: endpoints
  relabel_configs:
    - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape]
      action: keep
      regex: true
    - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scheme]
      action: replace
      target_label: __scheme__
      regex: (https?)
    - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_path]
      action: replace
      target_label: __metrics_path__
      regex: (.+)
    - source_labels: [__address__, __meta_kubernetes_service_annotation_prometheus_io_port]
      action: replace
      target_label: __address__
      regex: ([^:]+)(?::\d+)?;(\d+)
      replacement: $1:$2
    - action: labelmap
      regex: __meta_kubernetes_service_label_(.+)
    - source_labels: [__meta_kubernetes_namespace]
      action: replace
      target_label: kubernetes_namespace
    - source_labels: [__meta_kubernetes_service_name]
      action: replace
      target_label: kubernetes_name
    - source_labels: [__meta_kubernetes_service_name]
      action: drop
      regex: 'node-exporter'
- job_name: 'kubernetes-pods'
  kubernetes_sd_configs:
    - role: pod
  relabel_configs:
    - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
      action: keep
      regex: true
    - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
      action: replace
      target_label: __metrics_path__
      regex: (.+)
    - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
      action: replace
      regex: ([^:]+)(?::\d+)?;(\d+)
      replacement: $1:$2
      target_label: __address__
    - action: labelmap
      regex: __meta_kubernetes_pod_label_(.+)
    - source_labels: [__meta_kubernetes_namespace]
      action: replace
      target_label: kubernetes_namespace
    - source_labels: [__meta_kubernetes_pod_name]
      action: replace
      target_label: kubernetes_pod_name
```

havulv:
This is an addon to #3913 and also fixes #3853, but in a different way. This PR validates all generated configurations, so that we never attempt to create a secret that is an invalid prometheus config file.

If this is expensive, then we can move the check closer to creating/updating the secret and only validate on creation or when there is a difference between the secret and the generated config.

Some of the tests were generating invalid configs, so I fixed those as well.

```release-note
Validate all generated configs before attempting to create or update the secret.
```


lilic:
Thanks for your PR! I understand where you are coming from but to me this seems like a big change to add validation, Prometheus should be the one validating the configuration and we should just load everything into the Secret. 

You could have validation on your end for CustomResources you create before they get created in CI/CD, but this essentially does it twice once in prom-operator and once again in Prometheus. 

It might work if we add the error in the Status subresource of each CustomResource but right now users would need to look at logs in any case.

With that being said, these are just my thoughts letting @prometheus-operator/prometheus-operator-reviewers chime in as well.

brancz:
I would agree with @lilic, this essentially moves the error from the prometheus log to the prometheus-operator. However, I think if we did this validation as sort of a dry-run at create/update time, I think that would be a very valuable improvement for user experience as it would fail when people do kubectl apply already.

lilic:
> However, I think if we did this validation as sort of a dry-run at create/update time, I think that would be a very valuable improvement for user experience as it would fail when people do kubectl apply already.

Yes, nice suggestion! The only thing with this there might be a problem with versions compatibility, we would import `"github.com/prometheus/prometheus/config"` to do this which is versioned but we tend to support more versions backwards and forwards. Not opposed to that, but just bringing up a possible edge case here. I guess you can always force with no validations to avoid this?

havulv:
I totally agree that this PR is not necessarily the best solution -- it was mostly put up for a more-or-less proof of concept that it could be done with minimal changes.

I do think there are two problems that it solves, though:
1. An invalid configuration will be saved into the configuration secret and prometheus will crash loop on it. To explain the scenario more thoroughly (and please correct me if my understanding of what happens is wrong here):
    1. A bad change is made to a prometheus or alertmanger configuration (non-unique name, non-existent receiver, etc.)
    2. `prometheus-operator` sees the change, generates the configuration, updates the secret.
    3. `prometheus` reloads the change and fails to get the reload
    4. There is no check for prometheus failing to reload, and `prometheus` goes into permanent `CrashLoopBackoff` until an end user intervenes.
2. An invalid configuration has the possiblity to be created. In general, it would be nice to follow the rule of `invalid configurations are impossible to create` and refuse to allow users to create invalid prometheus configurations. This has the advantage of making the above state less likely, and everything a bit more friendly for the user.

This PR solves both of those problems but, as pointed out, there are issues when it comes to validating twice and moving the error from prometheus to prometheus-operator. 

Separating the two problems above, I think a validating webhook or something that tells the user that a configuration is invalid at apply time would be very helpful (with a `--force` escape hatch as well) and solve the second problem. I am not totally sure how versioning can be solved here (calling a separate binary compiled for each config version depending on the version of prometheus that is detected?), but some gatekeeping would be helpful for small mistakes.

For the first problem, prometheus-operator should save the old configuration and, after X period of time, check that the reloaded prometheus stateful set has not gone into `CrashLoopBackoff` (with the default assumption being, when not  able to determine pod state, that everything is okay) and only rolling back once at maximum. Validating webhooks can't catch everything, but this would ensure that prometheus doesn't remain in a permanently crashed state from invalid configuration.

Though, this does move the error from prometheus to prometheus-operator, so I am unsure as to whether the prevailing opinion is:
1. Leave the error to prometheus and the user to figure out why prometheus is failing, or 
2. Protect prometheus from failing via prometheus-operator

havulv:
Looks like there is no movement on this, so I am going to close this out. If the dry run functionality mentioned is needed, then it should probably be considered  as an independent issue rather than through this PR.

Thank you @simonpasquier, @brancz, and @lilic for reviewing and reading through the PR!

PR #3933, opened 2021-03-25
Optimize asset secret update logic.

mikechengwei:
Optimize asset secret update logic , when secret data equal to exist data , the asset can avoid updates and reduce the pressure on k8s apiserver.

```release-note:REPLACEME
Optimize asset secret update logic.
```


pkg/prometheus/operator.go
```diff
@@ -1591,8 +1591,12 @@ func (c *Operator) createOrUpdateTLSAssetSecret(ctx context.Context, p *monitori
...
 
 	} else {
-		err = k8sutil.UpdateSecret(ctx, sClient, tlsAssetsSecret)
-		level.Debug(c.logger).Log("msg", "updated tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
+		eq := reflect.DeepEqual(secret.Data, tlsAssetsSecret.Data)
+		if !eq {
+			err = k8sutil.UpdateSecret(ctx, sClient, tlsAssetsSecret)
+			level.Debug(c.logger).Log("msg", "updated tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
+		}
+
```
simonpasquier, line 1599:
```suggestion
		if reflect.DeepEqual(secret.Data, tlsAssetsSecret.Data) {
			return nil
		}
		err = k8sutil.UpdateSecret(ctx, sClient, tlsAssetsSecret)
		level.Debug(c.logger).Log("msg", "updated tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
```

simonpasquier (commented):
the same change should happen in pkg/alertmanager/operator.go (and maybe a good time to factorize some code).

mikechengwei:
> the same change should happen in pkg/alertmanager/operator.go (and maybe a good time to factorize some code).
I have change code in `pkg/alertmanager/operator.go` again. Do you mean to factorize some code, for example, to extract the common code and put it in the util class?

For example , extract below code to a common method .
```
secret, err := sClient.Get(ctx, tlsAssetsSecret.Name, metav1.GetOptions{})
	if err != nil {
		if !apierrors.IsNotFound(err) {
			return errors.Wrapf(
				err,
				"failed to check whether tls assets secret already exists for Prometheus %v in namespace %v",
				p.Name,
				p.Namespace,
			)
		}
		_, err = sClient.Create(ctx, tlsAssetsSecret, metav1.CreateOptions{})
		level.Debug(c.logger).Log("msg", "created tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)

	} else {
		if reflect.DeepEqual(secret.Data, tlsAssetsSecret.Data) {
			return nil
		}
		err = k8sutil.UpdateSecret(ctx, sClient, tlsAssetsSecret)
		level.Debug(c.logger).Log("msg", "updated tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
	}

	if err != nil {
		return errors.Wrapf(err, "failed to create TLS assets secret for Prometheus %v in namespace %v", p.Name, p.Namespace)
	}
```

simonpasquier:
> Do you mean to factorize some code, for example, to extract the common code and put it in the util class?

yes it would be neat :) I think that you can repurpose `k8sutil.UpdateSecret()` to be `k8sutil.CreateOrUpdateSecret()`: the latter can skip the update if existing and updated secrets are identical in terms of data and metadata (labels/annotations).

pkg/k8sutil/k8sutil.go
```diff
@@ -175,20 +176,33 @@ func UpdateStatefulSet(ctx context.Context, sstClient clientappsv1.StatefulSetIn
...
+				desired.Name,
+				desired.Namespace,
+			)
+		}
+		_, err = secretClient.Create(ctx, desired, metav1.CreateOptions{})
 		return err
 	}
-
+	// If secret exist,
+	mutated := existingSecret.DeepCopyObject().(*v1.Secret)
```
simonpasquier, line 195: I don't think we need to deep-copy the object

pkg/k8sutil/k8sutil.go
```diff
@@ -175,20 +176,33 @@ func UpdateStatefulSet(ctx context.Context, sstClient clientappsv1.StatefulSetIn
...
+
+	mergeMetadata(&desired.ObjectMeta, mutated.ObjectMeta)
+	mutated.Data = desired.Data
+	// diff and whether update
+	if !apiequality.Semantic.DeepEqual(existingSecret, mutated) {
```
simonpasquier, line 200: why not reflect.DeepEqual() on the Data fields?

pkg/k8sutil/k8sutil_test.go
```diff
@@ -76,32 +76,6 @@ func TestMergeMetadata(t *testing.T) {
 		modifiedLabels      map[string]string
 		modifiedAnnotations map[string]string
 	}{
-		{
```
simonpasquier, line 79: why did you remove these test cases?

pkg/prometheus/operator.go
```diff
@@ -1529,29 +1529,7 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *mon
...
-			return nil
-		}
-		level.Debug(c.logger).Log("msg", "current Prometheus configuration has changed")
-	} else {
-		level.Debug(c.logger).Log("msg", "no current Prometheus configuration secret found", "currentConfigFound", curConfigFound)
-	}
-
-	level.Debug(c.logger).Log("msg", "updating Prometheus configuration secret")
-	return k8sutil.UpdateSecret(ctx, sClient, s)
+	return k8sutil.CreateOrUpdateSecret(ctx, sClient, s)
```
simonpasquier, line 1532:
```suggestion
        level.Debug(c.logger).Log("msg", "updating Prometheus configuration secret")
	return k8sutil.CreateOrUpdateSecret(ctx, sClient, s)
```

pkg/k8sutil/k8sutil.go
mikechengwei: We also need to compare labels and annoations. `apiequality.Semantic.DeepEqual` method is like reflect.DeepEqual, but focused on semantic equality instead of memory equality.

pkg/k8sutil/k8sutil_test.go
mikechengwei: Sorry , my test result.

pkg/k8sutil/k8sutil.go
mikechengwei: We must use deepcopy a new secret from old secret , because `mergeMetadata` will change existingSecret value .

mikechengwei: We need to merge a new secret to compare with old secret.

pkg/k8sutil/k8sutil.go
simonpasquier: :+1: looking quickly at the definition of `apiequality.Semantic.DeepEqual()`, I thought that it was only checking a few fields but now I see that it uses reflect.DeepEqual() when needed underneath :)

pkg/k8sutil/k8sutil.go
```diff
@@ -175,20 +176,30 @@ func UpdateStatefulSet(ctx context.Context, sstClient clientappsv1.StatefulSetIn
...
-
+	mutated := existingSecret.DeepCopyObject().(*v1.Secret)
+	mergeMetadata(&desired.ObjectMeta, mutated.ObjectMeta)
+	// diff and whether update
+	if !apiequality.Semantic.DeepEqual(existingSecret, desired) {
+		_, err = secretClient.Update(ctx, desired, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+	}
```
simonpasquier, line 202: Lets add secret name/namespace. Also a bit nicer to read IMHO (less nested levels).

```suggestion
	if apiequality.Semantic.DeepEqual(existingSecret, desired) {
                return nil
        }
	
	if _, err = secretClient.Update(ctx, desired, metav1.UpdateOptions{}); err != nil {
		return errors.Wrapf(
			err,
			"failed to update secret %q in namespace %q",
			desired.Name,
			desired.Namespace,
		)
	}
	
	return nil
```

pkg/k8sutil/k8sutil.go
```diff
@@ -175,20 +176,30 @@ func UpdateStatefulSet(ctx context.Context, sstClient clientappsv1.StatefulSetIn
...
+		if !apierrors.IsNotFound(err) {
+			return errors.Wrapf(
+				err,
+				"failed to check whether secret %q in namespace %q already exists",
+				desired.Name,
+				desired.Namespace,
+			)
+		}
+		_, err = secretClient.Create(ctx, desired, metav1.CreateOptions{})
 		return err
```
simonpasquier, line 192: lets wrap here too

```suggestion
		return errors.Wrapf(err, "failed to create secret %q in namespace %q", desired.Name, desired.Namespace)
```

pkg/alertmanager/operator.go
```diff
@@ -830,23 +830,7 @@ func (c *Operator) createOrUpdateGeneratedConfigSecret(ctx context.Context, am *
...
-		_, err = sClient.Create(ctx, generatedConfigSecret, metav1.CreateOptions{})
-		level.Debug(c.logger).Log("msg", "created generated config secret", "secretname", generatedConfigSecret.Name)
-	} else {
-		err = k8sutil.UpdateSecret(ctx, sClient, generatedConfigSecret)
-		level.Debug(c.logger).Log("msg", "updated generated config secret", "secretname", generatedConfigSecret.Name)
-	}
-
+	err := k8sutil.CreateOrUpdateSecret(ctx, sClient, generatedConfigSecret)
 	if err != nil {
 		return errors.Wrapf(err, "failed to update generated config secret for Alertmanager %v in namespace %v", am.Name, am.Namespace)
```
simonpasquier, line 835: after looking at the code I think we can remove the name and namespace of the Alertmanager resource because it s already added by the caller (e.g. `sync()`):

```suggestion
		return errors.Wrap(err, "failed to update generated config secret")
```

pkg/alertmanager/operator.go
```diff
@@ -1313,24 +1297,7 @@ func (c *Operator) createOrUpdateTLSAssetSecret(ctx context.Context, am *monitor
...
-		level.Debug(c.logger).Log("msg", "created tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
-
-	} else {
-		err = k8sutil.UpdateSecret(ctx, sClient, tlsAssetsSecret)
-		level.Debug(c.logger).Log("msg", "updated tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
-	}
-
+	err := k8sutil.CreateOrUpdateSecret(ctx, sClient, tlsAssetsSecret)
 	if err != nil {
 		return errors.Wrapf(err, "failed to create TLS assets secret for Alertmanager %v in namespace %v", am.Name, am.Namespace)
```
simonpasquier, line 1302: ditto

pkg/prometheus/operator.go
```diff
@@ -1580,24 +1559,7 @@ func (c *Operator) createOrUpdateTLSAssetSecret(ctx context.Context, p *monitori
...
-		level.Debug(c.logger).Log("msg", "created tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
-
-	} else {
-		err = k8sutil.UpdateSecret(ctx, sClient, tlsAssetsSecret)
-		level.Debug(c.logger).Log("msg", "updated tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
-	}
-
+	err := k8sutil.CreateOrUpdateSecret(ctx, sClient, tlsAssetsSecret)
 	if err != nil {
 		return errors.Wrapf(err, "failed to create TLS assets secret for Prometheus %v in namespace %v", p.Name, p.Namespace)
```
simonpasquier, line 1564: ditto about removing name and namespace

pkg/prometheus/operator.go
```diff
@@ -1580,24 +1559,7 @@ func (c *Operator) createOrUpdateTLSAssetSecret(ctx context.Context, p *monitori
...
-		}
-		_, err = sClient.Create(ctx, tlsAssetsSecret, metav1.CreateOptions{})
-		level.Debug(c.logger).Log("msg", "created tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
-
-	} else {
-		err = k8sutil.UpdateSecret(ctx, sClient, tlsAssetsSecret)
-		level.Debug(c.logger).Log("msg", "updated tlsAssetsSecret", "secretname", tlsAssetsSecret.Name)
-	}
-
+	err := k8sutil.CreateOrUpdateSecret(ctx, sClient, tlsAssetsSecret)
```
simonpasquier, line 1562: we can merge with the next block

```suggestion
	if err := k8sutil.CreateOrUpdateSecret(ctx, sClient, tlsAssetsSecret); err != nil {
```

simonpasquier (commented):
Thanks for your patience :) just a few minor comments to make the code more consistent but overall :+1: for me.

mikechengwei:
@simonpasquier Thank you very much for your guidance and learn together. Please help me  review again.

mikechengwei:
@simonpasquier

simonpasquier (approved):

simonpasquier:
Looks good to me, thanks a lot!

PR #3934, opened 2021-03-25
*: fix gosimple errors

simonpasquier:

```release-note:none

```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,6 +5,7 @@ linters:
   disable-all: true
   enable:
   - golint
+  - gosimple
   - govet
   - staticcheck
   - unused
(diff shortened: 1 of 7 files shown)


.golangci.yml
```diff
@@ -5,6 +5,7 @@ linters:
   disable-all: true
   enable:
   - golint
+  - gosimple
```
lilic, line 8: Noticed you opened a few of these lately. I think its amazing! 👏  But my suggestion is it might be amazing to also open issues for other external folks to do, as its really great first issue. What do you think?

lilic (approved):
lgtm
Thanks!

.golangci.yml
simonpasquier: It happens that I had all the changes already prepared on my machine and the last one would be for the errcheck linter.

PR #3943, opened 2021-03-29
alertmanager: Fix handleAlertmanagerUpdate version check

Jbaudon:
handleAlertmanagerUpdate has to check if the resource has actually been updated

```release-note:REPLACEME

```


Jbaudon:
That patch effectively fixes a big Kubernetes API overuse issue. In my use-case (EKS) that issue led to a increase in the billing.

simonpasquier (commented):
lgtm, nice catch!

PR #3944, opened 2021-03-30
pkg/alertmanager: added hostAliases to pod spec (fixes #3488)

### LINKED ISSUE #3488: Add an option to add hostAlias for alert manager
Hi Team,

For us to talk to a webhook , we have a init container which loads secrets from vault. The issue we are facing is that the alert manager pod is not able to resolve the dns for the same.

Also we run alert manager as non toot user so we are not able to change the host entries during the pod start lifecycle.

If we have any alternatives for this please let me know

suvl:
Fixes #3488.

- release-note:feature

```release-note
Adds the ability to configure the `hostAlias` array in the `alertManagerSpec` object that 
in turn will set the same field in the `PodSpec` of the running AlertManager pods.
```


coderanger:
I think a more sustainable solution to this is some kind of deep-merge approach using PodSpec like the `container` field is now. Otherwise this will just result in duplicating the majority of PodSpec eventually.

Grounz:
Hello, we need this feature so what's the status ?

suvl:
#4787 beated me to it. closing.

simonpasquier:
@suvl thanks for the effort!

Issue #3061, opened 2020-03-04
Prometheus pod cannot find `prometheus.env.yaml` on start


Yangstatus:
Please help me to check this error, thank you! 

Below is the description of my pod

[root@k8s-master-1 alertmanager]# kubectl describe  pod prometheus-prometheus-operator-158278-prometheus-0 
Name:         prometheus-prometheus-operator-158278-prometheus-0
Namespace:    default
Priority:     0
Node:         k8s-node-2/172.29.1.102
Start Time:   Mon, 02 Mar 2020 16:30:42 +0800
Labels:       app=prometheus
              controller-revision-hash=prometheus-prometheus-operator-158278-prometheus-674c57cc8c
              prometheus=prometheus-operator-158278-prometheus
              statefulset.kubernetes.io/pod-name=prometheus-prometheus-operator-158278-prometheus-0
Annotations:  cni.projectcalico.org/podIP: 10.100.140.127/32
Status:       Running
IP:           10.100.140.127
IPs:
  IP:           10.100.140.127
Controlled By:  StatefulSet/prometheus-prometheus-operator-158278-prometheus
Containers:
  prometheus:
    Container ID:  docker://bbec5226b9c3610a265a5cee046f76044e5ff26149ad5079f774f2a49f46f1e1
    Image:         quay.io/prometheus/prometheus:v2.15.2
    Image ID:      docker-pullable://quay.io/prometheus/prometheus@sha256:914525123cf76a15a6aaeac069fcb445ce8fb125113d1bc5b15854bc1e8b6353
    Port:          9090/TCP
    Host Port:     0/TCP
    Args:
      --web.console.templates=/etc/prometheus/consoles
      --web.console.libraries=/etc/prometheus/console_libraries
      --config.file=/etc/prometheus/config_out/prometheus.env.yaml
      --storage.tsdb.path=/prometheus
      --storage.tsdb.retention.time=10d
      --web.enable-lifecycle
      --storage.tsdb.no-lockfile
      --web.external-url=http://prometheus-operator-158278-prometheus.default:9090
      --web.route-prefix=/
    State:       Running
      Started:   Mon, 02 Mar 2020 16:30:44 +0800
    Last State:  Terminated
      Reason:    Error
      Message:    caller=main.go:648 msg="Starting TSDB ..."
level=info ts=2020-03-02T08:30:43.925Z caller=web.go:506 component=web msg="Start listening for connections" address=0.0.0.0:9090
[...]

PR #3945, opened 2021-03-30
Added flag: one-off

zsuzhengdu:
Fix https://github.com/prometheus-operator/prometheus-operator/issues/3061#issuecomment-619812179

<!-- AUTOCHANGELOG for Downstream PRs.

Please select one of the following "release-note:" headings:
    - release-note:enhancement
    - release-note:bug

```release-note:REPLACEME

```


coderanger:
I don't think this would work as you expect.

zsuzhengdu:
Cancel apply upon watchInterval == 0 without the need of one-off option in bool.

PR #3946, opened 2021-03-31
Update prometheus and thanos versions

### LINKED ISSUE #3950: Remote write configuration not propagated unless prometheus version is explicitly specified
**What happened?**

Remote write headers did not get passed to prometheus. I put them in prometheus spec manifest but they did not get propagated into prometheus configuration. After looking at the code related to this I realized that this is caused by logic which checks if I have prometheus version which supports this feature.

```
		if len(spec.Headers) > 0 && version.GTE(semver.MustParse("2.25.0")) {
			cfg = append(cfg, yaml.MapItem{Key: "headers", Value: stringMapToMapSlice(spec.Headers)})
		}
```
https://github.com/prometheus-operator/prometheus-operator/blob/7f94a06b86d41c20176f0d5b53aa0100fdc361e6/pkg/prometheus/promcfg.go#L1404

We don't specify `version` in prometheus spec since it not required but after bit more digging I realized that if I don't explicitly specify it falls back on `PrometheusCompatibilityMatrix` which doesn't contain the `2.25.0` version.
https://github.com/prometheus-operator/prometheus-operator/blob/v0.46.0/pkg/operator/defaults.go
  
This essentially disable feature because field which is not required is not filled in. I think this is quite bad since afaik it's not documented, it also makes wonder where else is this logic reused. I would suggest updating at least the description for remoteWrite headers. 

Since I was running prometheus version which supports this feature I was expecting them to get propagated to the prometheus configuration.

**Environment**

    `v0.46.0`
    

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}

Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.6-eks-49a6c0", GitCommit:"49a6c0bf091506e7bafcdb1b142351b69363355a", GitTreeState:"clean", BuildDate:"2020-12-23T22:10:21Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```
    

* Kubernetes cluster kind:

    GKE

* Manifests:

```
  remoteWrite:
  - headers:
      THANOS-TENANT: dev1
    name: thanos-receive
    tlsConfig:
      caFile: /tls/ca.pem
      certFile: /tls/prometheus.crt
      keyFile: /tls/prometheus.key
    url: https://ednpoint/api/v1/receive
```

* Prometheus Operator Logs:

```
I don't have logs for this unfortunately 
```

QuentinBisson:
This PR updates the prometheus and thanos versions to the latest ones
closes #3950 

```release-note:enhancement
Update Prometheus and Thanos versions
```

```release-note:bug
Fix configuration of remote write headers
```

### DIFF
--- a/pkg/operator/defaults.go
+++ b/pkg/operator/defaults.go
@@ -20,7 +20,7 @@ const (
 	DefaultAlertmanagerVersion   = "v0.21.0"
 	DefaultAlertmanagerBaseImage = "quay.io/prometheus/alertmanager"
 	DefaultAlertmanagerImage     = DefaultAlertmanagerBaseImage + ":" + DefaultAlertmanagerVersion
-	DefaultThanosVersion         = "v0.17.2"
+	DefaultThanosVersion         = "v0.18.0"
 	DefaultThanosBaseImage       = "quay.io/thanos/thanos"
 	DefaultThanosImage           = DefaultThanosBaseImage + ":" + DefaultThanosVersion
 )
@@ -67,6 +67,10 @@ var (
 		"v2.23.0",
 		"v2.24.0",
 		"v2.24.1",
+		"v2.25.0",
+		"v2.25.1",
+		"v2.25.2",
+		"v2.26.0",
 	}
 	DefaultPrometheusVersion   = PrometheusCompatibilityMatrix[len(PrometheusCompatibilityMatrix)-1]
 	DefaultPrometheusBaseImage = "quay.io/prometheus/prometheus"
--- a/Documentation/compatibility.md
+++ b/Documentation/compatibility.md
@@ -57,6 +57,10 @@ The versions of Prometheus compatible to be run with the Prometheus Operator are
 * v2.23.0
 * v2.24.0
 * v2.24.1
+* v2.25.0
+* v2.25.1
+* v2.25.2
+* v2.26.0
 
 ## Alertmanager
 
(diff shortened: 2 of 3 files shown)


pkg/operator/defaults.go
```diff
@@ -67,6 +67,10 @@ var (
 		"v2.23.0",
 		"v2.24.0",
 		"v2.24.1",
+		"v2.25.0",
+		"v2.25.1",
+		"v2.25.2",
+		"v2.26.0",
```
simonpasquier, line 73: v2.26.0 isn't released yet :)

QuentinBisson: Well, the tag is already there so this should only be a matter of hours I hope :)
https://github.com/prometheus/prometheus/releases/tag/v2.26.0

QuentinBisson: Now it is released :)

simonpasquier: :smile: relaunched the CI

simonpasquier:
Ah I thought that we've already got rid of the Alertmanager v1 API but apparently not :) 

This line needs an update from v1 to v2 since Prometheus 2.26 now defaults to v2:
https://github.com/prometheus-operator/prometheus-operator/blob/4826c904d86c93cffaed293c5ce85ad6b2051995/test/e2e/prometheus_test.go#L3643

QuentinBisson:
Thanks @simonpasquier this solved the issue :D

simonpasquier (approved):
Thanks!

PR #3955, opened 2021-04-07
Added init container which applies config file once by setting `--watch-interval` equals to `0`

zsuzhengdu:
Fix https://github.com/prometheus-operator/prometheus-operator/issues/3061

Depends on a new Thanos release that includes https://github.com/thanos-io/thanos/pull/3994


coderanger:
At the point that the init container runs, the primary containers have not started. So prometheus won't actually be up yet for this to trigger a reload on.

zsuzhengdu:
Good call. Updated to skip making reload request to :locahost:9090/-/reload when --watch-interval==0; please see https://github.com/thanos-io/thanos/pull/4032
Also skip starting metrics server in cmd/prometheus-config-reloader/main.go when --watch-interval==0.

zsuzhengdu:
@coderanger  @brancz 

Would you please help to review the changes?  Thanks!

simonpasquier (commented):
can you update the description of the `--watch-interval` flag to explain that when 0, the program runs only once and exits.

pkg/operator/statefulset.go
```diff
@@ -28,6 +28,7 @@ const configReloaderPort = 8080
 // container. No shard environment variable will be passed to the reloader
 // container if `-1` is passed to the shards parameter.
 func CreateConfigReloader(
```
simonpasquier, line 30: It might be more readable to change the way we pass options to `CreateConfigReloader()` and use [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) instead:
* more explicit than passing a zero  watch interval to enable the "run once" mode 
* some options don't make sense in this case (e.g. reload URL and ports).

Instead we could write something like:

```
initReloaderContainer := operator.CreateConfigReloader(
    "init-config-reloader",
    operator.ReloaderResources(c.ReloaderConfig),
    operator.ReloaderRunOnce(),
)

...

reloaderContainer := operator.CreateConfigReloader(
    "config-reloader",
    operator.ReloaderResources(c.ReloaderConfig),
    operator.ReloaderURL(url.URL{...}),
    ...,
)


```

zsuzhengdu:
@simonpasquier  @coderanger  Can you please take another review on this PR?

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+	listenLocal    bool
+	localHost      string
+	logFormat      string
+	logLevel       string
+	additionalArgs []string
+	volumeMounts   []v1.VolumeMount
+	shard          int32
+}
+
+type option = func(*ConfigReloader)
```
simonpasquier, line 40:
```suggestion
type ReloaderOption = func(*ConfigReloader)
```

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+	return func(c *ConfigReloader) {
+		var listenAddress string
+		if c.listenLocal {
+			listenAddress = c.localHost
+		}
+		c.additionalArgs = append(c.additionalArgs, fmt.Sprintf("--listen-address=%s:%d", listenAddress, configReloaderPort))
+	}
+}
+
+func ReloadURL() option {
```
simonpasquier, line 58: I would remove this function and move the logic to `NewConfigReloader()` which could add the `--reload-url` argument if c.reloadURL.String() isn't empty.

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+
+func Shard(shard int32) option {
+	return func(c *ConfigReloader) {
+		c.shard = shard
+	}
+}
+
 // CreateConfigReloader returns the definition of the config-reloader
 // container. No shard environment variable will be passed to the reloader
 // container if `-1` is passed to the shards parameter.
```
simonpasquier, line 120:
```suggestion
// CreateConfigReloader returns the definition of the config-reloader
// container.
```

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+	name           string
+	config         ReloaderConfig
+	reloadURL      url.URL
+	listenLocal    bool
+	localHost      string
+	logFormat      string
+	logLevel       string
+	additionalArgs []string
+	volumeMounts   []v1.VolumeMount
+	shard          int32
```
simonpasquier, line 37:
```suggestion
	shard          *int32
```

so we know whether a shard value has been provided or not.

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+	}
+}
+
+func AdditionalArgs(args []string) option {
+	return func(c *ConfigReloader) {
+		c.additionalArgs = append(c.additionalArgs, args...)
+	}
+}
+
+func VolumeMount(mounts []v1.VolumeMount) option {
```
simonpasquier, line 106:
```suggestion
func VolumeMounts(mounts []v1.VolumeMount) option {
```

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+	return func(c *ConfigReloader) {
+		c.listenLocal = listenLocal
+	}
+}
+
+func LocalHost(localHost string) option {
+	return func(c *ConfigReloader) {
+		c.localHost = localHost
+	}
+}
```
simonpasquier, line 86: It might make sense to keep only one function and move the logic to the caller

```suggestion
func ListenLocal(localhost string) option {
	return func(c *ConfigReloader) {
		c.localHost = localHost
	}
}
```

Then in pkg/prometheus/statefulset.go

```go
if p.Spec.ListenLocal {
  reloaderOpts = append(reloaderOpts, operator.ListenLocal(c.Localhost))
}
```

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
-				Name:          "reloader-web",
-				ContainerPort: configReloaderPort,
-				Protocol:      v1.ProtocolTCP,
-			},
-		)
-	}
-	args = append(args, fmt.Sprintf("--listen-address=%s:%d", listenAddress, configReloaderPort))
-
-	args = append(args, fmt.Sprintf("--reload-url=%s", reloadURL.String()))
+	ports = append(
```
simonpasquier, line 133: we shouldn't expose a port if it listens on localhost.

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+
+type option = func(*ConfigReloader)
+
+func ReloaderRunOnce() option {
+	return func(c *ConfigReloader) {
+		c.additionalArgs = append(c.additionalArgs, fmt.Sprintf("--watch-interval=%d", 0))
+	}
+}
+
+func ListenAddress() option {
```
simonpasquier, line 48: I don't think we need this option for now.

pkg/operator/statefulset.go
```diff
@@ -24,74 +24,154 @@ import (
...
+	}
+}
+
+func LogLevel(logLevel string) option {
+	return func(c *ConfigReloader) {
+		c.logLevel = logLevel
+	}
+}
+
+func AdditionalArgs(args []string) option {
```
simonpasquier, line 100: could we get rid of `AdditionalArgs()` and add functions for things that aren't currently represented as option?

simonpasquier (commented):
A few suggestions but great work!

pkg/operator/statefulset.go
zsuzhengdu: Thanks for the review and great feedbacks, @simonpasquier!

`func additionalArgs` created to take of options of running config-reloader; like `configReloaderArgs` in `func makeStatefulSetSpec`.

If removed; to run config-reloader just once, a bool field named, e.g.,`runOnce` might be needed in struct `ConfigReloader` for `func ReloaderRunOnce` and used in `func CreateConfigReloader` as following
```
if c.runOnce {
    args  = append(args, fmt.Sprintf("--watch-interval=%d", 0)
}
```
Addition fields in string would be needed for  `--watched-dir`, `--config-file` and etc in struct ConfigReloader. 

Feel free to correct me if misunderstand the comment and point me the enlightening path to address the comment.

zsuzhengdu: @simonpasquier Would you please give another review on this PR?

zsuzhengdu:
Merge?

zsuzhengdu:
Can not reproduce e2e test failure locally using minikube, 

```
PASS
ok  	github.com/prometheus-operator/prometheus-operator/test/e2e	1795.114s
```

@simonpasquier  @paulfantom Would you please shed some light on https://github.com/prometheus-operator/prometheus-operator/runs/2516336987?check_suite_focus=true#step:6:305?

zsuzhengdu:
Ping @paulfantom  Can you help to trigger the rerun please?

Makefile
```diff
@@ -270,7 +270,7 @@ test/e2e/remote_write_certs/ca.key test/e2e/remote_write_certs/ca.crt test/e2e/r
 .PHONY: test-e2e
 test-e2e: KUBECONFIG?=$(HOME)/.kube/config
 test-e2e: test/instrumented-sample-app/certs/cert.pem test/instrumented-sample-app/certs/key.pem
-	go test -timeout 55m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG) --operator-image=$(REPO):$(TAG) -count=1
+	go test -timeout 100m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG) --operator-image=$(REPO):$(TAG) -count=1
```
fpetkovski, line 273: Is this change related to the issue, or is it something independent?

test/framework/prometheus.go
```diff
@@ -287,7 +287,7 @@ func (f *Framework) UpdatePrometheusAndWaitUntilReady(ns string, p *monitoringv1
 	if err != nil {
 		return nil, err
 	}
-	if err := f.WaitForPrometheusReady(result, 5*time.Minute); err != nil {
+	if err := f.WaitForPrometheusReady(result, 10*time.Minute); err != nil {
```
fpetkovski, line 290: Is this change related to the issue, or is it something independent?

fpetkovski (changes requested):
Thanks for the contribution, the PR looks quite good so far. 

As one last improvement, could you also add tests to the [prometheus](https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/prometheus/statefulset_test.go) and [alertmanager](https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/alertmanager/statefulset_test.go) test suites to make sure the init container is added to the respective statefulsets?

pkg/operator/statefulset.go
```diff
@@ -17,98 +17,194 @@ package operator
...
 	"net/url"
+	"strconv"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 const configReloaderPort = 8080
 
+type ConfigReloader struct {
```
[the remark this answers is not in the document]
zsuzhengdu: Shall we just rename the file from `statefulset.go` to `config-reloader.go`,  but kept under the same `pkg/operator` as other files that are used for composing operator instead of creating another `pkg`?

Will add godocs as per suggested :-)

test/framework/prometheus.go
zsuzhengdu: The wait timeout was increated to pass e2e tests; which might be related to under powered nodes. 

Noticed intermittent failures, which I could not reproduce locally; thus increased the timeout for status polling.  With the polling timeout doubled, the `go test` timeout increased as well as commented above.

pkg/operator/statefulset.go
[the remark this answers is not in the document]
fpetkovski: > Shall we just rename the file from statefulset.go to config-reloader.go, but kept under the same pkg/operator as other files that are used for composing operator instead of creating another pkg?

My understanding is that config-reloader is not used for composing the operator, but rather as a sidecar for prometheus and alertmanager. You can see that the `CreateConfigReloader` function is referenced from the alertmanager and prometheus packages which creates a bit of a circular dependency. 

In any case, I would not mind just renaming the file to `config-reloader.go` and leaving it in the same package since that would already be an improvement. I leave the decision up to you depending on how much effort you think each approach will require :)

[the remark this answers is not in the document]
zsuzhengdu: You are right; I should not have used the wording `composing the operator` and apologize for that!

Yah, `CreateCofnigReloader` is called in funcs named `makeStatefulSetSpec` in files named `statefulset.go` under `pkg/prometheus` , `pkg/alertmanager` and `pkg/thanos`...which **might** be  the reason why the file name is ALSO called `statefulset.go` under `pkg/operator`...LoL

As funcs in other files, such as `image.go` under `pkg/operator` callled by `makeStatefulSetSpec`; renaming seams the minial effort to take for improvment.

Makefile
zsuzhengdu: Rolled the change back and tests still passing....

zsuzhengdu:
@fpetkovski  Are these changes good enough and would you please give another review?

fpetkovski:
Hey @zsuzhengdu, this looks good to me now. I also tested the changes locally and they look solid. Thanks for the contribution!

fpetkovski (approved):

zsuzhengdu:
Thank you for the review and local try out, @fpetkovski!

paulfantom (approved):
lgtm from my side, leaving final say to @simonpasquier

simonpasquier:
Thanks a lot for your perseverance and sorry about the delayed review from my side! Also thanks to @fpetkovski for taking over :)

I have 2 last comments:
* I don't think we need to add the init container to the Alertmanager statefulset because the config reloader doesn't substitute environment variables in the Alertmanager config file.
* It would be good to mention the new  `init-config-reloader` container in the API documentation. In practice, this [comment](https://github.com/prometheus-operator/prometheus-operator/blob/43f5825becbf7f345f595cc350ca6600491ec5f4/pkg/apis/monitoring/v1/types.go#L269-L275) should be expanded to match what's already documented for [regular containers](https://github.com/prometheus-operator/prometheus-operator/blob/43f5825becbf7f345f595cc350ca6600491ec5f4/pkg/apis/monitoring/v1/types.go#L261-L267) generated by the operator.

simonpasquier (approved):
:+1: thanks a lot! :tada:

simonpasquier:
@zsuzhengdu it took longer than expected but finally it's merged! Thanks a lot for your patience :bow:

PR #3956, opened 2021-04-07
*: bump Go dependencies

simonpasquier:
In preparation of the v0.47 release...

```release-note:none

```


paulfantom (approved):
lgtm

chadlwilson:
Hi folks - this is probably the wrong place to comment on this, and I am second-guessing myself because I can't be the first person to come across this, however this seems to have bumped the `prometheus/common` dependencies from `0.15` to 0.20`. Within this is addition of  `follow_redirects: true` to the default http config if I am inferring from https://github.com/prometheus/common/commit/47ee79a16821e03c31ca741659d4f7919aed91c1 correctly.

Unfortunately AlertManager `0.21.0` doesn't understand this setting (seems to be a `0.22.0` thing?) and seems to barf when trying to load config off the config reloader when the user doesn't specify anything for `follow_redirects`, because the default has it specified.

Am I barking up the wrong tree? Is the issue to be raised there or here (indirectly referred to at https://github.com/prometheus-operator/prometheus-operator/issues/4009) or .... other option?

simonpasquier:
@chadlwilson could you please open an issue in this repository? It seems like an annoying regression :thinking:

simonpasquier:
FWIW I've reopened #4009

Issue #3801, opened 2021-01-19
Shard count won't get translated properly


sjentzsch:
If I set `shards: 3` in my Prometheus CR, it will only spin up one additional shard:

```
monitoring       prometheus-kube-prometheus-stack-prometheus-0               3/3     Running   1          2m34s
monitoring       prometheus-kube-prometheus-stack-prometheus-shard-1-0       3/3     Running   1          2m34s
```
If I change shards back to 2, then back to 3 again, it will spin up the third one:

```
monitoring       prometheus-kube-prometheus-stack-prometheus-0               3/3     Running   1          7m46s
monitoring       prometheus-kube-prometheus-stack-prometheus-shard-1-0       3/3     Running   1          7m44s
monitoring       prometheus-kube-prometheus-stack-prometheus-shard-2-0       3/3     Running   1          4m7s
```
Looks to me like a controller issue?

**Environment**

* Prometheus Operator version: v0.44.1

* Kubernetes version information: 1.19.7

lilic:
Thanks for the issue! How long did you wait for, as we do have a reconcile period, did it eventually (after 5-10 minutes) reconcile or did it never?

sjentzsch:
I am afraid it does never, I waited 10min+
Only if I actively change the shards (3 -> 2 -> 3) it will add the third one.

lilic:
Interesting this does indeed sound like a bug! Can you share some logs of prom-operator and the Prometheus instances that are created, e.g. the CRs? Hmm is this the helm chart and can you try to reproduce without the helm chart?

Gokul-Radhakrishnan:
Even I am facing this issue (I did not use helm chart to install prometheus-operator). 

The following is the CR I used 

```
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
metadata:
  name: prometheus
  labels:
    prometheus: `prometheus`
spec:
  serviceAccountName: prometheus
  serviceMonitorSelector:
    matchLabels:
      team: frontend
  shards: 5
  replicas: 2
```
The following are the logs from prometheus operator pod
```
level=info ts=2021-01-25T11:27:46.2592206Z caller=main.go:233 msg="Starting Prometheus Operator" version="(version=0.45.0, branch=refs/tags/v0.45.0, revision=5555f492df250168657b72bb8cb60bec071de71f)"
[...]

PR #3974, opened 2021-04-11
Fix inconsistency question between the number of pods and shard configuration

mikechengwei:
#3801 
I found some log by enable debug log level.

```
level=debug ts=2021-04-11T16:10:13.24178926Z caller=operator.go:1190 component=prometheusoperator msg="reconciling statefulset" statefulset=prometheus-main shard=0
level=debug ts=2021-04-11T16:10:13.242372418Z caller=operator.go:1225 component=prometheusoperator msg="new statefulset generation inputs match current, skipping any actions"
``` 

After my analysis, I found a bug. 

When the first shard is created, it exits the loop.

https://github.com/prometheus-operator/prometheus-operator/blob/fb16b71909dfe3feff2b5e77214002e10bad301d/pkg/prometheus/operator.go#L1214-L1219

If the hash value of the first shard in the next loop has not changed, it will exit the loop directly and it will not create the next shard.
https://github.com/prometheus-operator/prometheus-operator/blob/fb16b71909dfe3feff2b5e77214002e10bad301d/pkg/prometheus/operator.go#L1224-L1227

```release-note:REPLACEME
Fix inconsistency question between the number of pods and shard configuration.
```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1223,7 +1223,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		oldSSetInputHash := obj.(*appsv1.StatefulSet).ObjectMeta.Annotations[sSetInputHashName]
 		if newSSetInputHash == oldSSetInputHash {
 			level.Debug(c.logger).Log("msg", "new statefulset generation inputs match current, skipping any actions")
-			return nil
+			continue
 		}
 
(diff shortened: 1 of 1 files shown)


mikechengwei:
@lilic @simonpasquier /cc

brancz:
Good catch! lgtm :+1:

PR #3999, opened 2021-04-22
thanos-ruler: Always add web port to pod specs

ricoberger:
When the Thanos Ruler CR is used together with Istio, we have to enable the listenLocal option, so that the Thanos Ruler isn't bind against the Pod IP.

When this option is enabled the "web" port isn't added to the Pod specification, so that the corresponding endpoints for the Thanos Ruler service doesn't contain the port. Because of this we can not use a ServiceMonitor to scrape the Prometheus metrics of the Thanos Ruler.

As workaround we can add the following in the Thanos Ruler CR:

```
spec:
  podMetadata:
    annotations:
      prometheus.io/scrape: "true"
      prometheus.io/port: "10902"
      prometheus.io/path: "/metrics"
```

When the `enablePrometheusMerge` option of Istio is enabled, this allows us to get the Thanos Ruler metrics via the metrics endpoint of the Istio Proxy.

```release-note: change
Add web port to Pod specification, when "listenLocal" option is "true"
```

### DIFF
--- a/pkg/thanos/statefulset.go
+++ b/pkg/thanos/statefulset.go
@@ -196,22 +196,25 @@ func makeStatefulSetSpec(tr *monitoringv1.ThanosRuler, config Config, ruleConfig
 		trCLIArgs = append(trCLIArgs, fmt.Sprintf("--alert.label-drop=%s", lb))
 	}
 
+	if tr.Spec.PortName == "" {
+		tr.Spec.PortName = defaultPortName
+	}
+
 	ports := []v1.ContainerPort{
+		{
+			Name:          tr.Spec.PortName,
+			ContainerPort: 10902,
+			Protocol:      v1.ProtocolTCP,
+		},
 		{
 			Name:          "grpc",
 			ContainerPort: 10901,
 			Protocol:      v1.ProtocolTCP,
 		},
 	}
+
 	if tr.Spec.ListenLocal {
 		trCLIArgs = append(trCLIArgs, "--http-address=localhost:10902")
-	} else {
-		ports = append(ports,
-			v1.ContainerPort{
-				Name:          tr.Spec.PortName,
-				ContainerPort: 10902,
-				Protocol:      v1.ProtocolTCP,
-			})
 	}
 
 	if tr.Spec.LogLevel != "" && tr.Spec.LogLevel != "info" {


s-urbaniak:
@ricoberger thank you for the contribution! I think, just changing this for Thanos Ruler might not be enough. We have this logic consistently across all other operator logic (alertmanager, prometheus) as well, so I suggest to change it there as well.

@prometheus-operator/prometheus-operator-reviewers generally: do you have concerns on exposing a web port despite listening to local only?

simonpasquier:
hmm interesting :) From what I can tell, @brancz added the ability to listen on localhost in https://github.com/prometheus-operator/prometheus-operator/pull/1051 and decided to not expose the container port in this case. I'm not sure there is another reason than "no need to declare a port that is only accessible from localhost". 

From the kube API docs about the `ports` field:


> List of ports to expose from the container. Exposing a port here gives the system additional information about the network connections a container uses, but is primarily informational. Not specifying a port here DOES NOT prevent that port from being exposed. Any port which is listening on the default "0.0.0.0" address inside a container will be accessible from the network. Cannot be updated.

ricoberger:
@s-urbaniak I can change this for the other components too, if everyone agrees, that this would make sens in general.

@simonpasquier I didn't verified this, but my assumption is that Prometheus doesn't add the Pods as target, when the port is missing in the specs, because the `targetPort` in the `thanos-ruler-operated` can not be "resolved":

```yaml
apiVersion: v1
kind: Service
metadata:
  name: thanos-ruler-operated
  namespace: monitoring
spec:
  ports:
  - name: http-web
    port: 10902
    protocol: TCP
    targetPort: http-web
  - name: grpc
    port: 10901
    protocol: TCP
    targetPort: grpc
  ...
```

I also checked ourI compared this with our Alertmanager setup, which is deployed using the [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack). The Helm chart creates an additional service `prometheus-alertmanager` next to the `alertmanager-operated` service:

```yaml
apiVersion: v1
kind: Service
  name: prometheus-alertmanager
  namespace: monitoring
spec:
  ports:
  - name: web
    port: 9093
    protocol: TCP
    targetPort: 9093
  ...
```

This service uses the the port number instead of the port name as `targetPort` and can then be used together with a ServiceMonitor.

So another solution also could be to just create an additional service manually.

ricoberger:
I would close this, because with Istio 1.10.x the `listenLocal` option isn't required anymore and without this options everything works like expected.

PR #4001, opened 2021-04-25
Thanos sidecar add ReadinessProbe and LivenessProbe

mikechengwei:
Thanos sidecar add ReadinessProbe and LivenessProbe.

```release-note:REPLACEME
Thanos sidecar add ReadinessProbe and LivenessProbe.
```


lilic (commented):
Thanks! What is the background behind this feature?

mikechengwei:
I think thanos sidecar  also need to check health status , so i submit this feature. WDYT? @lilic

PR #4008, opened 2021-04-27
Update troubleshooting.md

juandspy:
Fix the flow diagram as it was pointing to an invalid path.

```
release-note:bug
```


lilic (approved):
Nice thank you!

Not sure it needs to go to release notes as it's not user-facing API, but docs.

metalmatze:
I think it was the correct path on the website though, which in turn is now broken again... :grin: 
Might be better to revert this commit and leave a comment.

PR #4066, opened 2021-06-03
feat: unset storage.tsdb.no-lockfile flag

### LINKED ISSUE #4048: Setting the `--storage.tsdb.no-lockfile` isn't necessary anymore
**What happened?**

The operator sets the `--storage.tsdb.no-lockfile` flag, which needlessly disables the check for multiple instances of Prometheus using the same data directory.

Not set the `--storage.tsdb.no-lockfile` flag.

Start Prometheus, check out the UI, go to Status → Command Line Flags. Search for `storage.tsdb.no-lockfile`. It is set to `true`.

**Environment**

Any version after PR #855 gat merged.

Irrelevant.

* Kubernetes cluster kind:

Irrelevant.

* Manifests:

None are relevant.

* Prometheus Operator Logs:

N/A.

#855 introduced the flag because Prometheus back then had an unreliable locking implementation. It later reverted to the locking behavior of Prometheus 1.x, which AFAIK never caused any problem anywhere.

Jammicus:
Unsets storage.tsdb.no-lockfile flag as locking behaviour has been reverted.

Close #4048

```release-note:change
Flag "storage.tsdb.no-lockfile" will now default to false
```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -349,7 +349,6 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 		fmt.Sprintf("-storage.tsdb.path=%s", storageDir),
 		retentionTimeFlag+p.Spec.Retention,
 		"-web.enable-lifecycle",
-		"-storage.tsdb.no-lockfile",
 	)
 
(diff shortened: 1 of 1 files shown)


simonpasquier:
Thanks! lgtm

PR #4098, opened 2021-06-10
Add Wise to adopters list

jalev:
About time we added ourselves to this list :D

```release-note:none

```


simonpasquier (commented):
:+1: thanks for sharing, quite impressive numbers :)

PR #4117, opened 2021-06-21
Remove context.TODO() from client_go function calls

### LINKED ISSUE #3158: Remove `context.TODO()` used in Kubernetes client function calls
**What is missing?**
After https://github.com/coreos/prometheus-operator/pull/3154 merges, the prometheus-operator repo will contain `context.TODO()` in all Kubernetes client function calls.

We should follow up by plumbing real contexts through the call stack so that `context.TODO()` can be removed.

**Why do we need it?**
See https://github.com/coreos/prometheus-operator/pull/3154#issuecomment-617758087

**Environment**
Kubernetes 1.18+

**Anything else we need to know?**: Don't think so :)

slashpai:
Fixes #3158 

context.TODO() now remains only with informers which cannot be updated since they are generated code

```bash
░▒▓    ~/gi/s/prometheus-operator  on   remove_context_todo  find . -name "*go" | xargs grep -l context.TODO                PIPE ✘  took 38s   3.9.5   1.16.4   2.7.3   at admin ⎈  at 15:52:10  ▓▒░
./pkg/client/informers/externalversions/monitoring/v1/podmonitor.go
./pkg/client/informers/externalversions/monitoring/v1/prometheusrule.go
./pkg/client/informers/externalversions/monitoring/v1/servicemonitor.go
./pkg/client/informers/externalversions/monitoring/v1/alertmanager.go
./pkg/client/informers/externalversions/monitoring/v1/probe.go
./pkg/client/informers/externalversions/monitoring/v1/prometheus.go
./pkg/client/informers/externalversions/monitoring/v1/thanosruler.go
./pkg/client/informers/externalversions/monitoring/v1alpha1/alertmanagerconfig.go

░▒▓    ~/gi/s/prometheus-operator  on   remove_context_todo      
```

```release-note:REPLACEME

Remove context.TODO() from framework, e2e tests, amcg_test and k8sutil_test
```

### DIFF
--- a/pkg/alertmanager/amcfg_test.go
+++ b/pkg/alertmanager/amcfg_test.go
@@ -39,6 +39,7 @@ func TestGenerateConfig(t *testing.T) {
 		baseConfig alertmanagerConfig
 		amConfigs  map[string]*monitoringv1alpha1.AlertmanagerConfig
 		expected   string
+		ctx        context.Context
 	}
 
 	globalSlackAPIURL, err := url.Parse("http://slack.example.com")
@@ -61,6 +62,7 @@ receivers:
 - name: "null"
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "skeleton base with global send_revolved, no CRs",
@@ -81,6 +83,7 @@ receivers:
 - name: "null"
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "skeleton base with global smtp_require_tls set to false, no CRs",
@@ -101,6 +104,7 @@ receivers:
 - name: "null"
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "skeleton base with global smtp_require_tls set to true, no CRs",
@@ -121,6 +125,7 @@ receivers:
 - name: "null"
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "base with sub route, no CRs",
@@ -147,6 +152,7 @@ receivers:
 - name: custom
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "skeleton base, simple CR",
@@ -181,6 +187,7 @@ receivers:
 - name: mynamespace-myamc-test
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "skeleton base, CR with inhibition rules only",
@@ -231,6 +238,7 @@ receivers:
 - name: "null"
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "base with subroute, simple CR",
@@ -269,6 +277,7 @@ receivers:
 - name: mynamespace-myamc-test
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name: "CR with Pagerduty Receiver",
@@ -327,6 +336,7 @@ receivers:
   - routing_key: 1234abc
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name:    "CR with Webhook Receiver",
@@ -372,6 +382,7 @@ receivers:
   - url: http://test.url
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name: "CR with Opsgenie Receiver",
@@ -430,6 +441,7 @@ receivers:
   - api_key: 1234abc
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name: "CR with Opsgenie Team Responder",
@@ -495,6 +507,7 @@ receivers:
       type: team
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 			name: "CR with WeChat Receiver",
@@ -555,6 +568,7 @@ receivers:
     corp_id: wechatcorpid
 templates: []
 `,
+			ctx: context.Background(),
 		},
 		{
 
@@ -627,6 +641,7 @@ receivers:
         text: text
 templates: []
 `,
+			ctx: context.Background(),
 		},
 	}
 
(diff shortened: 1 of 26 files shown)


pkg/alertmanager/amcfg_test.go
```diff
@@ -39,6 +39,7 @@ func TestGenerateConfig(t *testing.T) {
 		baseConfig alertmanagerConfig
 		amConfigs  map[string]*monitoringv1alpha1.AlertmanagerConfig
 		expected   string
+		ctx        context.Context
```
dgrisonnet, line 42: I think it would make more sense to have the context global to the test function. I don't think we need to test different contexts in the cases, it should remain the same for each of them.

test/e2e/alertmanager_instance_namespaces_test.go
```diff
@@ -27,7 +27,8 @@ import (
 )
 
 func testAlertmanagerInstanceNamespacesAllNs(t *testing.T) {
-	ctx := framework.NewTestCtx(t)
+	testCtx := framework.NewTestCtx(t)
```
dgrisonnet, line 30: This variable is mostly unused, how about we make `NewTestCtx` return a pointer?

dgrisonnet (commented):
Just a few comments from my side, but it looks good overall.

To add more context around why the context of the generated informers can't be changed, it is because of the template they are generated from https://github.com/kubernetes/code-generator/blob/master/cmd/informer-gen/generators/informer.go#L143-L167. We could hack [generate-crds.go](https://github.com/prometheus-operator/prometheus-operator/blob/master/scripts/generate-crds.go) to patch the contexts, but I don't think it is worth diverging from what is generated.

simonpasquier:
@dgrisonnet regarding https://github.com/kubernetes/code-generator, I assume that it makes sense to address it upstream and add a `context.Context` parameter to the `NewFiltered*Informer()` funcs?

slashpai:
@dgrisonnet Addressed the comments

@simonpasquier Can you review too?

test/e2e/alertmanager_instance_namespaces_test.go
```diff
@@ -28,6 +28,7 @@ import (
 
 func testAlertmanagerInstanceNamespacesAllNs(t *testing.T) {
 	ctx := framework.NewTestCtx(t)
+
```
dgrisonnet, line 31: nit: there is no need for a new line here, usually we tend to put an object creation and its deferred deletion next to each other. It helps not forgetting about the cleanup.

dgrisonnet (approved):
Just a nitpick from my side, otherwise looks good to me :tada:.

> I assume that it makes sense to address it upstream and add a context.Context parameter to the NewFiltered*Informer() funcs?

Makes complete sense, I will open an issue.

dgrisonnet:
Opened https://github.com/kubernetes/code-generator/issues/128

test/e2e/alertmanager_instance_namespaces_test.go
slashpai: Fixed :)

slashpai: The test that was passing before failing after the change. To make sure its not because of new commit I will try removing it once and add back

slashpai: It was still failing so I added back the commit

slashpai: Understood the error now I removed the extra spaces in func calls. I was not aware of that looks like my IDE added those :|. Generate step is passed now

dgrisonnet (approved):
Looks good from my side :+1:

slashpai:
@simonpasquier Could you please take a look at this when you get chance

test/framework/namespace.go
```diff
@@ -39,27 +39,27 @@ func CreateNamespace(kubeClient kubernetes.Interface, name string) (*v1.Namespac
 	return namespace, nil
 }
 
-func (ctx *TestCtx) CreateNamespace(t *testing.T, kubeClient kubernetes.Interface) string {
+func (f *Framework) CreateNamespace(t *testing.T, ctx *TestCtx) string {
```
simonpasquier, line 42: we could collapse the `CreateNamespace()` function above in this function. AFAICT it's only used here so there's little point in having it separate.

test/framework/namespace.go
```diff
@@ -39,27 +39,27 @@ func CreateNamespace(kubeClient kubernetes.Interface, name string) (*v1.Namespac
...
 	}
 
 	ctx.AddFinalizerFn(namespaceFinalizerFn)
 
 	return name
 }
 
-func DeleteNamespace(kubeClient kubernetes.Interface, name string) error {
-	return kubeClient.CoreV1().Namespaces().Delete(context.TODO(), name, metav1.DeleteOptions{})
+func DeleteNamespace(kubeClient kubernetes.Interface, ctx context.Context, name string) error {
```
simonpasquier, line 57: We could make it a receiver function and get rid of the first args?

```suggestion
func (f *Framework) DeleteNamespace(name string) error {
```

test/framework/namespace.go
```diff
@@ -39,27 +39,27 @@ func CreateNamespace(kubeClient kubernetes.Interface, name string) (*v1.Namespac
...
+	return kubeClient.CoreV1().Namespaces().Delete(ctx, name, metav1.DeleteOptions{})
 }
 
-func AddLabelsToNamespace(kubeClient kubernetes.Interface, name string, additionalLabels map[string]string) error {
-	ns, err := kubeClient.CoreV1().Namespaces().Get(context.TODO(), name, metav1.GetOptions{})
+func AddLabelsToNamespace(kubeClient kubernetes.Interface, ctx context.Context, name string, additionalLabels map[string]string) error {
```
simonpasquier, line 61: same here

test/framework/namespace.go
```diff
@@ -72,16 +72,16 @@ func AddLabelsToNamespace(kubeClient kubernetes.Interface, name string, addition
...
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func RemoveLabelsFromNamespace(kubeClient kubernetes.Interface, name string, labels ...string) error {
-	ns, err := kubeClient.CoreV1().Namespaces().Get(context.TODO(), name, metav1.GetOptions{})
+func RemoveLabelsFromNamespace(kubeClient kubernetes.Interface, ctx context.Context, name string, labels ...string) error {
```
simonpasquier, line 83: same here

slashpai:
@simonpasquier Updated with changes
Hopefully the flaky test passes :crossed_fingers: in this run

simonpasquier (commented):
lgtm

simonpasquier:
Thanks!

PR #4123, opened 2021-06-22
chore: not need to return error

zhaogaolong:
change kube config not need return err

```release-note:REPLACEME
remove set kube config  return err
```


simonpasquier:
:wave: thanks for your PR. These files are auto-generated so you can't modify them manually. If anything it would be an issue against https://github.com/kubernetes/code-generator.

zhaogaolong:
> 👋 thanks for your PR. These files are auto-generated so you can't modify them manually. If anything it would be an issue against https://github.com/kubernetes/code-generator.

ok, I close it。

PR #4143, opened 2021-07-02
Add timeout to the informers cache synchronization

dgrisonnet:
During heavy apiserver disruptions, the informers cache synchronization
done by the operator before running the informers might never return.
This results in the operator being stuck forever waiting for its
informers caches to be synchronized.
To prevent that, we introduce a timeout of 10 minutes in the context
passed to client-go to make sure that the sync will always return at
some point.

**Release Note**

```release-note:bug
Add a 10 minutes timeout to the informers cache synchronization
```

/cc @simonpasquier

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -325,26 +325,27 @@ func SanitizeSTS(sts *appsv1.StatefulSet) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


paulfantom (approved):
lgtm on green

PR #4148, opened 2021-07-06
Backport #4143: Add timeout to the informers cache synchronization 

dgrisonnet:
Backport #4143 to release-0.48

During heavy apiserver disruptions, the informers cache synchronization
done by the operator before running the informers might never return.
This result in the operator being stuck forever waiting for its
informers caches to be synchronized.
To prevent that, we introduce a timeout of 10 minutes in the context
passed to client-go to make sure that the sync will always return at
some point.

```release-note:bug
Add a 10 minutes timeout to the informers cache synchronization
```

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -325,25 +325,28 @@ func SanitizeSTS(sts *appsv1.StatefulSet) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


dgrisonnet:
After discussing this backport with @simonpasquier, the bug has a very low impact on the overall prometheus-operator community, so it might not be a good candidate for an upstream backport.

PR #4149, opened 2021-07-06
pkg: exit early if caches can't be synced

simonpasquier:
If a cache sync times out, there's no need to continue since the
operator will exit eventually.

```release-note:none

```


paulfantom (approved):
/lgtm

dgrisonnet:
I think it would make sense to also backport this change to 0.48 along with #4148.

simonpasquier:
@dgrisonnet since it will part of the upcoming v0.49.0, I'm not sure I would worry about a v0.48.2.

dgrisonnet:
I will need the fix made to the informers cache synchronization in 0.48, so I opened a backport PR after discussing with @paulfantom.
Since this change also improves the cache synchronization workflow, I think it would great to backport it alongside the addition of the timeout, but that's not required to fix the issue I have in 0.48.

PR #4159, opened 2021-07-09
pkg/prometheus: add SendExemplars in RemoteWrite

coryrjohnson:
Adds support to the remote write component to send exemplars


mvkrishna86 (commented):
Changes are simple and good. But looks like this is not merged to master. Any plans to merge?

PR #3571, opened 2020-10-07
feat: add TargetLimit and EnforcedTargetLimit

arghya88:
Fixes #3499

PR #4163, opened 2021-07-13
docs: Update EnforcedTargetLimit description

bison:
The EnforcedTargetLimit configuration option was added in #3571, but
it seems where the docs say "if TargetLimit is higher that value will
be taken instead", that should instead read "if TargetLimit is lower."

```release-note:none
No release notes.
```


bison:
I think the doc for this is incorrect. The actual logic is a bit more complex, i.e. checking if either value is zero, but it seems it should read that the individual TargetLimit will be preferred if it is *lower* not *higher*. Let me know if I'm misreading.

Documentation/api.md
```diff
@@ -653,7 +653,7 @@ PrometheusSpec is a specification of the desired behavior of the Prometheus clus
...
 | allowOverlappingBlocks | AllowOverlappingBlocks enables vertical compaction and vertical query merge in Prometheus. This is still experimental in Prometheus so it may change in any upcoming release. | bool | false |
-| enforcedTargetLimit | EnforcedTargetLimit defines a global limit on the number of scraped targets. This overrides any TargetLimit set per ServiceMonitor or/and PodMonitor. It is meant to be used by admins to enforce the TargetLimit to keep overall number of targets under the desired limit. Note that if TargetLimit is higher that value will be taken instead. | *uint64 | false |
+| enforcedTargetLimit | EnforcedTargetLimit defines a global limit on the number of scraped targets. This overrides any TargetLimit set per ServiceMonitor or/and PodMonitor. It is meant to be used by admins to enforce the TargetLimit to keep overall number of targets under the desired limit. Note that if TargetLimit is lower that value will be taken instead. | *uint64 | false |
```
lilic, line 656:
```suggestion
| enforcedTargetLimit | EnforcedTargetLimit defines a global limit on the number of scraped targets. This overrides any TargetLimit set per ServiceMonitor or/and PodMonitor. It is meant to be used by admins to enforce the TargetLimit to keep the overall number of targets under the desired limit. Note that if TargetLimit is lower, that value will be taken instead. | *uint64 | false |
```

Lets then add a note that if the enforcedTargetLimit is 0, this means no limit is set. This is due to how Prometheus target_limit works https://prometheus.io/docs/prometheus/latest/configuration/configuration/

lilic (approved):
One suggestion otherwise lgtm, thanks!

Documentation/api.md
bison: Updated. Let me know if this is better.

bison: Ah, sorry I missed your other suggested change. I can update that as well.

lilic (approved):
/lgtm

Thanks!

Documentation/api.md
bison: Done.

PR #4166, opened 2021-07-14
Add API documentation backlinks

fpetkovski:
When browsing the API docs for the types provided by the operator,
it is hard to find out where exactly each type is used.

This commit modifies the code generation tool to add links under every
type to the types where it is referenced.

Here is an example of what the docs for `SecretOrConfigMap` would look like: https://github.com/fpetkovski/prometheus-operator/blob/add-api-docs-backlinks/Documentation/api.md#secretorconfigmap

```release-note:none

```


simonpasquier:
Neat! LGTM

PR #4183, opened 2021-07-23
Remove context field from framework struct in tests

slashpai:
## Description

Remove the context field from framework struct and updating method signatures to pass context instead. Doing this change since its not recommended to have context as a field in struct

## Type of change

## Changelog entry

```release-note
Remove context field from framework struct and pass as an argument instead of methods
```

cc: @simonpasquier

### DIFF
--- a/test/e2e/alertmanager_instance_namespaces_test.go
+++ b/test/e2e/alertmanager_instance_namespaces_test.go
@@ -27,7 +27,6 @@ import (
 )
 
 func testAlertmanagerInstanceNamespacesAllNs(t *testing.T) {
-	ctx := context.Background()
 	testCtx := framework.NewTestCtx(t)
 	defer testCtx.Cleanup(t)
 
@@ -41,37 +40,36 @@ func testAlertmanagerInstanceNamespacesAllNs(t *testing.T) {
 	//
 	// 3. "nonInstance" ns:
 	//   - hosts an Alertmanager CR which must not be reconciled
-	operatorNs := framework.CreateNamespace(ctx, t, testCtx)
-	instanceNs := framework.CreateNamespace(ctx, t, testCtx)
-	nonInstanceNs := framework.CreateNamespace(ctx, t, testCtx)
-	framework.SetupPrometheusRBACGlobal(ctx, t, testCtx, instanceNs)
+	operatorNs := framework.CreateNamespace(context.Background(), t, testCtx)
+	instanceNs := framework.CreateNamespace(context.Background(), t, testCtx)
+	nonInstanceNs := framework.CreateNamespace(context.Background(), t, testCtx)
+	framework.SetupPrometheusRBACGlobal(context.Background(), t, testCtx, instanceNs)
 
-	_, err := framework.CreatePrometheusOperator(ctx, operatorNs, *opImage, nil, nil, nil, []string{instanceNs}, false, true)
+	_, err := framework.CreatePrometheusOperator(context.Background(), operatorNs, *opImage, nil, nil, nil, []string{instanceNs}, false, true)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	am := framework.MakeBasicAlertmanager("non-instance", 3)
 	am.Namespace = nonInstanceNs
-	_, err = framework.MonClientV1.Alertmanagers(nonInstanceNs).Create(ctx, am, metav1.CreateOptions{})
+	_, err = framework.MonClientV1.Alertmanagers(nonInstanceNs).Create(context.Background(), am, metav1.CreateOptions{})
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	am = framework.MakeBasicAlertmanager("instance", 3)
 	am.Namespace = instanceNs
-	if _, err := framework.CreateAlertmanagerAndWaitUntilReady(ctx, instanceNs, am); err != nil {
+	if _, err := framework.CreateAlertmanagerAndWaitUntilReady(context.Background(), instanceNs, am); err != nil {
 		t.Fatal(err)
 	}
 
-	sts, err := framework.KubeClient.AppsV1().StatefulSets(nonInstanceNs).Get(ctx, "alertmanager-instance", metav1.GetOptions{})
+	sts, err := framework.KubeClient.AppsV1().StatefulSets(nonInstanceNs).Get(context.Background(), "alertmanager-instance", metav1.GetOptions{})
 	if !api_errors.IsNotFound(err) {
 		t.Fatalf("expected not to find an Alertmanager statefulset, but did: %v/%v", sts.Namespace, sts.Name)
 	}
 }
 
 func testAlertmanagerInstanceNamespacesDenyNs(t *testing.T) {
-	ctx := context.Background()
 	testCtx := framework.NewTestCtx(t)
 	defer testCtx.Cleanup(t)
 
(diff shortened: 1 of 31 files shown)


test/framework/framework.go
```diff
@@ -178,29 +176,31 @@ func (f *Framework) CreatePrometheusOperator(ns, opImage string, namespaceAllowl
 	namespaceDenylist, prometheusInstanceNamespaces, alertmanagerInstanceNamespaces []string,
 	createRuleAdmissionHooks, createClusterRoleBindings bool) ([]FinalizerFn, error) {
 
+	ctx := context.Background()
```
dgrisonnet, line 179: wouldn't it be best to pass the context to `CreatePrometheusOperator`?

test/e2e/alertmanager_instance_namespaces_test.go
```diff
@@ -26,8 +27,9 @@ import (
 )
 
 func testAlertmanagerInstanceNamespacesAllNs(t *testing.T) {
-	ctx := framework.NewTestCtx(t)
-	defer ctx.Cleanup(t)
+	ctx := context.Background()
+	testCtx := framework.NewTestCtx(t)
```
dgrisonnet, line 31: Since `testCtx` is oftentimes passed down to functions along with `ctx`, I think we should put the context inside of the TestCtx structure. This is a valid use case for putting context inside of a structure since the structure itself is meant only to define the lifecycle of the test so its goal is clear.

slashpai: I remember we had added ctx like that previously inside testCtx. But since it's not recommended approach should we avoid that?

dgrisonnet: Yes, it is not recommended to add a context to structs since it would lead to some confusion but in our case, we are just adding it to a structure that defines a context, so I think it is a valid use case.

Also, maybe a better approach would be to update the TextCtx structure to extend the context.Context interface with the [testCtx Cleanup](https://github.com/prometheus-operator/prometheus-operator/blob/a10ba4a4e57144a21c9be6c4e5fad456b0433518/test/framework/context.go#L59-L69) function as a cancel function.

simonpasquier: Sorry for the late review. I think it would even be clearer to instantiate a new `context.Background()` for each function call that expect a context (e.g no need to pass a unique context from top-level to down). 

Instead of 
https://github.com/prometheus-operator/prometheus-operator/blob/83fe36566f4e0894eb5ffcd2638a0f039a17bdeb/test/framework/alertmanager.go#L124

We can do

```golang
 _, err = f.KubeClient.CoreV1().Secrets(ns).Create(context.Background(), s, metav1.CreateOptions{})
```

From https://pkg.go.dev/context#Background:

> Background returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and **tests**, and as the top-level Context for incoming requests.

slashpai: @simonpasquier  I have passed ctx variable which is declared in each test 

like in https://github.com/prometheus-operator/prometheus-operator/pull/4183/files#diff-06a2e1bc05a7f1185bff95c6029e8b10a257bee7b89ffa57b226ca2d19f24498R47

Do you think current change is ok?

simonpasquier: My current idea would be to avoid passing a top-level context around because in the end, it's always `context.Background()` anyway. 
In practice, the change would be to remove the `Ctx` field from the `framework.Framework` struct and everywhere we used to have `f.Ctx`, replace by `context.Background()`.

slashpai: Since current update is creating a variable to hold context.Background() like [here](https://github.com/prometheus-operator/prometheus-operator/pull/4183/files#diff-bd371abb25ee1c95090cf7508fc3133e1fb6faae26c4c01fdac53dd759052d3eR31). I felt that looks good to read than creating a new context object in every function call needing a context to be passed.

slashpai: After discussing offline we decided to use context.Background places where both ctx and testCtx is passed for better readability of code

test/framework/namespace.go
```diff
@@ -25,10 +26,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-func (f *Framework) CreateNamespace(t *testing.T, ctx *TestCtx) string {
-	name := ctx.GetObjID()
+func (f *Framework) CreateNamespace(ctx context.Context, t *testing.T, testCtx *TestCtx) string {
```
simonpasquier, line 29:
```suggestion
func (f *Framework) CreateNamespace(t *testing.T, testCtx *TestCtx) string {
```

test/framework/namespace.go
```diff
@@ -25,10 +26,10 @@ import (
...
+	name := testCtx.GetObjID()
 
-	_, err := f.KubeClient.CoreV1().Namespaces().Create(f.Ctx, &v1.Namespace{
+	_, err := f.KubeClient.CoreV1().Namespaces().Create(ctx, &v1.Namespace{
```
simonpasquier, line 32:
```suggestion
	_, err := f.KubeClient.CoreV1().Namespaces().Create(context.Background(), &v1.Namespace{
```

test/framework/namespace.go
```diff
@@ -39,20 +40,20 @@ func (f *Framework) CreateNamespace(t *testing.T, ctx *TestCtx) string {
...
 
-	ctx.AddFinalizerFn(namespaceFinalizerFn)
+	testCtx.AddFinalizerFn(namespaceFinalizerFn)
 
 	return name
 }
 
-func (f *Framework) DeleteNamespace(name string) error {
-	return f.KubeClient.CoreV1().Namespaces().Delete(f.Ctx, name, metav1.DeleteOptions{})
+func (f *Framework) DeleteNamespace(ctx context.Context, name string) error {
```
simonpasquier, line 51: ditto about removing `ctx` here

test/framework/namespace.go
```diff
@@ -39,20 +40,20 @@ func (f *Framework) CreateNamespace(t *testing.T, ctx *TestCtx) string {
...
+	return f.KubeClient.CoreV1().Namespaces().Delete(ctx, name, metav1.DeleteOptions{})
 }
 
-func (f *Framework) AddLabelsToNamespace(name string, additionalLabels map[string]string) error {
-	ns, err := f.KubeClient.CoreV1().Namespaces().Get(f.Ctx, name, metav1.GetOptions{})
+func (f *Framework) AddLabelsToNamespace(ctx context.Context, name string, additionalLabels map[string]string) error {
```
simonpasquier, line 55: same here

simonpasquier (commented):
Good! I think we can take it even further: practically we shouldn't any function in `test/e2` or `test/framework` taking a `ctx context.Context` parameter.

simonpasquier:
It needs a rebase :)

test/framework/namespace.go
slashpai: > Good! I think we can take it even further: practically we shouldn't any function in `test/e2` or `test/framework` taking a `ctx context.Context` parameter.

mm but looks like then we are replacing context.TODO() which we had before with context.Background() in all client-go calls. If I remember correctly we wanted to pass them to method calls so instead of context.TODO() we can pass context created at one place. 

For methods that doesn't involve testCtx do you think its fine to pass context object like we have now?. I find it more better :)

slashpai: @simonpasquier Please take a look when you get a chance :)

slashpai:
@simonpasquier Can you take a look at this one :)

simonpasquier:
sorry for the lag :) can you rebase? I should be able to merge then.

slashpai:
> sorry for the lag :) can you rebase? I should be able to merge then.

@simonpasquier  Updated, PTAL

simonpasquier:
Some e2e tests don't compile :)

simonpasquier:
Thanks!

PR #4193, opened 2021-07-29
Update rules in clusterrole

April-Q:
## Description
According to the rule:   [user-facing clusterroles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles), this pull request update the rules in edit role.


paulfantom (changes requested):
This would mean prometheus-operator is no longer able to get objects that it is supposed to manage.

PR #4205, opened 2021-08-11
Fix one logical error

mikechengwei:
I find one logical error.

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -201,11 +201,11 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


coderanger:
Why do you think this is an error?

mikechengwei:
https://github.com/prometheus-operator/prometheus-operator/blob/fb24ca8fd26597c5a2d69ccac949aa5a76cf29ce/pkg/prometheus/promcfg.go#L184-L210

I think this code is a bit complicated and should be simplified.
I understood from the beginning that if the user does not spec the external label name, we will add a default value. I checked the code carefully and found that if we spec the external label name, we will set a default value.
I think we should simplify it. @coderanger

mikechengwei:
Sorry, I think you are right after i review the code again.

PR #4034, opened 2021-05-13
Root AlertmanagerConfig name in Alertmanager spec, skips namespace matcher

Shashankft9:
This PR addresses the first objective here: https://github.com/prometheus-operator/prometheus-operator/issues/4033
Other issue: https://github.com/prometheus-operator/prometheus-operator/issues/3737
Motivation from the work done here: https://github.com/prometheus-operator/prometheus-operator/pull/3821

Usage: 
Add this to the Alertmanager Spec:
```
  baseAlertmanagerConfiguration:
    alertmanagerConfigName: config-example
```
And then create an AlertmanagerConfig in the same namespace with the name `config-example`, all the routes mentioned here wont have the namespace matcher and will be used to setup base alertmanager secret (`baseConfig.Route`), that is - unlike other `alertmanagerConfigs` where the `spec.Route` transforms into `baseConfig.Route.Routes[]`

 
**Release Note Template (will be copied)**

```
Specify a root AlertmangerConfig in the Alertmanager spec, for which namespace matcher will be skipped. 
```

Issue #4033, opened 2021-05-13
Root AlertmanagerConfig and namespace matcher as optional


Shashankft9:
This is to track the objectives of comment added here: 
https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-797577859

As I can understand, these are the pointers on which work could be done: 
- adding the `baseAlertmanagerConfiguration` with root `alertmanagerConfigName` as a field in the alertmanager CRD spec. For this particular `alertmanagerConfig`, namespace matcher will be skipped.
- adding `global` and `templates` keys.
- adding `enforcedNamespaceLabel`.

LiuVII:
and update?

simonpasquier:
I believe that the main concerns have been resolved with the latest v0.61.1:
* One can define the main configuration using  `AlertmanagerConfig` CRD (with the `spec.alertmanagerConfiguration` field).
* It is possible to disable the namespace enforcement when AlertmanagerConfig objects are merged in the resulting configuration using `spec.alertmanagerConfigMatcherStrategy.type: None`.

PR #4220, opened 2021-08-18
✨ feat(alertmanager): global alertmanager config

sunlintong:
## Description

As disscussed at #3737, this commit support config global alertmanager config. Now support:
```yaml
kind: Alertmanager
metadata:
  name: myalertmanager
spec:
  ...
  globalAlertmanagerConfig:
    name: myGlobalAlertmanagerConfig # the global alertmanagerconfig name which is in the namespace
    global:
      resolve_timeout: 10m 
    templates:
      - email.yaml
```
In my environment, is works well. And
- `global` only support `resolve_timeout` now, if need, I can add more,
- if you think this way is appropriate, I will complete the unit test and doc.

## Type of change


sunlintong:
Is there any one to have a look at ?

Shashankft9:
hey @sunlintong please note that I had updated PR #4034 just couple of weeks ago, so it was NOT abandoned, just that there wasn't any update by a maintainer. 
@simonpasquier we can go ahead with this PR if this is the better solution, happy to close my PR, the feature is most important and seems to be very much sought out, so please go ahead with a review here.

sunlintong:
Hi @Shashankft9, I noticed your PR #4034 two weeks ago. It can also resolve my problem, but they didn't merge it so I am so confused.
I noticed that they seems to want `global` and `template`  key which your PR not refer, so I opened this PR.
But anyway ,any PR related to #3737 merged I'll be very happy and close my PR.

Shashankft9:
@simonpasquier ping :)

simonpasquier:
When we discussed #3737, I've been convinced by @brancz that embedding an AlertmanagerConfig type in the Alertmanager CRD  is a better solution than referencing a resource by name. I think that it's more explicit for users and minimizes the risk of errors (like mistyping the resource name). As I said earlier I'd be fine with duplicating in `pkg/apis/monitoring/v1alpha1` the few structs from `pkg/apis/monitoring/v1` that AlertmanagerConfig depends on.

sunlintong:
I understand what you mean. Do you need me to finish this in a new PR or just in this PR?

simonpasquier:
If you can continue on this PR, it should be ok :+1:

sunlintong:
> regarding the cyclic dependencies, from what I can tell, the 2 structs imported in the `v1alpha1` package from the `v1` package are `BasicAuth` and `SafeTLSConfig`. I suggest that we duplicate the struct definitions in the `v1alpha1` package.

I have tried as you say, and I found it's not simple.
- First, if duplicate whese struct, the validate method also need to be duplicated. eg. 
```go
func (c *SafeTLSConfig) Validate() error
...
func (c *SafeAuthorization) Validate() error
...
func (c *Authorization) Validate() error
...
func (c *SecretOrConfigMap) Validate() error
...
```
- Secondly, It's a problem to choose which packages to use when they are some mothod's parameter, like in `amcfg.go`:
```go
func (cg *configGenerator) convertTLSConfig(ctx context.Context, in *monitoringv1.SafeTLSConfig, crKey types.NamespacedName) tlsConfig 
```
In this case, should we use `*monitoringv1.SafeTLSConfig` or `*v1.SafeTLSConfig`? Maybe we need to use both.
- Thirdly, as I said above
> Embedding an AlertmanagerConfig type in the Alertmanager CRD maybe make the Alertmanager complicated and bloated. I think the Alertmanager CRD shoud focus on it's own funtion ---- config the alertmanager instance, not the alert config

**So, I think referencing the global resource by name is a better solution.**

simonpasquier:
@sunlintong I hear your concerns. Lets try first to make it work by referencing an AlertmanagerConfig resource by name. I'll try to think about if/how we can proceed with embedding the resource...

simonpasquier (commented):
IMO we need to process the global config apart the namespaced AlertmanagerConfig resources because it should be appended

https://github.com/prometheus-operator/prometheus-operator/blob/af9c197bc89da0428fa3c72645299977fd982390/pkg/alertmanager/amcfg.go#L121-L126

Also I assume that the raw config secret and the global config are exclusive: we should use one or the other but not both at the same time. I would say that if the secret is defined it should take precedence?

sunlintong:
Oh..., I forget there is a raw config secret.
Yes I agree with you that we should only use one of the raw config secret and the global alertmanager config. They have some repetitive functions. 
So, if we want some alertmanager config to ignore namespace match, as you said at 
[#3737#issuecomment-797577859](https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-797577859) 
>A further improvement would be to add an enforcedNamespaceLabel flag to the Alertmanager CRD that enables/disables the namespace enforcement in the routing tree for user-defined AlertmanagerConfig resources as well as a way to exclude enforcement for some AlertmanagerConfig resources if namespace enforcement is enabled (this would be consistent with the enforcedNamespaceLabel and prometheusRulesExcludedFromEnforce fields from the Prometheus CRD). Whether namespaces are enforced or not, all first level routes would be set with continue: true to make sure that alerts are evaluated by all AlertmanagerConfigs.

May be a better way.
We just specify `enforcedNamespaceLabel` to add namespace matcher, and this value default true` to avoid excessive impact.
Here to check this value:
https://github.com/prometheus-operator/prometheus-operator/blob/af9c197bc89da0428fa3c72645299977fd982390/pkg/alertmanager/amcfg.go#L110
And here to add matcher:
https://github.com/prometheus-operator/prometheus-operator/blob/af9c197bc89da0428fa3c72645299977fd982390/pkg/alertmanager/amcfg.go#L149-L152

simonpasquier:
Hmm I'd really like that users can get rid of the raw config secret: it should be possible to fully define the configuration with Alertmanager and AlertmanagerConfig custom resources (e.g. including global and templates). 

I think that we need to identify *one* AlertmanagerConfig which will be used for the root of the routing tree so I'm not sure that `alertmanagerConfigsExcludedFromEnforce` would work.

sunlintong:
OK, let me have a try, I'll  process the global config apart the namespaced AlertmanagerConfig resources.
How about make global config take precedence?, So global config will overwrite the raw config secret, so the further version we can drop the secret.

simonpasquier:
>  How about make global config take precedence?,So global config will overwrite the raw config secret, the further version we can drop the secret.

I agree, if you reference a global configuration, you want it to be applied. We won't drop the raw secret anytime soon though since it would imply a new API version.

sunlintong:
@simonpasquier  I updated the code, if you agree this way. I'll complete the 
```go
func convertGlobal(in *monitoringv1.GlobalConfig) (*globalConfig, error) 
```
function to support all globalConfig. Then complete the unit test.

james-callahan:
Alternative implementation idea: what if there was an annotation you could add to `AlertmanagerConfig` objects to say "this is global"?

superfly-git:
Any news on the above seems like it stuck on the last step?? This would be very helpfull if it gets merged for me as well.

sunlintong:
Some things bothered me for a few weeks, I'll update this PR in recent days anyway.

sunlintong:
@simonpasquier I've pushed again to resolve the conflict, can you have a look at?

simonpasquier (commented):
Thanks for resuming the PR. I've focused my review on the CRD fields for now but keep it going!

pkg/apis/monitoring/v1/types.go
```diff
@@ -1603,6 +1603,25 @@ type AlertmanagerSpec struct {
 	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
 	// +optional
 	MinReadySeconds *uint32 `json:"minReadySeconds,omitempty"`
+	// GlobalAlertmanagerConfig used to set the global alertmanager config.
+	GlobalAlertmanagerConfig *GlobalAlertmanagerConfig `json:"globalAlertmanagerConfig,omitempty"`
+}
+
+type GlobalAlertmanagerConfig struct {
```
PhilipGough, line 1610: Are we ok with adding and promoting these types straight into `v1`? Perhaps it is fine, but I am just wondering what kind of stability guarantees for this feature we are advertising by doing so?

simonpasquier: You're right: the usual way is to mark the field as EXPERIMENTAL (we have this mention for the `shards` field of the Prometheus CRD for instance). It isn't ideal but this is the best we can do because the other option would be to have a v2alpha CRD. 
Having said that, we could improve the situation a bit by using feature gates so that operator admins control exactly what users can/can't do (e.g. `--feature-gates=GlobalAlertmanagerConfig=true`).

pkg/apis/monitoring/v1/types.go
```diff
@@ -1603,6 +1603,25 @@ type AlertmanagerSpec struct {
...
+	// Name specify the gloal AlertmanagerConfig name which is in the same namespace with alertmanager.
+	Name string `json:"name,omitempty"`
+	// Global specify the global alertmanager config, current only support resolve_timeout.
+	// More info: https://prometheus.io/docs/alerting/latest/configuration
+	Global *GlobalConfig `json:"global,omitempty"`
+	// Templates specify the file paths which custom notification template definitions are read.
+	Templates []string `json:"templates,omitempty"`
```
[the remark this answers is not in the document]
sunlintong: The `templates` config here used to overwrite the raw config secret. But now I'll drop it too. Maybe only support globalAlertmanagerConfig name first is a better way
```yaml
spec:
  globalAlertmanagerConfig:
    name: myGlobalAlertmanagerConfig 
```

sunlintong:
@simonpasquier I have updated the code again refer to your suggestions and pull #4424.
BTW, like `enforceNamespaceForRoute`
https://github.com/prometheus-operator/prometheus-operator/blob/787f54b055f797464d5832ace1c7f8318321c87a/pkg/alertmanager/amcfg.go#L179-L181
For inhibitRules, maybe also need  a function `enforceNamespaceForInhibitRule`. So it's  easier to support no force namespace inhibitRule.

bergerx:
@simonpasquier sorry for pinging you, I know you are busy with all your daily struggles. But do you mind having a look here when you have some free cycles?
I pinged you just because you requested some changes previously that are addressed by @sunlintong, and thought you may just have missed the last push and comment.

pkg/apis/monitoring/v1/types.go
```diff
@@ -1603,6 +1603,13 @@ type AlertmanagerSpec struct {
 	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
 	// +optional
 	MinReadySeconds *uint32 `json:"minReadySeconds,omitempty"`
+	// GlobalAlertmanagerConfig used to set the global alertmanager config.
+	GlobalAlertmanagerConfig *GlobalAlertmanagerConfig `json:"globalAlertmanagerConfig,omitempty"`
+}
+
+type GlobalAlertmanagerConfig struct {
+	// Name specify the gloal AlertmanagerConfig name which is in the same namespace with alertmanager.
```
james-callahan, line 1611:
```suggestion
	// Name specify the global AlertmanagerConfig name which is in the same namespace as alertmanager.
```

jsonnet/prometheus-operator/alertmanagers-crd.json
```diff
@@ -1906,6 +1906,16 @@
                     "description": "ForceEnableClusterMode ensures Alertmanager does not deactivate the cluster mode when running with a single replica. Use case is e.g. spanning an Alertmanager cluster across Kubernetes clusters with a single replica in each.",
                     "type": "boolean"
                   },
+                  "globalAlertmanagerConfig": {
+                    "description": "GlobalAlertmanagerConfig used to set the global alertmanager config.",
+                    "properties": {
+                      "name": {
+                        "description": "Name specify the gloal AlertmanagerConfig name which is in the same namespace with alertmanager.",
```
james-callahan, line 1913:
```suggestion
                        "description": "Name specify the global AlertmanagerConfig name which is in the same namespace as alertmanager.",
```

example/prometheus-operator-crd/monitoring.coreos.com_alertmanagers.yaml
```diff
@@ -2176,6 +2176,15 @@ spec:
                   Use case is e.g. spanning an Alertmanager cluster across Kubernetes
                   clusters with a single replica in each.
                 type: boolean
+              globalAlertmanagerConfig:
+                description: GlobalAlertmanagerConfig used to set the global alertmanager
+                  config.
+                properties:
+                  name:
+                    description: Name specify the gloal AlertmanagerConfig name which
+                      is in the same namespace with alertmanager.
```
james-callahan, line 2185:
```suggestion
                    description: Name specify the global AlertmanagerConfig name which
                      is in the same namespace as alertmanager.
```

bundle.yaml
```diff
@@ -5023,6 +5023,15 @@ spec:
                   Use case is e.g. spanning an Alertmanager cluster across Kubernetes
                   clusters with a single replica in each.
                 type: boolean
+              globalAlertmanagerConfig:
+                description: GlobalAlertmanagerConfig used to set the global alertmanager
+                  config.
+                properties:
+                  name:
+                    description: Name specify the gloal AlertmanagerConfig name which
+                      is in the same namespace with alertmanager.
```
james-callahan, line 5032:
```suggestion
                    description: Name specify the global AlertmanagerConfig name which
                      is in the same namespace as alertmanager.
```

Documentation/api.md
```diff
@@ -358,6 +360,19 @@ Endpoint defines a scrapeable endpoint serving Prometheus metrics.
...
+## GlobalAlertmanagerConfig
+
+
+
+
+<em>appears in: [AlertmanagerSpec](#alertmanagerspec)</em>
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| name | Name specify the gloal AlertmanagerConfig name which is in the same namespace with alertmanager. | string | false |
```
james-callahan, line 372:
```suggestion
| name | Name specify the global AlertmanagerConfig name which is in the same namespace as alertmanager. | string | false |
```

james-callahan (changes requested):

sunlintong:
@james-callahan Thanks for suggestions. 
And @simonpasquier , I force pushed again refer to your PR #4441 and @james-callahan suggestions, do you mind have a look at?

james-callahan (approved):
Seems good enough to me.
I'd be happy to have it without the feature gate @simonpasquier mentioned: this feature degrades to a no-op.

sunlintong:
Update: Remove the global alertmanager config after generateGlobalConfig to avoid generating again.
As: https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/b1fee0a9b4b5b8ee8bd697cf87d8506243eefd87

sunlintong:
Doc check seems failed, what's the wrong?
```log
>> checking formatting and local/remote links
/home/runner/work/prometheus-operator/prometheus-operator/tmp/bin/mdox fmt --soft-wraps --check -l --links.localize.address-regex="https://prometheus-operator.dev/.*" --links.validate.config-file=.mdox.validate.yaml Documentation/design.md Documentation/custom-configuration.md Documentation/rbac-crd.md Documentation/exposing-metrics.md Documentation/rbac.md Documentation/additional-scrape-config.md Documentation/high-availability.md Documentation/troubleshooting.md Documentation/thanos.md Documentation/user-guides/getting-started.md Documentation/user-guides/exposing-prometheus-and-alertmanager.md Documentation/user-guides/basic-auth.md Documentation/user-guides/shards-and-replicas.md Documentation/user-guides/strategic-merge-patch.md Documentation/user-guides/linting.md Documentation/user-guides/storage.md Documentation/user-guides/running-exporters.md Documentation/user-guides/monitoring-kubernetes-ingress.md Documentation/user-guides/alerting.md Documentation/user-guides/webhook.md Documentation/network-policies.md ADOPTERS.md CHANGELOG.md CONTRIBUTING.md MAINTAINERS.md README.md RELEASE.md SECURITY.md code-of-conduct.md governance.md
error: 		ADOPTERS.md:174: "https://polarsignals.com/" not accessible even after retry; status code 0: Get "https://www.polarsignals.com/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
make: *** [Makefile:278: check-docs] Error 1
Error: Process completed with exit code 2.
```

PhilipGough:
@sunlintong > Doc check seems failed, what's the wrong?

Try rebase on main, should resolve the issue - regardless, it is not a required check

simonpasquier:
@sunlintong as I commented on another linked issue, I'll free up some time next week to review your pull request. Thanks again for your patience!

Documentation/api.md
```diff
@@ -237,6 +238,7 @@ AlertmanagerSpec is a specification of the desired behavior of the Alertmanager
 | alertmanagerConfigSelector | AlertmanagerConfigs to be selected for to merge and configure Alertmanager with. | *[metav1.LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#labelselector-v1-meta) | false |
 | alertmanagerConfigNamespaceSelector | Namespaces to be selected for AlertmanagerConfig discovery. If nil, only check own namespace. | *[metav1.LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#labelselector-v1-meta) | false |
 | minReadySeconds | Minimum number of seconds for which a newly created pod should be ready without any of its container crashing for it to be considered available. Defaults to 0 (pod will be considered available as soon as it is ready) This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate. | *uint32 | false |
+| globalAlertmanagerConfig | GlobalAlertmanagerConfig used to set the global alertmanager config. The specified global alertmanager config will not force add a namespace label in routes and inhibitRules. | *[GlobalAlertmanagerConfig](#globalalertmanagerconfig) | false |
```
PhilipGough, line 241: Additionally I'd lean towards the inclusion of a standalone section in https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/alerting.md perhaps to which defines and describes the behaviours and the use case?

pkg/apis/monitoring/v1/types.go
```diff
@@ -1605,6 +1605,16 @@ type AlertmanagerSpec struct {
...
 	MinReadySeconds *uint32 `json:"minReadySeconds,omitempty"`
+	// GlobalAlertmanagerConfig used to set the global alertmanager config.
+	// The specified global alertmanager config will not force add a namespace label in routes and inhibitRules.
+	GlobalAlertmanagerConfig *GlobalAlertmanagerConfig `json:"globalAlertmanagerConfig,omitempty"`
+}
+
+// GlobalAlertmanagerConfig used to set the global alertmanager config.
+// +k8s:openapi-gen=true
+type GlobalAlertmanagerConfig struct {
+	// Name specify the global AlertmanagerConfig name which is in the same namespace as alertmanager.
```
PhilipGough, line 1616:
```suggestion
// Name specifies the global AlertmanagerConfig name which must be in the same namespace as Alertmanager.
```

pkg/alertmanager/amcfg.go
```diff
@@ -128,6 +128,56 @@ func validateConfigInputs(am *monitoringv1.Alertmanager) error {
...
+			mti, err := convertMuteTimeInterval(&muteTimeInterval, crKey)
+			if err != nil {
+				return errors.Wrapf(err, "AlertmanagerConfig %s", crKey.String())
+			}
+			baseConfig.MuteTimeIntervals = append(baseConfig.MuteTimeIntervals, mti)
+		}
+
+		// Remove the global alertmanager config to avoid generating again
+		delete(amConfigs, crKey.String())
+	}
```
PhilipGough, line 176: I think we should have some handling for the `!ok` case here also. Perhaps simply logging the issue is ok, but if I have configured the fields and the lookup fails, I would expect to be informed.

PhilipGough:
@sunlintong, sorry about the delay - added some comments, overall shape looks good to me. Think we will need some e2e tests in https://github.com/prometheus-operator/prometheus-operator/blob/main/test/e2e/alertmanager_test.go and some additional unit tests.

sunlintong:
@PhilipGough I've updated code in commit [db295f4](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/db295f4012c9d626a2fefef890c251ba7df36d63) as suggestions. 

For unit tests and e2e tests, Maybe I can complete in the next few days when I'm free :)

Documentation/user-guides/alerting.md
```diff
@@ -91,6 +91,24 @@ spec:
       alertmanagerConfig: example
 ```
 
+## Specify Global Alertmanager Config
```
simonpasquier, line 94: I think that the document already needs a rewrite to separate clearly the configuration of Alertmanager and the AlertmanagerConfig CRD but it can be addressed after this PR merges.

pkg/apis/monitoring/v1/types.go
simonpasquier: It was discussed in a previous contributors office hours meeting and the consensus was not to rely on feature gates (as long as the field is marked as EXPERIMENTAL).

pkg/apis/monitoring/v1/types.go
```diff
@@ -1605,6 +1605,16 @@ type AlertmanagerSpec struct {
 	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
 	// +optional
 	MinReadySeconds *uint32 `json:"minReadySeconds,omitempty"`
+	// GlobalAlertmanagerConfig used to set the global alertmanager config.
```
simonpasquier, line 1608:
```suggestion
	// EXPERIMENTAL: globalAlertmanagerConfig specifies the global Alertmanager configuration.
	// If defined, it takes precedence over the `configSecret` field.
	// This field may change in future releases.  
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1605,6 +1605,16 @@ type AlertmanagerSpec struct {
...
+	// The specified global alertmanager config will not force add a namespace label in routes and inhibitRules.
+	GlobalAlertmanagerConfig *GlobalAlertmanagerConfig `json:"globalAlertmanagerConfig,omitempty"`
+}
+
+// GlobalAlertmanagerConfig used to set the global alertmanager config.
+// +k8s:openapi-gen=true
+type GlobalAlertmanagerConfig struct {
+	// Name specifies the global AlertmanagerConfig name which must be in the same namespace as Alertmanager.
```
simonpasquier, line 1616:
```suggestion
	// The name of the AlertmanagerConfig resource which holds the global configuration.
	// It must be in the same namespace as the Alertmanager.
```

pkg/alertmanager/amcfg.go
```diff
@@ -128,6 +128,58 @@ func validateConfigInputs(am *monitoringv1.Alertmanager) error {
 	return nil
 }
 
+func (cg *configGenerator) generateGlobalConfig(
+	ctx context.Context,
+	baseConfig *alertmanagerConfig,
+	am *monitoringv1.Alertmanager,
+	amConfigs map[string]*monitoringv1alpha1.AlertmanagerConfig,
+) error {
```
simonpasquier, line 136: The global `AlertmanagerConfig` object might not exist in the amConfigs map (for instance, `spec.alertmanagerConfigNamespaceSelector` might not select the Alertmanager's namespace). It should be up to the caller to retrieve the `AlertmanagerConfig` object, handle the error case and pass it to this method.

```suggestion
func (cg *configGenerator) generateGlobalConfig(
	ctx context.Context,
	amConfig *monitoringv1alpha1.AlertmanagerConfig,
) (*alertmanagerConfig, error) {
```

pkg/alertmanager/operator.go
```diff
@@ -920,6 +920,9 @@ receivers:
 	}
 
 	generator := newConfigGenerator(namespacedLogger, version, store)
+	if err := generator.generateGlobalConfig(ctx, baseConfig, am, amConfigs); err != nil {
```
simonpasquier, line 923: I think that the operator should branch earlier depending on whether `spec.globalAlertmanagerConfig` is defined or not:
* if defined, retrieve the global `AlertmanagerConfig` object and bail out if something returns an error (missing ref, invalid object).
* otherwise work with the raw configuration secret.

pkg/apis/monitoring/v1/types.go
```diff
@@ -1605,6 +1605,16 @@ type AlertmanagerSpec struct {
...
+	Name string `json:"name,omitempty"`
```
simonpasquier, line 1617: name shouldn't be empty
```suggestion
	// +kubebuilder:validation:MinLength=1
	Name string `json:"name,omitempty"`
```

simonpasquier (commented):
Thanks for the hard work @sunlintong! I made a few comments but I like that the code changes are reasonable.
Something else to keep in mind is that we should probably remove the global AlertmanagerConfig object from the selected AlertmanagerConfigs if it is present. Otherwise the generated configuration would include it twice (both with and without namespace enforcement).

sunlintong:
@simonpasquier  I've updated code in commit [3822b82](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/3822b82085ec1ce783648b69958e6def94bcabb2) as your suggestions.
I force pushed because I rebase branch main again, hahah.
At the same time, e2e tests are still not finished and doc in [alerting.md]( https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/alerting.md) maybe need a new PR :)

PhilipGough:
> doc in [alerting.md](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/alerting.md) maybe need a new PR :)

@sunlintong - that is totally fine, and even preferable :)

pkg/alertmanager/amcfg.go
```diff
@@ -128,6 +128,56 @@ func validateConfigInputs(am *monitoringv1.Alertmanager) error {
...
+	globalAlertmanagerConfig := &alertmanagerConfig{
+		Receivers: []*receiver{
+			{
+				Name: "null",
+			},
+		},
+		Route: &route{
+			Receiver: "null",
+		},
+	}
```
simonpasquier, line 144: We shouldn't pre-set any fields.
```suggestion
	globalAlertmanagerConfig := &alertmanagerConfig{}
```

pkg/alertmanager/amcfg.go
```diff
@@ -128,6 +128,56 @@ func validateConfigInputs(am *monitoringv1.Alertmanager) error {
...
+	// Add inhibitRules to baseConfig.InhibitRules without enforce namespace
+	for _, inhibitRule := range amConfig.Spec.InhibitRules {
+		globalAlertmanagerConfig.InhibitRules = append(globalAlertmanagerConfig.InhibitRules, cg.convertInhibitRule(&inhibitRule))
+	}
+
+	// Add routes to baseConfig.Route.Routes without enforce namespace
+	globalAlertmanagerConfig.Route.Routes = []*route{
+		cg.convertRoute(amConfig.Spec.Route, crKey),
+	}
```
simonpasquier, line 157: We should update the root entry
```suggestion
	globalAlertmanagerConfig.Route = cg.convertRoute(amConfig.Spec.Route, crKey)
```

pkg/alertmanager/operator.go
```diff
@@ -933,6 +947,43 @@ receivers:
...
+// find and delete global AlertmanagerConfig from amConfigs to avoid generating configuration twice
+func findAndDeleteGlobalAlertmanagerConfig(
+	am *monitoringv1.Alertmanager,
+	amConfigs map[string]*monitoringv1alpha1.AlertmanagerConfig,
+) (*monitoringv1alpha1.AlertmanagerConfig, error) {
+	crKey := types.NamespacedName{
+		Namespace: am.Namespace,
+		Name:      am.Spec.GlobalAlertmanagerConfig.Name,
+	}
+	amConfig, ok := amConfigs[crKey.String()]
```
simonpasquier, line 959: The global AlertmanagerConfig object might not be present in the `amConfigs` map. The operator should get it from the API.

sunlintong:
@simonpasquier I updates again at [eab3e45](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/eab3e45b2d7c74d2c53ec6f9391b3a36c3731389) :D
BTW, I add root route continue check, ref:
https://github.com/prometheus/alertmanager/blob/4017d1a478e546909a2b4b4173b6aab3e7dd2dbe/config/config.go#L184-L187

pkg/alertmanager/operator.go
```diff
@@ -920,6 +921,22 @@ receivers:
 	}
 
 	generator := newConfigGenerator(namespacedLogger, version, store)
+	// If defined global AlertmanagerConfig, find the global AlertmanagerConfig object from amConfigs.
+	// Otherwise work with the raw configuration secret.
+	if am.Spec.GlobalAlertmanagerConfig != nil {
+		globalAmConfig, err := c.getGlobalAlertmanagerConfig(am, amConfigs)
+		if err != nil {
+			return errors.Wrap(err, "failed to find and delete global AlertmanagerConfig")
```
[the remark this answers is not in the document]
sunlintong: yes, I think so.

pkg/alertmanager/operator.go
```diff
@@ -920,6 +921,22 @@ receivers:
 	}
 
 	generator := newConfigGenerator(namespacedLogger, version, store)
+	// If defined global AlertmanagerConfig, find the global AlertmanagerConfig object from amConfigs.
+	// Otherwise work with the raw configuration secret.
+	if am.Spec.GlobalAlertmanagerConfig != nil {
```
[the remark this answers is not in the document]
sunlintong: In the current code logic, won't. But I think we hope it will. I'll update the code, thanks for the tip.

pkg/apis/monitoring/v1/types.go
```diff
@@ -1637,6 +1637,20 @@ type AlertmanagerSpec struct {
 	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
 	// +optional
 	MinReadySeconds *uint32 `json:"minReadySeconds,omitempty"`
+	// EXPERIMENTAL: globalAlertmanagerConfig specifies the global Alertmanager configuration.
+	// If defined, it takes precedence over the `configSecret` field.
+	// This field may change in future releases.
+	// The specified global alertmanager config will not force add a namespace label in routes and inhibitRules.
+	GlobalAlertmanagerConfig *GlobalAlertmanagerConfig `json:"globalAlertmanagerConfig,omitempty"`
```
simonpasquier, line 1644: (nit) I don't think we need the `Global` prefix but using `AlertmanagerConfig` could be confusing with the CRD name. `AlertmanagerConfiguration` might be a good middleground.

```suggestion
	alertmanagerConfiguration *AlertmanagerConfiguration `json:"alertmanagerConfiguration,omitempty"`
```

pkg/alertmanager/operator.go
```diff
@@ -920,6 +921,22 @@ receivers:
 	}
 
 	generator := newConfigGenerator(namespacedLogger, version, store)
+	// If defined global AlertmanagerConfig, find the global AlertmanagerConfig object from amConfigs.
+	// Otherwise work with the raw configuration secret.
```
simonpasquier, line 925: We need to get the object directly from the API since we already know its namespace and name (as I commented earlier, the object may not be in the map depending on the selectors).
And rather than deleting it from the map, `selectAlertmanagerConfigs()` can filter it out directly.

pkg/apis/monitoring/v1/types.go
sunlintong: fixed at [903c261](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/903c2610d710f4d5c5ac574abfb0bae021a8db36)

pkg/alertmanager/operator.go
sunlintong: fixed at [903c261](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/903c2610d710f4d5c5ac574abfb0bae021a8db36)

pkg/alertmanager/operator.go
```diff
@@ -920,6 +920,23 @@ receivers:
...
+		if err != nil {
+			return errors.Wrap(err, "get AlertmanagerConfig failed")
+		}
+		baseConfig, err = generator.generateGlobalConfig(ctx, globalAmConfig)
+		if err != nil {
+			return errors.Wrap(err, "failed to generate global AlertmangerConfig")
+		}
+		if err := checkAlertmanagerConfigRootRoute(baseConfig.Route); err != nil {
+			return errors.Wrap(err, "check AlertmanagerConfig root route failed")
+		}
```
simonpasquier, line 937: This should be moved to `generator.generateGlobalConfig()` IMHO.

sunlintong: Yes, fixed at [292dd07](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/292dd07617f2e310908af5aa0cf0fa4edc950c06)

sunlintong:
> @sunlintong thank you very much for the timely updates. I think it's looking very good. Before merging this would deserve an end-to-end test validating the new functionality. Would you be able to add something along the lines of [testUserDefinedAlertmanagerConfig()](https://github.com/prometheus-operator/prometheus-operator/blob/main/test/e2e/alertmanager_test.go#L1277-L1350) but using the new `alertmanagerConfiguration` field instead?

@simonpasquier  That's absolutely no problem. I planed to do this tomorrow. I also want to take this opportunity to learn the e2e test :D

PhilipGough:
@sunlintong thanks again for continuing to see this through. As Simon mentioned, a simple e2e test that tests the happy path will be sufficient to get this one over the line IMO.

I just wanted to add that I built and deployed from your branch and have run through a sanity test on a cluster.

I tested first the creation of an `AlertManager` that references an `AlertManagerConfig` that did not exist. As expected, the StatefulSet never rolls out and we see the operator log the error appropriately.  (@simonpasquier - this is something we can't really protect with the webhook either but might be able to improve the flow when #4575 gets implemented)

I went on to create the following `AlertManagerConfig`.

```yaml
kind: AlertmanagerConfig
metadata:
  name: my-test-config
  namespace: monitoring
spec:
  route:
    groupBy: ['job']
    groupWait: 30s
    groupInterval: 5m
    repeatInterval: 12h
    receiver: 'webhook'
  receivers:
  - name: 'webhook'
    webhookConfigs:
    - url: 'https://eowvcgu3tplymxm.m.pipedream.net'
```

This resulted in the following configuration in alertmanager

```yaml
global:
  resolve_timeout: 5m
  http_config:
    follow_redirects: true
  smtp_hello: localhost
  smtp_require_tls: true
  pagerduty_url: https://events.pagerduty.com/v2/enqueue
  opsgenie_api_url: https://api.opsgenie.com/
  wechat_api_url: https://qyapi.weixin.qq.com/cgi-bin/
  victorops_api_url: https://alert.victorops.com/integrations/generic/20131114/alert/
route:
  receiver: monitoring-my-test-config-webhook
  group_by:
  - job
  continue: false
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 12h
receivers:
- name: monitoring-my-test-config-webhook
  webhook_configs:
  - send_resolved: true
    http_config:
      follow_redirects: true
    url: https://eowvcgu3tplymxm.m.pipedream.net
    max_alerts: 0
templates: []
```

Alert was delivered as expected

alert

sunlintong:
@simonpasquier @PhilipGough I've update the e2e test case at [8c3ca1d](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/8c3ca1d2f308d0607eb96422ce21ff9a0bb06246). And the `testUserDefinedAlertmanagerConfig` test case passed in my local environment.

I used the image built from this branch too. It works well and hope we can use this feature well in the future. If we plan to add more fields in `AlertmanagerConfiguration`, I am very pleasure to contribute 😄 .

simonpasquier:
@sunlintong I think we want to preserve the existing `testUserDefinedAlertmanagerConfig()` function and add another test that verifies specifically the new `alertmanagerConfig` field. I'd recommend the following:
* Rename `testUserDefinedAlertmanagerConfig()` to ``testUserDefinedAlertmanagerConfigFromSecret()`
* Add a new `testUserDefinedAlertmanagerConfigFromCustomResource()` which is your [fc4b8a3](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/fc4b8a309281fe3eec95c6cc4b22d1e9ac9bdac3) commit except that it doesn't need to create the `amConfig` secret.

> If we plan to add more fields in AlertmanagerConfiguration, I am very pleasure to contribute 

sure, support for templates and global configuration would be the next on the list but lets get this one merged first :)

sunlintong:
> Rename testUserDefinedAlertmanagerConfig() to ``testUserDefinedAlertmanagerConfigFromSecret()`
> Add a new testUserDefinedAlertmanagerConfigFromCustomResource() which is your [fc4b8a3](https://github.com/prometheus-operator/prometheus-operator/pull/4220/commits/fc4b8a309281fe3eec95c6cc4b22d1e9ac9bdac3) commit except that it doesn't need to create the amConfig secret.

I can't agree more with this.


> sure, support for templates and global configuration would be the next on the list but lets get this one merged first :)

At my code at https://github.com/sunlintong/prometheus-operator/blob/feature_global_alertmanager_config/pkg/alertmanager/operator.go#L931,
If the raw config secret defined Global or Template, they will be set to null here. Because I didn't support to config these field in AlertmanagerConfiguration.

Should I only assign Route, InhibitRules, Receivers and MuteTimeIntervals to baseConfig ? @simonpasquier

sunlintong:
I update and rebase commits for e2e-test to one commit: b8c092a14004787dd5669b4450732f15448cc3e9 

Reason for update the test case name is to aviod ns name like `allns-x-amuserdefinedalertmanagerconfigfromcustomresource-r7r821-0` be more than 63 characters.

simonpasquier (approved):
LGTM. I'll let @PhilipGough another chance to review it.

PhilipGough (approved):
lgtm

Thanks a lot for seeing this through. Great contribution.

I will leave this open for final review over the weekend and we can merge it Monday pending no remaining issues.

Shashankft9:
hey, thanks for concluding this, appreciate it! 
slightly confused around the usage of global alertmanager config - will it skip the namespace matcher? its not very clear from the updated description of the PR.
Also, thanks for creating an issue for global settings and templates, any plans on adding `enforcedNamespaceLabel` as well, so that namespace matcher can be skipped for all the AlertmanagerConfig CRDs? If yes, we can help with it as well! 

I am taking reference of this issue here: https://github.com/prometheus-operator/prometheus-operator/issues/4033

PR #4230, opened 2021-08-25
Make shard take effect on additionalScrapeConfigs

MegaEle:
## Description
Automatically add shards relabel_configs for additionalScrapeConfigs on Prometheus CRD

## Type of change

## Changelog entry

```release-note
Let the 'shards' also take effect on additionalScrapeConfigs to avoid unexpected repetitive series when using Thanos queries
```


simonpasquier:
I see how it can be useful but at the same time, the operator updating the `additionalScrapeConfigs` field can be unexpected from a user standpoint. Thoughts from @prometheus-operator/prometheus-operator-reviewers welcome :)

PR #4246, opened 2021-09-06
Adds `minReadySeconds` field to AlertManager and Prometheus custom resource spec.

### LINKED ISSUE #3711: Add .spec.minReadySeconds as configurable option
**What is missing?**
.spec.minReadySeconds

**Why do we need it?**
In a HA setup using Cortex or Thanos, I want to have 0 downtime deployments. 
Best-case now is to run 3 replica's of Prometheus. 

Yet if there was an option to rollout slowly between instances, it would be fine to run a replication of 2. 

Current behaviour: 

- Prometheus 1 gets terminated -> starts again
- Loads WAL and eventually (scrape) configs. (I'm not really sure yet if a Pod already has the 'ready' status after reading wal or after reading dynamic scrape configs.)
- The moment the pod is ready, prometheus 2 gets terminated.
- Prometheus 1 has no data yet from "current" period. 
- For example Grafana instantly alerts "no data" as there is no evaluation property on that alert: https://github.com/grafana/grafana/issues/29397

What is possible, is to add a successThreshold like:

```
spec:
  containers:
  - name: prometheus
    readinessProbe:
      successThreshold: 30
      periodSeconds: 5
```
This gives Prometheus time to: 
- Load dynamic config (again, could be wrong here)
- Start scraping

Yet this is not enough. Because components need to know about the now ready state of this prometheus, there is a little delay. It's hard to pinpoint the delay but I would estimate it would be < 0.2 second. Enough to let Grafana at-least throw an alert: no data.

By implementing a minReadySeconds combined with an increased successThreshold, it would be possible for Prometheus to get a proper ready state (from my point of view ;) ). Yet also allowing services to catch up before the other replica gets terminated.

PhilipGough:
Support `minReadySeconds` for the deployed statefulsets and Fixes #3711

```release-note:feature
Adds `minReadySeconds` field to AlertManager and Prometheus custom resource spec.
```

### Verification

I created a `kind` cluster version `1.22.1` (on mac) with the following configuration:

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: kind
featureGates:
  "StatefulSetMinReadySeconds": true
networking:
  ipFamily: ipv6
  apiServerAddress: 127.0.0.1
```

I created an AlertManager with the following CR:

```yaml
kind: Alertmanager
metadata:
  name: example
spec:
  replicas: 3
  minReadySeconds: 10
```

Run the following command:

```shell
kubectl -n test get statefulsets.apps alertmanager-example -o yaml
```

Produces: 
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  annotations:
    prometheus-operator-input-hash: "4777008599300358879"
  generation: 2
  name: alertmanager-example
  namespace: test
  ownerReferences:
  - apiVersion: monitoring.coreos.com/v1
    blockOwnerDeletion: true
    controller: true
    kind: Alertmanager
    name: example
    uid: 40316435-341d-405f-8ada-d7fd2abe6a79
  resourceVersion: "2085"
  uid: f3015b6d-9fd2-4635-8a5b-6b23bdb539e1
spec:
  minReadySeconds: 10
  podManagementPolicy: Parallel
  replicas: 3
  ...
```

I created a `Prometheus` with the following CR:

```yaml
kind: Prometheus
metadata:
  name: example
spec:
  replicas: 2
  minReadySeconds: 5
```

```shell
kubectl -n test get statefulsets.apps prometheus-example -o yaml
```

Produces:
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  annotations:
    prometheus-operator-input-hash: "12556768881199655466"
  generation: 1
  labels:
    operator.prometheus.io/name: example
    operator.prometheus.io/shard: "0"
  name: prometheus-example
  namespace: test
  ownerReferences:
  - apiVersion: monitoring.coreos.com/v1
    blockOwnerDeletion: true
    controller: true
    kind: Prometheus
    name: example
    uid: d225fa16-16b2-4bc6-809a-98d4f53e1cea
  resourceVersion: "3085"
  uid: 0044bb99-e2b2-4409-9e03-e912f1c65a61
spec:
  minReadySeconds: 5
  podManagementPolicy: Parallel
  replicas: 2
  ...
status:
  availableReplicas: 2
  collisionCount: 0
  currentReplicas: 2
  currentRevision: prometheus-example-7bd947d766
  observedGeneration: 1
  readyReplicas: 2
  replicas: 2
  updateRevision: prometheus-example-7bd947d766
  updatedReplicas: 2
```


pkg/prometheus/statefulset_test.go
```diff
@@ -1838,6 +1836,35 @@ func TestExpectedStatefulSetShardNames(t *testing.T) {
...
+		Spec: monitoringv1.PrometheusSpec{
+			MinReadySeconds: &expect,
+		},
+	}, defaultTestConfig, nil, "", 0)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if statefulSet.Spec.MinReadySeconds != 5 {
```
simonpasquier, line 1863:
```suggestion
	if statefulSet.Spec.MinReadySeconds != expect {
```

pkg/prometheus/statefulset_test.go
```diff
@@ -1838,6 +1836,35 @@ func TestExpectedStatefulSetShardNames(t *testing.T) {
...
+		t.Fatalf("expected MinReadySeconds to be 5 but got %d", statefulSet.Spec.MinReadySeconds)
```
simonpasquier, line 1864:
```suggestion
		t.Fatalf("expected MinReadySeconds to be %d but got %d", expect, statefulSet.Spec.MinReadySeconds)
```

example/user-guides/alerting/prometheus-example.yaml
```diff
@@ -4,15 +4,4 @@ metadata:
...
-      name: alertmanager-example
-      port: web
-  serviceMonitorSelector:
-    matchLabels:
-      team: frontend
-  ruleSelector:
-    matchLabels:
-      role: alert-rules
-      prometheus: example
+  minReadySeconds: 400
```
simonpasquier: hmm not sure why this got modified?

pkg/apis/monitoring/v1/types.go
```diff
@@ -401,6 +401,12 @@ type PrometheusSpec struct {
 	// entire scrape will be treated as failed. 0 means no limit.
 	// Only valid in Prometheus versions 2.27.0 and newer.
 	EnforcedLabelValueLengthLimit *uint64 `json:"enforcedLabelValueLengthLimit,omitempty"`
+	// Minimum number of seconds for which a newly created pod should be ready
+	// without any of its container crashing for it to be considered available.
+	// Defaults to 0 (pod will be considered available as soon as it is ready)
+	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
+	// +optional
+	MinReadySeconds *uint32 `json:"minReadySeconds,omitempty"`
```
simonpasquier, line 409: Could it be `uint32`? Given that the default value is 0, I don't think we need a pointer here.

simonpasquier (commented):
I would assume that when the feature gate isn't enabled, the API accepts the request and ignores the minReadySeconds field?

PhilipGough:
> I would assume that when the feature gate isn't enabled, the API accepts the request and ignores the minReadySeconds field?

That was my experience when I tested it locally at least

example/user-guides/alerting/prometheus-example.yaml
PhilipGough: pbkac error

pkg/apis/monitoring/v1/types.go
PhilipGough: It could. Happy to change it. Was just following the couple of fields above and thought being explicit about being optional might be the way but I'm good with changing it.

test/e2e/kind-conf.yaml
```diff
@@ -0,0 +1,9 @@
+# See https://kind.sigs.k8s.io/docs/user/configuration/
+# This allows starting a kind cluster with specific feature gates enabled
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+name: kind
+featureGates:
+  "StatefulSetMinReadySeconds": true
+networking:
+  ipFamily: ipv6
```
PhilipGough, line 9: We need to commit this if we want to run e2e tests that are feature gated but wasn't sure if this is the best directory?

It kind of makes sense here I think because it will serve as a point of reference for someone who wants to run the full e2e suite also?

pkg/apis/monitoring/v1/types.go
simonpasquier: You're talking about the `enforced*` fields? I think that nobody really pay attention initially and we've kept copying the same pattern... I assume that we could switch to non-pointer values without user-facing impact.

PhilipGough: yep, done

simonpasquier:
BTW for consistency ThanosRuler should be updated too.

test/e2e/kind-conf.yaml
simonpasquier: The other option would be under `scripts/` but no strong opinion from my side so lets keep it there.

pkg/apis/monitoring/v1/thanos_types.go
```diff
@@ -185,6 +185,12 @@ type ThanosRulerSpec struct {
 	// of all alerts.
 	// Maps to the '--alert.query-url' CLI arg.
 	AlertQueryURL string `json:"alertQueryUrl,omitempty"`
+	// Minimum number of seconds for which a newly created pod should be ready
+	// without any of its container crashing for it to be considered available.
+	// Defaults to 0 (pod will be considered available as soon as it is ready)
+	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
+	// +optional
+	MinReadySeconds uint32 `json:"minReadySeconds,omitempty"`
```
sthaha, line 193: I wonder if it is better to use `*uint32`. In this particular case it shouldn't matter too much since the default is 0 but if we are to ever change the default to say `5` then it will help with determining if the value was explicitly set by the user to `0` or if it was left empty (nil).

simonpasquier: On second thought, I agree with you. Sorry @PhilipGough for the back and forth :)

PhilipGough: no worries

simonpasquier:
hmm interesting the `PromAlertmanagerDiscovery` test fails because the check doesn't work for IPv6 addresses...

```
2021-09-07T12:42:49.3493742Z 2021/09/07 12:42:49 Expected Alertmanager Targets: []string{"http://fd00:10:244::8c:9093/api/v2/alerts", "http://fd00:10:244::8d:9093/api/v2/alerts", "http://fd00:10:244::8e:9093/api/v2/alerts"}
2021-09-07T12:42:49.3495291Z 2021/09/07 12:42:49 Existing Alertmanager Targets: []string{"http://[fd00:10:244::8c]:9093/api/v2/alerts", "http://[fd00:10:244::8d]:9093/api/v2/alerts", "http://[fd00:10:244::8e]:9093/api/v2/alerts"}
```

simonpasquier (approved):
lgtm

PR #4247, opened 2021-09-06
*: make log messages more consistent

simonpasquier:
## Description

When skipping fields in the generation of Alertmanager and Prometheus
configuration because they aren't supported by the current version,
ensure that we log the name and namespace of the resource as well as the
minimum required version and the current version.

This change also reduce code duplication in the generation of the
Prometheus configuration when enforcing sample, target and label limit.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -131,8 +131,8 @@ func addSafeAuthorizationToYaml(
 		// extract current cfg section from assetStoreKey, assuming
 		// "<component>/something..."
 		component := strings.Split(assetStoreKey, "/")[0]
-		level.Warn(logger).Log("msg", fmt.Sprintf("%s: found authorization section, but prometheus is < 2.26.0, ignoring", component),
-			component)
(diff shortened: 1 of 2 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -1838,3 +1752,55 @@ func (cg *ConfigGenerator) generateRemoteWriteConfig(
...
+	labelValueLengthLimitKey = limitKey{
+		specField:       "labelValueLengthLimit",
+		prometheusField: "label_value_length_limit",
+		minVersion:      "2.27.0",
+	}
+)
+
+type limitEnforcer struct {
+	logger  log.Logger
+	version semver.Version
```
fpetkovski, line 1791: nit: since this is a field of the `limitEnforcer`, it might be better to rename this to prometheusVersion to indicate that it's not the enforcer version.

fpetkovski (approved):
just one small nit, otherwise lgtm

PR #4260, opened 2021-09-13
alertmanager: enable empty basic auth username

mahumpula:
## Description

Some services apparently don't have username in auth basic authentication. Enable specifying empty username so the webhook receiver can work with them.

Sample usecase: sending monitoring alert heartbeat to OpsGenie via webhook.

* [OpsGenie authentication](https://docs.opsgenie.com/docs/authentication)

## Type of change

- [x ] `ENHANCEMENT` (non-breaking change which improves existing functionality)

## Changelog entry

```release-note
- Allow empty username for HTTP Auth Basic
```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -814,7 +814,7 @@ func (cg *configGenerator) convertHTTPConfig(ctx context.Context, in monitoringv
 			return nil, errors.Errorf("failed to get BasicAuth password key %q from secret %q", in.BasicAuth.Password.Key, in.BasicAuth.Password.Name)
 		}
 
-		if username != "" && password != "" {
+		if username != "" || password != "" {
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
lgtm

PhilipGough:
@mahumpula this change will also allow a username without a password, is that acceptable and desired? Since it is basic auth I would have assumed we always need a password even if a username is not provided.

mahumpula:
@PhilipGough you are right. I just could not come up with some reason, why someone else would not get idea about using only username and not password. So if you think it's best to keep it more strict, I will rewrite the condition.

PhilipGough:
@mahumpula - I don't think there is any real risk or anything with the existing change so it's probably fine. On the other hand, simply changing the statement to  `if password != ""` will likely give you what you need :)

mahumpula:
@PhilipGough, @simonpasquier what can I do to get this MR merged?

PR #4270, opened 2021-09-20
Add a probe instruction

GuoJiafeng:
## Description

I haven't found any documents about how to use probe, and I can't even find useful documents on the Internet. The purpose of using probe at present is: I have an etcd independent of k8s cluster and need to use SSL certificate. Podmonitor and servicemonitor can't read the target, and additionalsnapeconfigs can't set external SSL certificate (I didn't find a way to mount external certificates). Finally, probe saved me, so I added a document for this scenario, hoping to help more people

## Type of change

Add a probe instruction

### DIFF
--- a/Documentation/user-guides/probe-instructions.md
+++ b/Documentation/user-guides/probe-instructions.md
@@ -1,0 +1,78 @@
+## Probe
+
+The `Probe` custom resource definition (CRD) allows to declarative define how groups of ingresses and static targets should be monitored. Besides the target, the `Probe` object requires a `prober` which is the service that monitors the target and provides metrics for Prometheus to scrape. The following document describes one of the application scenarios.
+
+
+
+## Usage scenario
+
++ HTTPS certificate needs to be configured
++ Systems outside the k8s cluster, such as etcd
++ Static target
+
+## Example
+
++ Prepare SSL certificate file, execute the following command and generate related resources
+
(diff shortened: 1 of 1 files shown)


simonpasquier:
> additionalsnapeconfigs can't set external SSL certificate (I didn't find a way to mount external certificates)

Have you tried mounting additional secrets in the Prometheus pod using the `secrets` field of [`PrometheusSpec`](https://prometheus-operator.dev/docs/operator/api/#prometheusspec)? 
While using Probes like you did works technically, it isn't exactly the use case that is targeted.

uweswrtz:
The [design document()](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/design.md#probe) mentions the 'Probe' (together with the blackbox exporter). What's then their purpose?

For my use case it worked.

But I also agree - Pawel's 3 options look and seem more k8s style.

GuoJiafeng:
I don't know why I didn't find this 『secrets』 configuration for a long time. Now my problem has been solved, but it's not through the normal way

GuoJiafeng:
I don't know why I didn't find this configuration before. This configuration was added four years ago. I shouldn't have been unable to find it. It's been stirred up for a long time. Now my problem has been solved, but it's not through the normal way. Thank you and thank you for your patience

PR #4272, opened 2021-09-21
Fix enforce namespace label

simonpasquier:
## Description

Whenever `enforcedNamespaceLabel` is defined, the operator should add a final relabeling configuration that sets the enforced label to the namespace's value. It was effective for the `relabel_configs` field but not for `metric_relabeling_configs`. The assumption was that dropping metric relabeling entries that tried to alter the enforced label was enough but it appears that the `labelmap` action could be used to bypass the enforcement.

## Type of change

## Changelog entry

```release-note
Ensure that the namespace label is always enforced with metricRelabelConfigs.
```

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -3784,6 +3784,148 @@ func testPromMinReadySeconds(t *testing.T) {
 	}
 }
 
+// testPromEnforcedNamespaceLabel checks that the enforcedNamespaceLabel field
+// is honored even if a user tries to bypass the enforcement.
+func testPromEnforcedNamespaceLabel(t *testing.T) {
+	t.Parallel()
+
+	for i, tc := range []struct {
+		relabelConfigs       []*monitoringv1.RelabelConfig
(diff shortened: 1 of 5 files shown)


test/e2e/prometheus_test.go
```diff
@@ -3784,6 +3784,155 @@ func testPromMinReadySeconds(t *testing.T) {
...
+func testPromEnforcedNamespaceLabel(t *testing.T) {
+	t.Parallel()
+
+	for i, tc := range []struct {
+		relabelConfigs       []*monitoringv1.RelabelConfig
+		metricRelabelConfigs []*monitoringv1.RelabelConfig
+	}{
+		{
+			// drop label with relabel_configs.
+			metricRelabelConfigs: []*monitoringv1.RelabelConfig{
```
PhilipGough, line 3798: should this one be `relabelConfigs`?

simonpasquier: nice catch!

PhilipGough (approved):
/lgtm

PR #4275, opened 2021-09-22
Support enforced body_size_limit at Prometheus CRD Level

slashpai:
## Description

A new feature was added to [Prometheus v2.28.0](https://github.com/prometheus/prometheus/blob/main/CHANGELOG.md#2280--2021-06-21) which adds the ability to the size of the body that a scrape can have. This is currently experimental upstream but can be really useful for resiliency purposes. The exact goal of the feature is to add a safety net to prevent Prometheus from killing a node because of a malicious target by capping the maximum ingestion with `body_size_limit` based on the size of the cluster.

Added this as enforced limit for Prometheus since service monitor owners may not know how to set bodySizeLimit and they already have sampleLimit if needed

## Type of change

## Changelog entry

```release-note
Support EnforcedBodySizeLimit configuration field at the Prometheus CRD level
```


pkg/prometheus/promcfg_test.go
```diff
@@ -5596,3 +5596,241 @@ alerting:
...
+		expected              string
+	}{
+		{
+			version:               "v2.27.0",
+			enforcedBodySizeLimit: -1,
+			expected:              expectNoLimit,
+		},
+		{
+			version:               "v2.28.0",
+			enforcedBodySizeLimit: -1,
```
dgrisonnet, line 5758: Isn't 0 supposed to be the value for no limit?

pkg/prometheus/promcfg.go
```diff
@@ -1802,5 +1816,9 @@ func (l *limitEnforcer) addLimitsToYAML(cfg yaml.MapSlice, k limitKey, limit uin
 		return cfg
 	}
 
-	return append(cfg, yaml.MapItem{Key: k.prometheusField, Value: getLimit(limit, enforcedLimit)})
+	// Since BodySizeLimit is defined only in PrometheusCRD
+	if k.specField != "enforcedBodySizeLimit" {
```
simonpasquier, line 1820: do we need a special case? Since `limit` is always 0 for `enforcedBodySizeLimit`, I would assume that `getLimit(limit, enforcedLimit)` would return `enforcedLimit` if it's > 0. Anyway it would make sense to have the comment on the caller side explaining why we pass a zero value.

simonpasquier (commented):
Nice

pkg/prometheus/promcfg.go
dgrisonnet, line 1820: we don't need a special case for this scenario since `getLimit` will handle the fact that the BodySizeLimit is only defined at the PrometheusCRD level.

pkg/prometheus/promcfg.go
```diff
@@ -726,6 +730,7 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	cfg = enforcer.addLimitsToYAML(cfg, labelLimitKey, m.Spec.LabelLimit, enforcedLabelLimit)
 	cfg = enforcer.addLimitsToYAML(cfg, labelNameLengthLimitKey, m.Spec.LabelNameLengthLimit, enforcedLabelNameLengthLimit)
 	cfg = enforcer.addLimitsToYAML(cfg, labelValueLengthLimitKey, m.Spec.LabelValueLengthLimit, enforcedLabelValueLengthLimit)
+	cfg = enforcer.addLimitsToYAML(cfg, bodySizeLimitKey, 0, enforcedBodySizeLimit)
```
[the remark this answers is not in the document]
dgrisonnet: For consistency purposes, I think it would make sense, but it is still very unlikely that users will define this limit per ServiceMonitor since it is a very advanced configuration option.

pkg/prometheus/promcfg_test.go
slashpai: This is testing a negative case that it doesn't set this limit passed since its invalid

pkg/prometheus/promcfg.go
slashpai: Ya it was my bad I read the conditional in getLimits() wrong. I updated the pR

pkg/prometheus/promcfg_test.go
slashpai: We don't need to set if default is already set by prometheus right?

pkg/prometheus/promcfg.go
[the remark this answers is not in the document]
slashpai: ya I felt that we won't be using it much even if we provide option in monitors and probe. But I am open to add it in them if we need to

pkg/apis/monitoring/v1/types.go
```diff
@@ -401,6 +401,12 @@ type PrometheusSpec struct {
 	// entire scrape will be treated as failed. 0 means no limit.
 	// Only valid in Prometheus versions 2.27.0 and newer.
 	EnforcedLabelValueLengthLimit *uint64 `json:"enforcedLabelValueLengthLimit,omitempty"`
+	// An uncompressed response body larger than this many bytes will cause the
+	// scrape to fail. 0 means no limit. Example: 100MB.
+	// This is an experimental feature, this behaviour could
+	// change or be removed in the future.
+	// Only valid in Prometheus versions 2.28.0 and newer.
```
simonpasquier, line 408:
```suggestion
	// EnforcedBodySizeLimit defines the maximum size of uncompressed response body 
	// that will be accepted by Prometheus. Targets responding with a body larger than this many bytes
	// will cause the scrape to fail. Example: 100MB.
	// If defined, the limit will apply to all service/pod monitors and probes.
	// This is an experimental feature, this behaviour could
	// change or be removed in the future.
	// Only valid in Prometheus versions 2.28.0 and newer.
```

pkg/prometheus/promcfg.go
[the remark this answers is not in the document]
simonpasquier: IMHO it's fine to wait and see if there is demand for adding it to service/pod monitors and probes since it's a new feature. And the main use case is to avoid Prometheus from blowing up when a target starts exposing hundreds megabytes of data by accident.

slashpai:
Addressed review comments, please review again :)

pkg/prometheus/promcfg.go
```diff
@@ -727,6 +731,16 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	cfg = enforcer.addLimitsToYAML(cfg, labelNameLengthLimitKey, m.Spec.LabelNameLengthLimit, enforcedLabelNameLengthLimit)
 	cfg = enforcer.addLimitsToYAML(cfg, labelValueLengthLimitKey, m.Spec.LabelValueLengthLimit, enforcedLabelValueLengthLimit)
 
+	// Since BodySizeLimit is defined only in PrometheusCRD
+	if enforcedBodySizeLimit != "" {
+		if version.Major == 2 && version.Minor >= 28 {
+			cfg = append(cfg, yaml.MapItem{Key: "body_size_limit", Value: enforcedBodySizeLimit})
+		} else {
+			level.Warn(cg.logger).Log("msg", "PodMonitor 'body_size_limit' is only available from prometheus 2.28.",
+				"podMonitor", m.Name)
```
simonpasquier, line 740: `logger` is already decorated with `podMonitor` and `namespace` keys. 

```suggestion
			level.Warn(logger).Log("msg", "'body_size_limit' is only available from prometheus 2.28.",
				"version", version)
```

pkg/prometheus/promcfg.go
```diff
@@ -727,6 +731,16 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	cfg = enforcer.addLimitsToYAML(cfg, labelNameLengthLimitKey, m.Spec.LabelNameLengthLimit, enforcedLabelNameLengthLimit)
 	cfg = enforcer.addLimitsToYAML(cfg, labelValueLengthLimitKey, m.Spec.LabelValueLengthLimit, enforcedLabelValueLengthLimit)
 
+	// Since BodySizeLimit is defined only in PrometheusCRD
+	if enforcedBodySizeLimit != "" {
+		if version.Major == 2 && version.Minor >= 28 {
```
simonpasquier, line 736: (nit) more consistent and robust (e.g. it will continue to work with v3.x if it still supports the field)
```suggestion
		if version.GTE(semver.MustParse("2.28.0")) {
```

pkg/prometheus/promcfg_test.go
```diff
@@ -5596,3 +5596,231 @@ alerting:
...
+    action: keep
+  body_size_limit: %s
+alerting:
+  alert_relabel_configs:
+  - action: labeldrop
+    regex: prometheus_replica
+  alertmanagers: []
+`
+
+	for _, tc := range []struct {
```
simonpasquier, line 5746: could you add a check with an empty `enforcedBodySizeLimit`?

simonpasquier:
lgtm

pkg/prometheus/promcfg.go
```diff
@@ -1804,3 +1819,16 @@ func (l *limitEnforcer) addLimitsToYAML(cfg yaml.MapSlice, k limitKey, limit uin
...
+		return cfg
+	}
+
+	if l.prometheusVersion.LT(semver.MustParse("2.28.0")) {
+		level.Warn(l.logger).Log("msg", "body_size_limit is only available starting from prometheus 2.28.0",
+			"version", l.prometheusVersion)
+		return cfg
+	}
+	return append(cfg, yaml.MapItem{Key: "body_size_limit", Value: enforcedLimit})
+}
```
PhilipGough, line 1834: @simonpasquier I'm not as familiar with this part of the code, but I wonder should this take on a similar pattern that we implemented in AlertManager config where we sanitized post config generation?

If it is applicable we could take care of it in a different PR

slashpai: It won't add this field if version doesn't have minimum required version. So we don't have to remove the fields if unsupported

simonpasquier: The Prometheus controller takes another approach: instead of duplicating the Prometheus config struct, it builds a `yaml.MapSlice` struct dynamically. Honestly I'm not sure which approach is the less terrible :)

PhilipGough: @simonpasquier  me neither, although I do see a lot of version checks scattered throughout the file so maybe consolidating them to a single function is preferable. 

@slashpai not a blocker here, was just curious if we should align the approach

slashpai:
rebase done @simonpasquier

PhilipGough (approved):
lgtm

simonpasquier (approved):

PR #4284, opened 2021-09-27
*: fix action and script publishing images

simonpasquier:
The workflow didn't work properly when tagging releases because it
wasn't enabled for tag events. We exclude all tags that start by `pkg`
since they are only relevant for Go sub-modules.

## Type of change

## Changelog entry

```release-note

```


fpetkovski:
lgtm

Issue #4282, opened 2021-09-24
Ensure that duration fields are valid


simonpasquier:
**What is missing?**
The operator doesn't prevent invalid values for duration fields like `PrometheusSpec.Retention` (same is true for `EnforcedBodySizeLimit`).

**Why do we need it?**
Avoid breaking Prometheus :)

**Environment**

* Prometheus Operator version: N/A

**Anything else we need to know?**:

slashpai:
I had thought about that while creating PR but didn't get enough time back then. I will fix that

simonpasquier:
Cool! FWIW the AlertmanagerConfig CRD also has duration fields like `groupInterval`, `groupWait` and `repeatInterval`. The check should happen when the operator selects the resources and rejects the ones that are invalid.

simonpasquier:
The same validation should be run on
* `ScrapeInterval`, `ScrapeTimeout`, `EvaluationInterval`, `MetadataConfig.SendInterval`, `QuerySpec.Timeout`, `Thanos.ReadyTimeout`, `RemoteWriteSpec.RemoteTimeout`, `RemoteReadSpec.RemoteTimeout` and `AlertmanagerEndpoints.Timeout` for Prometheus resources.
* `Interval` and `ScrapeTimeout` for service/pod monitors and probes.
* `ClusterGossipInterval`, `ClusterPushpullInterval` and `ClusterPeerTimeout` for Alertmanager.
* `EvaluationInterval` for ThanosRuler.

PhilipGough:
@simonpasquier is this closed by #4308

slashpai:
@PhilipGough We will be able to close once https://github.com/prometheus-operator/prometheus-operator/pull/4352 is merged as well, that was last one part of this fix :)

PR #4285, opened 2021-09-27
pkg/prometheus/promcfg*.go: Validate EnforcedBodySizeLimit

slashpai:
## Description

Added the input validation for EnforcedBodySizeLimit

Related to #4282
## Type of change

## Changelog entry

```release-note
Added the input validation for EnforcedBodySizeLimit
```

cc: @simonpasquier


pkg/prometheus/promcfg.go
```diff
@@ -1804,10 +1804,22 @@ func (l *limitEnforcer) addBodySizeLimitsToYAML(cfg yaml.MapSlice, enforcedLimit
 		return cfg
 	}
 
+	if !validInput("^(0[K,M,G,T,P,E]*B$|[1-9]\\d+[K,M,G,T,P,E]*B$)", enforcedLimit) {
```
fpetkovski, line 1807: Should we use the same library as prometheus [1][2] instead of crafting our own regex? 

[1] https://github.com/prometheus/prometheus/blob/e582e907f8a97ffae27123b9632a254e5018a4a6/config/config.go#L400
[2] https://pkg.go.dev/github.com/alecthomas/units

simonpasquier: Yep using the same library makes sense.

slashpai: I initially created code using units but I dropped that since other fields also needed to be validated even though in this PR only bodysizelimit is validated. So thought to have a general validator method.

simonpasquier:
I think that the operator should be more explicit and fail the reconciliation if the provided parameter is invalid.

pkg/prometheus/promcfg.go
slashpai: https://github.com/prometheus-operator/prometheus-operator/compare/master...slashpai:input_validator#diff-453ce9c721c51bf9591c2b212e9ce50157b9c40d1b211177bf502be268099656R1808-R1816

fpetkovski: I think a general validator can be useful as a fallback when more specific ones are not available. Regular expressions can often be hard to verify and troubleshoot, so I would keep things simple if possible.

slashpai: Sure that makes sense, I will change back the validation using units

slashpai:
> I think that the operator should be more explicit and fail the reconciliation if the provided parameter is invalid.

I propagated the error from validation that should fail sync

slashpai:
@fpetkovski @simonpasquier Can you review again?

pkg/prometheus/promcfg_test.go
```diff
@@ -5856,9 +5882,6 @@ alerting:
 				nil,
 				nil,
 			)
-			if err != nil {
```
fpetkovski, line 5859: Maybe we want to refine this assertion like 
```
expectErr := expected == ""
if expectErr && err == nil &&  {
  t.Fatal("prometheus config generation should return an error")
}

if !expectErr && err != nil {
  t.Fatal("prometheus config generation should not return an error")
}
```
Alternatively, we could also add a `expectErr` field to the test cases and use it instead of relying on the value of `expected`

slashpai: @fpetkovski Thank you :) . I updated the test

fpetkovski:
Nice work, this LGMT!

simonpasquier (commented):
IMHO the check should happen near the top of `*ConfigGenerator.GenerateConfig()` since it's a PrometheusSpec field?

https://github.com/prometheus-operator/prometheus-operator/blob/c6b5cc2e178002683ab3afe8d023a8bb2508ba19/pkg/prometheus/promcfg.go#L219

slashpai:
> IMHO the check should happen near the top of `*ConfigGenerator.GenerateConfig()` since it's a PrometheusSpec field?
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/c6b5cc2e178002683ab3afe8d023a8bb2508ba19/pkg/prometheus/promcfg.go#L219

Good point. I added a func for validating the fields at prometheus crd level. Added the func call in operator.go before calling GenerateConfig

@simonpasquier Can you take a look?

pkg/prometheus/operator.go
```diff
@@ -1592,6 +1592,13 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *mon
 		return errors.Wrap(err, "loading additional alert manager configs from Secret failed")
 	}
 
+	// Validate Prometheus Config Inputs at Prometheus CRD level
+	err = c.configGenerator.ValidateConfigInputs(p)
```
simonpasquier, line 1596: I'm not sure we need to create another receiver method for `*ConfigGenerator`. IMHO the validation should be done by `GenerateConfig()` directly so the caller doesn't have to worry about validating th configuration before generating.

pkg/prometheus/promcfg.go
```diff
@@ -215,6 +216,31 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
 	return stringMapToMapSlice(m)
 }
 
+// ValidateConfigInputs validates if inputs passed are valid
+func (cg *ConfigGenerator) ValidateConfigInputs(p *v1.Prometheus) error {
+	logger := log.With(cg.logger, "Validating config inputs", "EnforcedBodySizeLimit")
```
simonpasquier, line 221: Since the logger isn't used directly in this func, no need to create a new contextual logger. And the "key" parameter should be snake case instead of a full sentence.

pkg/prometheus/promcfg.go
```diff
@@ -215,6 +216,31 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
...
+
+// Validate EnforcedBodySizeLimit
+func validateBodySizeLimit(enforcedLimit string, logger log.Logger) error {
+	// To validate if given value is parsable for the acceptable body_size_limit values
+	_, err := units.ParseBase2Bytes(enforcedLimit)
+
+	if err != nil {
+		level.Warn(logger).Log("msg", "body_size_limit provided is invalid it should be a format of size",
+			"enforcedBodySizeLimit", enforcedLimit)
+		return errors.Wrap(err, "Invalid body_size_limit value specified")
```
simonpasquier, line 239: In general you should either return an error or log a message but not doing both.
See "Only handle errors once" in https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully. And the rest of the post is only worth reading ;-)

```suggestion
		return errors.Wrap(err, "invalid enforcedBodySizeLimit value specified")
```

simonpasquier (commented):
Can you base your change against release-0.51 so it can be part of v0.51.2

pkg/prometheus/operator.go
slashpai: Did you mean to call the ValidateInput() func from inside GenerateConfig() method?

simonpasquier: yes and we don't need to make `ValidateConfigInputs()` public (e.g. `validateConfigInputs()` instead).

slashpai: thank you made it public since I was trying to call from operator.go. I am updating with changes

slashpai:
@simonpasquier I updated with changed and changed the base for PR

slashpai:
> @simonpasquier I updated with changed and changed the base for PR
just realized I didn't update error string, updating soon

slashpai:
Done

slashpai:
@simonpasquier Take a look when you get chance

pkg/prometheus/promcfg.go
```diff
@@ -227,6 +250,13 @@ func (cg *ConfigGenerator) GenerateConfig(
 	additionalAlertManagerConfigs []byte,
 	ruleConfigMapNames []string,
 ) ([]byte, error) {
+	// Validate Prometheus Config Inputs at Prometheus CRD level
+	err := validateConfigInputs(p)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid config value passed")
+	}
```
simonpasquier, line 258: I think we can avoid wrapping the error, the returned error should already be clear enough.

```suggestion
	if err := validateConfigInputs(p); err != nil {
		return nil, err
	}
```

pkg/prometheus/promcfg.go
```diff
@@ -215,6 +216,28 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
...
+	if p.Spec.EnforcedBodySizeLimit != "" {
+		err := validateBodySizeLimit(p.Spec.EnforcedBodySizeLimit)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Validate EnforcedBodySizeLimit
```
simonpasquier, line 230:
```suggestion
```

pkg/prometheus/promcfg.go
```diff
@@ -215,6 +216,28 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
 	return stringMapToMapSlice(m)
 }
 
+// ValidateConfigInputs validates if inputs passed are valid
```
simonpasquier, line 219:
```suggestion
// validateConfigInputs runs extra validation on the Prometheus fields which can't be done at the CRD schema validation level.
```

pkg/prometheus/promcfg.go
```diff
@@ -215,6 +216,28 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
...
+func validateConfigInputs(p *v1.Prometheus) error {
+	if p.Spec.EnforcedBodySizeLimit != "" {
+		err := validateBodySizeLimit(p.Spec.EnforcedBodySizeLimit)
+		if err != nil {
+			return err
+		}
```
simonpasquier, line 225:
```suggestion
		if err := validateBodySizeLimit(p.Spec.EnforcedBodySizeLimit); err != nil {
			return err
		}
```

pkg/prometheus/promcfg.go
```diff
@@ -215,6 +216,28 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
...
+}
+
+// Validate EnforcedBodySizeLimit
+func validateBodySizeLimit(enforcedLimit string) error {
+	// To validate if given value is parsable for the acceptable body_size_limit values
+	_, err := units.ParseBase2Bytes(enforcedLimit)
+
+	if err != nil {
+		return errors.Wrap(err, "invalid enforcedBodySizeLimit value specified")
+	}
```
simonpasquier, line 237:
```suggestion
	if _, err := units.ParseBase2Bytes(enforcedLimit); err != nil {
		return errors.Wrap(err, "invalid enforcedBodySizeLimit value specified")
	}
```

slashpai:
@simonpasquier Thank you for the review. I updated the PR. Hope its good now :)

simonpasquier (approved):
lgtm

slashpai:
@simonpasquier Curious to know would this change be synced with master only before release of 0.51.2?

simonpasquier:
@slashpai yes this is the process.

PR #4307, opened 2021-10-05
alertmanager: Support updated matching syntax in raw config for inhib…

PhilipGough:
…it rules

See https://prometheus.io/docs/alerting/latest/configuration/#matcher

This change allows the raw config to parse the updated matching syntax
and will allow us to pave the way to move away from the deprecated
target_match, target_match_re, source_match, source_match_re fields.

## Description

## Type of change

## Changelog entry

```release-note
Support updated matching syntax in raw config for `inhibit_rules`
```


pkg/alertmanager/amcfg.go
```diff
@@ -1017,3 +1020,37 @@ func (whc *webhookConfig) sanitize(amVersion semver.Version, logger log.Logger)
...
+	if usingDeprecatedSyntax := checkNotEmptyMap(ir.SourceMatch, ir.TargetMatch, ir.SourceMatchRE, ir.TargetMatchRE); usingDeprecatedSyntax {
+		msg := "inhibit rule is using a deprecated match syntax which will be removed in future versions"
+		level.Warn(logger).Log("msg", msg)
+	}
+
+	if !matchersV2Allowed && checkNotEmptySlice(ir.SourceMatchers, ir.TargetMatchers) {
+		ir.SourceMatchers = nil
+		ir.TargetMatchers = nil
+		msg := `invalid syntax in 'inhibit_rules', target_matchers and source_matchers matching is supported in AlertManager >= 0.22.0 only - dropping config`
+		level.Warn(logger).Log("msg", msg)
```
simonpasquier, line 1036: It might be better to either drop the inhibition rule or fail completely since it might not do what you expect (e.g. it may inhibit more alerts than needed).

pkg/alertmanager/amcfg.go
```diff
@@ -1017,3 +1020,37 @@ func (whc *webhookConfig) sanitize(amVersion semver.Version, logger log.Logger)
 func (wcc *weChatConfig) sanitize(amVersion semver.Version, logger log.Logger) {
 	wcc.HTTPConfig.sanitize(amVersion, logger)
 }
+
+func (ir *inhibitRule) sanitize(amVersion semver.Version, logger log.Logger) {
+	matchersV2Allowed := amVersion.GTE(semver.MustParse("0.22.0"))
+
+	if usingDeprecatedSyntax := checkNotEmptyMap(ir.SourceMatch, ir.TargetMatch, ir.SourceMatchRE, ir.TargetMatchRE); usingDeprecatedSyntax {
```
simonpasquier, line 1027:
```suggestion
	if usingDeprecatedSyntax := checkNotEmptyMap(ir.SourceMatch, ir.TargetMatch, ir.SourceMatchRE, ir.TargetMatchRE); usingDeprecatedSyntax && matchersV2Allowed {
```

simonpasquier (commented):
Do you plan for another PR for route matchers?

pkg/alertmanager/amcfg.go
PhilipGough: for the other sanitize functions we just log and drop the config..happy to proceed with that or would it be preferable to error out in this case?

PR #4309, opened 2021-10-06
alertmanager: Support updated matching syntax in raw config for route

philipgough:
## Description

See https://prometheus.io/docs/alerting/latest/configuration/#matcher

This change allows the raw config to parse the updated matching syntax
and will allow us to pave the way to move away from the deprecated `match`, `match_re` fields.

Closes #4174 ?

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [x] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Support updated matching syntax in raw config for `route`
```

PR #4332, opened 2021-10-13
Support updated matching syntax for routes in AlertManagerConfig CRD

PhilipGough:
## Description

Depends on #4309 
Related to #4087 

## Type of change

## Changelog entry

```release-note
Support updated matching syntax for routes in AlertManagerConfig CRD
```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -137,7 +137,7 @@ func (cg *configGenerator) generateConfig(
 			continue
 		}
 
-		subRoutes = append(subRoutes, convertRoute(amConfigs[amConfigIdentifier].Spec.Route, crKey, true))
+		subRoutes = append(subRoutes, cg.convertRoute(amConfigs[amConfigIdentifier].Spec.Route, crKey, true))
 
 		for _, receiver := range amConfigs[amConfigIdentifier].Spec.Receivers {
 			receivers, err := cg.convertReceiver(ctx, &receiver, crKey)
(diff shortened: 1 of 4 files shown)


pkg/alertmanager/amcfg.go
```diff
@@ -161,37 +161,50 @@ func (cg *configGenerator) generateConfig(
...
 		if matcher.Regex {
 			matchRE[matcher.Name] = matcher.Value
 		} else {
 			match[matcher.Name] = matcher.Value
 		}
 	}
 	if firstLevelRoute {
-		match["namespace"] = crKey.Namespace
+		if matchersV2Allowed {
+			if !contains(v2NamespaceMatcher, matchers) {
```
simonpasquier, line 199: side-note: we could even not check if the matcher is already present and always append it.

pkg/alertmanager/amcfg.go
```diff
@@ -161,37 +161,50 @@ func (cg *configGenerator) generateConfig(
...
+				matchers = append(matchers, v2NamespaceMatcher)
+			}
+		} else {
+			match["namespace"] = crKey.Namespace
+		}
 		delete(matchRE, "namespace")
```
simonpasquier, line 205: as for the previous comment, there's no harm to leave a regex matcher on namespace...

simonpasquier (commented):
Don't we need to extend `checkAlertmanagerConfig()` to drop AlertmanagerConfig resources that use the new matchers syntax with old Alertmanager versions?

https://github.com/prometheus-operator/prometheus-operator/blob/6dfe2c320963277d3e6aa9da25d2a347a776535f/pkg/alertmanager/operator.go#L1043-L1056

simonpasquier (approved):
lgtm

PR #4352, opened 2021-10-29
Add validations for duration fields in alertmanager and thanos

slashpai:
## Description

Add validations for duration fields in alertmanager and thanos

## Type of change

## Changelog entry

```release-note
Added input validations for duration fields in alertmanager and thanos
```

cc: @simonpasquier


paulfantom (approved):
LGTM.

Side note: it looks like we should put those in validating webhook to prevent incorrect configuration from being applied. 

@prometheus-operator/kube-prometheus-reviewers WDYT about moving validation logic into webhook (or pkg) and bundling webhook definition in `bundle.yaml`?

PhilipGough (approved):
/lgtm

Maybe we can take care of @paulfantom suggestion in a follow up PR?

slashpai:
@paulfantom 

I can work on validating webhook if we decide so :)

But do you think we can merge this so it can go in 0.52 release this week? :)

simonpasquier:
I think that more validating webhooks make sense.

PR #4373, opened 2021-11-07
Split prometheus e2e tests

fpetkovski:
## Description

The current Prometheus e2e tests have both sequential and parallel
subtests. The parallel subtests expect a prometheus-operator configured
to run against all namespaces. These tests run before the sequential ones
and block their execution.

This commit splits the two subtests into dedicated workflows so that
they can run in parallel. The outcome is a reduction of the e2e suite
to from 60m to 40m.

## Type of change

## Changelog entry

```release-note

```


test/e2e/prometheus_test.go
```diff
@@ -305,6 +305,7 @@ func createK8sAppMonitoring(
 }
 
 func testPromRemoteWriteWithTLS(t *testing.T) {
+	t.Parallel()
```
fpetkovski, line 308: This test alone takes around 10-12 minutes to run. It has 20 subtests each of which takes around one minute. We should see if we there are redundant subtests that we can remove, or look for a way to optimize them.

simonpasquier (approved):
Neat! LGTM

test/e2e/prometheus_test.go
simonpasquier: I checked quickly a while ago but my recollection was that all combinations made sense to be tested.

fpetkovski: sgtm, I'll dive deeper into that test once we merge this PR. Just waiting for @paulfantom to chime as well

paulfantom (approved):
LGTM, let's just squash commits on merge.

PR #4374, opened 2021-11-07
added support for alertmanager sns receiver

### LINKED ISSUE #4236: Alertmanager SNS receiver not working
**What happened?**

We want to use the new SNS receiver from alertmanager 0.23.0, but the operator fails with an unmashal error. A quick look at the code showed that the sns receiver structs are missing.

A general question: why is the alert manager config rebuilt if the corresponding structures are exported in the alert manager module and there is already a dependency on the alert manager anyway?

alertmanager config with sns receivers should work without errors.

1. Deploy prometheus operator 0.50.0
2. Deploy alertmanager with sns receiver via the kube-prometheus-stack helm chart (config is stored in a secret)

**Environment**

    `0.50.0`
    
    

* Manifests:

```
...
receivers:
- name: 'null'
- name: 'cluster-admin'
   sns_configs:
   - send_resolved: True
     topic_arn: arn:aws:sns:eu-central-1:123456789:my-alert-topic
...
```

* Prometheus Operator Logs:

```
level=error ts=2021-08-30T11:17:30.620996291Z caller=klog.go:116 component=k8s_client_runtime func=ErrorDepth msg="Sync \"monitoring/kube-prometheus-stack-alertmanager\" failed: provision alertmanager configuration: base config from Secret could not be parsed: yaml: unmarshal errors:\n  line 6: field sns_configs not found in type alertmanager.receiver\n  line 11: field sns_configs not found in type alertmanager.receiver"
```

The operator and the alert manager is deployed via the kube prometheus stack helm chart

phin1x:
## Description

this pr adds support for the new sns receiver which was released with alertmanager 0.23.0.

fixes #4236

## Type of change

## Changelog entry

```release-note
Add support for alertmanager sns receiver
```


simonpasquier:
Thanks for the PR! Can you add validation for the SNS configurations like we have for the other [integrations](https://github.com/prometheus-operator/prometheus-operator/blob/fe746efdf6ee4fa483590af7e8366acc1daf8871/pkg/alertmanager/validation.go#L47-L78)?

pkg/alertmanager/types.go
```diff
@@ -309,3 +310,24 @@ type victorOpsConfig struct {
...
+	TargetARN     string            `yaml:"target_arn,omitempty" json:"target_arn,omitempty"`
+	Subject       string            `yaml:"subject,omitempty" json:"subject,omitempty"`
+	Message       string            `yaml:"message,omitempty" json:"message,omitempty"`
+	Attributes    map[string]string `yaml:"attributes,omitempty" json:"attributes,omitempty"`
+}
+
+type sigV4Config struct {
+	Region    string        `yaml:"region,omitempty" json:"region,omitempty"`
+	AccessKey string        `yaml:"access_key,omitempty" json:"access_key,omitempty"`
+	SecretKey config.Secret `yaml:"secret_key,omitempty" json:"secret_key,omitempty"`
```
PhilipGough, line 330: I think this needs to be a string as it will be essentially hidden when marshalled.

PhilipGough (changes requested):
Thanks for the contribution. I've left a comment that will need resolving for this to work as intended. A unit test will be good to prove the end result as well as the required validation func.

phin1x:
i'll have into the unit test and the validation function.

PhilipGough:
hey @phin1x just wanted to see if you were able to progress this or you need some help with it?

phin1x:
Hi @PhilipGough i didn't had time time to work on this pr. I hope i find some time at the end of next week.

pkg/alertmanager/types.go
phin1x: i changed the type to string

PhilipGough:
@phin1x - thanks for following up. What I would like to do here is split this PR into two.

First, we want to support the raw sns receiver via raw config (as you have done [here](https://github.com/prometheus-operator/prometheus-operator/pull/4374/files#diff-deb09cd4cf9ffbd98a5d6a4502ac0fd7f8764215f0d349c630909fe53f3156c3R186)) which I feel is nearly there (perhaps missing a small unit test).

Secondly, we will want to support it via the CRD (which you have started [here](https://github.com/prometheus-operator/prometheus-operator/pull/4374/files#diff-ca1bb1a59e5c40fa6d7fe5a21bba214234558c9ab699f3de0fdc95d7b302dc35R725)) but that needs a small bit of work. For example we shouldnt pass in secrets directly but use `SecretKetSelector` like https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L186

Does that approach make sense?

phin1x:
@PhilipGough this is fine to me. The unit test for the raw config should be in https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/alertmanager/amcfg_test.go?

pkg/alertmanager/validation.go
```diff
@@ -81,11 +81,77 @@ func validateReceivers(receivers []monitoringv1alpha1.Receiver) (map[string]stru
...
+			return errors.New("either 'phone_number', 'target_arn' or 'topic_arn' must be set")
+		}
+
+		if config.PhoneNumber != "" {
+			if len(config.PhoneNumber) > 16 || config.PhoneNumber[0] != '+' {
+				return errors.New("'phone_number' must start with a plus sign followed by a maximum of 15 digits")
+			}
+			if _, err := strconv.Atoi(config.PhoneNumber[1:]); err != nil {
+				return errors.New("'phone_number' must start with a plus sign followed by a maximum of 15 digits")
+			}
```
simonpasquier, line 105: IIUC the checks aren't present in Alertmanager. Also since the phone number string can be a template, they may prevent valid configurations.

pkg/alertmanager/validation.go
```diff
@@ -81,11 +81,77 @@ func validateReceivers(receivers []monitoringv1alpha1.Receiver) (map[string]stru
...
+		}
+
+		if config.TargetARN != "" {
+			if !strings.HasPrefix(config.TopicARN, "arn:") {
+				return errors.New("'target_arn' must start with 'arn:'")
+			}
+			if err := isAwsArn(config.TopicARN); err != nil {
+				return err
+			}
+		}
```
simonpasquier, line 124: same here

pkg/alertmanager/validation.go
PhilipGough: yes, you are correct, sorry I missed that.

pkg/alertmanager/validation.go
```diff
@@ -81,11 +81,77 @@ func validateReceivers(receivers []monitoringv1alpha1.Receiver) (map[string]stru
...
+	// aws arn contains six sections delimited by a ':'
+	if strings.Count(arn, ":") < 5 {
+		return errors.New("aws arn does not contain enough sections")
+	}
+	return nil
+}
+
+func validateSigV4Config(config *monitoringv1alpha1.SigV4Config) error {
+	if (config.AccessKey != "" && config.SecretKey == "") || (config.AccessKey == "" && config.SecretKey != "") {
+		return errors.New("both 'access_key' and 'secret_key' must be set")
```
PhilipGough, line 150: I think we can say both of these secrets must exist since since I don't believe we have a mechanism for injecting environment variables into alertmanager?

PhilipGough: `if config.AccessKey  == "" || config.SecretKey == ""`

pkg/alertmanager/validation.go
```diff
@@ -81,11 +81,77 @@ func validateReceivers(receivers []monitoringv1alpha1.Receiver) (map[string]stru
...
+		if config.TargetARN != "" {
+			if !strings.HasPrefix(config.TopicARN, "arn:") {
+				return errors.New("'target_arn' must start with 'arn:'")
+			}
+			if err := isAwsArn(config.TopicARN); err != nil {
+				return err
+			}
+		}
+
+		if config.APIURL != "" {
```
PhilipGough, line 126: this one is also a templated string

phin1x:
@PhilipGough Thanks for completing my work. i will close this pr and open a new one for the crd.

PR #4380, opened 2021-11-09
add a flag to enable local prometheus compaction when sidecar upload …

yeya24:
…is configured

Signed-off-by: Ben Ye <ben.ye@bytedance.com>

## Description

TSDB compaction is useful for improving Prometheus Query Performance. In the case of some edge sites with sidecar and Prometheus, we want to enable TSDB compaction for better query performance as well as uploading data to the object storage. For more detail, please check out https://github.com/thanos-io/thanos/issues/2620.

This pr provides a flag to enable local TSDB compaction for Prometheus for the use case described above.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -752,9 +752,8 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 				},
 			)
 
-			// NOTE(bwplotka): As described in https://thanos.io/components/sidecar.md/ we have to turn off compaction of Prometheus
-			// to avoid races during upload, if the uploads are configured.
-			disableCompaction = true
+			// By default, Prometheus local TSDB compaction is disabled when uploads are configured.
+			disableCompaction = !p.Spec.Thanos.EnableLocalCompaction
 		}
 
 		if p.Spec.Thanos.TracingConfig != nil || len(p.Spec.Thanos.TracingConfigFile) > 0 {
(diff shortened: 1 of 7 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -752,9 +752,8 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 				},
 			)
 
-			// NOTE(bwplotka): As described in https://thanos.io/components/sidecar.md/ we have to turn off compaction of Prometheus
```
fpetkovski, line 755: Should we add this warning to the documentation as well?

yeya24: Which documentation do you mean? The race condition happens rarely in some edge cases.

fpetkovski: I meant if we should move this comment to the documentation for the newly added field so that people are aware of the potential problems that can arise when the feature is enabled. Even if it could happen rarely in some cases, I think there's value in making that obvious for users.

simonpasquier (commented):
I guess we need @bwplotka to weigh in whether it's actually safe to enable compaction with the Thanos sidecar :)
Also having 2 ways to enable/disable compaction in the CRD feels error-prone and confusing, e.g.

```
spec:
  disableCompaction: true
  thanos:
    enableLocalCompaction: true
```

yeya24:
Yeah. Having two ways to control this seems wrong. Do you have any suggestions?
Using the one in Prometheus CRD is confusing for me as well

Kampe:
any resolution on this?  When shipping to thanos receive (instead of using the sidecare) should we still disable compaction all together on the shipping prometheus?

simonpasquier:
> When remote-writing to thanos receive (instead of using the sidecar) should we still disable compaction all together on the shipping prometheus?

My understanding is that compactions have no consequence on the remote write scenario.

pkg/prometheus/statefulset.go
yeya24: Yes, that makes sense. I will add the comment to the document and let users know. Users can take their own risks and make decision.

Migueljfs:
Any news on this? Anything was decided meanwhile? @yeya24 @simonpasquier 

I would like to use compaction as I want prometheus to be as fast as possible

PR #4383, opened 2021-11-09
test: Reduce runtime of e2e tests

fpetkovski:
## Description

This commit reduces the runtime of Prometheus e2e tests from 40 minutes
to less than 30 minutes. It does so by implementing two changes:
* reducing the scraping time for the tlsRemoteWrite test from 30s to 5s
* increasing the parallelism for e2e tests from 2 to 4.

According to the docs[1], the Github runners have 2 CPUs and our tests
do not seem to be CPU bound. Since each Github Actions job runs on a
dedicated VM, doubling the parallelism should not be problematic for now.

[1] https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

## Type of change

## Changelog entry

```release-note

```


paulfantom (approved):
lgtm

CI failures are unrelated and non-blocking.

PR #4387, opened 2021-11-09
Restore Prometheus StatefulSet liveness probe

yann-soubeyrand:
## Description

This PR modifies the Prometheus StatefulSet to restore the liveness probe (follow-up of https://github.com/prometheus-operator/prometheus-operator/pull/3502 and https://github.com/prometheus-operator/prometheus-operator/pull/4369).

## Type of change

(I’m not sure if this is only a bugfix.)

## Changelog entry

```release-note
Restore Prometheus StatefulSet liveness probe so that deadlocks are detected and recovered from.
```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -631,8 +631,13 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 		FailureThreshold: 60,
 	}
 
-	// TODO(paulfantom): Re-add livenessProbe when kubernetes 1.21 is available.
-	// This would be a follow-up to https://github.com/prometheus-operator/prometheus-operator/pull/3502
+	livenessProbe := &v1.Probe{
+		Handler:          probeHandler("/-/healthy"),
+		TimeoutSeconds:   probeTimeoutSeconds,
+		PeriodSeconds:    5,
+		FailureThreshold: 6,
+	}
+
 	readinessProbe := &v1.Probe{
 		Handler:          probeHandler("/-/ready"),
 		TimeoutSeconds:   probeTimeoutSeconds,
@@ -864,6 +869,7 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 			Args:                     promArgs,
 			VolumeMounts:             promVolumeMounts,
 			StartupProbe:             startupProbe,
+			LivenessProbe:            livenessProbe,
 			ReadinessProbe:           readinessProbe,
 			Resources:                p.Spec.Resources,
 			TerminationMessagePolicy: v1.TerminationMessageFallbackToLogsOnError,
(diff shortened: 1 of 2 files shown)


PhilipGough:
@yann-soubeyrand can you try rebase this to see if it rectifies the failing tests as the test may have been executed atop a flake (https://github.com/prometheus-operator/prometheus-operator/pull/4390).

PhilipGough:
Also we might want to revisit the changes in #4369 based on https://github.com/prometheus-operator/prometheus-operator/pull/4369#discussion_r756258595.

yann-soubeyrand:
> Based on that discussion, my biggest concern is that we may have introduced a regression by changing the FailureThreshold of the readiness probe from 120 -> 3

I don’t think so, the readiness probe will only get more reactive but should not cause harm since it’s only effect will be to remove the pod from the service. What do you think?

PhilipGough:
@yann-soubeyrand - sorry, I had to re-review and re-check the ready implementation on the Prometheus side. The change to the readiness probe is ok I think.

Having a dig around last night (from various sources):
> Startup probes are useful for Pods that have containers that take a long time to come into service. Rather than set a long liveness interval, you can configure a separate configuration for probing the container as it starts up, allowing a time longer than the liveness interval would allow.

> Sometimes, you have to deal with legacy applications that might require an additional startup time on their first initialisation. In such cases, it can be tricky to set up liveness probe parameters without compromising the fast response to deadlocks that motivated such a probe. The trick is to set up a startup probe with the same command, HTTP or TCP check, with a failureThreshold * periodSeconds long enough to cover the worse case startup time.

> Liveness probes do not wait for readiness probes to succeed. If you want to wait before executing a liveness probe you should use initialDelaySeconds or a startupProbe.

> Startup probes are similar to readiness probes but only executed at startup. They are optimised for slow starting containers or applications with unpredictable initialisation processes.

I put us in the "unpredictable initialisation process" category and reading those makes me sway towards @simonpasquier's suggestion of switching startup probe to the `ready` endpoint and for the worst case scenario. Since we know neither liveness or readiness comes into effect until the startup probe passes, wdyt?

yann-soubeyrand:
@PhilipGough I followed your and @simonpasquier’s suggestion and reworked my PR to use `/-/ready` for the startup probe ;-)

simonpasquier:
thanks! I'd prefer to have the handler change for the startup probe in a separate PR (it will easier to track it down for future ourselves) so I've submitted https://github.com/prometheus-operator/prometheus-operator/pull/4433.

simonpasquier (approved):
lgtm

leaving @paulfantom the final word :)

PhilipGough:
Just my final 2 cents. I think this is ok now since we merged #4433 

1. We know that liveness and readiness probes are ineffective until startup probe completes
2. We know the startup probe will not succeed unless the TSDB initialisation is complete

At that point, we expect the container to be ready to serve and we know that health endpoint should be succeed and cause no flakiness with pod restarts.

So lgtm

yann-soubeyrand:
Should I try to rebase and force push to see if the tests pass (IIRC, the failing tests are flaky, right?)?

PhilipGough (approved):
lgtm

PR #4406, opened 2021-11-16
test:e2e:Prometheus: Fix flake in check for minReadySeconds

### LINKED ISSUE #4401: PromMinReadySeconds e2e test is flaky
I've seen the test fail occasionally because the modification of the minReadySeconds field hasn't been propagated to the statefulset. Maybe the update didn't happen yet and the test should retry? 

```
2021-11-09T14:28:31.2359914Z === CONT  TestAllNS/y/PromMinReadySeconds
2021-11-09T14:28:53.4706077Z     prometheus_test.go:3754: expected MinReadySeconds to be 10 but got 5
```

https://github.com/prometheus-operator/prometheus-operator/blob/6dfe2c320963277d3e6aa9da25d2a347a776535f/test/e2e/prometheus_test.go#L3742-L3755

cc @PhilipGough

PhilipGough:
## Description

Fixes #4401 

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -3740,18 +3740,27 @@ func testPromMinReadySeconds(t *testing.T) {
 	}
 
 	var updated uint32 = 10
+	var got int32
 	prom.Spec.MinReadySeconds = &updated
 	if _, err = framework.UpdatePrometheusAndWaitUntilReady(context.Background(), ns, prom); err != nil {
 		t.Fatal("Updating prometheus failed: ", err)
 	}
 
-	promSS, err = kubeClient.AppsV1().StatefulSets(ns).Get(context.Background(), "prometheus-basic-prometheus", metav1.GetOptions{})
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
+		promSS, err := kubeClient.AppsV1().StatefulSets(ns).Get(context.Background(), "prometheus-basic-prometheus", metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
 
-	if promSS.Spec.MinReadySeconds != int32(updated) {
-		t.Fatalf("expected MinReadySeconds to be %d but got %d", updated, promSS.Spec.MinReadySeconds)
+		if promSS.Spec.MinReadySeconds != int32(updated) {
+			got = promSS.Spec.MinReadySeconds
+			return false, nil
+		}
+		return true, nil
+	})
+
+	if err != nil {
+		t.Fatalf("expected MinReadySeconds to be %d but got %d", updated, got)
 	}
 }
 


test/e2e/prometheus_test.go
```diff
@@ -3730,13 +3730,22 @@ func testPromMinReadySeconds(t *testing.T) {
...
 
-	promSS, err := kubeClient.AppsV1().StatefulSets(ns).Get(context.Background(), "prometheus-basic-prometheus", metav1.GetOptions{})
-	if err != nil {
-		t.Fatal(err)
-	}
+	var got int32
+	err = wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
+		promSS, err := kubeClient.AppsV1().StatefulSets(ns).Get(context.Background(), "prometheus-basic-prometheus", metav1.GetOptions{})
+		if err != nil {
+			return false, err
```
fpetkovski, line 3737: I think this that returning an error will short-circuit the wait.Poll function and won't really reduce the flake. What we want is 
```
if err != nil {
  return false, nil
}
```

PhilipGough: Hey @fpetkovski - I'm not entirely sure I agree since from https://github.com/prometheus-operator/prometheus-operator/pull/4406/files#diff-b3d3450b51af206bb0009c2b9c2f863c3d03a3575d64cbdb18510a2bc834144aR3728 we expect the statefulset to exist which is why I kept the hard error. I don't *think* that is what was causing the flake, it was because the spec hadn't yet been updated which the retry logic should solve.

Our other tests don't retry on error for the `Get Resource` calls so I would not expect this to behave or be handled any differently.

That said, I don't think changing it to retry on error is a big issue so I can do it if you prefer. Just telling you my thought process :)

fpetkovski: I see, I was under the impression that the flake was caused because we didn't wait long enough. In that case, it's correct to do what you've done and return an error if the sts is missing :+1:

fpetkovski (approved):
Since these sorts of flakes are hard to debug, I'm fine with merging this and monitoring the CI for a bit.

test/e2e/prometheus_test.go
```diff
@@ -3730,13 +3730,22 @@ func testPromMinReadySeconds(t *testing.T) {
 		t.Fatal("Creating prometheus failed: ", err)
 	}
 
-	promSS, err := kubeClient.AppsV1().StatefulSets(ns).Get(context.Background(), "prometheus-basic-prometheus", metav1.GetOptions{})
-	if err != nil {
-		t.Fatal(err)
-	}
+	var got int32
+	err = wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
```
simonpasquier, line 3734: I don't think that a wait.Poll is needed here since the sts should be created with the expected value.

PhilipGough: good point

simonpasquier (approved):
approve

PR #4413, opened 2021-11-17
Ensure group_by values are unique in alertmanager config

### LINKED ISSUE #4402: AlertmanagerConfig with duplicate group_by labels generates an invalid configuration
The following AlertmanagerConfig definition would trigger an invalid configuration:

```yaml
kind: AlertmanagerConfig
metadata:
  name: config-example
  labels:
    alertmanagerConfig: example
spec:
  route:
    groupBy: ['job', 'job']
    groupWait: 30s
    groupInterval: 5m
    repeatInterval: 12h
    receiver: 'null'
  receivers:
  - name: 'null'
```

Alertmanager logs:

```
level=error ts=2021-11-15T15:42:06.617Z caller=coordinator.go:118 component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/config/alertmanager.yaml err="duplicated label \"job\" in group_by"
```

cc @PhilipGough

PhilipGough:
## Description

Elements in the routes `groupBy` list must be unique and furthermore, when the special value '...' is provided, there must be no other elements in the list.

Fixes #4402 

## Type of change

## Changelog entry

```release-note
Ensure group_by values are unique in alertmanager config
```

### DIFF
--- a/pkg/alertmanager/validation.go
+++ b/pkg/alertmanager/validation.go
@@ -218,6 +218,19 @@ func validateAlertManagerRoutes(r *monitoringv1alpha1.Route, receivers map[strin
 		return errors.Errorf("receiver %q not found", r.Receiver)
 	}
 
+	if groupLen := len(r.GroupBy); groupLen > 0 {
+		groupedBy := make(map[string]struct{}, groupLen)
+		for _, str := range r.GroupBy {
+			if _, found := groupedBy[str]; found {
+				return errors.Errorf("duplicate values not permitted in route 'groupBy': %v", r.GroupBy)
+			}
+			groupedBy[str] = struct{}{}
+		}
+		if _, found := groupedBy["..."]; found && groupLen > 1 {
+			return errors.Errorf("'...' must be a sole value in route 'groupBy': %v", r.GroupBy)
+		}
+	}
+
 	children, err := r.ChildRoutes()
 	if err != nil {
 		return err
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -1532,7 +1532,7 @@ Route defines a node in the routing tree.
 | Field | Description | Scheme | Required |
 | ----- | ----------- | ------ | -------- |
 | receiver | Name of the receiver for this route. If not empty, it should be listed in the `receivers` field. | string | true |
-| groupBy | List of labels to group by. | []string | false |
+| groupBy | List of labels to group by. Labels must not be repeated (unique list). Special label \"...\" (aggregate by all possible labels), if provided, must be the only element in the list. | []string | false |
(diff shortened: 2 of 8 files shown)


pkg/alertmanager/validation.go
```diff
@@ -218,6 +218,19 @@ func validateAlertManagerRoutes(r *monitoringv1alpha1.Route, receivers map[strin
 		return errors.Errorf("receiver %q not found", r.Receiver)
 	}
 
+	if groupLen := len(r.GroupBy); groupLen > 0 {
```
fpetkovski, line 221: For my understanding, what is the reason for using both the OpenAPI validation and implementing our own operator-side validation?

PhilipGough: From my understanding, there are at least a couple of reasons:

1. There are some cases that the OpenAPi validation (at least with the kubebuilder annotations) do not cover. This PR actually contains an example of one of those. We want unique items in the slice (covered with `UniqueItems`), but we also want to validate that if `'...'` is in the list, there are no other items, so we need to validate ourselves server-side (unless I've missed something).
2. We implement a validating webhook, so it made sense to split out the entire validation and have the admission controller call that.

PhilipGough: The error message from the failing test is as follows: `alertmanager_instance_namespaces_test.go:136: initialize AlertmanagerConfig CRD: create CRD alertmanagerconfigs on the apiserver: create CRD: AlertmanagerConfig: CustomResourceDefinition.apiextensions.k8s.io "alertmanagerconfigs.monitoring.coreos.com" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[route].properties[groupBy].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic`

I went digging around and found https://github.com/kubernetes/community/pull/708/ and https://github.com/kubernetes/kubernetes/commit/fd09c3dbb67374e4f4b46b820b5cd3d9a3124cda#diff-6d0d1a9c59e6d17eea2b3ccdf2d6eb7d6169efdf7ed5d5e1f5c09ce5632efd18R205 which basically means we cannot have the `UniqueItems:=true`, so again a custom validation func is required. 

I am surprised this is not supported for basic types like int and string but here we are.

fpetkovski: Alright, this explains it then. Thanks for the input.

fpetkovski (approved):

pkg/alertmanager/validation.go
sttts: Make it list-type set and it will be checked for duplicates.

PR #4416, opened 2021-11-19
pkg/prometheus: refactor ConfigGenerator

simonpasquier:
## Description

This change makes ConfigGenerator aware of the Prometheus version so
that we don't need to pass this information around.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -43,17 +43,34 @@ var (
 	invalidLabelCharRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
 )
 
-// ConfigGenerator is used to create Prometheus configurations from operator resources.
+// ConfigGenerator knows how to generate a Prometheus configuration which is
+// compatible with a given Prometheus version.
 type ConfigGenerator struct {
-	logger log.Logger
+	logger  log.Logger
+	version semver.Version
 }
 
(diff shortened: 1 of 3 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -43,17 +43,34 @@ var (
...
-		logger: logger,
+// NewConfigGenerator creates a ConfigGenerator for the provided Prometheus resource.
+func NewConfigGenerator(logger log.Logger, p *v1.Prometheus) (*ConfigGenerator, error) {
+	promVersion := operator.StringValOrDefault(p.Spec.Version, operator.DefaultPrometheusVersion)
+	version, err := semver.ParseTolerant(promVersion)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse prometheus version")
+	}
+
+	logger = log.With(logger, "version", promVersion)
```
PhilipGough, line 61: would the logger benefit from having the name of `p` attached here?

simonpasquier: The logger already has a `key` field inherited from `sync()` which is formatted as `namespace/name`.

https://github.com/prometheus-operator/prometheus-operator/blob/1c913200060b0dbfb2cdb41bbc4956b1437f5678/pkg/prometheus/operator.go#L1220

fpetkovski (approved):

Issue #3989, opened 2021-04-16
Prometheus Agent Support


kakkoyun:
Prometheus soon is going to have a lightweight version as an agent. https://github.com/prometheus/prometheus/issues/8305
It would be awesome to support this in the promethues-operator. 

I've opened the issue to start the discussion and know what the maintainers think?

@prometheus-operator/kube-prometheus-reviewers

P.S: I'd be happy to work on this.

simonpasquier:
+1 for me definitely. Because the agent won't have all capabilities of the Prometheus server (e.g. no rule evaluation,no alertmanager config, no API probably) and the plan is to have it as another binary, it might make sense to have another CRD for it. Obviously we need more details on what the agent will actually look like.

brancz:
I think it might be a little early as the details haven't been fleshed/implemented out on Prometheus side even. It doesn't hurt to think about a high-level strategy of the CRD though.

I realize that this might be overloading things a bit, but this is sort of an opportunity for a new start, where we could re-think our mistakes from the existing Prometheus CRD/API. If you have time, I would very much welcome it if we could think about the ScrapeConfig CRD as the basis of all configuration, and PodMonitors/ServiceMonitors/Probes only compile down to the ScrapeConfig CRD and all the PrometheusAgent CRD does is pick up ScrapeConfig CRDs directly. Rough thoughts here: https://github.com/prometheus-operator/prometheus-operator/issues/2787

ArthurSens:
This issue came up during the contributor office hours, so just backporting the discussion here:

It’s already possible to run the agent once it's released with the Prometheus CR so it's not like we have a huge urgency on this, but a dedicated CR for it would definitely help with validation and a better UX.

* The idea for the new Agent CR is to re-use everything we can from the Prometheus CR. 
* Creating a new Agent CR will not prevent people from using Prometheus CR with the agent configuration, and we should not block this from happening. The main reason is because avoiding maintaining extra logic for little benefit

simonpasquier:
Now that we have a beta release of Prometheus with the agent feature flag ([v2.32.0-beta.0](https://github.com/prometheus/prometheus/releases/tag/v2.32.0-beta.0)), I gave it a try using the existing Prometheus CRD.

TL;DR: it's not working out of the box but it shouldn't be too hard to adjust a bit the operator to "support" the agent mode with the existing Prometheus CR. 

My CR looks like this:

```
apiVersion: monitoring.coreos.com/v1
[...]

PR #4417, opened 2021-11-19
pkg/prometheus: Make CR compatible for agent mode

ArthurSens:
## Description

Prometheus in agent mode fail-fast if alerting or rules_files is present in its configuration.
This commit removes those sections from the config if not necessary, aiming to make it possible to run Prometheus in agent mode with the Prometheus Custom Resouce.

I'll leave this PR as a draft until there is consensus if we really want to do this, or if we prefer to only support agent mode with a dedicated PrometheusAgent Custom Resource.

Changes were made based on this comment: https://github.com/prometheus-operator/prometheus-operator/issues/3989#issuecomment-974137486

## Type of change

## Changelog entry

```release-note
Add possibility to run Prometheus in agent mode with the Prometheus Custom Resource
```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -359,14 +359,16 @@ func (cg *ConfigGenerator) GenerateConfig(
 
 	cfg = append(cfg, yaml.MapItem{Key: "global", Value: globalItems})
 
-	ruleFilePaths := []string{}
-	for _, name := range ruleConfigMapNames {
-		ruleFilePaths = append(ruleFilePaths, rulesDir+"/"+name+"/*.yaml")
+	if p.Spec.RuleSelector != nil {
+		ruleFilePaths := []string{}
+		for _, name := range ruleConfigMapNames {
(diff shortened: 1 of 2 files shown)


simonpasquier:
Thanks for tackling this! I submitted https://github.com/prometheus-operator/prometheus-operator/pull/4416 because I wanted to clean up a bit the current code before proceeding with a similar change but I'm happy to merge your PR first and follow up with mine :)

simonpasquier:
lgtm

> I'll leave this PR as a draft until there is consensus if we really want to do this, or if we prefer to only support agent mode with a dedicated PrometheusAgent Custom Resource.

Definitely +1 on unblocking users that want to run Prometheus in agent mode now because I feel that having a PrometheusAgent CRD would take time. Also the change is only an "optimization" of the generated config and it doesn't change the behavior for the rest of our users.

paulfantom (approved):

ArthurSens:
Alright, if you both already agreed with enabling this with the Prometheus CR then I'm removing the draft mode.

I'll let you folks move forward with the merge because this is literally my first PR to prometheus-operator repository, I'm not sure about the approval process works here 😅

PR #4418, opened 2021-11-20
Makefile: use /usr/bin/env bash instead of /bin/bash

flokli:
This allows invoking `make generate` from systems where bash is not in
/bin/bash (such as NixOS).

Instead, this will pick up `bash` from $PATH.

See https://www.reddit.com/r/linuxadmin/comments/975nok/binbash_vs_usrbinenv_bash/
## Type of change

## Changelog entry

```release-note
The `Makefile` will now pick up `bash` from `$PATH`, instead of hardcoding it to /bin/bash. This will fix running it on systems where bash is elsewhere.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-SHELL=/bin/bash -o pipefail
+SHELL=/usr/bin/env bash -o pipefail
 
 GOOS?=$(shell go env GOOS)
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
lgtm

simonpasquier:
thanks!

PR #4422, opened 2021-11-22
pkg/k8sutil: fix annotation propagation

### LINKED ISSUE #4392: spec.podMetadata.annotations changes not applied starting with v0.50.0
**What happened?**

Starting with v0.50.0 when the `Prometheus` resource has an updated value for `spec.podMetadata.annotations` the changes are not applied to the StatefulSet. Prior versions updated as expected.

Expected the StatefulSet to have the new annotations applied to the Pods.

1. Create a Prometheus resource without a value for `spec.podMetadata.annotations`
2. Add a value to `spec.podMetadata.annotations`

and

1. Create a Prometheus resource with a value for `spec.podMetadata.annotations`
2. Modify the value or remove the value from `spec.podMetadata.annotations`

**Environment**

0.50.0, 0.51.1, 0.51.2, v0.52.0

Shouldn't be relevant but I tested with 1.19.x & 1.20.x clusters

I believe this was introduced via https://github.com/prometheus-operator/prometheus-operator/pull/4185

fpetkovski:
## Description

A bug was introduced in 0.50.0 where annotations set in the Prometheus
podMetadata spec were not properly propagated to statefulset pods.

This commit fixes the bug and adds additional unit tests to verify the
functionality works as expected.

Fixes https://github.com/prometheus-operator/prometheus-operator/issues/4392

## Type of change

## Changelog entry

```release-note
Propagate annotations set in spec.podMetadata to the statefulset pods. Fixes #4392.
```

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -168,7 +168,7 @@ func UpdateStatefulSet(ctx context.Context, sstClient clientappsv1.StatefulSetIn
 
 		mergeMetadata(&sset.ObjectMeta, existingSset.ObjectMeta)
 		// Propagate annotations set by kubectl on spec.template.annotations. e.g performing a rolling restart.
-		mergeKubectlAnnotations(&sset.Spec.Template.ObjectMeta, existingSset.Spec.Template.ObjectMeta)
+		mergeKubectlAnnotations(&existingSset.Spec.Template.ObjectMeta, sset.Spec.Template.ObjectMeta)
 
 		_, err = sstClient.Update(ctx, sset, metav1.UpdateOptions{})
 		return err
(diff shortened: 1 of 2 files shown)


pkg/k8sutil/k8sutil.go
```diff
@@ -254,12 +254,16 @@ func mergeMaps(new map[string]string, old map[string]string) map[string]string {
 }
 
 func mergeKubectlAnnotations(new *metav1.ObjectMeta, old metav1.ObjectMeta) {
-	new.SetAnnotations(mergeMapsByPrefix(new.Annotations, old.Annotations, "kubectl.kubernetes.io/"))
+	new.SetAnnotations(mergeMapsByPrefix(old.Annotations, new.Annotations, "kubectl.kubernetes.io/"))
```
simonpasquier, line 257: The `old` and `new` parameters are mixed up here compared to the function declaration and it's a bit confusing.

fpetkovski: Yes, that could be confusing. I swapped them here instead: https://github.com/prometheus-operator/prometheus-operator/pull/4422/files#diff-0128909c1627157f3265add5bfeacace324a45ac2aa7c81bf9e91e887edf67faR171.

We could also rename `old` and `new` into `from` and `to` to make it more clear what is being merged into what.

simonpasquier (approved):
lgtm

fpetkovski:
Hm it looks like I might have introduced a regression, let me take another look

fpetkovski:
Seems to have been a flake. 

@paulfantom any other comments or should we merge it?

paulfantom (approved):
lgtm

PR #4433, opened 2021-12-02
pkg/prometheus: use /-/ready for the startup probe

simonpasquier:
## Description

The /-/ready handler returns OK only after the TSDB initialization has
completed. The WAL replay can take a significant time for large setups
hence we enable the startup probe with a generous failure threshold (15
minutes) to ensure that the readiness probe only comes into effect once
Prometheus is effectively ready. We don't want to use the /-/healthy
handler here because it returns OK as soon as the web server is started
(irrespective of the WAL replay).

## Type of change

## Changelog entry

```release-note
use /-/ready endpoint for the startup probe
```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -617,8 +617,15 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 		return handler
 	}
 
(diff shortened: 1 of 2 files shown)


fpetkovski (approved):

PhilipGough (approved):
lgtm

PR #4434, opened 2021-12-02
alertmanager: Sync validation with upstream alertmanager

PhilipGough:
## Description

The changes here close the gap between the upstream alertmanager validations (code, docs) and our internal representation of those types.

Changes are made with the intention of reducing the chances
of accepting an invalid CR representation.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -428,6 +428,9 @@ func (cg *configGenerator) convertSlackConfig(ctx context.Context, in monitoring
 			return nil, errors.Errorf("failed to get key %q from secret %q", in.APIURL.Key, in.APIURL.Name)
 		}
 		out.APIURL = strings.TrimSpace(url)
+		if _, err := ValidateURL(out.APIURL); err != nil {
+			return nil, errors.Wrapf(err, "invalid 'apiURL' %s in slack config", out.APIURL)
+		}
 	}
 
 	var actions []slackAction
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -41,7 +41,12 @@ import (
 
 const inhibitRuleNamespaceKey = "namespace"
 
-func loadCfg(s string) (*alertmanagerConfig, error) {
+// alertmanagerConfigFrom returns a valid global alertmanagerConfig from s
+// or returns an error if
+// 1. s fails validation provided by upstream
+// 2. s fails to unmarshal into internal type
+// 3. the unmarshalled output is invalid
+func alertmanagerConfigFrom(s string) (*alertmanagerConfig, error) {
 	// Run upstream Load function to get any validation checks that it runs.
 	_, err := config.Load(s)
 	if err != nil {
@@ -54,6 +59,22 @@ func loadCfg(s string) (*alertmanagerConfig, error) {
 		return nil, err
 	}
 
+	rootRoute := cfg.Route
+	if rootRoute == nil {
+		return nil, errors.New("root route must exist")
+	}
+
+	if rootRoute.Receiver == "" {
+		return nil, errors.New("root route's receiver must exist")
+	}
+
+	if len(rootRoute.Matchers) > 0 || len(rootRoute.Match) > 0 || len(rootRoute.MatchRE) > 0 {
+		return nil, errors.New("'matchers' not permitted on root route")
+	}
+
+	if len(rootRoute.MuteTimeIntervals) > 0 {
+		return nil, errors.New("'mute_time_intervals' not permitted on root route")
+	}
 	return cfg, nil
 }
 
@@ -393,14 +414,15 @@ func (cg *configGenerator) convertWebhookConfig(ctx context.Context, in monitori
 	if in.URLSecret != nil {
 		url, err := cg.getValidURLFromSecret(ctx, crKey.Namespace, *in.URLSecret)
 		if err != nil {
-			return nil, errors.Wrap(err, "convertWebhookConfig failed")
+			return nil, err
 		}
 		out.URL = url
 	} else if in.URL != nil {
-		if _, err := ValidateURL(*in.URL); err != nil {
-			return nil, errors.Wrap(err, "convertWebhookConfig failed")
+		url, err := ValidateURL(*in.URL)
+		if err != nil {
+			return nil, err
 		}
-		out.URL = *in.URL
+		out.URL = url.String()
 	}
 
 	if in.HTTPConfig != nil {
@@ -443,7 +465,7 @@ func (cg *configGenerator) convertSlackConfig(ctx context.Context, in monitoring
 	if in.APIURL != nil {
 		url, err := cg.getValidURLFromSecret(ctx, crKey.Namespace, *in.APIURL)
 		if err != nil {
-			return nil, errors.Wrap(err, "convertSlackConfig failed. 'apiURL' is invalid")
+			return nil, err
 		}
 		out.APIURL = url
 	}
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -34,6 +34,8 @@ import (
 	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/timeinterval"
 	"gopkg.in/yaml.v2"
+
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
@@ -196,6 +198,19 @@ func (cg *configGenerator) enforceNamespaceForRoute(r *route, namespace string) 
 	r.Continue = true
 
 	return r
+}
+
+func (cg *configGenerator) getValidURLFromSecret(ctx context.Context, namespace string, selector v1.SecretKeySelector) (string, error) {
+	url, err := cg.store.GetSecretKey(ctx, namespace, selector)
+	if err != nil {
+		return "", errors.Errorf("failed to get key %q from secret %q", selector.Key, selector.Name)
+	}
+
+	url = strings.TrimSpace(url)
+	if _, err := ValidateURL(url); err != nil {
+		return url, errors.Wrapf(err, "invalid url %s in secret %s config", url, selector.Name)
+	}
+	return url, nil
 }
 
 func (cg *configGenerator) convertRoute(in *monitoringv1alpha1.Route, crKey types.NamespacedName) *route {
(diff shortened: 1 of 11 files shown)


paulfantom (approved):
lgtm

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -37,10 +36,6 @@ const (
 	AlertmanagerConfigKindKey = "alertmanagerconfig"
 )
 
-var (
-	opsGenieTypeRe = regexp.MustCompile("^(team|user|escalation|schedule)$")
```
simonpasquier, line 41: The regexp isn't used anywhere so I assume that it got lost in some refactoring a while ago but the validation should apply to the OpsGenieConfigResponder's type.

pkg/alertmanager/operator.go
```diff
@@ -1040,6 +1044,21 @@ func (c *Operator) selectAlertmanagerConfigs(ctx context.Context, am *monitoring
 	return res, nil
 }
 
+func checkAlertmanagerRootConfig(ctx context.Context, conf *alertmanagerConfig) error {
+	if conf == nil || conf.Route == nil {
+		return nil
+	}
```
simonpasquier, line 1050: Alertmanager wants at least one route, and that route needs a receiver.

pkg/alertmanager/operator.go
```diff
@@ -1040,6 +1044,21 @@ func (c *Operator) selectAlertmanagerConfigs(ctx context.Context, am *monitoring
...
+
+	if len(conf.Route.Matchers) > 0 {
+		return errors.Errorf("'matchers' not permitted on root route: %v", conf.Route.Matchers)
+	}
+
+	if len(conf.Route.MuteTimeIntervals) > 0 {
+		return errors.Errorf("'mute_time_intervals' not permitted on root route: %v", conf.Route.MuteTimeIntervals)
+	}
```
simonpasquier, line 1058: another option would be to unset the 2 fields?

pkg/alertmanager/validation.go
```diff
@@ -96,6 +124,12 @@ func validateOpsGenieConfigs(configs []monitoringv1alpha1.OpsGenieConfig) error
 		if err := config.Validate(); err != nil {
 			return err
 		}
+		if config.APIURL != "" {
+			_, err := ValidateURL(config.APIURL)
+			if err != nil {
```
simonpasquier, line 129: (nit)

```suggestion
			if _, err := ValidateURL(config.APIURL); err != nil {
```

pkg/alertmanager/validation.go
```diff
@@ -88,6 +105,17 @@ func validateReceivers(receivers []monitoringv1alpha1.Receiver) (map[string]stru
 
 // validatePagerDutyConfigs is a no-op
 func validatePagerDutyConfigs(configs []monitoringv1alpha1.PagerDutyConfig) error {
+	for _, conf := range configs {
+		if conf.URL != "" {
+			_, err := ValidateURL(conf.URL)
+			if err != nil {
```
simonpasquier, line 111: (nit)
```suggestion
			if _, err := ValidateURL(conf.URL); err != nil {
```

pkg/alertmanager/amcfg.go
```diff
@@ -428,6 +428,9 @@ func (cg *configGenerator) convertSlackConfig(ctx context.Context, in monitoring
 			return nil, errors.Errorf("failed to get key %q from secret %q", in.APIURL.Key, in.APIURL.Name)
 		}
 		out.APIURL = strings.TrimSpace(url)
+		if _, err := ValidateURL(out.APIURL); err != nil {
```
simonpasquier, line 431: since there are several places where we do 1) get secret key ref and 2) validate URL, maybe it's worth having a common method? it could also trim spaces as done here (but not at the other places currently).

pkg/alertmanager/operator.go
```diff
@@ -903,6 +903,10 @@ receivers:
 		return errors.Wrap(err, "base config from Secret could not be parsed")
 	}
 
+	if err := checkAlertmanagerRootConfig(ctx, baseConfig); err != nil {
```
simonpasquier, line 906: could be delegated to `loadCfg` (which would be good to have renamed BTW)?

pkg/alertmanager/operator.go
```diff
@@ -1040,6 +1044,21 @@ func (c *Operator) selectAlertmanagerConfigs(ctx context.Context, am *monitoring
...
+
+	if len(conf.Route.MuteTimeIntervals) > 0 {
+		return errors.Errorf("'mute_time_intervals' not permitted on root route: %v", conf.Route.MuteTimeIntervals)
+	}
+	return nil
+}
+
 // checkAlertmanagerConfig verifies that an AlertmanagerConfig object is valid
 // and has no missing references to other objects.
 func checkAlertmanagerConfig(ctx context.Context, amc *monitoringv1alpha1.AlertmanagerConfig, amVersion semver.Version, store *assets.Store) error {
```
simonpasquier, line 1064: side-note: I find that the function name is confusing, my first instinct it to assume that it checks the Alertmanager global configuration. What about checkAlertmanagerConfigResource instead?

pkg/alertmanager/validation.go
```diff
@@ -197,10 +242,13 @@ func validatePushoverConfigs(configs []monitoringv1alpha1.PushoverConfig) error
 		}
 
 		if config.Retry != "" {
-			_, err := time.ParseDuration(config.Retry)
+			d, err := time.ParseDuration(config.Retry)
 			if err != nil {
 				return errors.New("invalid retry duration")
 			}
+			if d < 30*time.Second {
```
simonpasquier, line 249: i missed the reference for this check

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
PhilipGough: I don't think it should since from https://prometheus.io/docs/alerting/latest/configuration/#responder it is a template string

```yaml
# "team", "user", "escalation" or schedule".
type: <tmpl_string>
```

pkg/alertmanager/operator.go
PhilipGough: ok, I was trying to avoid a panic but armed with that knowledge I will improve the check :)

pkg/alertmanager/operator.go
PhilipGough: I thought about that also, we could also do that in sanitise func I think, but I preferred this solution, since I thought it was a large enough issue to force a hard error. I am not tied to it though and can strip it if that is preferable

pkg/alertmanager/operator.go
PhilipGough: good point

pkg/alertmanager/operator.go
PhilipGough: yes, when I added the other func I thought the same, I expected this to take care of it at first glance. will update

pkg/alertmanager/validation.go
PhilipGough:
```yaml
# How often the Pushover servers will send the same notification to the user.
# Must be at least 30 seconds.
[ retry: <duration> | default = 1m ]
```

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
simonpasquier: ah my bad!

pkg/alertmanager/operator.go
simonpasquier: lets keep it this way. I think that fixing bad inputs would be useful but it probably deserves a separate discussion.

pkg/alertmanager/validation.go
simonpasquier: ack but I don't think that Alertmanager would enforce the limit.

PhilipGough: ok fair enough, in that case then I think we can remove the duration checks entirely in this block since they are covered now with OpenApi validation?

simonpasquier: sgtm

pkg/alertmanager/amcfg.go
```diff
@@ -52,6 +59,22 @@ func loadCfg(s string) (*alertmanagerConfig, error) {
 		return nil, err
 	}
 
+	rootRoute := cfg.Route
+	if rootRoute == nil {
+		return nil, errors.New("root route must exist")
+	}
+
+	if rootRoute.Receiver == "" {
+		return nil, errors.New("root routes receiver must exist")
```
simonpasquier, line 68:
```suggestion
		return nil, errors.New("root route's receiver must exist")
```

pkg/alertmanager/amcfg.go
```diff
@@ -52,6 +59,22 @@ func loadCfg(s string) (*alertmanagerConfig, error) {
...
+	}
+
+	if len(rootRoute.Matchers) > 0 {
```
simonpasquier, line 71: should check `rootRoute.Match` and `routeRoute.MatchRE` too?

pkg/alertmanager/amcfg.go
```diff
@@ -52,6 +59,22 @@ func loadCfg(s string) (*alertmanagerConfig, error) {
...
+		return nil, errors.Errorf("'matchers' not permitted on root route: %v", rootRoute.Matchers)
```
simonpasquier, line 72:
```suggestion
		return nil, errors.New("'matchers' not permitted on root route")
```

pkg/alertmanager/amcfg.go
```diff
@@ -52,6 +59,22 @@ func loadCfg(s string) (*alertmanagerConfig, error) {
...
+	}
+
+	if len(rootRoute.MuteTimeIntervals) > 0 {
+		return nil, errors.Errorf("'mute_time_intervals' not permitted on root route: %v", rootRoute.MuteTimeIntervals)
```
simonpasquier, line 76:
```suggestion
		return nil, errors.New("'mute_time_intervals' not permitted on root route")
```

pkg/alertmanager/amcfg.go
```diff
@@ -376,12 +412,15 @@ func (cg *configGenerator) convertWebhookConfig(ctx context.Context, in monitori
 	}
 
 	if in.URLSecret != nil {
-		url, err := cg.store.GetSecretKey(ctx, crKey.Namespace, *in.URLSecret)
+		url, err := cg.getValidURLFromSecret(ctx, crKey.Namespace, *in.URLSecret)
 		if err != nil {
-			return nil, errors.Errorf("failed to get key %q from secret %q", in.URLSecret.Key, in.URLSecret.Name)
+			return nil, errors.Wrap(err, "convertWebhookConfig failed")
```
simonpasquier, line 417: I think that the caller already wraps the error indicating that it's a WebhookConfig resource.

pkg/alertmanager/validation.go
```diff
@@ -15,16 +15,18 @@
...
-	"net/url"
-	"strings"
-	"time"
-
 	"github.com/pkg/errors"
 	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
+	"github.com/prometheus/alertmanager/config"
+	"net"
+	"regexp"
+	"strings"
```
simonpasquier, line 25: the imports got mixed up :)

simonpasquier (commented):
a few nits

simonpasquier (commented):
just a few minor comments. Could you also fix the wrapping message below

https://github.com/prometheus-operator/prometheus-operator/blob/009606736ffdcc14b23c77831a5ded26b3c8db43/pkg/alertmanager/validation.go#L64-L66

simonpasquier (commented):
lgtm, merge on green

PR #4443, opened 2021-12-08
Add CRDs without the description and use those in the bundle

sathieu:
## Description

Fixes: #4439

## Type of change

## Changelog entry

```release-note
- Add CRDs without the description and use those in the bundle to fix prometheus CRD being too big for kubectl apply
```


paulfantom (changes requested):
This is a breaking change when we factor in `kubectl explain` usage of CRD descriptions.

sathieu:
Abandoning this PR, I'll work on implementing server-side apply in ArgoCD instead (https://github.com/argoproj/gitops-engine/pull/363).

PR #4445, opened 2021-12-10
Consolidate logging setup into a standalone package

PhilipGough:
## Description

Looking at #4437 was the motivation for this since we will need to, for a third time, redeclare those types and logging bootstrap boilerplate. 

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -30,6 +30,7 @@ import (
 	"syscall"
 	"time"
 
+	logging "github.com/prometheus-operator/prometheus-operator/internal/log"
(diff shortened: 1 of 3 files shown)


simonpasquier (commented):
lgtm but maybe we can also move the flag handling to the new log package?

PhilipGough:
@simonpasquier we could, both components are using different packages however for flags

simonpasquier:
> both components are using different packages however for flags

ah obsviously I missed that :)
lets not worry about it yet then.

simonpasquier (approved):

PR #4446, opened 2021-12-10
Prefer generating Operator CLI MD with mdox

PhilipGough:
## Description

I noticed there are several failings in the existing `po-docgen` tool when it comes to generating the MD at [tip](https://github.com/prometheus-operator/prometheus-operator/blob/335ebbc2f6ecf10b699821fa8cebcbff4a718ca7/Documentation/operator.md).

For example the description for `log-*` are missing as are many of the default values. The problem is that the tool is using https://pkg.go.dev/go/ast to traverse the code and it is currently unable to handle strings or constants defined in other packages, as well as descriptions that are built using a function.

While this is not really a big issue, and can likely be ignored as an annoyance, this change suggests we might use one of our existing tools in the chain https://github.com/bwplotka/mdox to build output from the `--help` command.

While the output is not as pretty, it is good enough IMO for a reference document and it works out of the box while allowing us to deprecate part of `po-docgen` entirely, as we discussed also in https://github.com/prometheus-operator/prometheus-operator/issues/4376

## Type of change

## Changelog entry

```release-note

```


scripts/generate/cli-docs.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# exit immediately when a command fails
+set -e
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+
+if [ "$1" == "operator" ]; then
+  # Print the output via usage but remove the line containing the executable info
+  go run cmd/operator/main.go --help 2> >(grep -v 'Usage of')
+fi
```
paulfantom, line 10: It is unclear to me where this script is used in the docs generation pipeline (in Makefile). Could you clarify why we need this bash script?

paulfantom (commented):
I like this approach a lot! :+1: 
Just one nit

scripts/generate/cli-docs.sh
PhilipGough: Yes, I should have marked this WIP really, sorry for the confusion. 

So this is being called [here](https://github.com/prometheus-operator/prometheus-operator/pull/4446/files#diff-138201325bbc7cecc4d59e865c71badd1692f1dcaa52e02f3e0d596884207321R7).

The reason I have created a script for it is because the `--help` command produces an extraneous line about the executable, which I believe will be different for macOS and linux for example. I was trying to strip that here as it did not appear to work correctly when embedded directly in the markdown.

Furthermore, this script seems to prune that line correctly now in OS X but still the check fails on ubuntu which I need to investigate why that is still producing a diff.

PhilipGough: I had to rethink this because occasionally, the build info was getting spewed into the markdown since it had to download some modules depending on timing. The latest commit requires us building the binary but avoids us having to create a script since the output is now predictable.

Documentation/operator.md
```diff
@@ -1,57 +1,87 @@
...
-draft: false
-images: []
-menu:
-  docs:
-    parent: "operator"
-weight: 1000
-toc: false
----
+```
+# Operator CLI Flags
```
simonpasquier, line 2: this isn't probably how you expected it to render

image

Eventually I think that the page should be included in https://prometheus-operator.dev/ so I'm not sure that we should remove the [front matter](https://gohugo.io/content-management/front-matter/) header.

PhilipGough: noted and restored

paulfantom (approved):
lgtm on green, but let's hold off from merging this until 0.53 release is done

simonpasquier:
v0.53.0 is released, I'm merging this PR. Thanks!

PR #4449, opened 2021-12-14
Add tls-assets sharding to alertmanager

gottwald:
## Description

k8s secrets have a default size limitation of 1048576 bytes. Once this
size gets exceeded the k8s API server will refuse to update the secret.
This introduces the sharding approach used in prometheus controller
to the alermanager controller.

Fixes #4353
Refs #4427

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [x] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
- Add tls-assets secret sharding for alertmanager to support a high amount of tls assets
```

PR #4453, opened 2021-12-15
cmd/po-docgen: fix links to the Kube API documentation

simonpasquier:
## Description

As noticed in #4449, the Kubernetes API website for 1.17 isn't accessible anymore. Eventually we'll get rid of po-docgen (see #4376) so this is only fixing the issue at hand.

## Type of change

## Changelog entry

```release-note

```


paulfantom (approved):
lgtm

PR #4468, opened 2021-12-24
Support SNS Receiver in AlertmanagerConfig CR

PhilipGough:
## Description

Support SNS Receiver in AlertmanagerConfig CR

Follows on from the work originally started by @phin1x  in #4374 - thanks for your contribution and I have again added you as co-author here for credit.

Is related to #4236

## Type of change

## Changelog entry

```release-note
Support SNS Receiver in AlertmanagerConfig CR
```


pkg/alertmanager/validation.go
```diff
@@ -241,6 +244,15 @@ func validatePushoverConfigs(configs []monitoringv1alpha1.PushoverConfig) error
 	return nil
 }
 
+func validateSnsConfigs(configs []monitoringv1alpha1.SNSConfig) error {
+	for _, config := range configs {
+		if (config.TargetARN == "") != (config.TopicARN == "") != (config.PhoneNumber == "") {
```
fpetkovski, line 249: What would happen here if none of the values are set? By my interpretation this condition will be true, but I'm not sure if that's what we want.

pkg/alertmanager/amcfg.go
```diff
@@ -888,6 +901,48 @@ func (cg *configGenerator) convertPushoverConfig(ctx context.Context, in monitor
...
+		if err != nil {
+			return nil, err
+		}
+		out.HTTPConfig = httpConfig
+	}
+
+	if in.Sigv4 != nil {
+		accessKey, err := cg.store.GetSecretKey(ctx, crKey.Namespace, *in.Sigv4.AccessKey)
+		if err != nil {
+			return nil, errors.Errorf("failed to get secret %q", in.Sigv4.AccessKey)
```
simonpasquier, line 927: it would be better to wrap `err` since it should already contain the name and key of the reference

```suggestion
			return nil, errors.Wrap(err, "failed to get access key")
```

pkg/alertmanager/amcfg.go
```diff
@@ -888,6 +901,48 @@ func (cg *configGenerator) convertPushoverConfig(ctx context.Context, in monitor
...
+		}
+
+		secretKey, err := cg.store.GetSecretKey(ctx, crKey.Namespace, *in.Sigv4.SecretKey)
+		if err != nil {
+			return nil, errors.Errorf("failed to get secret %q", in.Sigv4.SecretKey)
```
simonpasquier, line 932:
```suggestion
			return nil, errors.Wrap(err, "failed to get AWS secret key")
```

pkg/alertmanager/validation.go
simonpasquier: It would evaluate to `true != true != true` -> `false != true` -> `true` so it works as expected but I had to think twice :sweat_smile: 
FWIW the test is taken from the upstream Alertmanager code base.

fpetkovski: Oh wow, totally unexpected but it does seem to work out in the end :smile:

PhilipGough: Yes, as mentioned this (and some other validations) are taken directly from the alertmanager project. Agree that it is not the easiest code to read so we can change it here if ye prefer?

fpetkovski: I am fine with leaving it as it is, I learned something new from it :)

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -717,6 +719,45 @@ type PushoverConfig struct {
 	HTTPConfig *HTTPConfig `json:"httpConfig,omitempty"`
 }
 
+// SNSConfig configures notifications via aws sns.
```
simonpasquier, line 722: (nit)
```suggestion
// SNSConfig configures notifications via AWS SNS.
```

simonpasquier (commented):
lgtm, just one final suggestion.

simonpasquier (approved):
lgtm

Issue #4470, opened 2021-12-30
JobLabel comment may cause misunderstanding


qixiaobo:
https://github.com/prometheus-operator/prometheus-operator/blob/9c0db5656f04e005de6a0413fd8eb8f11ec99757/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml#L492
>jobLabel:
>                description: "Chooses the label of the Kubernetes `Endpoints`. ItsMaksim Nabokikh, 7 months ago: • Bump golangci->lint and switch to the revive (…
>                 value will be used for the `job`-label's value of the created metrics.
>                  \n Default & fallback value: the name of the respective Kubernetes
>                 `Endpoint`."
>                type: string

It may leads us to add label to endpoints like below

```yaml
apiVersion: v1
kind: Endpoints
metadata:
  name: xxx
  namespace: xxx
  labels:
    app: jenkins-master
    xxxlabel: xxx
subsets:
  - addresses:
      - ip: xxx
```
We should just point out this is service

PR #4471, opened 2021-12-30
Just add some comment to avoid misunderstanding joblabel

qixiaobo:
https://github.com/prometheus-operator/prometheus-operator/issues/4470

## Description

Just add some comments to describe clearly about joblabel of  servicemonitor

## Type of change

## Changelog entry

```release-note
Make joblabel of servicemonitor comment more clearly
```

### DIFF
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -889,7 +889,7 @@ type ServiceMonitor struct {
 // +k8s:openapi-gen=true
 type ServiceMonitorSpec struct {
 	// Chooses the label of the Kubernetes `Endpoints`.
-	// Its value will be used for the `job`-label's value of the created metrics.
+	// Its value which comes from the same name service's label will be used for the `job`-label's value of the created metrics.
 	//
 	// Default & fallback value: the name of the respective Kubernetes `Endpoint`.
 	JobLabel string `json:"jobLabel,omitempty"`
--- a/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
+++ b/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
@@ -490,7 +490,8 @@ spec:
                 type: array
               jobLabel:
                 description: "Chooses the label of the Kubernetes `Endpoints`. Its
-                  value will be used for the `job`-label's value of the created metrics.
+                  value which comes from the same name service's label will be used 
+                  for the `job`-label's value of the created metrics.
                   \n Default & fallback value: the name of the respective Kubernetes
                   `Endpoint`."
                 type: string


simonpasquier:
:wave: you need to do the change in https://github.com/prometheus-operator/prometheus-operator/blob/2634a425951534e0447b499ec9aa71babf21d3dd/pkg/apis/monitoring/v1/types.go#L891-L894, then run `make generate` and commit all the files.

qixiaobo:
> 



> 👋 you need to do the change in
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/2634a425951534e0447b499ec9aa71babf21d3dd/pkg/apis/monitoring/v1/types.go#L891-L894
> 
> , then run `make generate` and commit all the files.

Thanks for remind. I will modify this file https://github.com/prometheus-operator/prometheus-operator/blob/2634a425951534e0447b499ec9aa71babf21d3dd/pkg/apis/monitoring/v1/types.go#L891-L894 too

pkg/apis/monitoring/v1/types.go
```diff
@@ -889,7 +889,7 @@ type ServiceMonitor struct {
 // +k8s:openapi-gen=true
 type ServiceMonitorSpec struct {
 	// Chooses the label of the Kubernetes `Endpoints`.
-	// Its value will be used for the `job`-label's value of the created metrics.
+	// Its value which comes from the same name service's label will be used for the `job`-label's value of the created metrics.
 	//
 	// Default & fallback value: the name of the respective Kubernetes `Endpoint`.
```
simonpasquier, line 894: I believe that the default value is the name of the associated service?

pkg/apis/monitoring/v1/types.go
```diff
@@ -889,7 +889,7 @@ type ServiceMonitor struct {
 // +k8s:openapi-gen=true
 type ServiceMonitorSpec struct {
 	// Chooses the label of the Kubernetes `Endpoints`.
```
simonpasquier, line 891:
```suggestion
	// Label of the Kubernetes service to be used for the target's `job` label value.
	// If the service's label is missing or the value is empty, the `job` value will be set to the service's name.
```

simonpasquier (commented):
I think that we should rewrite the full section to be more accurate. IIUC the operator only relies on services, not endpoints.

pkg/apis/monitoring/v1/types.go
qixiaobo: > I believe that the default value is the name of the associated service?

Yeah, but actually this value as same as endpoint's name

pkg/apis/monitoring/v1/types.go
qixiaobo: Very appreciate

pkg/apis/monitoring/v1/types.go
simonpasquier: sure but the prometheus configuration defaults to the service's name so I think it's better to mention "service" rather than "endpoint".

https://github.com/prometheus-operator/prometheus-operator/blob/86df45bbbf699f23271007797297792389fef24f/pkg/prometheus/promcfg.go#L1309-L1313

PR #4482, opened 2022-01-06
pkg/prometheus replace storage.tsdb command-line args for agent

Cheshirez:
## Description

Generating statefulsets by prometheus-operator does contain `--storage.tsdb.path` and ` --storage.tsdb.retention.time` when pod is  expecting only `--storage.agent.path` or nothing instead when  enabled `--enable-feature=agent`. This put  Prometheus to error state, preventing boot.
This PR switches default storage command-line args when feature `agent` is enabled.

Related to #3989  [(comment)](https://github.com/prometheus-operator/prometheus-operator/issues/3989#issuecomment-974137486)

>The important bit is that the Prometheus arguments need to be overridden because the storage flags are different (TSDB path and retention in particular)

## Type of change

## Changelog entry

```release-note
Set correct storage command-line args to Prometheus StatefulSet in Agent mode
```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -308,6 +308,7 @@ func makeStatefulSetSpec(
 	// Prometheus may take quite long to shut down to checkpoint existing data.
 	// Allow up to 10 minutes for clean termination.
 	terminationGracePeriod := int64(600)
+	prometheusAgentMode := false
 
 	promVersion := operator.StringValOrDefault(p.Spec.Version, operator.DefaultPrometheusVersion)
 	pImagePath, err := operator.BuildImagePath(
@@ -353,9 +354,14 @@ func makeStatefulSetSpec(
 
 	promArgs = append(promArgs,
 		monitoringv1.Argument{Name: "config.file", Value: path.Join(confOutDir, configEnvsubstFilename)},
-		monitoringv1.Argument{Name: "storage.tsdb.path", Value: storageDir},
 		monitoringv1.Argument{Name: "web.enable-lifecycle"},
 	)
+
+	if prometheusAgentMode {
+		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.agent.path", Value: storageDir})
+	} else {
+		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.path", Value: storageDir})
+	}
 
 	if p.Spec.Rules.Alert.ForOutageTolerance != "" {
 		promArgs = cg.WithMinimumVersion("2.4.0").AppendCommandlineArgument(promArgs, monitoringv1.Argument{Name: "rules.alert.for-outage-tolerance", Value: p.Spec.Rules.Alert.ForOutageTolerance})
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -344,10 +344,17 @@ func makeStatefulSetSpec(
 		retentionTimeFlagValue = defaultRetention
 	}
 
-	if retentionTimeFlagValue != "" {
+	for _, f := range p.Spec.EnableFeatures {
+		switch f {
+		case "agent":
+			prometheusAgentMode = true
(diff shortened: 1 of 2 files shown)


simonpasquier:
The current consensus among maintainers is that we want a Prometheus Agent CRD to support the agent mode properly. While I understand where this PR is going to, it would create a maintenance cost in the existing Prometheus CRD that we're not ready to commit.

cc @fpetkovski @paulfantom

paulfantom (changes requested):
As @simonpasquier said, we would like to have proper support for Prometheus Agent by designing a separate CRD for it. The consensus was that we shouldn't add patches like this to the Prometheus CRD.

Cheshirez:
@paulfantom I've missed there's route on dedicated PrometheusAgent CRD, that makes this patch redundant indeed.

However benefits of new CRD for Agents only aren't clear for me.

From user standpoint Agent remains the same Prometheus and any his params are in Prometheuses CR (scrape configs, selectors, resources and affinty/toleration, secrets mount, replicas and shards, external labels), when remote_write haven't any Agent-related instructions.
The difference is in `enableFeatures: ["agent"]`, which directly translates to `--enable-feature=agent`, and skipping alerting stuff.

In that case dedicated CRD for Agents would help users to not put breaking configuration parts which are obviously redundant (and Prometheus explicitly reporting about it in log), but require to control two CR types of about the same and adjust them all if any change come to both regular and Agent mode. Same from maintaining - it doubles work to support both regular and agent CRD.

Would appreciate help to clarify benefits of PrometheusAgent CRD or point to related doc/discussion.


As a user running 100+ k8s clusters with heavy metrics flows I want Agents support for prometheus-operator badly, since it allow to have reliable remote_writes with lightweight Agents (combining sharding and replicas) what makes uninterruptible data on LTS because Agent replicas starting in seconds and free from storage-related issues, that negates impact on monitoring when partial outages in k8s clusters happens and all that without major resource overhead. Would like to help, but need more clarity :)

fpetkovski:
@Cheshirez here is the doc we started last week, feel free to add your comments there: https://docs.google.com/document/d/1LnnxGOaypmwgczuTp_nLyWG0EiF6ohNCsZf3XQHDvkc/edit?usp=sharing

Cheshirez:
Thank you very much for document. Checking it and #2787  i see that's not an Agent itself awaiting to be designed, but rather ways to utilize scrape configs and generic `SrapeConfig` CRD for custom cases, when new `PrometheuAgent` CRD is going to be like starting point and testing platform to reimplement Prometheuses CRD in result.

From this standpoint current patch is not adding maintenance costs as not interfere with scrape configs and `PrometheusAgent` CRD, but rather fixing a hardcoded value of generating args. 

This shouldn't break using PrometheusAgent with new paradigm, but unblock users experimenting with Agent mode on existing resources to switch on polished solution once ready. Benefit of this - Agent is a new feature of Prometheus, so as earlier it became available - as more cases community can generate, more issues collect and better suggestions for further development make.

Cheshirez:
Updated with resolving merge conflicts.
@paulfantom @fpetkovski 
ScrapeConfig CRD for prometheus-operator take a while, however Prometheus Agents is very anticipated feature. Could you please review this PR with assumptions from last comments?

JoaoBraveCoding:
@Cheshirez Thanks for this contribution, currently some work has started in order to provide Agent support, we are still in the very beginning but I hope that eventually your request will be fulfilled.

I'm linking here the PR for visibility https://github.com/prometheus-operator/prometheus-operator/pull/5310

Disclaimer: I'm not a maintainer

Cheshirez:
@JoaoBraveCoding Glad to hear the actual work on ScrapeConfig has been moved from doc review stage! 
Would like to test when done.
Till that the forked operator with the fix in current PR just works fine for us for more than year, allowing to use Prometheus agents with Prometheus operator.

simonpasquier:
The PrometheusAgent CRD is available with the latest release of prometheus-operator.

Issue #4481, opened 2022-01-05
Validating webhook for service monitors, pod monitors and probes


simonpasquier:
**What is missing?**

Though the operator will discard "invalid" resources before reconciling the Prometheus configuration, the user isn't notified in advance that the resource isn't correct.

For example:

https://github.com/prometheus-operator/prometheus-operator/blob/56e89dfa0bac7a6b38a962059127992e64f69301/pkg/apis/monitoring/v1/types.go#L1195-L1211

In practice, we should push as much as we can to the OpenAPI spec but there are things like "at least one of X or Y should be present" or "X can't be present if Y is defined" that can't be modeled with OpenAPI and have to rely on a validating webhook configuration.

Also related #4437 and #4476.

**Why do we need it?**

Faster feedback for users when they try to provision invalid resources.

slashpai:
Can you assign this to me?

rexagod:
@slashpai Can I take this up (if you're not working on this currently)? 🙂

slashpai:
This is on my TODO list. Unfortunately was busy with other important ones

simonpasquier:
A short design doc (in https://github.com/prometheus-operator/prometheus-operator/tree/main/Documentation/proposals) might be helpful as it might require several iterations.

slashpai:
While I was trying to refresh on this issue found that validation rules are moved to GA in [Kubernetes 1.29](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules), we can try to move possible validations over there and for anything that is not possible we can implement it in admission webhook

simonpasquier:
Superseded by #5079

PR #4483, opened 2022-01-06
pkg/prometheus: Add validation for proberSpec url field

### LINKED ISSUE #4476: The operator should validate the "url" field in the Probe resource
**What happened?**

From https://github.com/openshift/runbooks/pull/33#issuecomment-1004035511

Create a probe resource like the following:

```
kind: Probe
metadata:
  name: example
spec:
  interval: 10s
  jobName: httpstat-test
  module: http_2xx
  prober:
    path: /probe
    scheme: https
    url: https://blackbox-exporter.example.com
  targets:
    staticConfig:
      relabelingConfigs:
        - replacement: test-app
          targetLabel: app
        - replacement: endpoint-name
          targetLabel: name
        - action: labeldrop
          regex: namespace
      static:
        - 'https://httpstat.us/200'
```

Prometheus will fail to load the generated scrape configuration because the `url` contains the `https://` prefix. The value should be formatted like `host:port` or `host`.

The operator should prevent the invalid input from getting into the generated configuration. 

See above.

**Environment**

`main` branch.

N/A

* Kubernetes cluster kind:

N/A

* Manifests:

See above.

* Prometheus Operator Logs:

N/A

slashpai:
Fixes #4476

## Description

Add validation for proberSpec url field to avoid invalid url being passed

## Type of change

## Changelog entry

```release-note
Added validation for proberSpec url field in ProbeSpec
```


pkg/prometheus/operator.go
```diff
@@ -2157,3 +2163,11 @@ func validateRelabelConfig(rc monitoringv1.RelabelConfig) error {
 	}
 	return nil
 }
+
+func validateProberURL(url string) bool {
+	if strings.Contains(url, ":") {
+		hostPort := strings.Split(url, ":")
+		return govalidator.IsHost(hostPort[0]) && govalidator.IsPort(hostPort[1])
```
fpetkovski, line 2170: Can `hostPort[0]` be an IP. If so, we might need something like 
```
validHost := govalidator.IsHost(hostPort[0]) || govalidator.IsIP(hostPort[0])
return validHost && govalidator.IsPort(hostPort[1])
```

pkg/prometheus/operator_test.go
```diff
@@ -394,3 +394,73 @@ func TestValidateRelabelConfig(t *testing.T) {
...
 	}
 }
+
+func TestValidateProberUrl(t *testing.T) {
+	for _, tc := range []struct {
+		proberSpec     monitoringv1.ProberSpec
+		validProberURL bool
+	}{
+		{
+			proberSpec: monitoringv1.ProberSpec{
```
fpetkovski, line 404: Can you add a name to the test cases to indicate what each test case covers?  You can then use the name in `t.Run`.

pkg/prometheus/operator_test.go
```diff
@@ -394,3 +394,73 @@ func TestValidateRelabelConfig(t *testing.T) {
...
+		},
+		{
+			proberSpec: monitoringv1.ProberSpec{
+				URL: "192.168.178.3",
+			},
+			validProberURL: true,
+		},
+		{
+			proberSpec: monitoringv1.ProberSpec{
+				URL: "192.168.178.3:9090",
```
fpetkovski, line 429: Hm, how does this test case pass if the validation only checks whether the first part is a host?

pkg/prometheus/operator.go
```diff
@@ -2052,6 +2053,11 @@ func (c *Operator) selectProbes(ctx context.Context, p *monitoringv1.Prometheus,
 				}
 			}
 		}
+		if !validateProberURL(probe.Spec.ProberSpec.URL) {
+			err := errors.Errorf("%s url specified in proberSpec is invalid, it should be of format `hostname` or `hostname:port`", probe.Spec.ProberSpec.URL)
```
fpetkovski, line 2057:
```suggestion
			err := errors.Errorf("%s url specified in proberSpec is invalid, it should be of the format `hostname` or `hostname:port`", probe.Spec.ProberSpec.URL)
```

pkg/prometheus/operator.go
slashpai: ya good point :)

pkg/prometheus/operator_test.go
slashpai: It checks either hostname or Ip https://github.com/asaskevich/govalidator/blob/f21760c49a8d602d863493de796926d2a5c1138d/validator.go#L851-L853

pkg/prometheus/operator.go
slashpai: ok I double checked the [IsHost() method](https://github.com/asaskevich/govalidator/blob/f21760c49a8d602d863493de796926d2a5c1138d/validator.go#L851-L853 ), it checks if input is either dnsname or IP. I just need IsHost() method then I can remove IsIP() from validate function itself

slashpai:
@fpetkovski I updated, can you review again?

fpetkovski (approved):
Nice :+1:

pkg/prometheus/operator.go
```diff
@@ -2157,3 +2163,11 @@ func validateRelabelConfig(rc monitoringv1.RelabelConfig) error {
 	}
 	return nil
 }
+
+func validateProberURL(url string) bool {
```
simonpasquier, line 2167: IMHO it would be better if `validateProberURL` returns an error. This way the returned message would be clearer (e.g. "invalid host" vs. "invalid port") and it would be consistent with the other validate*() functions.

simonpasquier (commented):
side-note: `validateProberURL()` is more strict than what Prometheus currently checks but it's fine for me.

pkg/prometheus/operator.go
slashpai: I actually wrote with error as return but found it a bit tricky when checked with different types of inputs. Like the case url passed is `http://blackbox-exporter.example.com:8080` in this case when we split the string on `:` it will take `/blackbox-exporter.example.com` as hostPort[1] and if we validate and return error as invalid port it could be more confusing

pkg/prometheus/operator.go
```diff
@@ -2157,3 +2163,11 @@ func validateRelabelConfig(rc monitoringv1.RelabelConfig) error {
...
+	if strings.Contains(url, ":") {
+		hostPort := strings.Split(url, ":")
+		return govalidator.IsHost(hostPort[0]) && govalidator.IsPort(hostPort[1])
+	}
+	return govalidator.IsHost(url)
```
simonpasquier, line 2172: Taking into account #4481, it would be more convenient if the function returns an error (only one place where the error message is constructed).
Changing the function to return an error instead of a bool could look like:

```suggestion
        hostPort := strings.Split(url, ":")
	if !govalidator.IsHost(hostPort[0]) {
		return errors.Errorf("invalid host: %q", hostPort[0])
	}
	
	if len(hostPort[1]) > 1 && govalidator.IsPort(hostPort[1]) {
		return errors.Errorf("invalid port: %q", hostPort[1])
	}
	
	return nil
```

slashpai: Thank you. I added a case for handling url scheme specified as well. Else error may be reported as invalid port for a case like

``` 
      proberSpec: monitoringv1.ProberSpec{
        URL: "https://blackbox-exporter.example.com",
      }
```

slashpai: discussed offline and decided we don't need to handle each case separately so removing handling the url scheme case separately

slashpai:
@simonpasquier I updated as per your suggestion, please review :)

pkg/prometheus/operator.go
```diff
@@ -2157,3 +2163,19 @@ func validateRelabelConfig(rc monitoringv1.RelabelConfig) error {
 	}
 	return nil
 }
+
+func validateProberURL(url string) error {
+	hostPort := strings.Split(url, ":")
```
PhilipGough, line 2168: https://pkg.go.dev/net#SplitHostPort might work here? Not sure if it will negate the need for the dependency on `govalidator`

slashpai: I had looked at that but we can specify in combinations of host/ip/host:port/ip:port it wont work
https://go.dev/play/p/HsReNGKhoqX

simonpasquier (approved):
lgtm

PR #4491, opened 2022-01-11
pkg/prometheus: Add validation for ScrapeTimeout

### LINKED ISSUE #4485: Invalid scrapeTimeout value specified causes failure during prometheus configuration reload and preventing changes from being applied
**What happened?**

Create a servicemonitor like below

``` yaml
{
    "apiVersion": "monitoring.coreos.com/v1",
    "kind": "ServiceMonitor",
    "metadata": {
        "name": "example-app",
        "namespace": "test"
    },
    "spec": {
        "endpoints": [
            {
                "interval": "60s",
                "port": "http",
                "scrapeTimeout": "120s"
            }
        ],
        "jobLabel": "app.kubernetes.io/name",
        "selector": {
            "matchLabels": {
                "app": "example-app"
            }
        }
    }
}
```

Operator should validate scrapeTimeout value specified

Try creating servicemonitor like above example

**Environment**

`0.50` and above

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* Prometheus Operator Logs:

log shows error message like below

```
level=error ts=2022-01-10T12:51:02.928Z caller=main.go:763 msg="Error reloading config" err="couldn't load configuration (--config.file=\"/etc/prometheus/config_out/prometheus.env.yaml\"): parsing YAML file /etc/prometheus/config_out/prometheus.env.yaml: scrape timeout greater than scrape interval for scrape config with job name \"serviceMonitor/test/example-app/0\"" reported and all future changes won't be applied because of this problem
```

slashpai:
This change adds validation for scrapeTimeout to avoid specifying
higher value than scrapeInterval at prometheus, servicemonitor, podmonitor and probe levels

Fixes #4485

## Description

## Type of change

## Changelog entry

```release-note
Added scrapeTimeout validation to avoid invalid value being accepted
```

### DIFF
--- a/pkg/operator/validations.go
+++ b/pkg/operator/validations.go
@@ -16,6 +16,7 @@ package operator
 
 import (
 	"github.com/alecthomas/units"
+	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 )
 
@@ -32,5 +33,29 @@ func ValidateDurationField(durationField string) error {
 	if _, err := model.ParseDuration(durationField); err != nil {
 		return err
 	}
+	return nil
+}
+
+// CompareScrapeTimeoutToScrapeInterval validates value of scrapeTimeout based on scrapeInterval
+func CompareScrapeTimeoutToScrapeInterval(scrapeTimeout, scrapeInterval string) error {
+	var err error
+	var si, st model.Duration
+
+	// since default scrapeInterval set by operator is 30s
+	if scrapeInterval == "" {
+		scrapeInterval = "30s"
+	}
+
+	if si, err = model.ParseDuration(scrapeInterval); err != nil {
+		return errors.Wrapf(err, "invalid scrapeInterval %q", scrapeInterval)
+	}
+	if st, err = model.ParseDuration(scrapeTimeout); err != nil {
+		return errors.Wrapf(err, "invalid scrapeTimeout: %q", scrapeTimeout)
+	}
+
+	if st > si {
+		return errors.Errorf("scrapeTimeout %q greater than scrapeInterval %q", scrapeTimeout, scrapeInterval)
+	}
+
 	return nil
 }
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1811,6 +1811,10 @@ func (c *Operator) selectServiceMonitors(ctx context.Context, p *monitoringv1.Pr
 				break
 			}
 
+			if err = validateScrapeIntervalAndTimeout(p, endpoint.Interval, endpoint.ScrapeTimeout); err != nil {
+				break
+			}
+
 			for _, rl := range endpoint.RelabelConfigs {
 				if rl.Action != "" {
 					if err = validateRelabelConfig(*rl); err != nil {
@@ -1924,6 +1928,10 @@ func (c *Operator) selectPodMonitors(ctx context.Context, p *monitoringv1.Promet
 
 			pmAuthKey := fmt.Sprintf("podMonitor/auth/%s/%s/%d", pm.GetNamespace(), pm.GetName(), i)
 			if err = store.AddSafeAuthorizationCredentials(ctx, pm.GetNamespace(), endpoint.Authorization, pmAuthKey); err != nil {
+				break
+			}
+
+			if err = validateScrapeIntervalAndTimeout(p, endpoint.Interval, endpoint.ScrapeTimeout); err != nil {
 				break
 			}
 
@@ -2053,6 +2061,11 @@ func (c *Operator) selectProbes(ctx context.Context, p *monitoringv1.Prometheus,
 		}
 
 		if err = store.AddOAuth2(ctx, probe.GetNamespace(), probe.Spec.OAuth2, pnKey); err != nil {
+			rejectFn(probe, err)
+			continue
+		}
+
+		if err = validateScrapeIntervalAndTimeout(p, probe.Spec.Interval, probe.Spec.ScrapeTimeout); err != nil {
 			rejectFn(probe, err)
 			continue
 		}
@@ -2190,4 +2203,14 @@ func validateProberURL(url string) error {
 		}
 	}
 	return nil
+}
+
+func validateScrapeIntervalAndTimeout(p *monitoringv1.Prometheus, scrapeInterval, scrapeTimeout string) error {
+	if scrapeTimeout == "" {
+		return nil
+	}
+	if scrapeInterval == "" {
+		scrapeInterval = p.Spec.ScrapeInterval
+	}
+	return operator.CompareScrapeTimeoutToScrapeInterval(scrapeTimeout, scrapeInterval)
 }
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -80,7 +80,8 @@ func TestConfigGeneration(t *testing.T) {
 }
 
 func TestGlobalSettings(t *testing.T) {
-	type testCase struct {
+	for _, tc := range []struct {
+		Scenario           string
 		EvaluationInterval string
 		ScrapeInterval     string
 		ScrapeTimeout      string
@@ -88,12 +89,11 @@ func TestGlobalSettings(t *testing.T) {
 		QueryLogFile       string
 		Version            string
 		Expected           string
-		ExpectedErr        error
-	}
-
-	testcases := []testCase{
+		ExpectError        bool
+	}{
 		{
-			Version: "v2.15.2",
+			Scenario: "valid config",
+			Version:  "v2.15.2",
 			Expected: `global:
   evaluation_interval: 30s
   scrape_interval: 30s
@@ -104,6 +104,7 @@ scrape_configs: []
 `,
 		},
 		{
+			Scenario:           "valid evaluation interval specified",
 			Version:            "v2.15.2",
 			EvaluationInterval: "60s",
 			Expected: `global:
@@ -116,16 +117,19 @@ scrape_configs: []
 `,
 		},
 		{
+			Scenario:           "invalid evaluation interval specified #1",
 			Version:            "v2.15.2",
 			EvaluationInterval: "60 s",
-			ExpectedErr:        errors.New("invalid evaluationInterval value specified: not a valid duration string: \"60 s\""),
+			ExpectError:        true,
 		},
 		{
+			Scenario:           "invalid evaluation interval specified #2",
 			Version:            "v2.28.0",
 			EvaluationInterval: "randomvalue",
-			ExpectedErr:        errors.New("invalid evaluationInterval value specified: not a valid duration string: \"randomvalue\""),
+			ExpectError:        true,
 		},
 		{
+			Scenario:       "valid scrape interval",
 			Version:        "v2.15.2",
 			ScrapeInterval: "60s",
 			Expected: `global:
@@ -138,30 +142,48 @@ scrape_configs: []
 `,
 		},
 		{
+			Scenario:       "invalid scrape interval",
 			Version:        "v2.28.0",
 			ScrapeInterval: "30 k",
-			ExpectedErr:    errors.New("invalid scrapeInterval value specified: not a valid duration string: \"30 k\""),
+			ExpectError:    true,
 		},
 		{
-			Version:       "v2.15.2",
-			ScrapeTimeout: "30s",
+			Scenario:      "invalid scrape timeout",
+			Version:       "v2.29.0",
+			ScrapeTimeout: "some value",
+			ExpectError:   true,
+		},
+		{
+			Scenario:      "invalid scrape timeout specified when scrape interval not specified to compare with default value",
+			Version:       "v2.30.0",
+			ScrapeTimeout: "120s",
+			ExpectError:   true,
+		},
+		{
+			Scenario:       "invalid scrape timeout specified when scrape interval specified",
+			Version:        "v2.30.0",
+			ScrapeInterval: "30s",
+			ScrapeTimeout:  "60s",
+			ExpectError:    true,
+		},
+		{
+			Scenario:       "valid scrape timeout along with valid scrape interval specified",
+			Version:        "v2.15.2",
+			ScrapeInterval: "60s",
+			ScrapeTimeout:  "10s",
 			Expected: `global:
   evaluation_interval: 30s
-  scrape_interval: 30s
+  scrape_interval: 60s
   external_labels:
     prometheus: /
     prometheus_replica: $(POD_NAME)
-  scrape_timeout: 30s
+  scrape_timeout: 10s
 scrape_configs: []
 `,
 		},
 		{
-			Version:       "v2.29.0",
-			ScrapeTimeout: "some value",
-			ExpectedErr:   errors.New("invalid scrapeTimeout value specified: not a valid duration string: \"some value\""),
-		},
-		{
-			Version: "v2.15.2",
+			Scenario: "external label specified",
+			Version:  "v2.15.2",
 			ExternalLabels: map[string]string{
 				"key1": "value1",
 				"key2": "value2",
@@ -178,6 +200,7 @@ scrape_configs: []
 `,
 		},
 		{
+			Scenario:     "query log file",
 			Version:      "v2.16.0",
 			QueryLogFile: "test.log",
 			Expected: `global:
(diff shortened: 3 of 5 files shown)


pkg/operator/validations.go
```diff
@@ -34,3 +35,19 @@ func ValidateDurationField(durationField string) error {
 	}
 	return nil
 }
+
+// ValidateScrapeTimeout validates value of scrapeTimeout based on scrapeInterval
+func ValidateScrapeTimeout(scrapeTimeout, scrapeInterval string) error {
+	// since default scrapeInterval set by operator is 1m
+	if scrapeInterval == "" {
+		scrapeInterval = "60s"
```
PhilipGough, line 43: We could set this via OpenAPI in https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/types.go#L174 via an annotation:

`// +kubebuilder:default:=60s`

What do you think?

pkg/prometheus/operator.go
```diff
@@ -1763,39 +1763,75 @@ func (c *Operator) selectServiceMonitors(ctx context.Context, p *monitoringv1.Pr
...
-				break
+				rejectFn(sm, err)
+				continue
 			}
 
 			smAuthKey := fmt.Sprintf("serviceMonitor/auth/%s/%s/%d", sm.GetNamespace(), sm.GetName(), i)
 			if err = store.AddSafeAuthorizationCredentials(ctx, sm.GetNamespace(), endpoint.Authorization, smAuthKey); err != nil {
-				break
+				rejectFn(sm, err)
+				continue
```
PhilipGough, line 1813: For this and the similar changes above, I am wondering how they relate to validation of scrape timeout? Should they be dealt with in a different PR?

Also wondering what drove this change? Why do we continue to process the rest of the service monitor endpoints in case of error as opposed to dropping the whole CR? Wouldn't the log `skipping servicemonitor` be misleading since we may end up with a "partially reconciled" ServiceMonitor?

simonpasquier: I agree with @PhilipGough: we should drop the whole service monitor (same for pod monitors).

pkg/prometheus/promcfg_test.go
```diff
@@ -206,32 +228,32 @@ scrape_configs: []
...
-				t.Fatal("expected error and actual error do not match")
+			if err != nil && !tc.ExpectedErr {
+				t.Fatalf("expected no error, got: %v", err)
+			}
+			if err == nil && tc.ExpectedErr {
+				t.Fatalf("expected an error, got nil")
+			}
+			result := string(cfg)
+			if tc.Expected != string(cfg) {
+				fmt.Println(pretty.Compare(tc.Expected, result))
```
simonpasquier, line 252:
```suggestion
				t.Log(pretty.Compare(tc.Expected, result))
```

pkg/prometheus/promcfg_test.go
```diff
@@ -206,32 +228,32 @@ scrape_configs: []
...
-		if tc.ExpectedErr != nil {
-			if tc.ExpectedErr.Error() != err.Error() {
-				t.Logf("\n%s", pretty.Compare(tc.ExpectedErr.Error(), err.Error()))
-				t.Fatal("expected error and actual error do not match")
+			if err != nil && !tc.ExpectedErr {
+				t.Fatalf("expected no error, got: %v", err)
+			}
+			if err == nil && tc.ExpectedErr {
+				t.Fatalf("expected an error, got nil")
+			}
```
simonpasquier, line 249: If we expect an error, no need to check for the returned configuration.

```suggestion
			if tc.ExpectedErr {
				if err == nil {
					t.Fatalf("expected an error, got nil")
				}
				return
			}
```

pkg/prometheus/operator.go
slashpai: @PhilipGough  Sorry I was still updating the PR :) 

I was trying to make it like as `selectProbes`  because I wanted to wrap the error returned from validate function.

https://github.com/prometheus-operator/prometheus-operator/blob/14b5ca17052ea16b22bacee31354d46de4430ab4/pkg/prometheus/operator.go#L2004-L2056

I think we would need to update `selectProbe` too then?

pkg/operator/validations.go
slashpai: The purpose of checking if scrapeTimeout > 60s was incase if scrapeInterval is not specified operator sets to `60s` per https://github.com/prometheus-operator/prometheus-operator/blob/9c0db5656f04e005de6a0413fd8eb8f11ec99757/pkg/apis/monitoring/v1/types.go#L173

I am checking where it sets this value by default

slashpai: It actually sets `30s` by default, I could confirm that in minikube cluster as well. 

https://github.com/prometheus-operator/prometheus-operator/blob/5f72fbc655b416b9b50ede04ed2c0a64cda22933/pkg/prometheus/promcfg.go#L339

@simonpasquier We need to correct this in `types.go` comments since it  shows in `kubectl explain` ? or Am I missing something?

slashpai: Like @PhilipGough  suggested we can set this via OpenAPI instead of setting here

PhilipGough: I personally think we should make the OpenAPI spec the source of truth to avoid situations where we have drift between the docs and the code like you have found here.

If @simonpasquier agrees with that sentiment we can just do it for `scrapeInterval ` as part of this PR and create an issue to investigate and tackle other scenarios.

simonpasquier: I agree, lets define the default value of scrapeInterval in the OpenAPI spec. We can also add a default value for scrapeTimeout.

slashpai: @simonpasquier @PhilipGough  I tested with `//+kubebuilder:default="30s"` for interval and `//+kubebuilder:default="`10s"` for timeout. But it requires updating most of the unit tests since this default value will be set runtime as I understand


https://github.com/prometheus-operator/prometheus-operator/blob/5f72fbc655b416b9b50ede04ed2c0a64cda22933/pkg/prometheus/promcfg_test.go#L97-L103

Can we do that in separate PR by handling all the default values in openapi spec as possible?

simonpasquier (commented):
I'm fine with updating the OpenAPI spec in another PR but it might be easier to do it with this change though. When validating the scrape interval/timeout for an endpoint, we need also to take into account `spec.scrapeInterval` and `spec.scrapeTimeout` from the Prometheus resource:
* if the endpoint defines both scrape interval and timeout values then compare both directly.
* if the endpoint defines only the scrape interval value then compare it with the global scrape timeout (nothing to do if unset).
* if the endpoint defines only the scrape timeout value then compare with the global scrape interval (default to "30s" if unset).

pkg/prometheus/operator.go
```diff
@@ -1799,6 +1799,30 @@ func (c *Operator) selectServiceMonitors(ctx context.Context, p *monitoringv1.Pr
...
+				if err = validateScrapeIntervalAndTimeout(endpoint.Interval, p.Spec.ScrapeTimeout); err != nil {
+					break
+				}
+			}
+
+			if endpoint.Interval == "" && endpoint.ScrapeTimeout != "" {
+				if err = validateScrapeIntervalAndTimeout(p.Spec.ScrapeInterval, endpoint.ScrapeTimeout); err != nil {
+					break
+				}
+			}
```
simonpasquier, line 1824: I would encapsulate the logic into `validateScrapeIntervalAndTimeout()` so that we don't have to repeat the same tests in different places:

```
func validateScrapeIntervalAndTimeout(p *v1.Prometheus, scrapeInterval, scrapeTimeout string) error {
    ...
}
```

slashpai:
@simonpasquier Updated per suggestions can you review again?

pkg/operator/validations.go
```diff
@@ -34,3 +35,27 @@ func ValidateDurationField(durationField string) error {
...
+	var err error
+	var si, st model.Duration
+
+	// since default scrapeInterval set by operator is 30s
+	if scrapeInterval == "" {
+		scrapeInterval = "30s"
+	}
+
+	if si, err = model.ParseDuration(scrapeInterval); err != nil {
+		return errors.Wrapf(err, "invalid scrapeInterval value `%s` specified", scrapeInterval)
```
simonpasquier, line 50:
```suggestion
		return errors.Wrapf(err, "invalid scrapeInterval: %q", scrapeInterval)
```

pkg/operator/validations.go
```diff
@@ -34,3 +35,27 @@ func ValidateDurationField(durationField string) error {
...
+	}
+	if st, err = model.ParseDuration(scrapeTimeout); err != nil {
+		return errors.Wrapf(err, "invalid scrapeTimeout value `%s` specified", scrapeTimeout)
```
simonpasquier, line 53:
```suggestion
		return errors.Wrapf(err, "invalid scrapeTimeout: %q", scrapeTimeout)
```

pkg/operator/validations.go
```diff
@@ -34,3 +35,27 @@ func ValidateDurationField(durationField string) error {
...
+	}
+
+	if st > si {
+		return errors.Errorf("scrapeTimeout value specified `%s` should be lesser than scrapeInterval `%s`", scrapeTimeout, scrapeInterval)
```
simonpasquier, line 57:
```suggestion
		return errors.Errorf("scrapeTimeout %q greater than scrapeInterval %q", scrapeTimeout, scrapeInterval)
```

pkg/prometheus/operator.go
```diff
@@ -1799,6 +1799,12 @@ func (c *Operator) selectServiceMonitors(ctx context.Context, p *monitoringv1.Pr
 				break
 			}
 
+			if endpoint.Interval != "" || endpoint.ScrapeTimeout != "" {
```
simonpasquier, line 1802: we may have a situation where `endpoint.ScrapeTimeout` is set, `endpoint.Interval` is empty and the former is greater than the value of `p.Spec.ScrapeInterval` which would cause an error. In short I don't think that we need a test here, `validateScrapeIntervalAndTimeout` should take care of everything.

pkg/operator/validations.go
```diff
@@ -34,3 +35,27 @@ func ValidateDurationField(durationField string) error {
...
+
+// CompareScrapeTimeoutToScrapeInterval validates value of scrapeTimeout based on scrapeInterval
+func CompareScrapeTimeoutToScrapeInterval(scrapeTimeout, scrapeInterval string) error {
+	var err error
+	var si, st model.Duration
+
+	// since default scrapeInterval set by operator is 30s
+	if scrapeInterval == "" {
+		scrapeInterval = "30s"
+	}
```
simonpasquier, line 47: it's only the default for the Prometheus CRD so I don't think it should be checked here.
```suggestion
```

pkg/prometheus/operator.go
```diff
@@ -2179,3 +2198,20 @@ func validateProberURL(url string) error {
...
+	if scrapeInterval != "" && scrapeTimeout == "" {
+		if p.Spec.ScrapeTimeout != "" {
+			return operator.CompareScrapeTimeoutToScrapeInterval(p.Spec.ScrapeTimeout, scrapeInterval)
+		}
+	}
+
+	if scrapeInterval == "" && scrapeTimeout != "" {
+		return operator.CompareScrapeTimeoutToScrapeInterval(scrapeTimeout, p.Spec.ScrapeInterval)
+	}
+	return nil
```
simonpasquier, line 2216: I think that it can be simplified a bit. If the scrape timeout isn't defined (or 0) then there's nothing to check because Prometheus will do the right thing (if scrape interval > global timeout then timeout = scrape interval otherwise timeout = global timeout).

```suggestion
	if scrapeTimeout == "" {
		return nil
	}

	if scrapeInterval == "" {
		scrapeInterval = p.Spec.ScrapeInterval
		if scrapeInterval == "" {
			scrapeInterval = "30s"
		}
	}

	return operator.CompareScrapeTimeoutToScrapeInterval(scrapeTimeout, scrapeInterval)
```

pkg/operator/validations.go
slashpai: The default value is set here https://github.com/prometheus-operator/prometheus-operator/blob/5f72fbc655b416b9b50ede04ed2c0a64cda22933/pkg/prometheus/promcfg.go#L339 which comes after this validation so we need to check here right?

pkg/prometheus/operator.go
```diff
@@ -2179,3 +2194,16 @@ func validateProberURL(url string) error {
 	}
 	return nil
 }
+
+func validateScrapeIntervalAndTimeout(p *monitoringv1.Prometheus, scrapeInterval, scrapeTimeout string) error {
+	if scrapeTimeout == "" {
+		return operator.CompareScrapeTimeoutToScrapeInterval(p.Spec.ScrapeTimeout, scrapeInterval)
```
slashpai, line 2200: I modified it to handle the case

 `if the endpoint defines only the scrape interval value then compare it with the global scrape timeout (nothing to do if unset).`

pkg/prometheus/operator.go
```diff
@@ -2179,3 +2194,18 @@ func validateProberURL(url string) error {
 	}
 	return nil
 }
+
+func validateScrapeIntervalAndTimeout(p *monitoringv1.Prometheus, scrapeInterval, scrapeTimeout string) error {
+	if scrapeTimeout == "" {
+		if scrapeInterval != "" {
+			return operator.CompareScrapeTimeoutToScrapeInterval(p.Spec.ScrapeTimeout, scrapeInterval)
```
slashpai, line 2201: I modified it to handle the case

`if the endpoint defines only the scrape interval value then compare it with the global scrape timeout (nothing to do if unset).`

pkg/prometheus/operator.go
slashpai: We need this part in `CompareScrapeTimeoutToScrapeInterval()` since we use that in prometheus crd level validation as well https://github.com/prometheus-operator/prometheus-operator/pull/4491/files#diff-453ce9c721c51bf9591c2b212e9ce50157b9c40d1b211177bf502be268099656R254-R256

```go
if scrapeInterval == "" {
   scrapeInterval = "30s"
}
```

slashpai: @simonpasquier I updated per above above suggestion but this wont handle 2nd case mentioned right?. 

`if the endpoint defines only the scrape interval value then compare it with the global scrape timeout (nothing to do if unset).`

https://github.com/prometheus-operator/prometheus-operator/pull/4491#pullrequestreview-851515889

slashpai: Reviewed it here so we are good https://github.com/prometheus/prometheus/blob/2c61d29b2a865db7e1a8d188e08b68ed154929c8/config/config.go#L292-L301

PhilipGough: Unless I am missing something, when I read the API docs it isn't clear to me that the endpoint config, if not specified will be inherited from the Prometheus CR or that if one field is provided it will be compared against the global field so maybe we should call that out?

That feels slightly obtuse in the "self-service" model however. Perhaps we could take an all or nothing approach and declare than unless both fields are required to be either omitted or provided as a pair, but that would be a breaking change i guess.

slashpai: From what I understand thats default set by prometheus when scrape interval not specified in job section it sets to global scrape interval and thats not handled by operator https://github.com/prometheus/prometheus/blob/2c61d29b2a865db7e1a8d188e08b68ed154929c8/config/config.go#L289-L290

slashpai: I can update that in separate pr

slashpai:
@simonpasquier addressed comments

PhilipGough:
small nit, looks good for the most part. I'd prefer we tackle the docs change mentioned above here though or have an issue at least for an overhaul.

pkg/prometheus/promcfg_test.go
```diff
@@ -80,20 +80,20 @@ func TestConfigGeneration(t *testing.T) {
...
 		ScrapeTimeout      string
 		ExternalLabels     map[string]string
 		QueryLogFile       string
 		Version            string
 		Expected           string
-		ExpectedErr        error
-	}
-
-	testcases := []testCase{
+		ExpectedErr        bool
```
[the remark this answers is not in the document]
slashpai: Both looks good to me but I can update

slashpai:
@PhilipGough @simonpasquier I have logged #4524 for improving documentation and moving validations to OpenAPI. I can work on that once this is merged, please feel free to assign to me

simonpasquier:
lgtm

PhilipGough (approved):
lgtm

simonpasquier:
leaving the final say to @fpetkovski if he wants to include it in #4525 or not :)

PR #4500, opened 2022-01-17
Don't propogate fluxcd labels to statefulset

timmow:
## Description

We use [fluxcd](https://github.com/fluxcd/flux2) with a
[kustomization](https://fluxcd.io/docs/components/kustomize/kustomization/)
referring to manifests generated by
[kube-prometheus](https://github.com/prometheus-operator/kube-prometheus)

Flux sets these labels on objects
it manages. If an object it manages exists on the cluster, but does not
exist in the defined kustomization, fluxcd will purge the object. As the
operator manages the statefulset, this does not exist in the
kustomization and flux will delete the statefulset when it reconciles.
We currently work around this by disabling flux garbage collection on
Prometheus objects, but it would be good if the operator didnt propogate
this label and we didnt need this workaround.

## Type of change

## Changelog entry

```release-note
Stop propogating fluxcd labels to prometheus statefulset to improve process of managing operator with fluxcd
```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -155,7 +155,7 @@ func makeStatefulSet(
 	// pruned by kubectl
 	annotations := make(map[string]string)
 	for key, value := range p.ObjectMeta.Annotations {
-		if !strings.HasPrefix(key, "kubectl.kubernetes.io/") {
+		if !(strings.HasPrefix(key, "kubectl.kubernetes.io/") || strings.HasPrefix(key, "kustomize.toolkit.fluxcd.io")) {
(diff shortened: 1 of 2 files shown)


fpetkovski:
I wonder if we should make this improvement through a command line argument or CRD field so that we don't have to change the operator every time we need to make an exception. @simonpasquier @paulfantom

simonpasquier:
We can keep a set of prefixes in the operator for well-known annotation prefixes but a CLI argument is a good idea to e.
We should also factorize the logic across prometheus, alertmanager and thanos ruler controllers so there's only one place where this happens.

PR #4499, opened 2022-01-17
alertmanager: Add follow_redirects and oauth fields to httpClientConfig

PhilipGough:
## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

Close the drift between internal httpClientConfig and [the latest `HTTPClientConfig` from Prometheus ](https://github.com/prometheus/common/blob/00591a3ea9c0d18f6bc983818a23901d4154077f/config/http_config.go#L178-L199)

Add follow_redirects and support oauth config.

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [x] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Support following redirects and Oauth2 in HTTP Client config in raw alertmanager config secret
```

PR #4501, opened 2022-01-18
Support Oauth2 in AlertmanagerConfig CRD

PhilipGough:
## Description

Follows on from #4499 

## Type of change

## Changelog entry

```release-note
Support Oauth2 in AlertmanagerConfig CRD
```


pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -545,6 +548,10 @@ type HTTPConfig struct {
 	// Optional proxy URL.
 	// +optional
 	ProxyURL string `json:"proxyURL,omitempty"`
+	// FollowRedirects specifies whether the client should follow HTTP 3xx redirects.
+	// +optional
+	// +kubebuilder:default:=true
+	FollowRedirects bool `json:"followRedirects"`
```
simonpasquier, line 554: don't we want a follow-up PR to address the lack of `FollowRedirects` for all existing CRDs?

PhilipGough: yes, I think we can add it to `Prometheus` also. Again, I can add it here but it might be better to tackle the different API and code generation in a follow up?

simonpasquier: I agree, lets remove it and tackle `FollowRedirects` for all CRDs in a follow-up PR.

pkg/alertmanager/amcfg.go
```diff
@@ -1140,6 +1140,25 @@ func (cg *configGenerator) convertHTTPConfig(ctx context.Context, in monitoringv
 		out.BearerToken = bearerToken
 	}
 
+	if in.OAuth2 != nil {
```
simonpasquier, line 1143: we need a version check in `sanitize()` since OAuth2 is supported with >= 0.22.0

pkg/alertmanager/validation.go
```diff
@@ -115,6 +115,10 @@ func validatePagerDutyConfigs(configs []monitoringv1alpha1.PagerDutyConfig) erro
 		if conf.RoutingKey == nil && conf.ServiceKey == nil {
 			return errors.New("one of 'routingKey' or 'serviceKey' is required")
 		}
+
+		if err := conf.HTTPConfig.Validate(); err != nil {
```
simonpasquier, line 119: hmm so we didn't validate the HTTP config for all receivers?

pkg/apis/monitoring/v1alpha1/validation.go
```diff
@@ -22,6 +22,25 @@ import (
...
+	if (hc.OAuth2 != nil || hc.BasicAuth != nil) && hc.BearerTokenSecret != nil {
+		return fmt.Errorf("only one of basic auth, oauth2 or bearer token can be specified")
+	}
+
+	if (hc.OAuth2 != nil || hc.BasicAuth != nil) && hc.Authorization != nil {
+		return fmt.Errorf("only one of basic auth, oauth2 or authorization can be specified")
+	}
+
+	if hc.Authorization != nil && hc.BearerTokenSecret != nil {
+		return fmt.Errorf("authorization is not compatible bearer token")
```
simonpasquier, line 39:
```suggestion
		return fmt.Errorf("authorization is not compatible with bearer token")
```

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
PhilipGough: I was thinking initially that since this is the only one that touches `v1alpha1` I would include it here and update and generate the `v1` API(s) together?

pkg/alertmanager/validation.go
PhilipGough: we do validate, but down at the "check" level (not here at admission webhook level)- https://github.com/prometheus-operator/prometheus-operator/blob/1090bed455c51b7525f78a9c26e942d839bc0eb5/pkg/alertmanager/operator.go#L1371

my thoughts were that we were reading the secrets etc at that point but looking closer at some of those functions I think we might be able to include them here also

pkg/alertmanager/amcfg.go
PhilipGough: ack..missed that

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
simonpasquier: my point is more to say that if we expose the field in the CRD then we need to wire it into `configGenerator.convertHTTPConfig()` + proper version sanitization.

pkg/alertmanager/validation.go
PhilipGough: ok so i've included what is safe to validate here also. We cant remove the validation from where it already exists since other packages rely on it also. We could potentially rethink the overall "validation strategy" of the project if we wanted to.

pkg/alertmanager/amcfg.go
```diff
@@ -1226,21 +1246,30 @@ func (hc *httpClientConfig) sanitize(amVersion semver.Version, logger log.Logger
...
-	// we don't need to do any sanitization in this case and return early
-	if hc.Authorization == nil {
-		return
-	}
 
-	if hc.BasicAuth != nil {
-		msg := "'basicAuth' and 'authorization' are mutually exclusive, 'basicAuth' has taken precedence"
-		level.Warn(logger).Log("msg", msg)
+	if hc.Authorization != nil && !amVersion.GTE(semver.MustParse("0.22.0")) {
+		msg := "'authorization' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
```
simonpasquier, line 1251:
```suggestion
		msg := "'authorization' set in 'http_config' but supported in AlertManager >= 0.22.0 only - dropping field from provided config"
```

pkg/alertmanager/amcfg.go
```diff
@@ -1226,21 +1246,30 @@ func (hc *httpClientConfig) sanitize(amVersion semver.Version, logger log.Logger
...
-		msg := "'authorization' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
+
+	if hc.OAuth2 != nil && !amVersion.GTE(semver.MustParse("0.22.0")) {
+		msg := "'oauth2' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
+		level.Warn(logger).Log("msg", msg, "current_version", amVersion.String())
+		hc.OAuth2 = nil
+	}
+
+	if hc.FollowRedirects != nil && !amVersion.GTE(semver.MustParse("0.22.0")) {
+		msg := "'follow_redirects' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
```
simonpasquier, line 1263:
```suggestion
		msg := "'follow_redirects' set in 'http_config' but supported in AlertManager >= 0.22.0 only - dropping field from provided config"
```

pkg/alertmanager/amcfg.go
```diff
@@ -1226,21 +1246,30 @@ func (hc *httpClientConfig) sanitize(amVersion semver.Version, logger log.Logger
...
+		msg := "'authorization' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
+		level.Warn(logger).Log("msg", msg, "current_version", amVersion.String())
 		hc.Authorization = nil
 	}
-	// we could have returned here but useful to grab the log and bubble up the warning
-	if httpAuthzAllowed := amVersion.GTE(semver.MustParse("0.22.0")); !httpAuthzAllowed {
-		msg := "'authorization' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
+
+	if hc.OAuth2 != nil && !amVersion.GTE(semver.MustParse("0.22.0")) {
+		msg := "'oauth2' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
```
simonpasquier, line 1257:
```suggestion
		msg := "'oauth2' set in 'http_config' but supported in AlertManager >= 0.22.0 only - dropping field from provided config"
```

pkg/alertmanager/amcfg.go
```diff
@@ -1226,21 +1246,30 @@ func (hc *httpClientConfig) sanitize(amVersion semver.Version, logger log.Logger
...
+		hc.OAuth2 = nil
+	}
+
+	if hc.FollowRedirects != nil && !amVersion.GTE(semver.MustParse("0.22.0")) {
+		msg := "'follow_redirects' set in 'http_config' but  supported in AlertManager >= 0.22.0 only - dropping field from provided config"
 		level.Warn(logger).Log("msg", msg, "current_version", amVersion.String())
+		hc.FollowRedirects = nil
+	}
+
+	if hc.BasicAuth != nil && (hc.OAuth2 != nil || hc.Authorization != nil) {
```
simonpasquier, line 1268: don't we need to check for `hc.OAuth2` and `hc.Authorization` being not nil simultaneously? Also wondering if we shouldn't return an error instead since it's not like you want to use a field that isn't supported.

pkg/apis/monitoring/v1alpha1/validation.go
```diff
@@ -22,6 +22,37 @@ import (
...
+func (hc *HTTPConfig) Validate() error {
+	if hc == nil {
+		return nil
+	}
+
+	if (hc.OAuth2 != nil || hc.BasicAuth != nil) && hc.BearerTokenSecret != nil {
+		return fmt.Errorf("only one of basic auth, oauth2 or bearer token can be specified")
+	}
+
+	if (hc.OAuth2 != nil || hc.BasicAuth != nil) && hc.Authorization != nil {
```
simonpasquier, line 34: same as above, OAuth2 and BasicAuth being non-nil should be rejected?

pkg/alertmanager/amcfg.go
PhilipGough: In regards to the error, I was returning an error here originally but I believe we discussed that offline and reverted since right now in https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/alertmanager/amcfg.go#L1243 we are simply dropping the field and we were aligning the behaviour.

Saying that, my initial instinct here was that the right thing to do was to error so maybe we should do a u-turn? :)

simonpasquier: sorry for introducing confusion then :) I agree that returning an error makes more sense since the operator can't really decide which option makes sense than the others.

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
simonpasquier, line 554: https://github.com/prometheus-operator/prometheus-operator/pull/4563 uses a *bool field instead. This has the advantage to delegate the handling of the default value to Prometheus.

simonpasquier (commented):
lgtm

PR #4504, opened 2022-01-19
Address race condition in recreate flow for statefulset

### LINKED ISSUE #4519: Operator can enter a hot-loop when trying to recreate Statefulset(s)
**What happened?**

We had two similar issues reported downstream which caused us to investigate a potential race in the operator during upgrade. 

1. https://bugzilla.redhat.com/show_bug.cgi?id=2030539
1. https://bugzilla.redhat.com/show_bug.cgi?id=2042441

The `app` selector label was dropped in #4350 - this would require us to delete and recreate the statefulset which is a pattern that pre-existed in the operator code https://github.com/prometheus-operator/prometheus-operator/pull/3939#issuecomment-830080014

We believe this caused the operator to enter a hot-loop of trying to delete and recreate the sts in each sync. 
The sts is specified with foreground deletion option which means it might remain available in the api while its resources get garbage collected.

We can observe that behaviour in the following logs which get repeated over and over:

```
level=info ts=2022-01-20T16:28:49.333082906Z caller=operator.go:1218 component=prometheusoperator key=openshift-monitoring/k8s msg="sync prometheus"
level=info ts=2022-01-20T16:28:50.174976129Z caller=operator.go:741 component=alertmanageroperator key=openshift-monitoring/main msg="sync alertmanager"
level=error ts=2022-01-20T16:28:50.309593173Z caller=operator.go:1301 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="updating StatefulSet failed" reason="Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden"
level=info ts=2022-01-20T16:28:50.309727902Z caller=operator.go:1308 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="deleting and recreating StatefulSet because the update operation wasn't possible"
level=info ts=2022-01-20T16:28:50.326837954Z caller=operator.go:1218 component=prometheusoperator key=openshift-monitoring/k8s msg="sync prometheus"
level=error ts=2022-01-20T16:28:51.779818335Z caller=operator.go:1301 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="updating StatefulSet failed" reason="Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden"
level=info ts=2022-01-20T16:28:51.78036565Z caller=operator.go:1308 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="deleting and recreating StatefulSet because the update operation wasn't possible"
level=info ts=2022-01-20T16:28:53.643943094Z caller=operator.go:1218 component=prometheusoperator key=openshift-monitoring/k8s msg="sync prometheus"
level=info ts=2022-01-20T16:28:54.829240542Z caller=operator.go:741 component=alertmanageroperator key=openshift-monitoring/main msg="sync alertmanager"
level=error ts=2022-01-20T16:28:54.938966916Z caller=operator.go:1301 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="updating StatefulSet failed" reason="Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden"
level=info ts=2022-01-20T16:28:54.939019642Z caller=operator.go:1308 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="deleting and recreating StatefulSet because the update operation wasn't possible"
level=info ts=2022-01-20T16:28:54.949241093Z caller=operator.go:1218 component=prometheusoperator key=openshift-monitoring/k8s msg="sync prometheus"
level=error ts=2022-01-20T16:28:55.53075434Z caller=operator.go:1301 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="updating StatefulSet failed" reason="Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden"
level=info ts=2022-01-20T16:28:55.530903104Z caller=operator.go:1308 component=prometheusoperator key=openshift-monitoring/k8s statefulset=prometheus-k8s shard=0 msg="deleting and recreating StatefulSet because the update operation wasn't possible"
```

 It is not apparent why this issue never surfaced before or has not been reported. Perhaps something at the kube layer in later versions?

Expect the recreate flow to execute with minimal downtime and not see repreated delete requests for resources marked for deletion.

**Environment**
kube 1.22 -> 1.23
operator 0.49.0 -> 0.53.1

PhilipGough:
## Description

```
We have observed an issue where the operator fails to update
a sts and must delete, recreate, a race can occur.

The delete is a foreground cascading delete which causes
the sts to remain visible via the API server for an unknown
amount of time.

We have observed for all resources, the operator is receiving a
multitude of update requests before the delete is finalized.

This is causing a hot-loop of delete, recreate cycles which in
some cases appear never to resolve without manual intervention
i.e. (deleting the sts in the background).

Since the foreground delete will add a deletionTimestamp to the
existing spec, we can check its existence to determine we should
not continue processing the update.

While this change does not guarantee zero downtime of the pods
owned by the sts, it should at least minimize it and not cause
the problem to exacerbate.
```

This bug was reported to us downstream and thanks @simonpasquier who did the initial investigation of the problem. 

I carried out an upgrade on kube v1.22 -> 1.23 taking the operator from 0.49.0 to 0.53.1 with no apparent ill effects.

It should be noted that this change will not guarantee zero downtime, it just prevents the problem from growing and keeps potential downtime to a minimum. I observed the following on the above upgrade in Thanos:

Screenshot 2022-01-21 at 14 36 02

Fixes #4519 

## Type of change

## Changelog entry

```release-note
Avoid race during recreation of StatefulSet(s)
```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -655,6 +655,10 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 	podLabels := map[string]string{
 		"app.kubernetes.io/version": version.String(),
 	}
+	// In cases where an existing selector label is modified, or a new one is added, new sts cannot match existing pods.
+	// We should try to avoid removing such immutable fields whenever possible since doing
+	// so forces us to enter the 'recreate cycle' and can potentially lead to downtime.
+	// The requirement to make a change here should be carefully evaluated.
 	podSelectorLabels := map[string]string{
 		"app.kubernetes.io/name":       "prometheus",
 		"app.kubernetes.io/managed-by": "prometheus-operator",
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -763,13 +763,20 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return errors.Wrap(err, "failed to retrieve statefulset")
 	}
 
-	oldSpec := appsv1.StatefulSetSpec{}
+	existingStatefulSet := &appsv1.StatefulSet{}
 	if obj != nil {
-		ss := obj.(*appsv1.StatefulSet)
-		oldSpec = ss.Spec
+		existingStatefulSet = obj.(*appsv1.StatefulSet)
+		if existingStatefulSet.DeletionTimestamp != nil {
+			level.Info(logger).Log(
+				"msg", "halting update of StatefulSet",
+				"reason", "resource has been marked for deletion",
+				"resource_name", existingStatefulSet.GetName(),
+			)
+			return nil
+		}
 	}
 
-	newSSetInputHash, err := createSSetInputHash(*am, c.config, tlsAssets, oldSpec)
+	newSSetInputHash, err := createSSetInputHash(*am, c.config, tlsAssets, existingStatefulSet.Spec)
 	if err != nil {
 		return err
 	}
@@ -782,11 +789,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 
 	ssetClient := c.kclient.AppsV1().StatefulSets(am.Namespace)
 
-	var oldSSetInputHash string
-	if obj != nil {
-		oldSSetInputHash = obj.(*appsv1.StatefulSet).ObjectMeta.Annotations[sSetInputHashName]
-	}
-	if newSSetInputHash == oldSSetInputHash {
+	if newSSetInputHash == existingStatefulSet.ObjectMeta.Annotations[sSetInputHashName] {
 		level.Debug(logger).Log("msg", "new statefulset generation inputs match current, skipping any actions")
 		return nil
 	}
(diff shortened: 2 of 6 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -655,6 +655,9 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 	podLabels := map[string]string{
 		"app.kubernetes.io/version": version.String(),
 	}
+	// In cases where a selector label is removed, new sts cannot match existing pods.
```
simonpasquier, line 658: This isn't restricted to removals but even adding or updating a label would trigger a recreation of the statefulset. I think that it should be reinforced that the statefulset's selector is immutable and any change to the list should be carefully evaluated.
Also can we add the same warning to Alertmanager and Thanos?

fpetkovski (approved):

simonpasquier (approved):
lgtm

PR #4535, opened 2022-01-28
Support the endpointslice role for Prometheus

### LINKED ISSUE #3862: Support the endpointslice role for Prometheus
**What is missing?**

The operator doesn't support the new `endpointslice` role available in Prometheus since v2.21.0 (https://github.com/prometheus/prometheus/pull/6838).

**Why do we need it?**

[Endpoint slices](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) are required to scale services with many pods.

After discussing on Slack with @brancz, the recommended approach would be that the operator uses seamlessly endpoint slices in lieu of endpoints when it detects that the Kube cluster supports them and Prometheus version is >= 2.21.0.

mikechengwei:
## Description

Fixes #3862.
Support the endpointslice role for Prometheus

## Type of change

## Changelog entry

```release-note
Support the endpointslice role for Prometheus
```


simonpasquier (commented):
Thanks for the PR! I wonder if using https://pkg.go.dev/k8s.io/client-go@v0.23.3/discovery#DiscoveryClient wouldn't be cleaner/safer?
Also I assume that we can check once at startup if the API supports endpointslices instead of at every sync.

mikechengwei:
> Thanks for the PR! I wonder if using https://pkg.go.dev/k8s.io/client-go@v0.23.3/discovery#DiscoveryClient wouldn't be cleaner/safer? Also I assume that we can check once at startup if the API supports endpointslices instead of at every sync.

Thanks，i am optimizing it.

mikechengwei:
@simonpasquier  please help me review this pr again,thanks.

pkg/prometheus/operator.go
```diff
@@ -310,6 +311,16 @@ func New(ctx context.Context, conf operator.Config, logger log.Logger, r prometh
 		c.nsPromInf = newNamespaceInformer(c, c.config.Namespaces.PrometheusAllowList)
 	}
 
+	isSupportEndpointSliceResource := false
+	_, err = c.kclient.DiscoveryV1beta1().EndpointSlices(kubeletObjectNamespace).List(context.TODO(), metav1.ListOptions{})
```
fpetkovski, line 315:
```suggestion
	_, err = c.kclient.DiscoveryV1beta1().EndpointSlices(kubeletObjectNamespace).List(ctx, metav1.ListOptions{})
```

pkg/prometheus/operator.go
```diff
@@ -310,6 +311,16 @@ func New(ctx context.Context, conf operator.Config, logger log.Logger, r prometh
...
+	if err != nil {
```
fpetkovski, line 316: Is this intended to automatically detect whether the endpointslice resource is present in the cluster?

pkg/prometheus/promcfg.go
```diff
@@ -320,6 +321,7 @@ func (cg *ConfigGenerator) Generate(
 	additionalAlertRelabelConfigs []byte,
 	additionalAlertManagerConfigs []byte,
 	ruleConfigMapNames []string,
+	isSupportEndpointSliceResource bool,
```
fpetkovski, line 324: A better variable name could be either `endpointSliceSupported` or `isEndpointSliceSupported`, both here and in `operator.go`.

pkg/prometheus/promcfg.go
```diff
@@ -1131,7 +1135,14 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 	}
 
 	selectedNamespaces := getNamespacesFromNamespaceSelector(&m.Spec.NamespaceSelector, m.Namespace, ignoreNamespaceSelectors)
-	cfg = append(cfg, cg.generateK8SSDConfig(selectedNamespaces, apiserverConfig, store, kubernetesSDRoleEndpoint))
+
+	role := kubernetesSDRoleEndpoint
+	if cg.version.GTE(semver.MustParse("2.21.0")) {
+		if isSupportEndpointSliceResource {
```
fpetkovski, line 1141:
```suggestion
	if cg.version.GTE(semver.MustParse("2.21.0")) && isSupportEndpointSliceResource {
```
Could we inline the two conditions to avoid nesting?

pkg/prometheus/promcfg.go
```diff
@@ -1252,17 +1269,26 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 		}
 	}
 
+	var sourceLabelItem yaml.MapItem
+	if isSupportEndpointSliceResource {
+		sourceLabelItem = yaml.MapItem{Key: "source_labels", Value: []string{"__meta_kubernetes_endpointslice_address_target_kind", "__meta_kubernetes_endpointslice_address_target_name"}}
+
+	} else {
+		sourceLabelItem = yaml.MapItem{Key: "source_labels", Value: []string{"__meta_kubernetes_endpoint_address_target_kind", "__meta_kubernetes_endpoint_address_target_name"}}
+
```
fpetkovski, line 1278:
```suggestion
```

pkg/prometheus/promcfg.go
```diff
@@ -1252,17 +1269,26 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 		}
 	}
 
+	var sourceLabelItem yaml.MapItem
+	if isSupportEndpointSliceResource {
+		sourceLabelItem = yaml.MapItem{Key: "source_labels", Value: []string{"__meta_kubernetes_endpointslice_address_target_kind", "__meta_kubernetes_endpointslice_address_target_name"}}
+
```
fpetkovski, line 1275:
```suggestion
```

pkg/prometheus/operator.go
mikechengwei: yes

mikechengwei:
@simonpasquier @fpetkovski please help me review this pr again.

pkg/prometheus/operator.go
```diff
@@ -1629,6 +1635,7 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *mon
 		additionalAlertRelabelConfigs,
 		additionalAlertManagerConfigs,
 		ruleConfigMapNames,
+		c.endpointSliceSupported,
```
simonpasquier, line 1638: The ConfigGenerator struct could have a new `enableEndpointSlices` bool field so we don't have to pass the flag around. Another advantage is that we can also check that the Prometheus version supports the endpointslice role.

pkg/prometheus/promcfg.go
```diff
@@ -1252,17 +1267,24 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 		}
 	}
 
+	var sourceLabelItem yaml.MapItem
+	if endpointSliceSupported {
+		sourceLabelItem = yaml.MapItem{Key: "source_labels", Value: []string{"__meta_kubernetes_endpointslice_address_target_kind", "__meta_kubernetes_endpointslice_address_target_name"}}
+	} else {
+		sourceLabelItem = yaml.MapItem{Key: "source_labels", Value: []string{"__meta_kubernetes_endpoint_address_target_kind", "__meta_kubernetes_endpoint_address_target_name"}}
+	}
```
simonpasquier, line 1275:
```suggestion
	sourceLabels := []string{"__meta_kubernetes_endpoint_address_target_kind", "__meta_kubernetes_endpoint_address_target_name"}
	if endpointSliceSupported {
		sourceLabels = []string{"__meta_kubernetes_endpointslice_address_target_kind", "__meta_kubernetes_endpointslice_address_target_name"}
	}
```

pkg/prometheus/promcfg.go
```diff
@@ -1231,9 +1240,15 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 
 	// Filter targets based on correct port for the endpoint.
 	if ep.Port != "" {
+		var sourceLabelItem yaml.MapItem
+		if endpointSliceSupported {
+			sourceLabelItem = yaml.MapItem{Key: "source_labels", Value: []string{"__meta_kubernetes_endpointslice_port_name"}}
+		} else {
+			sourceLabelItem = yaml.MapItem{Key: "source_labels", Value: []string{"__meta_kubernetes_endpoint_port_name"}}
+		}
```
simonpasquier, line 1248: (nit) a bit less duplication IMO
```suggestion
		sourceLabels := []string{"__meta_kubernetes_endpoint_port_name"}
		if endpointSliceSupported {
		        sourceLabels = []string{"__meta_kubernetes_endpointslice_port_name"}
		}
```

pkg/prometheus/operator.go
```diff
@@ -310,6 +311,12 @@ func New(ctx context.Context, conf operator.Config, logger log.Logger, r prometh
 		c.nsPromInf = newNamespaceInformer(c, c.config.Namespaces.PrometheusAllowList)
 	}
 
+	endpointSliceSupported := false
+	_, err = c.kclient.DiscoveryV1beta1().EndpointSlices(kubeletObjectNamespace).List(ctx, metav1.ListOptions{})
```
simonpasquier, line 315: you could also use [k8s.io/client-go/discovery](https://pkg.go.dev/k8s.io/client-go@v0.23.3/discovery). The advantage is that it would fallback to the "legacy" mode if for some reason the API returned an error (e.g. temporary network hiccup).
It would also be worth having an info log reporting whether the API supports the endpointslices API group or not.

simonpasquier (commented):
Good job!

pkg/prometheus/operator.go
mikechengwei: Ok, I think we can use the discovery serverinfo api to sure if stable endpointslices are supported,ref [doc](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/),WDYT?

mikechengwei:
@simonpasquier plz help review it again,thks.

pkg/prometheus/promcfg.go
```diff
@@ -1249,9 +1257,13 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 
 	// Filter targets based on correct port for the endpoint.
 	if ep.Port != "" {
+		sourceLabels := []string{"__meta_kubernetes_endpoint_port_name"}
```
PhilipGough, line 1260: this will likely need to be addressed in the *deprecated* `TargetPort` block below also

pkg/prometheus/promcfg.go
```diff
@@ -1142,7 +1145,12 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 	}
 
 	selectedNamespaces := getNamespacesFromNamespaceSelector(&m.Spec.NamespaceSelector, m.Namespace, ignoreNamespaceSelectors)
-	cfg = append(cfg, cg.generateK8SSDConfig(selectedNamespaces, apiserverConfig, store, kubernetesSDRoleEndpoint))
+
+	role := kubernetesSDRoleEndpoint
+	if cg.version.GTE(semver.MustParse("2.21.0")) && cg.endpointSliceSupported {
+		role = kubernetesSDRoleEndpointSlice
+	}
```
simonpasquier, line 1152: Part of it could be moved to the ConfigGenerator struct for reuse. Something like:

```golang
func (cg *ConfigGenerator) EndpointSliceSupported() bool {
    return cg.version.GTE(semver.MustParse("2.21.0")) && cg.endpointSliceSupported 
}
```

pkg/prometheus/promcfg.go
```diff
@@ -1249,9 +1257,13 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 
 	// Filter targets based on correct port for the endpoint.
 	if ep.Port != "" {
+		sourceLabels := []string{"__meta_kubernetes_endpoint_port_name"}
+		if cg.endpointSliceSupported {
```
simonpasquier, line 1261: this is also depending on the Prometheus version

pkg/prometheus/promcfg.go
simonpasquier: The other block refers to `__meta_kubernetes_pod_container_port_name` so I'm not sure it needs an update?

pkg/prometheus/promcfg.go
```diff
@@ -1270,17 +1282,22 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 		}
 	}
 
+	sourceLabels := []string{"__meta_kubernetes_endpoint_address_target_kind", "__meta_kubernetes_endpoint_address_target_name"}
+	if cg.endpointSliceSupported {
```
simonpasquier, line 1286: same as above, the Prometheus version should be checked too.

pkg/prometheus/promcfg.go
PhilipGough: yes, you are right, sorry

mikechengwei:
@simonpasquier plz help me review this pr again , thks.

pkg/prometheus/operator.go
```diff
@@ -310,6 +311,11 @@ func New(ctx context.Context, conf operator.Config, logger log.Logger, r prometh
 		c.nsPromInf = newNamespaceInformer(c, c.config.Namespaces.PrometheusAllowList)
 	}
 
+	endpointSliceSupported, err := k8sutil.IsAPIGroupVersionResourceSupported(c.kclient.Discovery(), "discovery.k8s.io", "endpointslices")
+	if err != nil {
+		level.Error(c.logger).Log("msg", "query server whether support endpointslices resource err ", err)
```
simonpasquier, line 316: A warning is better since the program will proceed.

```suggestion
		level.Warn(c.logger).Log("msg", "failed to check if the API supports the endpointslice resources", "err ", err)
```

pkg/prometheus/promcfg.go
```diff
@@ -1270,17 +1286,22 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
...
+	sourceLabels := []string{"__meta_kubernetes_endpoint_address_target_kind", "__meta_kubernetes_endpoint_address_target_name"}
+	if cg.EndpointSliceSupported() {
+		sourceLabels = []string{"__meta_kubernetes_endpointslice_address_target_kind", "__meta_kubernetes_endpointslice_address_target_name"}
+	}
+
 	// Relabel namespace and pod and service labels into proper labels.
 	relabelings = append(relabelings, []yaml.MapSlice{
-		{ // Relabel node labels for pre v2.3 meta labels
-			{Key: "source_labels", Value: []string{"__meta_kubernetes_endpoint_address_target_kind", "__meta_kubernetes_endpoint_address_target_name"}},
+		{
```
simonpasquier, line 1296: could you preserve the comments?

mikechengwei: I have preserve it.

pkg/prometheus/promcfg.go
```diff
@@ -1270,17 +1286,22 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
...
 	}
 
+	sourceLabels := []string{"__meta_kubernetes_endpoint_address_target_kind", "__meta_kubernetes_endpoint_address_target_name"}
+	if cg.EndpointSliceSupported() {
+		sourceLabels = []string{"__meta_kubernetes_endpointslice_address_target_kind", "__meta_kubernetes_endpointslice_address_target_name"}
+	}
+
 	// Relabel namespace and pod and service labels into proper labels.
 	relabelings = append(relabelings, []yaml.MapSlice{
 		{ // Relabel node labels for pre v2.3 meta labels
```
simonpasquier, line 1296: This isn't your fault but the comment is misleading :)
could you fix it?

```suggestion
		{ // Relabel node labels with meta labels available with Prometheus >= v2.3.
```

pkg/prometheus/operator.go
```diff
@@ -310,6 +311,11 @@ func New(ctx context.Context, conf operator.Config, logger log.Logger, r prometh
 		c.nsPromInf = newNamespaceInformer(c, c.config.Namespaces.PrometheusAllowList)
 	}
 
+	endpointSliceSupported, err := k8sutil.IsAPIGroupVersionResourceSupported(c.kclient.Discovery(), "discovery.k8s.io", "endpointslices")
+	if err != nil {
+		level.Warn(c.logger).Log("msg", "failed to check if the API supports the endpointslice resources", "err ", err)
+	}
+	c.endpointSliceSupported = endpointSliceSupported
```
simonpasquier, line 318: I thought that I already asked for it but apparently my comment got lost somewhere. It would be nice to have an info log reporting if the API supports endpointslices.

```suggestion
        level.Info(c.logger).Log("msg", "Kubernetes API capabilities", "endpointslices", endpointSliceSupported)
	c.endpointSliceSupported = endpointSliceSupported
```

simonpasquier (commented):
2 minor comments otherwise LGTM. Thanks!

mikechengwei:
@simonpasquier Thanks for you help , i have fixed the comments.

simonpasquier (approved):
Thanks a lot, this is a very useful addition!

PR #4505, opened 2022-01-19
docs: Update link in adopters file which fixes ci failures

PhilipGough:
We are repeatedly failing link checking in `make check-docs`.

I believe we are being bitten by https://github.com/bwplotka/mdox/issues/22 since polarsignals.com redirects to www.polarsignals.com

Updated the link to avoid the redirect and circumvent bug in tooling.

## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

PR #4536, opened 2022-01-28
Make `mdox` less error-prone

saswatamcode:
## Description

This PR tries to addresses the recent uptick of `mdox` CI failures (e.g, #4532, #4505) by increasing the request timeouts for `mdox` to one minute from the default 10s (as defined in [colly](https://github.com/gocolly/colly/blob/e63893d27834f749d7e2035c29ffe6c75c4c6ad4/colly.go#L980) which is the scraping framework used by mdox) and haven't come across the usual `net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)` error yet.
But other errors such as a TLS handshake timeout do happen from time to time.

Also, removes `ADOPTERS.md` from the list of checked files as content is not modified frequently, but is still error-prone.

Tested this a number of times locally and also on [GitHub Actions in my fork](https://github.com/saswatamcode/prometheus-operator/runs/4981220629?check_suite_focus=true).

## Type of change

## Changelog entry

```release-note
Increase timeout for mdox
```

### DIFF
--- a/.mdox.validate.yaml
+++ b/.mdox.validate.yaml
@@ -1,4 +1,5 @@
 version: 1
+timeout: "1m"
 
 validators:
   # Cloudflare protection, so returns 503 if not in browser. Cannot curl as well.
--- a/Makefile
+++ b/Makefile
@@ -265,7 +265,7 @@ check-golang: $(GOLANGCILINTER_BINARY)
 	$(GOLANGCILINTER_BINARY) run
 
 MDOX_VALIDATE_CONFIG?=.mdox.validate.yaml
-MD_FILES_TO_FORMAT=$(filter-out $(FULLY_GENERATED_DOCS), $(shell find Documentation -name "*.md")) $(shell ls *.md)
+MD_FILES_TO_FORMAT=$(filter-out $(FULLY_GENERATED_DOCS), $(shell find Documentation -name "*.md")) $(filter-out ADOPTERS.md, $(shell ls *.md))
 
 .PHONY: docs
 docs: $(MDOX_BINARY)


simonpasquier:
Thanks for looking into this! Unfortunately it doesn't seem to help, maybe there's something else at play?
```
 error: 		ADOPTERS.md:174: "https://www.polarsignals.com/" not accessible even after retry; status code 0: Get "https://www.polarsignals.com/": net/http: TLS handshake timeout
```

saswatamcode:
Nearly, all the errors in prometheus-operator CI seem to be `Client.Timeout` errors the last few weeks. The handshake one seems to be less common, but is still there (only got it twice out of 11 runs that I tested on fork). 

Oddly, many of these failures started a few days ago and are usually adopter homepage links. A few reasons that I could think of were protections on these websites reacting to automated requests, or some sort of GitHub Actions issue with networks which isn't uncommon (e.g, https://github.com/actions/virtual-environments/issues/4896). 🤔

For the time being, maybe we can skip checking the adopters markdown file along with increasing timeouts? As that isn't a very frequently touched file but many errors seem to be from the links there? 🙂

PhilipGough:
@saswatamcode agree with your analysis. Skipping the adopters.md file seems like a reasonable solution to me because:
1. The content is pretty static as you mentioned
2. If a link is *really* broken there, it has no real impact on the core documentation and hence virtually no impact on the end user.

I'd prefer to skip it and get our CI green.

simonpasquier:
I agree that skipping the link verification for ADOPTERS.md makes sense.

saswatamcode:
Will update PR and change the title as well!

PR #4398, opened 2021-11-12
Generate API reference docs with gen-crd-api-reference-docs

PhilipGough:
## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

PoC for using https://github.com/ahmetb/gen-crd-api-reference-docs as API reference documentation.
This seems to work well, and the resulting file for v1 is much smaller than with `crdoc` tool in #4396

```bash
 ls -lh Documentation/apis/v1/api.md                                                                                                                    
-rw-r--r--  1 pgough  staff   207K Nov 12 10:58 Documentation/apis/v1/api.md
```

Related to #4376

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #4539, opened 2022-01-31
Extract common Prometheus fields in a separate struct

fpetkovski:
## Description

In preparation for adding the Prometheus Agent CRD, this commit extracts
fields common to the Prometheus server and agent into a dedicated struct
that can be embedded into the two CRDs. 

The commit also refactors `po-docgen` and extracts the logic for parsing
structs and fields into a dedicate module. Furthermore, It improves
`po-docgen` to hide types which are only used as embedded, and merge
their fields into the types which extend them. This change makes sure that 
the newly introduced `CommonPrometheusFields` type is not shown in the docs, 
and its fields are flattened into the Prometheus type.

## Type of change

## Changelog entry

```release-note

```


simonpasquier:
This looks good to me. Maybe @PhilipGough could chime in about doc generation since he looked at potential replacements for po-docgen (https://github.com/prometheus-operator/prometheus-operator/issues/4376) and the potential impact?

PhilipGough:
@simonpasquier - thanks for giving me a kick to revisit https://github.com/prometheus-operator/prometheus-operator/pull/4398 :) 

TL;DR - I evaluated a couple of tools - https://github.com/ahmetb/gen-crd-api-reference-docs looks like the best of them (I think @fpetkovski might have some prior experience with this tool already?)

In #4398 I cherry-picked over this commit and regenerated the docs with the new "would-be" tooling to get an idea as to what it would look like by default.

In that case, the `CommonFields` is displayed in doc and actually with the way I have split the reference docs for the APIs there (v1, v1alpha1) into individual markdown files, that is actually beneficial because we want internal links across those files ( v1alpha1 -> v1)

I don't see an issue with what is generated, I actually prefer it but YMMV.

It probably would be good to get feedback on the other tool and decide if we are going to move forward with it before deciding to make these changes to our current docs tool.

fpetkovski:
Looking at the generated docs from the third-party tool, I am not too excited about the way embedded fields are shown. They're linked to a different section instead of having their fields merged into the parent struct. Although not a blocker, for me this makes it a slightly to read the docs and also exposes internals in the documentation. Nevertheless, I understand the benefits of using a third party tool and removing the burden of maintaining our own docs generation machinery. So if everyone is fine with the resulting format, we can adopt the new tool and get rid of po-docgen.

My concern with doing deep(er) evaluation of documentation tools as part of this PR is that we can easily end up with scope creep and add additional delays to the prometheus agent work. We can always replace our docs tooling at a later point, or even work on the investigation in parallel. So to summarize, my preference would be to treat the agent work and documentation investigation as independent work streams.

PhilipGough:
@fpetkovski - your concerns sound reasonable to me and it makes sense that it should not block this work based on that IMO. 

Bear in mind, the output in the linked PR is what we get by default. It does rely on templates however and we may find a way to output docs in the preferred format in regards to embedded fields.

simonpasquier:
> So to summarize, my preference would be to treat the agent work and documentation investigation as independent work streams.

I'm +1 on your suggestion. I just wanted to make sure that we're all on the same page and we don't end up maintaining po-docgen forever because we've discovered after the fact that the planned successor couldn't work with the new code reorg.

fpetkovski:
@PhilipGough 
> Bear in mind, the output in the linked PR is what we get by default. It does rely on templates however and we may find a way to output docs in the preferred format in regards to embedded fields.

Yeah that makes complete sense. We can also make contributions if we lack a particular feature.

@simonpasquier Yes that's a great point. From my perspective, this specific third party tools produces docs that are satisfactory. The few issues I noticed are in no way a blocker given the benefits we get by outsourcing docs generation.

fpetkovski:
@simonpasquier @PhilipGough this has been rebased against main and is ready for review. Since we're making a change to the library, should we reflect this in the changelog entry?

PR #4552, opened 2022-02-02
Add securityContext to managed containers

ArthurSens:
## Description

Following what has been done in https://github.com/prometheus-operator/kube-prometheus/pull/1610, https://github.com/prometheus-operator/kube-prometheus/pull/1600 and https://github.com/prometheus-operator/kube-prometheus/pull/1593. This PR also sets the same securityContext to all managed containers (I believe I haven't missed any 😅)

Prometheus statefulset still doesn't have `readonlyRootFilesystem: true`, because of https://github.com/prometheus-operator/prometheus-operator/issues/4562

## Type of change

## Changelog entry

```release-note
Improve security with container Security Context
```

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -538,6 +538,8 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 	var watchedDirectories []string
 	watchedDirectories = append(watchedDirectories, reloadWatchDirs...)
 
+	boolFalse := false
+	boolTrue := true
 	defaultContainers := []v1.Container{
 		{
 			Args:           amArgs,
@@ -548,6 +550,13 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 			LivenessProbe:  livenessProbe,
 			ReadinessProbe: readinessProbe,
 			Resources:      a.Spec.Resources,
+			SecurityContext: &v1.SecurityContext{
+				AllowPrivilegeEscalation: &boolFalse,
+				ReadOnlyRootFilesystem:   &boolTrue,
+				Capabilities: &v1.Capabilities{
+					Drop: []v1.Capability{"ALL"},
+				},
+			},
 			Env: []v1.EnvVar{
 				{
 					// Necessary for '--cluster.listen-address' flag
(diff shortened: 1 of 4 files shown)


simonpasquier:
I think you missed Thanos Ruler? I'll try to run it against the OpenShift CI to see if it triggers any issue there.

simonpasquier:
https://github.com/openshift/prometheus-operator/pull/156 fyi

PhilipGough:
cc @ArthurSens - PTAL - https://github.com/openshift/prometheus-operator/pull/156#issuecomment-1030079286

simonpasquier:
It looks like the OpenShift e2e tests reveal an interesting edge case :)
In one of the tests, we configure the Prometheus object with `spec.queryLogFile: /tmp/test.log` which fails now that the root FS is read-only. I would argue that we should add an emptyDir volume mounted under `/tmp` but it might hit other users too.
I'm wondering if there are other ways that could trigger a similar problem. At least we need to emphasize the change and document it carefully. WDYT?

ArthurSens:
I nice catch by openshift tests!

> I would argue that we should add an emptyDir volume mounted under /tmp but it might hit other users too.

I'm not sure I understand how adding an emptyDir could affect other users, I believe this should be the way to move forward here 🤔

simonpasquier:
sorry I wasn't clear. I meant that we could solve it in OpenShift with the monitoring operator (managing the Prometheus object) adding an emptyDir volume. But regardless of the OpenShift case, upstream users that specify `queryLogFile` today might be affected once/if we merge this PR.

ArthurSens:
Would it solve if I add an emptyDir volume to this PR?

PhilipGough:
> Would it be solved if I add an emptyDir volume to this PR?

I am not sure it would, since that would depend on where they write the log and where we would mount the dir. IIUC this could be a breaking change in that respect.

ArthurSens:
Alright yeah makes sense

I'll remove the readonlyRootFilesystem from this PR and file an issue to remind us to tackle this later

ArthurSens:
Issue created: https://github.com/prometheus-operator/prometheus-operator/issues/4562

simonpasquier (commented):
LGTM, maybe add a FIXME comment to the Prometheus statefulset's SecurityContext referring to https://github.com/prometheus-operator/prometheus-operator/issues/4562 so that we don't forget in the future and add it by accident?

PhilipGough (approved):
lgtm

Yes, we can merge this first and follow up at that point.

PR #4554, opened 2022-02-03
pkg/prometheus: add relabel_configs and metric_relabel_configs

ghost:
## Description

This PR adds the possibility to configure relabeling for all metrics and scrape targets in the Prometheus spec.
This allows to easily reuse some  general relabeling rules for an entire Prometheus instance.

We are using `Secret` references in the configuration, exactly like `AdditionalScrapeConfigs`, the `Secrets` get appended to the existing configuration without further validation.
The following note from the PR #1246 that added the `AdditionalScrapeConfigs` field, holds for this feature as well:

> As scrape configs are appended, the user is responsible to make sure it is
> valid. Note that using this feature may expose the possibility to break upgrades
> of Prometheus. It is advised to review Prometheus release notes to ensure that
> no incompatible scrape configs are going to break Prometheus after the upgrade.

## Type of change

## Changelog entry

```release-note
Add relabel_configs and metric_relabel_configs fields to the Prometheus CRD for configuring general relabeling rules.
```


simonpasquier:
I'm not convinced that it's something we should do. In particular, it might be difficult for users to understand the discrepancy between their service monitor definition and what's returned by Prometheus eventually. If I were to implement something like this, I'd probably look after a mutating webhook.
Can you provide more details about your use case?

karl-pip:
We are hosting a internal Kubernetes as a service platform, running hundreds of Kubernetes clusters on premise.

We got hundreds of Prometheus custom resources in one cluster created by another operator. Each Prometheus pod scrapes platform metrics (not application specific ones) from remote customer clusters. Every Prometheus sends metrics to a central Thanos longterm metric store. We need a common set of labels on every metric from every discovered scrape target to properly work with longterm data. This is even more important with alert stack where alert routing depends on these common labels.

Such labels involve:
- cluster id to find affected clusters in alerts
- geo region and availability zone to derive data center local issues from metrics
- platform release version to find out which release introduced issues
- release stage to prioritize during outages

Having those common labels on metrics heavily simplifies those queries.

karl-pip:
We are open for other solutions, we just thought that talking about ideas as code is easier than just opening an issue.

> If I were to implement something like this, I'd probably look after a mutating webhook.

Our scrape configs are configured in secret values, these secrets are reconciled by another operator. We are concerned that changing these values by a mutating webhook will lead the operator to revert these changes.

simonpasquier:
Thanks for the details, it's definitely helpful! Let me ask a few questions :-)

> We got hundreds of Prometheus custom resources in one cluster created by another operator. Each Prometheus pod scrapes platform metrics (not application specific ones) from remote customer clusters. Every Prometheus sends metrics to a central Thanos longterm metric store. We need a common set of labels on every metric from every discovered scrape target to properly work with longterm data. This is even more important with alert stack where alert routing depends on these common labels.

IIUC you run many Prometheus instances in the internal Kubernetes cluster. Each instance scrapes platform metrics for a specific remote on-premises cluster and forwards the metrics to a central Thanos. You want every metric/alert coming from these Prometheus instances to carry the information about the "remote" cluster. Is this correct?
Have you evaluated external labels? It looks like a viable approach but I may miss some context.

> Our scrape configs are configured in secret values, these secrets are reconciled by another operator. 

Does it mean that you use only `additionalScrapeConfigs` and no service/pod monitors?

> We are concerned that changing these values by a mutating webhook will lead the operator to revert these changes.

Which operator? human or your own operator working on top of the prometheus operator?

lauraseidler:
Since the original author doesn't seem to exist anymore, I assume this won't move forward, but I wanted to add my two cents anyways, in case someone else stumbles upon this, as we have this use case too.

We're a platform team managing one `Prometheus` instance per cluster, which is then used by all teams, who can deploy `ServiceMonitor`s on their own if needed. However, we also have some "global" services that are managed by us - think sealed-secrets, cert-manager, kube-state-metrics etc. We have configured metrics ingestion for those via `ServiceMonitor`s.

However, we sometimes want to skip metrics ingestion for specific namespaces all together, for example because they are used for temporary environments, and we have no use for the metrics. Now, we _can_ (and do) throw a `metricRelabelings` config on all the `ServiceMonitor`s, but it would be much nicer to do this on the `Prometheus` instance itself, so that we don't forget to do it if we deploy a new global service.

lauraseidler:
@simonpasquier yeah, that would work. Though tbh - our infrastructure is not that complex. Our current config matches namespaces like `(.*)-temp-(.*)` - this will never be relevant for user deployed service monitors, so we'd be perfectly fine with just applying this globally without further filtering.

JoaoBraveCoding:
Disclaimer: I'm not a maintainer

> However, we sometimes want to skip metrics ingestion for specific namespaces all together, for example because they are used for temporary environments, and we have no use for the metrics. Now, we can (and do) throw a metricRelabelings config on all the ServiceMonitors, but it would be much nicer to do this on the Prometheus instance itself, so that we don't forget to do it if we deploy a new global service.

IIUC @lauraseidler goal is to skip the ingestion of metrics altogether for specific projects. Couldn't we achieve this with the `matchExpression` field in the `NamespaceSelector`?
This way Prometheus would always pick up monitors ([Probe/Pod/Service]Monitors) from a namespace unless they had a specific label (in the example I'm giving `k8s.io/monitor: false`) 

```yaml
    matchExpressions:
    - key: k8s.io/monitor
      operator: NotIn
      values:
      - "false"
```

Not against the feature, just brainstorming/trying to help :)

sathieu:
We need this also.

We have tenants each being an aggregation of namespaces.

This would look like this:

```yaml
    - sourceLabels: [namespace]
      targetLabel: tenant
      regex: (team1-.*)
      replacement: team1
      action: replace
    - sourceLabels: [namespace]
      targetLabel: tenant
      regex: (team2-.*)
      replacement: team2
      action: replace
```

PR #4555, opened 2022-02-03
*: mark remoteWrite and remoteRead as non-experimental

simonpasquier:
## Description

The remote read and write features are established enough in Prometheus
that we can drop the "EXPERIMENTAL" warning in the Prometheus CRD and
support them officially from now on.

If the format of the remote write/read configuration in Prometheus
changes in a breaking way, the operator should deal with it based on the
`spec.version` field of the Prometheus resource, just like it does for
other configuration fields.

## Type of change

## Changelog entry

```release-note
`remoteWrite` and `remoteRead` fields of the Prometheus CRD not considered experimental anymore.
```

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -361,15 +361,15 @@ Endpoint defines a scrapeable endpoint serving Prometheus metrics.
 
 ## MetadataConfig
 
-Configures the sending of series metadata to remote storage.
+MetadataConfig configures the sending of series metadata to the remote storage.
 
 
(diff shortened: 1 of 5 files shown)


simonpasquier:
cc @paulfantom @fpetkovski @PhilipGough

PhilipGough (approved):
lgtm

and thanks for the additional cleanup :)

simonpasquier:
>Worth adding a note to the changelog? I know we haven't added or removed anything but maybe officially moving out of experimental is worth calling out?

makes sense

Issue #4557, opened 2022-02-04
Add followredirect to Endpoint configuration so that it can be disabled.


primeroz:
**What is missing?**
Prometheus support an option to disable Redirect on the scrape endpoints 

```
# Configure whether scrape requests follow HTTP 3xx redirects.
[ follow_redirects: <bool> | default = true ]
```

but is not exposed through the endpoint spec

**Why do we need it?**

When scraping VAULT, for example, the standby nodes do not return metrics but instead return a redirect to the Public endpoint

**Environment**
Kubernetes 

* Prometheus Operator version: 0.53

    `Insert image tag or Git SHA here`
    
    

**Anything else we need to know?**:

PR #4563, opened 2022-02-07
pkg/promcfg: Add support for follow_redirects in endpoints

primeroz:
## Description

Add support for setting `follow_redirects` in endpoint configuration.

Since the prometheus default value is `true` only set the value when is set to `false`

Closes https://github.com/prometheus-operator/prometheus-operator/issues/4557

## Type of change

- [ x ] `FEATURE` (non-breaking change which adds functionality)

## Changelog entry

```release-note
- Add support for follow_redirects in endpoint scrape configuration. Prometheus default value is true so only set when the requested value is false

```

### DIFF
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -965,6 +965,8 @@ type Endpoint struct {
 	RelabelConfigs []*RelabelConfig `json:"relabelings,omitempty"`
 	// ProxyURL eg http://proxyserver:2195 Directs scrapes to proxy through this endpoint.
 	ProxyURL *string `json:"proxyUrl,omitempty"`
+	// FollowRedirects Configure whether scrape requests follow HTTP 3xx redirects.
+	FollowRedirects bool `json:"followRedirects,omitempty"`
 }
 
 // PodMonitor defines monitoring for a set of pods.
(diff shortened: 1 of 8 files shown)


pkg/apis/monitoring/v1/types.go
```diff
@@ -1049,6 +1051,8 @@ type PodMetricsEndpoint struct {
 	RelabelConfigs []*RelabelConfig `json:"relabelings,omitempty"`
 	// ProxyURL eg http://proxyserver:2195 Directs scrapes to proxy through this endpoint.
 	ProxyURL *string `json:"proxyUrl,omitempty"`
+	// FollowRedirects Configure whether scrape requests follow HTTP 3xx redirects.
```
simonpasquier, line 1054: ditto

pkg/prometheus/promcfg.go
```diff
@@ -657,6 +661,10 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	if ep.Scheme != "" {
 		cfg = append(cfg, yaml.MapItem{Key: "scheme", Value: ep.Scheme})
 	}
+	// Defaults to true, so only add when value is set to false
+	if ep.FollowRedirects != nil && !ptrToBool(ep.FollowRedirects) {
+		cfg = append(cfg, yaml.MapItem{Key: "follow_redirects", Value: false})
+	}
```
simonpasquier, line 667: I would only check if the value isn't nil and if so, pass the value directly. What do you think?

```suggestion
	if ep.FollowRedirects != nil {
		cfg = append(cfg, yaml.MapItem{Key: "follow_redirects", Value: *ep.FollowRedirects})
	}
```

primeroz: i have some local changes to commit ( basically the tests ) but i can add that. 

I did not add it because the value `true` is the default but i guess better to be explicit

primeroz: done

simonpasquier (approved):
thanks!

PR #4566, opened 2022-02-07
pkg/prometheus/statefulset: Add VolumeMount for query-log-file

### LINKED ISSUE #4562: QueryLog should create a volumeMount
**What is missing?**
To follow more strict security best practices, the Prometheus container should have the security context `readonlyRootFilesystem` set to `true`.

Today that would be a breaking change to the feature QueryLog, where users are able to write down queries performed by Prometheus to a path chosen by the user.

To be able to enable `readonlyRootFilesystem: true`, when setting QueryLog in the Prometheus spec, the operator needs to create a volume mount to the specified path.

See also: https://hub.armo.cloud/docs/c-0017

ArthurSens:
Signed-off-by: ArthurSens <arthursens2005@gmail.com>

## Description

Creates a volumeMount to Prometheus container when QueryLogFile is set.

This is an effort to follow security best practices, allowing us to enable `readonlyRootFilesystem` to Prometheus' container.

Fixes #4562 

## Type of change

## Changelog entry

```release-note
Follow security best practices by adding a volumeMount when QueryLogFile is set
```


pkg/prometheus/statefulset.go
```diff
@@ -548,6 +557,14 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 		})
 	}
 
+	if p.Spec.QueryLogFile != "" {
+		promVolumeMounts = append(promVolumeMounts, v1.VolumeMount{
+			Name:      "query-log-file",
+			ReadOnly:  false,
+			MountPath: p.Spec.QueryLogFile,
```
simonpasquier, line 564: it should the `filepath.Dir()` of the query log file

simonpasquier (commented):
After more thinking, I found a few edge cases. You could set `queryLogFile: /prometheus/query.log` to store the query logs on the persistent volume or `queryLogFile: /dev/stdout` to stream the logs to standard output. In both cases, you don't want/need the emptyDir.
But there might be more possibilites?

ArthurSens:
Alright, I believe I've changed the PR, code and description, to match what we discussed during the contributor office hours today...

How can we make sure users get informed about the change?

And I believe this is a breaking change for openshift, isn't it?

simonpasquier (commented):
> How can we make sure users get informed about the change?

I think that it would be worth updating the `queryLogFile` description to explain that if the filename has an empty path, the operator adds automatically an emptyDir volume mounted at `/to_be_defined`. If the user provides a full path, it should be writable and knowing that the root filesystem is read-only, they probably have to configure additional volume/volumeMount.

https://github.com/prometheus-operator/prometheus-operator/blob/9a07ea21084952f3188d1d988b34aa8654e3c44d/pkg/apis/monitoring/v1/types.go#L364-L370

> And I believe this is a breaking change for openshift, isn't it?

We'll deal with it at the [cluster monitoring operator](https://github.com/openshift/cluster-monitoring-operator/) level.

pkg/apis/monitoring/v1/types.go
```diff
@@ -362,7 +362,10 @@ type PrometheusSpec struct {
 	// Make sure both ruleNamespace and ruleName are set for each pair
 	PrometheusRulesExcludedFromEnforce []PrometheusRuleExcludeConfig `json:"prometheusRulesExcludedFromEnforce,omitempty"`
 	// QueryLogFile specifies the file to which PromQL queries are logged.
-	// Note that this location must be writable, and can be persisted using an attached volume.
+	// If the filename has an empty path, e.g. 'query.log', prometheus-operator will will mount the file into an
+	// emptyDir volume. If a full path is provided, e.g. /var/log/prometheus/query.log, you must mount a volume 
+	// in the specified directory and it must be writable. The reason is beacuse the prometheus container is created
+	// with 'readonlyRootFilesystem: true', aiming to tighten security.
```
simonpasquier, line 368:
```suggestion
	// in the specified directory and it must be writable. The reason being that the prometheus container runs with a read-only root filesystem for security reasons.
```

ArthurSens:
Thanks for all the reviews so far simon! I believe it ready to remove from draft :)

Documentation/api.md
```diff
@@ -783,7 +783,7 @@ PrometheusSpec is a specification of the desired behavior of the Prometheus clus
...
+| queryLogFile | QueryLogFile specifies the file to which PromQL queries are logged. If the filename has an empty path, e.g. 'query.log', prometheus-operator will mount the file into an emptyDir volume at `/var/log/prometheus`. If a full path is provided, e.g. /var/log/prometheus/query.log, you must mount a volume in the specified directory and it must be writable. The reason being that the prometheus container runs with a read-only root filesystem for security reasons. Alternatively, the location can be set to a stdout location such as `/dev/stdout` to log query information to the default Prometheus log stream. This is only available in versions of Prometheus >= 2.16.0. For more details, see the Prometheus docs (https://prometheus.io/docs/guides/query-log/) | string | false |
```
simonpasquier, line 786: Ah I introduced a repetition in my suggestion :)
```suggestion
| queryLogFile | QueryLogFile specifies the file to which PromQL queries are logged. If the filename has an empty path, e.g. 'query.log', prometheus-operator will mount the file into an emptyDir volume at `/var/log/prometheus`. If a full path is provided, e.g. /var/log/prometheus/query.log, you must mount a volume in the specified directory and it must be writable. This is because the prometheus container runs with a read-only root filesystem for security reasons. Alternatively, the location can be set to a stdout location such as `/dev/stdout` to log query information to the default Prometheus log stream. This is only available in versions of Prometheus >= 2.16.0. For more details, see the Prometheus docs (https://prometheus.io/docs/guides/query-log/) | string | false |
```

ArthurSens:
Merge conflicts resolved 🙂

simonpasquier (approved):
lgtm

Issue #4572, opened 2022-02-11
The fix of #4429 seems to break existing labelMap-actions


simonpasquier:
The fix of #4429 seems to break existing labelMap-actions. With v0.54 using kube-stack we are losing some PodMonitors (in this case from our Pulsar Cluster.

Log from the operator
```
level=warn ts=2022-02-10T21:36:06.190554556Z caller=operator.go:1957 component=prometheusoperator msg="skipping podmonitor" error="\"\" is invalid 'replacement' for labelmap action" podmonitor=pulsar/pulsar-zookeeper namespace=observability prometheus=prometheus-community-kube-prometheus
level=warn ts=2022-02-10T21:36:06.190682758Z caller=operator.go:1957 component=prometheusoperator msg="skipping podmonitor" error="\"\" is invalid 'replacement' for labelmap action" podmonitor=pulsar/pulsar-broker namespace=observability prometheus=prometheus-community-kube-prometheus
```

From looking at the podMonitors
```
...
    relabelings:
        - action: labelmap
          regex: __meta_kubernetes_pod_label_(.+)
```

this should be the problematic action. 

Source definition: https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/templates/autorecovery-podmonitor.yaml#L41

Workaround seems to be to revert to v0.53.

_Originally posted by @mkoertgen in https://github.com/prometheus-operator/prometheus-operator/issues/3942#issuecomment-1035552992_

simonpasquier:
@slashpai if you want to have a look :)

slashpai:
We can close this since it's fixed now

PR #4429, opened 2021-12-01
pkg/prometheus: Add validations for relabel configs

slashpai:
Fixes #2503, #3942

Signed-off-by: Jayapriya Pai <slashpai9@gmail.com>

## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

This PR is to add validation for relabelling configs

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [x] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Added validations for relabel configs
```

Issue #4524, opened 2022-01-26
Improve documentation and move possible validations via OpenAPI


slashpai:
**What is missing?**

While working on adding [validation for scrapeTimeout](https://github.com/prometheus-operator/prometheus-operator/pull/4491#discussion_r782103988), we noted that its a bit confusing and documentation is not clear how default values are set for CRD fields. We need to improve documentation on how default values are set and move validations possible using [OpenAPI](https://book.kubebuilder.io/reference/markers/crd-validation.html)

**Why do we need it?**

This is needed for better understanding and troubleshooting incase of issues

**Environment**

NA

* Prometheus Operator version:

    `Insert image tag or Git SHA here`
    
    

0.53.1

**Anything else we need to know?**:

See discussion in the PR #4491

slashpai:
@simonpasquier I think we have covered possible options in PRs linked, good to close?

simonpasquier:
Yep, I think we're good.

PR #4574, opened 2022-02-14
pkg/prometheus: Update validation for labelmap

slashpai:
## Description

pkg/apis: Set default values for relabelConfigs

We had implemented validation logic for relabelling as in [prometheus](https://github.com/prometheus/prometheus/blob/9a2e93228e63a3fbac950ef0a639f754c3598c5d/model/relabel/relabel.go#L94-L133). Since default values for relabel configs are set [here](https://github.com/prometheus/prometheus/blob/9a2e93228e63a3fbac950ef0a639f754c3598c5d/model/relabel/relabel.go#L31-L36) in prometheus it works well but since default values in operator for these configs are not set explicitly [validation logic](https://github.com/prometheus-operator/prometheus-operator/pull/4429) implemented for relabelling will fail ~~hence handling it at CRD level so that default values are set correctly and validation will not throw error when these values are not set in config.~~

As a fix updating the validation method logic instead of api level as per discussion in this PR

Fixes https://github.com/prometheus-operator/prometheus-operator/issues/4572

## Type of change

## Changelog entry

```release-note
Fixed relabelConfigs for labelmap action
```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -2170,8 +2170,10 @@ func validateRelabelConfig(rc monitoringv1.RelabelConfig) error {
 	if rc.Action == string(relabel.Replace) && !relabelTarget.MatchString(rc.TargetLabel) {
 		return errors.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
 	}
-	if rc.Action == string(relabel.LabelMap) && !relabelTarget.MatchString(rc.Replacement) {
-		return errors.Errorf("%q is invalid 'replacement' for %s action", rc.Replacement, rc.Action)
+	if rc.Action == string(relabel.LabelMap) {
+		if rc.Replacement != "" && !relabelTarget.MatchString(rc.Replacement) {
+			return errors.Errorf("%q is invalid 'replacement' for %s action", rc.Replacement, rc.Action)
+		}
 	}
 	if rc.Action == string(relabel.HashMod) && !model.LabelName(rc.TargetLabel).IsValid() {
 		return errors.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
--- a/pkg/prometheus/operator_test.go
+++ b/pkg/prometheus/operator_test.go
@@ -322,18 +322,27 @@ func TestValidateRelabelConfig(t *testing.T) {
 		{
 			scenario: "invalid labelmap config",
 			relabelConfig: monitoringv1.RelabelConfig{
-				Action: "labelmap",
-				Regex:  "__meta_kubernetes_service_label_(.+)",
+				Action:      "labelmap",
+				Regex:       "__meta_kubernetes_service_label_(.+)",
+				Replacement: "some-name-value",
 			},
 			expectedErr: true,
 		},
-		// Test valid labelmap relabel config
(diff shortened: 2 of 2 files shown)


slashpai:
cc: @simonpasquier @PhilipGough

simonpasquier (commented):
For this one, I'm on the fence whether we should enforce defaults at the API level or modify the validation code. Adding defaults at the API level will change the generated configuration and while in theory it shouldn't make a difference, I'm worried that it might break existing users (I don't have a concrete example though).

PhilipGough:
IIUC , right now, we don't currently set the default in the operator codebase, we just delegate to Prometheus? 

In that case I agree we don't really need to add this default here, although I also don't see the change breaking anything (not to say I haven't overlooked something however).

What I would think is worth considering however, in cases where we are delegating to Prometheus, perhaps we should remove the "default" value from our docs?

simonpasquier:
> In that case I agree we don't really need to add this default here, although I also don't see the change breaking anything (not to say I haven't overlooked something however).

Like I said earlier, I don't have a concrete example so my fear may be unfounded.

> What I would think is worth considering however, in cases where we are delegating to Prometheus, perhaps we should remove the "default" value from our docs?

100% agree

slashpai:
> IIUC , right now, we don't currently set the default in the operator codebase, we just delegate to Prometheus? 
> 

Yes some of the fields which says default value is set by Prometheus that's a bit confusing though so like you said we should remove those from docs

If we don't want to handle this at API level and handle it at validation method we should just ignore the check if `replacement` is unset.

I will also try to test if this breaks the existing config because I have another issue #4524 to set defaults and validations at API level. So if we do that that could also be a problem for existing users?

slashpai:
> I think the idea would be to evaluate any of the existing validations in the code base as a candidate for a conversion to OpenAPI validation - that would be a like for like change as opposed to an addition which is what was proposed here and should negate any concern

We had also discussed about moving default values to API level as well in [PR for Add validation for ScrapeTimeout](https://github.com/prometheus-operator/prometheus-operator/pull/4491#discussion_r782103988). From what I understand that would be a configuration change as well like here

slashpai:
I removed the default value set at API level and updated the condition in validation function to fix the bug for labelmap

simonpasquier:
@slashpai can you submit the PR against the release-0.53? I think this deserves a patch release given it breaks existing setups. We can revisit whether we should move the defaults to the API spec when you tackle https://github.com/prometheus-operator/prometheus-operator/issues/4524.

cc @fpetkovski

slashpai:
Updated the PR to change base branch to release-0.54 for patch release

simonpasquier (commented):
LGTM on green

fpetkovski:
Should we make the same PR against main?

slashpai:
Ya I can add one to main also

simonpasquier:
the workflow is to merge back release-0.54 into main.

simonpasquier:
I'll defer to @fpetkovski for merging this one since he's the 0.54 release shepherd :)

slashpai:
@simonpasquier We have another PR to handle default values [#4579](https://github.com/prometheus-operator/prometheus-operator/pull/4579). 

May be we should create an upgrade test so we can catch these issues earlier :)

PR #4579, opened 2022-02-16
pkg/prometheus: support default relabel config in validation

0robustus1:
## Description

Prometheus has a default relabel config that is different
from the natural zero values for the individual types.

The current validations create a situation where a valid configuration
that is marshaled and unmarshaled (e.g. when generating a
ServiceMonitor config) is suddenly no longer valid.

This changes the validation to accept both the zero values (as that
makes the most sense when writing the config by hand) and the
default config as it is used by Prometheus (see also
https://github.com/prometheus/prometheus/blob/v2.33.3/model/relabel/relabel.go#L122-L130).

## Type of change

## Changelog entry

```release-note
Updated relabelConfig validation to accept Prometheus default config on labeldrop relabelConfig.
```

### DIFF
--- a/pkg/prometheus/operator_test.go
+++ b/pkg/prometheus/operator_test.go
@@ -21,6 +21,7 @@ import (
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
+	"github.com/prometheus/prometheus/model/relabel"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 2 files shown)


fpetkovski (approved):
Thanks!

metalmatze (approved):
Nice

slashpai:
@simonpasquier  @fpetkovski  I think we should include this change as well in 0.54.1

simonpasquier:
@slashpai I agree that #4574 and #4579 should be merged in the same `release-0.54` branch.
@fpetkovski WDYT?

fpetkovski:
Makes sense. @0robustus1 could you change the target branch of the PR to `release-0.54`?

pkg/prometheus/operator_test.go
```diff
@@ -267,6 +268,20 @@ func TestValidateRemoteWriteConfig(t *testing.T) {
...
+	defaultRegexp, err := relabel.DefaultRelabelConfig.Regex.MarshalYAML()
+	if err != nil {
+		t.Errorf("Could not marshal relabel.DefaultRelabelConfig.Regex: %v", err)
+	}
+	defaultRegex, ok := defaultRegexp.(string)
+	if !ok {
+		t.Errorf("Could not assert marshaled defaultRegexp as string: %v", defaultRegexp)
+	}
+
+	defaultSourceLabels := []monitoringv1.LabelName{}
```
slashpai, line 280: LabelName was introduced in this [PR](https://github.com/prometheus-operator/prometheus-operator/pull/4568) it is not yet in 0.54 branch so you would need to use `[]string` type here

fpetkovski (approved):
Thanks!

PR #4587, opened 2022-02-21
Support probe shards

### LINKED ISSUE #4126: Probe Sharding
**What is missing?**

I am missing the ability to shard metrics from Probes. 

**Why do we need it?**

We are using an instance of prometheus-operater with 2 shards and our Probe scrapes are getting duplicated. I see that this has been handled for PodMetrics and ServiceMontiors. 

**Environment**

I am willing to do the work but would like some guidance on how to approach it. Since there is already a function to generate the shard relabel configs, I see 2 possibilites:

1. Apply the config as it is normally applied with `__address__` and the user will need to add a `__address__` label
2. Extend the function and the Probe CRD to have a shardLabel field that is passed to the shard function instead of `__address__`

mikechengwei:
## Description

Fix #4126 

## Type of change

## Changelog entry

```release-note

```


pkg/prometheus/promcfg.go
```diff
@@ -1405,9 +1408,13 @@ func getLimit(user uint64, enforced *uint64) uint64 {
 	return user
 }
 
-func generateAddressShardingRelabelingRules(relabelings []yaml.MapSlice, shards int32) []yaml.MapSlice {
+func generateAddressShardingRelabelingRules(relabelings []yaml.MapSlice, shards int32, isProbe bool) []yaml.MapSlice {
```
simonpasquier, line 1411: It might be easier to use different function names

```suggestion
func generateAddressShardingRelabelingRules(relabelings []yaml.MapSlice, shards int32) []yaml.MapSlice {
   return generateAddressShardingRelabelingRulesWithSourceLabel(relabelings, shards, "__address__")
}

func generateAddressShardingRelabelingRulesForProbes(relabelings []yaml.MapSlice, shards int32) []yaml.MapSlice {
   return generateAddressShardingRelabelingRulesWithSourceLabel(relabelings, shards, "__param_target")
}

func generateAddressShardingRelabelingRulesWithSourceLabel(relabelings []yaml.MapSlice, shards int32, shardLabel string) []yaml.MapSlice {
```

simonpasquier (commented):
`TestProbe*()` tests need a fix.

mikechengwei:
@simonpasquier I have optimize the code and add a unit test .

pkg/prometheus/promcfg.go
```diff
@@ -1094,7 +1096,9 @@ func (cg *ConfigGenerator) generateProbeConfig(
 			enforcedNamespaceLabel: cg.spec.EnforcedNamespaceLabel,
 		}
 		relabelings = append(relabelings, rcg.generate(m.Spec.Targets.Ingress.RelabelConfigs)...)
-
+		if shards > 1 {
+			relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)
+		}
```
simonpasquier, line 1101: To be consistent with how we do for service/pod monitors already.

```suggestion
		relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)
```

simonpasquier (approved):
lgtm

Issue #4273, opened 2021-09-21
Support for using HTTPS in the Prometheus web UI is incomplete


jphx:
**What happened?**

I've been trying to use the feature of having the Prometheus web UI use HTTPS, via config in the `Prometheus` resource like the following:

```yaml
web:
  tlsConfig:
    keySecret:
      name: cert-secret
      key: tls.key
    cert:
      secret:
        name: cert-secret
        key: tls.crt
```

This appears to partially work.  It does configure Prometheus to use HTTPS, and it does change the readiness probe for the Prometheus container to use HTTPS.  But it _doesn't_ change the `config-reloader` container to use HTTPS when it contacts the Prometheus container (via the `--reload-url` option), and it _doesn't_ change the `thanos-sidecar` container to use HTTP when it contacts the Prometheus container (via the `--prometheus.url` option).

The `config-reloader` log fills up with messages like:
```
function failed. Retrying in next tick" err="trigger reload: received non-200 response: 400 Bad Request; have you set "--web.enable-lifecycle" Prometheus flag?
```
and the `thanos-sidecar` log fills up with messages like:
```
failed to get Prometheus flags. Is Prometheus running? Retrying" err="got non-200 response code: 400, response: Client sent an HTTP request to an HTTPS server.
```
I've reproduced this with version v0.50.0 of the Prometheus Operator.

**Did you expect to see something different?**

The `config-reloader` and `thanos-sidecar` containers should be configured to contact Prometheus using HTTPS instead of HTTPS, when the Prometheus web UI is using HTTPS.

One other important point:  My setup uses a self-signed certificate for the Prometheus web UI, so the `config-loader` and `thanos-sidecar` would have to (at least optionally) suppress certificate validation when they make their connections.

**How to reproduce it (as minimally and precisely as possible)**:

Use some config lines as above in the `Prometheus` resource to ask for HTTPS support in Prometheus.

**Environment**:

* Prometheus Operator version:  v0.50.0

* Kubernetes version information: v1.20.10

* Kubernetes cluster kind:  IBM Cloud Kubernetes cluster

fpetkovski:
This is indeed a bug, but changing the config reloader to use https can be problematic for certificates issued by an untrusted CA. We either need to mount the CA's certificate in `/etc/ssl/certs/` or add a feature to config-reloader for accept insecure connections.

@simonpasquier @PhilipGough thoughts on this one?

PhilipGough:
[...]

PR #4592, opened 2022-02-23
TLS certificate verification skip for prometheus-config-reloader

RemcoBuddelmeijer:
## Description

As discussed in #4273: proper configurable setup for config-reloader to support setting TLS parameter "skip TLS verify" when talking to Prometheus on a config reload. 

The only change is to the config-reloader command for skipping TLS verify. 
No flag is added since the solution intent was scoped to only supporting the config reloader for use in combination with the prometheus operator; no standalone application.

Support for this is then also not given. (Discussion can be found in the comments)

## Type of change

## Changelog entry

```release-note
- Add skip TLS verify for the config-reloader HTTP client when informing Prometheus on a config reload.
```

### DIFF
--- a/cmd/prometheus-config-reloader/main.go
+++ b/cmd/prometheus-config-reloader/main.go
@@ -19,6 +19,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	stdlog "log"
+	"net"
 	"net/http"
 	"os"
 	"regexp"
@@ -133,7 +134,7 @@ func main() {
 			},
 		)
 
-		client := createHTTPClient()
+		client := createHTTPClient(watchInterval)
 		rel.SetHttpClient(client)
 
 		g.Add(func() error {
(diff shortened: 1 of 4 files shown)


simonpasquier:
> Either allow setting TLS parameters or to completely ignore TLS verify

I'm fine with skipping TLS verification unconditionally since the config reloader connects directly to localhost. IIUC this would be equivalent to the way Kubernetes works with HTTPS probes.

@fpetkovski @paulfantom @PhilipGough thoughts?

RemcoBuddelmeijer:
> > Either allow setting TLS parameters or to completely ignore TLS verify
> 
> I'm fine with skipping TLS verification unconditionally since the config reloader connects directly to localhost. IIUC this would be equivalent to the way Kubernetes works with HTTPS probes.
> 
> @fpetkovski @paulfantom @PhilipGough thoughts?

I agree. However, since this is a newly introduced change it would be better to make this conditional. The default behaviour should not change just for this. My proposition would be to add a flag for this, f.e: --insecure-skip-tls-verify. 

Thoughts @simonpasquier?

fpetkovski:
I also think that it's safe to always skip TLS verification in this casem since both the config reloader and certificate are owned and managed by the user.

simonpasquier:
In the current state, the config-reloader doesn't work when Prometheus is configured with TLS so I'm not sure that we need to introduce a flag. Turning off TLS verification for the reloader wouldn't change anything when plain HTTP is used anyway.

RemcoBuddelmeijer:
> I also think that it's safe to always skip TLS verification in this casem since both the config reloader and certificate are owned and managed by the user.

It's a good point that the config reloader is managed and owned by the user. While this change is scoped only to the setup for the operator, aren't there any use cases where the certificates and prometheus instances are not owned by the user? Perhaps in situations where the config-reloader is a standalone application? 
This is my main reason to keep it toggle-able. 😄

fpetkovski:
I don't think it would be easy to run the config reloader as a standalone application right now and I am not sure the config reloader is even intended to run as a standalone or external application.

simonpasquier:
Yep I haven't heard that anybody uses the config-reloader outside of the prometheus-operator project (but I may be surprised).

RemcoBuddelmeijer:
@simonpasquier @fpetkovski Could the workflows be ran one last time to see if everything is up to standards? 🙂

cmd/prometheus-config-reloader/main.go
```diff
@@ -155,6 +159,19 @@ func main() {
 	}
 }
 
+func createHTTPClient() http.Client {
+	config := &tls.Config{
+		// TLS certificate verification is disabled by default
+		InsecureSkipVerify: true,
+	}
+
+	transport := &http.Transport{
```
simonpasquier, line 168: It would be good to clone `http.DefaultTransport` since this is what the reloader would use if no HTTP client is given.

```golang
        transport := (http.DefaultTransport.(*http.Transport)).Clone()
        transport.TLSClientConfig = config
```

RemcoBuddelmeijer: @simonpasquier I have gone ahead and done this alongside some changes to the transport timeout and max (idle) connections.
I based timeouts on my own tests on an AKS cluster, and local running Kind cluster. These values were within the margin (100ms margin).
Added comments to explain my changes and some can be removed/replaced once accepted.

cmd/prometheus-config-reloader/main.go
```diff
@@ -155,6 +160,31 @@ func main() {
 	}
 }
 
+func createHTTPClient() http.Client {
+	config := &tls.Config{
+		InsecureSkipVerify: true, // TLS certificate verification is disabled by default.
+	}
```
simonpasquier, line 166:
```suggestion
```

RemcoBuddelmeijer:
> Overall I wouldn't care about tuning the various timeouts since we've been using the default values and nobody complained about it so far. Tuning the max number of connections per host and disabling keep-alives are fine.

I have gone ahead and inlined the TLS config struct.
Looking at the use case and the "retrying every tick when failing" I decided to give changing the timeouts a go. 
Personally I find the Timeout on the client itself rather important since we shouldn't want it to wait until the timeout for watch interval without any logging. Especially since it will _always_ connect to localhost.
The other timeouts are just personal preference, so should we keep them out for now?

RemcoBuddelmeijer:
> > Personally I find the Timeout on the client itself rather important since we shouldn't want it to wait until the timeout for watch interval without any logging. Especially since it will always connect to localhost.
> > The other timeouts are just personal preference, so should we keep them out for now?
> 
> I'm not opposed to timeout tuning in theory but I'd like to change only one thing at a time and see how it goes.

Let's keep the changes as small as possible then. Removed the timeouts and cloned the Dialer timeout to be equal to the DefaultTransport.

simonpasquier (approved):
lgtm

PR #4595, opened 2022-02-24
Prepare 0.54.1 release

fpetkovski:
Prepares the 0.54.1 release.


PhilipGough (commented):
@fpetkovski - will we want to bump https://github.com/prometheus-operator/prometheus-operator/blob/main/example/admission-webhook/deployment.yaml also?

fpetkovski:
Looks like that file is not in the 0.54 release, or am I missing something?

PhilipGough:
> Looks like that file is not in the 0.54 release, or am I missing something?

No, you are fully correct :)

PhilipGough (approved):
lgtm

paulfantom (approved):
image

PR #4609, opened 2022-03-02
Add new AlertRelabelConfig type and CRD

bison:
## Description

This adds a new `AlertRelabelConfig` CRD for adding additional alert relabel configs to a Prometheus instance. This allows alert relabel configs to more easily be added from multiple sources when compared to the current approach of a single key in a Secret object. The current implementation configures alert relabelling with the combined set of configs loaded from the selected CRD instances and whatever may be loaded from the secret.

Example:
```yaml
---
kind: AlertRelabelConfig
metadata:
  labels:
    app.kubernetes.io/instance: k8s
  name: watchdog
  namespace: monitoring
spec:
  configs:
  - sourceLabels: [alertname, severity]
    regex: "Watchdog;none"
    targetLabel: severity
    replacement: critical
    action: replace
```

## Type of change

## Changelog entry

```release-note
- Add AlertRelabelConfig CRD for configuring alert relabelling alongside existing secret-based implementation.
```


bison:
Marking this as a draft. Sorry to drop a biggish PR. Just wanted to get it out there early to see if there was any interest.

It seems to work, but still needs tests at a minimum.

simonpasquier:
hey! sorry for the lag on reviewing this. While I see how it could be useful, I'm not convinced that we need to make it a CRD. My main concern is that the generated relabeling configs would not be really predictable.
What I could see though is an new field in the Prometheus CRD to express relabelings in a more convenient manner:

```
spec:
  alertRelabelConfigs:
  - sourceLabels: [alertname, severity]
    regex: "Watchdog;none"
    targetLabel: severity
    replacement: critical
    action: replace
  - sourceLabels: [alertname, severity]
    regex: "SomeAlert;none"
    targetLabel: severity
    replacement: warning
    action: replace
```

The benefit would be that the operator could enforce the validity of the configs using the same logic as what is done for the service/pod monitors.

bison:
Hey, thanks for taking a look. I like that idea, but ultimately the problem I was trying to solve with this CRD was allowing easier configuration of relabel configs from multiple sources. I think that's still a problem if they are all contained in a `Prometheus` instance.

Having thought about this more though, I actually agree that this is a bit unpredictable as-is, and probably isn't a good fit for the operator. I think I've come to the conclusion that his is better solved at a higher layer that just generates the secret the operator already knows how to read. So I guess I'm going to close this. Thanks again for having a look though.

PR #4626, opened 2022-03-09
*: avoid name collisions with AlertmanagerConfig

simonpasquier:
## Description

The change ensures that 2 AlertmanagerConfig objects don't generate
identical references in the final Alertmanager configuration. One
situation would be:
* AlertmanagerConfig `foo-bar` in namespace `ns` defines a receiver with name
  `fred`.
* AlertmanagerConfig `bar` in namespace `ns-foo` defines a receiver with name
  `fred`.

Without this patch, the resulting configuration would have 2 receivers
named `ns-foo-bar-fred` which is invalid.

With this patch, the receivers would be named `ns/foo-bar/fred` and
`ns-foo/bar/fred`. The `/` character has been chosen because it isn't
allowed for namespaces.

The issue is similar to what has been described in #3853 for pod and
service monitors.

## Type of change

## Changelog entry

```release-note
Avoid name collisions in Alertmanager configuration when merging AlertmanagerConfig objects.
```


PhilipGough (approved):
lgtm

Issue #4634, opened 2022-03-11
"converting (v1.APIGroup) to (v1.APIResourceList): unknown conversion" warning in the logs


simonpasquier:
**What happened?**

I've deployed the latest bundle.yaml on minikube. Checking the logs I see a warning about endpointslice support:

```
level=warn ts=2022-03-11T16:07:34.588477555Z caller=operator.go:316 component=prometheusoperator msg="failed to check if the API supports the endpointslice resources" err="converting (v1.APIGroup) to (v1.APIResourceList): unknown conversion"
```

**Did you expect to see something different?**

No error.

**Environment**

* Prometheus Operator version:

v0.55.0

* Kubernetes version information:

v1.23.1

* Kubernetes cluster kind:

minikube

* Manifests:

bundle.yaml from the repository.

* Prometheus Operator Logs:

See above.

**Anything else we need to know?**:

Thinking more about endpoint vs. endpointslice roles, it's very likely that the operator shouldn't automatically switch to the endpointslices role since the service account may not have permissions on the resource...

simonpasquier:
Closing in favor of https://github.com/prometheus-operator/prometheus-operator/issues/3862

PR #4635, opened 2022-03-11
[WIP] pkg/prometheus: fix endpointslice support

simonpasquier:
## Description

Putting on hold for now, I'm quite sure that it won't pass the e2e tests since we didn't add the permissions on the endpointslice resource...

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -311,12 +311,27 @@ func New(ctx context.Context, conf operator.Config, logger log.Logger, r prometh
 		c.nsPromInf = newNamespaceInformer(c, c.config.Namespaces.PrometheusAllowList)
 	}
 
-	endpointSliceSupported, err := k8sutil.IsAPIGroupVersionResourceSupported(c.kclient.Discovery(), "discovery.k8s.io", "endpointslices")
-	if err != nil {
-		level.Warn(c.logger).Log("msg", "failed to check if the API supports the endpointslice resources", "err ", err)
+	var (
+		endpointSliceSupported bool
+		groupVersion           string
+	)
+	for _, v := range []string{"v1", "v1beta1"} {
+		groupVersion = fmt.Sprintf("discovery.k8s.io/%s", v)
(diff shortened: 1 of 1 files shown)


PhilipGough:
@simonpasquier I started working on this and missed that you had started to tackle it.

I added the RBAC in #4636 but happy to close mine

simonpasquier:
Yeah, sorry I forgot to comment on #4634 that I had started looking into it.

I think that the "fix" is more involved than extending the RBAC permissions: in a model where service/pod monitors and prometheus objects aren't managed by the same person, the Prometheus service account may not have permissions on all `endpointslices` resources by default. This is the case in OpenShift where each component is responsible for setting up the service monitor + RBAC (example with the etcd operator with the [service monitor](https://github.com/openshift/cluster-etcd-operator/blob/master/manifests/0000_90_etcd-operator_03_servicemonitor.yaml), [role](https://github.com/openshift/cluster-etcd-operator/blob/master/manifests/0000_90_etcd-operator_01_prometheusrole.yaml) and [role binding](https://github.com/openshift/cluster-etcd-operator/blob/master/manifests/0000_90_etcd-operator_02_prometheusrolebinding.yaml)). If we were to let the operator decide which of the `endpoints` or `endpointslice` role should be used in the Kubernetes SD configs by only looking at what's supported by the Kube API, it would break these environments.

I can see at least 3 options (not exclusive):
* Add a field to the `ServiceMonitor` and `PodMonitor` CRDs to let users control which role should be used.
* Add a field to the `Prometheus` CRD to define the default Kubernetes SD role (defaults to `endpoints`). In case the user has full control on the monitoring objects and RBAC permissions, they can easily switch to endpointslices.
* The operator checks the permissions of the Prometheus SA using `LocalSubjectAccessReview` and selects the appropriate role based on the response.

simonpasquier:
Closing for now since it's more involved as described in https://github.com/prometheus-operator/prometheus-operator/issues/3862#issuecomment-1068260430.

PR #4645, opened 2022-03-16
Update time range for alert PrometheusOperatorWatchErrors

JoaoBraveCoding:
## Description

The alert expression has a time range of 10 minutes so it will not resolve "quickly" if the operator experiences a transient burst of watch errors.
Given that metrics are scraped every 30 seconds, it would be better to use a time range of 5 minutes (consistent with what Prometheus & Alertmanager are doing).

## Type of change

## Changelog entry

```release-note
Updated the time range for the alert PrometheusOperatorWatchErrors from 10m to 5m
```

### DIFF
--- a/example/mixin/alerts.yaml
+++ b/example/mixin/alerts.yaml
@@ -17,7 +17,7 @@ groups:
         in {{$labels.namespace}} namespace.
       summary: Errors while performing watch operations in controller.
     expr: |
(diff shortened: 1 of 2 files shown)


simonpasquier (approved):
lgtm

simonpasquier:
thanks! I'll merge on green :)

PR #4646, opened 2022-03-16
*: use consistent image name for the admision webhook

simonpasquier:
## Description

The end-to-end tests and jsonnet code were using
`quay.io/prometheus-operator/prometheus-admission-webhook` which doesn't
exist.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -68,7 +68,7 @@ jobs:
       run: |
         kind load docker-image quay.io/prometheus-operator/prometheus-operator:$(git rev-parse --short HEAD)
         kind load docker-image quay.io/prometheus-operator/prometheus-config-reloader:$(git rev-parse --short HEAD)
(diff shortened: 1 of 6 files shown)


simonpasquier:
cc @PhilipGough

simonpasquier:
So I'm not changing the image name in the public registries (they're already using the short name with the prometheus- prefix: https://quay.io/repository/prometheus-operator/admission-webhook and https://github.com/prometheus-operator/prometheus-operator/pkgs/container/admission-webhook). I've discovered the inconsistency when trying to run the e2e tests locally. It wasn't noticed before because the e2e workflow was building from the Makefile which used the prefixed version by default. On the other hand, the CI job publishing the images uses `scripts/push-docker-image.sh` which overrides the image names.

I don't think that the [README](https://github.com/prometheus-operator/prometheus-operator#running-end-to-end-tests-on-local-kind-cluster) needs an update since it's already using the correct name.

PhilipGough (approved):
LGTM

PR #4652, opened 2022-03-17
add support for basic auth web config

### LINKED ISSUE #4200: prometheus basic auth
**What did you do?**

Looked at previous issues but `basic-auth` flags a few but they all appear to be prometheus -> endpoint rather than say grafana -> prometheus.

I'm looking to setup prometheus per cluster i have and then have a single grafana instance hit the prometheuses as seperate data-sources. In order for me to feel comfortable about exposing prometheus apis externally (even behind a whitelist) i would ideally configure basic-auth on prometheus itself. See https://prometheus.io/docs/guides/basic-auth/

I was expecting to see something around setting a user/password combo or list of usernames passwords for an instance of prometheus.

Documentation that states its possible to enable basic auth on prometheus itself (again not endpoints prometheus hits). https://prometheus.io/docs/guides/basic-auth/

I can also see in Grafana that it is possible to configure basic auth for the prometheus data source:
image

Hades32:
## Description

Extends WebSpec with basic-auth user information to secure the Prometheus API

This fixes #4200 

## Type of change

## Changelog entry

Add support for configuring HTTP basic auth. See [Prometheus docs](https://prometheus.io/docs/guides/basic-auth/) for details.

## Draft: Open Points
* [x] refactor to use secrets to store basic auth users
* [ ] add internal user for health-check and config reloader

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -304,7 +304,7 @@ Authorization contains optional `Authorization` header configuration. This secti
 BasicAuth allow an endpoint to authenticate over basic authentication More info: https://prometheus.io/docs/operating/configuration/#endpoints
 
 
-<em>appears in: [APIServerConfig](#apiserverconfig), [Endpoint](#endpoint), [PodMetricsEndpoint](#podmetricsendpoint), [ProbeSpec](#probespec), [RemoteReadSpec](#remotereadspec), [RemoteWriteSpec](#remotewritespec)</em>
+<em>appears in: [APIServerConfig](#apiserverconfig), [Endpoint](#endpoint), [PodMetricsEndpoint](#podmetricsendpoint), [ProbeSpec](#probespec), [RemoteReadSpec](#remotereadspec), [RemoteWriteSpec](#remotewritespec), [WebSpec](#webspec)</em>
 
 | Field | Description | Scheme | Required |
 | ----- | ----------- | ------ | -------- |
@@ -1168,6 +1168,7 @@ WebSpec defines the query command line flags when starting Prometheus.
 | Field | Description | Scheme | Required |
 | ----- | ----------- | ------ | -------- |
 | pageTitle | The prometheus web page title | *string | false |
+| basicAuthUsers | Defines the basic auth users to access the Prometheus API/UI. See https://prometheus.io/docs/guides/basic-auth/ | [][BasicAuth](#basicauth) | false |
 | tlsConfig |  | *[WebTLSConfig](#webtlsconfig) | false |
 
 [Back to TOC](#table-of-contents)
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -19432,6 +19432,53 @@ spec:
                 description: WebSpec defines the web command line flags when starting
                   Prometheus.
                 properties:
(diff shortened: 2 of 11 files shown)


simonpasquier (commented):
Thanks for your PR!

We shouldn't expose credentials directly in the custom resource for security reasons. ~~One approach would be to have a list of references to secret keys like for other sensitive fields.~~ Lets reuse the existing BasicAuth struct:

```
type Prometheus struct {
	...
	BasicAuthUsers []BasicAuth `json:"basicAuthUsers,omitempty"`
	...
}
```

Another point to keep in mind is that the config reloader uses the Prometheus web service so it needs to be configured with valid authn credentials.

simonpasquier:
I see that @fpetkovski commented on the issue that we could reuse the existing [basicAuth](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#basicauth) type which makes sense indeed.

Hades32:
Thanks for your comments! Yes, I realized just now, that also the health check is actually affected by this...
But I'll continue working on this over the next few days

Hades32:
@simonpasquier OK, so changing to the existing `BasicAuth` type was straightforward, but fixing health-checks and config reloader will be trickier. My idea would be like this:
* always create a basic auth user secret for each `Prometheus` object (something like `internal-operator-user`, with a UUID as password)
* if the list of basic auth users exists and has entries, add this user 
* somehow add this user to the health-check and config reloader
  * here your input would be appreciated

WDYT?

simonpasquier:
My instinct would be to generate a user/password for the config reloader and append it to the list provided by the user (if not empty). It sounds better than reusing an existing user/password (cleaner separation).
 
Liveness/readiness probes are another issue indeed... In this case, we could use an [exec probe](https://stackoverflow.com/questions/33484942/how-to-use-basic-authentication-in-a-http-liveness-probe-in-kubernetes/33488076#33488076) (already the case when Prometheus listens on localhost) and inject the credentials via environment variables. Ideally Prometheus shouldn't require authentication for the `/-/ready` and `/-/healthy` endpoints.

RomainPhil:
Hi any news concerning this feature ?

Hades32:
Unfortunately, I didn't have time to finish this and solved it by simply running an Nginx sidecar container that only does basic-auth

shenxuxie:
> 



> Thanks for your PR!
> 
> We shouldn't expose credentials directly in the custom resource for security reasons. ~One approach would be to have a list of references to secret keys like for other sensitive fields.~ Lets reuse the existing BasicAuth struct:
> 
> ```
> type Prometheus struct {
> 	...
> 	BasicAuthUsers []BasicAuth `json:"basicAuthUsers,omitempty"`
> 	...
> }
> ```
> 
> Another point to keep in mind is that the config reloader uses the Prometheus web service so it needs to be configured with valid authn credentials.

how to configur the config reloader with authncredentials

ArthurSens:
Hey @Hades32, I'm closing this one since you've communicated that you won't be able to continue. Thanks for the effort and feel free to re-open if you feel like continuing in the future :)

PR #4661, opened 2022-03-18
pkg/apis/monitoring/v1: Add validations for size settings at CRD level

slashpai:
pkg/apis/monitoring/v1: Add validations for size settings at CRD level
test: Add e2e test for CRD validation

This is needed to avoid invalid values from user which would
cause problem for application to load or not work as expected

Related-to prometheus-operator#4524

## Description

pkg/apis/monitoring/v1: Add validations for size settings at CRD level

https://regex101.com/r/sRlRj5/1

## Type of change

## Changelog entry

```release-note
Added validations for size settings at CRD level
```


simonpasquier (commented):
Would it work if we define an intermediate type (`BytesSize`) so there's only one validation regex?

pkg/apis/monitoring/v1/types.go
```diff
@@ -317,6 +317,7 @@ type CommonPrometheusFields struct {
 	// This is an experimental feature, this behaviour could
 	// change or be removed in the future.
 	// Only valid in Prometheus versions 2.28.0 and newer.
+	// +kubebuilder:validation:Pattern:="^0|[0-9]+B|([K|M|G|T|E|P]B)$"
```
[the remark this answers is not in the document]
slashpai: I initially included both iB and B in regex but I saw like this in [prometheus](https://github.com/prometheus/prometheus/blob/main/cmd/prometheus/main.go#L296-L297)  which says it allows these units only

```
serverOnlyFlag(a, "storage.tsdb.retention.size", "Maximum number of bytes that can be stored for blocks. A unit is required, supported units: B, KB, MB, GB, TB, PB, EB. Ex: \"512MB\". Based on powers-of-2, so 1KB is 1024B.").
		BytesVar(&cfg.tsdb.MaxBytes)
```

 [unitsBase2Bytes](https://github.com/alecthomas/units/blob/b94a6e3cc13755c0a75fffecbb089eb346fc4289/bytes.go) allows both iB and B variations but I see another conversion done using [BytesVar()](https://github.com/alecthomas/kingpin/blob/17f40c25f4977ef446a79e34e46270665dffa263/parsers.go#L131-L134). I am trying to parse a few values to test valid/invalid ones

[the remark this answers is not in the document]
slashpai: Since both units worked I updated regex

slashpai:
We would need to update regex to allow values like `1.5GB`

slashpai:
@simonpasquier I updated the PR with regex to support floating point values as well like `1.5GB`
https://regex101.com/r/sRlRj5/1

slashpai:
@simonpasquier Can you review again?

pkg/apis/monitoring/v1/types.go
```diff
@@ -317,6 +317,8 @@ type CommonPrometheusFields struct {
 	// This is an experimental feature, this behaviour could
 	// change or be removed in the future.
 	// Only valid in Prometheus versions 2.28.0 and newer.
+	// Supported units: B, KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB Ex: `512MB`.
+	// +kubebuilder:validation:Pattern:="(^0|([0-9]*[.])?[0-9]+((K|M|G|T|E|P)i?)?B)$"
```
simonpasquier, line 321: instead of duplicating the regex here and a few lines below, it would be good to create an intermediate type (`type BytesSize string`) so we don't have to update the expression at several places.

slashpai:
@simonpasquier I have updated per suggestion

pkg/prometheus/promcfg.go
```diff
@@ -363,13 +363,13 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
 // validateConfigInputs runs extra validation on the Prometheus fields which can't be done at the CRD schema validation level.
 func validateConfigInputs(p *v1.Prometheus) error {
 	if p.Spec.EnforcedBodySizeLimit != "" {
-		if err := operator.ValidateSizeField(p.Spec.EnforcedBodySizeLimit); err != nil {
+		if err := operator.ValidateSizeField(string(p.Spec.EnforcedBodySizeLimit)); err != nil {
```
simonpasquier, line 366: eventually we can get rid of the redundant validation. To handle the transition, we can remove it in a couple of releases (lets say in v0.58.0).

@PhilipGough thoughts? If we agree on this, we need a TODO here.

pkg/prometheus/promcfg.go
```diff
@@ -363,13 +363,13 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
...
 			return errors.Wrap(err, "invalid enforcedBodySizeLimit value specified")
 		}
 	}
 
 	if p.Spec.RetentionSize != "" {
-		if err := operator.ValidateSizeField(p.Spec.RetentionSize); err != nil {
+		if err := operator.ValidateSizeField(string(p.Spec.RetentionSize)); err != nil {
```
simonpasquier, line 372: same here

simonpasquier (commented):
It would be good to check that the validation works as expected. From what I can tell, it has to be done in the e2e tests. Could you add one test with valid values and another test with incorrect values?

pkg/prometheus/promcfg.go
PhilipGough: Yeah, adding a TODO sounds like the safest thing to do in this case. It would be good to identify any similar cases and aim to remove them all in the single release.

pkg/prometheus/promcfg.go
```diff
@@ -362,14 +362,16 @@ func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
 
 // validateConfigInputs runs extra validation on the Prometheus fields which can't be done at the CRD schema validation level.
 func validateConfigInputs(p *v1.Prometheus) error {
+	// TODO(slashpai): Remove this validation since this is handled at CRD level
```
simonpasquier, line 365:
```suggestion
	// TODO(slashpai): Remove this validation after v0.57 since this is handled at CRD level
```

slashpai: done

test/e2e/prometheus_test.go
```diff
@@ -3905,6 +3905,100 @@ func testPromEnforcedNamespaceLabel(t *testing.T) {
...
+		{
+			name: "invalidRetentionSize2",
+			prometheusSpec: monitoringv1.PrometheusSpec{
+				CommonPrometheusFields: commonFields,
+				RetentionSize:          "1000",
+			},
+			expectedError: true,
+		},
+		{
+			name: "invalidRetentionSize3",
```
simonpasquier, line 3970:
```suggestion
			name: "IEC unit with missing symbol",
```

test/e2e/prometheus_test.go
```diff
@@ -3905,6 +3905,100 @@ func testPromEnforcedNamespaceLabel(t *testing.T) {
...
+		},
+		{
+			name: "iec-unit",
+			prometheusSpec: monitoringv1.PrometheusSpec{
+				CommonPrometheusFields: commonFields,
+				RetentionSize:          "100MiB",
+			},
+		},
+		{
+			name: "legacy-missing-symbol",
```
slashpai, line 3954: I had to cut short the suggestion to avoid error while creating namespace since we use this name

``` 
namespace.go:39: failed to create namespace with name allns-y-prometheuscrdvalidation-legacy-unit-with-missing-symbol-r971yd-0: Namespace "allns-y-prometheuscrdvalidation-legacy-unit-with-missing-symbol-r971yd-0" is invalid: [metadata.name: Invalid value: "allns-y-prometheuscrdvalidation-legacy-unit-with-missing-symbol-r971yd-0": must be no more than 63 characters, metadata.labels: Invalid value: "allns-y-prometheuscrdvalidation-legacy-unit-with-missing-symbol-r971yd-0": must be no more than 63 characters]
```

slashpai:
@simonpasquier Please another look I updated the PR

test/framework/prometheus.go
```diff
@@ -63,6 +63,17 @@ type PromRemoteWriteTestConfig struct {
 	ShouldSuccess      bool
 }
 
+func (f *Framework) MakeBasicPrometheusCRForValidation(ns, name string, spec monitoringv1.PrometheusSpec) *monitoringv1.Prometheus {
```
simonpasquier, line 66: not sure that we need the extra function if it's only called from one place.

test/e2e/prometheus_test.go
```diff
@@ -3905,6 +3905,100 @@ func testPromEnforcedNamespaceLabel(t *testing.T) {
...
+			prom := framework.MakeBasicPrometheusCRForValidation(ns, name, test.prometheusSpec)
+			prom.Namespace = ns
+
+			_, err := framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, prom)
+
+			if test.expectedError && err == nil {
+				t.Fatalf("expected error but got nil")
+			}
+			t.Logf("%v", err)
+		})
```
simonpasquier, line 3998: Don't we miss the opposite check (e.g. non-nil error returned but not expected)?
We could also ensure that the returned error is about invalid input using  [`errors.IsInvalid`](https://pkg.go.dev/k8s.io/apimachinery@v0.23.5/pkg/api/errors#IsInvalid)).

```suggestion
			p := framework.MakeBasicPrometheusCRForValidation(ns, name, test.prometheusSpec)
			if tc.expectedError {
				_, err := framework.MonClientV1.Prometheuses(ns).Create(ctx, p, metav1.CreateOptions{})
				if err == nil {
					t.Fatal("expected error but got nil")
				}
				if !errors.IsInvalid(err) {
					t.Fatalf("expected Invalid error but got %v", err)
				}
				return
			}

			_, err := framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, prom)

			if err != nil {
				t.Fatalf("expected no error but got %v", err)
			}
		})
```

test/framework/prometheus.go
simonpasquier: I would argue that in this case, the test can create the Prometheus object directly without an helper function.

slashpai:
@simonpasquier updated per review

simonpasquier (approved):
lgtm

simonpasquier:
thanks!

PR #4681, opened 2022-03-25
Make termination grace seconds configurable

yeya24:
Signed-off-by: Ben Ye <ben.ye@bytedance.com>

## Description

We have the usecase to configure the termination grace seconds value of the prometheus statefulset. Now the value is hardcoded as 10m. 

## Type of change

## Changelog entry

```release-note
Make terminationGracePeriodSeconds configurable for the Prometheus Statefulset
```


pkg/apis/monitoring/v1/types.go
```diff
@@ -335,6 +335,16 @@ type CommonPrometheusFields struct {
...
+	// Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request.
+	// Value must be non-negative integer. The value zero indicates stop immediately via
+	// the kill signal (no opportunity to shut down).
+	// If this value is nil, the default grace period will be used instead.
+	// The grace period is the duration in seconds after the processes running in the pod are sent
+	// a termination signal and the time when the processes are forcibly halted with a kill signal.
+	// Set this value longer than the expected cleanup time for your process.
+	// Defaults to 600 seconds.
+	// +optional
+	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
```
slashpai, line 347: Should this be better to use uint to avoid negative values

slashpai: Also add default value using validation marker so we can remove the conditional from statefulset code

`// +kubebuilder:default:="600"`

yeya24: Done. Thanks for the review.

pkg/apis/monitoring/v1/types.go
```diff
@@ -338,13 +338,14 @@ type CommonPrometheusFields struct {
...
-	// If this value is nil, the default grace period will be used instead.
+	// If this value is nil, the default grace period will be used instead. Default value is set to
+	// 10 min because Prometheus may take quite long to shutdown to checkpoint existing data.
 	// The grace period is the duration in seconds after the processes running in the pod are sent
 	// a termination signal and the time when the processes are forcibly halted with a kill signal.
 	// Set this value longer than the expected cleanup time for your process.
-	// Defaults to 600 seconds.
 	// +optional
-	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
+	// +kubebuilder:default:="600"
```
slashpai, line 347: Since its int type we don't need quotes

`// +kubebuilder:default:=600`

pkg/prometheus/statefulset.go
```diff
@@ -907,10 +900,16 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 		}
 	}
 
-	var minReadySeconds int32
+	var (
+		minReadySeconds        int32
+		terminationGracePeriod int64
```
slashpai, line 905: This should be `uint` too?

slashpai (commented):
Can you also add tests?

pkg/prometheus/statefulset.go
yeya24: No it is casted from uint64

pkg/apis/monitoring/v1/types.go
yeya24: Thanks. Updated

yeya24:
I have added tests. Can you please help take a look? Thanks!

PhilipGough (approved):
lgtm

yeya24:
As I mentioned in https://github.com/thanos-io/thanos/pull/5255, the issue is on our CNI side. It enables graceful termination on the IPVS side so the connection will remain for 10min in prometheus operator case. If the promethues itself is down, because of this problem, thanos sidecar is still available, causing a lot of partial query errors from our Thanos Query for 10m.

simonpasquier:
@yeya24 sorry for the late follow-up but I'm not sure to understand exactly the scenario. You want to configure a shorter termination grace period in case Prometheus is stuck?

yeya24:
> @yeya24 sorry for the late follow-up but I'm not sure to understand exactly the scenario. You want to configure a shorter termination grace period in case Prometheus is stuck?

Yes. If it is stuch in case the backend storage like Ceph is not responsive, then it doesn't make sense to wait for 10m to ensure data is written successfully because the storage is done.
In this case, I want to stop it ASAP.

paulfantom (changes requested):
> If the promethues itself is down somehow, then thanos sidecar is still available

That sounds like a thanos sidecar issue, not prometheus-operator or prometheus one.

---
I would be against configurable `terminationGracePeriodSeconds`. In the majority of cases tweaking it can lead to unexpected data loss. I also understand the case that @yeya24 is making about CNI and CSI being unresponsive and that in those cases fast termination is beneficial. However those issues are edge cases which can happen in particular critical scenarios which are most likely handled directly by user via `kubectl`. And if that is the case then `kubectl delete pod <prometheus> --grace-period=0 ` is likely a better way.

yeya24:
> In the majority of cases tweaking it can lead to unexpected data loss

If data loss is fine and users want to have a termination time < 10m. I think this use case is still valid since 10m might not fit all users. In our case we want a smaller duration like 5m. What about other use cases that think 10m is too short and they want to have a longer duration like 1h?

My point is that an operator should provide some way to allow users to configure k8s native fields like `terminationGracePeriodSeconds`. Users can take this risk if they really want to like us.

rnaveiras:
Hi, team; it's been a while since the last messages in this PR. 

I'm interested in the feature. However, my use case is different. We want to increase beyond the default 10 minutes because, in some cases, that is not enough for a graceful shutdown if you've enabled the feature flag for snapshot in-memory chunks. See details https://github.com/prometheus/prometheus/pull/7229

Our setup required more than 10 minutes to complete the chunk snapshots successfully. 

/cc @yeya24 I'm happy to collaborate to get this PR in good shape again.

simonpasquier:
@rnaveiras more than 10 minutes for the chunks snapshot seems a bit extreme. Do you have an explanation why it takes so long? Have you tried reporting the issue to prometheus/prometheus?

Having said that, we had many requests in the past to customize the termination grace periods and though sometimes the justification could be challenged, we also agreed in #4691 that we shouldn't block such customization if there was high demand from the community and no alternative existed.
In summary, feel free to resurrect this pull request (sharing credits with @yeya24 of course).

PR #4683, opened 2022-03-25
Fix Prometheus configuration when query log file has an empty path

simonpasquier:
## Description

A Prometheus object with `spec.queryLogFile: "query.log"` can't be successfully rolled out because the `query_log_file` field in the Prometheus configuration doesn't include the full path to the emptyDir volume's mount.

## Type of change

## Changelog entry

```release-note
Fixed Prometheus configuration when  `spec.queryLogFile` has no path ("query.log" for instance)
```


PhilipGough (approved):
lgtm

PR #4698, opened 2022-03-29
Fix k8s.io/client-go version to have semantic versioning

regarm:
## Description

This repository generates error if included as a dependency in strict semantic versioned enviroment.

```
go list -mod=readonly -m -json all': go list -m: k8s.io/client-go@v12.0.0+incompatible: invalid version: +incompatible suffix not allowed: module contains a go.mod file, so semantic import versioning is required
```

## Type of change

## Changelog entry

Update k8s.io/client-go to v0.23.5 from v12.0.0+incompatible for semantic version

```release-note
Update k8s.io/client-go version to v0.23.5.
```


simonpasquier (commented):
can you also remove the `replace` [directive](https://github.com/prometheus-operator/prometheus-operator/blob/4f02938492a804600adae5aeace96c59434c4094/go.mod#L134-L136)? It looks like it's not needed anymore.

PhilipGough (approved):
lgtm

regarm:
> can you also remove the `replace` [directive](https://github.com/prometheus-operator/prometheus-operator/blob/4f02938492a804600adae5aeace96c59434c4094/go.mod#L134-L136)? It looks like it's not needed anymore.

Done

fpetkovski (approved):

Issue #4677, opened 2022-03-23
Add v1beta1 AlertmanagerConfig CRD


simonpasquier:
The AlertmanagerConfig CRD has been introduced as v1alpha1 in v0.43.0, one and half years ago. We've collected great feedback so far and the CRD now supports all Alertmanager configuration options. We'd like more users to benefit from it and offering a v1beta1 version of the CRD would increase adoption (beta version means that the project won't remove the API in the future and provides increased support).

I want to use this issue as a way to share and track the work towards this goal.

## Overview

Though alpha API versions may be removed in any release without prior deprecation notice, we want the migration from v1alpha1 to v1beta1 to be easy and seamless. The project will implement conversion webhooks to convert v1alpha1 objects to v1beta1 (and vice-versa). Initially the CRD will continue to use v1alpha1 as the stored version (e.g. the representation persisted in etcd) but to encourage users to switch to the new API version, it will switch in a couple of releases to v1beta1.

# Changes

## API changes

* Remove the `regex` field from the [Matcher](https://prometheus-operator.dev/docs/operator/api/#matcher) type.
  * When converting from v1alpha1 to v1beta1
    * "empty" matchType + false regex turns into `matchType: "="`
    * "empty" matchType + true regex turns into matchType: "=~"
  * When converting from v1beta1 to v1alpha1
    * Nothing changes
* Anticipate support for active time intervals (see #4656)
  * Rename `spec.muteTimeIntervals` to `spec.timeIntervals`
    * When converting from v1alpha1 to v1beta1
      * Copy `spec.muteTimeIntervals` to `spec.timeIntervals`
    * When converting from v1beta1 to v1alpha1
      * Copy `spec.timeIntervals` to `spec.muteTimeIntervals`
  * Rename the MuteTimeInterval type to TimeInterval
  * Rename the TimeInterval type to TimePeriod

## Improvements

* Push field validations to the open API specification instead of when reconciling the resources.

## Supported Alertmanager versions

We continue to support all Alertmanager versions without limitation. 

## Deprecation plan

* [v0.57.0] -> [v0.58.0] (~3 months): Alertmanager CRD v1beta1 is opt-in (e.g. not included in `bundle.yaml`).
* [v0.59.0] -> [v0.62.0] (~6 months): Alertmanager CRD v1beta1 is deployed by default, v1alpha1 marked as deprecated but still defined as the stored version and used by the operator.
[...]

PR #4709, opened 2022-04-08
*: add v1beta1 for AlertmanagerConfig CRD

simonpasquier:
## Description

This change implements #4677. It includes an extension of the webhook admission service that enables converting from v1alpha1 to v1beta1 (and vice-versa). The stored version is v1alpha1 to enable smooth migration.

As discussed during the project office hours and documented in the associated issue, v1beta1 is going to be opt-in for 2 releases. This is to let time for users to install the conversion webhook as long as they don't want/need to support v1beta1.

- [X] Implement v1beta1 as a 1:1 copy of v1alph1.
- [X] Implement conversion webhook.
- [x] Unit tests for conversion webhook.
- [x] Remove the regex field from the [Matcher](https://prometheus-operator.dev/docs/operator/api/#matcher) type. 
- [x] Rename spec.muteTimeIntervals to spec.timeIntervals
- [x] Rename the MuteTimeInterval type to TimeInterval (and TimeInterval type to TimePeriod).
- [x] Replace `v1.SecretKeySelector` by `SecretKeySelector` (`optional` field removed + `name` and `key` required).
- [x] Update `Documentation/user-guides/webhook.md`
- [x] ~~Fix Documentation/api.md~~ (edit: v1beta1 being opt-in, the documentation part can wait a bit and it would be best tackled as part of https://github.com/prometheus-operator/prometheus-operator/issues/4376). 

## Type of change

## Changelog entry

```release-note
Added v1beta1 version of AlertmanagerConfig CRD.
```


simonpasquier:
@PhilipGough @paulfantom @fpetkovski can you please review this?

Documentation/user-guides/webhook.md
```diff
@@ -1,50 +1,89 @@
...
+volume and the container arguments should be modified to include the
+certificate and key:
 
-* `--web.key-file` to load the associate key.
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: prometheus-operator-admission-webhook
+  namespace: default
```
PhilipGough, line 52: do you think it might be a good idea to remove the `namespace` field from these examples entirely.

Documentation/user-guides/webhook.md
```diff
@@ -146,3 +187,42 @@ webhooks:
...
+            },
+            "conversionReviewVersions": [
+               "v1beta1",
+               "v1alpha1"
+            ]
+         }
+      }
+   }
+}
+```
```
PhilipGough, line 228: Nice!

Makefile
```diff
@@ -42,6 +43,7 @@ K8S_GEN_ARGS:=--go-header-file $(shell pwd)/.header --v=1 --logtostderr
 K8S_GEN_DEPS:=.header
 K8S_GEN_DEPS+=$(TYPES_V1_TARGET)
 K8S_GEN_DEPS+=$(TYPES_V1ALPHA1_TARGET)
+K8S_GEN_DEPS+=$(TYPES_V1BETA1_TARGET)
```
PhilipGough, line 46: do we need to retain the alpha targets?

Documentation/user-guides/webhook.md
simonpasquier: yes it might make sense

Makefile
simonpasquier: yes because v1alpha1 and v1beta1 are going to co-exist for a while.

Documentation/user-guides/webhook.md
simonpasquier: after more thinking, I realize that the webhook and conversion configurations expect the service's namespace so it's probably better to keep `default`. It's easy enough for the reader to adapt the snippets to their environment.

simonpasquier:
@prometheus-operator/prometheus-operator-reviewers please have a look. As discussed during our last office hours, this requires explicit action from cluster admins to expose the v1beta1 version so this is quite safe for the majority of users.

Documentation/user-guides/webhook.md
```diff
@@ -79,7 +118,8 @@ webhooks:
 
 ### `/admission-prometheusrules/mutate`
 
-The endpoint `/admission-prometheusrules/mutate` ensures that integers and boolean yaml data elements are cooerced into strings.
+The endpoint `/admission-prometheusrules/mutate` ensures that integers and
+boolean yaml data elements are coerced into strings.
```
PhilipGough, line 122: Although it is already clear to me, perhaps since for the `/validate` endpoint, you mentioned explicitly it is for `PrometheusRule`, for consistency we might do the same?

simonpasquier: good point!

PhilipGough (approved):

Issue #4722, opened 2022-04-15
Add support to configure Prometheus' Tracing


ArthurSens:
**What is missing?**
Tracing was introduced to Prometheus in [version 2.18.0](https://github.com/prometheus/prometheus/releases/tag/v2.18.0), using the Jaeger library.

Jaeger's lib was deprecated and has finally been sunset so it didn't make sense to commit support for something that was definitely going to be replaced.

In [version 2.34.0](https://github.com/prometheus/prometheus/releases/tag/v2.34.0), Prometheus changed its library to OpenTelemetry, and with it a lower likelihood of changes. Maybe it is a good time to add support to the operator as well?

**Why do we need it?**

Prometheus users might need to collect traces to troubleshoot Prometheus.

ArthurSens:
Fixed by #5591

PR #4723, opened 2022-04-15
WIP: Tracing config support

ArthurSens:
## Description
Started working on #4722, but opening as a draft already to get feedback on the API changes. Does that seems appropriate?

Here is the link for `tracing_config` docs, in case it helps: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tracing_config

## Type of change

## Changelog entry

```release-note
Add support for configuring tracing in Prometheus CR
```


pkg/apis/monitoring/v1/types.go
```diff
@@ -545,6 +548,57 @@ type ArbitraryFSAccessThroughSMsConfig struct {
...
+	// Compression key for supported compression types. Supported compression: gzip
+	//+kubebuilder:validation:Enum=gzip
+	Compression PrometheusTracingCompression `json:"compression"`
+
+	// Maximum time the exporter will wait for each batch export. Default '10s'
+	// +kubebuilder:default:="10s"
+	Timeout Duration `json:"timeout"`
+
+	// TLS Config to use when sending traces.
+	TLSConfig TLSConfig `json:"tlsConfig"`
```
simonpasquier, line 580:
```suggestion
	TLSConfig *TLSConfig `json:"tlsConfig"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -545,6 +548,57 @@ type ArbitraryFSAccessThroughSMsConfig struct {
...
+type PrometheusTracingClientType string
+
+const (
+	HTTPClientType PrometheusTracingClientType = "http"
+
+	GRPCClientType PrometheusTracingClientType = "grpc"
+)
+
+// Compression key for supported compression types. Supported compression: gzip
+//+kubebuilder:validation:Enum=gzip
```
simonpasquier, line 595: IIUC it can be empty or `gzip`?

pkg/apis/monitoring/v1/types.go
```diff
@@ -545,6 +548,57 @@ type ArbitraryFSAccessThroughSMsConfig struct {
 	Deny bool `json:"deny,omitempty"`
 }
 
+type PrometheusTracingConfig struct {
+	// Client used to export the traces. Options are 'http' or 'grpc'.
+	//+kubebuilder:validation:Enum=http;grpc
+	// +optional
+	ClientType PrometheusTracingClientType `json:"clientType"`
```
simonpasquier, line 555: Unless the same type is used elsewhere, I don't think we need a `PrometheusTracingClientType` type.

```suggestion
	ClientType string `json:"clientType"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -545,6 +548,57 @@ type ArbitraryFSAccessThroughSMsConfig struct {
...
+	// If disabled, the client will use a secure connection.
+	// +optional
+	Insecure *bool `json:"insecure"`
+
+	// Key-value pairs to be used as headers associated with gRPC or HTTP requests.
+	Headers map[string]string `json:"headers"`
+
+	// Compression key for supported compression types. Supported compression: gzip
+	//+kubebuilder:validation:Enum=gzip
+	Compression PrometheusTracingCompression `json:"compression"`
```
simonpasquier, line 573: same here 
```suggestion
	Compression string `json:"compression"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -545,6 +548,57 @@ type ArbitraryFSAccessThroughSMsConfig struct {
...
+
+	// Maximum time the exporter will wait for each batch export. Default '10s'
+	// +kubebuilder:default:="10s"
```
simonpasquier, line 576: I would leave the default to "".

pkg/apis/monitoring/v1/types.go
ArthurSens: Yep sounds fair 👍 

This is not required to add validation at the CRD level, right?

pkg/apis/monitoring/v1/types.go
ArthurSens: So maybe it could be `//+optional` instead? Is there any difference?

ArthurSens:
Simon, thanks for the early feedback :) I was expecting to have time to work on this in the last week but life had other plans for me 😅. I think I won't have time to continue this one for the next 3 weeks, so no rush for more feedback, but have some questions:

* I'm not really sure where I should be adding the field validations (and the tests). I ended up validating clientType, Compression, endpoint and samplingFraction while rendering the config, but I think it would be a lot better to validate with webhookvalidations when creating the CRD. Is that possible for Prometheus-operator? If yes, where should this code live?

* If we go with webhookvalidations, do we have unit tests? Or are e2e tests enough?

mateuszdrab:
Hey @ArthurSens 
I'd be interested in seeing this feature, I don't think there is a way to configure Prometheus via the operator and add this config in, is there? I reviewed the API doc and can't see any options for 'extraConfig' or similar.

ArthurSens:
https://github.com/prometheus-operator/prometheus-operator/pull/5591 is continuing this one :)

PR #4758, opened 2022-04-29
test: Add e2e test for Prometheus Operator Upgrade Path

### LINKED ISSUE #4067: Add end-to-end test to verify the upgrade from the current stable version to the next one

heylongdacoder:
Fixes #4067

Signed-off-by: heylongdacoder <heylongdacoder@gmail.com>

## Description

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/test/e2e/upgradepath_test.go
+++ b/test/e2e/upgradepath_test.go
@@ -127,7 +127,7 @@ func testOperatorUpgrade(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	_, err = previousVersionFramework.CreateServiceAndWaitUntilReady(context.Background(), ns, &alertmanagerService)
+	_, err = previousVersionFramework.CreateOrUpdateServiceAndWaitUntilReady(context.Background(), ns, &alertmanagerService)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -140,7 +140,7 @@ func testOperatorUpgrade(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	_, err = previousVersionFramework.CreateServiceAndWaitUntilReady(context.Background(), ns, &prometheusService)
+	_, err = previousVersionFramework.CreateOrUpdateServiceAndWaitUntilReady(context.Background(), ns, &prometheusService)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -150,7 +150,7 @@ func testOperatorUpgrade(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	_, err = previousVersionFramework.CreateServiceAndWaitUntilReady(context.Background(), ns, &thanosRulerService)
+	_, err = previousVersionFramework.CreateOrUpdateServiceAndWaitUntilReady(context.Background(), ns, &thanosRulerService)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/test/e2e/upgradepath_test.go
+++ b/test/e2e/upgradepath_test.go
@@ -17,18 +17,12 @@ package e2e
 import (
 	"context"
 	"fmt"
-	"io"
-	"os"
-	"strings"
 	"testing"
 	"time"
 
-	operatorFramework "github.com/prometheus-operator/prometheus-operator/test/framework"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
-
-	"github.com/blang/semver/v4"
 )
 
 func testOperatorUpgrade(t *testing.T) {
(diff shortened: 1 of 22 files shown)


PhilipGough:
@heylongdacoder thanks for doing this and sorry about the lag in terms of reviewing. I have this on my radar for next week :)

test/framework/prometheus.go
```diff
@@ -305,6 +305,42 @@ func (f *Framework) MakeThanosQuerierService(name string) *v1.Service {
 	return service
 }
 
+// CreatePrometheusAndWaitUntilReadyV055x create and wait prometheus for operator v0.55.x.
+// This is needed for the upgrade path e2e test from v0.55.x to v0.56.x.
+// It can be deprecated when the latest version reach 0.57.x.
```
PhilipGough, line 310: @heylongdacoder would you mind adding additional info to the comment as to why it is needed and perhaps a `#TODO` so it does not get lost.

heylongdacoder: @PhilipGough done, please check if the additional info is clear enough. I also added another #TODO [here](https://github.com/prometheus-operator/prometheus-operator/pull/4758/files#diff-4049027b7623040b566fa582bcbf90a0c469204019a56d366134c5090720c0afR168)

test/e2e/upgradepath_test.go
```diff
@@ -0,0 +1,234 @@
...
+		t.Fatal(err)
+	}
+
+	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, ns)
+	prometheus := framework.MakeBasicPrometheus(ns, name, name, 1)
+
+	// Method to wait for prometheus ready changed since v0.56.0.
+	// Hence we need to use the old way to check if prometheus ready for operator v0.55.x
+	// TODO(heylongdacoder): this condition checking can be removed when the latest operator version reach 0.57.0
+	if currentSemVer.LT(semver.MustParse("0.57.0")) {
```
PhilipGough, line 169: @heylongdacoder .57 was release this week, so I guess we can simplify this already? :)

PhilipGough (changes requested):
needs a rebase and we need to take into account that we are already at v0.57.0 but other than that it is looking good

heylongdacoder:
@PhilipGough thanks for the review! These are the few changes that I made in the latest commit, please have a look.

- add operatorVersion, opImage, exampleDir, resourcesDir as framework struct field
- create previousVersionFramework singleton in main_test.go
- set createResourceAdmissionHooks to true for CreateOrUpdatePrometheusOperator in upgradepath_test.go to also experience the upgrade of webhook server
- for deployment and service, remain the Create function(CreateDeployment) and create a new CreateOrUpdate function(CreateOrUpdateDeploymentAndWaitUntilReady), since the method to wait for a upgrade operation to be done is different from the create operation.
- take v0.57.0 into account

test/framework/admission-webhooks.go
```diff
@@ -23,40 +23,66 @@ import (
...
+		hook.ObjectMeta.ResourceVersion = h.ObjectMeta.ResourceVersion
+
+		// MutatingWebhookConfiguration already exists -> Update
+		_, err = f.KubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Update(ctx, hook, metav1.UpdateOptions{})
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to update mutating webhook %s", hook.Name)
+		}
+	} else {
+		// MutatingWebhookConfiguration doesn't exists -> Create
+		_, err = f.KubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Create(ctx, hook, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
heylongdacoder: @PhilipGough  I agree, I will do it. But currently I have some question regarding the error message part. I saw two different pattern when we want to return the error. [One](https://github.com/prometheus-operator/prometheus-operator/blob/main/test/framework/service_account.go#L34) will return the error directly, and [another ](https://github.com/prometheus-operator/prometheus-operator/blob/main/test/framework/deployment.go#L56) will wrap the error with additional info before return it. May I know I should follow which pattern?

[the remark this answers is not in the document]
heylongdacoder: Personally, I think since we are using third party library, and we do not know the actual error message is clear enough or not, so I think is better to wrap the error with some additional info?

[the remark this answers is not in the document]
heylongdacoder: @PhilipGough understand, then I remained the way to return error. It is done as well, please have a look :)

test/framework/service.go
```diff
@@ -52,6 +53,41 @@ func (f *Framework) CreateServiceAndWaitUntilReady(ctx context.Context, namespac
...
+		service.Spec.ClusterIPs = s.Spec.ClusterIPs
+		service.Spec.IPFamilies = s.Spec.IPFamilies
+
+		// Service already exists -> Update
+		if _, err := f.KubeClient.CoreV1().Services(namespace).Update(ctx, service, metav1.UpdateOptions{}); err != nil {
+			return finalizerFn, errors.Wrap(err, fmt.Sprintf("updating service %v failed", service.Name))
+		}
+
+		// Allow controller to update the endpoints
+		time.Sleep(15 * time.Second)
```
PhilipGough, line 82: do we really need this sleep? I'd expect `WaitForServiceReady` to be eventual?

PhilipGough: Sorry for the consistent lag on reviewing this PR and thanks for continuing to see it through. 

Would checking the revision version in such cases be useful to determine if there has been an update instead of the sleep or perhaps some other mechanism? Regardless, I think in the context of this being a test function we can and should remove it unless it begins to flake!

heylongdacoder: @PhilipGough agree. Done, please have a look. :D

heylongdacoder:
@PhilipGough seems like some of my code were deleted after the merge. I wondering why there is no merge conflict. 
Is it better if I open another branch based on the current main branch, and cherry-pick the changes to that branch and create another PR? Thanks :D

PhilipGough:
Hey, yes was surprised by this too as there was no conflict. It seemed safe to merge. I guess branching off main and porting over is the best option

PR #4762, opened 2022-05-03
pkg/prometheus: exclude status from hash computation

simonpasquier:
## Description

The operator computes a hash value which is derived from different
inputs such as the Prometheus object, the operator's configuration, the
StatefulSet spec, etc. The operator records the hash value as an
annotation on the generated StatefulSet then it compares the current and
new values to decide whether an update of the StatefulSet is needed or
not.

Because the operator updates the status subresource in parallel, the
status field of the Prometheus object should not be taken into account
to compute the hash value.

The issue was indirectly detected when bumping the downstream github.com/openshift/prometheus-operator to v0.56.0. After this, we noticed more failures in the CI that could be explained by the Prometheus operator doing too many updates of the Prometheus statefulset. After reverting the bump, the CI got healthier.

## Type of change

## Changelog entry

```release-note
Avoid unnecessary updates of the Prometheus StatefulSet object.
```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1625,6 +1625,10 @@ func checkPrometheusSpecDeprecation(key string, p *monitoringv1.Prometheus, logg
 }
 
 func createSSetInputHash(p monitoringv1.Prometheus, c operator.Config, ruleConfigMapNames []string, tlsAssets *operator.ShardedSecret, ss interface{}) (string, error) {
+	// The status field is updated only by the operator hence it shouldn't be
+	// taken into account for computing the hash value.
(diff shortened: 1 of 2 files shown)


PhilipGough (approved):
/lgtm

fpetkovski (approved):

simonpasquier:
I've gathered the number of statefulset updates performed by the prometheus-operator service account from different OpenShift CI runs:
* v0.56.0 => 151 operations
* revert of v0.56.0 (e.g. v0.55.0) => 77 operations
* v0.56.0 + this PR => 96 operations

So it looks like the change is good. I'm not sure about the 25% increase between v0.55.0 and this PR but it's not doubling the number of requests either.

PR #4792, opened 2022-05-15
Add support for attach_metadata into PodMonitor

### LINKED ISSUE #4779: Add support for `attach_metadata` into ServiceMonitor/PodMonitor
**What is missing?**

As of Prometheus 2.35 we can add the `attach_metadata` option to a `kubernetes_sd_config` to allow us to add `node` labels to pods discovered in the scrape config.

**Why do we need it?**

To be able to expose node metadata as a label on pod metrics, such as for `node_exporter` metrics. This is useful for monitoring items such as `nf_conntrack` on hosts and looking up the node by hostname rather than instance IP.

**Environment**

* Prometheus Operator version: `0.56.2`

mikechengwei:
## Description

Close #4779 .

## Type of change

## Changelog entry

```release-note
Add support for attach_metadata into PodMonitor
```


pkg/apis/monitoring/v1/types.go
```diff
@@ -1189,6 +1189,14 @@ type PodMonitorSpec struct {
 	// Per-scrape limit on length of labels value that will be accepted for a sample.
 	// Only valid in Prometheus versions 2.27.0 and newer.
 	LabelValueLengthLimit uint64 `json:"labelValueLengthLimit,omitempty"`
+	//Attaches node metadata to discovered targets. Only valid for role: pod.
+	//Only valid in Prometheus versions 2.35.0 and newer.
```
fpetkovski, line 1193:
```suggestion
	// Attaches node metadata to discovered targets. Only valid for role: pod.
	// Only valid in Prometheus versions 2.35.0 and newer.
```

pkg/prometheus/promcfg.go
```diff
@@ -1547,6 +1548,13 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 		// config as well, make sure to path the right namespace here.
 		k8sSDConfig = addTLStoYaml(k8sSDConfig, "", apiserverConfig.TLSConfig)
 	}
+	if attachMetadata != nil {
```
fpetkovski, line 1551: We should make sure to attach this key only when the version is greater or equal 2.35.0. You can see an example of this pattern here: https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L192

fpetkovski (commented):
Thanks for the contribution. Overall this looks good with just a few small nits

pkg/prometheus/promcfg.go
mikechengwei: ok.

mikechengwei:
@fpetkovski plz help me review it again.

fpetkovski (approved):
This looks good to me now 👍

simonpasquier (approved):
thanks!

PR #4793, opened 2022-05-16
pkg/prometheus: continue instead of return

simonpasquier:
When reconciling shards and encountering a statefulset for which
deletion is in progress, the operator should continue to the next shard
instead of returning from the sync() method.

## Description

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1364,16 +1364,19 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		if obj != nil {
 			existingStatefulSet = obj.(*appsv1.StatefulSet)
 			if existingStatefulSet.DeletionTimestamp != nil {
-				// We want to avoid entering a hot-loop of update/delete cycles here since the sts was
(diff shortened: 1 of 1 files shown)


fpetkovski:
Should we requeue the request until all shards are running to avoid some strange edge cases?

simonpasquier:
If I'm not wrong, the [Statefulset delete handler](https://github.com/prometheus-operator/prometheus-operator/blob/ab893fdb142ea0c0675b97ee4df835f5c6d10e39/pkg/prometheus/operator.go#L1208-L1215) should be called when the statefulset is effectively deleted and this should trigger another reconciliation and the recreation of the deleted shard.
But maybe I misunderstood your remark :)

fpetkovski (approved):

simonpasquier:
> I suppose it wouldn't be straightforward to test this "recreate" logic?

yes it would be quite tricky (at least I haven't find something easy).
I've also spotted another place where we should use a `continue` so I'll post another version soon.

PR #4804, opened 2022-05-23
pkg/prometheus: fix curl exec probe

simonpasquier:
## Description

By default, the `curl` command returns with a zero exit code (success)
if the endpoint doesn't reply with a 2xx status code. As a result, the
readiness probe would succeed as soon as the Prometheus web server is
ready to respond while `/-/ready` may return "503 Service Not Available"
(e.g. the server hasn't finished loading the TSDB). During a
StatefulSet update with a 2 (or more) replicas, it means that the
StatefulSet controller may recreate the next pod while the last updated
pod hasn't finished its initialization, leading to a service disruption.

The issue is fixed by adding the `--fail` flag to the command. The man
page warns that it may still return success in some edge cases (such as
when authentication is involved) which shouldn't happen here. The
alternative would be to output the returned status code and check the
with more shell scripting but I fear that it's going to be more brittle.

Note that the issue only affects Prometheus objects with
`spec.listenLocal: true` and that 'wget' probes were unaffected.

## Type of change

## Changelog entry

```release-note
Avoided service disruption during updates for Prometheus objects with `spec.listenLocal: true` and exec probes based on the `curl` command.
```

### DIFF
--- a/pkg/operator/prober.go
+++ b/pkg/operator/prober.go
@@ -1,0 +1,27 @@
+// Copyright 2022 The prometheus-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


PhilipGough (approved):
lgtm

PR #4806, opened 2022-05-23
pkg/prometheus: add generation to computed hash

simonpasquier:
## Description

The `spec.resources` field of the Prometheus CRD embeds the
`resource.Quantity` struct which doesn't work with the hashstructure
library because some of its fields aren't public. The consequence is
that 2 Prometheus objects that differ only by their resources fields
could produce the same hash value and the operator wouldn't reconcile
the statefulset definition. To circumvent the issue, the object's
generation is added to the hash computation.

## Type of change

## Changelog entry

```release-note
Fix changes detection on Prometheus objects. 
```


pkg/prometheus/operator.go
```diff
@@ -1702,19 +1702,21 @@ func createSSetInputHash(p monitoringv1.Prometheus, c operator.Config, ruleConfi
 	hash, err := hashstructure.Hash(struct {
 		PrometheusLabels      map[string]string
 		PrometheusAnnotations map[string]string
-		PrometheusSpec        monitoringv1.PrometheusSpec
-		Config                operator.Config
-		StatefulSetSpec       appsv1.StatefulSetSpec
-		RuleConfigMaps        []string `hash:"set"`
-		Assets                []string `hash:"set"`
+		//PrometheusSpec        monitoringv1.PrometheusSpec
```
simonpasquier, line 1705: We may want to keep the PrometheusSpec field in the hash computation for now but I wanted to give the CI a try with only the generation field.

PR #4813, opened 2022-05-25
Make max_exemplars configurable

### LINKED ISSUE #4791: Add support to configure the property max_exemplars
**What is missing?**
Add support to configure the property `max_exemplars`. 
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#exemplars
```
storage:
  exemplars:
    max_exemplars: 10000000
```

**Why do we need it?**
To increase the max allowed exemplars.

faustodavid:
## Description

This creates a new StorageSettings that allows configuring the max_exemplars

Fixes #4791

## Type of change

## Changelog entry
Creates a new StorageSettings that allows configuring the max_exemplars

```release-note
Creates a new StorageSettings that allows configuring the max_exemplars
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#exemplars
```


pkg/apis/monitoring/v1/types.go
```diff
@@ -474,6 +474,17 @@ type PrometheusSpec struct {
 	// Note that if TargetLimit is lower, that value will be taken instead,
 	// except if either value is zero, in which case the non-zero value will be
 	// used.  If both values are zero, no limit is enforced.
+
+	// Storage related settings that are runtime reloadable.
+	StorageSettings *StorageSettingsSpec `json:"storageSettingsSpec,omitempty"`
```
simonpasquier, line 479: I'm not sure that we should replicate the Prometheus configuration format. We can probably avoid the `storageSettingsSpec` key.

```
spec:
  examplars:
    maxSize: 10000
```

simonpasquier (commented):
The operator needs to check that the version of Prometheus supports the examplars configuration. Also what happens if you try to configure examplars storage but forget to pass the exemplar-storage feature flag?

pkg/apis/monitoring/v1/types.go
faustodavid: I will rename it :)

faustodavid:
> The operator needs to check that the version of Prometheus supports the examplars configuration. Also what happens if you try to configure examplars storage but forget to pass the exemplar-storage feature flag?

If we don't set the  `exemplar-storage` feature flag then Prometheus ignores this setting. If we check the metric `prometheus_tsdb_exemplar_max_exemplars` it will be 0.

faustodavid:
should I add the check for the flag @simonpasquier ?

pkg/apis/monitoring/v1/types.go
faustodavid: I miss read your last comment @simonpasquier , so you want it to be:

```yaml
PrometheusSpec:
    exemplars:
        max_exemplars: 10000
```

What if they add new settings in the storage? 

If this is a blocker I will update the PR right now :)

PR #4814, opened 2022-05-25
jsonnet/mixin: add config option for group labels

ceriath:
## Description

At the moment the alert metrics are hardcoded to have only two labels, controller and namespace. This might be inconvenient. For example when managing multiple clusters, you might want to add a cluster-id label to your alerts. This change makes it possible to add custom labels similar to custom selectors.

## Type of change

## Changelog entry

```release-note
Add groupLabels config option to jsonnet mixin for alert label customization
```


simonpasquier:
lgtm

@paulfantom could you have a look?

paulfantom (approved):

PR #4831, opened 2022-06-03
*: cut v0.57.0

simonpasquier:
## Description

As the title says :)

The changelog entry emphasizes on the new v1beta1 AlertmanagerConfig CRD since this is the biggest change from the previous release. The change is an opt-in for now so this isn't disruptive for users but at the same time, we want to give the heads-up to everybody.

## Type of change

## Changelog entry

```release-note

```


paulfantom (approved):
LGTM. Small nit, could you also update the release schedule list in RELEASE.md file? You can also add me as a volunteer for the next release :D

PhilipGough (approved):
Thanks Simon.

simonpasquier:
> could you also update the release schedule list in RELEASE.md file? You can also add me as a volunteer for the next release 

thanks I'll do that in a follow-up PR to not delay this one :)

PR #4833, opened 2022-06-03
opsgenie-config: optimise warning log message during sanitize

jijotj:
## Description
This is a small change to optimise noisy warning message when sanitizing opsgenie config. The warning should be logged only if both APIKey and APIKeyFile are configured. The PR adds a condition to check the same.

## Type of change

## Changelog entry
opsgenie-config: optimise warning log message during sanitize

```release-note

```


simonpasquier (approved):
thanks!

PR #4834, opened 2022-06-03
Extend the PrometheusSpec to allow to configure the max_exemplars

### LINKED ISSUE #4791: Add support to configure the property max_exemplars

faustodavid:
## Description

Extend the PrometheusSpec to allow to configure the max_exemplars.
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#exemplars

Fixes #4791

## Type of change

## Changelog entry
```release-note
Extend the PrometheusSpec to allow to configure the max_exemplars
```


faustodavid:
hey @simonpasquier I have updated the spec, It looks as follows now.

```
spec:
  examplars:
    maxSize: 10000
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -489,6 +489,13 @@ type PrometheusSpec struct {
 	// Note that if TargetLimit is lower, that value will be taken instead,
 	// except if either value is zero, in which case the non-zero value will be
 	// used.  If both values are zero, no limit is enforced.
```
slashpai, line 491: Nit: Its unrelated to your change but above comment seems to be removed as looks to be that of `EnforcedTargetLimit` which is defined at different place

pkg/prometheus/promcfg.go
```diff
@@ -607,6 +607,11 @@ func (cg *ConfigGenerator) Generate(
 		Value: append(scrapeConfigs, addlScrapeConfigs...),
 	})
```
slashpai, line 609: Instead of checking inside method if we add check here we can avoid the method invocation itself
 `if p.Spec.EnableFeatures != nil && p.Spec.Exemplars != nil `

slashpai (commented):
Overall looks good to me , a few comments

@simonpasquier Do you think we should have separate `StorageSettings` struct to support fields that may get added in future?

pkg/prometheus/promcfg.go
faustodavid: done :)

pkg/apis/monitoring/v1/types.go
faustodavid: should I remove this comment?

simonpasquier: Not needed in this PR :)

pkg/prometheus/promcfg.go
simonpasquier: I'd be ok without the additional check here as long as appendStorageSettingsConfig() exits early if it needs to do nothing. The additional invocation isn't problematic since it's not in the "hot" path ;)

pkg/prometheus/promcfg.go
```diff
@@ -623,6 +630,41 @@ func (cg *ConfigGenerator) Generate(
...
+							Key:   "max_exemplars",
+							Value: p.Spec.Exemplars.MaxSize,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	cfg = append(cfg, storageSettings...)
```
simonpasquier, line 663: We need to check for supported versions.
```suggestion
	cfg = cg.WithMinimumVersion("2.29.0").AppendMapItem(cfg, "storage", ...)
```

pkg/prometheus/promcfg.go
```diff
@@ -623,6 +630,41 @@ func (cg *ConfigGenerator) Generate(
...
+	isExemplarsEnabled := false
+	for _, featureFlag := range p.Spec.EnableFeatures {
+		if featureFlag == "exemplar-storage" {
+			isExemplarsEnabled = true
+			break
+		}
+	}
+
+	if !isExemplarsEnabled {
+		return cfg, nil
```
simonpasquier, line 643: Since Prometheus starts without problem if the exemplars config is defined and without the `exemplar-storage` flag, the operator could also not check anything.

pkg/apis/monitoring/v1/types.go
```diff
@@ -489,6 +489,13 @@ type PrometheusSpec struct {
...
+
+	// Exemplars related settings that are runtime reloadable.
+	Exemplars *Exemplars `json:"exemplars,omitempty"`
+}
+
+type Exemplars struct {
+	MaxSize uint `json:"maxSize,omitempty"`
```
simonpasquier, line 498: Prometheus uses `int64`. We need to make it a pointer to distinguish between "use Prometheus default value" (e.g. 100000) and "disable exemplar storage" (e.g. 0).

```suggestion
	// Maximum number of exemplars stored in memory for all series.
	// If not set, Prometheus uses its default value.
	// A value of zero or less than zero disables the storage.
	MaxSize *int64 `json:"maxSize,omitempty"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -489,6 +489,13 @@ type PrometheusSpec struct {
 	// Note that if TargetLimit is lower, that value will be taken instead,
 	// except if either value is zero, in which case the non-zero value will be
 	// used.  If both values are zero, no limit is enforced.
+
+	// Exemplars related settings that are runtime reloadable.
```
simonpasquier, line 493:
```suggestion
	// Exemplars related settings that are runtime reloadable.
	// It requires to enable the exemplar storage feature to be effective.
```

simonpasquier (commented):
A few suggestions but it's going in the right direction :)

pkg/prometheus/promcfg.go
faustodavid: perfect, I will apply your comments and fix the build.

faustodavid: done :)

pkg/prometheus/promcfg.go
faustodavid: I moved the check to the function `appendStorageSettingsConfig()`

pkg/prometheus/promcfg.go
```diff
@@ -623,6 +628,26 @@ func (cg *ConfigGenerator) Generate(
...
+	}
+
+	cfg = cg.WithMinimumVersion("2.29.0").AppendMapItem(cfg, "storage", yaml.MapSlice{
+		{
+			Key: "exemplars",
+			Value: yaml.MapSlice{
+				{
+					Key:   "max_exemplars",
+					Value: p.Spec.Exemplars.MaxSize,
```
simonpasquier, line 642: (nit) not really required but more explicit

```suggestion
					Value: *p.Spec.Exemplars.MaxSize,
```

pkg/prometheus/promcfg_test.go
```diff
@@ -7647,3 +7647,89 @@ scrape_configs:
 		})
 	}
 }
+
+func TestStorageSettingMaxExemplars(t *testing.T) {
```
simonpasquier, line 7651: could you add another test case with version < 2.29?

simonpasquier (commented):
almost there :)

pkg/prometheus/promcfg.go
faustodavid: done :)

pkg/prometheus/promcfg_test.go
faustodavid: done :)

PR #4844, opened 2022-06-08
Feature/none clusterrole

zhuangzhi-cc:
## Description

Support use Role instead of ClusterRole on deployment by deploy all promethues operator related pods at a single namespace.

Add an argument for prometheus operator `has-clusterrole` default is `true`, set `false` in case of no clusterrole.

## Type of change

- [x ] `FEATURE` (non-breaking change which adds functionality)

## Changelog entry

```release-note
Support deploy prometheus operator with Role instead of ClusterRole.
```


simonpasquier:
So ideally you'd like that the operator doesn't even need the `get` permission on a single namespace?

I would prefer that the operator does the right thing out of the box without introducing a new argument. E.g. if the operator is scoped to a single namespace with the `--namespace` argument then it doesn't need to watch for namespaces?

zhuangzhi-cc:
You means just use a function like:

PR #4862, opened 2022-06-22
`ServiceMonitorSpec.JobLabel`: example in description

Dominik-K:
## Description

I think an example better helps to understand the meaning of this field.

## Type of change

- `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- `FEATURE` (non-breaking change which adds functionality)
- `BUGFIX` (non-breaking change which fixes an issue)
- `ENHANCEMENT` (non-breaking change which improves existing functionality)

## Changelog entry

```release-note
Documentation: `ServiceMonitor`'s `jobLabel` now contains an example.
```

### DIFF
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -1097,10 +1097,10 @@ type ServiceMonitorSpec struct {
 	// The value of this field will be used as the value for the `job`-label in each created metric.
 	//
 	// For example:
-	// If in `ServiceMonitor.spec.jobLabel: foo` and in `Service.metadata.labels.foo: bar`°,
-	// then each metric gets `{job="bar"}`-label. (° to be precise: in the `Endpoint`)
+	// If in `ServiceMonitor.spec.jobLabel: foo` and in `Service.metadata.labels.foo: bar`,
+	// then each metric gets `{job="bar"}`-label.
 	//
-	// Default & fallback value: the name of the respective Kubernetes `Endpoint`.
+	// Default & fallback value: the name of the `Service`.
 	JobLabel string `json:"jobLabel,omitempty"`
 	// TargetLabels transfers labels from the Kubernetes `Service` onto the created metrics.
 	TargetLabels []string `json:"targetLabels,omitempty"`
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -1093,8 +1093,12 @@ type ServiceMonitor struct {
 // ServiceMonitorSpec contains specification parameters for a ServiceMonitor.
 // +k8s:openapi-gen=true
 type ServiceMonitorSpec struct {
-	// Chooses the label of the Kubernetes `Endpoints`.
-	// Its value will be used for the `job`-label's value of the created metrics.
+	// JobLabel chooses the label of the Kubernetes `Endpoints`.
+	// The value of this field will be used as the value for the `job`-label in each created metric.
+	//
+	// For example:
+	// If in `ServiceMonitor.spec.jobLabel: foo` and in `Service.metadata.labels.foo: bar`°,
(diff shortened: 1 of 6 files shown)


pkg/apis/monitoring/v1/types.go
```diff
@@ -1093,8 +1093,12 @@ type ServiceMonitor struct {
...
 // +k8s:openapi-gen=true
 type ServiceMonitorSpec struct {
-	// Chooses the label of the Kubernetes `Endpoints`.
-	// Its value will be used for the `job`-label's value of the created metrics.
+	// JobLabel chooses the label of the Kubernetes `Endpoints`.
+	// The value of this field will be used as the value for the `job`-label in each created metric.
+	//
+	// For example:
+	// If in `ServiceMonitor.spec.jobLabel: foo` and in `Service.metadata.labels.foo: bar`°,
+	// then each metric gets `{job="bar"}`-label. (° to be precise: in the `Endpoint`)
```
simonpasquier, line 1101: IIUC the Prometheus config looks at the service's labels so I wouldn't mention endpoints. 

```suggestion
	// If in `ServiceMonitor.spec.jobLabel: foo` and in `Service.metadata.labels.foo: bar`,
	// then each metric gets `{job="bar"}`-label.
```

simonpasquier (commented):
Nice! Just one small comment.

pkg/apis/monitoring/v1/types.go
Dominik-K: @simonpasquier Yes, but AFAICS the `prometheus-operator` sets the `role: endpoints` in `func (cg *ConfigGenerator) generateServiceMonitorConfig(...)`

https://github.com/prometheus-operator/prometheus-operator/blob/f1e140a982400e7019689cef7aad93d339e64884/pkg/prometheus/promcfg.go#L1165

Otherwise the [comment below](https://github.com/prometheus-operator/prometheus-operator/pull/4862/files#diff-255c0aca616a1cdf4385c88d1d063a6c86360fedf99f7fc76bb967034e68897cR1103) must be updated:
```golang
// Default & fallback value: the name of the respective Kubernetes `Endpoint`.
```

simonpasquier: I think the next sentence should be updated indeed. The relevant code is:

https://github.com/prometheus-operator/prometheus-operator/blob/f1e140a982400e7019689cef7aad93d339e64884/pkg/prometheus/promcfg.go#L1356-L1371

simonpasquier:
I think that the description of `PodMonitor.spec.jobLabel` would need some love, feel free to submit another PR after we merge this one. And thanks for the contribution!

pkg/apis/monitoring/v1/types.go
Dominik-K: Ah right, thx. 👍

pkg/apis/monitoring/v1/types.go
```diff
@@ -1093,10 +1093,14 @@ type ServiceMonitor struct {
 // ServiceMonitorSpec contains specification parameters for a ServiceMonitor.
 // +k8s:openapi-gen=true
 type ServiceMonitorSpec struct {
-	// Chooses the label of the Kubernetes `Endpoints`.
-	// Its value will be used for the `job`-label's value of the created metrics.
+	// JobLabel chooses the label of the Kubernetes `Endpoints`.
+	// The value of this field will be used as the value for the `job`-label in each created metric.
```
simonpasquier, line 1097:
```suggestion
	// JobLabel selects the label from the associated Kubernetes service which will be used as the `job` label for all metrics.
```

Dominik-K:
@simonpasquier Thx for improving the texts 👍

slashpai:
@Dominik-K You would need to run `make generate` after updating in types.go :)

simonpasquier (approved):

PR #4863, opened 2022-06-22
Prometheus: Add additionalArgs for Prometheus and Thanos sidecar

### LINKED ISSUE #4851: Add support for additional CLI arguments to Prometheus, Alertmanager and Thanos Ruler
**What is missing?**

Users have no practical way to pass additional CLI arguments to the operands. One could use [strategic merge patch](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/strategic-merge-patch.md) to override the argument list generated by the operator but this is brittle and hacky since the operation replaces all items.

For most arguments (e.g. retention time, log level, ...), it is expected that arguments are exposed and managed by the CRDs but:
* It takes time to support arguments introduced by a new version of the operand.
* There's no plan to support hidden flags in the CRDs.

Based on the situation it sounds reasonable to allow users passing additional arguments with the following caveats:
1. The operator doesn't perform any version or validation check of the arguments.
2. There should be a clear message in the documentation that the feature is for advanced use cases and that it should be used carefully (e.g. passing an unsupported/broken argument means that the container will fail to start).
3. It's not possible to override an argument which is already exposed by the CRD (e.g. `spec.retention`).
4. It's not possible to override an argument which is managed already by the operator (e.g. [`--web.console.*`](https://github.com/prometheus-operator/prometheus-operator/blob/b4814089738e0b3576a81abc5b5549befba868c3/pkg/prometheus/statefulset.go#L353-L354)).

For the latter 2 points, the operator should fail the reconciliation when it happens.

Proposed API:

```
spec:
  additionalArgs:
  - '--foo'
  - '--bar'
```

For the Prometheus CRD, the Thanos sidecar should also support

```
spec:
  additionalArgs:
  - '--foo'
  - '--bar'
  thanos:
    additionalArgs:
    - '--foo'
    - '--bar'
```

**Why do we need it?**

#4850
https://github.com/prometheus-operator/prometheus-operator/discussions/4809
And probably other requests that I'm not aware of :)

boernd:
## Description

Add the possibility to provide `additionalArgs` for Prometheus and Thanos sidecar. Fixes #4851 

## Type of change

## Changelog entry

```release-note
Add additionalArgs field to the Prometheus CRD for Prometheus and Thanos sidecar.
```


pkg/prometheus/statefulset.go
```diff
@@ -1080,3 +1097,48 @@ func queryLogFilePath(p *monitoringv1.Prometheus) string {
...
+func verifyAdditionalArgs(operatorArgs []string, additionalArgs []string) (e error) {
+	operatorArgKeys, _ := extractArgKeys(operatorArgs)
+	additionalArgKeys, invalidArgs := extractArgKeys(additionalArgs)
+	if len(invalidArgs) > 0 {
+		return errors.Errorf("invalid argument syntax for args: %s", invalidArgs)
+	}
+
+	i := intersection(operatorArgKeys, additionalArgKeys)
+	if len(i) > 0 {
+		return errors.Errorf("invalid additionalArgs configuration for already defined args: %s", i)
```
simonpasquier, line 1140: There are 2 options
* Fail the reconciliation in case of conflict (as you did here).
* Silently drop the additional argument which is already managed by the operator.

I agree that an explicit failure is better (users won't have to wonder why the operator dropped the additional arguments) but it also means that when a new version of the operator starts managing an argument, the reconciliation will break right after the upgrade. At least, the behavior should be documented in the API.

simonpasquier (commented):
Thanks for giving it a try! I think we need to settle on the semantics and behavior but I like where it's going :)

pkg/apis/monitoring/v1/types.go
```diff
@@ -341,6 +341,10 @@ type CommonPrometheusFields struct {
 	// +listType=map
 	// +listMapKey=ip
 	HostAliases []HostAlias `json:"hostAliases,omitempty"`
+	// AdditionalArgs allows setting additional arguments for the Prometheus container.
+	// It is intended for e.g. activating hidden flags which are not supported by
+	// the dedicated configuration options yet.
```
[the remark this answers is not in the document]
boernd: Extended the description

pkg/apis/monitoring/v1/types.go
```diff
@@ -341,6 +341,10 @@ type CommonPrometheusFields struct {
...
+	AdditionalArgs []string `json:"additionalArgs,omitempty"`
```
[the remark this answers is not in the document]
boernd: Adapted accordingly

pkg/prometheus/statefulset.go
boernd: Extended docs

boernd:
@simonpasquier I have made adaptions according to your suggestions, please let me know what you think.

pkg/apis/monitoring/v1/types.go
```diff
@@ -2086,3 +2102,12 @@ type AuthorizationValidationError struct {
 func (e *AuthorizationValidationError) Error() string {
 	return e.err
 }
+
+// Argument as part of the AdditionalArgs list.
+// +k8s:openapi-gen=true
+type Argument struct {
+	// Name of the argument, e.g. scrape.discovery-reload-interval
```
simonpasquier, line 2109:
```suggestion
	// Name of the argument, e.g. "scrape.discovery-reload-interval".
	// +kubebuilder:validation:MinLength=1
```

pkg/prometheus/statefulset.go
```diff
@@ -765,22 +754,23 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
 			bindAddress = "127.0.0.1"
 		}
 
-		thanosArgs := []string{"sidecar",
-			fmt.Sprintf("--prometheus.url=%s://%s:9090%s", prometheusURIScheme, c.LocalHost, path.Clean(webRoutePrefix)),
-			fmt.Sprintf("--grpc-address=%s:10901", bindAddress),
-			fmt.Sprintf("--http-address=%s:10902", bindAddress),
+		thanosContainerArgs := []string{"sidecar"}
```
simonpasquier, line 757: I think that we can get rid of this variable and insert `"sidecar"` when building the final args list (e.g. L878 in this version)

pkg/prometheus/statefulset.go
```diff
@@ -932,14 +932,23 @@ func makeStatefulSetSpec(p monitoringv1.Prometheus, c *operator.Config, shard in
...
 	}
 
+	if len(p.Spec.AdditionalArgs) > 0 {
+		if err := verifyAdditionalArgs(promArgs, p.Spec.AdditionalArgs); err != nil {
+			return nil, err
+		}
+		promArgs = append(promArgs, p.Spec.AdditionalArgs...)
+	}
+
+	containerArgs := createContainerArgs(promArgs)
```
simonpasquier, line 942: We can have a single function to merge and return the final args list.

```suggestion
	containerArgs, err := buildArgs(promArgs, p.Spec.AdditionalArgs...)
	if err != nil {
		return nil, err
	}
	...
}
	
func buildArgs(args []monitoringv1.Argument, additionalArgs monitoringv1.Argument...) ([]string, error) {
...
}
```

pkg/prometheus/statefulset.go
```diff
@@ -1080,3 +1089,54 @@ func queryLogFilePath(p *monitoringv1.Prometheus) string {
...
+	}
+
+	return k
+}
+
+func verifyAdditionalArgs(operatorArgs []monitoringv1.Argument, additionalArgs []monitoringv1.Argument) (e error) {
+	operatorArgKeys := extractArgKeys(operatorArgs)
+	additionalArgKeys := extractArgKeys(additionalArgs)
+
+	i := intersection(operatorArgKeys, additionalArgKeys)
```
simonpasquier, line 1122: we need to take care of bool args like `--storage.tsdb.no-lockfile`/`--no-storage.tsdb.no-lockfile`.

simonpasquier (commented):
Good job!

simonpasquier (commented):
could you regenerate the docs too? not sure why GitHub doesn't complain but #4709 has merged and Documentation/api.md should be completely different now...

pkg/apis/monitoring/v1/types.go
boernd: Updated

pkg/prometheus/statefulset.go
```diff
@@ -1080,3 +1081,53 @@ func queryLogFilePath(p *monitoringv1.Prometheus) string {
 
 	return filepath.Join(defaultQueryLogDirectory, p.Spec.QueryLogFile)
 }
+
+func intersection(a, b []string) (i []string) {
+	m := make(map[string]bool)
```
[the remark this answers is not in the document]
boernd: Updated

pkg/prometheus/statefulset.go
```diff
@@ -1080,3 +1081,53 @@ func queryLogFilePath(p *monitoringv1.Prometheus) string {
...
+
+	for _, item := range a {
+		m[item] = true
+	}
+
+	for _, item := range b {
+		if _, ok := m[item]; ok {
+			i = append(i, item)
+		}
```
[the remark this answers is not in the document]
boernd: Updated

boernd:
@simonpasquier Thanks for your review, I think I have covered every comment. I'm not sure the failing e2e tests are related to my changes (maybe flaky tests)?

PR #4897, opened 2022-07-09
Update monitoring.coreos.com_alertmanagerconfigs.yaml

zanhsieh:
Signed-off-by: MH <zanhsieh@gmail.com>

## Description

See:
https://github.com/prometheus-community/helm-charts/pull/2238

## Type of change

## Changelog entry

```release-note
This will fix a pyaml issue which causes the equal sign to throw a could not determine a constructor for the tag 'tag:yaml.org,2002:value' error in kube-prometheus-stack helm chart. See:

https://github.com/prometheus-community/helm-charts/pull/2238
```

### DIFF
--- a/example/prometheus-operator-crd-full/monitoring.coreos.com_alertmanagerconfigs.yaml
+++ b/example/prometheus-operator-crd-full/monitoring.coreos.com_alertmanagerconfigs.yaml
@@ -69,7 +69,7 @@ spec:
                               if non-empty.
                             enum:
                             - '!='
-                            - =
+                            - '='
                             - =~
                             - '!~'
                             type: string
@@ -102,7 +102,7 @@ spec:
                               if non-empty.
                             enum:
                             - '!='
-                            - =
+                            - '='
                             - =~
                             - '!~'
                             type: string
@@ -4312,7 +4312,7 @@ spec:
                             if non-empty.
                           enum:
                           - '!='
-                          - =
+                          - '='
                           - =~
                           - '!~'
                           type: string
@@ -4417,7 +4417,7 @@ spec:
                               >= v0.22.0.
                             enum:
                             - '!='
-                            - =
+                            - '='
                             - =~
                             - '!~'
                             type: string
@@ -4446,7 +4446,7 @@ spec:
                               >= v0.22.0.
                             enum:
                             - '!='
-                            - =
+                            - '='
                             - =~
                             - '!~'
                             type: string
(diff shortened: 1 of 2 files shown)


fpetkovski:
Thanks for the fix. Could you try making the change [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L847) and then run `make generate`?

zanhsieh:
> Thanks for the fix. Could you try making the change [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L847) and then run `make generate`?

Hello @fpetkovski I tried:

```
	// +kubebuilder:validation:Enum="!=";"=";"=~";"!~"
        // => same
	// +kubebuilder:validation:Enum=!=;"=";=~;!~
        // => same
	// +kubebuilder:validation:Enum='!=';'=';'=~';'!~'
        // => triple quote, e.g.
                            enum:
                            - '''!='''
                            - '''='''
                            - '''=~'''
                            - '''!~'''
```
So where did I do wrong? I tried go 1.17 and 1.18.3 darwin/arm64. Same.

fpetkovski:
I wonder if the problem is in the CRD generation tooling. Are you able to modify your python code to work around the issue?

zanhsieh:
> I wonder if the problem is in the CRD generation tooling. Are you able to modify your python code to work around the issue?

Sorry @fpetkovski . Could you point out where I should change the code? Searching Makefile and all files with .py but nothing came out.

simonpasquier:
hmm this looks like a bug in pyaml if the generated YAML manifest is syntactically correct?

zanhsieh:
> hmm this looks like a bug in pyaml if the generated YAML manifest is syntactically correct?

I am out of idea where to change. Could someone point where I did wrong?

Moep90:
This issue causes extra "update" work on our side with every update..

simonpasquier:
The yaml files are generated from the Go [source code](https://github.com/prometheus-operator/prometheus-operator/blob/e4509d44549b4b371c99b826cdd48acdb8fcc956/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L848) with [controller-gen](https://book.kubebuilder.io/reference/controller-gen.html) so changing the generated files isn't going to work. I don't think that we have easy ways to tweak the output of controller-gen and as I said earlier, the output is valid YAML so the best solution is probably to fix pyaml.

Moep90:
See: https://github.com/yaml/pyyaml/issues/89

PR #4725, opened 2022-04-16
Add validation for the OpsGenie's responders

mikechengwei:
## Description

Close #4659 .

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #4900, opened 2022-07-12
Revert "Add validation for the OpsGenie's responders (#4725)"

sbchinese:
zhi na gou, ni fang dai huan wan le me, hai lai guan xing wu ke lan, sha gou.


slashpai:
Thanks for the PR
Could you please describe why you are reverting the change?

PR #4925, opened 2022-07-21
Allow podman for building images

paramite:
## Description

Enables building images also using podman.

## Type of change

## Changelog entry

```release-note
Enable building images also using podman.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ else
 	ARCH=$(GOARCH)
 endif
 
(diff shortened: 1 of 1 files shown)


PhilipGough (approved):
lgtm

I guess alias docker -> podman would also be sufficient but I don't see any problem with this change.

simonpasquier:
could you update the README.md too? I think that the `sed -i 's/docker build/podman build/g' Makefile` hack can be removed :)

PR #4937, opened 2022-07-28
Makefile: add stripped-down-crds.yaml target

simonpasquier:
## Description

Workaround for `kubectl apply` failing because of annotation being too
long.

Running `make stripped-down-crds.yaml` will generate a version of the CRDs with all description fields removed.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -227,6 +227,15 @@ generate-docs: $(shell find Documentation -type f)
 bundle.yaml: generate-crds $(shell find example/rbac/prometheus-operator/*.yaml -type f)
 	scripts/generate-bundle.sh
 
+# stripped-down-crds.yaml is a version of the Prometheus Operator CRDs with all
+# description fields being removed. It is meant as a workaround for the issue
+# that `kubectl apply -f ...` might fail with the full version of the CRDs
+# because of too long annotations field.
+# See https://github.com/prometheus-operator/prometheus-operator/issues/4355
+stripped-down-crds.yaml: $(shell find example/prometheus-operator-crd/*.yaml -type f)
+	: > $@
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -226,6 +226,15 @@ generate-docs: $(shell find Documentation -type f)
...
 	scripts/generate-bundle.sh
 
+# stripped-down-crds.yaml is a version of the Prometheus Operator CRDs with all
+# description fields being removed. It is meant as a workaround for the issue
+# that `kubectl apply -f ...` might fail with the full version of the CRDs
+# because of too long annotations field.
+# See https://github.com/prometheus-operator/prometheus-operator/issues/4355
+stripped-down-crds.yaml: $(shell find example/prometheus-operator-crd/*.yaml -type f)
+	echo >> $@
+	for f in example/prometheus-operator-crd/*.yaml; do echo '---' >> $@; $(GOJSONTOYAML_BINARY) -yamltojson < $$f | jq 'walk(if type == "object" then with_entries(select(.key | test("description") | not)) else . end)'  >> $@; done
```
jzelinskie, line 236: You forgot to convert back into YAML:

```suggestion
	for f in example/prometheus-operator-crd/*.yaml; do echo '---' >> $@; $(GOJSONTOYAML_BINARY) -yamltojson < $$f | jq 'walk(if type == "object" then with_entries(select(.key | test("description") | not)) else . end)' | $(GOJSONTOYAML_BINARY) >> $@; done
```

simonpasquier: but json is YAML :smile:
thanks, this is fixed!

jzelinskie:
This has been working for us perfectly so far. Will bake it as long as you need to be confident.

PhilipGough (approved):
@simonpasquier this lgtm but wondering if we want to include it in CI somehow going forward?

simonpasquier:
@PhilipGough we could generate the file and attach it to the release as part of the release workflow?

PhilipGough:
@simonpasquier sounds like a reasonable suggestion to me!

PhilipGough (approved):

rufreakde:
Is this "make stripped-down-crds.yaml" version available as a chart in the helm repository as well? Could not find it.

PR #4941, opened 2022-07-29
Generalize Statefulset creation for Prometheus object

paramite:
## Description

This patch moves Statefulset object creation from prometheus package to operator package and makes it to accept interface instead of Prometheus object. This will enable us reuse the same logic for a future PrometheusAgent object. 

## Type of change

## Changelog entry

```release-note
Generalize Statefulset creation from Prometheus objects to enable reuse of the logic for agent mode too.
```


pkg/prometheus/statefulset.go
```diff
@@ -97,153 +71,6 @@ func prometheusNameByShard(name string, shard int32) string {
...
-	ruleConfigMapNames []string,
-	inputHash string,
-	shard int32,
-	tlsAssetSecrets []string,
-) (*appsv1.StatefulSet, error) {
-	// p is passed in by value, not by reference. But p contains references like
-	// to annotation map, that do not get copied on function invocation. Ensure to
-	// prevent side effects before editing p by creating a deep copy. For more
-	// details see https://github.com/prometheus-operator/prometheus-operator/issues/1659.
-	p = *p.DeepCopy()
```
simonpasquier, line 114: IIUC the operator already calls DeepCopy() at the beginning of sync() so there's probably no reason to clone the object again.

https://github.com/prometheus-operator/prometheus-operator/blob/363a4a7dfcd1e20c3a4737a6ba3f45e03a6d702b/pkg/prometheus/operator.go#L1378-L1382

simonpasquier: addressed in #4950

pkg/operator/statefulset.go
```diff
@@ -0,0 +1,916 @@
...
+	podAnnotations := map[string]string{}
+	podLabels := map[string]string{
+		"app.kubernetes.io/version": version.String(),
+	}
+	// In cases where an existing selector label is modified, or a new one is added, new sts cannot match existing pods.
+	// We should try to avoid removing such immutable fields whenever possible since doing
+	// so forces us to enter the 'recreate cycle' and can potentially lead to downtime.
+	// The requirement to make a change here should be carefully evaluated.
+	podSelectorLabels := map[string]string{
+		"app.kubernetes.io/name":       "prometheus",
```
simonpasquier, line 546: "prometheus" will be variable depending on the object's kind I guess (e.g. "prometheus" or "promagent").

pkg/operator/statefulset.go
```diff
@@ -0,0 +1,916 @@
...
+		"app.kubernetes.io/managed-by": "prometheus-operator",
+		"app.kubernetes.io/instance":   nc.BaseName(),
+		"prometheus":                   nc.BaseName(),
```
simonpasquier, line 549: same here

pkg/operator/statefulset.go
```diff
@@ -0,0 +1,916 @@
...
+		shardLabelName:                 fmt.Sprintf("%d", shard),
+		prometheusNameLabelName:        nc.BaseName(),
```
simonpasquier, line 551: This label is used to map statefulsets <-> prometheus resources and delete statefulsets in excess when scaling down the number of shards (see [here](https://github.com/prometheus-operator/prometheus-operator/blob/ae46c0810875b03811957e39adc436db13a6cb5f/pkg/prometheus/operator.go#L1513-L1534)). If we have a `foo` Prometheus resource and a `foo` PrometheusAgent resource then the operator may delete more statefulsets than needed.
We would need a different label for the agent resources or an additional label like `operator.prometheus.io/kind` to make them unique.

cmd/operator/main_test.go
```diff
@@ -32,7 +34,8 @@ func TestNamespacesType(t *testing.T) {
 	}
 
 	ns = namespaces{}
-	ns.Set(val)
+	err = ns.Set(val)
+	require.NoError(t, err)
```
simonpasquier, line 38: did you add this to pass golangci-lint locally?

.golangci.yml
```diff
@@ -18,3 +18,7 @@ issues:
 linters-settings:
   errcheck:
     exclude: scripts/errcheck_excludes.txt
+
+issues:
+  exclude:
+    - composite
```
simonpasquier, line 24: why is it needed?

pkg/operator/statefulset.go
paramite: ack, will update.

pkg/operator/statefulset.go
paramite: ack

pkg/operator/statefulset.go
paramite: Ah, got it. Will update.

cmd/operator/main_test.go
paramite: Linter here started complaining on that AFAIR.

.golangci.yml
paramite: This is required, so that CI stops complaining on wrapping Prometheus object to PrometheusServer. Eg. `operator.PrometheusServer(&promv1.Prometheus)` is being signaled by govet [1].

[1] https://github.com/alessio/golangci-lint/commit/68b8aaacd09b1591ddc982d5dfa5815f19b214a2

pkg/operator/statefulset.go
```diff
@@ -0,0 +1,916 @@
...
+				podAnnotations[k] = v
+			}
+		}
+	}
+
+	for k, v := range podSelectorLabels {
+		podLabels[k] = v
+	}
+
+	podAnnotations["kubectl.kubernetes.io/default-container"] = "prometheus"
```
JoaoBraveCoding, line 571: Maybe here the same from [comment](https://github.com/prometheus-operator/prometheus-operator/pull/4941/files#r936709304) also applies

mrunge:
Martin, would you be able to resolve the initial review comments?

paramite:
This PR is still WIP, but unfortunately I don't have much time any more to finish it. Feel free to close on your will.

ArthurSens:
Hi @paramite, @simonpasquier, @JoaoBraveCoding 

Do you mind if I continue this one, but re-using the same strategy I'm using at https://github.com/prometheus-operator/prometheus-operator/pull/5310?

paramite:
@ArthurSens I'm perfectly fine with it. Feel free to take over this PR.

JoaoBraveCoding:
Sounds good to me go for it @ArthurSens! 💪

simonpasquier:
Closing as it's replaced by #5334

Issue #4975, opened 2022-08-19
Reported conditions don't align with established conventions


thetechnick:
We are implementing a higher-level controller on top of github.com/rhobs/observability-operator, as part of github.com/openshift/addon-operator.
Now we want to use the new status conditions posted by OO and Prometheus Operator, but this left us confused as these conditions seem to behave different to other conditions.

1) Condition Status is either "True", "False" or "Unkown"
Prometheus Operator and also Observability Operator are reporting an extra "Degraded" status, which is doesn't map to any existing API convention and makes interpreting status harder.
Is there a difference between: Type: "Available", Status: "False" and Status: "Degraded"?
Is "Degraded" referring to the condition itself or the operator being degraded and thus should not be taken for granted?
[Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)

2) What state does the "Reconciled" condition represent?
If the controller was able to post that condition, it - by definition - reconciled _something_.
Whether the status is up to date is already represented by reporting .observedGeneration.

3) (nit/suggestion) Conditions could use standardized conditions from apimachinery and their helper functions to make handling easier:
[Standardized Conditions from k8s/apimachinery](https://github.com/kubernetes/apimachinery/blob/kubernetes-1.24.4/pkg/apis/meta/v1/types.go#L1467)
[Standardized Condition helper functions from k8s/apimachinery](https://github.com/kubernetes/apimachinery/blob/kubernetes-1.24.4/pkg/api/meta/conditions.go)

This is not a blocker for us, but to quote from API conventions:
> Conditions are most useful when they follow some consistent conventions

Right now it's very easy to miss-interpret status reported by Prometheus Operator and by extension OO.

Thanks for reading!

JoaoBraveCoding:
Hi Nico 👋 having read your issue I agree with you that we should try and follow the conventions! I'll review your PR but @simonpasquier will be the best person to discuss the design decisions. Thanks for the contribution!

simonpasquier:
> Is there a difference between: Type: "Available", Status: "False" and Status: "Degraded"?
> Is "Degraded" referring to the condition itself or the operator being degraded and thus should not be taken for granted?

Yes there's a difference:
* (Type: Available, Status: False) means that the Prometheus service is totally unavailable (e.g. no pods are ready).
[...]

PR #4976, opened 2022-08-19
Change status conditions to follow conventions

thetechnick:
## Description

Currently prometheus-operator status conditions don't follow kubernetes
api conventions. This PR replaces the custom condition type with a
definition from kubernetes/apimachinery.

It also uses helper functions from apimachinery, where appropriately to
reduce code duplication.

This change is to be considered API breaking(!), dependent on release
management and consistency guarantees.

Follow up from #4975 

Signed-off-by: Nico Schieder <nschieder@redhat.com>

## Type of change

## Changelog entry

```release-note
changed status conditions to be consistent with kubernetes api conventions
```


JoaoBraveCoding (approved):
LGTM 👍

apahim:
@simonpasquier can you please take a look?

simonpasquier (commented):
While I see where you're coming from, my first remark is that we can't change the actual API because the CRD version is v1. Having said that, I'm fine with adding things like `observedGeneration` and improving the API documentation.

thetechnick:
Closing in favor of a follow up PR just containing observedGeneration to not break backwards compatibility.
Thanks everyone for having a look!

Issue #4968, opened 2022-08-15
Secret or other resource in prometheus ns changes would cause additional secret loaded request secret list frequently 


gitgaoxiang:
**What happened?**
In my cluster, I have many secret and prometheus pod in the same ns，I often find that the K8S APiserver resource usage metrics are too high，I checked the API logs，found this.

<img width="1261" alt="lQLPJxaJfDQbpAzMus0J2rCY3n3sw1x1TgLiUm8qgIcA_2522_186" src="https://user-images.githubusercontent.com/19486917/184602947-a758bd96-cf0c-4eb4-9bbf-687f5fc6800b.png">

**Why is this a problem ?**
I check the pkg/prometheus/operator code ，found in handle method，if secret 、cm 、serviceMonitor、podmonitor or other resource have changed , it would cause additional secret loaded,  This will result in a secret list request。
![image](https://user-images.githubusercontent.com/19486917/184604844-76db2c55-a329-4639-811b-c8b11c379fe4.png)
if ns have many secret , this would cause  k8s apiserver metrics too high

Can I submit code to solve this problem

slashpai:
Updating the secret is needed when changes are made to ServiceMonitors/PodMonitors/Probe or the Prometheus resource itself.

More details in https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/design.md#prometheus

gitgaoxiang:
@slashpai whether it can be use sClient.Get instead of sClient.List. if so , when ns have a lot of secret ,it can reduce the pressure on the apiserver

JoaoBraveCoding:
Hello @gitgaoxiang I think I see your point. You basically want to replace the List with a set of Get's, from my side it looks reasonable. Unless I'm missing anything

> Can I submit code to solve this problem

I think you can always do it, you just can't expect it to always be merged

simonpasquier:
Closed by 3ad523d1a4

PR #4986, opened 2022-08-29
replace the List with a set of Get's

gitgaoxiang:
## Description

https://github.com/prometheus-operator/prometheus-operator/issues/4968

## Type of change

## Changelog entry

```release-note
change secret list to get , Resolve the pressure on Apiserver when NS secret are too many
```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -1958,14 +1958,13 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *mon
 		return errors.Wrap(err, "selecting Probes failed")
 	}
 	sClient := c.kclient.CoreV1().Secrets(p.Namespace)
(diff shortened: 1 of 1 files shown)


JoaoBraveCoding:
LGTM 👍 but a maintainer has to review it!
Thank you for the contribution 🥇

fpetkovski:
I am not sure if this reduces the load on the API server, since a `List` request should be going to the cache, while a `Get` request would often go directly to API server.

PhilipGough:
Agree with @fpetkovski and I don't think we could blindly merge this without some metrics or benchmarks to show an improvement.

IIUC, the issue stems from you deploying your monitoring components in the same namespace as your other workloads!

gitgaoxiang:
@PhilipGough We do create other services under the monitor ns, because resources in ns change frequently, the operator requests the apiserver frequently. this leads to insufficient memory.  However,can we optimizing list perform scene，request only when additional is configured. I submit merge for this.

PR #4988, opened 2022-08-30
Fix 4944 invalid vol name

sthaha:
## Description

Previously, adding secrets to Prometheus resource longer than 50
characters resulted in generation of statefulset that have
`volume.name` larger than 63 characters. This doesn't conform to
RFC-1123 (<= 63 chars) requirement and resulted in pod not getting
created.

This patch fixes the issue by ensuring that the volume name conform to
the requirement. Additionally it fixes an issue with SanitizeVolumeName
which used to truncate the name at DNS1123LabelMaxLength thus returing
same name for `<63-char>-foo` and `<63-char>-bar` to return unique name
by computing the sha1 and appending to the end.

Bug: https://github.com/prometheus-operator/prometheus-operator/issues/4944

Signed-off-by: Simon Pasquier <spasquie@redhat.com>
Signed-off-by: Sunil Thaha <sthaha@redhat.com>

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [x] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

Fixes prometheus/alertmanager pods not getting created when secrets name exceeds 63 characters.

```release-note
Fixes prometheus/alertmanager pods not getting created when secrets name exceeds 63 characters.
```

PR #5000, opened 2022-09-07
Fixed secret and configmap volume names that need to be mounted in additional containers

simonpasquier:
## Description

#4988 (84ffc50f68f619c0cb2e3e6999ea267a259b4635) introduced a change breaking users that pass additional secrets & configmaps and want to mount them in additional containers. In practice, the aforementioned commit appends a hash suffix to the volume name which makes it impossible for a user to reference the volume in a sidecar container.

This PR reverts to the old behavior for user-provided secrets/configmaps: the volume name is identical to the object name, except that invalid characters are replaced by `-` and the string is truncated to 63 chars.

## Type of change

## Changelog entry

```release-note
Fixed secret and configmap volume names that need to be mounted in additional containers.
```


PhilipGough (approved):
lgtm

PR #4931, opened 2022-07-22
Add CAP_FOWNER capability to thanos-sidecar container

bingikarthik:
## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

Stating promtheus-operator v0.56.0, by default thanos-sidecar container failed to upload blocks to object storage as explained in https://github.com/prometheus-operator/prometheus-operator/issues/4928. This PR adds necessary permissions to thanos-sidecar container under securityContext level so that hard link creation action will be successful as it was needed before uploading block to object storage

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [x] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Adds CAP_FOWNER capability to thanos-sidecar container
```

PR #5004, opened 2022-09-09
pkg/prometheus: add CAP_FOWNER capability to Thanos only if needed

simonpasquier:
## Description

When configuring the Thanos sidecar to upload blocks to object storage, the container needs to the CAP_FOWNER capability. Otherwise it isn't required. This is a small improvement on top of #4931 to avoid granting more capabilities than needed.

## Type of change

## Changelog entry

```release-note
Removed CAP_FOWNER capability for the Thanos sidecar if it isn't needed.
```


JoaoBraveCoding (approved):
LGTM 👍

PhilipGough (approved):
lgtm

PR #5008, opened 2022-09-10
Update webhook examples and docs to always use service port 443

dobesv:
## Description

While setting this up I noticed that the examples used a mix of not specifying the port (meaning apiserver uses port 443) and specifying port 8443.  

This meant that following the examples resulted in a broken setup because api server was connecting to an undefined port 443 on the service for the webhooks.

Since there were more cases of the default port, my suggestion here is to standardize on that for the docs and examples.

## Type of change

## Changelog entry

```release-note
- Fix examples / docs for webhook to consistently use the same service port (443)
```

### DIFF
--- a/Documentation/user-guides/webhook.md
+++ b/Documentation/user-guides/webhook.md
@@ -63,10 +63,10 @@ spec:
         - mountPath: /etc/tls/private
           name: tls-certificates
           readOnly: true
-     volumes:
-     - name: tls-certificates
-       secret:
-         secretName: admission-webhook-certs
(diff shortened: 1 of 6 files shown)


simonpasquier:
Files from the examples/ directory are generated from [scripts/generate/admission-webhook.jsonnet](https://github.com/prometheus-operator/prometheus-operator/blob/main/scripts/generate/admission-webhook.jsonnet). I'm not against standardizing on port 443 but you'd need to update the config in this file and regenerate the examples.

PR #5010, opened 2022-09-11
feat : enable support for "hostNetwork" in Prometheus CRD

### LINKED ISSUE #2630: Support hostNetwork api in prometheus and alertmanager crds

Rajpratik71:
## Description

Enabled support for "hostNetwork" in Prometheus Operator for Prometheus

Needed in many scenarios like:
1. To bypass the CNI network
2. In general to reduce the metrics traffic overhead on CNI network in clusters with large no of Nodes.

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

## Type of change

## Changelog entry

```release-note
feat : enable support for "hostNetwork" in Prometheus CRD
```

Fixes #2630

### DIFF
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -336,6 +336,10 @@ type CommonPrometheusFields struct {
 	// to be excluded from enforcing a namespace label of origin.
 	// Applies only if enforcedNamespaceLabel set to true.
 	ExcludedFromEnforcement []ObjectReference `json:"excludedFromEnforcement,omitempty"`
+	// Use the host's network namespace if true.
+	// Make sure to understand the security implications if you want to enable it.
+	// When hostNetwork is enabled, this will set dnsPolicy to ClusterFirstWithHostNet automatically.
+	HostNetwork bool `json:"hostNetwork,omitempty"`
 }
 
 // +genclient
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -215,6 +215,10 @@ func makeStatefulSet(
 
 	statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes, p.Spec.Volumes...)
 
+	if p.Spec.HostNetwork {
+		statefulset.Spec.Template.Spec.DNSPolicy = v1.DNSClusterFirstWithHostNet
+	}
+
 	return statefulset, nil
 }
 
@@ -1011,6 +1015,7 @@ func makeStatefulSetSpec(
 				Affinity:                      p.Spec.Affinity,
 				TopologySpreadConstraints:     p.Spec.TopologySpreadConstraints,
 				HostAliases:                   operator.MakeHostAliases(p.Spec.HostAliases),
+				HostNetwork:                   p.Spec.HostNetwork,
 			},
 		},
 	}, nil
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -2125,6 +2125,19 @@ Applies only if enforcedNamespaceLabel set to true.</p>
 </tr>
 <tr>
 <td>
+<code>hostNetwork</code><br/>
+<em>
+bool
+</em>
+</td>
+<td>
+<p>Use the host&rsquo;s network namespace if true.
+Make sure to understand the security implications if you want to enable it.
+When hostNetwork is enabled, this will set dnsPolicy to ClusterFirstWithHostNet automatically.</p>
+</td>
+</tr>
+<tr>
+<td>
 <code>baseImage</code><br/>
 <em>
 string
@@ -5661,6 +5674,19 @@ to be excluded from enforcing a namespace label of origin.
 Applies only if enforcedNamespaceLabel set to true.</p>
 </td>
 </tr>
+<tr>
+<td>
+<code>hostNetwork</code><br/>
+<em>
+bool
+</em>
+</td>
+<td>
+<p>Use the host&rsquo;s network namespace if true.
+Make sure to understand the security implications if you want to enable it.
+When hostNetwork is enabled, this will set dnsPolicy to ClusterFirstWithHostNet automatically.</p>
+</td>
+</tr>
 </tbody>
 </table>
 <h3 id="monitoring.coreos.com/v1.Duration">Duration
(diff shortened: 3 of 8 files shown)


Rajpratik71:
Failing build of "checks / Check Documentation formatting and links (pull_request)" is unrelated, i checked it is failing while opening a link which is taking long time while opening.

slashpai:
There is already discussion around `hostNetwork` support  in https://github.com/prometheus-operator/prometheus-operator/issues/2630 and I see that is added in agenda for next [office hours meeting](https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#office-hours-meetings)

pkg/apis/monitoring/v1/types.go
```diff
@@ -334,6 +334,8 @@ type CommonPrometheusFields struct {
 	// to be excluded from enforcing a namespace label of origin.
 	// Applies only if enforcedNamespaceLabel set to true.
 	ExcludedFromEnforcement []ObjectReference `json:"excludedFromEnforcement,omitempty"`
+	// Use the host networking
```
simonpasquier, line 337:
```suggestion
	// Use the host's network namespace if true.
	// Make sure to understand the security implications if you want to enable it. 
```

simonpasquier (commented):
I've never used `hostNetwork: true` but I think that the operator needs to modify the container's ports defintion.

From the ContainerPort documentation:

> hostPort | Number of port to expose on the host. If specified, this must be a valid port number, 0 < x < 65536. If HostNetwork is specified, this must match ContainerPort. Most containers do not need this.

pkg/apis/monitoring/v1/types.go
Rajpratik71: Done.

Rajpratik71:
> I've never used `hostNetwork: true` but I think that the operator needs to modify the container's ports defintion.
> 
> From the ContainerPort documentation:
> 
> > hostPort | Number of port to expose on the host. If specified, this must be a valid port number, 0 < x < 65536. If HostNetwork is specified, this must match ContainerPort. Most containers do not need this.

HI @simonpasquier ,

Please see the example at https://gist.github.com/Rajpratik71/632321a6fb8c6acc17593e2c887c45a9

When we specify "hostNetwork" for the Pod, there is no need to specify the "hostPort" seperately as when using "hostNetwork" , whole network namespace of Host will be shared, which will also expose the port.

See, below

image


Although, "hostNetwork" requires "dnsPolicy" to be set as "ClusterFirstWithHostNet".

So, raised a PR #5027 to "enable support for "dnsPolicy" in Prometheus CRD".

simonpasquier:
> Although, "hostNetwork" requires "dnsPolicy" to be set as "ClusterFirstWithHostNet".
> So, raised a PR https://github.com/prometheus-operator/prometheus-operator/pull/5027 to "enable support for "dnsPolicy" in Prometheus CRD".

After reviewing more carefully, I realize that we may not need to expose dnsPolicy in the CRD. If I understand correctly,  `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet` go hand in hand so the Prometheus operator could set dnsPolicy to the correct value automatically.
In other words, is there any reason why you'd want to set dnsPolicy to another value than `ClusterFirstWithHostNet` with host networking?

Rajpratik71:
> > Although, "hostNetwork" requires "dnsPolicy" to be set as "ClusterFirstWithHostNet".
> > So, raised a PR #5027 to "enable support for "dnsPolicy" in Prometheus CRD".
> 
> After reviewing more carefully, I realize that we may not need to expose dnsPolicy in the CRD. If I understand correctly, `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet` go hand in hand so the Prometheus operator could set dnsPolicy to the correct value automatically. In other words, is there any reason why you'd want to set dnsPolicy to another value than `ClusterFirstWithHostNet` with host networking?

Hi @simonpasquier , I created #5027 for the same purpose to configure the dnsPolicy . When it will get merged , i will rebase this branch to automate the process of setting "dnsPolicy" to "ClusterFirstWithHostNet" in case of "hostNetwork".

simonpasquier:
Sorry if I wasn't clear but my question is: do we need #5027 at all? What if the operator is the only one responsible for setting dnsPolicy?
In practice:
* `spec.hostNetwork: false` (or not set) in the Prometheus CR, the operator doesn't set anything for `spec.template.spec.dnsPolicy` on the statefulset.
* `spec.hostNetwork: true`, the operator always sets `spec.template.spec.dnsPolicy: ClusterFirstWithHostNet` on the statefulset.

pkg/prometheus/statefulset.go
```diff
@@ -215,6 +215,10 @@ func makeStatefulSet(
 
 	statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes, p.Spec.Volumes...)
 
+	if p.Spec.HostNetwork == boolTrue {
```
simonpasquier, line 218:
```suggestion
	if p.Spec.HostNetwork {
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -336,6 +336,10 @@ type CommonPrometheusFields struct {
 	// to be excluded from enforcing a namespace label of origin.
 	// Applies only if enforcedNamespaceLabel set to true.
 	ExcludedFromEnforcement []ObjectReference `json:"excludedFromEnforcement,omitempty"`
+	// Use the host's network namespace if true.
+	// Make sure to understand the security implications if you want to enable it.
+	// When hostNetwork is enabled, this will set dnsPolicy to ClusterFirstWithHostNet automatically
```
simonpasquier, line 341:
```suggestion
	// When hostNetwork is enabled, this will set dnsPolicy to ClusterFirstWithHostNet automatically.
```

simonpasquier (commented):
can you add a unit test with "hostNetwork: true"?

Rajpratik71:
> can you add a unit test with "hostNetwork: true"?

Hi @simonpasquier , added the test for "hostNetwork: true".

Also, tested the "dnsPolicy" . it is working fine as expected.

image

simonpasquier (commented):
thanks!

PR #5045, opened 2022-09-22
pkg/prometheus: add `{grpc,http}ListenLocal` for Thanos sidecar

simonpasquier:
## Description

When `spec.thanos.listenLocal` is true, both gRPC and HTTP endpoints are configured to listen on the loopback interface. It is sometimes needed to listen on all interfaces for gRPC and on the loopback interface only for HTTP.

This commit introduces the `spec.thanos.grpcListenLocal` and `spec.thanos.httpListenLocal` fields to enable all possible use cases. `spec.thanos.listenLocal` is marked as deprecated since the new fields offer equivalent possibilities with fine-grained controls.

## Type of change

## Changelog entry

```release-note
Added `grpcListenLocal` and `httpListenLocal` fields to the Thanos sidecar configuration.
```

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -11769,8 +11769,36 @@ bool
 </em>
 </td>
 <td>
-<p>ListenLocal makes the Thanos sidecar listen on loopback, so that it
-does not bind against the Pod IP.</p>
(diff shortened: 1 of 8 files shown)


PhilipGough (approved):
lgtm

Issue #4816, opened 2022-05-25
Prometheus jobs generated by PodMonitors do not exclude non-running pods


jutley:
**What happened?**

When using PodMonitors, it appears that the generated Prometheus jobs do not filter out non-running pods.

In some cases, this can have really ugly consequences. In our case, we had a handful of succeeded/failed pods which all had the same pod IP. Since they were completed, the IP was available to assigning, and it happened to get assigned to kube-state-metrics. This meant that we had many non-running pods that Prometheus was trying to scrape, and based on the target IP and port (which happened to match kube-state-metrics), we ended up 5 active scrapes of kube-state-metrics. This caused the Prometheus to crash because kube-state-metrics returned _way_ more data than the intended target.

**Did you expect to see something different?**

I expected to see that non-running pods were intentionally removed in the job relabellings. This can be done by filtering based on the `__meta_kubernetes_pod_phase` label.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a CronJob that runs every minute which does basically nothing. Make sure it keeps around a few succeeded jobs.
2. Create a PodMonitor that targets the CronJob's pods.
3. Check the Prometheus targets page. You'll see that the completed pods are failing to be scraped.

**Environment**

* Prometheus Operator version:

v0.47.1. It's fairly old, but I couldn't find any evidence in the changelog or source code that this behavior has been updated in newer versions.

* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.2", GitCommit:"8b5a19147530eaac9476b0ab82980b4088bbc1b2", GitTreeState:"clean", BuildDate:"2021-09-15T21:31:32Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.12", GitCommit:"696a9fdd2a58340e61e0d815c5769d266fca0802", GitTreeState:"clean", BuildDate:"2022-04-13T19:01:10Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:

Our own deployment (not using standard tools) onto AWS.

* Manifests:

```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  labels:
    prometheus: apps
  name: commissioner
  namespace: commissioner
spec:
  podMetricsEndpoints:
[...]

Issue #4877, opened 2022-06-29
New field in monitors to auto include ready status


snuggie12:
**What is missing?**
Curious if there's any interest in adding a field to `PodMonitor` and `ServiceMonitor` that would automatically add relabeling which checks if a resource is ready. 

**Why do we need it?**
Some of our devs get scraped early and that is messing with our progressive delivery system. Adding relabelling that uses `__meta_kubernetes_endpoint_ready` can fix that issue. Not sure we _need_ it since we can get most of our devs to update easily enough, but figured I'd bring it up to see if others have interest in making it a built in feature.

**Environment**

* Prometheus Operator version:

    `prometheus-operator:v0.55.1`
    
    

**Anything else we need to know?**:

snuggie12:
Yeah, it's similar but the endpoint maps back to servicemonitor which is what we use.

I also see loads of people have issues of targets not leaving where as this is about starting scraping too early. Didn't know if my ticket should be separate or not but happy to close and watch that issue if that'll track both podmonitor and servicemonitor.

slashpai:
I am ok with keeping this issue open but we could follow same approach for both.
Would you like to contribute for the fix?

simonpasquier:
I think that #4816 would also apply in your case but we can keep this one open until we have a clearer picture.

adabuleanu:
I ended up on this thread because I have experienced a similar issue with a slow starting pod  ~ 2 minutes (with all the probes configured), so I ended up with a lot of 503 error codes until the pod transitioned into ready state. I expected that the default behavior should have been to scrape ready endpoints. Analysing this thread and  https://github.com/prometheus-operator/prometheus-operator/issues/4816, I think they do not achieve the same result. In https://github.com/prometheus-operator/prometheus-operator/issues/4816, if I set `__meta_kubernetes_pod_phase` to `Running` (according to the [Pod Lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) this phase include containers that are in the process of starting) , I will still experience my issue (my pod takes 2 minutes to start and if I have scraping configured for 30 seconds, I will get 3-4 503 error codes). But, if I configure `__meta_kubernetes_endpoint_ready`, the scraping will not start until the pod is [Ready](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions) (which means it waits for all k8s condition + readinessProbe + startupProbe).

[...]

PR #5049, opened 2022-09-25
Add a filter for non-running pods

SuperQ:
## Description

Add a boolean option to PodMetricsEndpoint to filter out non-running phases.
* Enable by default.

Fixes: https://github.com/prometheus-operator/prometheus-operator/issues/4816

## Type of change

## Changelog entry

```release-note
- Add a filter to PodMonitors to filter out non-running pods
```


pkg/apis/monitoring/v1/types.go
```diff
@@ -1373,6 +1373,9 @@ type PodMetricsEndpoint struct {
 	FollowRedirects *bool `json:"followRedirects,omitempty"`
 	// Whether to enable HTTP2.
 	EnableHttp2 *bool `json:"enableHttp2,omitempty"`
+	// Drop pods that are not running. (Failed, Succeeded, Unknown).
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
+	FilterRunning *bool `json:"filterRunning,omitempty"`
```
PhilipGough, line 1378: @SuperQ do we want to comment that this is enabled by default? 

Also wondering that since it is enabled by default does it make sense to be a pointer? Perhaps alternatively we can use the `+kubebuilder:default` annotation?

SuperQ: Sure, I can do that.

As for the pointer, I wasn't quite sure how to make a bool true by default here.

simonpasquier: I'm not sure that `+kubebuilder:default` will play nice with a boolean field (`omitempty` might also come in the way) but it's worth checking.

SuperQ: Yea, I was tinkering around with this a bit and couldn't get it to work. The code generator always wants to make bools into pointers.

PhilipGough: From the [docs](https://book.kubebuilder.io/reference/markers/crd-validation.html)

```
// +kubebuilder:default
any
sets the default value for this field.
A default value will be accepted as any value valid for the field. Formatting for common types include: boolean: true, string: Cluster, numerical: 1.24, array: {1,2}, object: {policy: &#34;delete&#34;}). Defaults should be defined in pruned form, and only best-effort validation will be performed. Full validation of a default requires submission of the containing CRD to an apiserver.
```

but yeah, lets not get hung up on this, once we document the default value I think we are good.

simonpasquier (approved):
LGTM. We can follow up in another PR to do the same for ServiceMonitors.

PhilipGough:
@SuperQ can you run `make format generate`

slashpai:
Just adding a note here we have https://github.com/prometheus-operator/prometheus-operator/issues/4877 logged for  service monitor :)

PR #5084, opened 2022-10-11
add alertmanagerConfigMatcherStrategy toggle to disable namespace matchers

Fr3574:
## Description

This fix adds a toggle to the Alertmanager which disables the automatic generation of namespace matchers.
It resolves this feature request: https://github.com/prometheus-operator/prometheus-operator/issues/3737

## Type of change

## Changelog entry

```release-note
Add alertmanagerConfigMatcherStrategy to Alertmanager CRD in order to disable auto-generated namespace matchers.
```


PhilipGough:
@Fr3574 there has already been lengthy discussions around this and there is also already the ability to take care of this as described in https://github.com/prometheus-operator/prometheus-operator/issues/3737#issuecomment-1069866807

Fr3574:
Hi @PhilipGough, the commit you are referring to only enables this feature though for global AlertmanagerConfigs. If I however want to use e.g. a global aonfig and several other "normal" configs then for the other ones the namespace label is still enforced.

jojotong:
@Fr3574 You can put all config(which you don't want to force add namespace label) in the global AlertmanagerConfig.

pkg/alertmanager/amcfg.go
```diff
@@ -298,9 +299,16 @@ func (cb *configBuilder) enforceNamespaceForInhibitRule(ir *inhibitRule, namespa
 
 // enforceNamespaceForRoute modifies the route configuration to match alerts
 // originating only from the given namespace.
-func (cb *configBuilder) enforceNamespaceForRoute(r *route, namespace string) *route {
+func (cb *configBuilder) enforceNamespaceForRoute(r *route, namespace string, enabled bool) *route {
```
simonpasquier, line 302: The same thing would be needed for inhibition rules. Rather than adding a new bool parameter to all `enforceNamespaceFor...()` methods, it might be easier to refactor configBuilder and abstracting away the (lack of) enforcement logic into a separate component.

```golang
type enforcer interface {
        processRoute(*route) *route
        processInhibitRule(*inhibitRule) *inhibitRule
}

// No enforcement
type noopEnforcer struct{}
func (ne *noopEnforce) processRoute(r *route) *route { return r }
func (ne *noopEnforce) processInhibitRule(ir *inhibitRule) *inhibitRule { return ir }

// Enforcing the namespace label
type namespaceEnforcer struct { namespace string }
func (ne *namespaceEnforcer) processRoute(r *route) *route {
  // similar to enforceNamespaceForInhibitRule
}
func (ne *namespaceEnforcer) processInhibitRule(ir *inhibitRule) *inhibitRule {
  // similar to enforceNamespaceForRoute
}

type configBuilder struct {
        ...
        enforcer  enforcer
}

func (cb *configBuilder) convertRoute(in *monitoringv1alpha1.Route, crKey types.NamespacedName) *route {
   ...
   return cb.enforcer.processRoute(ir)
}

func (cb *configBuilder) convertInhibitRule(in *monitoringv1alpha1.InhibitRule) *inhibitRule {
    ...
    return cb.enforcer.processInhibitRule(ir)
}
```

simonpasquier (commented):
Overall I'm ok with adding the functionality. I've made some comments on how the API could be made more generic/future-proof. I've also have suggestions on the implementation. Let me know if this makes sense :)

pkg/alertmanager/amcfg.go
Fr3574: This pattern looks quite nice, I modified it so e.g. `cb.enforcer.processRoute` will no be called on the returned route in `convertRoute` but like this:
```
subRoutes = append(subRoutes,
    cb.enforcer.processRoute(
        cb.convertRoute(
	    amConfigs[amConfigIdentifier].Spec.Route,
	    crKey,
	),
    ),
)
```
This ensures that on routes from a global config (so the current behaviour is kept) and on child routes (seems to be obsolete) no route processing will be applied.

Fr3574:
@simonpasquier 
I overall agree with your suggestions as this enhances the code quality of the change. Let me know if this looks good to you?

Fr3574:
@jojotong You're right, we currently use the global config. We reached however a point where maintenance of this one file got quite messy and would really like to split it up. We're in a setup where we have teams deploying multiple services and therefore IMO it only makes sense that they maintain their AlertmanagerConfigs also themselves.

pkg/apis/monitoring/v1/types.go
```diff
@@ -2022,6 +2022,9 @@ type AlertmanagerSpec struct {
 	ForceEnableClusterMode bool `json:"forceEnableClusterMode,omitempty"`
 	// AlertmanagerConfigs to be selected for to merge and configure Alertmanager with.
 	AlertmanagerConfigSelector *metav1.LabelSelector `json:"alertmanagerConfigSelector,omitempty"`
+	// The AlertmanagerConfigMatcherStrategy defines how to add additional matchers to the AlertmanagerConfigs.
```
simonpasquier, line 2025:
```suggestion
	// The AlertmanagerConfigMatcherStrategy defines how AlertmanagerConfig objects match the alerts.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -2022,6 +2022,9 @@ type AlertmanagerSpec struct {
...
+	// In the future more options may be added.
+	AlertmanagerConfigMatcherStrategy *AlertmanagerConfigMatcherStrategy `json:"alertmanagerConfigMatcherStrategy,omitempty"`
```
simonpasquier, line 2027: We can avoid a pointer here.

```suggestion
	AlertmanagerConfigMatcherStrategy AlertmanagerConfigMatcherStrategy `json:"alertmanagerConfigMatcherStrategy,omitempty"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -2043,6 +2046,16 @@ type AlertmanagerSpec struct {
 	AlertmanagerConfiguration *AlertmanagerConfiguration `json:"alertmanagerConfiguration,omitempty"`
 }
 
+// AlertmanagerConfigMatcherStrategy defines the strategy that is used to match AlertmanagerConfigs.
```
simonpasquier, line 2049:
```suggestion
// AlertmanagerConfigMatcherStrategy defines the strategy used by AlertmanagerConfig objects to match alerts.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -2043,6 +2046,16 @@ type AlertmanagerSpec struct {
...
+type AlertmanagerConfigMatcherStrategy struct {
+	// If set to `OnNamespace` namespace matchers will be added to the routes.
```
simonpasquier, line 2051:
```suggestion
	// If set to `OnNamespace`, the operator injects a label matcher matching the namespace of the AlertmanagerConfig object for all its routes and inhibition rules.
```

pkg/alertmanager/amcfg.go
```diff
@@ -99,24 +99,50 @@ func (c alertmanagerConfig) String() string {
 	return string(b)
 }
 
+type enforcer interface {
+	processRoute(*route) *route
+	processInhibitRule(*inhibitRule) *inhibitRule
+	updateNamespace(namespace string)
+}
```
simonpasquier, line 106: I'm not sure that we need updateNamespace() here. To keep it generic maybe the 2 other functions need to have an additional parameter holding name+namespace of the referenced AlertmanagerConfig object?


```suggestion
type enforcer interface {
	processRoute(types.NamespacedName, *route) *route
	processInhibitRule(types.NamespacedName, *inhibitRule) *inhibitRule
}
```

pkg/alertmanager/amcfg.go
```diff
@@ -260,26 +287,30 @@ func (cb *configBuilder) addAlertmanagerConfigs(ctx context.Context, amConfigs m
 	return nil
 }
 
-// enforceNamespaceForInhibitRule modifies the inhibition rule to match alerts
-// originating only from the given namespace.
-func (cb *configBuilder) enforceNamespaceForInhibitRule(ir *inhibitRule, namespace string) *inhibitRule {
-	matchersV2Allowed := cb.amVersion.GTE(semver.MustParse("0.22.0"))
+func (ne *noopEnforcer) updateNamespace(namespace string) {}
```
simonpasquier, line 290: (nit) I'd move the methods next to the definition of the noopEnforcer struct.

pkg/alertmanager/amcfg.go
```diff
@@ -260,26 +287,30 @@ func (cb *configBuilder) addAlertmanagerConfigs(ctx context.Context, amConfigs m
...
 
+func (ne *namespaceEnforcer) updateNamespace(namespace string) { ne.namespace = namespace }
```
simonpasquier, line 292: same here but for namespaceEnforcer

simonpasquier (commented):
I'm quite pleased to see that the change is rather small :) I've got only a few nits and suggestions but this looks good to me in general.

pkg/apis/monitoring/v1/types.go
simonpasquier: My point was to say that if we use a value rather than a pointer, we don't need to check for nil pointers (as [here](https://github.com/prometheus-operator/prometheus-operator/blob/fa2ca180fe4a7fac2e9204622b6b9de98beb0914/pkg/alertmanager/amcfg.go#L199)).

Fr3574: Ah I see, sure my bad.

Fr3574:
@simonpasquier I've added your suggestions, could you give a final review?

simonpasquier (approved):

simonpasquier:
Thanks!

Fr3574:
Great! @simonpasquier Thank you for your guidance.

PR #5089, opened 2022-10-12
pkg/operator: stop reconciliation for objects being deleted

simonpasquier:
## Description

Whenever the operator receives an update for a monitoring CR or a statefulset that is marked for deletion, it is safer to wait for the object being completely deleted before trying to reconcile.

Without this change, the operator may try to recreate a dependent resource such as a configmap or secret owned by the being-deleted resource which would create a race with the resource garbage collector.

## Type of change

## Changelog entry

```release-note
Avoided unnecessary reconciliations when objects are being deleted.
```


pkg/operator/resource_reconciler.go
```diff
@@ -133,6 +133,17 @@ func NewResourceReconciler(
 	}
 }
 
+// deletionInProgress returns true if the object deletion has been requested.
+func (rr *ResourceReconciler) deletionInProgress(o metav1.Object) bool {
+	if o.GetDeletionTimestamp() != nil {
+		level.Debug(rr.logger).Log(
+			"msg", "object deletion in progress",
+			"object", fmt.Sprintf("%s/%s", o.GetNamespace(), o.GetName()),
```
amuraru, line 141: Is this missing "return true" ?

simonpasquier: :facepalm: thanks for spotting it!

PhilipGough (approved):
lgtm

Issue #5050, opened 2022-09-26
Ability to add labels and annotations to prometheus rules from PrometheusRule CR


gberche-orange:
**What is missing?**

Ability to add labels and annotations to prometheus rules from PrometheusRule CR

**Why do we need it?**

adding onto an existing off-the-shelf prometheus alert such as [host-and-hardware/node-exporter.yml](https://github.com/samber/awesome-prometheus-alerts/blob/master/dist/rules/host-and-hardware/node-exporter.yml#L214) 

1) a set of [prometheus alert rule label](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/)  such as the followings:
* `source: awesome-prometheus-alerts`: the source authoring for this alert
* `local-operators-adoption: alpha`: how much confidence/experience the current operation teams has on this alert

This is then used in the prometheus alert manager to group/filter out alerts from a new source, or of yet low confidence (despites their contributed `severity` level)

2) a set of prometheus alert rules annotations such as the followings:
* `source_url: https://github.com/samber/awesome-prometheus-alerts/blob/master/dist/rules/host-and-hardware/node-exporter.yml`: used by operators to access the git history of an alert definition, create issues, or PR for refinements

**Environment**

* Prometheus Operator version:

https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.59.2

**Anything else we need to know?**:

See more details in the conversation at https://github.com/prometheus-operator/prometheus-operator/issues/4495#issuecomment-1251026056

PR #5100, opened 2022-10-14
Adds labels and annotations to PrometheusRule CR

JoaoBraveCoding:
Signed-off-by: Joao Marcal <jmarcal@redhat.com>

## Description

Issue: https://github.com/prometheus-operator/prometheus-operator/issues/5050

Problem: described in the issue. Furthermore, it also fixes the function mergeKubectlAnnotations where previously it was working only because the to map was being modified inside the function mergeMapsByPrefix.

Solution: Additions of labels and annotations to PrometheusRule CR that are then merged into each rule. If a rule already defines a key it will not be overwritten. Regarding mergeKubectlAnnotations function now correctly sets the result of mergingMaps to "to" instead of "from". MergeMapsByPrefix now creates a new map so the original remain intact.

## Type of change

## Changelog entry

```release-note
Adds global labels and annotations to PrometheusRule CR that will be added to all rules in groups
```


simonpasquier:
I think that your change isn't exactly what was requested. The ask is to define labels/annotations at the PrometheusRule level that are injected in every alerting/recording rule of every rule group.

Given

```yaml
kind: PrometheusRule
metadata:
  name: prometheus-example-rules
spec:
  labels:
    label1: label_val1
  annotations:
    annotation1: annotation_val1
  groups:
  - name: example.rules
    rules:
    - alert: ExampleAlert
      expr: vector(1)
      labels:
        label1: overriden
    - record: zero
      expr: vector(0)
```

The resulting Prometheus config file would look like:

```yaml
groups:
- name: example.rules
  rules:
  - alert: ExampleAlert
    expr: vector(1)
    labels:
      label1: overriden # existing label value is retained.
    annotations:
      annotation1: annotation_val1
  - record: zero
      expr: vector(0)
    labels:
      label1: label_val1
    # no annotation because not supported for recording rules.
```

ping @gberche-orange to keep me honest.

gberche-orange:
thanks @JoaoBraveCoding and @simonpasquier. The wording by both of you seem correct to me, and I have a hard time reading the tests at https://github.com/prometheus-operator/prometheus-operator/pull/5100/files#diff-dee0fd2fb1a5d5093af4ed100731310b27b020ffda23cef4e1c97440fed04e23 to find out how tests differs from the above wording.

simonpasquier:
We clarified off-line with @JoaoBraveCoding: I was confused in my initial review and the PR is doing the right thing.
Quick feedback for now:
* Annotations shouldn't be copied to recording rules.
* The fields descriptions should be updated to make it clear that it applies to rules, not Kubernetes objects.

JoaoBraveCoding:
agreed 👍

paulfantom (commented):
Wouldn't https://github.com/prometheus/prometheus/pull/11474 essentially solve the root issue (ability to add metadata to all alerts in a group/CR)? If that PR gets implemented I don't see why we would need to add a 3rd level of defining labels/annotations for alerts.

And if we still need this, I would propose to do a cascading propagation through the levels. Basically if someone sets `spec.labels`, operator would only set `spec.groups[].labels` instead of going through all alerts.

simonpasquier:
@paulfantom agreed that we need to see where https://github.com/prometheus/prometheus/pull/11474 is going.

PR #5122, opened 2022-10-26
Modifies ThanosRuler arguments to be read from secrets instead of using ENV vars

JoaoBraveCoding:
## Description
Issue: https://issues.redhat.com/browse/OCPBUGS-2778

Problem: Some arguments are currently being passed to ThanosRuler using ENV vars. Those arguments are QueryConfig, AlertManagerConfig, ObjectStorageConfig and TracingConfig. These arguments should instead be read directly from the secrets instead of using ENV vars.

Solution: Modify the code to mount the secrets in the SecretKeySelector instead of assigning them to ENV variables

## Type of change

## Changelog entry

```release-note
-  ThanosRuler arguments (QueryConfig, AlertManagerConfig, ObjectStorageConfig and TracingConfig) are now directly read from the secrets instead of using ENV vars
```

### DIFF
--- a/pkg/thanos/statefulset.go
+++ b/pkg/thanos/statefulset.go
@@ -33,6 +33,7 @@ import (
 
 const (
 	rulesDir                  = "/etc/thanos/rules"
+	configDir                 = "/etc/thanos/config"
 	storageDir                = "/thanos/data"
 	governingServiceName      = "thanos-ruler-operated"
 	defaultPortName           = "web"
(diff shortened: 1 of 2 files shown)


simonpasquier (approved):
lgtm. We can probably improve the way volumes/volume mounts are populated (e.g. have an intermediate object that accumulates volumes & volume mounts during the building statefulset rather than bare slices of `v1.Volume`/`v1.VolumeMount`) but this can be a follow-up PR.

PR #5125, opened 2022-10-28
pkg/prometheus: Ignore PartialResponseStrategy for Prometheus

### LINKED ISSUE #5124: Prometheus log reports field partial_response_strategy not found in type rulefmt.RuleGroup when sets this field in PrometheusRule
**What happened?**
Prometheus log reports field partial_response_strategy not found in type rulefmt.RuleGroup when sets this field in PrometheusRule. This was getting logged till removed this field or set it to `""`

```
ts=2022-10-27T11:43:29.048Z caller=manager.go:980 level=error component="rule manager" msg="loading groups failed" err="/etc/prometheus/rules/prometheus-user-workload-rulefiles-0/ns1-example-alert-2f98a05b-dd27-4d1a-87bd-5ef74352cc3a.yaml: yaml: unmarshal errors:\n  line 3: field partial_response_strategy not found in type rulefmt.RuleGroup"
ts=2022-10-27T11:43:29.048Z caller=main.go:1227 level=error msg="Failed to apply configuration" err="error loading rules, previous rule set restored"
ts=2022-10-27T11:43:29.048Z caller=main.go:945 level=error msg="Error reloading config" err="one or more errors occurred while applying the new configuration (--config.file=\"/etc/prometheus/config_out/prometheus.env.yaml\")"
```

Since this field is only applicable for Thanos. Operator should reject this field for Prometheus and allow it only for Thanos

Create a PrometheusRule object with `partial_response_strategy` set to a valid accepted value

```yaml
kind: PrometheusRule
metadata:
  name: example-alert
  namespace: ns1
spec:
  groups:
  - name: example
    partial_response_strategy: "abort"
    rules:
    - alert: VersionAlert
      expr: version{job="prometheus-example-app"} == 0
```

**Environment**

All

    `kubectl version`
    

NA

* Kubernetes cluster kind:

NA

* Manifests:

```
```

NA

* Prometheus Operator Logs:

```
```

NA

As discussed in https://github.com/prometheus-operator/prometheus-operator/pull/4999#issuecomment-1280927196

slashpai:
partial_response_strategy field is only applicable for Thanos so if a user sets this by mistake in PrometheusRule object prometheus errors this field is not found in rule format.

This commit is to fix that and allow this field only for Thanos

Fixes #5124

## Description

## Type of change

## Changelog entry

```release-note
Ignore PartialResponseStrategy for Prometheus
```

### DIFF
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -1,0 +1,71 @@
+// Copyright 2022 The prometheus-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package operator
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/model/rulefmt"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	thanostypes "github.com/thanos-io/thanos/pkg/store/storepb"
+)
+
+func GenerateRulesConfiguration(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
+	content, err := yaml.Marshal(promRule)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to marshal content")
+	}
+	errs := ValidateRule(promRule)
+	if len(errs) != 0 {
+		const m = "Invalid rule"
+		level.Debug(logger).Log("msg", m, "content", content)
+		for _, err := range errs {
+			level.Info(logger).Log("msg", m, "err", err)
+		}
+		return "", errors.New(m)
+	}
+	return string(content), nil
+}
+
+// ValidateRule takes PrometheusRuleSpec and validates it using the upstream prometheus rule validator.
+func ValidateRule(promRule monitoringv1.PrometheusRuleSpec) []error {
+	for i, group := range promRule.Groups {
+		if group.PartialResponseStrategy == "" {
+			continue
+		}
+		// TODO(slashpai): Remove this validation after v0.65 since this is handled at CRD level
+		if _, ok := thanostypes.PartialResponseStrategy_value[strings.ToUpper(group.PartialResponseStrategy)]; !ok {
+			return []error{
+				fmt.Errorf("invalid partial_response_strategy %s value", group.PartialResponseStrategy),
+			}
+		}
+
+		// reset this as the upstream prometheus rule validator
+		// is not aware of the partial_response_strategy field.
+		promRule.Groups[i].PartialResponseStrategy = ""
+	}
+	content, err := yaml.Marshal(promRule)
+	if err != nil {
+		return []error{errors.Wrap(err, "failed to marshal content")}
+	}
+	_, errs := rulefmt.Parse(content)
+	return errs
+}
--- a/pkg/prometheus/rules.go
+++ b/pkg/prometheus/rules.go
@@ -22,19 +22,17 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/ghodss/yaml"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring"
 	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
-	"github.com/prometheus/prometheus/model/rulefmt"
+	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	namespacelabeler "github.com/prometheus-operator/prometheus-operator/pkg/namespace-labeler"
-	thanostypes "github.com/thanos-io/thanos/pkg/store/storepb"
 )
 
 const labelPrometheusName = "prometheus-name"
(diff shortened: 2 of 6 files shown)


pkg/prometheus/rules.go
```diff
@@ -332,20 +332,37 @@ func prometheusRuleConfigMapName(prometheusName string) string {
...
-func GenerateContent(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
+func GenerateContent(objectType string, promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
 	content, err := yaml.Marshal(promRule)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to marshal content")
 	}
+
 	errs := ValidateRule(promRule)
-	if len(errs) != 0 {
+	if errs != nil {
```
simonpasquier, line 342: Better to keep the original case IMHO as an empty slice might not be nil.

```suggestion
	if len(errs) != 0 {
```

pkg/prometheus/rules.go
```diff
@@ -332,20 +332,37 @@ func prometheusRuleConfigMapName(prometheusName string) string {
 }
 
 // GenerateContent takes a PrometheusRuleSpec and generates the rule content
-func GenerateContent(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
+func GenerateContent(objectType string, promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
```
simonpasquier, line 335: 2 functions would be more explicit IMHO. I'd also recommend moving the code to the pkg/operator package since it's shared between prometheus and thanos.

```suggestion
func GenerateRulesConfigurationForPrometheus(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
    // Unset partialResponseStrategy field.
    for _, for i := range promRule.Groups {
        ...
    }
    return generateRulesConfiguration(promRule, logger)
}

func GenerateRulesConfigurationForThanos(objectType string, promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
    return generateRulesConfiguration(promRule, logger)
}

func generateRulesConfiguration(objectType string, promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
    // same as GenerateContent
}
```

slashpai: Thanks for that suggestion :) It looks much better :)

slashpai:
@simonpasquier I have updated the PR with suggestions

pkg/operator/rules.go
```diff
@@ -0,0 +1,84 @@
...
+		level.Debug(logger).Log("msg", m, "content", content)
+		for _, err := range errs {
+			level.Info(logger).Log("msg", m, "err", err)
+		}
+		return "", errors.New(m)
+	}
+	return string(content), nil
+}
+
+// ValidateRule takes PrometheusRuleSpec and validates it using the upstream prometheus rule validator
```
simonpasquier, line 61:
```suggestion
// ValidateRule takes PrometheusRuleSpec and validates it using the upstream prometheus rule validator.
```

pkg/operator/rules.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		// TODO(slashpai): Remove this validation after v0.65 since this is handled at CRD level
+		if _, ok := thanostypes.PartialResponseStrategy_value[strings.ToUpper(group.PartialResponseStrategy)]; !ok {
+			return []error{
+				fmt.Errorf("invalid partial_response_strategy %s value", group.PartialResponseStrategy),
+			}
+		}
+
+		// reset this as the upstream prometheus rule validator
+		// is not aware of the partial_response_strategy field
```
simonpasquier, line 75:
```suggestion
		// is not aware of the partial_response_strategy field.
```

pkg/operator/rules.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/model/rulefmt"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	thanostypes "github.com/thanos-io/thanos/pkg/store/storepb"
+)
+
+func GenerateRulesConfigurationForPrometheus(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
```
simonpasquier, line 31: It's a bit inconsistent that the unit tests live in pkg/prometheus while the impementation is in pkg/operator. I believe that we could move `GenerateRulesConfigurationForPrometheus()` to the pkg/prometheus package (and rename it to `GenerateRulesConfiguration(...)`)

pkg/operator/rules.go
```diff
@@ -0,0 +1,84 @@
...
+	// Unset partialResponseStrategy field.
+	for i := range promRule.Groups {
+		promRule.Groups[i].PartialResponseStrategy = ""
+	}
+
+	return generateRulesConfiguration(promRule, logger)
+}
+
+func GenerateRulesConfigurationForThanos(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
```
simonpasquier, line 40: in the same vein, `GenerateRulesConfigurationForThanos` could go away and pkg/thanos can call `operator.GenerateRulesConfiguration(...)` directly.

pkg/operator/rules.go
```diff
@@ -0,0 +1,84 @@
...
+	return generateRulesConfiguration(promRule, logger)
+}
+
+func generateRulesConfiguration(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
```
simonpasquier, line 44:
```suggestion
func GenerateRulesConfiguration(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
```

pkg/thanos/rules_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
+)
+
+func TestMakeRulesConfigMapsThanos(t *testing.T) {
+	t.Run("shouldAcceptRuleWithValidPartialResponseStrategyValue", shouldAcceptRuleWithValidPartialResponseStrategyValue)
+	t.Run("shouldRejectRuleWithInvalidPartialResponseStrategyValue", shouldRejectRuleWithInvalidPartialResponseStrategyValue)
+	t.Run("shouldAcceptRuleWithPartialResponseStrategySet", shouldAcceptRuleWithPartialResponseStrategySet)
```
simonpasquier, line 32: Quite similar to `shouldAcceptRuleWithValidPartialResponseStrategyValue` so I suggest to remove this test.

pkg/operator/rules.go
slashpai: Sure can do that :). I created separated ones based on previous discussion https://github.com/prometheus-operator/prometheus-operator/pull/5125#discussion_r1008045422 since both Prometheus and Thanos  calls the method

slashpai: I think we would need `GenerateRulesConfigurationForPrometheus()` and `GenerateRulesConfigurationForThanos()` to differentiate who is calling `GenerateRulesConfiguration` that way we can say if `partialResponseStrategy` field is valid or not

slashpai:
@simonpasquier I updated the PR with changes. Moved tests to respective package as well where the code exists

simonpasquier (approved):
thanks!

PR #5126, opened 2022-10-28
Remove "replace" directive of github.com/prometheus/prometheus in go.mod

raptorsun:
## Description

In the `go.mod` file it uses replace directive for github.com/prometheus/prometheus to replace the placeholder version of 1.99 with the latest version 0.38.1.

This PR will remove the prometheus package out of the replace directive, so that we can use directly the version 0.x.y which follows the real version 2.x.y. Next time when we update go packages using `go get -u github.com/prometheus/prometheus`, it will pick the latest version without modifying the replace directive.

## Type of change

## Changelog entry

```release-note
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,8 +25,7 @@ require (
 	github.com/prometheus/alertmanager v0.24.0
 	github.com/prometheus/client_golang v1.13.0
 	github.com/prometheus/common v0.37.0
-	// This version is replaced using replace directive below
-	github.com/prometheus/prometheus v1.99.0
+	github.com/prometheus/prometheus v0.39.1
 	github.com/stretchr/testify v1.8.1
 	github.com/thanos-io/thanos v0.28.1
 	golang.org/x/sync v0.0.0-20220929204114-8fcdb60fdcc0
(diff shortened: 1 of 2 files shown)


slashpai:
Can you rebase the branch? :)

raptorsun:
> Can you rebase the branch? :)

Thanks, it is corrected now :)

go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/prometheus/client_golang v1.13.0
 	github.com/prometheus/common v0.37.0
 	// This version is replaced using replace directive below
```
PhilipGough, line 28: This comment should be removed if we intend to remove the replace directive

go.mod
```diff
@@ -124,7 +124,5 @@ require (
 replace (
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring => ./pkg/apis/monitoring
 	github.com/prometheus-operator/prometheus-operator/pkg/client => ./pkg/client
-	// A replace directive is needed for github.com/prometheus/prometheus to ensure running against the latest version of prometheus.
-	github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.38.1-0.20221005044927-6d7f26c46ff7
```
PhilipGough, line 128: I am not sure about this change to be honest. The comment prior to this provided more detail

```md
// A replace directive is needed for github.com/prometheus/prometheus
// because Thanos hasn't switched to the new v0.x tag and Go modules
// would select an "older" version.
```

cc @paulfantom

raptorsun: The 0.x.y version of Prometheus is now pointing to the 2.x.y version. When we update the package in go.mod, it will not upgraded to v1.99.0 automatically, should be safe to avoid older versions.

simonpasquier: This Go module depends on github.com/thanos-io/thanos@v0.28.1 which depends itself on github.com/prometheus/prometheus@v0.38.0 so it is safe now to drop the `replace` directive.

https://github.com/thanos-io/thanos/blob/41a8ce61c90734b81986b273e154e96358333f92/go.mod#L69-L70

PhilipGough (approved):

PR #5134, opened 2022-11-04
Prometheus hostnetwork: specify listen address on pod ip

lwangrabbit:
## Description

when prometheus.spec.hostNetwork=true, it will listen on 0.0.0.0:9090 of the node by default, which is not security.

To reduce the problem, just specify it only listen on the pod ip.

## Type of change

## Changelog entry

```
prometheus hostnetwork: specify listen address on pod ip
```

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -460,6 +460,9 @@ func makeStatefulSetSpec(
 	if p.Spec.ListenLocal {
 		promArgs = append(promArgs, monitoringv1.Argument{Name: "web.listen-address", Value: "127.0.0.1:9090"})
 	} else {
+		if p.Spec.HostNetwork {
+			promArgs = append(promArgs, monitoringv1.Argument{Name: "web.listen-address", Value: "[$(POD_IP)]:9090"})
+		}
 		ports = []v1.ContainerPort{
 			{
 				Name:          p.Spec.PortName,
@@ -956,6 +959,16 @@ func makeStatefulSetSpec(
 				AllowPrivilegeEscalation: &boolFalse,
 				Capabilities: &v1.Capabilities{
 					Drop: []v1.Capability{"ALL"},
+				},
+			},
+			Env: []v1.EnvVar{
+				{
+					Name: "POD_IP",
+					ValueFrom: &v1.EnvVarSource{
+						FieldRef: &v1.ObjectFieldSelector{
+							FieldPath: "status.podIP",
+						},
+					},
 				},
 			},
 		},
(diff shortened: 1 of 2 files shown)


simonpasquier:
lgtm. I'd like to get additional input from @Rajpratik71 and @adamsb6 as I don't have concrete experience using host network.

Rajpratik71:
HI @lwangrabbit  @simonpasquier ,

for security reasons advertising Prometheus on Pod IP is a good scenario but it will break and cause issues in other scenarios like:

-  In the cloud where generally we have private IP on the interface while a floating Public IP

- In environments with multiple networks and interfaces

there can be more.

By default enabling this will cause the issue, this can be done optionally through parameters.

lwangrabbit:
In our scenario, each node has internal ip and external ip，hostNetwork default make pod listen [::]:9090, both internal ip and external ip, but we only want it listen on internal ip. 

May be we can enable it through parameters.

Rajpratik71:
> In our scenario, each node has internal ip and external ip，hostNetwork default make pod listen [::]:9090, both internal ip and external ip, but we only want it listen on internal ip.
> 
> May be we can enable it through parameters.

I.e. why same I mentioned, for your use case it is fine to block access but for others, it will create an issue

adamsb6:
For our use case this is a non-issue, we don't attach a public IP to the instances we run Prometheus on.

However, I think this code will break us. We need to serve on multiple private IPs.

Additionally, I think this code may be grabbing the desired IP on accident. `podIP` is set in [pkg/kubelet/kuberuntime/kuberuntime_manager.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L822), it's just the first item in the `podIPs` array.

This maps to the the `Network.Ip`attribute yielded by the CRI's `PodSandBoxStatus()`function. As far as I can tell there's no rule for container runtimes on which IP gets the privilege of being *the* IP, so it's up to containerd/Docker/etc. and the behavior could differ. I haven't looked at the runtime code for containerd, but it's possible this could just be the first element of a sorted list, with link-locals filtered out.

To achieve your goal I think you'll need some way to indicate exactly which IP you prefer, and a flag so that we don't alter the current default behavior.

adamsb6 (changes requested):
I think this would break my use case, I need to serve on all available IPs. Can you make this behavior optional?

See also my comments on the origin of `POD_IP`.

lwangrabbit:
> I think this would break my use case, I need to serve on all available IPs. Can you make this behavior optional?
> 
> See also my comments on the origin of `POD_IP`.


Thanks！The POD_IP is import which I haven't noticed, which should be fixed on the network you wanted.

simonpasquier:
Thanks for chiming in! I agree that it needs to be configurable and we want to keep the current behavior (e.g. listen on all interfaces) as the default.

PR #5135, opened 2022-11-06
Support time_intervals and active_time_intervals

### LINKED ISSUE #4656: Add support for active time intervals in AlertmanagerConfig CRD
**What is missing?**

Alertmanager v0.24 (to be released soon) supports active time intervals (see https://github.com/prometheus/alertmanager/pull/2779). The functionality would need to be added to the AlertmanagerConfig CRD too.

Similar to what happens in Alertmanager, the existing `spec.muteTimeIntervals` field needs to be deprecated in favor of `spec.timeIntervals`. The operator could seamlessly merge both fields in the generated configuration:
* using `time_intervals` if the Alertmanager's `spec.version` is >= v0.24.
* `mute_time_intervals` otherwise.

**Why do we need it?**

Feature parity with upstream Alertmanager.

**Environment**

chtcvl:
Fixes #4656 

## Description

Add support for using `time_intervals` in Alertmanager global configuration and and `active_time_intervals` in Route configuration.

## Type of change

## Changelog entry

Support `time_intervals` and `active_time_intervals`

```release-note
Add support for using `time_intervals` in Alertmanager global configuration and and `active_time_intervals` in Route configuration.
```


pkg/alertmanager/amcfg.go
```diff
@@ -153,11 +153,19 @@ func (cb *configBuilder) initializeFromAlertmanagerConfig(ctx context.Context, g
 	}
 
 	for _, muteTimeInterval := range amConfig.Spec.MuteTimeIntervals {
-		mti, err := convertMuteTimeInterval(&muteTimeInterval, crKey)
+		ti, err := convertTimeInterval(&muteTimeInterval, crKey)
 		if err != nil {
 			return err
 		}
-		globalAlertmanagerConfig.MuteTimeIntervals = append(globalAlertmanagerConfig.MuteTimeIntervals, mti)
+		globalAlertmanagerConfig.TimeIntervals = append(globalAlertmanagerConfig.TimeIntervals, ti)
```
JoaoBraveCoding, line 160: The code is simply moving away from MuteTimeIntervals to TimeInterval but shouldn't MuteTimeInterval be the default and just use TimeInterval when the AM version is >= v0.24? Or am I missing something?

pkg/alertmanager/amcfg_test.go
```diff
@@ -1407,7 +1407,7 @@ templates: []
 							Receiver:          "test",
 							MuteTimeIntervals: []string{"test"},
 						},
-						MuteTimeIntervals: []monitoringv1alpha1.MuteTimeInterval{
```
JoaoBraveCoding, line 1410: It would be nice to extend the tests to test the new options

JoaoBraveCoding (commented):

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -131,6 +134,9 @@ type Route struct {
 	// MuteTimeIntervals is a list of MuteTimeInterval names that will mute this route when matched,
 	// +optional
```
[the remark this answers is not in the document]
chtcvl: This one is not deprecated, since both `mute_time_intervals` and `active_time_intervals` are supported in the route. Added at https://github.com/prometheus-operator/prometheus-operator/pull/5135/files#diff-ca1bb1a59e5c40fa6d7fe5a21bba214234558c9ab699f3de0fdc95d7b302dc35R81 to mark `mute_time_intervals` in global config deprecated

pkg/alertmanager/amcfg.go
chtcvl: Sorry, indeed it was a replacement. Fixed to support both syntax

pkg/alertmanager/amcfg_test.go
chtcvl: Done

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -78,8 +78,12 @@ type AlertmanagerConfigSpec struct {
...
 	InhibitRules []InhibitRule `json:"inhibitRules,omitempty"`
 	// List of MuteTimeInterval specifying when the routes should be muted.
+	// +kubebuilder:deprecatedversion
```
simonpasquier, line 81: I believe that this marker only applies to CRDs, not individual fields/types.

simonpasquier (commented):
Sorry about this but I realize now that #4656 doesn't have all the details and contained outdated information... I think we can have 2 PRs to simplify the work and the review. 

The first one will add support for `time_intervals` and `active_time_intervals` to the internal alertmanagerConfig struct:
https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/alertmanager/types.go#L30-L37
The goal is that a user can provide a raw Alertmanager configuration (via a secret) using the new `time_intervals` syntax and that the config generator checks/sanitizes the generated configuration depending on the Alertmanager version.

The second PR will add support for active time intervals to the AlertmanagerConfig CRD. Here we don't need to add a new `spec.timeIntervals`: it already exists in the [v1beta1](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1beta1/alertmanager_config_types.go#L78-L80) version and we can keep using  `spec.muteTimeIntervals` in v1alpha1 (the v1alpha1 <-> v1beta1 [conversion logic](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1beta1/conversion_from.go#L528-L536) already handles the translation from one field to the other).

What we need though is adding the `activeTimeIntervals` field to the Route struct for both [v1alpha1](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L86-L134) and [v1beta1](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1beta1/alertmanager_config_types.go#L84-L132).

chtcvl:
> Sorry about this but I realize now that #4656 doesn't have all the details and contained outdated information... I think we can have 2 PRs to simplify the work and the review.
> 
> The first one will add support for `time_intervals` and `active_time_intervals` to the internal alertmanagerConfig struct:
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/alertmanager/types.go#L30-L37
> 
> 
> The goal is that a user can provide a raw Alertmanager configuration (via a secret) using the new `time_intervals` syntax and that the config generator checks/sanitizes the generated configuration depending on the Alertmanager version.
> The second PR will add support for active time intervals to the AlertmanagerConfig CRD. Here we don't need to add a new `spec.timeIntervals`: it already exists in the [v1beta1](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1beta1/alertmanager_config_types.go#L78-L80) version and we can keep using `spec.muteTimeIntervals` in v1alpha1 (the v1alpha1 <-> v1beta1 [conversion logic](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1beta1/conversion_from.go#L528-L536) already handles the translation from one field to the other).
> 
> What we need though is adding the `activeTimeIntervals` field to the Route struct for both [v1alpha1](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L86-L134) and [v1beta1](https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/apis/monitoring/v1beta1/alertmanager_config_types.go#L84-L132).

Pushed change for this PR to add support for `time_intervals` and `active_time_intervals` to the internal alertmanagerConfig struct only, will create a new one to address remaining part.

JoaoBraveCoding (commented):
I might be mistaken but shouldn't this PR also include the part about

> the config generator checks/sanitizes the generated configuration depending on the Alertmanager version

Also, @simonpasquier shouldn't the CI have failed this commit? We seem to be missing the files from `make generate` but the CI passed.

simonpasquier:
@JoaoBraveCoding there's no change to the CRDs so no change to the generated code :)

simonpasquier:
We'd need to check in `*alertmanagerConfig) sanitize(`) that the Alertmanager's version supports the new time intervals (like we do for mute time intervals).

https://github.com/prometheus-operator/prometheus-operator/blob/ba7ec03a08a81a9a40b128524d8104604a47baf9/pkg/alertmanager/amcfg.go#L1379-L1383

The same in `*route.sanitize()`

https://github.com/prometheus-operator/prometheus-operator/blob/ba7ec03a08a81a9a40b128524d8104604a47baf9/pkg/alertmanager/amcfg.go#L1688-L1692

simonpasquier:
Additional tests in pkg/alertmanager/amcfg_test.go would be neat :)

chtcvl:
Added sanitizing code to drop `time_intervals` and `active_time_intervals` config on version < 0.24.0 and added tests.

simonpasquier (approved):
thanks!

PR #5142, opened 2022-11-08
Add filter_external_labels remote_read flag

dobrerazvan:
## Description

This adds the filter_external_labels remote_read flag to prom operator. In case remote metrics use the same external labels as you current prom deployment, then the remote metrics won't be read: issue described here https://github.com/prometheus/prometheus/issues/7192.

## Type of change

## Changelog entry

```release-note
Adding the filter_external_labels remote_read flag to prom operator.

```


pkg/prometheus/promcfg.go
```diff
@@ -1691,6 +1691,8 @@ func (cg *ConfigGenerator) generateRemoteReadConfig(
 			cfg = append(cfg, yaml.MapItem{Key: "proxy_url", Value: spec.ProxyURL})
 		}
 
+		cfg = append(cfg, yaml.MapItem{Key: "filter_external_labels", Value: spec.FilterExternalLabels})
```
simonpasquier, line 1694: There needs to be a check of the Prometheus version

```suggestion
		cfg = cg.WithMinimumVersion("2.34.0").AppendMapItem(cfg, yaml.MapItem{Key: "filter_external_labels", Value: spec.FilterExternalLabels})
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1122,6 +1122,9 @@ type RemoteReadSpec struct {
 	TLSConfig *TLSConfig `json:"tlsConfig,omitempty"`
 	// Optional ProxyURL.
 	ProxyURL string `json:"proxyUrl,omitempty"`
+	// Optional FilterExternalLabels. This translates in filter_external_labels
+	// in prometheus.yaml configuration.
+	FilterExternalLabels bool `json:"filterExternalLabels,omitempty"`
```
simonpasquier, line 1127: The default Prometheus value is `true` so we need a bool pointer (otherwise the flag would be automatically disabled for all users after upgrade).

```suggestion
	FilterExternalLabels *bool `json:"filterExternalLabels,omitempty"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1122,6 +1122,9 @@ type RemoteReadSpec struct {
 	TLSConfig *TLSConfig `json:"tlsConfig,omitempty"`
 	// Optional ProxyURL.
 	ProxyURL string `json:"proxyUrl,omitempty"`
+	// Optional FilterExternalLabels. This translates in filter_external_labels
+	// in prometheus.yaml configuration.
```
simonpasquier, line 1126:
```suggestion
	// Whether to use the external labels as selectors for the remote read endpoint.
	// Requires Prometheus v2.34.0 and above.
```

pkg/prometheus/promcfg_test.go
```diff
@@ -5474,6 +5475,26 @@ scrape_configs: []
 remote_read:
 - url: http://example.com
   remote_timeout: 30s
+  filter_external_labels: false
+`,
+		},
+		{
```
simonpasquier, line 5481: we would need more tests:
* with version < 2.34.0 and `filterExternalLabels` defined
* with version >= 2.34.0 and `filterExternalLabels` not defined
* with version >= 2.34.0 and `filterExternalLabels` set to `false`
* with version >= 2.34.0 and `filterExternalLabels` set to `true`

dobrerazvan: Thanks @simonpasquier for the review. I will add the suggested tests ;).

dobrerazvan:
@simonpasquier PR is ready for review. Thanks!

simonpasquier (approved):
thanks!

Issue #5130, opened 2022-11-01
Omitting access_key and secret_key in Alertmanager SNS configuration causes a crash


mlaitinen:
**What happened?**

I configured Alertmanager with an SNS receiver and defined a sigv4 configuration without `access_key` and `secret_key`. This causes Prometheus Operator to crash.

I have to define the sigv4 configuration, because otherwise I get this:
```
caller=notify.go:732 level=warn component=dispatcher receiver=prometheus/sns-email/sns-email-mycompany integration=sns[0] msg="Notify attempt failed, will retry later" attempts=1 err="region not configured in sns.sigv4.region or in default credentials chain"
```

**Did you expect to see something different?**

As mentioned in [Prometheus documentation](https://prometheus.io/docs/alerting/latest/configuration/#sigv4_config), `role_arn` can be used as "an alternative to using AWS API keys". I expected Prometheus Operator to ignore `access_key` and `secret_key`, because I had `role_arn` and `region` defined.

**How to reproduce it (as minimally and precisely as possible)**:

Create an `AlertmanagerConfig` manifest, define a `sigv4` config with `region` and `role_arn`, but omit `access_key` and `secret_key`. View Prometheus Operator logs.

**Environment**

* Prometheus Operator version:

    `v0.59.1`
    
    

* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.3", GitCommit:"816c97ab8cff8a1c72eccca1026f7820e93e0d25", GitTreeState:"clean", BuildDate:"2022-01-25T21:25:17Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23+", GitVersion:"v1.23.10-eks-15b7512", GitCommit:"cd6399691d9b1fed9ec20c9c5e82f5993c3f42cb", GitTreeState:"clean", BuildDate:"2022-08-31T19:17:01Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:

    EKS, managed by Terraform

* Manifests:

```
apiVersion: monitoring.coreos.com/v1alpha1
kind: AlertmanagerConfig
metadata:
  labels:
    alertmanagerConfig: sns-email
spec:
  receivers:
    - name: sns-email-mycompany
      snsConfigs:
        - sendResolved: true
          sigv4:
[...]

PR #5143, opened 2022-11-08
Fixes nil pointer when roleARN is used instead of access and secret keys

JoaoBraveCoding:
## Description

Issue: https://github.com/prometheus-operator/prometheus-operator/issues/5130

Problem: Alertmanager supports both roleARN and access and secret keys for Sigv4 authentication, however in the current implementation of prometheus-operator the code does not check if the user has provided access and secret keys and simply calls GetSecretKey, if the user uses roleARN this will generate access to a nil pointer

Solution: Before calling GetSecretKey validate that the user has indeed provided access and a secret key

## Type of change

## Changelog entry

```release-note
Fixes nil pointer when roleARN is used instead of access and secret keys in Alertmanager Config
```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -1049,23 +1049,25 @@ func (cb *configBuilder) convertSnsConfig(ctx context.Context, in monitoringv1al
 	}
 
 	if in.Sigv4 != nil {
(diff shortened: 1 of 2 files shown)


PhilipGough (approved):
lgtm

Issue #5116, opened 2022-10-24
Implement `spec.attachMetadata.node` for ServiceMonitor


simonpasquier:
**What is missing?**

https://github.com/prometheus-operator/prometheus-operator/pull/4792 added the ability to attach node metadata during target discovery for PodMonitor. It wasn't possible to do the same for ServiceMonitor because the feature was only available for the pod role at that time.
Now that https://github.com/prometheus/prometheus/pull/10759 merged in Prometheus v2.37.0, we can close the gap.

**Why do we need it?**

Parity with Prometheus.

**Environment**

* Prometheus Operator version: N/A

**Anything else we need to know?**: No

simonpasquier:
cc @fpetkovski I've just realized that we miss a small improvement for ServiceMonitor :)

PR #5147, opened 2022-11-09
Implements `spec.attachMetadata.node` for ServiceMonitor

JoaoBraveCoding:
## Description

Issue: https://github.com/prometheus-operator/prometheus-operator/issues/5116

Problem: `spec.attachMetadata.node` had been previously implemented for PodMonitor. With PR https://github.com/prometheus/prometheus/pull/10759 being merged it's now possible to also have this field for ServiceMonitor

Solution: add the field to the ServiceMonitor CR and pass it to generateK8SSDConfig instead of nil

## Type of change

## Changelog entry

```release-note
Adds `spec.attachMetadata.node` for ServiceMonitor
```


pkg/prometheus/promcfg.go
```diff
@@ -1120,7 +1120,7 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 		role = kubernetesSDRoleEndpointSlice
 	}
 
-	cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, store, role, nil))
+	cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, store, role, m.Spec.AttachMetadata))
```
PhilipGough, line 1123: I wonder do we need to version check here against `v2.37.0`

https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L1506-L1510 appears to take care of it generically but I am wondering if we need specific checks per CRD

JoaoBraveCoding: Good point! I think it might make sense, better be safe than sorry

JoaoBraveCoding: Should be done 👍

pkg/apis/monitoring/v1/types.go
```diff
@@ -1344,7 +1347,7 @@ type PodMonitorSpec struct {
 	// Per-scrape limit on length of labels value that will be accepted for a sample.
 	// Only valid in Prometheus versions 2.27.0 and newer.
 	LabelValueLengthLimit uint64 `json:"labelValueLengthLimit,omitempty"`
-	// Attaches node metadata to discovered targets. Only valid for role: pod.
+	// Attaches node metadata to discovered targets. Only valid for role: pod and service.
```
simonpasquier, line 1350: I think that the `Only valid for role: pod ...` part isn't really needed here because by definition, a PodMonitor configures the pod's role.

```suggestion
	// Attaches node metadata to discovered targets.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1250,6 +1250,9 @@ type ServiceMonitorSpec struct {
 	// Per-scrape limit on length of labels value that will be accepted for a sample.
 	// Only valid in Prometheus versions 2.27.0 and newer.
 	LabelValueLengthLimit uint64 `json:"labelValueLengthLimit,omitempty"`
+	// Attaches node metadata to discovered targets. Only valid for role: pod and service.
```
simonpasquier, line 1253:
```suggestion
	// Attaches node metadata to discovered targets.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -1250,6 +1250,9 @@ type ServiceMonitorSpec struct {
...
+	// Only valid in Prometheus versions 2.35.0 and newer.
```
simonpasquier, line 1254: I think that v2.37.0 at least is required.
```suggestion
	// Requires Prometheus v2.37.0 and above.
```

pkg/prometheus/promcfg.go
```diff
@@ -1120,7 +1120,11 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 		role = kubernetesSDRoleEndpointSlice
 	}
 
-	cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, store, role, nil))
+	if m.Spec.AttachMetadata != nil {
+		cfg = cg.WithMinimumVersion("2.37.0").AppendMapItem(cfg, "", cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, store, role, m.Spec.AttachMetadata))
```
simonpasquier, line 1124: hmm it would mean that the service monitor would generate no config at all if Prometheus < v2.37.0. We could rather extend `generateK8SSDConfig()` to pass the minimum version supporting node metadata (e.g. 2.35.0 for pod monitor and 2.37.0 for service monitor).

JoaoBraveCoding: Indeed, when I did this I did not realize it would only generate config if Prometheus > v2.37.0. However instead of extending the `generateK8SSDConfig()` I tried encapsulating the details of attachMetadata into a new type that we then pass to `generateK8SSDConfig()`. This to me makes slightly more sense but I'm up to just passing the minimum version if you feel it's better :)

simonpasquier (commented):
the unit tests are failing with a nil pointer panic :)

pkg/prometheus/promcfg.go
PhilipGough: sorry for the lag on re-reviewing this. My suggestion was to actually remove the version check from `generateK8SSDConfig` and have the calling function responsible for validation prior with the applicable version check.

Simons suggestion or your alternative works for me too though.

JoaoBraveCoding: Apologies at the time I miss understood it, I'm fine with implementing either :) tests should be fixed now

simonpasquier (approved):
lgtm

PR #5149, opened 2022-11-10
Adds a filter for non-running pods for ServiceMonitor

JoaoBraveCoding:
## Description

Issue: https://github.com/prometheus-operator/prometheus-operator/issues/4877

Problem: on a previous PR the capability to drop metrics from PodMonitors whose pods were in Failed or Succeded state was introduced. The same can also be done for ServiceMonitors since if a Service is backed by a Pod then all labels of the role: pod discovery are attached as per the documentation on https://prometheus.io/docs/prometheus/latest/configuration/configuration/#endpoints

Solution: Add a step to drop metrics from pods which are in the phase Failed or Succeeded

## Type of change

## Changelog entry

```release-note
- Adds a filter for non-running pods for ServiceMonitor
```


pkg/prometheus/promcfg.go
```diff
@@ -1289,6 +1289,14 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 		},
 	}...)
 
+	if ep.FilterRunning == nil || *ep.FilterRunning {
```
PhilipGough, line 1292: Its not a blocker but perhaps its worth extracting the existing https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L708-L714 into its own function?

PhilipGough (approved):
lgtm

PR #5159, opened 2022-11-14
Improve handling of rule configmaps

### LINKED ISSUE #5085: Improve handling of rule configmaps
**What is missing?**
Improved handling of rule configmaps in a grow-only mode. 

**Why do we need it?**
When reaching the size-limit of the first configmap and then falling below it again, the operator updates the STS and recreates the pods to go back to 1 configmap. This introduces disturbance - especially if the number of rules is around the limit, therefore re-triggering this behaviour multiple times.

I would propose just leaving the new ConfigMap empty in this case and not removing it, therefore reducing the impact on the running Prometheus.

mikechengwei:
## Description

Close #5085 .

## Type of change

## Changelog entry

```release-note
Improve handling of rule configmaps.
```


JoaoBraveCoding (commented):
IIUC the issue, I don't think the goal would be to expose something like the field `PreProvisionRuleConfigMapNumber` but rather hide that detail from users, especially because it's not clear to users how many ConfigMaps they will need. 
The operator should instead provision a set of extra ConfigMaps (2 or 3) that it would mount on the statefulset and use them to spread the configuration if need be. 
Furthermore I feel that the operator should still maintain the ability to provision more ConfigMaps if the configuration grows past the point where the initial ConfigMaps are not enough anymore 
I'll let one of the maintainers correct me if I misunderstood it

simonpasquier:
I confirm what @JoaoBraveCoding said: there's no need to expose the number of pre-provisioned configmaps to the end-user IMHO. As I hinted in the original issue, we might be able to use [optional configmaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#optional-configmaps) to always mount 3 configmaps (3 is an arbitrary number, my hunch is that most deployments won't need more) even though the operator needs only to create one configmap initially.
When more than 3 configmaps are needed, the operator will trigger a statefulset rollout but I think that it's ok. And to avoid subsequent rollout if the number of configmaps shrinks again, I'd record the max number of configmaps in a statefulset's annotation. The operator would use the annotation to know how many volume mounts should be provisioned. Does it make sense?

mikechengwei:
Ok let me think about it.

mikechengwei:
@simonpasquier @JoaoBraveCoding PLTA again.

pkg/prometheus/rules_test.go
```diff
@@ -129,7 +129,7 @@ func shouldReturnAtLeastOneConfigMap(t *testing.T) {
 		t.Fatalf("expected no error but got: %v", err.Error())
 	}
 
-	if len(configMaps) != 1 {
+	if len(configMaps) != 3 {
 		t.Fatalf("expected one ConfigMaps but got %v", len(configMaps))
```
JoaoBraveCoding, line 133: forgot to update the error message
```suggestion
		t.Fatalf("expected three ConfigMaps but got %v", len(configMaps))
```

pkg/prometheus/rules_test.go
```diff
@@ -159,7 +159,7 @@ func shouldSplitUpLargeSmallIntoTwo(t *testing.T) {
 		t.Fatalf("expected no error but got: %v", err)
 	}
 
-	if len(configMaps) != 2 {
+	if len(configMaps) != 3 {
 		t.Fatalf("expected rule files to be split up into two ConfigMaps, but got '%v' instead", len(configMaps))
```
JoaoBraveCoding, line 163: Forgot to update the error message. But I fee like this test should be updated to continue testing the code path where configuration exceeds the 3 ConfigMaps

JoaoBraveCoding (commented):
Good work! But I feel like it's still not there, please take a look at the comment [1] where I provide a bit more detail. 

Disclaimer: I'm not a maintainer so if you have a different opinion feel free to wait for the review of one :) 

[1] https://github.com/prometheus-operator/prometheus-operator/pull/5159#discussion_r1029204404

mikechengwei:
@JoaoBraveCoding plz help me review it again.

pkg/k8sutil/k8sutil.go
```diff
@@ -383,3 +385,45 @@ func mergeMapsByPrefix(from map[string]string, to map[string]string, prefix stri
 
 	return to
 }
+
+// DiffRulerConfigMap preProvisionRuleConfigMapNumber can't less than newConfigMaps number.
```
JoaoBraveCoding, line 389: Comment seems to be outdated
```suggestion
// DiffRulerConfigMap preProvisionRuleConfigMapNumber can't less than newConfigMaps number.
```

pkg/k8sutil/k8sutil.go
```diff
@@ -383,3 +385,45 @@ func mergeMapsByPrefix(from map[string]string, to map[string]string, prefix stri
...
+
+	if newConfigMapsNum < currentConfigMapsNum {
+		deleteConfigMaps = currentConfigMaps[newConfigMapsNum:currentConfigMapsNum]
+		createOrUpdateConfigMaps = newConfigMaps
+	}
+
+	return createOrUpdateConfigMaps, deleteConfigMaps
+}
+
+func CreateOrUpdateConfigMap(ctx context.Context, client clientv1.ConfigMapInterface, configMap *v1.ConfigMap) error {
```
JoaoBraveCoding, line 412: I'm not sure about this, but here couldn't we use [1] instead of maintaining this code. @simonpasquier might know what's best to do in this specific case.

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate

pkg/prometheus/rules.go
```diff
@@ -108,9 +110,9 @@ func (c *Operator) createOrUpdateRuleConfigMaps(ctx context.Context, p *monitori
 		return newConfigMapNames, nil
 	}
 
-	// Simply deleting old ConfigMaps and creating new ones for now. Could be
-	// replaced by logic that only deletes obsolete ConfigMaps in the future.
-	for _, cm := range currentConfigMaps {
+	createOrUpdateConfigMaps, deleteConfigMaps := k8sutil.DiffRulerConfigMap(currentConfigMaps, newConfigMaps)
+	// replaced by logic that only deletes obsolete ConfigMaps.
```
JoaoBraveCoding, line 114: Don't think this comment is necessary

pkg/prometheus/rules.go
```diff
@@ -122,13 +124,21 @@ func (c *Operator) createOrUpdateRuleConfigMaps(ctx context.Context, p *monitori
...
-		_, err = cClient.Create(ctx, &cm, metav1.CreateOptions{})
+	for _, cm := range createOrUpdateConfigMaps {
+		err = k8sutil.CreateOrUpdateConfigMap(ctx, cClient, &cm)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to create new ConfigMap '%v'", cm.Name)
+			return nil, errors.Wrapf(err, "failed to createOrUpdate ConfigMap '%v'", cm.Name)
 		}
 	}
 
+	// supple configmap number
```
JoaoBraveCoding, line 134: This comment has to be updated, it's important because it's the part of the code where we will return the extra empty configMaps that will be mounted optionally. This part of the code will not be obvious if the person reading the code does not know about #5085

pkg/prometheus/rules_test.go
```diff
@@ -130,7 +131,7 @@ func shouldReturnAtLeastOneConfigMap(t *testing.T) {
 	}
 
 	if len(configMaps) != 1 {
-		t.Fatalf("expected one ConfigMaps but got %v", len(configMaps))
+		t.Fatalf("expected three ConfigMaps but got %v", len(configMaps))
```
JoaoBraveCoding, line 134: Sorry, but since we reverted our change to `makeRulesConfigMaps` this should be reverted as well
```suggestion
		t.Fatalf("expected one ConfigMaps but got %v", len(configMaps))
```

pkg/prometheus/rules_test.go
```diff
@@ -160,10 +161,33 @@ func shouldSplitUpLargeSmallIntoTwo(t *testing.T) {
 	}
 
 	if len(configMaps) != 2 {
-		t.Fatalf("expected rule files to be split up into two ConfigMaps, but got '%v' instead", len(configMaps))
+		t.Fatalf("expected rule files to be split up into three ConfigMaps, but got '%v' instead", len(configMaps))
```
JoaoBraveCoding, line 164: Same as previous comment
```suggestion
		t.Fatalf("expected rule files to be split up into two ConfigMaps, but got '%v' instead", len(configMaps))
```

pkg/prometheus/rules_test.go
```diff
@@ -160,10 +161,33 @@ func shouldSplitUpLargeSmallIntoTwo(t *testing.T) {
...
+func shouldSplitUpLargeSmallIntoFour(t *testing.T) {
+	p := &monitoringv1.Prometheus{}
+	ruleFiles := map[string]string{}
+
+	ruleFiles["first"] = strings.Repeat("a", maxConfigMapDataSize)
+	ruleFiles["second"] = strings.Repeat("a", maxConfigMapDataSize)
+	ruleFiles["three"] = strings.Repeat("a", maxConfigMapDataSize)
+	ruleFiles["four"] = "a"
+
+	configMaps, err := makeRulesConfigMaps(p, ruleFiles)
```
JoaoBraveCoding, line 181: Since here we are testing only the function `makeRulesConfigMaps` I don't think adding this test makes sense as it's testing the same behaviour as `shouldSplitUpLargeSmallIntoTwo`. Instead, it might be better to test `createOrUpdateRuleConfigMaps` since that's the function that we actually changed the behaviour

pkg/prometheus/statefulset.go
```diff
@@ -508,14 +509,21 @@ func makeStatefulSetSpec(
...
 	}
 
-	for _, name := range ruleConfigMapNames {
+	for index, name := range ruleConfigMapNames {
+		var optional *bool
+		if index == 0 {
+			optional = pointer.Bool(false)
+		} else {
+			optional = pointer.Bool(true)
+		}
```
JoaoBraveCoding, line 518:
```suggestion
		optional := pointer.Bool(true)
		if index == 0 {
			optional = pointer.Bool(false)
		} 
```

JoaoBraveCoding (commented):
I think it's much closer to the proposition @simonpasquier left! Just a few more comments but I'm pretty happy with the direction! Great work and thank you for your work! 💪 

PS: the comments I left to `prometheus/(rules.go/statefulset.go)` should also apply to `thanos/(rules.go/statefulset.go)`

pkg/k8sutil/k8sutil.go
mikechengwei: I don't think it's good to use this method? If yes, why not all create resources using this package.

mikechengwei:
@simonpasquier  can u help me review it again?

pkg/prometheus/rules.go
```diff
@@ -122,13 +123,21 @@ func (c *Operator) createOrUpdateRuleConfigMaps(ctx context.Context, p *monitori
...
-		_, err = cClient.Create(ctx, &cm, metav1.CreateOptions{})
+	for _, cm := range createOrUpdateConfigMaps {
+		err = k8sutil.CreateOrUpdateConfigMap(ctx, cClient, &cm)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to create new ConfigMap '%v'", cm.Name)
+			return nil, errors.Wrapf(err, "failed to createOrUpdate ConfigMap '%v'", cm.Name)
 		}
 	}
 
+	// supple configmap volumeMount number, at least `default` number of rule configmaps mount.
```
JoaoBraveCoding, line 133:
```suggestion
	// supple configmap volumeMount number, at least `default` number of rule configmaps mount.
	// this is to prevent the StatefulSet Pods from being frequently redeployed when the size of the
	// configuration can easily fluctuate between the maximum size of a single ConfigMaps or not.
```

mikechengwei:
> Just a minor comment but apart from that it looks great! Great job, and thanks for addressing my comments even though I'm not a maintainer of the project 👍

Thanks for your suggestions ,happy cooperation.

JoaoBraveCoding:
@mikechengwei can you fix the failing CI

mikechengwei:
> @mikechengwei can you fix the failing CI



> @mikechengwei can you fix the failing CI

ok.

JoaoBraveCoding (approved):
lgtm 👍 

Thanks for all the work, hope we can merge this in the future

mikechengwei:
> lgtm 👍
> 
> Thanks for all the work, hope we can merge this in the future

I thought so too, but it seems that no one in the community has reviewed the PR.

PhilipGough:
@mikechengwei - sorry, I missed this one but will get it reviewed ASAP

mikechengwei:
> sorry, I missed this one but will get it reviewed ASAP

Is there any new progress? @PhilipGough

Issue #5067, opened 2022-10-05
Operator trying to recreate the statefulset in a loop


mcorbin:
**What happened?**

I'm deploying Prometheus operator using [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack).
When I add the variable `minReadySeconds` on `prometheusSpec` (for example: `minReadySeconds: 180`) the operator enters in a loop where it deletes and creates the statefulset all the time (pods are immediately terminated once created).

**Did you expect to see something different?**

I expected to have the operator starting the statefulset as expected

**How to reproduce it (as minimally and precisely as possible)**:

In my case, adding `minReadySeconds: 180` to the `Prometheus`resource spec is enough to trigger the bug.

**Environment**

* Prometheus Operator version:  quay.io/prometheus-operator/prometheus-operator:v0.59.2

* Kubernetes version information:

```
Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.11-eks-18ef993", GitCommit:"b9628d6d3867ffd84c704af0befd31c7451cdc37", GitTreeState:"clean", BuildDate:"2022-07-06T18:06:23Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: AWS EKS

* Manifests:

I'm using [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) latest version (`40.3.1`).

Adding `minReadySeconds: 180` to `prometheusSpec` (which add it to the Prometheus resource spec) triggers the issue. 

* Prometheus Operator Logs:

```
key=monitoring/prometheus-operator-kube-p-prometheus statefulset=prometheus-prometheus-operator-kube-p-prometheus shard=0 msg="recreating StatefulSet because the update operation wasn't possible" reason="Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden"
level=info ts=2022-10-05T09:51:53.488912666Z caller=operator.go:1559 component=prometheusoperator key=monitoring/prometheus-operator-kube-p-prometheus msg="update prometheus status"
level=info ts=2022-10-05T09:51:53.493656582Z caller=operator.go:1390 component=prometheusoperator key=monitoring/prometheus-operator-kube-p-prometheus msg="sync prometheus"
[...]

PR #5166, opened 2022-11-15
Improves comment on field MinReadySeconds for CommonPrometheusFields and AlertmanagerSpec

JoaoBraveCoding:
## Description

Close: https://github.com/prometheus-operator/prometheus-operator/issues/5067

Problem: the field StatefulSetMinReadySeconds was introduced in 1.22 and until 1.25 when it stopped being alpha this field is not reset in the statefulset code, this makes it so if the field is specified and the feature gate is not enabled the stateful will just crash loop

Solution: improve the field comments to prevent users from using it if the feature is not enabled

## Type of change

## Changelog entry

```release-note
Improves comment on field MinReadySeconds for CommonPrometheusFields and AlertmanagerSpec
```

### DIFF
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -313,7 +313,7 @@ type CommonPrometheusFields struct {
 	// Minimum number of seconds for which a newly created pod should be ready
 	// without any of its container crashing for it to be considered available.
 	// Defaults to 0 (pod will be considered available as soon as it is ready)
-	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
+	// This is an alpha field from kubernetes 1.22 until 1.24 which requires enabling StatefulSetMinReadySeconds feature gate.
 	// +optional
 	MinReadySeconds *uint32 `json:"minReadySeconds,omitempty"`
 	// Pods' hostAliases configuration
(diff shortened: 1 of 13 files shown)


pkg/apis/monitoring/v1/types.go
```diff
@@ -313,7 +313,7 @@ type CommonPrometheusFields struct {
 	// Minimum number of seconds for which a newly created pod should be ready
 	// without any of its container crashing for it to be considered available.
 	// Defaults to 0 (pod will be considered available as soon as it is ready)
-	// This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
+	// This is an alpha field from kubernetes 1.22 until 1.24 which requires enabling StatefulSetMinReadySeconds feature gate.
```
simonpasquier, line 316:
```suggestion
	// This is an alpha field from kubernetes 1.22 until 1.24 which requires enabling the StatefulSetMinReadySeconds feature gate.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -2037,7 +2037,7 @@ type AlertmanagerSpec struct {
...
```
simonpasquier, line 2040:
```suggestion
	// This is an alpha field from kubernetes 1.22 until 1.24 which requires enabling the StatefulSetMinReadySeconds feature gate.
```

simonpasquier:
You'd need to update the description of `minReadySeconds`  for ThanosRuler too (in pkg/apis/monitoring/v1/thanos_types.go).

> After updating the comment I'm not so sure if this enough to fix the issue, couldn't the operator also check to see if the feature is enabled? Like it's done in [1] if so I'm happy to also code that in for this PR :)

I don't think that there's a way for clients to know whether a feature gate is enabled or not unfortunately.

simonpasquier (approved):
thanks!

PR #5171, opened 2022-11-17
Standardize how the Prometheus and Alertmanager versions are calculated

JoaoBraveCoding:
## Description

Close: https://github.com/prometheus-operator/prometheus-operator/issues/4893

Problem: since the inception of the operator a lot of options were introduced to customize the image ran by the statefulset for both Prometheus and Alertmanager, this made it so some inconsistencies on how the version is calculated were introduced between the two.

Solution: this commit introduces the function GetVersion which has the goal of returning the tag of the image that will be used in the statefulsets. If the image is tagless or uses a SHA256 it will instead return the DefaultVersion for that component if Spec.Version is not defined.

## Type of change

## Changelog entry

```release-note
Standardize how the Prometheus and Alertmanager versions are calculated
```

### DIFF
--- a/pkg/operator/image.go
+++ b/pkg/operator/image.go
@@ -27,33 +27,33 @@ import (
 // If image contains a tag or digest then image will be returned.
 // Otherwise, return image suffixed by either SHA, tag or version(in that order).
 // Inspired by kubernetes code handling of image building.
-func BuildImagePath(specImage, image, version, tag, sha string) (string, error) {
+func BuildImagePath(specImage, baseImage, version, tag, sha string) (string, error) {
 	if strings.TrimSpace(specImage) != "" {
 		return specImage, nil
 	}
-	named, err := dockerref.ParseNormalizedNamed(image)
+	named, err := dockerref.ParseNormalizedNamed(baseImage)
 	if err != nil {
-		return "", fmt.Errorf("couldn't parse image reference %q: %v", image, err)
+		return "", fmt.Errorf("couldn't parse image reference %q: %v", baseImage, err)
 	}
 	_, isTagged := named.(dockerref.Tagged)
 	_, isDigested := named.(dockerref.Digested)
 	if isTagged || isDigested {
-		return image, nil
+		return baseImage, nil
 	}
 
 	if sha != "" {
-		return fmt.Sprintf("%s@sha256:%s", image, sha), nil
+		return fmt.Sprintf("%s@sha256:%s", baseImage, sha), nil
 	} else if tag != "" {
 		imageTag, err := dockerref.WithTag(named, tag)
 		if err != nil {
 			return "", err
 		}
 		return imageTag.String(), nil
-	} else if version != "" {
-		return image + ":" + version, nil
 	}
-
-	return image, nil
+	if version == "" {
+		return "", fmt.Errorf("couldn't generate an image reference since version is empty")
+	}
+	return baseImage + ":" + version, nil
 }
 
 // StringValOrDefault returns the default val if the
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -105,12 +105,6 @@ func makeStatefulSet(
 	shard int32,
 	tlsAssetSecrets []string,
 ) (*appsv1.StatefulSet, error) {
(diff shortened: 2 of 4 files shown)


pkg/prometheus/statefulset.go
```diff
@@ -310,23 +304,28 @@ func makeStatefulSetSpec(
...
-	prometheusImagePath, err := operator.BuildImagePath(
+	pVersion := operator.StringValOrDefault(p.Spec.Version, operator.DefaultPrometheusVersion)
+	pImagePath, err := operator.BuildImagePath(
 		operator.StringPtrValOrDefault(p.Spec.Image, ""),
 		operator.StringValOrDefault(p.Spec.BaseImage, c.PrometheusDefaultBaseImage),
-		p.Spec.Version,
-		p.Spec.Tag,
-		p.Spec.SHA,
+		pVersion,
+		operator.StringValOrDefault(p.Spec.Tag, ""),
```
slashpai, line 317: Do we need `operator.StringValOrDefault()` here since default string value is `""`?

pkg/prometheus/statefulset.go
```diff
@@ -310,23 +304,28 @@ func makeStatefulSetSpec(
...
+		operator.StringValOrDefault(p.Spec.SHA, ""),
```
slashpai, line 318: same here

pkg/prometheus/statefulset.go
```diff
@@ -310,23 +304,28 @@ func makeStatefulSetSpec(
...
 	ruleConfigMapNames []string,
 	tlsAssetSecrets []string,
-	version semver.Version,
 ) (*appsv1.StatefulSetSpec, error) {
 	// Prometheus may take quite long to shut down to checkpoint existing data.
 	// Allow up to 10 minutes for clean termination.
 	terminationGracePeriod := int64(600)
 
-	prometheusImagePath, err := operator.BuildImagePath(
+	pVersion := operator.StringValOrDefault(p.Spec.Version, operator.DefaultPrometheusVersion)
```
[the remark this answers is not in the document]
simonpasquier: agreed

pkg/prometheus/statefulset_test.go
```diff
@@ -855,13 +855,13 @@ func TestPrometheusDefaultBaseImageFlag(t *testing.T) {
 			Labels:      labels,
 			Annotations: annotations,
 		},
-	}, prometheusBaseImageConfig, nil, "", 0, nil)
+	}, operatorConfig, nil, "", 0, nil)
 
 	require.NoError(t, err)
 
 	image := sset.Spec.Template.Spec.Containers[0].Image
 	expected := "nondefaultuseflag/quay.io/prometheus/prometheus"
```
simonpasquier, line 863: In pkg/alertmanager/statefulset_test.go, we append `:<default version>` to the image path and check for equality. From my POV, it's a good approach (we want the version number to be present if not specified in the image field).

pkg/prometheus/statefulset.go
JoaoBraveCoding: Only trying to keep the code consistent since `alertmanager/statefulset.go` also implements the same behaviour. Will be happy to revert this, but if we do so I would be in favor of also removing this function call in the `alertmanager/statefulset.go` code

[1] https://github.com/prometheus-operator/prometheus-operator/blob/4e026ff8f9ccac7ab12a7118a94f316ae5e8f183/pkg/alertmanager/statefulset.go#L226-L227

simonpasquier: A follow-up PR to clean up everything sounds good to me :)

pkg/operator/image.go
```diff
@@ -27,33 +27,30 @@ import (
...
 		if err != nil {
 			return "", err
 		}
 		return imageTag.String(), nil
-	} else if version != "" {
-		return image + ":" + version, nil
 	}
-
-	return image, nil
+	return baseImage + ":" + version, nil
```
simonpasquier, line 53: Should we guard against an empty version value? I understand that it shouldn't be the case anymore given that all 3 controllers use the hardcoded default version value but it would help preventing regressions (or at least making it more obvious to detect).

PR #5174, opened 2022-11-18
Refactors version checks on prometheus/statefulset.go to use ConfigGenerator

JoaoBraveCoding:
## Description

Problem: depending on the Prometheus version used some command line
arguments are valid where others aren't and the operator should log
a warning. Previously this was done with an "if/else" statement, however
since the operator now has ConfigGenerator we can use it to simplify the
code

Solution: start using the ConfigGenerator in prometheus/statefulset.
Unfortunately, when providing a default for a certain version or above it's not
entirely possible to get rid of the "if" part of the original "if/else"
statement

## Type of change

## Changelog entry

```release-note
Refactors version checks on prometheus/statefulset.go to use ConfigGenerator
```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -147,6 +147,27 @@ func (cg *ConfigGenerator) AppendMapItem(m yaml.MapSlice, k string, v interface{
 	return append(m, yaml.MapItem{Key: k, Value: v})
 }
 
+// AppendCommandlineArgument appends the name/v argument to the given []monitoringv1.Argument and returns
+// the updated slice.
+func (cg *ConfigGenerator) AppendCommandlineArgument(m []monitoringv1.Argument, argument monitoringv1.Argument) []monitoringv1.Argument {
+	if cg.notCompatible {
+		level.Warn(cg.logger).Log("msg", fmt.Sprintf("ignoring command line argument %q not supported by Prometheus", argument.Name))
+		return m
+	}
+
+	return append(m, argument)
+}
+
+// IsCompatible return true or false depending if the version being used is compatible
+func (cg *ConfigGenerator) IsCompatible(name string) bool {
+	if cg.notCompatible {
+		level.Warn(cg.logger).Log("msg", fmt.Sprintf("ignoring %q not supported by Prometheus", name))
+		return false
+	}
+
+	return true
+}
+
 type limitKey struct {
 	specField       string
 	prometheusField string
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -140,7 +140,7 @@ func (cg *ConfigGenerator) WithMaximumVersion(version string) *ConfigGenerator {
 // the updated slice.
 func (cg *ConfigGenerator) AppendMapItem(m yaml.MapSlice, k string, v interface{}) yaml.MapSlice {
 	if cg.notCompatible {
-		level.Warn(cg.logger).Log("msg", fmt.Sprintf("ignoring %q not supported by Prometheus", k))
+		cg.Warn(k)
 		return m
 	}
 
@@ -159,13 +159,13 @@ func (cg *ConfigGenerator) AppendCommandlineArgument(m []monitoringv1.Argument, 
 }
 
 // IsCompatible return true or false depending if the version being used is compatible
-func (cg *ConfigGenerator) IsCompatible(name string) bool {
-	if cg.notCompatible {
-		level.Warn(cg.logger).Log("msg", fmt.Sprintf("ignoring %q not supported by Prometheus", name))
-		return false
-	}
+func (cg *ConfigGenerator) IsCompatible() bool {
+	return !cg.notCompatible
+}
 
-	return true
+// Warn logs a warning.
+func (cg *ConfigGenerator) Warn(field string) {
+	level.Warn(cg.logger).Log("msg", fmt.Sprintf("ignoring %q not supported by Prometheus", field))
 }
 
 type limitKey struct {
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -403,9 +403,12 @@ func makeStatefulSetSpec(
 		}
 	}
 
-	// TODO(simonpasquier): check that the Prometheus version supports the flag.
 	if p.Spec.Web != nil && p.Spec.Web.PageTitle != nil {
-		promArgs = append(promArgs, monitoringv1.Argument{Name: "web.page-title", Value: *p.Spec.Web.PageTitle})
+		if version.GTE(semver.MustParse("2.6.0")) {
+			promArgs = append(promArgs, monitoringv1.Argument{Name: "web.page-title", Value: *p.Spec.Web.PageTitle})
+		} else {
+			level.Warn(logger).Log("msg", "ignoring 'pageTitle' not supported by Prometheus", "version", version, "minimum_version", "2.6.0")
+		}
 	}
 
 	if p.Spec.EnableAdminAPI {
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -28,7 +28,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
-	"github.com/blang/semver/v4"
 	"github.com/go-kit/log"
 	"github.com/pkg/errors"
 
@@ -99,6 +98,7 @@ func makeStatefulSet(
 	name string,
 	p monitoringv1.Prometheus,
 	config *operator.Config,
+	cg *ConfigGenerator,
 	ruleConfigMapNames []string,
 	inputHash string,
 	shard int32,
@@ -116,7 +116,7 @@ func makeStatefulSet(
 		p.Spec.Replicas = &intZero
 	}
 
-	spec, err := makeStatefulSetSpec(logger, p, config, shard, ruleConfigMapNames, tlsAssetSecrets)
+	spec, err := makeStatefulSetSpec(logger, p, config, cg, shard, ruleConfigMapNames, tlsAssetSecrets)
 	if err != nil {
 		return nil, errors.Wrap(err, "make StatefulSet spec")
 	}
@@ -300,6 +300,7 @@ func makeStatefulSetSpec(
 	logger log.Logger,
 	p monitoringv1.Prometheus,
 	c *operator.Config,
+	cg *ConfigGenerator,
 	shard int32,
 	ruleConfigMapNames []string,
 	tlsAssetSecrets []string,
@@ -307,10 +308,6 @@ func makeStatefulSetSpec(
 	// Prometheus may take quite long to shut down to checkpoint existing data.
 	// Allow up to 10 minutes for clean termination.
 	terminationGracePeriod := int64(600)
-	cg, err := NewConfigGenerator(logger, &p, false)
-	if err != nil {
-		return nil, err
-	}
 
 	promVersion := operator.StringValOrDefault(p.Spec.Version, operator.DefaultPrometheusVersion)
 	pImagePath, err := operator.BuildImagePath(
@@ -324,13 +321,8 @@ func makeStatefulSetSpec(
 		return nil, err
 	}
 
-	version, err := semver.ParseTolerant(promVersion)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to parse prometheus version")
-	}
-
-	if version.Major != 2 {
-		return nil, errors.Errorf("unsupported Prometheus major version %s", version)
+	if cg.version.Major != 2 {
+		return nil, errors.Errorf("unsupported Prometheus major version %s", cg.version)
 	}
 
 	promArgs := []monitoringv1.Argument{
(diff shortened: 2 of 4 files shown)


simonpasquier (commented):
it would be the occasion to leverage the ConfigGenerator struct as it would avoid the repetitive "if version >= x { ... } else { log ... }" blocks (and also make it consistent).

pkg/prometheus/promcfg.go
```diff
@@ -147,6 +147,27 @@ func (cg *ConfigGenerator) AppendMapItem(m yaml.MapSlice, k string, v interface{
...
+	if cg.notCompatible {
+		level.Warn(cg.logger).Log("msg", fmt.Sprintf("ignoring command line argument %q not supported by Prometheus", argument.Name))
+		return m
+	}
+
+	return append(m, argument)
+}
+
+// IsCompatible return true or false depending if the version being used is compatible
+func (cg *ConfigGenerator) IsCompatible(name string) bool {
```
JoaoBraveCoding, line 162: At the moment this only avoids one `if/else` on line 539 of promehteus/stefulset.go, I'm also up for removing it and leaving the `if/else`

simonpasquier: sounds good to me.
Maybe modify the function to make it more obvious that it can also log a warning?

```suggestion
func (cg *ConfigGenerator) IsCompatible(field string) bool {
    return cg.notCompatible
}

func (cg *ConfigGenerator) IsCompatibleOrWarn(field string) bool {
...
}
```

pkg/prometheus/promcfg.go
```diff
@@ -147,6 +147,27 @@ func (cg *ConfigGenerator) AppendMapItem(m yaml.MapSlice, k string, v interface{
 	return append(m, yaml.MapItem{Key: k, Value: v})
 }
 
+// AppendCommandlineArgument appends the name/v argument to the given []monitoringv1.Argument and returns
+// the updated slice.
+func (cg *ConfigGenerator) AppendCommandlineArgument(m []monitoringv1.Argument, argument monitoringv1.Argument) []monitoringv1.Argument {
```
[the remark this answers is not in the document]
simonpasquier: :thinking: Go generics might be a solution here but there might be a more elegant solution...

```suggestion
func AppendToSlice[T any](cg *ConfigGenerator, slice []T, elems ...T) []T {
	return append(slice, elems...)
}
```

pkg/prometheus/statefulset.go
```diff
@@ -315,6 +314,10 @@ func makeStatefulSetSpec(
 	// Prometheus may take quite long to shut down to checkpoint existing data.
 	// Allow up to 10 minutes for clean termination.
 	terminationGracePeriod := int64(600)
+	cg, err := NewConfigGenerator(logger, &p, false)
```
simonpasquier, line 317: I would pass to the function the config generator already instantiated by the controller.

pkg/prometheus/statefulset.go
```diff
@@ -336,34 +339,18 @@ func makeStatefulSetSpec(
 		{Name: "web.console.libraries", Value: "/etc/prometheus/console_libraries"},
 	}
 
-	// TODO(simonpasquier): log a warning message if the Prometheus version
-	// doesn't support the flag (do it everywhere it needs to be, not only for
-	// this block).
-	retentionTimeFlag := monitoringv1.Argument{Name: "storage.tsdb.retention"}
+	retentionTimeFlag := monitoringv1.Argument{Name: "storage.tsdb.retention", Value: defaultRetention}
 	if version.GTE(semver.MustParse("2.7.0")) {
```
simonpasquier, line 343:
```suggestion
	if cg.WithMinimumVersion("2.7.0").IsCompatible() {
```

pkg/prometheus/statefulset.go
```diff
@@ -441,22 +416,21 @@ func makeStatefulSetSpec(
...
-		if p.Spec.LogFormat != "" && p.Spec.LogFormat != "logfmt" {
-			promArgs = append(promArgs, monitoringv1.Argument{Name: "log.format", Value: p.Spec.LogFormat})
-		}
+
+	if p.Spec.LogFormat != "" && p.Spec.LogFormat != "logfmt" {
+		promArgs = cg.WithMinimumVersion("2.6.0").AppendCommandlineArgument(promArgs, monitoringv1.Argument{Name: "log.format", Value: p.Spec.LogFormat})
 	}
 
-	if version.GTE(semver.MustParse("2.11.0")) && p.Spec.WALCompression != nil {
+	if p.Spec.WALCompression != nil {
```
simonpasquier, line 424: A bit less verbose to my taste :)
```
if p.Spec.WALCompression != nil {
    arg := "no-storage.tsdb.wal-compression"
    if *p.Spec.WALCompression {
        arg = "storage.tsdb.wal-compression"
    }
    cg.WithMinimumVersion("2.11.0").AppendCommandlineArgument(...)
}
```

pkg/prometheus/statefulset.go
```diff
@@ -558,16 +535,16 @@ func makeStatefulSetSpec(
...
+	var fields monitoringv1.WebConfigFileFields
+	if p.Spec.Web != nil && cg.WithMinimumVersion("2.24.0").IsCompatible("web.config.file") {
+		fields = p.Spec.Web.WebConfigFileFields
+	}
+
 	// Mount web config and web TLS credentials as volumes.
 	// We always mount the web config file for versions greater than 2.24.0.
 	// With this we avoid redeploying prometheus when reconfiguring between
 	// HTTP and HTTPS and vice-versa.
 	if version.GTE(semver.MustParse("2.24.0")) {
```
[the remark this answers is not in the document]
simonpasquier: If we offer 2 versions for checking compatibility (one that logs a warning and the other that doesn't), we could use ConfigGenerator everywhere?

```suggestion
	if cg.WithMinimumVersion("2.24.0").IsCompatible() {
```

pkg/prometheus/statefulset_test.go
```diff
@@ -1353,7 +1353,7 @@ func TestRetentionAndRetentionSize(t *testing.T) {
 		{"v2.5.0", "1d", "512MB", "--storage.tsdb.retention=1d", "--storage.tsdb.retention.size=512MB", true, false},
 		{"v2.7.0", "", "", "--storage.tsdb.retention.time=24h", "--storage.tsdb.retention.size=", true, false},
 		{"v2.7.0", "1d", "", "--storage.tsdb.retention.time=1d", "--storage.tsdb.retention.size=", true, false},
-		{"v2.7.0", "", "512MB", "--storage.tsdb.retention.time=24h", "--storage.tsdb.retention.size=512MB", false, true},
+		{"v2.7.0", "", "512MB", "--storage.tsdb.retention.time=24h", "--storage.tsdb.retention.size=512MB", true, true},
```
simonpasquier, line 1356: can you elaborate on why the test needs a change?

simonpasquier (commented):
nice! Left a few comments but looks good in general.

pkg/prometheus/statefulset_test.go
JoaoBraveCoding: Maybe I misunderstood but in this test we were validating that if when`retention.time` is not set in the CR then the operator should not set `retention.time`. However from the code what I understood was that we should set `--storage.tsdb.retention.time=24h` if nothing was defined in the CR. So either I understood that wrongfully or the test needs to be fixed.

simonpasquier: IIRC the semantics are:
* if retentionSize and retention are empty then the retention argument is set to 24h.
* if retention is empty but not retentionSize, the retention argument isn't set (e.g. the only limit is space).
* if retentionSize is empty but not retention, the retention argument is set (e.g. the only limit is time).
* if both aren't empty, both arguments are set.

And if the Prometheus version is < v2.7.0 then the operator assumes that retentionSize is empty.

Now the test itself isn't 100% correct: if we expect that no retention CLI argument is defined, it should ensure that no argument with the said prefix (e.g.  `--storage.tsdb.retention.time`) exists (right now it verifies that `--storage.tsdb.retention.time=24h` isn't present).

JoaoBraveCoding: Should be good now, reverted my change to this part of the test

JoaoBraveCoding:
> If we offer 2 versions for checking compatibility (one that logs a warning and the other that doesn't), we could use ConfigGenerator everywhere?

In theory sounds possible 👍

pkg/prometheus/statefulset.go
JoaoBraveCoding: This ended up generating a bit more work, due to having to re-write the parameters of the function but I agree that only having one ConfigGenerator is the way to go

pkg/prometheus/promcfg.go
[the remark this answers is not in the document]
JoaoBraveCoding: I tried it but didn't enjoy it so much, since it's already a big change I prefer to re-do this effort on another PR

pkg/prometheus/statefulset.go
```diff
@@ -321,48 +322,29 @@ func makeStatefulSetSpec(
...
+	if cg.WithMaximumVersion("2.7.0").IsCompatible() {
+		retentionTimeFlag.Name = "storage.tsdb.retention"
+	}
+	if p.Spec.Retention == "" && (p.Spec.RetentionSize == "" || cg.version.LT(semver.MustParse("2.7.0"))) {
+		promArgs = append(promArgs, retentionTimeFlag)
+	}
+	if p.Spec.Retention != "" {
+		retentionTimeFlag.Value = string(p.Spec.Retention)
+		promArgs = append(promArgs, retentionTimeFlag)
+	}
```
simonpasquier, line 344: (nit) maybe a bit easier to read?
```suggestion
	var (
		retentionTimeFlagName = "storage.tsdb.retention.time"
		retentionTimeFlagValue = p.Spec.Retention
	)
	if cg.WithMaximumVersion("2.7.0").IsCompatible() {
		retentionTimeFlagName = "storage.tsdb.retention"
		if p.Spec.Retention == "" {
			retentionTimeFlagValue = defaultRetention
		}
	} else if p.Spec.Retention == "" && p.Spec.RetentionSize == "" {
		retentionTimeFlagValue = defaultRetention
	}
	
	if retentionTimeFlagValue != "" {
		promArgs = append(promArgs, monitoringv1.Argument{
			Name: retentionTimeFlagName,
			Value: retentionTimeFlagValue,
		})
	}
```

pkg/prometheus/statefulset.go
```diff
@@ -437,22 +410,21 @@ func makeStatefulSetSpec(
...
-			promArgs = append(promArgs, monitoringv1.Argument{Name: "no-storage.tsdb.wal-compression"})
+			arg.Name = "storage.tsdb.wal-compression"
 		}
+		promArgs = cg.WithMinimumVersion("2.11.0").AppendCommandlineArgument(promArgs, arg)
 	}
 
-	if version.GTE(semver.MustParse("2.8.0")) && p.Spec.AllowOverlappingBlocks {
-		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.allow-overlapping-blocks"})
+	if p.Spec.AllowOverlappingBlocks {
+		promArgs = cg.WithMinimumVersion("2.11.0").AppendCommandlineArgument(promArgs, monitoringv1.Argument{Name: "storage.tsdb.allow-overlapping-blocks"})
```
simonpasquier, line 427: (nit) since 2.39.0, this flag has no effect and is deprecated.

simonpasquier (commented):
Some small nits otherwise lgtm

pkg/prometheus/statefulset.go
JoaoBraveCoding: I've added `.WithMaximumVersion("2.39.0")`

pkg/prometheus/statefulset.go
```diff
@@ -557,16 +529,16 @@ func makeStatefulSetSpec(
 		promVolumeMounts = append(promVolumeMounts, vmount)
 	}
 
+	var fields monitoringv1.WebConfigFileFields
+	if p.Spec.Web != nil && cg.WithMinimumVersion("2.24.0").IsCompatibleOrWarn("web.config.file") {
+		fields = p.Spec.Web.WebConfigFileFields
+	}
+
```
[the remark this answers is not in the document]
JoaoBraveCoding: We cannot just do `fields := p.Spec.Web.WebConfigFileFields` because of nil pointers so I've put a `p.Spec.Web != nil` inside

pkg/prometheus/statefulset.go
```diff
@@ -437,22 +414,21 @@ func makeStatefulSetSpec(
...
-			promArgs = append(promArgs, monitoringv1.Argument{Name: "no-storage.tsdb.wal-compression"})
+			arg.Name = "storage.tsdb.wal-compression"
 		}
+		promArgs = cg.WithMinimumVersion("2.11.0").AppendCommandlineArgument(promArgs, arg)
 	}
 
-	if version.GTE(semver.MustParse("2.8.0")) && p.Spec.AllowOverlappingBlocks {
-		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.allow-overlapping-blocks"})
+	if p.Spec.AllowOverlappingBlocks {
+		promArgs = cg.WithMinimumVersion("2.11.0").WithMaximumVersion("2.39.0").AppendCommandlineArgument(promArgs, monitoringv1.Argument{Name: "storage.tsdb.allow-overlapping-blocks"})
```
simonpasquier, line 431: I'm afraid that it won't do what we want for versions < 2.11.0

pkg/prometheus/promcfg.go
```diff
@@ -140,13 +140,45 @@ func (cg *ConfigGenerator) WithMaximumVersion(version string) *ConfigGenerator {
...
+}
+
+// IsCompatible return true or false depending if the version being used is compatible
+func (cg *ConfigGenerator) IsCompatible() bool {
+	return !cg.notCompatible
+}
+
+// IsCompatibleOrWarn return true or false depending if the version being used is compatible.
+// If false it will also log a warning.
+func (cg *ConfigGenerator) IsCompatibleOrWarn(field string) bool {
```
simonpasquier, line 168: not needed anymore?

pkg/prometheus/statefulset.go
JoaoBraveCoding: To validate I added a unit test and it works :D

simonpasquier: Nice! Unit tests FTW :tada:

simonpasquier (approved):
lgtm

Issue #4850, opened 2022-06-14
Add support for passing web.max-connections argument to Prometheus


vivekjainx86:
**What is missing?**
I didn't find any specification in PrometheusSpec to pass `--web.max-connections=<integer>` argument to Prometheus. Also, there is no support for passing configuration parameter to Prometheus which is not in PrometheusSpec.

**Why do we need it?**
It will be helpful if PrometheusSpec has all the configuration parameters that Prometheus has, or at least there should be a way to provide additional configuration parameters.

**Environment**

* Prometheus Operator version:

    `v2.32.1`

slashpai:
@vivekjainx86 would you like to contribute? :)

simonpasquier:
> or at least there should be a way to provide additional configuration parameters.

I've just filed #4851which is about this use case as it was an action item from our last office hours meetings :)

JoaoBraveCoding:
With #4851 being closed shouldn't this be closed as well? Or do we want dedicated support for this?

simonpasquier:
I'd keep the issue opened. The CRD should eventually allow setting this parameter: 

```
spec:
  web:
    maxConnections: 256
```

grv180290:
You can try something like this. It works for me in the values.yaml for helm. 

```
          prometheusSpec:
            additionalArgs:
            - name: web.max-connections
              value: "10000"
```

PR #5175, opened 2022-11-18
Adds Spec.Web.MaxConnections to Prometheus CRD

JoaoBraveCoding:
## Description

Closes: https://github.com/prometheus-operator/prometheus-operator/issues/4850

Problem: currently the Prometheus CRD only allows users to configure the flag web.max-connections using Spec.AdditionalArgs

Solution: introduce dedicated support for the flag under Spec.Web.MaxConnections

## Type of change

## Changelog entry

```release-note
Adds Spec.Web.MaxConnections to Prometheus CRD
```


pkg/prometheus/statefulset.go
```diff
@@ -405,8 +405,14 @@ func makeStatefulSetSpec(
...
 	// TODO(simonpasquier): check that the Prometheus version supports the flag.
-	if p.Spec.Web != nil && p.Spec.Web.PageTitle != nil {
-		promArgs = append(promArgs, monitoringv1.Argument{Name: "web.page-title", Value: *p.Spec.Web.PageTitle})
+	if p.Spec.Web != nil {
+		if p.Spec.Web.PageTitle != nil {
+			promArgs = append(promArgs, monitoringv1.Argument{Name: "web.page-title", Value: *p.Spec.Web.PageTitle})
+		}
+
+		if p.Spec.Web.MaxConnections != nil {
+			promArgs = append(promArgs, monitoringv1.Argument{Name: "web.max-connections", Value: fmt.Sprintf("%d", *p.Spec.Web.MaxConnections)})
```
JoaoBraveCoding, line 414: Not sure if I should've added a check on the prometheus version. This flag was introduced in prometheus 1.5.0 [1]

[1] https://github.com/prometheus/prometheus/blob/main/CHANGELOG.md#150--2017-01-23
[2] https://github.com/prometheus/prometheus/pull/2240

slashpai: Having removed support for `1.x` .  we don't need to check that IIUC

JoaoBraveCoding: Thanks for letting me know!

pkg/apis/monitoring/v1/types.go
```diff
@@ -796,6 +796,8 @@ type PrometheusWebSpec struct {
 	WebConfigFileFields `json:",inline"`
 	// The prometheus web page title
 	PageTitle *string `json:"pageTitle,omitempty"`
+	// Defines the maximum number of simultaneous connections
+	MaxConnections *int32 `json:"maxConnections,omitempty"`
```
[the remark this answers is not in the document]
JoaoBraveCoding: I've set the minimum for `--query.max-concurrency` to 1 since to me it seems to make the most sense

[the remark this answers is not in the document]
simonpasquier: Yep. I've checked starting Prometheus with `--query.max-concurrency=0` and Prometheus refused to stop. As for max connections, `--query.max-concurrency=-1` triggers a panic. I think that the panics need to be fixed in Prometheus (e.g. it should reject negative values).

simonpasquier (commented):
lgtm, it would be good to add tests for the new validations in https://github.com/prometheus-operator/prometheus-operator/blob/a168981ef61edfb46f6be79031eb4a2c22b0ed99/test/e2e/prometheus_test.go#L4158

test/framework/context.go
```diff
@@ -53,7 +53,11 @@ func (f *Framework) NewTestCtx(t *testing.T) *TestCtx {
 // cleanUpFns. This can e.g. be used to create multiple namespaces in the same
 // test context.
 func (ctx *TestCtx) GetObjID() string {
```
simonpasquier, line 55: To be honest, I never really looked into this code but I believe that we can refactor a bit and make the interfaces a bit clearer:
* I'd move the logic about truncating the string to 63 chars to `*Framework.CreateNamespace()`  since this is really where it matters.
* I'd also not expose the `ID` field (e.g. rename to `id` ) & rename this function to `ID()`  
* Leverage [UniqueVolumeName()](https://github.com/prometheus-operator/prometheus-operator/blob/4decb4b8690cbcea2308ef8c1a63f737d7f781fa/pkg/k8sutil/k8sutil.go#L271) to generate the unique name.

JoaoBraveCoding: Done 👍 I didn't do it because I wanted to check with you but I think it also makes sense to rename `UniqueVolumeName`  to something like `UniqueName` since it will probably be useful in other parts of the code.

test/framework/namespace.go
```diff
@@ -21,15 +21,21 @@ import (
...
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
 func (f *Framework) CreateNamespace(ctx context.Context, t *testing.T, testCtx *TestCtx) string {
-	name := testCtx.GetObjID()
+	name := testCtx.ID()
+	rn := k8sutil.ResourceNamer{}
+	name, err := rn.UniqueVolumeName(name)
```
simonpasquier, line 33: maybe rename `UniqueVolumeName()` to `DNS1123Label()`?

simonpasquier: UniqueVolumeName was ok-ish because it was obvious that you would use it for volume names. UniqueName doesn't convey the fact that it returns a string which is a valid DNS label. The Kubernetes [documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) already mentions `RFC 1123 Label Names`   and it's not even detailed which resources types follow this constraint :shrug:

JoaoBraveCoding: Alright, then I agree that it makes sense to go for `DNS1123Label` 👍

JoaoBraveCoding: In trying to make the functions clear, I've renamed:
- `UniqueVolumeName` -> `UniqueDNS1123Label`
- `VolumeName` -> `DNS1123Label`

simonpasquier (approved):
lgtm

PR #5183, opened 2022-11-23
Fixes regression introduced in thanos/statefulset.go when reading arguments from secrets

JoaoBraveCoding:
## Description

PR that introduced the regression: https://github.com/prometheus-operator/prometheus-operator/pull/5122

In the PR above Thanos started reading arguments from secrets instead of ENV vars, however, in the change, we forgot to specify a Path that is necessary when working with projection of secret keys to specific paths as can be seen in the k8s documentation [1]

[1] https://kubernetes.io/docs/concepts/configuration/secret/#projection-of-secret-keys-to-specific-paths

## Type of change

## Changelog entry

```release-note
Fixes regression introduced in thanos/statefulset.go when reading arguments from secrets
```

### DIFF
--- a/pkg/thanos/statefulset_test.go
+++ b/pkg/thanos/statefulset_test.go
@@ -208,7 +208,8 @@ func TestTracing(t *testing.T) {
 		secretName = "thanos-tracing-config-secret"
 		secretKey  = "config.yaml"
 		volumeName = "tracing-config"
-		mountPath  = "/etc/thanos/config/tracing-config.yaml"
+		mountPath  = "/etc/thanos/config/tracing-config"
+		fullPath   = "/etc/thanos/config/tracing-config/config.yaml"
 	)
 
 	sset, err := makeStatefulSet(&monitoringv1.ThanosRuler{
@@ -234,7 +235,7 @@ func TestTracing(t *testing.T) {
 		var containsVolume bool
 		for _, volume := range sset.Spec.Template.Spec.Volumes {
 			if volume.Name == volumeName {
-				if volume.Secret.SecretName == secretName && volume.Secret.Items[0].Key == secretKey {
+				if volume.Secret.SecretName == secretName && volume.Secret.Items[0].Key == secretKey && volume.Secret.Items[0].Path == secretKey {
 					containsVolume = true
 					break
 				}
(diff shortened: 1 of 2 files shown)


pkg/thanos/statefulset_test.go
```diff
@@ -234,7 +235,7 @@ func TestTracing(t *testing.T) {
 		var containsVolume bool
 		for _, volume := range sset.Spec.Template.Spec.Volumes {
 			if volume.Name == volumeName {
-				if volume.Secret.SecretName == secretName && volume.Secret.Items[0].Key == secretKey {
+				if volume.Secret.SecretName == secretName && volume.Secret.Items[0].Key == secretKey && volume.Secret.Items[0].Path == secretKey {
```
simonpasquier, line 238: shouldn't this be `volume.Secret.Items[0].Path == mountPath`?

JoaoBraveCoding: I don't think so. According to what is in the k8s documentation [1] `Path`  will be the suffix of the value in `mountPath`. So if we take:
```
spec:
  containers:
  - volumeMounts:
    - mountPath: "/etc/foo"
volumes:
  secrets:
    items:
    - key: "bar.yaml"
      path: "bar.yaml"
```

Then the file will be available under `/etc/foo/bar.yaml`, where if we were to use `mountPath` instead we would end up with:


```
spec:
  containers:
  - volumeMounts:
    - mountPath: "/etc/foo"
volumes:
  secrets:
    items:
    - key: "bar.yaml"
      path: "/etc/foo"
```

This would make it so the file would be under `/etc/foo/etc/foo` I think it would also work but it would probably be more confusing

[1] https://kubernetes.io/docs/concepts/configuration/secret/#projection-of-secret-keys-to-specific-paths

simonpasquier: My bad, sorry for the noise :)

simonpasquier (approved):
The issue was discovered when trying to bump the Prometheus operator version in OpenShift (see https://github.com/openshift/prometheus-operator/pull/209) . After applying this patch on top of v0.61.0 (see https://github.com/openshift/prometheus-operator/pull/211), the CI goes green.

@fpetkovski happy to prepare a v0.61.1 if you don't have time.

The regression wasn't caught beforehand because the e2e Thanos Ruler don't exercise `spec.queryConfig` and `spec.alertmanagersConfig` fields. @JoaoBraveCoding offered to file an issue so we remember that the e2e tests need improvements.

JoaoBraveCoding:
> The regression wasn't caught beforehand because the e2e Thanos Ruler don't exercise spec.queryConfig and spec.alertmanagersConfig fields. @JoaoBraveCoding offered to file an issue so we remember that the e2e tests need improvements.

Here is the issue https://github.com/prometheus-operator/prometheus-operator/issues/5187

Issue #2747, opened 2019-09-03
Support setting ImagePullPolicy for operand (prometheus, alertmanager) images


shettich:
A quick read of statefullset.go shows it never looks at the field from the CRD.

Could you propagate that over?

brancz:
Can you share the object you created?

shettich:
Not easily.  I could work out something close, but then I have no way to test it.

brancz:
Then it's going to be hard for us to help you :)

brancz:
would https://github.com/coreos/kube-prometheus help you?

PR #5203, opened 2022-12-04
Added ability to specify imagePullPolicy

michalschott:
## Description

Should be able to set imagePullPolicy for pods.

Fixes https://github.com/prometheus-operator/prometheus-operator/issues/2747

## Type of change

## Changelog entry

```release-note
Added ability to specify imagePullPolicy
```

### DIFF
--- a/pkg/apis/monitoring/v1/thanos_types.go
+++ b/pkg/apis/monitoring/v1/thanos_types.go
@@ -68,6 +68,10 @@ type ThanosRulerSpec struct {
 	PodMetadata *EmbeddedObjectMetadata `json:"podMetadata,omitempty"`
 	// Thanos container image URL.
 	Image string `json:"image,omitempty"`
+	// Image pull policy for the 'prometheus', 'init-config-reloader' and 'config-reloader' containers.
+	// See https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy for more details.
+	// +kubebuilder:validation:Enum="";Always;Never;IfNotPresent
+	ImagePullPolicy v1.PullPolicy `json:"imagePullPolicy,omitempty"`
 	// An optional list of references to secrets in the same namespace
 	// to use for pulling thanos images from registries
 	// see http://kubernetes.io/docs/user-guide/images#specifying-imagepullsecrets-on-a-pod
--- a/pkg/apis/monitoring/v1/thanos_types.go
+++ b/pkg/apis/monitoring/v1/thanos_types.go
@@ -68,7 +68,7 @@ type ThanosRulerSpec struct {
 	PodMetadata *EmbeddedObjectMetadata `json:"podMetadata,omitempty"`
 	// Thanos container image URL.
 	Image string `json:"image,omitempty"`
-	// Image pull policy for the 'prometheus', 'init-config-reloader' and 'config-reloader' containers.
+	// Image pull policy for the 'thanos', 'init-config-reloader' and 'config-reloader' containers.
 	// See https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy for more details.
 	// +kubebuilder:validation:Enum="";Always;Never;IfNotPresent
 	ImagePullPolicy v1.PullPolicy `json:"imagePullPolicy,omitempty"`
--- a/pkg/apis/monitoring/v1/types.go
+++ b/pkg/apis/monitoring/v1/types.go
@@ -96,6 +96,10 @@ type CommonPrometheusFields struct {
 	// Prometheus Operator knows what version of Prometheus is being
 	// configured.
 	Image *string `json:"image,omitempty"`
+	// Image pull policy for the 'prometheus', 'init-config-reloader' and 'config-reloader' containers.
+	// See https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy for more details.
+	// +kubebuilder:validation:Enum="";Always;Never;IfNotPresent
+	ImagePullPolicy v1.PullPolicy `json:"imagePullPolicy,omitempty"`
 	// An optional list of references to secrets in the same namespace
 	// to use for pulling prometheus and alertmanager images from registries
 	// see http://kubernetes.io/docs/user-guide/images#specifying-imagepullsecrets-on-a-pod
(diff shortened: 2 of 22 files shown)


pkg/operator/config.go
```diff
@@ -47,11 +48,12 @@ type Config struct {
...
-	CPULimit      string
-	MemoryRequest string
-	MemoryLimit   string
-	Image         string
+	CPURequest      string
+	CPULimit        string
+	MemoryRequest   string
+	MemoryLimit     string
+	Image           string
+	ImagePullPolicy v1.PullPolicy
```
simonpasquier, line 56: The image pull policy should be treated the same way as the other reloader options that depend on the Prometheus/Alertmanager object. E.g. it should have an option function like [LogFormat()](https://github.com/prometheus-operator/prometheus-operator/blob/49a12372e228f881ea388d480b45736fa6cc97cd/pkg/operator/config-reloader.go#L104-L109) that gets called when [generating](https://github.com/prometheus-operator/prometheus-operator/blob/49a12372e228f881ea388d480b45736fa6cc97cd/pkg/prometheus/statefulset.go#L915-L926) the config reloader containers.

pkg/apis/monitoring/v1/types.go
```diff
@@ -96,6 +96,8 @@ type CommonPrometheusFields struct {
 	// Prometheus Operator knows what version of Prometheus is being
 	// configured.
 	Image *string `json:"image,omitempty"`
+	// By not specyfing ImagePullPolicy default value will be used
```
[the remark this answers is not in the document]
michalschott: Actioned.

pkg/apis/monitoring/v1/types.go
```diff
@@ -1898,6 +1900,8 @@ type AlertmanagerSpec struct {
 	// Prometheus Operator knows what version of Alertmanager is being
 	// configured.
 	Image *string `json:"image,omitempty"`
+	// By not specyfing ImagePullPolicy default value will be used
```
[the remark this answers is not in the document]
michalschott: Actioned.

pkg/operator/config.go
michalschott: Actioned.

michalschott:
During manual test found out that `imagePullPolicy` for `config-reloader` containers is not set properly and it's always set to "IfNotPresent", struggling to spot out why is this happening. Am I missing something here?

pkg/apis/monitoring/v1/thanos_types.go
```diff
@@ -68,6 +68,10 @@ type ThanosRulerSpec struct {
 	PodMetadata *EmbeddedObjectMetadata `json:"podMetadata,omitempty"`
 	// Thanos container image URL.
 	Image string `json:"image,omitempty"`
+	// Image pull policy for the 'thanos', 'init-config-reloader' and 'config-reloader' containers.
+	// See https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy for more details.
+	// +kubebuilder:validation:Enum="";Always;Never;IfNotPresent
```
JoaoBraveCoding, line 73: Should we really allow users to set this to an empty string?
Also shouldn't we set a default? In my opinion it should be `Always`
```suggestion
	// +kubebuilder:validation:Enum=Always;Never;IfNotPresent
```

JoaoBraveCoding (commented):
Just a minor comment, apart from it looks good! Thank you for the contribution 👍

pkg/apis/monitoring/v1/thanos_types.go
michalschott: Agree this should be set to `Always` by default, altho I believe in k8s `IfNotPresent` being default is to either maintain compatibility for air gapped environments or being able to run containers despite not being able to reach out remote registry.

michalschott:
> During manual test found out that `imagePullPolicy` for `config-reloader` containers is not set properly and it's always set to `IfNotPresent`, struggling to spot out why is this happening. Am I missing something here?

Managed to fix it.

pkg/apis/monitoring/v1/thanos_types.go
simonpasquier: We want consistency with Kubernetes here. An empty `imagePullPolicy` doesn't necessarily default to `Always` (see https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting) and I don't think we want to surprise people here.

pkg/operator/config.go
simonpasquier, line 56:
```suggestion
```

michalschott: Is this intentional to not be able to set imagePullPolicy for config-reloader?

michalschott: Actually it is being set but that function.

pkg/operator/config-reloader.go
```diff
@@ -129,6 +129,13 @@ func Shard(shard int32) ReloaderOption {
 	}
 }
 
+// ImagePullPolicy sets the imagePullPolicy option for the config-reloader container
+func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
+	return func(c *ConfigReloader) {
+		c.config.ImagePullPolicy = imagePullPolicy
```
simonpasquier, line 135: imagePullPolicy needs to be a field of ConfigReloader (consistent with Shard() and other functions).

```suggestion
		c.imagePullPolicy = imagePullPolicy
```

pkg/operator/config.go
simonpasquier: Ah I see the misunderstanding now. See my comment above.

michalschott:
@simonpasquier now it makes sense, I though it would be good to keep `imagePullPolicy` together with `Image` config.

Actioned.

pkg/apis/monitoring/v1/types.go
```diff
@@ -1897,6 +1901,10 @@ type AlertmanagerSpec struct {
 	// Prometheus Operator knows what version of Alertmanager is being
 	// configured.
 	Image *string `json:"image,omitempty"`
+	// Image pull policy for the 'prometheus', 'init-config-reloader' and 'config-reloader' containers.
```
simonpasquier, line 1904:
```suggestion
	// Image pull policy for the 'alertmanager', 'init-config-reloader' and 'config-reloader' containers.
```

simonpasquier (commented):
lgtm minor one last nit. Thanks for the effort!

simonpasquier (approved):
Thanks again! This will merge on green.

michalschott:
> Thanks again! This will merge on green.

Just pushed changes made by `make generate`

michalschott:
@simonpasquier I guess you need to click a button again ;)

PhilipGough (approved):

PR #5221, opened 2022-12-12
pkg/operator: create Select method for rule selection 

### LINKED ISSUE #5191: Reject invalid PrometheusRule objects beforehand rather than failing the reconciliation
**What is missing?**

When the Thanos/Prometheus controllers detect an invalid PrometheusRule object (invalid PromQL expression for instance), they fail the reconciliation (and for Prometheus, the `Reconciled`  condition will be set to False). Instead the controllers should reject the invalid object and proceed (just like we do for service/pod monitors and probes).

**Why do we need it?**

In self-service environments, blocking the reconciliation means that one single user can block all other users if they push a bad rule.

**Environment**

Probably a good time to refactor the code handling the rules which is duplicated between pkg/prometheus and pkg/thanos.

slashpai:
Fixes #5191

## Description

## Type of change

## Changelog entry

```release-note
Update logic on selecting prometheus rules by rejecting invalid ones instead of failing reconciliation
```


slashpai:
Removing draft for initial review

pkg/prometheus/rules.go
```diff
@@ -190,32 +189,43 @@ func (c *Operator) selectRules(p *monitoringv1.Prometheus, namespaces []string)
...
-			return nil, marshalErr
+	}
+
+	var rejected int
+	rules := make(map[string]string, len(promRules))
+
+	for ruleName, promRule := range promRules {
+		var err error
+		if err := nsLabeler.EnforceNamespaceLabel(promRule); err != nil {
+			break
```
simonpasquier, line 212: this should be `continue` + you want something like we do in selectProbes()
https://github.com/prometheus-operator/prometheus-operator/blob/a382d0952fc48afa49bf71d4803186ce866956dd/pkg/prometheus/operator.go#L2220-L2229

pkg/prometheus/rules.go
```diff
@@ -190,32 +189,43 @@ func (c *Operator) selectRules(p *monitoringv1.Prometheus, namespaces []string)
...
+		}
+
+		content, err := generateRulesConfiguration(promRule.Spec, c.logger)
+		if err != nil {
+			rejected++
+			level.Warn(c.logger).Log(
+				"msg", "skipping prometheusrule",
+				"error", err.Error(),
+				"prometheusrule", promRule.Name,
+				"namespace", p.Namespace,
```
JoaoBraveCoding, line 222: shouldn't it be the rule namespace? Since users will want to know in which namespace is the prometheus rule that contains the error 

```suggestion
				"namespace", promRule.Namespace,
```

slashpai: Correct. my bad :)

pkg/prometheus/rules.go
slashpai: done

JoaoBraveCoding (approved):
lgtm 👍

slashpai:
Need to implement for Thanos as well. I will look in reducing duplicate code between Thanos and Prometheus

slashpai:
@simonpasquier Can you take another look at this?

simonpasquier (commented):
Can we factorize the 2 `selectRules()` methods (probably in pkg/operator)?

pkg/operator/rules.go
```diff
@@ -24,10 +24,28 @@ import (
...
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
+	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
+	namespacelabeler "github.com/prometheus-operator/prometheus-operator/pkg/namespace-labeler"
 	thanostypes "github.com/thanos-io/thanos/pkg/store/storepb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
 )
 
+type PromRuleCommonFields struct {
```
slashpai, line 37: Tried this approach to avoid passing all these fields to function. Also looking for better name for this struct :)

slashpai:
@simonpasquier PTAL when you get a chance :)

pkg/operator/rules.go
simonpasquier: I'd suggest to leave the initialization of the label selector and namespace labeler to the caller. It will make the struct more lightweight + Prometheus & Thanos don't create the same ns labeler (because of the partial response strategy).

```golang
type PrometheusRuleSelector struct {
   labelSelector labels.Selector
   nsLabeler namespacelabeler.Labeler
   rulesInfs *informers.ForResource
   logger log.Logger
   metrics *Metrics
}

func NewPrometheusRuleSelector(...) *PrometheusRuleSelector {
...
}

func (prs *PrometheusRuleSelector) Select(namespaces ...string) (map[string]string, error) {
...
}
```

pkg/operator/rules.go
```diff
@@ -69,3 +87,104 @@ func ValidateRule(promRule monitoringv1.PrometheusRuleSpec) []error {
...
+		}
+		content, err = GenerateRulesConfiguration(promRule.Spec, p.Logger)
+		if err != nil {
+			rejected++
+			level.Warn(p.Logger).Log(
+				"msg", "skipping prometheusrule",
+				"error", err.Error(),
+				"prometheusrule", promRule.Name,
+				"namespace", promRule.Namespace,
+				objectType, p.ResourceName,
```
simonpasquier, line 154: we can pass a wrapped logger for PrometheusRuleSelector (e.g. `log.With(c.logger, "prometheus", p.Name, "namespace", p.Namespace)`).

pkg/operator/rules.go
```diff
@@ -69,3 +87,104 @@ func ValidateRule(promRule monitoringv1.PrometheusRuleSpec) []error {
...
+		"msg", "selected Rules",
+		"rules", strings.Join(ruleNames, ","),
+		"namespace", p.ResourceName,
+		"prometheus", p.ResourceName,
+	)
+
+	if pKey, ok := keyFunc(p.Logger, p); ok {
+		p.Metrics.SetSelectedResources(pKey, monitoringv1.PrometheusRuleKind, len(rules))
+		p.Metrics.SetRejectedResources(pKey, monitoringv1.PrometheusRuleKind, rejected)
+	}
```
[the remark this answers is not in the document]
slashpai: I went ahead with this approach since this was my initial thought as well :)

pkg/operator/rules.go
```diff
@@ -25,9 +25,31 @@ import (
...
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
+	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
+	namespacelabeler "github.com/prometheus-operator/prometheus-operator/pkg/namespace-labeler"
 	thanostypes "github.com/thanos-io/thanos/pkg/store/storepb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type PrometheusRuleSelector struct {
+	objectType   string
```
slashpai, line 36: I see we still need this field for partial response strategy field

slashpai:
I updated with changes suggested, need to remove a few units as well. Updating that as well

pkg/operator/rules.go
```diff
@@ -25,9 +25,31 @@ import (
...
+	return &PrometheusRuleSelector{
+		objectType:   objectType,
+		ruleSelector: ruleSelector,
+		nsLabeler:    nsLabeler,
+		ruleInformer: ruleInformer,
+		logger:       logger,
+	}
+}
+
 func GenerateRulesConfiguration(promRule monitoringv1.PrometheusRuleSpec, logger log.Logger) (string, error) {
```
simonpasquier, line 53: we don't need this to be public anymore?

pkg/operator/rules.go
simonpasquier: We can make it something more obvious though

```suggestion
	ruleFormat   RuleConfigurationFormat
```

with

```golang
type RuleConfigurationFormat int

const (
    PrometheusFormat RuleConfigurationFormat = iota
    ThanosFormat
)
```

simonpasquier (commented):
just a few nits :)

pkg/thanos/rules.go
```diff
@@ -51,9 +51,34 @@ func (o *Operator) createOrUpdateRuleConfigMaps(ctx context.Context, t *monitori
...
+	// append the deprecated PrometheusRulesExcludedFromEnforce
+	for _, rule := range t.Spec.PrometheusRulesExcludedFromEnforce {
+		excludedFromEnforcement = append(excludedFromEnforcement,
+			monitoringv1.ObjectReference{
+				Namespace: rule.RuleNamespace,
+				Group:     monitoring.GroupName,
+				Resource:  monitoringv1.PrometheusRuleName,
+				Name:      rule.RuleName,
+			})
+	}
```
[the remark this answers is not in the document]
slashpai: This would be true for Prometheus and false for Thanos right? 

```
nsLabeler := namespacelabeler.New(
		t.Spec.EnforcedNamespaceLabel,
		append(t.Spec.ExcludedFromEnforcement, monitoringv1.ToObjectReferences(t.Spec.PrometheusRulesExcludedFromEnforce)...)
		false,
	)
```

slashpai:
@simonpasquier Addressed all comments except the nsLabeler (added comment in line)

pkg/operator/rules_test.go
```diff
@@ -62,9 +66,88 @@ func shouldAcceptRuleWithValidPartialResponseStrategyValue(t *testing.T) {
...
 		},
 	}}
-	content, _ := GenerateRulesConfiguration(rules, log.NewLogfmtLogger(log.NewSyncWriter(os.Stdout)))
+	content, _ := generateRulesConfiguration(ThanosFormat, rules, log.NewLogfmtLogger(log.NewSyncWriter(os.Stdout)))
 	if !strings.Contains(content, "partial_response_strategy: warn") {
 		t.Fatalf("expected `partial_response_strategy` to be set in PrometheusRule as `warn`")
 
 	}
 }
+
```
slashpai, line 75: These are tests moved from Prometheus package

pkg/thanos/rules.go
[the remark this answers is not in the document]
simonpasquier: yes correct, sorry for the confusion.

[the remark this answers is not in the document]
slashpai: Are we good with the changes?

JoaoBraveCoding (approved):
Sweet refactoring 👍 Great job JP!

simonpasquier (commented):
lgtm but you'll have to rebase on main :)

slashpai:
> lgtm but you'll have to rebase on main :)

@simonpasquier rebased

simonpasquier:
I realize that we don't have an e2e test verifying that the operator doesn't select invalid rules now. As the validating webhook config watches for all namespaces, an invalid rule can never reach the operator. I'd suggest to tweak the `namespaceSelector` field of the webhook configuration to exclude namespaces with a certain label:

```
namespaceSelector:
  matchExpressions:
  - key: excludeFromWebhook
    operator: NotIn
    values: ["true"]
```

https://github.com/prometheus-operator/prometheus-operator/blob/cb272a61310e7d163d47e8609cbcbad670d1b7df/test/framework/resources/prometheus-operator-validatingwebhook.yaml#L13

Then a rule created in a namespace with the `excludeFromWebhook: "true"` would skip the validation and we can check that the operator discards it.

test/e2e/rules_test.go
```diff
@@ -312,3 +312,24 @@ func TestPrometheusRuleApply(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	name := "invalidRule"
+	invalidAlert := "invalidAlert"
+
+	_, err = framework.MakeAndCreateInvalidRule(context.Background(), ns, name, invalidAlert)
+	if err == nil {
+		t.Fatal("Expected invalid prometheusrule to be rejected")
```
simonpasquier, line 333: Hmm the rule should be accepted by the Kube API server if the namespace is excluded from the webhook?

slashpai:
@simonpasquier  Rebased and updated the e2e test to check prometheus reconciliation status after invalid rule created

test/e2e/prometheus_test.go
```diff
@@ -1592,6 +1592,67 @@ func testInvalidRulesAreRejected(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("expected invalid rule to be created in namespace %v", err)
+	}
+
+	prom := framework.MakeBasicPrometheus(ns, "basic-prometheus", "test-group", 1)
+	if _, err = framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, prom); err != nil {
+		t.Fatal("Creating prometheus failed: ", err)
+	}
+
+	var pollErr error
```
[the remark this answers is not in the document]
slashpai: ya correct. I just needed updated prometheus object after it becomes ready and `f.WaitForPrometheusReady` doesn't return updated object. Should we update that method to return updated prometheus since that will contain status as well but that will need change on other tests as well to update the method call.

For now I am getting the object again in this test to check for reconcile status

slashpai:
Just for more validation I tested this change manually as well checked the e2e test against main branch on which refactoring is not there. Prometheus fails to reconcile without this change. Also the metric `prometheus_operator_managed_resources{controller="prometheus",resource="PrometheusRule",state="rejected"} 1` is updated correctly

slashpai:
@simonpasquier Can you take another look at this one, updated test :)

test/e2e/prometheus_test.go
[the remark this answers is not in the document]
slashpai: my bad I didnt notice `WaitForResourceAvailable()` is called by `WaitForPrometheusReady()` which checks the reconciliation status. I have removed the extra check now

simonpasquier:
Thanks!

PR #5262, opened 2023-01-04
feat: add bearerTokenFile to PodMonitor

raffis:
## Description

ServiceMonitor supports `bearerTokenFile`, I assumed PodMonitor does the same which apparently is not the case.
This PR will add support for bearerTokenFile to PodMonitor.

## Type of change

## Changelog entry

```release-note
Added `spec.podMetricsEndpoints[].bearerTokenFile` in the `PodMonitor` CRD.
```

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -7018,6 +7018,17 @@ PodMetricsEndpointTLSConfig
 </tr>
 <tr>
 <td>
+<code>bearerTokenFile</code><br/>
+<em>
+string
+</em>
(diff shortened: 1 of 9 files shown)


simonpasquier:
It was a conscious decision to avoid `bearerTokenFile` for PodMonitor: from a security standpoint, it would allow anybody with permissions to create PodMonitor resources to read any file in the Prometheus container. 
Instead you need to use `bearerTokenSecret`.

Issue #5190, opened 2022-11-25
webhook's AdmissionResponse should contain the field name(s) in case of failure


simonpasquier:
**What is missing?**

The admission webhook doesn't set the `field` value in the [StatusCause](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#StatusCause) when an error is detected.

https://github.com/prometheus-operator/prometheus-operator/blob/740129b60784d3c23689583cc8871dcb07cc6c68/pkg/admission/admission.go#L156-L159

**Why do we need it?**

This can be used by clients to understand where is the error.

**Environment**

* Prometheus Operator version: N/A

**Anything else we need to know?**:

JoaoBraveCoding:
So IIUC for us to achieve this we would have to modify `toAdmissionResponseFailure` to also have `field` in its parameters, or am I missing something?

```go
func toAdmissionResponseFailure(message, resource, field string, errors []error) *v1.AdmissionResponse {
```

Also if we were to follow this approach, I'm guessing we would only provide a `field` when we are able to, correct? If we get an error during `json.Unmarshal` it's not possible to know what field caused the error.

https://github.com/prometheus-operator/prometheus-operator/blob/1dfecab7c2706f3591d6bb6d5eb8f6b2518f1b45/pkg/admission/admission.go#L225-L228

simonpasquier:
> So IIUC for us to achieve this we would have to modify toAdmissionResponseFailure to also have `field` in its parameters

correct

> Also if we were to follow this approach, I'm guessing we would only provide a field when we are able to, correct?

yes. I think that deserialization failures are only possible if the PrometheusRule CRD and admission webhook's versions are different (and not backward compatible).

JoaoBraveCoding:
Looking further into this I have another question, regarding errors in this line
https://github.com/prometheus-operator/prometheus-operator/blob/21147bd93fa83f01b3fbf1542015d3508962561a/pkg/operator/rules.go#L69
From my POV it would be nice if we could provide the fields `group[i].rule[i]` where the error occurred, now the problem is that `rulefmt.Parse` only returns a list of errors. The errors themselves are rich enough that we could parse them to get this info, but I don't quite like this idea.
On the other hand, I thought we could also enhance the `rulefmt` lib to allow us to iterate over groups and rules and validate each rule individually.

simonpasquier:
[...]

PR #5271, opened 2023-01-06
Adds field to Prometheus Rule AdmissionResponse in case of failure

JoaoBraveCoding:
## Description

Issue: https://github.com/prometheus-operator/prometheus-operator/issues/5190

Problem: When the validation AdmissionWebhook rejects a PrometheusRule
resource, the error message sometimes might not indicate easily in the
resource where is the issue.

Solution: this commit aims at adding a field to the AdmissionResponse
when there is a issue with a PrometheusRule being created, it uses
reflection to build the path to the part of the resource where the issue
was found

## Type of change

## Changelog entry

```release-note
- Adds fields to AdmissionResponse in case of failure
```

### DIFF
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -16,6 +16,7 @@ package operator
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/go-kit/log"
@@ -75,12 +76,12 @@ func generateRulesConfiguration(ruleformat RuleConfigurationFormat, promRule mon
 	if err != nil {
 		return "", errors.Wrap(err, "failed to marshal content")
 	}
-	errs := ValidateRule(promRule)
-	if len(errs) != 0 {
+	rValidationErrors := ValidateRule(promRule)
+	if len(rValidationErrors) != 0 {
 		const m = "Invalid rule"
 		level.Debug(logger).Log("msg", m, "content", content)
-		for _, err := range errs {
-			level.Info(logger).Log("msg", m, "err", err)
+		for _, rValidationError := range rValidationErrors {
+			level.Info(logger).Log("msg", m, "field", rValidationError.Field, "err", rValidationError.Error)
 		}
 		return "", errors.New(m)
 	}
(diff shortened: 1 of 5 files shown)


pkg/operator/types.go
```diff
@@ -52,3 +60,10 @@ func MakeHostAliases(input []monitoringv1.HostAlias) []v1.HostAlias {
 
 	return output
 }
+
+func GetFieldJSONTag(parentField interface{}, fieldName string) string {
```
simonpasquier, line 64: parentField is a bit imprecise? It could also live in pkg/operator/rules.go if it's only used there.

```suggestion
func GetJSONTagForField(i interface{}, fieldName string) string {
```

pkg/operator/types.go
```diff
@@ -15,11 +15,19 @@
...
 import (
+	"reflect"
+	"strings"
+
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type AdmissionError struct {
```
simonpasquier, line 26:
```suggestion
type RuleValidationError struct {
```

simonpasquier (commented):
Looks good!

pkg/operator/rules.go
```diff
@@ -33,39 +33,60 @@ func GenerateRulesConfiguration(promRule monitoringv1.PrometheusRuleSpec, logger
...
-		return []error{errors.Wrap(err, "failed to marshal content")}
+		return []AdmissionError{{Error: errors.Wrap(err, "failed to marshal content")}}
 	}
+	admissionErrors := []AdmissionError{}
 	_, errs := rulefmt.Parse(content)
-	return errs
+
+	for _, err := range errs {
+		switch promErr := err.(type) {
+		case *rulefmt.Error:
```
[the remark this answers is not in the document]
JoaoBraveCoding: Indeed I forgot that part, maybe it's best to add a default
```golang
		default:
			admissionError := AdmissionError{
				Field: groupTag,
				Error: err,
			}
			admissionErrors = append(admissionErrors, admissionError)
		}
```
Since if it's not a `rulefmt.Error` we will not be able to tell anything else from the field apart from that it happened somewhere at the`groups` level.

JoaoBraveCoding:
Looks like the failed test is not related

JoaoBraveCoding:
@simonpasquier should be ready to merge

slashpai:
@JoaoBraveCoding Can you rebase? :)

PR #5290, opened 2023-01-20
Split CRDs into their own files

xiu:
## Description

This is purely cosmetic, I had noticed @paulfantom starting this back in 2021 in #4222 since I was wondering the same as I went through the codebase. Location might not be the greatest, I tried to keep the common resources or resources that could be reused in `types.go`

## Type of change

## Changelog entry

```release-note

```


simonpasquier (approved):
:+1:

PR #5291, opened 2023-01-20
Convert "discovery.k8s.io" to "discovery.k8s.io/v1"

jahnvigupta:
Fixes endpointslice support

## Description
As endpoint controller in the kubernetes version >1.20 supports only 1000 endpoints for a daemonset, the shift to endpointslice is required to not loose data. Changelogs claim that version 0.55 support endpointslices but that is not the case. The issue has been raised in the past as well.
https://github.com/prometheus-operator/prometheus-operator/issues/3862#issuecomment-1068260430
https://github.com/prometheus-operator/prometheus-operator/issues/3862#issuecomment-1259853713

This change fixes the incorrectly used discovery group version causing the issue.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -327,7 +327,7 @@ func New(ctx context.Context, conf operator.Config, logger log.Logger, r prometh
 		c.nsPromInf = newNamespaceInformer(c, c.config.Namespaces.PrometheusAllowList)
 	}
 
-	endpointSliceSupported, err := k8sutil.IsAPIGroupVersionResourceSupported(c.kclient.Discovery(), "discovery.k8s.io", "endpointslices")
+	endpointSliceSupported, err := k8sutil.IsAPIGroupVersionResourceSupported(c.kclient.Discovery(), "discovery.k8s.io/v1", "endpointslices")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


simonpasquier:
I'm afraid that it isn't going to work as-is, especially because of RBAC permissions which might be missing for the Prometheus service account. https://github.com/prometheus-operator/prometheus-operator/issues/3862#issuecomment-1068260430 lists the possible options and how we want to tackle this. TL;DR: we suggest to add a boolean `spec.serviceDiscoveryRole` field to the Prometheus CRD which would configure `role: endpointslices` when true.

Also @matej-g expressed interest in working on this feature during our last prometheus operator office hours.

iyashu:
@simonpasquier does that means existing feature of using endpointslices for service discovery is not useful since it'll not be able to discover endpointslices with current implementation ever (given we're validating existence of resource under `discovery.k8s.io` rather than `discovery.k8s.io/v1`)?

matej-g:
Just wanted to add @iyashu feel free to take this, this issue unfortunately keeps getting lower and lower on my to-do list.

jahnvigupta:
Sure, thanks @matej-g. We are working on it and will be able to raise a PR soon.

PR #5293, opened 2023-01-20
feat: Add spec.AdditionalArgs to ThanosRuler

### LINKED ISSUE #5260: Add spec.additionalArgs to ThanosRuler CRD
**What is missing?**

A way to pass additional arguments to the ThanosRuler CRD as we have for Prometheus CRD

**Why do we need it?**

Different environments have different needs, we should allow prom-operator users to be able to run ThanosRuler with arguments other than the ones set by the operator

**Environment**

xiu:
Fixes #5260

## Description

Adds `spec.AdditionalArgs` to ThanosRuler

## Type of change

## Changelog entry

```release-note
Add `spec.AdditionalArgs` to ThanosRuler
```


xiu:
I'm not sure why `TestOperatorUpgrade` is failing, I've successfully executed it in minikube. Let me know if you have any more insights into it.

pkg/thanos/statefulset.go
```diff
@@ -530,3 +536,62 @@ func mountSecret(secretSelector *v1.SecretKeySelector, volumeName string, trVolu
...
+		if arg.Value != "" {
+			containerArgs = append(containerArgs, fmt.Sprintf("--%s=%s", arg.Name, arg.Value))
+		} else {
+			containerArgs = append(containerArgs, fmt.Sprintf("--%s", arg.Name))
+
+		}
+	}
+
+	return containerArgs, nil
+}
```
xiu, line 597: This is a copy of the same functions from `pkg/prometheus/statefulset.go`. I'd rather have them elsewhere than copy them around, suggestions welcome :)

slashpai: I think `pkg/operator` is better place for common functions and methods

xiu: Thanks @slashpai! Moved it there 👍

pkg/thanos/statefulset.go
```diff
@@ -220,78 +220,84 @@ func makeStatefulSetSpec(tr *monitoringv1.ThanosRuler, config Config, ruleConfig
...
 	if tr.Spec.AlertQueryURL != "" {
-		trCLIArgs = append(trCLIArgs, fmt.Sprintf("--alert.query-url=%s", tr.Spec.AlertQueryURL))
+		trCLIArgs = append(trCLIArgs, monitoringv1.Argument{Name: "alert.query-url", Value: tr.Spec.AlertQueryURL})
+	}
+
+	containerArgs, err := operator.BuildArgs(trCLIArgs, tr.Spec.AdditionalArgs)
+	if err != nil {
+		return nil, err
 	}
+	containerArgs = append([]string{"rule"}, containerArgs...)
```
JoaoBraveCoding, line 300: nit: Maybe adding a small comment here would be nice for people not familiar with ThanosRuler config that has to start with `rule`

JoaoBraveCoding (approved):
Small nit but apart from that LGTM

pkg/thanos/statefulset.go
xiu: Good point, added it 👍

simonpasquier (approved):
nice!

PR #5296, opened 2023-01-22
fix: Unsupported types in log lines

### LINKED ISSUE #5253: misleading log for alertmanagerConfig
**What happened?**
I was set the alertmanagerconfig CR for the k8s cluster, and misleading by this log.
https://github.com/prometheus-operator/prometheus-operator/blob/bbd7400711a99eda6c140c31e17f770355617b66/pkg/alertmanager/amcfg.go#L1683-L1686
1. The r.match wouldn't be empty if "matchType" not set in alertmanagerconfig CR like this.
```
matchers:
  - name: key1
    value: value1
```
2. The log for r.Match and r.MatchRE is not string type, will lead to "unsupported value type"
The log will be like this
```
level=warn ts=2022-12-27T06:28:50.879469406Z caller=amcfg.go:1685 component=alertmanageroperator alertmanager=prometheus-kube-prometheus-alertmanager namespace=prometheus receiver=prometheus/liuer-alert-webhook-test/webhook msg="'matchers' field is using a deprecated syntax which will be removed in future versions" match="unsupported value type" match_re="unsupported value type"
```

v0.61.1

xiu:
Fixes #5253

## Description

Fixes unsupported types reported in #5253 

## Type of change

## Changelog entry

```release-note
Fix unsupported types in alertmanager route sanitizer log lines
```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -1918,18 +1918,18 @@ func (r *route) sanitize(amVersion semver.Version, logger log.Logger) error {
 
 	if matchersV2Allowed && checkNotEmptyMap(r.Match, r.MatchRE) {
 		msg := "'matchers' field is using a deprecated syntax which will be removed in future versions"
-		level.Warn(withLogger).Log("msg", msg, "match", r.Match, "match_re", r.MatchRE)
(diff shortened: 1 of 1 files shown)


JoaoBraveCoding (approved):
Not sure if there's a way to test these, but apart from that lgtm

PhilipGough (approved):

Issue #5295, opened 2023-01-22
webexConfig for Alertmanager is missing


eumel8:
Just wondering, the [new Webex receiver](https://github.com/prometheus/alertmanager/releases/tag/v0.25.0) was implemented in https://github.com/prometheus-operator/prometheus-operator/pull/5254, thx to @simonpasquier , but maybe I have overseen, there is no implementation in alertmanagerconfigs.monitoring.coreos.com/v1alpha1 or alertmanagerconfigs.monitoring.coreos.com/v1beta1

Is there a plan or just no time? I'm curious to have this feature :-)

xiu:
According to https://github.com/prometheus-operator/prometheus-operator/pull/5254#issuecomment-1370613809, the configuration generation part was done but the CRD still needs to be updated to support it.

This issue is similar to #5251.

eumel8:
help wanted ... I would do my best
@xiu is it allowed to change existing CRDs like v1alpha1? thx

xiu:
I'm not a maintainer but I would expect you can, @simonpasquier can you confirm?

slashpai:
All sanitise part is taken care in https://github.com/prometheus-operator/prometheus-operator/pull/5254. This is not yet supported in CRD.

xiu:
> help wanted ... I would do my best
@eumel8 are you taking this one on?

eumel8:
@xiu ^^

fctb:
any time when webex here working?

eumel8:
available in https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.68.0

a working example can be catched up[ here](https://github.com/caas-team/use-cases/tree/main#alertmanagerconfig-for-webex-teams)
the used markdown template is described [here](https://github.com/caas-team/caas-project-monitoring/blob/feat/51.0.3/values.yaml#L145-L243)

thanks to @puffitos to get this feature finaly running

PR #5254, opened 2022-12-28
pkg/alertmanager: sanitize configuration for v0.25.0

simonpasquier:
## Description

Alertmanager v0.25.0 introduces new notifiers (Discord, Webex) + configuration fields that the operator needs to sanitize.
    
Global configuration:
    * `smtp_auth_password_file`
    * `victorops_api_key_file`
    
HTTP client configuration:
    * `proxy_url` for OAuth2
    * `min_version` & `max_version` for TLS
    * `enable_http2`
    
Email configuration:
    * `auth_password_file`
    
PagerDuty configuration:
    * `routing_key_file`
    * `service_key_file`
    * `source`
    
VictorOps configuration:
    * `api_key_file`

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [X] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Supported new fields from Alertmanager v0.25.0
```

PR #4726, opened 2022-04-18
Add Telegram Receiver type to AlertmanagerConfig CRD

angelbarrera92:
## Description

Add Telegram Receiver type to AlertmanagerConfig CRD
Fixes #4703

There are a few things to consider while using this new feature:
- API URL is mandatory in the alertmanager go project, but docs say the opposite: https://github.com/prometheus/alertmanager/issues/2876
- The default parse mode, markdownv2, is not fully compatible yet, meanwhile consider using HTML: https://github.com/prometheus/alertmanager/issues/2866

## Type of change

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [X] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

Add Telegram Receiver type to AlertmanagerConfig CRD

```release-note
- Add Telegram Receiver type to AlertmanagerConfig CRD
```

PR #5305, opened 2023-01-26
Webex Feature/API

eumel8:
Signed-off-by: Frank Kloeker <f.kloeker@telekom.de>

## Description

Prometheus Alertmanager brings the functionality of Webex Notification. Start of Operator implementation was [here](https://github.com/prometheus-operator/prometheus-operator/pull/5254)
This is the part of API/CRD extension

## Type of change

ref:https://github.com/prometheus-operator/prometheus-operator/issues/5295

## Changelog entry

```release-note
Add Webex configs to AlertmanagerConfig CRD
```


eumel8:
the bundle version needs to bump up

xiu:
Great!

@eumel8 you can generate the bundle by using `make generate`. To help with formatting as well, `make format` and to check the formatting of the docs, `make check-docs` (https://github.com/prometheus-operator/prometheus-operator#contributing).

I see your tests are not passing, `make test-unit` can help there (https://github.com/prometheus-operator/prometheus-operator#testing).

simonpasquier:
Thanks for starting this! You'd need to update https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1beta1/conversion_from.go and https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1beta1/conversion_to.go to convert between v1alpha1 and v1beta1 types.

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -571,6 +573,16 @@ type HTTPConfig struct {
 	FollowRedirects *bool `json:"followRedirects,omitempty"`
 }
 
+// WebexConfig configures notification via Cisco Webex
+// See https://prometheus.io/docs/alerting/latest/configuration/#webex_config
+type WebexConfig struct {
+	SendResolved *bool       `json:"sendResolved,omitempty"`
+	APIURL       string      `json:"apiURL,omitempty"`
+	RoomID       string      `json:"roomID,omitempty"`
```
simonpasquier, line 581: room_id is a mandatory field.

```suggestion
	// ID of the Webex Teams room where to send the messages.
	// +kubebuilder:validation:MinLength=1
	RoomID       string      `json:"roomID,omitempty"`
```

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -571,6 +573,16 @@ type HTTPConfig struct {
 	FollowRedirects *bool `json:"followRedirects,omitempty"`
 }
 
+// WebexConfig configures notification via Cisco Webex
+// See https://prometheus.io/docs/alerting/latest/configuration/#webex_config
+type WebexConfig struct {
+	SendResolved *bool       `json:"sendResolved,omitempty"`
```
simonpasquier, line 579:
```suggestion
	// Whether to notify about resolved alerts.
	SendResolved *bool       `json:"sendResolved,omitempty"`
```

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -571,6 +573,16 @@ type HTTPConfig struct {
...
+	APIURL       string      `json:"apiURL,omitempty"`
```
simonpasquier, line 580:
```suggestion
	// The Webex Teams API URL if different from the default API URL.
	APIURL       string      `json:"apiURL,omitempty"`
```

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -571,6 +573,16 @@ type HTTPConfig struct {
...
+	RoomID       string      `json:"roomID,omitempty"`
+	Message      string      `json:"message,omitempty"`
+	HTTPConfig   *HTTPConfig `json:"httpConfig,omitempty"`
```
simonpasquier, line 583:
```suggestion
	// HTTP client configuration.
	HTTPConfig   *HTTPConfig `json:"httpConfig,omitempty"`
```

eumel8:
thx for starting review, @xiu @simonpasquier

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -571,6 +573,16 @@ type HTTPConfig struct {
 	FollowRedirects *bool `json:"followRedirects,omitempty"`
 }
 
+// WebexConfig configures notification via Cisco Webex
+// See https://prometheus.io/docs/alerting/latest/configuration/#webex_config
+type WebexConfig struct {
+	APIURL       string      `json:"apiURL,omitempty"`
+	HTTPConfig   *HTTPConfig `json:"httpConfig,omitempty"`
+	Message      string      `json:"message,omitempty"`
+	RoomID       string      `json:"roomID"`
```
simonpasquier, line 582:
```suggestion
	// +kubebuilder:validation:MinLength=1
	RoomID       string      `json:"roomID"`
```

pkg/apis/monitoring/v1beta1/alertmanager_config_types.go
```diff
@@ -565,6 +567,16 @@ type HTTPConfig struct {
 	FollowRedirects *bool `json:"followRedirects,omitempty"`
 }
 
+// WebexConfig configures notification via Cisco Webex
+// See https://prometheus.io/docs/alerting/latest/configuration/#webex_config
+type WebexConfig struct {
+	SendResolved *bool       `json:"sendResolved,omitempty"`
+	APIURL       string      `json:"apiURL,omitempty"`
+	RoomID       string      `json:"roomID"`
```
simonpasquier, line 575:
```suggestion
	// +kubebuilder:validation:MinLength=1
	RoomID       string      `json:"roomID"`
```

simonpasquier (changes requested):
Can you add comments for each field? Next step is to wire the webex integration in https://github.com/prometheus-operator/prometheus-operator/blob/6be89b4010d06cbd2561b6650b6399438900a944/pkg/alertmanager/amcfg.go#L464

eumel8:
just wondering, `make generate` fails on Ubuntu:

```
mv github.com/prometheus-operator/prometheus-operator/pkg/client/{versioned,informers,listers} pkg/client
mv: cannot stat 'github.com/prometheus-operator/prometheus-operator/pkg/client/informers': No such file or directory
mv: cannot stat 'github.com/prometheus-operator/prometheus-operator/pkg/client/listers': No such file or directory
make: *** [Makefile:142: k8s-client-gen] Error 1
```

xiu:
> just wondering, `make generate` fails on Ubuntu:
> 
> ```
> mv github.com/prometheus-operator/prometheus-operator/pkg/client/{versioned,informers,listers} pkg/client
> mv: cannot stat 'github.com/prometheus-operator/prometheus-operator/pkg/client/informers': No such file or directory
> mv: cannot stat 'github.com/prometheus-operator/prometheus-operator/pkg/client/listers': No such file or directory
> make: *** [Makefile:142: k8s-client-gen] Error 1
> ```

This is not the issue as I see it. Check https://github.com/prometheus-operator/prometheus-operator/actions/runs/4088515306/jobs/7050273000, the step is checking whether all generated content has been committed. It thus runs `make generate` and does a global diff. In this case, it found changes.

Can you run `make generate` and commit the changes to your branch and try again?

eumel8:
@xiu This was the error on my local Ubuntu WSL. I couldn't found out the cause and made make generate on another machine. Maybe there are some hints for build requirements beside make and Go.

xiu:
> @xiu This was the error on my local Ubuntu WSL. I couldn't found out the cause and made make generate on another machine. Maybe there are some hints for build requirements beside make and Go.

Oh, I don't have a Windows machine to try that out unfortunately. Glad you made it work!

eumel8:
@simonpasquier I'm not sure, anything else to change here?

simonpasquier:
> @simonpasquier I'm not sure, anything else to change here?

You need to update the [convertReceiver()](https://github.com/prometheus-operator/prometheus-operator/blob/6be89b4010d06cbd2561b6650b6399438900a944/pkg/alertmanager/amcfg.go#L464) method to convert the Webex configuration from the CRD to the native Alertmanager configuration.

You can take a look at https://github.com/prometheus-operator/prometheus-operator/pull/4726 which did the same thing but for the Telegram integration. Does it make sense?

fctb:
any time when it will get unblocked?

eumel8:
@fctb It's progressing slowly. I'm a bit overwhelmed with the feature. Any help is appreciated, hopefully it will be over soon.

puffitos:
Hi @simonpasquier, I'll be picking up from where @eumel8 left off. Please let me know if there's still something to be done. I've gone through the telegram MR and it seems to me we've covered everything. The changes you'd requested are also already implemented. I've just pulled the latest commits from the upstream and fixed a logic bug in the validation. And also a small grammar mistake I found :)

The only thing I've noticed is that the E2E tests don't successfully run on my ubuntu pc. The webex ones do, but some fail, unfortunately.

pkg/alertmanager/validation/v1alpha1/validation.go
```diff
@@ -300,6 +304,25 @@ func validateTelegramConfigs(configs []monitoringv1alpha1.TelegramConfig) error
...
+	for _, config := range configs {
+		if config.APIURL == "" {
+			if _, err := validation.ValidateURL(config.APIURL); err != nil {
+				return errors.Wrap(err, "invalid 'apiURL'")
+			}
+		}
+
+		if config.RoomID == "" {
+			return fmt.Errorf("mandatory field %q is empty", "roomID")
+		}
```
ArthurSens, line 317: The API definition already says that RoomID is a required field, right? Do we need this validation anyway?

ArthurSens (commented):
Thanks @eumel8 for starting and @puffitos for picking it up, I think it's almost done :)

pkg/apis/monitoring/v1beta1/alertmanager_config_types.go
```diff
@@ -565,6 +567,26 @@ type HTTPConfig struct {
...
+	HTTPConfig *HTTPConfig `json:"httpConfig,omitempty"`
+	// Message template
+	// +optional
+	Message string `json:"message,omitempty"`
+	// ID of the Webex Teams room where to send the messages.
+	RoomID string `json:"roomID"`
+	// Whether to notify about resolved alerts.
+	// +optional
+	SendResolved *bool `json:"sendResolved,omitempty"`
+}
```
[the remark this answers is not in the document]
puffitos: I've changed the optional fields to pointers and accommodated the `convertWebexConfig` function to work with the new definition.

pkg/alertmanager/validation/v1alpha1/validation.go
puffitos: Good point; i've removed the unnecessary check :+1:

pkg/apis/monitoring/v1beta1/alertmanager_config_types.go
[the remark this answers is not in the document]
puffitos: I've also changed the internal `webexConfig` struct in the `alertmanager` package to use pointers. 

E2E tests are still failing. The YAML config can't be parsed for the following reason:

```
ts=2023-06-17T06:10:30.122Z caller=coordinator.go:118 level=error component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/config_out/alertmanager.env.yaml err="missing webex_configs.http_config.authorization"
```

Why is this happening? The e2e tests for webex were written in a similar fashion to the telegram ones, and the e2e were passing before moving to pointers.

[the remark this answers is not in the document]
puffitos: OK... I needed to take a look on how the global config for the tests is being created, then everything was clear. I had to make some changes:

I had to change the HTTPConfig field from optional to mandatory, as the Bearer Token of the Bot is expected to be there. The AlertManager code always checks for its presence:

https://github.com/prometheus/alertmanager/blob/90a8acd27661db09c9d9f1b4a2d4574a24466c76/config/notifiers.go#L196-L213
```go
// UnmarshalYAML implements the yaml.Unmarshaler interface.
func (c *WebexConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
	*c = DefaultWebexConfig
	type plain WebexConfig
	if err := unmarshal((*plain)(c)); err != nil {
		return err
	}

	if c.RoomID == "" {
		return fmt.Errorf("missing room_id on webex_config")
	}

	if c.HTTPConfig == nil || c.HTTPConfig.Authorization == nil {
		return fmt.Errorf("missing webex_configs.http_config.authorization")
	}

	return nil
}
``` 

I decided against expecting from the user to put the Authorization for the webex bot into their global config and made the http_config a mandatory field.

After that, the tests were still failing, because of missing smpt fields in the global AlertManager config. I've added an smpt_smarthost and a smpt_from, which are also mandatory fields for a global config (the config couldn't be unmarshalled for else wise).

Finally, I noticed that the `testPromOperatorStartsWithoutScrapeConfigCRD` was still failing. The test was checking whether the returned map with pod restarts of the prometheus operator pod was empty:

```goland
	require.Emptyf(t, restarts, "expected to get 1 container but got %d", len(restarts))
// this produced the funny looking error: expected to get 1 container but got 1
```

I've changed this to check whether only 1 container was really returned.

Sorry the long comment; just wanted to make sure I've explained all the changes.

ArthurSens:
Nice work @puffitos! I want to apologize, I'm a bit busy this week and I'm struggling to find time to give a proper review

You found a tricky situation there with the HTTP Config! As you noticed, HTTP Config can be optional in WebexConfig if we follow Alertmanager's API definition because it can assume the value of Global HTTP Config. 

You give an interesting proposal here by making it required, it would certainly avoid that we have a nil HTTP config in case both Webex and Global HTTP config are nil, but we're breaking the case where the user really wants to re-use the HTTP config from Global config.

I think we'll need to keep HTTP Config as an optional field but validate it when reconciling the CRD.

Any other ideas @prometheus-operator/prometheus-operator-reviewers ?

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -571,6 +573,25 @@ type HTTPConfig struct {
 	FollowRedirects *bool `json:"followRedirects,omitempty"`
 }
 
+// WebexConfig configures notification via Cisco Webex
+// See https://prometheus.io/docs/alerting/latest/configuration/#webex_config
+type WebexConfig struct {
+	// The Webex Teams API URL i.e. https://webexapis.com/v1/messages
+	// +optional
+	APIURL *string `json:"apiURL,omitempty"`
```
simonpasquier, line 581: Let's create a URL type for proper validation
```suggestion
	APIURL *URL `json:"apiURL,omitempty"`
```

then

````
// +kubebuilder:validation:Pattern=`^https?://.+$`
type URL string
````

File: pkg/alertmanager/types.go [hunk 10]
```diff
@@ -354,8 +355,8 @@ type discordConfig struct {
 type webexConfig struct {
 	VSendResolved *bool             `yaml:"send_resolved,omitempty"`
 	HTTPConfig    *httpClientConfig `yaml:"http_config,omitempty"`
-	APIURL        string            `yaml:"api_url,omitempty"`
-	Message       string            `yaml:"message,omitempty"`
+	APIURL        *string           `yaml:"api_url,omitempty"`
+	Message       *string           `yaml:"message,omitempty"`
```
simonpasquier, line 359: I wouldn't change the types here as they come from the Alertmanager upstream configuration format.

pkg/alertmanager/operator.go
```diff
@@ -1397,6 +1402,28 @@ func checkWechatConfigs(
...
 
+func checkWebexConfigs(
+	ctx context.Context,
+	configs []monitoringv1alpha1.WebexConfig,
+	namespace string,
+	key string,
+	store *assets.Store,
+	amVersion semver.Version,
+) error {
+	for i, config := range configs {
```
simonpasquier, line 1413: we need a check on the amVersion too (e.g. return an error if version < v0.25.0 similar to the check in `checkTelegramConfigs()`).

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
puffitos: Should I also replace other URL-kind fields as with the new URL type (maybe in a separate PR)?

test/e2e/scrapeconfig_test.go
```diff
@@ -216,7 +216,7 @@ func testPromOperatorStartsWithoutScrapeConfigCRD(t *testing.T) {
 
 	restarts, err := framework.GetPodRestartCount(context.Background(), ns, pl.Items[0].GetName())
 	require.NoError(t, err)
-	require.Emptyf(t, restarts, "expected to get 1 container but got %d", len(restarts))
+	require.Lenf(t, restarts, 1, "expected to get 1 container but got %d", len(restarts))
```
ArthurSens, line 219: Actually we're counting the amount of restarts here, not the amount of containers. The error message is certainly misleading!

That said, the test is correct. We don't expect restarts here

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
simonpasquier: > Does a dedicated URL type offer any other advantages, which I'm not aware of?

yes, `kubebuilder:validation:Pattern` is applied by the Kubernetes API without requiring any validating webhook configuration.

simonpasquier:
@eumel8 I've added a few commits on top of your PR to address my comments.

test/e2e/scrapeconfig_test.go
```diff
@@ -216,7 +216,7 @@ func testPromOperatorStartsWithoutScrapeConfigCRD(t *testing.T) {
 
 	restarts, err := framework.GetPodRestartCount(context.Background(), ns, pl.Items[0].GetName())
 	require.NoError(t, err)
-	require.Lenf(t, restarts, 1, "expected to get 1 container but got %d", len(restarts))
+	require.Emptyf(t, restarts, "expected to get 1 container but got %d", len(restarts))
```
puffitos: This test was failing on my machine with the message "expected to get 1 container but got 1". On line 213 one Pod is explicitly expected.  The GetPodRestartCount will therefore always return a map with at least one container, which should have 0 restarts. The `restarts` map has therefore a length of at least 1. 

Emptyf checks whether the map is empty, which it shouldn't. The error msg is even more cryptic, as it refers to the length of the map, which should be one according to the message. But here we're checking whether the map's nil / empty? 

I can't get my head around as to how this tests works :smile: I just wanted to explain, why I made this change and I why I didn't consider it unnecessary.

test/framework/resources/alertmanager-main-secret.yaml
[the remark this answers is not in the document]
puffitos: I've reverted the change - if the E2E suite passes, it is indeed not necessary. I'd only changed it, as it didn't work when I ran the E2E tests locally and after debugging I'd found out the issue were the missing fields in the AM Config :shrug:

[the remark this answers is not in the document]
puffitos: Tests passes so thanks for reviewing this change as well.

ArthurSens (approved):

pkg/alertmanager/operator.go
```diff
@@ -1433,6 +1438,36 @@ func checkWechatConfigs(
...
+	key string,
+	store *assets.Store,
+	amVersion semver.Version,
+) error {
+	if len(configs) == 0 {
+		return nil
+	}
+
+	if amVersion.LT(semver.MustParse("0.25.0")) {
+		return fmt.Errorf(`webexConfigs' is available in Alertmanager >= 0.24.0 only - current %s`, amVersion)
```
simonpasquier, line 1454: (nit)
```suggestion
		return fmt.Errorf(`webexConfigs' is available in Alertmanager >= 0.25.0 only - current %s`, amVersion)
```

simonpasquier (commented):
very good! One last nit...

ArthurSens (approved):
Oof, I approved and forgot to merge this one after 3 weeks... sorry! 😅 

Approving again and I'll let simon take another look if he wants to :)

I'm putting a reminder to merge it this week if there is no more change requests

Issue #5289, opened 2023-01-20
Documentation about volume expansion may trigger a recreation of the statefulset


simonpasquier:
> https://prometheus-operator.dev/docs/operator/storage/#resizing-volumes.
> First, update the storage request in the spec.storage field of the custom resource (assuming a Prometheus resource named example):

I tested it according to this document, and it seems that the statefulset is recrated in step 1.

> The operator should recreate the StatefulSet immediately, there will be no service disruption thanks to the orphan strategy and the volumes mounted in the Pods should have the updated size.

Doesn't this mean there's no restart of pods and no downtimes?

In order to do this without service disruption, it seems that we need to make replicas of the prometheus operator to zero because of blocking the recreation of the statefulset.

_Originally posted by @parkjeongryul in https://github.com/prometheus-operator/prometheus-operator/issues/4079#issuecomment-1396465595_

simonpasquier:
Following comment from @parkjeongryul

```
Sure. Belows are our storage class and it supports volume expansion.

$ k get  storageclass rbd-ssd -o yaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    meta.helm.sh/release-name: rbd
    meta.helm.sh/release-namespace: kube-system
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2022-12-05T04:16:48Z"
  labels:
    app: rbd
    app.kubernetes.io/managed-by: Helm
    chart: rbd-v1
    heritage: Helm
    release: rbd
  name: rbd-ssd
  resourceVersion: "60047"
  uid: 224e347d-2908-4101-8e8f-00aaf44328ac
parameters:
  cephLogDir: /var/log/ceph
  clusterID: 96f95cff-c001-4a8f-87f0-fe5494b13c13
  csi.storage.k8s.io/controller-expand-secret-name: nv-rbd-ssd
  csi.storage.k8s.io/controller-expand-secret-namespace: kube-system
  csi.storage.k8s.io/fstype: xfs
  csi.storage.k8s.io/node-stage-secret-name: nv-rbd-ssd
  csi.storage.k8s.io/node-stage-secret-namespace: kube-system
  csi.storage.k8s.io/provisioner-secret-name: nv-rbd-ssd
  csi.storage.k8s.io/provisioner-secret-namespace: kube-system
  imageFeatures: layering
[...]

PR #5306, opened 2023-01-26
Fix documentation about volume expansion may trigger a recreation of the statefulset

parkjeongryul:
## Description

Close: https://github.com/prometheus-operator/prometheus-operator/issues/5289

Current documentation about volume expansion may trigger a recreation of the statefulset.
Documentation should provide instructions that incur zero downtime.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/Documentation/user-guides/storage.md
+++ b/Documentation/user-guides/storage.md
@@ -165,8 +165,9 @@ issue](https://github.com/kubernetes/enhancements/issues/661)).
 
 It is still possible to fix the situation manually.
 
-First, update the storage request in the `spec.storage` field of the custom
-resource (assuming a Prometheus resource named `example`):
+First, update `spec.paused` field to `true`.
+It should prevent the operator from recreating the statefulset.
+(assuming a Prometheus resource named `example`):
 
 ```yaml
 apiVersion: monitoring.coreos.com/v1
@@ -175,13 +176,12 @@ metadata:
   name: example
 spec:
   replicas: 1
-  storage:
-    volumeClaimTemplate:
-      spec:
-        resources:
-          requests:
-            storage: 10Gi
+  paused: true
 ```
+
+Directly modifying the storage request in the Prometheus CRD would
+recreate the StatefulSet with the new spec and, unfortunately, all the pods
+managed by the StatefulSet would be recreated at the same time incurring downtime.
 
 Next, patch every PVC with the updated storage request (10Gi in this example):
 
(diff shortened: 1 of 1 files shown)


Documentation/user-guides/storage.md
```diff
@@ -175,14 +176,13 @@ metadata:
   name: example
 spec:
   replicas: 1
-  storage:
-    volumeClaimTemplate:
-      spec:
-        resources:
-          requests:
-            storage: 10Gi
+  paused: true
```
simonpasquier, line 179: What if we keep the existing manifest but adding `paused: true`?

```suggestion
  paused: true
  storage:
      volumeClaimTemplate:
      spec:
        resources:
          requests:
            storage: 10Gi
```

parkjeongryul: I think it's good. However, it seems strange that it is set to 10 Gi in the first step. Shouldn't we use a smaller value  that 10Gi in the first step because we're going to increase it to 10 Gi in the next step? (because this is guide to increase pvc size)

So, how about specifying initial storage value less than 10Gi? (like 5Gi)
```suggestion
  paused: true
    storage:
      volumeClaimTemplate:
      spec:
        resources:
          requests:
            storage: 5Gi
```

JoaoBraveCoding: I might be wrong but I'm understanding correctly @simonpasquier is proposing in the first step adding the pause and updating the size of the volume all in 1 go. 

Then in the last step, we would just remove the pause and we could omit the `storage` section.

JoaoBraveCoding (commented):
Apart from the discussion going on lgtm 👍

Documentation/user-guides/storage.md
parkjeongryul: > I might be wrong but I'm understanding correctly @simonpasquier is proposing in the first step adding the pause and updating the size of the volume all in 1 go.
Then in the last step, we would just remove the pause and we could omit the storage section.

I understand now. I think the procedure you suggested will work well also. I'll apply your feedback.

JoaoBraveCoding (approved):
lgtm 👍

Documentation/user-guides/storage.md
simonpasquier: What isn't very explicit in the current documentation is that we assume the initial value is less than 10Gi. Maybe using a patch command would be clearer?

```
kubectl patch prometheus/example --patch '{"spec": {"paused": true, "storage": {"volumeClaimTemplate": {"spec": {"resources": {"requests": {"storage":"10Gi"}}}}}}}' --type merge
```

parkjeongryul: It's okay to change it to a patch command, but I think it became clear now without changing the yaml file to a patch command. 
  
What I misunderstood at first was that I wrote a guide to change the pvc size in the last step, but I didn't understand that your suggestion meant to change it in the first step.
  
Now, I think it's pretty clear because I modified it to change the pvc size in the first stage.
and mentioned like below at first step:
> And update the storage request in the `spec.storage` field of the custom

parkjeongryul: I think it's now clear, but if you think it's better to change it to patch command, please let me know.

Documentation/user-guides/storage.md
```diff
@@ -191,12 +195,24 @@ for p in $(kubectl get pvc -l operator.prometheus.io/name=example -o jsonpath='{
...
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: example
+spec:
+  replicas: 1
+  paused: false
+```
```
simonpasquier, line 214:
```suggestion
```bash
kubectl patch prometheus/example --patch '{"spec": {"paused": false}}' --type merge
```
```

parkjeongryul: I applied your suggestion 👍

simonpasquier (approved):
thanks!

PR #5310, opened 2023-01-30
Generate Prometheus configuration through interface

ArthurSens:
## Description

This PR creates a common interface that can be used by `Prometheus` and, in the future, by `PrometheusAgent` CRDs. I'm also refactoring the config generation to use the new interface instead of the Prometheus object, as discussed in https://github.com/prometheus-operator/prometheus-operator/issues/3989#issuecomment-1408948890.

This is probably one of many PRs needed for us to finally introduce the PrometheusAgent CRD

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -8228,11 +8228,6 @@ string
 </tr>
 </tbody>
 </table>
-<h3 id="monitoring.coreos.com/v1.PrometheusInterface">PrometheusInterface
-</h3>
-<div>
-<p>PrometheusInterface is used by Prometheus and PrometheusAgent to share common methods, e.g. config generation.</p>
-</div>
 <h3 id="monitoring.coreos.com/v1.PrometheusRuleExcludeConfig">PrometheusRuleExcludeConfig
 </h3>
 <p>
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -8228,6 +8228,11 @@ string
 </tr>
 </tbody>
 </table>
+<h3 id="monitoring.coreos.com/v1.PrometheusInterface">PrometheusInterface
+</h3>
+<div>
+<p>PrometheusInterface is used by Prometheus and PrometheusAgent to share common methods, e.g. config generation.</p>
+</div>
 <h3 id="monitoring.coreos.com/v1.PrometheusRuleExcludeConfig">PrometheusRuleExcludeConfig
 </h3>
 <p>
--- a/pkg/apis/monitoring/v1/prometheus_types.go
+++ b/pkg/apis/monitoring/v1/prometheus_types.go
@@ -28,6 +28,17 @@ const (
 	PrometheusKindKey = "prometheus"
 )
 
+// PrometheusInterface is used by Prometheus and PrometheusAgent to share common methods, e.g. config generation.
+// +k8s:deepcopy-gen=false
+type PrometheusInterface interface {
+	metav1.ObjectMetaAccessor
+	GetCommonPrometheusFields() CommonPrometheusFields
+}
+
+func (l *Prometheus) GetCommonPrometheusFields() CommonPrometheusFields {
+	return l.Spec.CommonPrometheusFields
+}
+
 // CommonPrometheusFields are the options available to both the Prometheus server and agent.
 // +k8s:deepcopy-gen=true
 type CommonPrometheusFields struct {
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -55,17 +55,17 @@ type ConfigGenerator struct {
 	logger                 log.Logger
 	version                semver.Version
 	notCompatible          bool
-	spec                   *v1.PrometheusSpec
+	prom                   v1.PrometheusInterface
 	endpointSliceSupported bool
 }
 
 // NewConfigGenerator creates a ConfigGenerator for the provided Prometheus resource.
-func NewConfigGenerator(logger log.Logger, p *v1.Prometheus, endpointSliceSupported bool) (*ConfigGenerator, error) {
+func NewConfigGenerator(logger log.Logger, p v1.PrometheusInterface, endpointSliceSupported bool) (*ConfigGenerator, error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
 
-	promVersion := operator.StringValOrDefault(p.Spec.Version, operator.DefaultPrometheusVersion)
+	promVersion := operator.StringValOrDefault(p.GetCommonPrometheusFields().Version, operator.DefaultPrometheusVersion)
 	version, err := semver.ParseTolerant(promVersion)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to parse Prometheus version")
@@ -76,7 +76,7 @@ func NewConfigGenerator(logger log.Logger, p *v1.Prometheus, endpointSliceSuppor
 	return &ConfigGenerator{
 		logger:                 logger,
 		version:                version,
-		spec:                   p.Spec.DeepCopy(),
+		prom:                   p,
 		endpointSliceSupported: endpointSliceSupported,
 	}, nil
 }
@@ -89,7 +89,7 @@ func (cg *ConfigGenerator) WithKeyVals(keyvals ...interface{}) *ConfigGenerator 
 		logger:                 log.WithSuffix(cg.logger, keyvals),
 		version:                cg.version,
 		notCompatible:          cg.notCompatible,
-		spec:                   cg.spec,
+		prom:                   cg.prom,
 		endpointSliceSupported: cg.endpointSliceSupported,
 	}
 }
@@ -106,7 +106,7 @@ func (cg *ConfigGenerator) WithMinimumVersion(version string) *ConfigGenerator {
 			logger:                 log.WithSuffix(cg.logger, "minimum_version", version),
 			version:                cg.version,
 			notCompatible:          true,
-			spec:                   cg.spec,
+			prom:                   cg.prom,
 			endpointSliceSupported: cg.endpointSliceSupported,
 		}
 	}
@@ -126,7 +126,7 @@ func (cg *ConfigGenerator) WithMaximumVersion(version string) *ConfigGenerator {
 			logger:                 log.WithSuffix(cg.logger, "maximum_version", version),
 			version:                cg.version,
 			notCompatible:          true,
-			spec:                   cg.spec,
+			prom:                   cg.prom,
 			endpointSliceSupported: cg.endpointSliceSupported,
 		}
 	}
@@ -219,8 +219,9 @@ func (cg *ConfigGenerator) AddLimitsToYAML(cfg yaml.MapSlice, k limitKey, limit 
 // For backwards compatibility with Prometheus <2.9.0 we don't set
 // honor_timestamps.
 func (cg *ConfigGenerator) AddHonorTimestamps(cfg yaml.MapSlice, userHonorTimestamps *bool) yaml.MapSlice {
+	cpf := cg.prom.GetCommonPrometheusFields()
 	// Fast path.
-	if userHonorTimestamps == nil && !cg.spec.OverrideHonorTimestamps {
+	if userHonorTimestamps == nil && !cpf.OverrideHonorTimestamps {
 		return cfg
 	}
 
@@ -229,13 +230,13 @@ func (cg *ConfigGenerator) AddHonorTimestamps(cfg yaml.MapSlice, userHonorTimest
 		honor = *userHonorTimestamps
 	}
 
-	return cg.WithMinimumVersion("2.9.0").AppendMapItem(cfg, "honor_timestamps", honor && !cg.spec.OverrideHonorTimestamps)
+	return cg.WithMinimumVersion("2.9.0").AppendMapItem(cfg, "honor_timestamps", honor && !cpf.OverrideHonorTimestamps)
 }
 
 // AddHonorLabels adds the honor_labels field into scrape configurations.
 // if OverrideHonorLabels is true then honor_labels is always false.
 func (cg *ConfigGenerator) AddHonorLabels(cfg yaml.MapSlice, honorLabels bool) yaml.MapSlice {
-	if cg.spec.OverrideHonorLabels {
+	if cg.prom.GetCommonPrometheusFields().OverrideHonorLabels {
 		honorLabels = false
 	}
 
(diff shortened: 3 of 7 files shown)


ArthurSens:
Does anyone know how to make `controller-gen` happy with interfaces? 😅

pkg/prometheus/promcfg.go
```diff
@@ -421,7 +423,7 @@ func compareScrapeTimeoutToScrapeInterval(scrapeTimeout, scrapeInterval monitori
 
 // Generate creates a serialized YAML representation of a Prometheus configuration using the provided resources.
 func (cg *ConfigGenerator) Generate(
```
JoaoBraveCoding, line 425: I know `ConfigGenerator` type already has a field called `spec` where we can extract `cpf`. Since we are refactoring a bit how hard is it to make this consistent? (I'm not sure, just raising the question) 
Either we could always the necessary structure (`PrometheusInterface`) as function arguments or have them as a type in `ConfigGenerator`

JoaoBraveCoding (commented):
Good start! 👍

pkg/prometheus/promcfg.go
ArthurSens: Thanks for the initial review Joao! Could you elaborate a little bit here?

Do you mean changing the ConfigGenerator struct to something like this?
```go
type ConfigGenerator struct {
	logger                 log.Logger
	version                semver.Version
	notCompatible          bool
	//spec                   *v1.PrometheusSpec
	spec                   *v1.PrometheusInterface
	endpointSliceSupported bool
}
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -28,6 +28,21 @@ const (
 	PrometheusKindKey = "prometheus"
 )
 
+// PrometheusInterface is used by Prometheus and PrometheusAgent to share common methods, e.g. config generation.
+// +k8s:deepcopy-gen=false
+type PrometheusInterface interface {
+	GetCommonPrometheusFields() CommonPrometheusFields
+	GetObjectMeta() metav1.ObjectMeta
```
simonpasquier, line 35: suggestion: leverage the existing [ObjectMetaAccessor](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMetaAccessor) interface.

```suggestion
	metav1.ObjectMetaAccessor
```

pkg/prometheus/promcfg.go
```diff
@@ -554,22 +565,25 @@ func (cg *ConfigGenerator) Generate(
...
-		return nil, errors.Wrap(err, "generating storage_settings configuration failed")
+	if len(cpf.RemoteWrite) > 0 {
+		cfg = append(cfg, cg.generateRemoteWriteConfig(p.GetCommonPrometheusFields(), store, p.GetObjectMeta().Namespace))
 	}
 
-	cfg, err = cg.appendAlertingConfig(cfg, p, additionalAlertRelabelConfigs, additionalAlertManagerConfigs, store)
-	if err != nil {
-		return nil, errors.Wrap(err, "generating alerting configuration failed")
-	}
+	switch prom := p.(type) {
```
simonpasquier, line 572: I'm not a fan of type casting in Go :)
Since alerting isn't applicable to the Prometheus agent, I'd suggest to split the Generate() function into more composable fragments. The server and agent controllers can pipe them up as needed. 

```golang
func (cg *ConfigGenerator) AppendScrapeConfigs(
  cfg yaml.MapSlice,
  ..., // required inputs for building the scrape_config
) (yaml.MapSlice, error) {
  ...
}

func (cg *ConfigGenerator) AppendAlertingConfig(
  cfg yaml.MapSlice,
...
) (yaml.MapSlice, error) {
  ...
}

func (cg *ConfigGenerator) AppendGlobalConfig(
  cfg yaml.MapSlice,
...
) (yaml.MapSlice, error) {
  ...
}

func (cg *ConfigGenerator) AppendRemoteWrite(
  cfg yaml.MapSlice,
...
) (yaml.MapSlice, error) {
  ...
}
```

pkg/prometheus/promcfg.go
simonpasquier: I think that it makes sense to have the PrometheusInterface field in the config generator. Any parameter that is specific to the Prometheus server can be passed to the config generator's methods.
E.g. [appendStorageSettingsConfig](https://github.com/prometheus-operator/prometheus-operator/blob/fcd34a44045503bc45093887776c4b9ca675f6ed/pkg/prometheus/promcfg.go#L578-L607) would become

```golang
func (cg *ConfigGenerator) appendStorageSettingsConfig(cfg yaml.MapSlice, exemplars *v1.Exemplars, tsdb v1.TSBD) (yaml.MapSlice, error) {
...
}
```

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: Oh nice, I wasn't aware this existed :)

pkg/prometheus/promcfg.go
ArthurSens: I don't see how we avoid typecasting with this approach 🤔

If both controllers use the same ConfigGenerator, we need to know the resource type to decide if we're calling server-specific configuration or not, don't we?

simonpasquier: For the record, my idea is that each controller will build its own Generate() function based on primitives offered by the ConfigGenerator struct.

pkg/prometheus/promcfg.go
```diff
@@ -1929,3 +1864,157 @@ func (cg *ConfigGenerator) generateRemoteWriteConfig(
...
+				store,
+				shards,
+			),
+		)
+	}
+
+	return slices
+}
+
+func (cg *ConfigGenerator) AppendAdditionalScrapeConfigs(slice yaml.MapSlice, additionalScrapeConfigs []byte, shards int32) ([]yaml.MapSlice, error) {
```
simonpasquier, line 2014: `slice` isn't used anywhere in the function. Did you mean to append to the input slice and pass `scrapeConfigs` to the function?

pkg/prometheus/promcfg.go
```diff
@@ -431,170 +438,91 @@ func (cg *ConfigGenerator) Generate(
...
-	apiserverConfig := p.Spec.APIServerConfig
-	shards := int32(1)
-	if p.Spec.Shards != nil && *p.Spec.Shards > 1 {
-		shards = *p.Spec.Shards
-	}
+	// Rule Files config
+	cfg = cg.AppendRuleFiles(cfg, ruleConfigMapNames, ruleSelector)
 
+	// Scrape config
 	var scrapeConfigs []yaml.MapSlice
```
simonpasquier, line 463: nit (GitHub won't let me comment on the 3 lines)
```golang
	var (
		scrapeConfigs []yaml.MapSlice
		apiserverConfig = cpf.APIServerConfig
		shards = int32(1)
	)
```

pkg/prometheus/promcfg.go
```diff
@@ -431,170 +438,91 @@ func (cg *ConfigGenerator) Generate(
...
+	if cpf.Shards != nil && *cpf.Shards > 1 {
+		shards = *cpf.Shards
 	}
-
-	var addlScrapeConfigs []yaml.MapSlice
-	addlScrapeConfigs, err := cg.generateAdditionalScrapeConfigs(additionalScrapeConfigs, shards)
+	scrapeConfigs = cg.AppendServiceMonitorConfigs(scrapeConfigs, sMons, apiserverConfig, store, shards)
+	scrapeConfigs = cg.AppendPodMonitorConfigs(scrapeConfigs, pMons, apiserverConfig, store, shards)
+	scrapeConfigs = cg.AppendProbeConfigs(scrapeConfigs, probes, apiserverConfig, store, shards)
+	addlScrapeConfigs, err := cg.AppendAdditionalScrapeConfigs(cfg, additionalScrapeConfigs, shards)
```
simonpasquier, line 472: see my comment below on `AppendAdditionalScrapeConfigs` but I think that we want this
```suggestion
	addlScrapeConfigs, err := cg.AppendAdditionalScrapeConfigs(cfg, scrapeConfigs, shards)
```

pkg/prometheus/promcfg.go
```diff
@@ -431,170 +438,91 @@ func (cg *ConfigGenerator) Generate(
...
 	if err != nil {
 		return nil, errors.Wrap(err, "generate additional scrape configs")
 	}
-
 	cfg = append(cfg, yaml.MapItem{
 		Key:   "scrape_configs",
 		Value: append(scrapeConfigs, addlScrapeConfigs...),
```
simonpasquier, line 478: see comment above
```suggestion
		Value: append(scrapeConfigs),
```

pkg/prometheus/promcfg_test.go
```diff
@@ -2003,19 +2081,13 @@ func TestAdditionalScrapeConfigs(t *testing.T) {
...
-				},
-				Spec: monitoringv1.PrometheusSpec{
-					CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
-						Shards:         shards,
-						ScrapeInterval: "30s",
-					},
-					EvaluationInterval: "30s",
-				},
-			},
+			"30s",
```
simonpasquier, line 2084: nit: maybe create the Prometheus object before like for the other tests? I find it easier to read/understand what we're testing.

Documentation/api.md
```diff
@@ -8228,6 +8228,11 @@ string
 </tr>
 </tbody>
 </table>
+<h3 id="monitoring.coreos.com/v1.PrometheusInterface">PrometheusInterface
+</h3>
+<div>
+<p>PrometheusInterface is used by Prometheus and PrometheusAgent to share common methods, e.g. config generation.</p>
+</div>
```
simonpasquier, line 8235: I think you need to tweak https://github.com/prometheus-operator/prometheus-operator/blob/main/scripts/docs/config.json to hide the interface type from the docs.

simonpasquier (commented):
Wow, this is nice! Just a few nits.

Documentation/api.md
ArthurSens: Do you know where can I find documentation for this tool? I couldn't find it in the original repo.

I've tried adding `PrometheusInterface` to `hideTypePatterns`, but it it made no changes to the API docs :(

pkg/prometheus/promcfg.go
ArthurSens: Oh, I think that was just copy-paste mistake 😅

pkg/prometheus/promcfg_test.go
ArthurSens: I just noticed that we have a variable `p` of type Prometheus just some lines above. I'm not sure why a new one was being created 😅

pkg/prometheus/promcfg.go
```diff
@@ -370,30 +371,30 @@ func (cg *ConfigGenerator) addAuthorizationToYaml(
 	return cg.WithMinimumVersion("2.26.0").WithKeyVals("component", strings.Split(assetStoreKey, "/")[0]).AppendMapItem(cfg, "authorization", authCfg)
 }
 
-func buildExternalLabels(p *v1.Prometheus) yaml.MapSlice {
+func buildExternalLabels(cpf v1.CommonPrometheusFields, name, namespace string) yaml.MapSlice {
```
JoaoBraveCoding, line 374: Do we have a reason not to make this function a method of the struct `ConfigGenerator`? This would allow us to do something like the following no? Or am I missing something 🤔 
```suggestion
func (cg *ConfigGenerator) buildExternalLabels() yaml.MapSlice {
```

pkg/prometheus/promcfg.go
```diff
@@ -431,170 +438,93 @@ func (cg *ConfigGenerator) Generate(
...
-
-	if p.Spec.QueryLogFile != "" {
-		globalItems = cg.WithMinimumVersion("2.16.0").AppendMapItem(globalItems, "query_log_file", queryLogFilePath(p))
-	}
-
+	globalItems := yaml.MapSlice{}
+	globalItems = cg.AppendEvaluationInterval(globalItems, evaluationInterval)
+	globalItems = cg.AppendScrapeIntervals(globalItems)
+	globalItems = cg.AppendExternalLabels(globalItems)
+	globalItems = cg.AppendQueryLogFile(globalItems, queryLogFile)
```
JoaoBraveCoding, line 456: nit: since these methods will only be used inside of this package we could not export them, right?
```suggestion
	globalItems = cg.appendEvaluationInterval(globalItems, evaluationInterval)
	globalItems = cg.appendScrapeIntervals(globalItems)
	globalItems = cg.appendExternalLabels(globalItems)
	globalItems = cg.appendQueryLogFile(globalItems, queryLogFile)
```

pkg/prometheus/promcfg.go
```diff
@@ -421,7 +422,13 @@ func compareScrapeTimeoutToScrapeInterval(scrapeTimeout, scrapeInterval monitori
...
 // Generate creates a serialized YAML representation of a Prometheus configuration using the provided resources.
 func (cg *ConfigGenerator) Generate(
-	p *v1.Prometheus,
+	evaluationInterval v1.Duration,
+	queryLogFile string,
+	ruleSelector *metav1.LabelSelector,
+	exemplars *v1.Exemplars,
+	tsdb v1.TSDBSpec,
+	alerting *v1.AlertingSpec,
+	remoteRead []v1.RemoteReadSpec,
```
JoaoBraveCoding, line 431: non-blocking comment, doing it just in case someone has a good solution: I'm a bit afraid that passing the non-agent Prometheus Config like this might become a bit annoying to maintain 🤔 not sure if putting these inside of a struct might make sense.

JoaoBraveCoding (commented):
Great work @ArthurSens, super happy to see this progressing! Just a few comments

pkg/prometheus/promcfg.go
ArthurSens: Very good point! I don't think there is any special reason 🤔

pkg/prometheus/promcfg.go
ArthurSens: I think this is still manageable but I can definitely see your point! My preference here would be to do what the current maintainers prefer, so... wdyt @prometheus-operator/prometheus-operator-reviewers?

JoaoBraveCoding (commented):
Just one more nit, but apart from that lgtm 👍

pkg/prometheus/promcfg.go
```diff
@@ -431,170 +440,93 @@ func (cg *ConfigGenerator) Generate(
...
-					pMons[identifier], ep, i,
-					apiserverConfig,
-					store,
-					shards,
-				),
-			)
-		}
-	}
+	// Rule Files config
+	cfg = cg.AppendRuleFiles(cfg, ruleConfigMapNames, ruleSelector)
```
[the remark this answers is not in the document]
ArthurSens: Whooops, not sure how I missed that one 😅

Documentation/api.md
ArthurSens: Finally got it!

pkg/prometheus/promcfg.go
simonpasquier: This is fine by me. When we move to implementing the Prometheus agent part, I imagine that we'll rename `*ConfigGenerator.Generate()` to `*ConfigGenerator.GenerateServerConfiguration()` and create a `*ConfigGenerator.GenerateAgentConfiguration()` method

simonpasquier (approved):
:+1:

PR #5311, opened 2023-01-31
Use tmpfs to store Prometheus/Alertmanager config

ArthurSens:
## Description

Changes the emptyDir that is used to store Prometheus and Alertmanager configuration to use `tmpfs`. While the content is still readable if the host gets invaded while running, the data is wiped once the host is turned off, preventing leaking data from disks.

## Type of change

## Changelog entry

```release-note
Use tmpfs to store Prometheus and Alertmanager configuration
```

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -465,7 +465,10 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 		{
 			Name: "config-out",
 			VolumeSource: v1.VolumeSource{
-				EmptyDir: &v1.EmptyDirVolumeSource{},
+				EmptyDir: &v1.EmptyDirVolumeSource{
+					// tmpfs is used here to avoid writing sensitive data into disk.
+					Medium: v1.StorageMediumMemory,
+				},
 			},
 		},
 	}
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -480,7 +480,10 @@ func makeStatefulSetSpec(
 		{
 			Name: "config-out",
 			VolumeSource: v1.VolumeSource{
-				EmptyDir: &v1.EmptyDirVolumeSource{},
+				EmptyDir: &v1.EmptyDirVolumeSource{
+					// tmpfs is used here to avoid writing sensitive data into disk.
+					Medium: v1.StorageMediumMemory,
+				},
 			},
 		},
 	}
--- a/pkg/prometheus/statefulset_test.go
+++ b/pkg/prometheus/statefulset_test.go
@@ -364,7 +364,9 @@ func TestStatefulSetVolumeInitial(t *testing.T) {
 						{
 							Name: "config-out",
 							VolumeSource: v1.VolumeSource{
-								EmptyDir: &v1.EmptyDirVolumeSource{},
+								EmptyDir: &v1.EmptyDirVolumeSource{
+									Medium: v1.StorageMediumMemory,
+								},
 							},
 						},
 						{


simonpasquier:
This could be considered to be a CHANGE as the used space will count against the memory usage of the pod. Otherwise LGTM, once it's green ;)

ArthurSens:
> This could be considered to be a CHANGE as the used space will count against the memory usage of the pod. Otherwise LGTM, once it's green ;)

good point!

simonpasquier (approved):

shay-berman:
Hello

The whole idea is to avoid exposing the secrets in the prometheus configuration to the k8s node folesystem.  Syandard Emptydir is not enough because it stored as volume local in the host disk.

Changing it to tmpfs idea is to make sire its not on the disk bit only in memory.

But o am wondering if tjis is enough from security stand point?

Maybe it will be good to allow the use choose what will be the storage class behind the empty disk. So they may choose to use encrypted disk instead for more security.
Maybe other aspects to make it more secure?

paulfantom:
> The whole idea is to avoid exposing the secrets in the prometheus configuration to the k8s node folesystem.

I don't think this is possible. This data has to live in the filesystem in one way or another.

Also, a file with secrets (configuration file) needs to be readable to prometheus during runtime, thus this file has to be mounted and available in the filesystem. As such allowing storage class configuration won't improve security as the encrypted storage class provides only additional security for data at rest not during runtime and thus it offers similar security guarantees as tmpfs in this case.

I would like to ask what would be the vector of attack for this data? From what I see this data can now be retrieved only with administrative access (either to a node or to cluster) and at this point, it is moot to secure it further. It was also the case before as `/var/lib/kubelet` is owned by `root` and not world-readable, but using `tmpfs` allows better handling of data at rest.

shay-berman:
Thanks paul for your quick response.

The point is that an operator that read a k8s secret and manipulation it and then store it into prometheus container as regular emptydir its not good because it means the secret is not handled in a secure way (and this is what prometheus operator does).

So probably on tmpfs is a good enough solution to make sure it note storing it on the node filesystem but only in the RAM.

BTW About `Let's keep in mind that even env variables are stored in the filesystem`.  I dont think that evc variable stored in node filesystem, it is stored in /proc which is tmpfs.

Anyway thanks a lot on your responcivness and the fix.

paulfantom:
> I dont think that evc variable stored in node filesystem, it is stored in /proc which is tmpfs.

`/proc` has it's own filesystem called [procfs](https://en.wikipedia.org/wiki/Procfs) and it is not using tmpfs.

> The point is that an operator that read a k8s secret and manipulation it and then store it into prometheus container as regular emptydir its not good because it means the secret is not handled in a secure way

Could you explain why do you think this is not a secure way?

shay-berman:
1. I think that /proc is not stored on disk, it lives only in RAM ([link](https://askubuntu.com/questions/1112177/proc-directory-stored-in-a-ram-or-hard-drive)).

2. You can check in k8s documentation about how Secrets handled in k8s and you will see that secrets mounted into the pod using tmpfs "so that the confidential data is not written to durable storage" [HERE](https://kubernetes.io/docs/concepts/configuration/secret/#information-security-for-secrets).
image

Anyway really thanks for this fix, I wonder in which Prometheus operator image this fix will be available?

simonpasquier:
@shay-berman thanks again for reporting your concerns! The change will be available in the next minor release of prometheus-operator (v0.63.0) which is planned for next week (see https://github.com/prometheus-operator/prometheus-operator/blob/main/RELEASE.md#release-schedule).

With respect to environment variables, there are audit policies that discourage passing secrets via env vars:
https://www.tenable.com/audits/items/CIS_Kubernetes_v1.20_v1.0.1_Level_2_Master.audit:98de3da69271994afb6211cf86ae4c6b

shay-berman:
Thanks a lot for your quick fix.
Appreciate your effort.

PR #5316, opened 2023-02-02
Renames ReloaderConfig to ReloaderOperatorConfig and refactors pkg/*/statefulset_test

JoaoBraveCoding:
## Description

Problem: The name ReloaderConfig conflicted with the function under pkg/operator/config_reloader.go that would set ReloaderConfig, this function was previously called ReloaderResources which did not indicate clearly that other fields like image were also being configured. Furthermore each pkg/*/statefulset_test.go file was testing extensively the resources for the ReloadConfig sidecar which created a lot of code duplication.

Solution: Rename the type ReloaderConfig to ReloaderOperatorConfig since the config comes from the operator and refactor tests to extract the logic to the resources tests for ReloadConfig sidecar into a dedicated lib under pkg/operator/config_reloader_test_lib.go

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -31,7 +31,7 @@ type Config struct {
 	TLSInsecure                  bool
 	TLSConfig                    rest.TLSClientConfig
 	ServerTLSConfig              server.TLSServerConfig
-	ReloaderConfig               ReloaderConfig
+	ReloaderConfig               ContainerConfig
 	AlertmanagerDefaultBaseImage string
 	PrometheusDefaultBaseImage   string
 	ThanosDefaultBaseImage       string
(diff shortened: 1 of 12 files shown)


pkg/operator/config.go
```diff
@@ -46,7 +46,7 @@ type Config struct {
 	SecretListWatchSelector      string
 }
 
-type ReloaderConfig struct {
+type ReloaderOperatorConfig struct {
```
simonpasquier, line 49: this struct is about configuration options for the container.
```suggestion
type ContainerConfig struct {
```

simonpasquier (approved):
lgtm

PR #5326, opened 2023-02-08
Fix path of flow diagram image

mghantous:
## Description

This fixes the path of the flow diagram image in the troubleshooting documentation

## Type of change

## Changelog entry

```release-note
- Fixes the path of the flow diagram image in the troubleshooting documentation
```

### DIFF
--- a/Documentation/troubleshooting.md
+++ b/Documentation/troubleshooting.md
@@ -51,7 +51,7 @@ A common problem related to `ServiceMonitor` identification by Prometheus is rel
 
 <!-- do not change this link without verifying that the image will display correctly on https://prometheus-operator.dev -->
 
-![flow diagram](/img/custom-metrics-elements.png)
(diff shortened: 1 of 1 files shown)


slashpai:
Thanks for the PR

The diagram will be displayed correctly in https://prometheus-operator.dev/docs/operator/troubleshooting/ if we change it here it wont be displayed correctly over there

mghantous:
Thanks for the review. I see, so, how do you think we should proceed in this case? It seems there are at least 3 options but I don't believe any of them are ideal:

1. Use absolute image urls to either `github` or `prometheus-operator.dev`
2. Move images around on either `github` or `prometheus-operator.dev` so that the relative urls work on both sites
3. Make no change. Close this PR and live with the fact that images are broken if viewed from `github`.

simonpasquier:
We get PRs like this so frequently that we need to fix it. Having said that, the source of truth for the documentation is https://prometheus-operator.dev/.

PR #5334, opened 2023-02-10
Refactor: Use PrometheusInterface to generate statefulset

ArthurSens:
## Description

This PR was built on top of https://github.com/prometheus-operator/prometheus-operator/pull/5310, so I could re-use the `PrometheusInterface` interface.

It follows the same strategy, arguments that are not present in `CommonPrometheusFields` are being passed as extra arguments to the functions `makeStatefulSet` and `makeStatefulSetSpec`.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/apis/monitoring/v1/prometheus_types.go
+++ b/pkg/apis/monitoring/v1/prometheus_types.go
@@ -32,11 +32,21 @@ const (
 // +k8s:deepcopy-gen=false
 type PrometheusInterface interface {
 	metav1.ObjectMetaAccessor
+	GetTypeMeta() metav1.TypeMeta
 	GetCommonPrometheusFields() CommonPrometheusFields
+	SetCommonPrometheusFields(CommonPrometheusFields)
 }
 
 func (l *Prometheus) GetCommonPrometheusFields() CommonPrometheusFields {
 	return l.Spec.CommonPrometheusFields
+}
+
+func (l *Prometheus) SetCommonPrometheusFields(f CommonPrometheusFields) {
+	l.Spec.CommonPrometheusFields = f
+}
+
+func (l *Prometheus) GetTypeMeta() metav1.TypeMeta {
+	return l.TypeMeta
 }
 
 // CommonPrometheusFields are the options available to both the Prometheus server and agent.
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -472,7 +472,7 @@ func makeStatefulSetSpec(
 	if disableCompaction {
 		thanosBlockDuration := "2h"
 		if thanos != nil {
-			thanosBlockDuration = string(thanos.BlockDuration)
+			thanosBlockDuration = operator.StringValOrDefault(string(thanos.BlockDuration), thanosBlockDuration)
 		}
 		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.max-block-duration", Value: thanosBlockDuration})
 		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.min-block-duration", Value: thanosBlockDuration})
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -470,7 +470,10 @@ func makeStatefulSetSpec(
 	}
 
 	if disableCompaction {
-		thanosBlockDuration := operator.StringValOrDefault(string(thanos.BlockDuration), "2h")
+		thanosBlockDuration := "2h"
+		if thanos != nil {
+			thanosBlockDuration = string(thanos.BlockDuration)
+		}
 		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.max-block-duration", Value: thanosBlockDuration})
 		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.min-block-duration", Value: thanosBlockDuration})
 	}
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -96,7 +96,17 @@ func prometheusNameByShard(name string, shard int32) string {
 func makeStatefulSet(
 	logger log.Logger,
 	name string,
-	p monitoringv1.Prometheus,
+	p monitoringv1.PrometheusInterface,
+	baseImage, tag, sha string,
+	retention monitoringv1.Duration,
+	retentionSize monitoringv1.ByteSize,
+	rules monitoringv1.Rules,
+	query *monitoringv1.QuerySpec,
+	allowOverlappingBlocks bool,
+	enableAdminAPI bool,
+	queryLogFile string,
+	thanos *monitoringv1.ThanosSpec,
+	disableCompaction bool,
 	config *operator.Config,
 	cg *ConfigGenerator,
 	ruleConfigMapNames []string,
@@ -104,19 +114,26 @@ func makeStatefulSet(
 	shard int32,
 	tlsAssetSecrets []string,
 ) (*appsv1.StatefulSet, error) {
-	if p.Spec.PortName == "" {
-		p.Spec.PortName = defaultPortName
+	cpf := p.GetCommonPrometheusFields()
+	objMeta := p.GetObjectMeta()
+	typeMeta := p.GetTypeMeta()
+
+	if cpf.PortName == "" {
+		cpf.PortName = defaultPortName
 	}
 
-	if p.Spec.Replicas == nil {
-		p.Spec.Replicas = &minReplicas
+	if cpf.Replicas == nil {
+		cpf.Replicas = &minReplicas
 	}
 	intZero := int32(0)
-	if p.Spec.Replicas != nil && *p.Spec.Replicas < 0 {
-		p.Spec.Replicas = &intZero
+	if cpf.Replicas != nil && *cpf.Replicas < 0 {
+		cpf.Replicas = &intZero
 	}
 
-	spec, err := makeStatefulSetSpec(logger, p, config, cg, shard, ruleConfigMapNames, tlsAssetSecrets)
+	// We need to re-set the common fields because cpf is only a copy of the original object.
+	// We set some defaults if some fields are not present, and we want those fields set in the original Prometheus object before building the StatefulSetSpec.
+	p.SetCommonPrometheusFields(cpf)
+	spec, err := makeStatefulSetSpec(logger, baseImage, tag, sha, retention, retentionSize, rules, query, allowOverlappingBlocks, enableAdminAPI, queryLogFile, thanos, disableCompaction, p, config, cg, shard, ruleConfigMapNames, tlsAssetSecrets)
 	if err != nil {
 		return nil, errors.Wrap(err, "make StatefulSet spec")
 	}
@@ -125,17 +142,17 @@ func makeStatefulSet(
 	// do not transfer kubectl annotations to the statefulset so it is not
 	// pruned by kubectl
 	annotations := make(map[string]string)
-	for key, value := range p.ObjectMeta.Annotations {
+	for key, value := range objMeta.GetAnnotations() {
 		if !strings.HasPrefix(key, "kubectl.kubernetes.io/") {
 			annotations[key] = value
 		}
 	}
 	labels := make(map[string]string)
-	for key, value := range p.ObjectMeta.Labels {
+	for key, value := range objMeta.GetLabels() {
 		labels[key] = value
 	}
 	labels[shardLabelName] = fmt.Sprintf("%d", shard)
-	labels[prometheusNameLabelName] = p.Name
+	labels[prometheusNameLabelName] = objMeta.GetName()
 
 	statefulset := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
@@ -144,12 +161,12 @@ func makeStatefulSet(
 			Annotations: annotations,
 			OwnerReferences: []metav1.OwnerReference{
 				{
-					APIVersion:         p.APIVersion,
+					APIVersion:         typeMeta.APIVersion,
 					BlockOwnerDeletion: &boolTrue,
 					Controller:         &boolTrue,
-					Kind:               p.Kind,
-					Name:               p.Name,
-					UID:                p.UID,
+					Kind:               typeMeta.Kind,
+					Name:               objMeta.GetName(),
+					UID:                objMeta.GetUID(),
 				},
 			},
 		},
@@ -164,13 +181,13 @@ func makeStatefulSet(
 		statefulset.ObjectMeta.Annotations[sSetInputHashName] = inputHash
 	}
 
-	if p.Spec.ImagePullSecrets != nil && len(p.Spec.ImagePullSecrets) > 0 {
-		statefulset.Spec.Template.Spec.ImagePullSecrets = p.Spec.ImagePullSecrets
+	if cpf.ImagePullSecrets != nil && len(cpf.ImagePullSecrets) > 0 {
+		statefulset.Spec.Template.Spec.ImagePullSecrets = cpf.ImagePullSecrets
 	}
-	storageSpec := p.Spec.Storage
+	storageSpec := cpf.Storage
 	if storageSpec == nil {
 		statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes, v1.Volume{
-			Name: volumeName(p.Name),
+			Name: volumeName(objMeta.GetName()),
 			VolumeSource: v1.VolumeSource{
 				EmptyDir: &v1.EmptyDirVolumeSource{},
 			},
@@ -178,7 +195,7 @@ func makeStatefulSet(
 	} else if storageSpec.EmptyDir != nil {
 		emptyDir := storageSpec.EmptyDir
 		statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes, v1.Volume{
-			Name: volumeName(p.Name),
+			Name: volumeName(objMeta.GetName()),
 			VolumeSource: v1.VolumeSource{
 				EmptyDir: emptyDir,
 			},
@@ -186,7 +203,7 @@ func makeStatefulSet(
 	} else if storageSpec.Ephemeral != nil {
 		ephemeral := storageSpec.Ephemeral
 		statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes, v1.Volume{
-			Name: volumeName(p.Name),
+			Name: volumeName(objMeta.GetName()),
 			VolumeSource: v1.VolumeSource{
 				Ephemeral: ephemeral,
 			},
@@ -194,7 +211,7 @@ func makeStatefulSet(
 	} else {
 		pvcTemplate := operator.MakeVolumeClaimTemplate(storageSpec.VolumeClaimTemplate)
 		if pvcTemplate.Name == "" {
-			pvcTemplate.Name = volumeName(p.Name)
+			pvcTemplate.Name = volumeName(objMeta.GetName())
 		}
 		if storageSpec.VolumeClaimTemplate.Spec.AccessModes == nil {
 			pvcTemplate.Spec.AccessModes = []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce}
@@ -206,9 +223,9 @@ func makeStatefulSet(
 		statefulset.Spec.VolumeClaimTemplates = append(statefulset.Spec.VolumeClaimTemplates, *pvcTemplate)
 	}
 
-	statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes, p.Spec.Volumes...)
+	statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes, cpf.Volumes...)
 
-	if p.Spec.HostNetwork {
+	if cpf.HostNetwork {
 		statefulset.Spec.Template.Spec.DNSPolicy = v1.DNSClusterFirstWithHostNet
 	}
 
@@ -298,7 +315,17 @@ func makeStatefulSetService(p *monitoringv1.Prometheus, config operator.Config) 
 
 func makeStatefulSetSpec(
 	logger log.Logger,
-	p monitoringv1.Prometheus,
+	baseImage, tag, sha string,
+	retention monitoringv1.Duration,
+	retentionSize monitoringv1.ByteSize,
+	rules monitoringv1.Rules,
+	query *monitoringv1.QuerySpec,
+	allowOverlappingBlocks bool,
+	enableAdminAPI bool,
+	queryLogFile string,
+	thanos *monitoringv1.ThanosSpec,
+	disableCompaction bool,
+	p monitoringv1.PrometheusInterface,
 	c *operator.Config,
 	cg *ConfigGenerator,
 	shard int32,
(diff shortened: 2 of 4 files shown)


ArthurSens:
It is worth mentioning that I've noticed some fields that could be used with PrometheusAgent were not present in `CommonPrometheusFields`. I've moved them, but I'm not sure if that is a breaking change for any reason... could someone clarify?

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
 
 	return filepath.Join(defaultQueryLogDirectory, queryLogFile)
 }
+
+// buildCommonPrometheusArgs builds an slice of arguments that are common between Prometheus Server and Agent.
```
JoaoBraveCoding, line 680:
```suggestion
// buildCommonPrometheusArgs builds a slice of arguments that are common between Prometheus Server and Agent.
```

pkg/prometheus/statefulset.go
```diff
@@ -325,222 +351,30 @@ func makeStatefulSetSpec(
...
-		if *p.Spec.WALCompression {
-			arg.Name = "storage.tsdb.wal-compression"
-		}
-		promArgs = cg.WithMinimumVersion("2.11.0").AppendCommandlineArgument(promArgs, arg)
-	}
-
-	if p.Spec.AllowOverlappingBlocks {
-		promArgs = cg.WithMinimumVersion("2.11.0").WithMaximumVersion("2.39.0").AppendCommandlineArgument(promArgs, monitoringv1.Argument{Name: "storage.tsdb.allow-overlapping-blocks"})
+	if cpf.RoutePrefix != "" {
+		webRoutePrefix = cpf.RoutePrefix
```
JoaoBraveCoding, line 356: What if we flip the condition and then we just stop passing `webRoutePrefix` as a parameter and we use instead `cpf.RoutePrefix`?
```suggestion
	if cpf.RoutePrefix == "" {
		cpf.RoutePrefix = webRoutePrefix
```

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+		if cpf.Web.MaxConnections != nil {
+			promArgs = append(promArgs, monitoringv1.Argument{Name: "web.max-connections", Value: fmt.Sprintf("%d", *cpf.Web.MaxConnections)})
+		}
+	}
+
+	if cpf.EnableAdminAPI {
+		promArgs = append(promArgs, monitoringv1.Argument{Name: "web.enable-admin-api"})
+	}
+
+	if cpf.EnableRemoteWriteReceiver {
```
JoaoBraveCoding, line 703: Is agent mode really okay with this argument? 🤔

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+			*promArgs = append(*promArgs, monitoringv1.Argument{Name: "query.timeout", Value: string(*query.Timeout)})
+		}
+	}
+
+	if allowOverlappingBlocks {
+		*promArgs = cg.WithMinimumVersion("2.11.0").WithMaximumVersion("2.39.0").AppendCommandlineArgument(*promArgs, monitoringv1.Argument{Name: "storage.tsdb.allow-overlapping-blocks"})
+	}
+}
+
+// AddCommonVolumes adds volumes to the statefulset spec that are common between Prometheus Server and Agent
```
JoaoBraveCoding, line 807:
```suggestion
// buildCommonVolumes returns a set of volumes to be mounted on the statefulset that are common between Prometheus Server and Agent
```

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+func buildCommonVolumes(cpf monitoringv1.CommonPrometheusFields, promName, volName string, tlsAssetSecrets []string) ([]v1.Volume, []v1.VolumeMount, error) {
+	assetsVolume := v1.Volume{
```
JoaoBraveCoding, line 809: And I suspect we might also be able to stop passing `volName` 👀 but didn't look deep into it
```suggestion
func buildCommonVolumes(p monitoringv1.PrometheusInterface, volName string, tlsAssetSecrets []string) ([]v1.Volume, []v1.VolumeMount, error) {
        cpf := p.GetCommonPrometheusFields()
        promName := p.GetObjectMeta().GetName()
	assetsVolume := v1.Volume{
```

JoaoBraveCoding (commented):
First pass-through, great work!

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -345,6 +355,13 @@ type CommonPrometheusFields struct {
 	HostNetwork bool `json:"hostNetwork,omitempty"`
 	// PodTargetLabels are added to all Pod/ServiceMonitors' podTargetLabels
 	PodTargetLabels []string `json:"podTargetLabels,omitempty"`
+	// Enable access to prometheus web admin API. Defaults to the value of `false`.
+	// WARNING: Enabling the admin APIs enables mutating endpoints, to delete data,
+	// shutdown Prometheus, and more. Enabling this should be done with care and the
+	// user is advised to add additional authentication authorization via a proxy to
+	// ensure only clients authorized to perform these actions can do so.
+	// For more information see https://prometheus.io/docs/prometheus/latest/querying/api/#tsdb-admin-apis
+	EnableAdminAPI bool `json:"enableAdminAPI,omitempty"`
```
simonpasquier, line 364: I tried using the admin API in Prometheus agent mode but it fails with `{"status":"error","errorType":"execution","error":"unavailable with Prometheus Agent"}`. I guess there's no point in moving the field to the common struct.

pkg/prometheus/statefulset.go
```diff
@@ -96,27 +96,41 @@ func prometheusNameByShard(name string, shard int32) string {
...
+	baseImage, tag, sha string,
+	retention monitoringv1.Duration,
+	retentionSize monitoringv1.ByteSize,
+	rules monitoringv1.Rules,
+	query *monitoringv1.QuerySpec,
+	allowOverlappingBlocks bool,
+	queryLogFile string,
+	thanos *monitoringv1.ThanosSpec,
+	disableCompaction bool,
+	p monitoringv1.PrometheusInterface,
```
simonpasquier, line 108: nit: I'd move this parameter right after the `name string` parameter.

pkg/prometheus/statefulset.go
```diff
@@ -96,27 +96,41 @@ func prometheusNameByShard(name string, shard int32) string {
 func makeStatefulSet(
 	logger log.Logger,
 	name string,
-	p monitoringv1.Prometheus,
+	baseImage, tag, sha string,
```
simonpasquier, line 99: (nit) I'd pass a fully constructed image location to the function. It would be up to the caller (e.g. pkg/prometheus/operator.go) to build it as it needs.

pkg/prometheus/statefulset.go
```diff
@@ -96,27 +96,41 @@ func prometheusNameByShard(name string, shard int32) string {
...
 	}
 	intZero := int32(0)
-	if p.Spec.Replicas != nil && *p.Spec.Replicas < 0 {
-		p.Spec.Replicas = &intZero
+	if cpf.Replicas != nil && *cpf.Replicas < 0 {
+		cpf.Replicas = &intZero
 	}
 
-	spec, err := makeStatefulSetSpec(logger, p, config, cg, shard, ruleConfigMapNames, tlsAssetSecrets)
+	p.SetCommonPrometheusFields(cpf)
```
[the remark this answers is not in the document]
simonpasquier: same question for me.

pkg/prometheus/statefulset.go
```diff
@@ -298,7 +312,16 @@ func makeStatefulSetService(p *monitoringv1.Prometheus, config operator.Config)
 
 func makeStatefulSetSpec(
 	logger log.Logger,
-	p monitoringv1.Prometheus,
+	baseImage, tag, sha string,
```
simonpasquier, line 315: same remarks about parameters :)

pkg/prometheus/statefulset.go
simonpasquier: I would defer any logic change to a subsequent PR to minimize the risk of regression.

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+
+// appendServerArgs appends arguments that are only valid to the Prometheus server.
+func appendServerArgs(
+	promArgs *[]monitoringv1.Argument,
+	cg *ConfigGenerator,
+	retention monitoringv1.Duration,
+	retentionSize monitoringv1.ByteSize,
+	rules monitoringv1.Rules,
+	query *monitoringv1.QuerySpec,
+	allowOverlappingBlocks bool) {
```
simonpasquier, line 748: suggestion: it's more idiomatic to use the `x = append(x, ...)` pattern.
```suggestion
func appendServerArgs(
	promArgs []monitoringv1.Argument,
	cg *ConfigGenerator,
	retention monitoringv1.Duration,
	retentionSize monitoringv1.ByteSize,
	rules monitoringv1.Rules,
	query *monitoringv1.QuerySpec,
	allowOverlappingBlocks bool) []monitoringv1.Argument {
```

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+	}
+
+	if cpf.ListenLocal {
+		promArgs = append(promArgs, monitoringv1.Argument{Name: "web.listen-address", Value: "127.0.0.1:9090"})
+	}
+
+	return promArgs
+}
+
+// appendServerArgs appends arguments that are only valid to the Prometheus server.
```
simonpasquier, line 740:
```suggestion
// appendServerArgs appends arguments that are only valid for the Prometheus server.
```

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+		if cpf.Web.PageTitle != nil {
+			promArgs = cg.WithMinimumVersion("2.6.0").AppendCommandlineArgument(promArgs, monitoringv1.Argument{Name: "web.page-title", Value: *cpf.Web.PageTitle})
+		}
+
+		if cpf.Web.MaxConnections != nil {
+			promArgs = append(promArgs, monitoringv1.Argument{Name: "web.max-connections", Value: fmt.Sprintf("%d", *cpf.Web.MaxConnections)})
+		}
+	}
+
+	if cpf.EnableAdminAPI {
```
simonpasquier, line 699: see above

pkg/prometheus/statefulset.go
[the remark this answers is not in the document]
ArthurSens: We make a copy of CommonPrometheusFields when we call:
```go
cpf := p.GetCommonPrometheusFields()
```

Then we make changes to this copy, setting defaults if some fields aren't set:
```go
if cpf.PortName == "" {
	cpf.PortName = defaultPortName
}

if cpf.Replicas == nil {
	cpf.Replicas = &minReplicas
}
intZero := int32(0)
if cpf.Replicas != nil && *cpf.Replicas < 0 {
	cpf.Replicas = &intZero
}
```

Later we call `makeStatefulSetSpec`, but we don't use `cpf` as an argument, we use the same `p` object. Since `cpf` is a copy, the fields we just changed above aren't changed in the `p` object. So that's why `p.SetCommonPrometheusFields(cpf)` is needed 🙂

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+			ReadOnly:  true,
+			MountPath: configmapsDir + c,
+		})
+	}
+
+	return volumes, promVolumeMounts, nil
+}
+
+// addServerVolumes adds volumes to the statefulset spec that are specific to Prometheus Server
+func addServerVolumes(volumes *[]v1.Volume, volumeMounts *[]v1.VolumeMount, queryLogFile string, ruleConfigMapNames []string) {
```
simonpasquier, line 924: same suggestion about the function signature
```suggestion
func appendServerVolumes(volumes []v1.Volume, volumeMounts *[]v1.VolumeMount, queryLogFile string, ruleConfigMapNames []string) []v1.Volume {
```

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+			}
+		}
+	}
+
+	return podAnnotations, podLabels
+}
+
+func configureThanos(
+	additionalContainers *[]v1.Container,
+	disableCompaction *bool,
```
simonpasquier, line 1013: Ideally we should avoid modifying the `disableCompaction` parameter in this function but it can be done later (better to stay close to the current code).

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +676,488 @@ func queryLogFilePath(queryLogFile string) string {
...
+	cpf monitoringv1.CommonPrometheusFields,
+	thanos *monitoringv1.ThanosSpec,
+	c *operator.Config,
+	prometheusURIScheme, webRoutePrefix, volName string) error {
```
simonpasquier, line 1017: suggestion: the caller can append the container to the slice.
```suggestion
func createThanosContainer(
	disableCompaction *bool,
	cpf monitoringv1.CommonPrometheusFields,
	thanos *monitoringv1.ThanosSpec,
	c *operator.Config,
	prometheusURIScheme, webRoutePrefix, volName string) ([]v1.Container, error) {
```

simonpasquier (commented):
Again this is epic! I did a quick pass but this looks very promising :)

pkg/prometheus/statefulset.go
ArthurSens: That's a very interesting question, my guess is that it shouldn't work 🤔

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: I tested starting the Prometheus Agent with the flag, which worked, but I haven't tried to actually use the endpoint. Good thing that you tried 😛

pkg/prometheus/statefulset.go
ArthurSens: Nice, that sounds like a good approach!

pkg/prometheus/statefulset.go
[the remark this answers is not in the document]
JoaoBraveCoding: Then couldn't we just do `cpf := &p.GetCommonPrometheusFields()`? Otherwise, if we want to follow this path I would at least advocate for a comment explaining the set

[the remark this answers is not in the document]
ArthurSens: Unfortunately we can't 😬 
image


Good point on adding a comment 👍

pkg/prometheus/statefulset.go
ArthurSens: We could get rid of volName, but it's needed again when creating the Thanos container.

ArthurSens: Looking at the createThanosContainer signature, we could replace 2 arguments with one `p PrometheusInterface`, maybe it's worth reducing signature for better readability 🙂

pkg/prometheus/statefulset.go
ArthurSens: Actually, this is a bit harder than I thought when I started implementing it. `p.Spec.Image` is a string pointer instead of a single string.

A ton of tests started to fail with nil pointer exceptions and I could find a good approach to make them pass 😅.

Another strategy I tried was changing the signature of `operator.BuildImagePath` to return `*string` instead of `string` but, that requires adding `pointer.String(...)` to a lot of returns... not sure what is more readable 😬

pkg/prometheus/statefulset.go
ArthurSens: Indeed it works as simon described!

I've made the following experiment:

Started a prometheus server with the following command line `./prometheus --config.file=documentation/examples/prometheus.yml --web.enable-remote-write-receiver`, and the config file below:

<details><summary>Prometheus server config</summary>

```yaml
global:
  scrape_interval: 15s
  evaluation_interval: 15s 

scrape_configs:
  - job_name: "prometheus"
    static_configs:
      - targets: ["localhost:9090"]
```

</details>


Next I started an agent, with remote-write receiver enabled `./prometheus --web.listen-address="0.0.0.0:9091" --enable-feature=remote-write-receiver,agent --config.file ../../documentation/examples/prometheus-agent.yml`

<details><summary>Prometheus agent1 config</summary>

```yaml
global:
  scrape_interval: 15s 
  evaluation_interval: 15s
  external_labels:
    prometheus_relay: true

scrape_configs:
  - job_name: "prometheus"
    static_configs:
      - targets: ["localhost:9091"]

remote_write:
  - url: "http://localhost:9090/api/v1/write"
```

</details>

Finally a second agent, with the following command `./prometheus --web.listen-address="0.0.0.0:9092" --enable-feature=agent --config.file ../../documentation/examples/prometheus-agent2.yml --storage.agent.path=data-agent2/`

<details><summary>Prometheus agent2 config</summary>

```yaml
global:
  scrape_interval: 15s 
  evaluation_interval: 15s
  external_labels:
    prometheus_relay: true

scrape_configs:
  - job_name: "prometheus"
    static_configs:
      - targets: ["localhost:9092"]

remote_write:
  - url: "http://localhost:9091/api/v1/write"
```

</details>

When querying the Prometheus server, I can see metrics with both agents' external labels, which proves that the first agent worked as a proxy between the server and the second agent 🙂 
image

pkg/prometheus/statefulset.go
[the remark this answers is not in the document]
simonpasquier: Probably all the default values + validation should be handled at the OpenAPI spec level (like in https://github.com/prometheus-operator/prometheus-operator/pull/5350 for spec.portName). This way we can start with a clean state for the PrometheusAgent CRD and only set the defaults in the Prometheus controller. But like for the previous comment, this is fine for now.

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +680,497 @@ func queryLogFilePath(queryLogFile string) string {
...
+		}
+	}
+
+	if allowOverlappingBlocks {
+		promArgs = cg.WithMinimumVersion("2.11.0").WithMaximumVersion("2.39.0").AppendCommandlineArgument(promArgs, monitoringv1.Argument{Name: "storage.tsdb.allow-overlapping-blocks"})
+	}
+	return promArgs
+}
+
+// AddCommonVolumes returns a set of volumes to be mounted on statefulset spec that are common between Prometheus Server and Agent
```
simonpasquier, line 812:
```suggestion
// buildCommonVolumes returns a set of volumes to be mounted on statefulset spec that are common between Prometheus Server and Agent
```

pkg/prometheus/statefulset.go
```diff
@@ -1076,3 +680,497 @@ func queryLogFilePath(queryLogFile string) string {
...
+func buildCommonVolumes(p monitoringv1.PrometheusInterface, tlsAssetSecrets []string) ([]v1.Volume, []v1.VolumeMount, error) {
+	cpf := p.GetCommonPrometheusFields()
+	promName := p.GetObjectMeta().GetName()
+	volName := volumeName(promName)
```
simonpasquier, line 816: (nit) can you move it next to where it's used (L856)?

simonpasquier (commented):
a few minor comments but otherwise :+1:

pkg/prometheus/statefulset.go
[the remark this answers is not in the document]
ArthurSens: Yeah, that would make a lot more sense!

ArthurSens:
Thanks for all the reviews @JoaoBraveCoding and @simonpasquier.

I went through all the comments and I believe I've addressed everything, but do let me know if I missed anything 🙂

ArthurSens:
Oof, I'm really confused with all this test failures, I'm running them locally and they pass 🤨

simonpasquier:
@ArthurSens can you rebase on top of `main`? It seems that there's a problem with #5360 that merged recently.

ArthurSens:
Indeed that was it :)

pkg/prometheus/statefulset.go
```diff
@@ -565,311 +400,80 @@ func makeStatefulSetSpec(
...
-
-		additionalContainers = append(additionalContainers, container)
-	}
-	if disableCompaction {
-		thanosBlockDuration := operator.StringValOrDefault(string(p.Spec.Thanos.BlockDuration), "2h")
-		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.max-block-duration", Value: thanosBlockDuration})
-		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.min-block-duration", Value: thanosBlockDuration})
-	}
+	if disableCompaction {
+		thanosBlockDuration := operator.StringValOrDefault(string(thanos.BlockDuration), "2h")
```
simonpasquier, line 473: there's a potential nil pointer exception which I didn't catch in #5360 :)
```suggestion
		thanosBlockDuration := "2h"
		if thanos != nil {
			thanosBlockDuration := string(thanos.BlockDuration)
		}
```

simonpasquier (commented):
lgtm minor a small comment

pkg/prometheus/statefulset.go
ArthurSens: good catch

simonpasquier (approved):
:+1:

simonpasquier:
````
2023-02-23T14:01:05.0255127Z     statefulset_test.go:1160: Prometheus is missing expected argument: --storage.tsdb.max-block-duration=2h
2023-02-23T14:01:05.0255553Z --- FAIL: TestThanosObjectStorage (0.00s)
2023-02-23T14:01:05.0255819Z === RUN   TestThanosObjectStorageFile
2023-02-23T14:01:05.0256413Z     statefulset_test.go:1235: Prometheus is missing expected argument: --storage.tsdb.max-block-duration=2h
2023-02-23T14:01:05.0256842Z --- FAIL: TestThanosObjectStorageFile (0.00s)
````

The unit test is failing because the blockDuration is empty but in the "real" life, the spec API has a default value of "2h" so empty block duration shouldn't arise.

ArthurSens:
> TestThanosObjectStorageFile

Not sure if I understood it correctly, but I've changed the input to make them pass.

Do I understand it correctly that those adding the extra input sound I bit like we're faking the behavior that API spec would have in a real scenario? I.e. a Prometheus instance without `thanos.BlockDuration` set, should have the blockduration argument set anyway.

JoaoBraveCoding:
> Not sure if I understood it correctly, but I've changed the input to make them pass.
> 
> Do I understand it correctly that those adding the extra input sound I bit like we're faking the behavior that API spec would have in a real scenario? I.e. a Prometheus instance without `thanos.BlockDuration` set, should have the blockduration argument set anyway.

Yup that's what I understood as well, I think you did the correct thing

JoaoBraveCoding (approved):
lgtm, great work @ArthurSens 💪

simonpasquier (commented):
Sorry for the wrong advice but I realized that if the CRDs aren't updated to the last version (e.g. with the `spec.thanos.blockDuration` field), it would generate a bad arguments list. In other words, we still need a fallback default in the Go code.

simonpasquier (approved):
I'll merge on green CI. Prior implementing the agent controller, you might want to reorganize the pkg/prometheus directory?
My suggestion would be
* pkg/prometheus :arrow_right: code to be shared between agent & server (pkg/prometheus/promcfg*.go and pkg/prometheus/statefulset*.go).
* pkg/prometheus/server :arrow_right: code specific to server (e.g. pkg/prometheus/operator*.go and pkg/prometheus/rules*.go).
As we move forward with the agent implementation, we'll probably delineate things more precisely...

PR #5340, opened 2023-02-14
Set Read/ReadHeaderTimeouts on http.Server

mrueg:
## Description
via gosec linter - G112: Potential slowloris attack

https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6

## Type of change

## Changelog entry

```release-note
Set Read/ReadHeaderTimeouts on http.Server instances
```

### DIFF
--- a/cmd/admission-webhook/main.go
+++ b/cmd/admission-webhook/main.go
@@ -166,8 +166,10 @@ func newSrv(logger log.Logger, tlsConf *tls.Config) *srv {
 	return &srv{
 		logger: logger,
 		s: &http.Server{
-			Handler:   mux,
-			TLSConfig: tlsConf,
+			Handler:           mux,
+			TLSConfig:         tlsConf,
+			ReadHeaderTimeout: 30 * time.Second,
+			ReadTimeout:       30 * time.Second,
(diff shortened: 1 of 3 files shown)


JoaoBraveCoding (approved):
TIL, that's neat! Thanks for the contribution!

PhilipGough:
I am not convinced about this change if I am honest, particularly the hardcoded nature of the value. Perhaps making them configurable might make sense but my gut feeling tells me that this could/should be mitigated at ingress/proxy layer if required.

simonpasquier (commented):
From what I can tell, the timeout for webhook configurations has to be between 1 second and 30 seconds, with a default of 10s for v1 ([ref](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#validatingwebhook-v1-admissionregistration-k8s-io)). I'd be ok with a hardcoded timeout value of 30s so we don't introduce a potentially breaking change.
In addition we should evaluate if there's value in enabling the `gosec` linter. My first impression is that it's a bit noisy but I haven't looked very deep.

PhilipGough (approved):
lgtm

simonpasquier:
thanks!

PR #5371, opened 2023-02-24
Prepare package organization for PrometheusAgent

ArthurSens:
## Description

This PR only splits the `pkg/prometheus` into 2. Where code that will be shared between prometheus servers and agents remains on the original package, what what is specific to the server moves to the new package.

We expect a new package for agent-specific code to live nearby once it gets implemented

## Type of change

## Changelog entry

```release-note

```


ArthurSens:
I guess we could move some methods created in #5334 to the new package already?

e.g. `createThanosContainer`, `appendServerVolumes` and `appendServerArgs`

simonpasquier:
> I guess we could move some methods created in #5334 to the new package already?
> 
> e.g. `createThanosContainer`, `appendServerVolumes` and `appendServerArgs`

sounds good to me

ArthurSens:
Hmmm, actually I just moved and got errors regarding `go import cycle`.

> e.g. createThanosContainer, appendServerVolumes and appendServerArgs

The server's operator calls the original package for several operations. If I move only the methods mentioned above, the `makeStatefulSet` function cannot call methods moved to the server package because one package calls the other.


---

What would be your preferred way forward? From what I see we have two options:

* Leave server-specific code in the original package, making this PR minimal just to introduce the new package

or

* Move almost all of `statefulset.go` code to the new package. The server operator will call makeStatefulSet from its own package, and whatever can be shared with the agent will be called from the original package.



I'm fine with both, I'm just saying because I originally thought this was a small PR and just realize it might not 😅

simonpasquier:
I'd go with option 2 ("Move almost all of statefulset.go code to the new package").

ArthurSens:
Moved :)

simonpasquier:
The CI failure is https://github.com/prometheus-operator/prometheus-operator/issues/5375

pkg/prometheus/promcfg.go
```diff
@@ -53,7 +53,7 @@ func sanitizeLabelName(name string) string {
 // compatible with a given Prometheus version.
 type ConfigGenerator struct {
 	logger                 log.Logger
-	version                semver.Version
+	Version                semver.Version
```
simonpasquier, line 56: I'd rather keep it unexported and update the code that rely on it. In practice,

https://github.com/prometheus-operator/prometheus-operator/blob/ef5fad77131af9f162b4b13234446e721d74818d/pkg/prometheus/statefulset.go#L352-L354

This check should be done earlier and/or differently:

```golang
	// Anything before v2.0.0-alpha.0 isn't supported by the operator.
	if cg := cg.WithMinimumVersion("v2.0.0-alpha.0"); !cg.IsCompatible() {
		return nil, errors.Errorf("unsupported Prometheus version %s", cg.Version())
	}

// pkg/prometheus/promcfg.go
func (cg *ConfigGenerator) Version() string {
	return cg.version.String()
}
```

pkg/prometheus/collector.go
```diff
@@ -47,32 +47,32 @@ var (
...
 )
 
-type prometheusCollector struct {
+type Collector struct {
 	stores []cache.Store
 }
 
-func newPrometheusCollectorForStores(s ...cache.Store) *prometheusCollector {
-	return &prometheusCollector{stores: s}
+func NewPrometheusCollectorForStores(s ...cache.Store) *Collector {
```
simonpasquier, line 54:
```suggestion
func NewCollectorForStores(s ...cache.Store) *Collector {
```

pkg/prometheus/statefulset.go
```diff
@@ -21,55 +21,49 @@ import (
...
 	managedByOperatorLabel            = "managed-by"
 	managedByOperatorLabelValue       = "prometheus-operator"
-	managedByOperatorLabels           = map[string]string{
+	ManagedByOperatorLabels           = map[string]string{
 		managedByOperatorLabel: managedByOperatorLabelValue,
 	}
-	shardLabelName                = "operator.prometheus.io/shard"
-	prometheusNameLabelName       = "operator.prometheus.io/name"
-	probeTimeoutSeconds     int32 = 3
+	ShardLabelName                = "operator.prometheus.io/shard"
```
simonpasquier, line 61: this could be moved to pkg/prometheus/server/statefultset.go as it's only used there?

pkg/prometheus/statefulset.go
```diff
@@ -21,55 +21,49 @@ import (
...
+	PrometheusNameLabelName       = "operator.prometheus.io/name"
```
simonpasquier, line 62: same

pkg/prometheus/statefulset.go
```diff
@@ -21,55 +21,49 @@ import (
...
+	ProbeTimeoutSeconds     int32 = 3
```
simonpasquier, line 63: same

pkg/prometheus/statefulset.go
```diff
@@ -21,55 +21,49 @@ import (
...
-	secretsDir                      = "/etc/prometheus/secrets/"
-	configmapsDir                   = "/etc/prometheus/configmaps/"
-	configFilename                  = "prometheus.yaml.gz"
-	configEnvsubstFilename          = "prometheus.env.yaml"
-	sSetInputHashName               = "prometheus-operator-input-hash"
-	defaultPortName                 = "web"
-	defaultQueryLogDirectory        = "/var/log/prometheus"
-	defaultQueryLogVolume           = "query-log-file"
+	//TODO: RulesDir should be moved to the server package, since it is not used by the agent.
+	// It is here at the moment because promcfg uses it, and moving as is will cause import cycle error.
```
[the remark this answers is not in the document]
simonpasquier: I think that we reorganize the code to remove the dependency but it can be in another PR.

pkg/prometheus/statefulset.go
```diff
@@ -21,55 +21,49 @@ import (
...
+	RulesDir                 = "/etc/prometheus/rules"
+	secretsDir               = "/etc/prometheus/secrets/"
+	configmapsDir            = "/etc/prometheus/configmaps/"
+	ConfigFilename           = "prometheus.yaml.gz"
+	ConfigEnvsubstFilename   = "prometheus.env.yaml"
+	SSetInputHashName        = "prometheus-operator-input-hash"
```
simonpasquier, line 48: this could be moved to pkg/prometheus/server/statefultset.go as it's only used there?

simonpasquier (commented):
Maybe rename `(cg *ConfigGenerator) Generate(...)` to `(cg *ConfigGenerator) GenerateServerConfiguration(...)` to make it more obvious that it generates the server config? Otherwise just a few nits and I agree that we can adjust the details as we move forward with the agent implementation :smiley:

pkg/prometheus/promcfg.go
JoaoBraveCoding: Agree we should try to keep the `version` field unexported

JoaoBraveCoding (approved):
From my side its mostly lgtm, once we have the agent part it will be easier to clearly see what belongs where

pkg/prometheus/promcfg.go
ArthurSens: Looks good!

Small nit that it should return semver.Version because we need to compare `Major` :P

pkg/prometheus/statefulset.go
ArthurSens: Yeah, but it will be used by the agent too, won't it? I'm not entirely familiar with the several agent deployment models, but I'm assuming HA is one of then 🤔

pkg/prometheus/statefulset.go
ArthurSens: Same as above

pkg/prometheus/server/statefulset.go
```diff
@@ -0,0 +1,771 @@
...
+		operator.StringValOrDefault(tag, ""),
+		operator.StringValOrDefault(sha, ""),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if cg.Version().Major != 2 {
+		return nil, errors.Errorf("unsupported Prometheus major version %s", cg.Version())
+	}
```
JoaoBraveCoding, line 218: Can't we move this `if` to `NewConfigGenerator`? From my POV the promcfg package should be responsible for these types of checks

ArthurSens: That makes perfect sense!

pkg/prometheus/statefulset.go
simonpasquier: right, let's leave it there for now.

simonpasquier (approved):

PR #5383, opened 2023-03-01
Add SlackAPIURL to Alertmanager global configuration

mcbenjemaa:
## Description

Configuring receivers with a lot of slack configs is cumbersome.
Having a global field that defines the `slackAPIURL` is much easier, 

This  will Add `spec.alertmanagerConfiguration.global.slackApiUrl` 

## Type of change

## Changelog entry

```release-note
Add `spec.alertmanagerConfiguration.global.slackApiUrl` to the `Alertmanager` CRD.
```


pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -273,6 +273,9 @@ type AlertmanagerGlobalConfig struct {
 
 	// HTTP client configuration.
 	HTTPConfig *HTTPConfig `json:"httpConfig,omitempty"`
+
+	// Slack Api Url
+	SlackAPIURL *string `json:"slackApiUrl,omitempty"`
```
simonpasquier, line 278: This needs to be a `*SecretKeySelector` type because the Slack API URL can't be stored in clear-text.

mcbenjemaa: Alright, I was a bit confused tho...

mcbenjemaa: I will change it..

mcbenjemaa:
Ready

simonpasquier (commented):
Changes to go.sum files need to be reverted. Ideally an additional sub-test in `TestInitializeFromAlertmanagerConfig()` would be welcome but it feels too much work for you, I can add it later.

mcbenjemaa:
let me give the test a try...

mcbenjemaa:
I added a test case, would you check?

simonpasquier (approved):

simonpasquier:
Thanks! If you have some free cycles, feel free to submit PRs to implement the other global parameters that are still missing ;-)

mcbenjemaa:
Will do)

PR #5392, opened 2023-03-03
Introduce ResourceSelector interface to be used by diferent operators

ArthurSens:
## Description

Introduce a ResourceSelector interface to be used by different operators, e.g. Prometheus Server and the future Prometheus Agent

## Type of change

## Changelog entry

```release-note

```


pkg/prometheus/operator.go
```diff
@@ -0,0 +1,591 @@
...
+type OperatorInterface interface {
+	GetLogger() log.Logger
+	GetServiceMonitorInformers() *informers.ForResource
+	GetPodMonitorInformers() *informers.ForResource
+	GetProbeInformers() *informers.ForResource
+	GetNamespaceInformers() cache.SharedIndexInformer
+	GetMetrics() *operator.Metrics
+}
+
+func SelectServiceMonitors(ctx context.Context, po OperatorInterface, p monitoringv1.PrometheusInterface, store *assets.Store) (map[string]*monitoringv1.ServiceMonitor, error) {
```
simonpasquier, line 52: I would create a struct and turn the Select* functions into receiver's methods.

````golang
type ResourceSelector struct {
    l log.Logger
    store *assets.Store
    podMonitorLister ListAllByNamespacer
    serviceMonitorLister ListAllByNamespacer
    ...
}

type ListAllByNamespacer interface {
    ListAllByNamespace(namespace string, selector labels.Selector, appendFn cache.AppendFunc) error
}

func NewResourceSelector(l log.Logger, p monitoringv1.PrometheusInterface, store *assets.Store, podMonitorLister ListAllByNamespacer, ...) *ResourceSelector {
    ...
}

func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context) (map[string]*monitoringv1.ServiceMonitor, error) {
    ...
}
````

Comment (ArthurSens) [reply pkg/prometheus/operator.go:52]: Nice idea!

Could you clarify what is the purpose of the interface `ListAllByNamespace`, not sure if I'm lacking context, but it is not obviously clear what it is used for 🤔

File: pkg/prometheus/resource_selector.go [hunk 2]
```diff
@@ -0,0 +1,591 @@
...
+type ResourceSelector struct {
+	l                    log.Logger
+	p                    monitoringv1.PrometheusInterface
+	store                *assets.Store
+	podMonitorLister     ListAllByNamespacer
+	serviceMonitorLister ListAllByNamespacer
+	probeLister          ListAllByNamespacer
+	namespaceInformers   cache.SharedIndexInformer
+	metrics              *operator.Metrics
+	accessor             *operator.Accessor
```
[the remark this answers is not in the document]
ArthurSens: Hmmmm, fews kinda strange that we have a logger in the resourceSelector and another in the accessor 🤔 

I feel like I missed something... the logger in the accessor should have been exported?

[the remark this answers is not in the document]
simonpasquier: I think that we can omit the `accessor` parameter as it can be instantiated in the ResourceSelector constructor.

pkg/prometheus/resource_selector.go
```diff
@@ -0,0 +1,591 @@
...
+		rs.metrics.SetSelectedResources(pKey, monitoringv1.ServiceMonitorsKind, len(res))
+		rs.metrics.SetRejectedResources(pKey, monitoringv1.ServiceMonitorsKind, rejected)
+	}
+
+	return res, nil
+}
+
+// ListMatchingNamespaces lists all the namespaces that match the provided
+// selector.
+func ListMatchingNamespaces(selector labels.Selector, nsMonInf cache.SharedIndexInformer) ([]string, error) {
```
simonpasquier, line 208: (nit)
```suggestion
func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInformer) ([]string, error) {
```

pkg/thanos/rules.go
```diff
@@ -148,7 +149,7 @@ func (o *Operator) selectRuleNamespaces(p *monitoringv1.ThanosRuler) ([]string,
 			return namespaces, errors.Wrap(err, "convert rule namespace label selector to selector")
 		}
 
-		namespaces, err = o.listMatchingNamespaces(ruleNamespaceSelector)
+		namespaces, err = prompkg.ListMatchingNamespaces(ruleNamespaceSelector, o.nsRuleInf)
```
simonpasquier, line 152: to avoid a dependency from pkg/thanos to pkg/prometheus, I'd move `ListMatchingNamespaces()` to pkg/operator.

pkg/prometheus/resource_selector.go
```diff
@@ -0,0 +1,577 @@
...
+		podMonitorLister:     podMonitorLister,
+		serviceMonitorLister: serviceMonitorLister,
+		probeLister:          probeLister,
+		namespaceInformers:   namespaceInformers,
+		metrics:              metrics,
+		accessor:             operator.NewAccessor(l),
+	}
+}
+
+func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context) (map[string]*monitoringv1.ServiceMonitor, error) {
```
JoaoBraveCoding, line 70: nit: should we add a method comment outlining the general strategy for selecting the monitor? This method is so large that I feel it would help

JoaoBraveCoding (commented):
Small optional nit, but apart from that it's awesome seeing it all come together

pkg/prometheus/resource_selector.go
ArthurSens: Could you elaborate on how this comment would look like?

Personally, I don't think a comment explaining a function step-by-step helps to make things simple. If the code changes, we often forget about the comment, which gets outdated.

JoaoBraveCoding: I was thinking of something to just allude to people to what is being done it doesn't need to be very descriptive. I'm not 100% sure about the last phrase.

```suggestion
// SelectServiceMonitors selects ServiceMonitors based on the selectors in the Prometheus CR and filters them
// returning only those with a valid configuration. This function also populates authentication stores.
func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context) (map[string]*monitoringv1.ServiceMonitor, error) {
```

ArthurSens: I feel like breaking this method into two separate ones would make more sense, making sure they do what they are named after 😅

With that said, I think this goes beyond what this PR is intended to change, so I'll just add the comment for now :)

JoaoBraveCoding: Yeah, I feel the same, but that's precisely why such a comment from my POV makes the difference here. Just by reading the name of the method, a developer will most probably think this method just selects ServiceMonitors when in reality it does at least 3 things. Select them, filter them and adds data to store.

I'm still not 100% sold that this should be separated, I would have to see how it turns out, but I agree that regardless this should be done in a follow-up

JoaoBraveCoding (approved):
lgtm :shipit:

simonpasquier (approved):
:partying_face:

PR #5404, opened 2023-03-09
updateResourceVersion

yccharles:
## Description

Add 'resourceVersion' to ListOptions when get/list resource  througth client-go. 
Apiserver return data by using cache  instead of etcd server.
It's help improve response time and reduce stress of etcd server when node number is more than 1000  in k8s cluster.

## Type of change

## Changelog entry

```
Add 'resourceVersion=0' to ListOptions when get/list resource througth client-go.
```

### DIFF
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -614,7 +614,7 @@ func (c *Operator) syncNodeEndpoints(ctx context.Context) error {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
Thanks for the PR! IIUC the performance issue is mainly when listing nodes?
I'd recommend to update `syncNodeEndpoints()` first and cache the latest-known resource version in the Operator struct.

PR #5405, opened 2023-03-09
docs: Update steps on running operator outside cluster

slashpai:
## Description

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,11 @@ kubectl create -f bundle.yaml
 To run the Operator outside of a cluster:
 
 ```sh
-make
+kubectl create -f example/prometheus-operator-crd-full
+```
+
+```sh
+make operator
 scripts/run-external.sh <kubectl cluster name>
 ```
 


README.md
```diff
@@ -116,7 +116,11 @@ kubectl create -f bundle.yaml
 To run the Operator outside of a cluster:
 
 ```sh
-make
+kubectl create -f example/prometheus-operator-crd-full
+```
+
+```sh
```
simonpasquier, line 122: The command at L111 would already create/update the CRDs and also apply the needed RBAC objects.

```suggestion
```

sthaha: How about write this section as .. 

```
kubectl delete -f bundle.yaml   # to remove already running PO 

kubectl create -f example/prometheus-operator-crd-full
make operator
./scripts/run-external.sh
```

sthaha: I know this is beyond the scope of this change but .. since user has to run `make operator` every single tinme before running the script, how about we add that to the script itself? 
In fact all of these can be incorportated  into the script so that to run, PO, all someone has to do is to run the script.

sthaha: Here is an example -  https://github.com/prometheus-operator/prometheus-operator/pull/5415 which I am using locally to run operator out of cluster.

simonpasquier: Thinking more about it, I'd move all developer instructions to CONTRIBUTING.md and have only user-oriented content in the README.

slashpai:
Closing in favour of https://github.com/prometheus-operator/prometheus-operator/pull/5415

PR #5441, opened 2023-03-23
Configure Prebuilds and extension installation for Gitpod workspaces

ArthurSens:
## Description

I've been using Gitpod to develop changes to the Operator for quite a while, and every time I open a new workspace I have to manually install the extensions 😅. I decided that it might be best to just properly configure the project so workspaces have prebuilds and extensions installed whenever someone opens them :)

In the Gitpod YAML file, I've configured [Gitpod Prebuilds](https://www.gitpod.io/docs/configure/projects/prebuilds) to only build the operator binary and image and I've enabled that pull requests from forks also get prebuilds. The full configuration surface can be seen [here](https://www.gitpod.io/docs/references/gitpod-yml).

To open a gitpod workspace, there are a couple of ways:

1. Prefix the repo/branch/issue/PR URL with https://gitpod.io/#
2. Download [gitpod browser extension](https://www.gitpod.io/docs/configure/user-settings/browser-extension), which injects a button to Github to open the workspace. A shortcut can also be enable to open the workspace by pressing `.`
3. We could add a button to the README like the one below. (I'm not really a fan)

[Open in Gitpod](https://gitpod.io/#https://github.com/prometheus-operator/prometheus-operator)

The extension is by far my preferred way 

## Type of change

## Changelog entry

```release-note

```


.gitpod.yml
JoaoBraveCoding, line 1: Si IIUC this will make two things:

1. If someone uses GitPod it will have the `extensions` automatically installed in VSCode
2. If we configure prebuilds for the project, then all PR will automatically run `make build image` if the user uses GitPod to look at the PR

Right?

Personally, regarding 1. I'm fine with it, regarding 2. I don't know enough about GitPod to comment on it... are there any costs associated? How many people have to actually build the image when they open a PR?

ArthurSens: Regarding 2)

A gitpod workspace is a container image that can be attached to an IDE, Gitpod manages the connection between the container and the IDE of choice

A prebuild is the process of building this container when there is a push event to a remote fit repository. Gitpod will build the container image in advance so when the reviewer open the workspace, the operator binary and image are already built, saving a bit of time :)

As far as I know Gitpod is free for open source, but I can check it with them again.

JoaoBraveCoding: Sounds good to me, I don't have any objections :)

.gitpod.yml
```diff
@@ -0,0 +1,16 @@
...
+tasks:
+  - init: make build image
+github:
+  prebuilds:
+    pullRequestsFromForks: true
+vscode:
+  extensions:
+    - "golang.go"
+    - "GitHub.copilot"
+    - "Grafana.vscode-jsonnet"
```
sthaha, line 16: I am not a vscode user, so pardon these stupid questions :) 

Our of curiosity, what happens if someone without copilot opens the pod? 
Also why is "Grafana.vcode-jsonnet" needed? 

I am also not too sure if we should include editor specific configs (maintenance cost) in a repo but I don't have a strong opinions as well :)

ArthurSens: > I am not a vscode user, so pardon these stupid questions :)

No worries! In case you're using Jetbrain's IDEs, I believe it is possible to automate the installation of plugins (I think they the same as vscode extensions)

https://www.gitpod.io/docs/references/gitpod-yml#jetbrainsproductplugins

> Our of curiosity, what happens if someone without copilot opens the pod?

That's a very good question, I'm not 100% sure, but I believe it will ask for authentication and will just do nothing if access isn't granted

>  Also why is "Grafana.vcode-jsonnet" needed?

We do offer a [jsonnet library](https://github.com/prometheus-operator/prometheus-operator/tree/main/jsonnet) for downstream projects, e.g. kube-prometheus. For sure development of jsonnet is much less frequent than go, but it doesn't hurt to let this extension enabled for when it is needed :)

> I am also not too sure if we should include editor specific configs (maintenance cost) in a repo but I don't have a strong opinions as well :)

That's a fair point, maintaining those lines of code will only benefit Gitpod users. So far, I believe it is only me 😅 but I do invite everyone else to try it out. Once I got used to the workflow, I have never used local development again :P

I don't want to force anything though, it doesn't hurt me much to install the extensions every time I open a workspace and I'm far from being the most frequent reviewer. If people don't use Gitpod, then prebuilds won't save anyone's time

ArthurSens:
Thinking again about the usage here, Gitpod supports running containers but doesnt support kubernetes.

I usually use codespaces when I need k8s, so maybe this PR isnt useful afterall

Issue #5411, opened 2023-03-14
Fix image links in documentation to display correctly on both prometheus-operator.dev and GitHub


simonpasquier:
Though `Documentation/troubleshooting.md` has a [comment](https://github.com/prometheus-operator/prometheus-operator/blob/e2ce57ad6c383a886a7d5fb518a677e1dd6a6115/Documentation/troubleshooting.md?plain=1#L52-L54) saying that the image should render properly at https://prometheus-operator.dev/docs/operator/troubleshooting/, we keep receiving pull requests that want to "fix" the image being broken when viewing the Markdown file from GitHub (https://github.com/prometheus-operator/prometheus-operator/pull/5259 and https://github.com/prometheus-operator/prometheus-operator/pull/5406).

We should update the documentation's file layout to avoid the confusion and have the image displaying properly for both websites.

PR #5459, opened 2023-03-28
fix(docs): use external images to fix display on github

dotdc:
## Description

This is a way to fix the broken images while browsing documentation on GitHub directly whithout breaking them on https://prometheus-operator.dev.

In order to work, the org owner will need to create an `assets` repository and move the `img` folder and it's content to it. 

New repo: `https://github.com/prometheus-operator/assets`

This will fix:
- https://github.com/prometheus-operator/prometheus-operator/issues/5411
- https://github.com/prometheus-operator/prometheus-operator/pull/5326

Tradeoff: image will not be available from a local copy without internet access, but this should be pretty acceptable these days !

Let me know if you want to adapt repository name for the assets, or anything else!

## Type of change

## Changelog entry

```release-note
- Fix: use external images to fix display on GitHub documentation
```

### DIFF
--- a/Documentation/troubleshooting.md
+++ b/Documentation/troubleshooting.md
@@ -51,7 +51,7 @@ A common problem related to `ServiceMonitor` identification by Prometheus is rel
 
 <!-- do not change this link without verifying that the image will display correctly on https://prometheus-operator.dev -->
 
-![flow diagram](/img/custom-metrics-elements.png)
+![flow diagram](https://raw.githubusercontent.com/prometheus-operator/assets/main/img/custom-metrics-elements.png)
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
I'd rather avoid having a separate repository for assets... Having said that, I've got no ready-to-go solution to fix the issue.

dotdc:
Having a separate repo for these kind of assets is a good way to solve the issue. It's also a good way to free the other repositories from the weight, even if it's not huge in this case...

There is probably other way to solve the problem like:
- rework the repo structure, but this may require more work...
- duplicate the img folder, but... 😅
- ...

Anyway this was just an idea, feel free to close.

simonpasquier:
@dotdc sorry if my remark sounded like I dismissed your contribution, it wasn't the case at all!

> rework the repo structure, but this may require more work...

This is what I'd recommend exploring but I agree that it can be challenging.

dotdc:
> @dotdc sorry if my remark sounded like I dismissed your contribution, it wasn't the case at all!
> 
> > rework the repo structure, but this may require more work...
> 
> This is what I'd recommend exploring but I agree that it can be challenging.

Don't worry, your remark was not offensive in any way 😉
Like I said, this was just an idea...

PR #5464, opened 2023-03-29
Remove unnecessary carriage return

boatrainlsz:
## Description

Remove unnecessary carriage return in api.md

## Type of change

## Changelog entry

```release-note
Remove unnecessary carriage return in Description column with Field apiVersion
```

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -2652,10 +2652,7 @@ PrometheusRuleSpec
 <code>apiVersion</code><br/>
 string</td>
 <td>
-<code>
-monitoring.coreos.com/v1
-</code>
-</td>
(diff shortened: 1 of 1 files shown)


boatrainlsz:
> Thank you for the contribution but this page is automatically generated, even if we were to accept this it would just be overwritten once we ran the doc generation tool.

Thanks for you reply, is there any approach to resolve this? maybe more info about doc generation tool would help, I would love to do research.

simonpasquier (commented):
the template responsible for this snippet is https://github.com/prometheus-operator/prometheus-operator/blob/aa8222d7e9b66e9293ed11c9291ea70173021029/scripts/docs/templates/type.tpl#L62-L66

Issue #5456, opened 2023-03-28
Test resource name colision between PrometheusAgent and server


ArthurSens:
one good test might be to create a prometheus object and a prometheusagent object with the same name in the same namespace and verify that the 2 controllers don't fight for the same resources :)

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/5385#pullrequestreview-1348479023_

PR #5471, opened 2023-03-30
Test PrometheusAgent creation and deletion

ArthurSens:
## Description

Adds a simple e2e test that creates and deletes a PrometheusAgent resource.

Opening it as a draft because, even though tests pass and the code looks correct, I've noticed that the test doesn't behave as expected. The PrometheusAgent resource isn't reconciled for some reason and the test pass anyway

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/test/framework/prometheusagent.go
+++ b/test/framework/prometheusagent.go
@@ -90,7 +90,7 @@ func (f *Framework) WaitForPrometheusAgentReady(ctx context.Context, p *monitori
 			return resourceStatus{
 				expectedReplicas: expected,
 				generation:       current.Generation,
-				replicas:         current.Status.AvailableReplicas,
+				replicas:         current.Status.UpdatedReplicas,
 				conditions:       current.Status.Conditions,
 			}, nil
 		},
--- a/test/framework/prometheusagent.go
+++ b/test/framework/prometheusagent.go
@@ -1,0 +1,129 @@
+// Copyright 2023 The prometheus-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
+	"github.com/prometheus-operator/prometheus-operator/pkg/prometheus/agent"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas int32) *monitoringv1alpha1.PrometheusAgent {
+	return &monitoringv1alpha1.PrometheusAgent{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Namespace:   ns,
+			Annotations: map[string]string{},
+		},
+		Spec: monitoringv1alpha1.PrometheusAgentSpec{
+			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
+				Replicas: &replicas,
+				Version:  operator.DefaultPrometheusVersion,
+				ServiceMonitorSelector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"group": group,
+					},
+				},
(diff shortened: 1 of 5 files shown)


ArthurSens:
I'm actually struggling really hard on this one 😓 

In this last 3 weeks I've continuously tried to debug why this is happening and I'm still unable to understand why PrometheusAgent resources aren't being reconciled :/

JoaoBraveCoding:
Sorry @ArthurSens I haven't had free time to be able to look at this 😞 I'll try to see if this week I have some extra time if this is not solved by then

test/framework/prometheusagent.go
```diff
@@ -0,0 +1,129 @@
...
+				conditions:       current.Status.Conditions,
+			}, nil
+		},
+		timeout,
+	); err != nil {
+		return errors.Wrapf(err, "prometheus-agent %v/%v failed to become available", p.Namespace, p.Name)
+	}
+
+	return nil
+}
```
ArthurSens, line 103: Overall, the main problem is that I can't understand why Prometheus-Operator is not reconciling `PrometheusAgent` resources.

Also, there is a bug in this function that I couldn't understand yet, but it returns no errors even if the PrometheusAgent resource never got reconciled. Since it was a copy past from the Prometheus test pkg, the bug is probably there as well

test/framework/prometheusagent.go
```diff
@@ -0,0 +1,129 @@
...
+		ctx,
+		func(context.Context) (resourceStatus, error) {
+			current, err := f.MonClientV1alpha1.PrometheusAgents(p.Namespace).Get(ctx, p.Name, metav1.GetOptions{})
+			if err != nil {
+				return resourceStatus{}, err
+			}
+			return resourceStatus{
+				expectedReplicas: expected,
+				generation:       current.Generation,
+				replicas:         current.Status.AvailableReplicas,
```
ArthurSens, line 93: the `WaitForPrometheusAgentReady` fuction was copy pasted from the Prometheus test framework and this line specifically was setting `replicas: current.StatusUpdatedReplicas` instead of `AvailableReplicas`.

A month later and I finally realized that UpdatedReplicas = 1 doesn't mean that prometheus-agent is ready and we should be looking at AvailableReplicas instead 😅

ArthurSens (commented):
Ok, this PR took a lot longer than I wanted it to 😅.

I was facing an issue where `CreatePrometheusAgentAndWaitUntilReady` would create the resource, but would not wait for the Agent to be available even without errors. The following line would call `DeletePrometheusAgentAndWaitUntilGone`, and it passes instantly as well because the Agent pod was never created.

The test was passing, even though 0 Pods were ever created. Today I finally found the culprit 🥳, but it makes me wonder if the Prometheus test framework should be considered bogus or not 🤔

test/framework/prometheusagent.go
ArthurSens: After a few more hours of debugging, I've found out that https://github.com/prometheus-operator/prometheus-operator/blob/906669420e97573a4ea7534fb3fdf0ce8192ae03/test/framework/status.go#L50-L64 is where the bug lives

image

Notice in the image above how the "available" variable is of type "reconciled" instead of available. It is set to `true`, even though kubectl is showing false for the `AVAILABLE` field.

I'm still trying to understand why this is happening, but I suspect it has something to do with the pointer assignments

ArthurSens: Found it 🥳 That was a really hard one, what the hell

The problem was that we assign the address of `cond` to the variable `available`. On the second loop, the address of `cond` changes, and so does the value of `available`.

I've pushed a fix in 53c708ab8f10ca78b65fe689ac732a38f2f37492, making sure we're not changing values of `available` and `reconciled` on every loop

ArthurSens: Gosh, that was a month and a half of headaches 😅

JoaoBraveCoding: OK! I think I got it! So in the second loop if the resource is already reconciled this function will return `True` because both `available` and `reconciled` will be pointing to the conditions `reconciled` which is already `true`. Damn, that's indeed a thought one! Great job Arthur!

test/framework/status.go
```diff
@@ -48,6 +48,9 @@ func (f *Framework) WaitForResourceAvailable(ctx context.Context, getResourceSta
 
 		var reconciled, available *monitoringv1.Condition
 		for _, cond := range status.conditions {
+			// We need to create a new address for 'cond' inside the loop, otherwise we change the value of
+			// 'available' and 'reconciled' will have their pointer changed on every loop iteration.
```
JoaoBraveCoding, line 52: If you feel like it helps to clarify!
```suggestion
			// 'available' and 'reconciled' will have their pointer changed on every loop iteration.
			// https://medium.com/swlh/use-pointer-of-for-range-loop-variable-in-go-3d3481f7ffc9
```

JoaoBraveCoding (approved):
lgtm! Mega work detective @ArthurSens 🕵️

test/framework/prometheusagent.go
xiu: Thanks for that, @ArthurSens! I actually had the same issue trying to create some for ScrapeConfig and didn't understand why it wasn't waiting for the pods to be ready.

test/framework/status.go
ArthurSens: It sure does :)

ArthurSens:
This PR was more to understand why creating a PrometheusAgent resource and waiting for reconciliation had such odd behavior.

After getting this one in, I'll refactor the test to finally solve https://github.com/prometheus-operator/prometheus-operator/issues/5456. Testing name collision will cover Agent creation and deletion

simonpasquier (approved):
lgtm! Sorry for introducing the bug in the first place :-/

ArthurSens:
> lgtm! Sorry for introducing the bug in the first place :-/

Happens to all of us!

PR #5472, opened 2023-03-30
prometheus/agent/operator: Log when missing permissions

### LINKED ISSUE #5457: Log missing permissions if PrometheusAgent controller doesn't start
it would be neat to log exactly which permission(s) is(are) missing but I'll defer it to a follow-up.

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/5385#discussion_r1142091009_

ArthurSens:
## Description

Fix #5457 

## Type of change

## Changelog entry

```release-note
Log missing permissions if PrometheusAgent controller doesn't start due to lack of authorization
```

### DIFF
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -292,12 +292,12 @@ func (c *Operator) Run(ctx context.Context) error {
 		return nil
 	}
 
-	allowed, err := c.checkAgentAuthorization(ctx)
+	missingPermissions, err := c.getMissingPermissions(ctx)
 	if err != nil {
 		return err
 	}
-	if !allowed {
-		level.Info(c.logger).Log("msg", "Prometheus agent controller disabled because it lacks the required permissions on PrometheusAgent objects")
+	if len(missingPermissions) > 0 {
+		level.Warn(c.logger).Log("msg", "Prometheus agent controller disabled because it lacks the required permissions on PrometheusAgent objects", "missingpermissions", fmt.Sprintf("%v", missingPermissions))
 		return nil
 	}
 
@@ -1333,9 +1333,9 @@ func (c *Operator) handleMonitorNamespaceUpdate(oldo, curo interface{}) {
 	}
 }
 
-// checkAgentAuthorization checks if the operator has access to PrometheusAgent CRD
-// It checks individual namespaces if an allowlist was provided, otherwise it checks cluster-wide.
-func (c *Operator) checkAgentAuthorization(ctx context.Context) (bool, error) {
+// getMissingPermissions returns the RBAC permissions that the controller would need to be
+// granted to fulfill its mission. An empty map means that everything is ok.
+func (c *Operator) getMissingPermissions(ctx context.Context) (map[string][]string, error) {
 	verbs := map[string][]string{
 		monitoringv1alpha1.PrometheusAgentName:                           {"get", "list", "watch"},
 		fmt.Sprintf("%s/status", monitoringv1alpha1.PrometheusAgentName): {"update"},
@@ -1343,6 +1343,8 @@ func (c *Operator) checkAgentAuthorization(ctx context.Context) (bool, error) {
 	var ssar *authv1.SelfSubjectAccessReview
 	var ssarResponse *authv1.SelfSubjectAccessReview
 	var err error
+
+	missingPermissions := map[string][]string{}
 
 	for ns := range c.config.Namespaces.PrometheusAllowList {
 		for resource, verbs := range verbs {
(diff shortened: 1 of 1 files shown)


pkg/prometheus/agent/operator.go
```diff
@@ -1344,6 +1344,11 @@ func (c *Operator) checkAgentAuthorization(ctx context.Context) (bool, error) {
 	var ssarResponse *authv1.SelfSubjectAccessReview
 	var err error
 
+	missingPermissions := map[string][]string{
+		monitoringv1alpha1.PrometheusAgentName:                           {},
+		fmt.Sprintf("%s/status", monitoringv1alpha1.PrometheusAgentName): {},
+	}
```
sthaha, line 1350: wouldn't it better if we kept it empty and add `missing[resource] = append(missing[resource], verb)` since append handles nil fine. 

```
	missing := map[string][]string{}
	missing["a"] = append(missing["a"], "bar")
```
works fine

pkg/prometheus/agent/operator.go
```diff
@@ -1360,14 +1365,21 @@ func (c *Operator) checkAgentAuthorization(ctx context.Context) (bool, error) {
...
+					return nil, err
 				}
 				if !ssarResponse.Status.Allowed {
-					return false, nil
+					missingPermissions[resource] = append(missingPermissions[resource], verb)
 				}
 			}
+			if len(missingPermissions[resource]) == 0 {
+				delete(missingPermissions, resource)
+			}
```
sthaha, line 1376: we can remove this if we followed the pattern above, `missingPermission[resource]` won't be created if the append above wasn't called.

pkg/prometheus/agent/operator.go
```diff
@@ -292,12 +292,12 @@ func (c *Operator) Run(ctx context.Context) error {
...
 	}
 
-	allowed, err := c.checkAgentAuthorization(ctx)
+	missingPermissions, err := c.getMissingPermissions(ctx)
 	if err != nil {
 		return err
 	}
-	if !allowed {
-		level.Info(c.logger).Log("msg", "Prometheus agent controller disabled because it lacks the required permissions on PrometheusAgent objects")
+	if missingPermissions != nil {
```
sthaha, line 299: instead of checking for nil, I think it is better to use `len` 

```
if len(missingPermission) != 0 {
}
```

pkg/prometheus/agent/operator.go
ArthurSens: Oh awesome, thanks for the suggestion :)

Indeed it looks a lot cleaner this way

pkg/prometheus/agent/operator.go
```diff
@@ -292,12 +292,12 @@ func (c *Operator) Run(ctx context.Context) error {
...
+		level.Info(c.logger).Log("msg", "Prometheus agent controller disabled because it lacks the required permissions on PrometheusAgent objects", "missingpermissions", fmt.Sprintf("%v", missingPermissions))
```
JoaoBraveCoding, line 300: Should this be a Warn? I kinda feel that this is more important than just info 🤔 but I'm also fine with `Info`

ArthurSens: Good point!

pkg/prometheus/agent/operator.go
```diff
@@ -1360,14 +1362,14 @@ func (c *Operator) checkAgentAuthorization(ctx context.Context) (bool, error) {
 				}
 				ssarResponse, err = c.kclient.AuthorizationV1().SelfSubjectAccessReviews().Create(ctx, ssar, metav1.CreateOptions{})
 				if err != nil {
-					return false, err
+					return map[string][]string{}, err
```
sthaha, line 1365: nil, should be sufficient and semantically correct as well

pkg/prometheus/agent/operator.go
```diff
@@ -1333,9 +1333,9 @@ func (c *Operator) handleMonitorNamespaceUpdate(oldo, curo interface{}) {
 	}
 }
 
-// checkAgentAuthorization checks if the operator has access to PrometheusAgent CRD
+// getMissingPermissions checks if the operator has access to PrometheusAgent CRD
```
simonpasquier, line 1336:
```suggestion
// getMissingPermissions returns the RBAC permissions that the controller would need to be granted to fulfill its mission. An empty map means that everything is ok.
```

sthaha (commented):
looks good to me :-)

JoaoBraveCoding (approved):
lgtm, great job @ArthurSens

simonpasquier (approved):
thanks!

PR #5476, opened 2023-04-04
fix: Add validation method for controller and crd's 

slashpai:
Move controller creation pre-requisites to validation method `CheckPrerequisites()` to avoid creating
controller object incase of validation failure. This is used for prometheus-agent controller as part of
this commit. These methods can be re-used for future CRD validation
and controllers

This change is result of alert fired during e2e test for prometheus-operator version upgrade in ocp https://github.com/openshift/prometheus-operator/pull/223. `PrometheusOperatorNotReady` was fired for `prometheus-agent` controller since CRD was not installed. Same alert will fire for upstream users who doesn't install prometheus-agent crd. This commit creates agent controller object only when validations are met.

## Description

## Type of change

## Changelog entry

```release-note
Add validation method for controller and crd's and don't create prometheusagent controller incase of validation failure
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -33,6 +33,7 @@ import (
 	logging "github.com/prometheus-operator/prometheus-operator/internal/log"
 	"github.com/prometheus-operator/prometheus-operator/pkg/admission"
 	alertmanagercontroller "github.com/prometheus-operator/prometheus-operator/pkg/alertmanager"
+	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
 	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
 	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
 	prometheusagentcontroller "github.com/prometheus-operator/prometheus-operator/pkg/prometheus/agent"
@@ -266,11 +267,39 @@ func Main() int {
 		return 1
 	}
 
-	pao, err := prometheusagentcontroller.New(ctx, cfg, log.With(logger, "component", "prometheusagentoperator"), r)
+	var pao *prometheusagentcontroller.Operator
+	verbs := map[string][]string{
+		monitoringv1alpha1.PrometheusAgentName:                           {"get", "list", "watch"},
+		fmt.Sprintf("%s/status", monitoringv1alpha1.PrometheusAgentName): {"update"},
+	}
+
+	cc, err := k8sutil.NewCRDChecker(cfg.Host, cfg.TLSInsecure, &cfg.TLSConfig)
 	if err != nil {
-		fmt.Fprint(os.Stderr, "instantiating prometheus-agent controller failed: ", err)
+		level.Error(logger).Log("msg", "failed to create new CRDChecker object ", "err", err)
 		cancel()
 		return 1
+	}
+
+	err = cc.CheckPrerequisites(ctx,
+		namespaces(cfg.Namespaces.AllowList).asSlice(),
+		verbs,
+		monitoringv1alpha1.SchemeGroupVersion.String(),
+		monitoringv1alpha1.PrometheusAgentName)
+
+	switch {
+	case errors.Is(err, k8sutil.ErrPrerequiresitesFailed):
+		level.Warn(logger).Log("msg", "Prometheus agent controller disabled because prerequisites not met", "err", err)
+	case err != nil:
+		level.Error(logger).Log("msg", "failed to check prerequisites for prometheus-agent controller ", "err", err)
+		cancel()
+		return 1
+	default:
+		pao, err = prometheusagentcontroller.New(ctx, cfg, log.With(logger, "component", "prometheusagentoperator"), r)
+		if err != nil {
+			level.Error(logger).Log("msg", "instantiating prometheus-agent controller failed", "err", err)
+			cancel()
+			return 1
+		}
 	}
 
 	ao, err := alertmanagercontroller.New(ctx, cfg, log.With(logger, "component", "alertmanageroperator"), r)
@@ -360,7 +389,9 @@ func Main() int {
 	}))
 
 	wg.Go(func() error { return po.Run(ctx) })
-	wg.Go(func() error { return pao.Run(ctx) })
+	if pao != nil {
+		wg.Go(func() error { return pao.Run(ctx) })
+	}
 	wg.Go(func() error { return ao.Run(ctx) })
 	wg.Go(func() error { return to.Run(ctx) })
 
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -16,7 +16,6 @@ package k8sutil
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -25,11 +24,13 @@ import (
 	"strings"
 
 	"github.com/cespare/xxhash/v2"
+	"github.com/pkg/errors"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
 	monitoringv1beta1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1beta1"
 	promversion "github.com/prometheus/common/version"
 	appsv1 "k8s.io/api/apps/v1"
+	authv1 "k8s.io/api/authorization/v1"
 	v1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -37,6 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/kubernetes"
 	clientappsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
@@ -51,10 +53,29 @@ var invalidDNS1123Characters = regexp.MustCompile("[^-a-z0-9]+")
 
 var scheme = runtime.NewScheme()
 
+var ErrPrerequiresitesFailed = errors.New("unmet prerequisites")
+
 func init() {
 	_ = monitoringv1.SchemeBuilder.AddToScheme(scheme)
 	_ = monitoringv1alpha1.SchemeBuilder.AddToScheme(scheme)
 	_ = monitoringv1beta1.SchemeBuilder.AddToScheme(scheme)
+}
+
+type CRDChecker struct {
+	kclient kubernetes.Interface
+}
+
+func NewCRDChecker(host string, tlsInsecure bool, tlsConfig *rest.TLSClientConfig) (*CRDChecker, error) {
+	cfg, err := NewClusterConfig(host, tlsInsecure, tlsConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "instantiating cluster config failed")
+	}
+
+	kclient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, errors.Wrap(err, "instantiating kubernetes client failed")
+	}
+	return &CRDChecker{kclient: kclient}, nil
 }
 
 // PodRunningAndReady returns whether a pod is running and each container has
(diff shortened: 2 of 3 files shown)


pkg/prometheus/agent/operator.go
```diff
@@ -82,14 +83,64 @@ type Operator struct {
...
+	if err != nil {
+		level.Warn(logger).Log("msg", "failed to check if the API supports the PrometheusAgent CRD", "err ", err)
+		return false, err
+	}
+	if !crdInstalled {
+		level.Info(logger).Log("msg", "Prometheus agent controller disabled because the PrometheusAgent CRD isn't installed")
+		return false, err
+	}
+
+	allowed, err := checkAgentAuthorization(ctx, kclient, conf)
```
simonpasquier, line 107: same here I'd also move `checkAgentAuthorization`  to the pkg/k8sutil package for later re-use. 

````
func CheckPermissions(
  ctx context.Context,
  kclient *kubernetes.Clientset,
  attrs []authv1.ResourceAttributes,  # list of attributes to check
  isAllowedFn func(authv1.ResourceAttributes, bool), # function called for every attribute to tell if it's allowed or not
) error {
  ...
}
````

File: cmd/operator/main.go [hunk 2]
```diff
@@ -265,11 +265,20 @@ func Main() int {
 		return 1
 	}
 
-	pao, err := prometheusagentcontroller.New(ctx, cfg, log.With(logger, "component", "prometheusagentoperator"), r)
-	if err != nil {
-		fmt.Fprint(os.Stderr, "instantiating prometheus-agent controller failed: ", err)
-		cancel()
-		return 1
+	var pao *prometheusagentcontroller.Operator
+	ok, err := prometheusagentcontroller.Runnable(ctx, cfg, log.With(logger, "component", "prometheusagentoperator"))
```
simonpasquier, line 269: On second thought, it might be more readable to have the core logic in main() + the pkg/k8sutil package instead of mixing between here and pkg/prometheus/agent.

pkg/prometheus/agent/operator.go
```diff
@@ -82,14 +83,64 @@ type Operator struct {
...
+		return false, errors.Wrap(err, "instantiating kubernetes client failed")
+	}
+
+	crdInstalled, err := k8sutil.IsAPIGroupVersionResourceSupported(kclient.Discovery(), monitoringv1alpha1.SchemeGroupVersion.String(), monitoringv1alpha1.PrometheusAgentName)
+	if err != nil {
+		level.Warn(logger).Log("msg", "failed to check if the API supports the PrometheusAgent CRD", "err ", err)
+		return false, err
+	}
+	if !crdInstalled {
+		level.Info(logger).Log("msg", "Prometheus agent controller disabled because the PrometheusAgent CRD isn't installed")
```
JoaoBraveCoding, line 103: In my opinion this should be `Warn` as well 
```suggestion
		level.Warn(logger).Log("msg", "Prometheus agent controller disabled because the PrometheusAgent CRD isn't installed")
```

simonpasquier: I'm not sure: you might not want to deploy a v1alpha1 CRD on your cluster. In which case, the warning log might might be confusing.
But if the CRD is deployed, not having the required permissions is worth a warning.

JoaoBraveCoding: Yes, that makes sense! It's indeed two separate situations that should be handled differently! Thanks for clarifying!

cmd/operator/main.go
slashpai: Move `Runnable()` to main?

pkg/prometheus/agent/operator.go
```diff
@@ -1335,18 +1358,14 @@ func (c *Operator) handleMonitorNamespaceUpdate(oldo, curo interface{}) {
 
 // getMissingPermissions returns the RBAC permissions that the controller would need to be
 // granted to fulfill its mission. An empty map means that everything is ok.
-func (c *Operator) getMissingPermissions(ctx context.Context) (map[string][]string, error) {
-	verbs := map[string][]string{
-		monitoringv1alpha1.PrometheusAgentName:                           {"get", "list", "watch"},
-		fmt.Sprintf("%s/status", monitoringv1alpha1.PrometheusAgentName): {"update"},
-	}
+func getMissingPermissions(ctx context.Context, kclient *kubernetes.Clientset, conf operator.Config, verbs map[string][]string) (map[string][]string, error) {
```
slashpai, line 1361: we cannot move this to k8sutil since it will create cyclic imports, we need to move to another package

JoaoBraveCoding: Which imports are causing the cycle? I'm asking because at first it's not clear

cmd/operator/main.go
JoaoBraveCoding: I believe so since AFAIS apart from the `verbs` there is nothing specific to prometheus agent so most of the code should be able to be extracted and here we could maybe just have a mapping between the CR and the necessary verbs and call `Runnable` with that.

```
	verbs := map[string][]string{
		monitoringv1alpha1.PrometheusAgentName:                           {"get", "list", "watch"},
		fmt.Sprintf("%s/status", monitoringv1alpha1.PrometheusAgentName): {"update"},
	}
        ok, err := k8sutil.Runnable(ctx, cfg, log.With(logger, "component", "prometheusagentoperator"), "PrometheusAgent", verbs)

File: pkg/prometheus/agent/operator.go (hunk 4 above)
Comment (slashpai) [reply pkg/prometheus/agent/operator.go:1361]: we are importing `k8sutil` in operator package and if we move this func to `k8sutil` we need to import operator package over there since we pass operator.Config

Comment (simonpasquier) [reply pkg/prometheus/agent/operator.go:1361]: Since the function only needs `conf.Namespaces.PrometheusAllowList`, we can modify the signature to be 

```suggestion
func getMissingPermissions(ctx context.Context, kclient *kubernetes.Clientset, namespaces []string, verbs map[string][]string) (map[string][]string, error) {
```

cmd/operator/main.go
simonpasquier: basically what Joao said.

slashpai:
Making a few changes hence moving to draft

pkg/prometheus/agent/operator.go
```diff
@@ -1332,44 +1312,3 @@ func (c *Operator) handleMonitorNamespaceUpdate(oldo, curo interface{}) {
 		)
 	}
 }
-
-// getMissingPermissions returns the RBAC permissions that the controller would need to be
```
slashpai, line 1336: Moved to k8sutil

slashpai:
@simonpasquier @JoaoBraveCoding This should be good to review again :)

ArthurSens (approved):
Brilliant :)

Just a nit

pkg/k8sutil/k8sutil.go
```diff
@@ -114,6 +137,83 @@ func NewClusterConfig(host string, tlsInsecure bool, tlsConfig *rest.TLSClientCo
...
+	}
+
+	return true, nil
+}
+
+func (c CRConfig) IsCRDInstalled(kclient *kubernetes.Clientset, logger log.Logger) (bool, error) {
+	crdInstalled, err := IsAPIGroupVersionResourceSupported(kclient.Discovery(), c.schemeGroupVersion, c.resourceName)
+	if err != nil {
+		level.Warn(logger).Log("msg", fmt.Sprintf("failed to check if the API supports the %s CRD", c.kind), "err ", err)
+		return false, err
```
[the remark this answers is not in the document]
slashpai: surely better that way :)

slashpai:
awaiting https://github.com/prometheus-operator/prometheus-operator/pull/5476 to be merged

pkg/k8sutil/k8sutil.go
```diff
@@ -114,6 +136,80 @@ func NewClusterConfig(host string, tlsInsecure bool, tlsConfig *rest.TLSClientCo
 	return cfg, nil
 }
 
+func (c CRConfig) Runnable(ctx context.Context, host string, tlsInsecure bool, tlsConfig *rest.TLSClientConfig, logger log.Logger) (bool, error) {
```
JoaoBraveCoding, line 139: Feel like here `cr` is more readable than `c` no? Since in main.go we also reference it using `cr` WDYT?
```suggestion
func (cr CRConfig) Runnable(ctx context.Context, host string, tlsInsecure bool, tlsConfig *rest.TLSClientConfig, logger log.Logger) (bool, error) {
```

pkg/k8sutil/k8sutil.go
```diff
@@ -114,6 +136,80 @@ func NewClusterConfig(host string, tlsInsecure bool, tlsConfig *rest.TLSClientCo
...
+		return false, err
+	}
+	if len(missingPermissions) > 0 {
+		return false, errors.Wrapf(err, "controller disabled because it lacks the required permissions: missingpermissions %v", missingPermissions)
+	}
+
+	return true, nil
+}
+
+func (c CRConfig) IsCRDInstalled(kclient *kubernetes.Clientset) (bool, error) {
```
JoaoBraveCoding, line 166: I don't think we want to expose  `IsCRDInstalled` and `GetMissingPermissions` outside of the package no? At least for now I don't see them being used outside

cmd/operator/main.go
```diff
@@ -266,11 +267,28 @@ func Main() int {
...
-	pao, err := prometheusagentcontroller.New(ctx, cfg, log.With(logger, "component", "prometheusagentoperator"), r)
-	if err != nil {
-		fmt.Fprint(os.Stderr, "instantiating prometheus-agent controller failed: ", err)
-		cancel()
-		return 1
+	var pao *prometheusagentcontroller.Operator
+	verbs := map[string][]string{
+		monitoringv1alpha1.PrometheusAgentName:                           {"get", "list", "watch"},
+		fmt.Sprintf("%s/status", monitoringv1alpha1.PrometheusAgentName): {"update"},
+	}
```
JoaoBraveCoding, line 274: nit: part of me feels like these might belong to either `monitoringv1alpha1` or `prometheusagentcontroller` inside a function like `GetPrometheusAgentVerbs` but I'll let @simonpasquier give his 2 cents, this might just be me being picky

JoaoBraveCoding (commented):
Looks great! Just some small comments

pkg/k8sutil/k8sutil.go
slashpai: Ya right. I kept it that way for ScrapeConfig validation. But we can enable later if needed

cmd/operator/main.go
simonpasquier: I see your point. I think we can leave it for now and see later (as we reuse the same pattern) if it makes sense to move it back.

slashpai:
@JoaoBraveCoding @simonpasquier are we good with the changes? :)

cmd/operator/main.go
```diff
@@ -266,11 +267,28 @@ func Main() int {
...
+		monitoringv1alpha1.SchemeGroupVersion.String(),
+		monitoringv1alpha1.PrometheusAgentsKind,
+		monitoringv1alpha1.PrometheusAgentName)
+
+	err = cr.Runnable(ctx, cfg.Host, cfg.TLSInsecure, &cfg.TLSConfig)
 	if err != nil {
-		fmt.Fprint(os.Stderr, "instantiating prometheus-agent controller failed: ", err)
-		cancel()
-		return 1
+		fmt.Fprint(os.Stderr, "pre-requistes for prometheus-agent controller failed: ", err)
```
simonpasquier, line 284: can we use the logger instead?

cmd/operator/main.go
```diff
@@ -266,11 +267,28 @@ func Main() int {
...
+	} else {
+		pao, err = prometheusagentcontroller.New(ctx, cfg, log.With(logger, "component", "prometheusagentoperator"), r)
+		if err != nil {
+			fmt.Fprint(os.Stderr, "instantiating prometheus-agent controller failed: ", err)
```
simonpasquier, line 288: same here

pkg/k8sutil/k8sutil.go
```diff
@@ -57,6 +59,25 @@ func init() {
...
+// CRConfig is CustomResource config
+type CRConfig struct {
+	nsAllowList        map[string]struct{}
+	verbs              map[string][]string
+	schemeGroupVersion string
+	kind               string
+	resourceName       string
+}
+
+func NewCRConfig(nsList map[string]struct{}, verbs map[string][]string, sgv, kind, resource string) CRConfig {
```
simonpasquier, line 71: I'd recommend inverting the parameters: the struct should require API connection parameters and the method should require which CRD, namespaces and verbs to check. It will make it easier to reuse.

simonpasquier: In which case:

```suggestion
func NewCRDChecker(host string, tlsInsecure bool, tlsConfig *rest.TLSClientConfig) (CRDChecker, error) {
```

pkg/k8sutil/k8sutil.go
```diff
@@ -114,6 +135,79 @@ func NewClusterConfig(host string, tlsInsecure bool, tlsConfig *rest.TLSClientCo
 	return cfg, nil
 }
 
+func (cr CRConfig) Runnable(ctx context.Context, host string, tlsInsecure bool, tlsConfig *rest.TLSClientConfig) error {
```
simonpasquier, line 138:
```suggestion
func (cc *CRDChecker) CheckPrerequisites(ctx context.Context, ...) error {
```

cmd/operator/main.go
```diff
@@ -266,11 +267,28 @@ func Main() int {
...
+	}
+
+	cr := k8sutil.NewCRConfig(cfg.Namespaces.AllowList,
+		verbs,
+		monitoringv1alpha1.SchemeGroupVersion.String(),
+		monitoringv1alpha1.PrometheusAgentsKind,
+		monitoringv1alpha1.PrometheusAgentName)
+
+	err = cr.Runnable(ctx, cfg.Host, cfg.TLSInsecure, &cfg.TLSConfig)
 	if err != nil {
```
simonpasquier, line 283: we need to distinguish between 2 classes of error:
* the API server returned an unexpected error -> the operator needs to fail
* the prerequisites aren't met -> the controller can't be started.

I'd recommend wrapping a sentinel error (see https://bitfieldconsulting.com/golang/wrapping-errors) to detect the second case.

cmd/operator/main.go
slashpai: sure I thought so as well. The reason I kept Stderr is for other controllers as well stderr is used. But we can update for them in another PR

cmd/operator/main.go
```diff
@@ -266,13 +267,44 @@ func Main() int {
...
 		return 1
 	}
 
+	err = cc.CheckPrerequisites(ctx,
+		cfg.Namespaces.AllowList,
+		verbs,
+		monitoringv1alpha1.SchemeGroupVersion.String(),
+		monitoringv1alpha1.PrometheusAgentsKind,
+		monitoringv1alpha1.PrometheusAgentName)
+	if err != nil {
```
simonpasquier, line 289: (nit) we could have a swtich/case to simplify the reading

```
switch {
    case error.Is(err, ...):
        level.Error(logger).Log(...)
    case err != nil:
        ...
        return 1
     default:
        pao, err = prometheusagentcontroller.New(...)
        ...
}
```

pkg/k8sutil/k8sutil.go
```diff
@@ -51,12 +53,33 @@ var invalidDNS1123Characters = regexp.MustCompile("[^-a-z0-9]+")
...
+var ErrorControllerPreRequistesNotMet = errors.New("controller pre-requisites are not met")
+
 func init() {
 	_ = monitoringv1.SchemeBuilder.AddToScheme(scheme)
 	_ = monitoringv1alpha1.SchemeBuilder.AddToScheme(scheme)
 	_ = monitoringv1beta1.SchemeBuilder.AddToScheme(scheme)
 }
 
+type CRDChecker struct {
+	kclient *kubernetes.Clientset
```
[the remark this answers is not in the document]
slashpai: Just curious to know why we need to change this :)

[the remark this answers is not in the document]
simonpasquier: it's better in the long term to use an interface than a concrete class (e.g. if we want to unit test in the future).

pkg/k8sutil/k8sutil.go
```diff
@@ -114,6 +135,69 @@ func NewClusterConfig(host string, tlsInsecure bool, tlsConfig *rest.TLSClientCo
 	return cfg, nil
 }
 
+func (cc CRDChecker) CheckPrerequisites(ctx context.Context, nsAllowList []string, verbs map[string][]string, sgv, kind, resource string) error {
```
JoaoBraveCoding, line 138:
```suggestion
// CheckPrerequisites for a given k8s resource assert if it's installed in the cluster and if the operator service
// account has the necessary RBAC verbs, in the namespace list, to reconcile it.
func (cc CRDChecker) CheckPrerequisites(ctx context.Context, nsAllowList []string, verbs map[string][]string, sgv, kind, resource string) error {
```

JoaoBraveCoding (commented):
One small nit but lgtm!

cmd/operator/main.go
```diff
@@ -266,13 +267,42 @@ func Main() int {
...
+		verbs,
+		monitoringv1alpha1.SchemeGroupVersion.String(),
+		monitoringv1alpha1.PrometheusAgentsKind,
+		monitoringv1alpha1.PrometheusAgentName)
+
+	switch {
+	case errors.Is(err, k8sutil.ErrPrerequiresitesFailed):
+		level.Error(logger).Log("msg", "pre-requistes for prometheus-agent controller failed", "err", err)
+	case err != nil:
+		level.Error(logger).Log("msg", "unexpected error", "err", err)
```
simonpasquier, line 294:
```suggestion
		level.Error(logger).Log("msg", "failed to check prerequisites for prometheus-agent controller ", "err", err)
```

cmd/operator/main.go
```diff
@@ -266,13 +267,42 @@ func Main() int {
...
+	err = cc.CheckPrerequisites(ctx,
+		namespaces(cfg.Namespaces.AllowList).asSlice(),
+		verbs,
+		monitoringv1alpha1.SchemeGroupVersion.String(),
+		monitoringv1alpha1.PrometheusAgentsKind,
+		monitoringv1alpha1.PrometheusAgentName)
+
+	switch {
+	case errors.Is(err, k8sutil.ErrPrerequiresitesFailed):
+		level.Error(logger).Log("msg", "pre-requistes for prometheus-agent controller failed", "err", err)
```
simonpasquier, line 292: Previously we had an info log if the CRD isn't available and a warn log if permissions were not granted. To avoid more back-and-forth for this PR, I'm ok with sticking to warn.
 
```suggestion
		level.Warn(logger).Log("msg", "Prometheus agent controller disabled because prerequisites not met", "err", err)
```

slashpai:
@simonpasquier updated 🤞🏻

simonpasquier (approved):

simonpasquier:
thanks!

rafilkmp3:
How can I add permission to prometheus agent  CRD ? I installed using helm.

ArthurSens:
> How can I add permission to prometheus agent CRD ? I installed using helm.

The necessary permissions can be seen [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/example/rbac/prometheus-operator/prometheus-operator-cluster-role.yaml). I'm not really familiar how to do so using helm. 

Maybe it's worth asking at the [helm repository](https://github.com/prometheus-community/helm-charts)?

PR #5477, opened 2023-04-04
Ensure securityContext.seccompProfile.type is set

jcpunk:
## Description
This sets the default value for `securityContext.seccompProfile.type`.  When running with warnings for `PodSecurity "restricted:1.26.0` a message is emitted that this is unset.

Message
```
Warning: would violate PodSecurity "restricted:1.26.0": seccompProfile (pod or container "prometheus-operator" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
deployment.apps/prometheus-operator created
```

As `RuntimeDefault` is already the used default this should have no user facing changes.

## Type of change

## Changelog entry

```release-note
Set a default for `securityContext.seccompProfile.type` on `deployment.apps/prometheus-operator`
```

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -37773,6 +37773,8 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: prometheus-operator
 ---
 apiVersion: v1
--- a/Documentation/user-guides/webhook.md
+++ b/Documentation/user-guides/webhook.md
@@ -157,6 +157,8 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: prometheus-operator-admission-webhook
       volumes:
       - name: tls-certificates
--- a/example/admission-webhook/deployment.yaml
+++ b/example/admission-webhook/deployment.yaml
@@ -63,6 +63,8 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: prometheus-operator-admission-webhook
       volumes:
       - name: tls-certificates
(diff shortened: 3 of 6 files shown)


bundle.yaml
```diff
@@ -37773,6 +37773,8 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
+        seccompProfile:
+          type: RuntimeDefault
```
sthaha, line 37777: Thanks for the patch! `bundle.yaml` is a generated file (see makefile target `bundle.yaml`). Below may be what you are looking for, followed by make generate. 

```
diff --git a/jsonnet/prometheus-operator/prometheus-operator.libsonnet b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
index 7e2688e06..49fa5b37c 100644
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -33,7 +33,7 @@ function(params) {
                                                  if po.config.enableAlertmanagerConfigV1beta1 then
                                                    (import 'alertmanagerconfigs-v1beta1-crd.libsonnet')
                                                  else {},
-  '0prometheusagentCustomResourceDefinition': import 'prometheusagents-crd.json',
+  '0prometheusagentCustomResourceDefinition': import 'prometheusagents-crd.json',
   '0prometheusCustomResourceDefinition': import 'prometheuses-crd.json',
   '0servicemonitorCustomResourceDefinition': import 'servicemonitors-crd.json',
   '0podmonitorCustomResourceDefinition': import 'podmonitors-crd.json',
@@ -178,6 +178,7 @@ function(params) {
             securityContext: {
               runAsNonRoot: true,
               runAsUser: 65534,
+              seccomProfile: 'RuntimeDefault',
             },
             serviceAccountName: po.config.name,
             automountServiceAccountToken: true,
```

jcpunk: I think I've made the requested changes, the libsonnet does make a lot more sense.

sthaha (approved):
looks good to me. Will request @simonpasquier for final review before merging.

Issue #5452, opened 2023-03-24
Add support for Silence CRD


mcbenjemaa:
**What is missing?**

Definitely having a way of managing Silences through the K8s API, by adding support for Silence CRD. 

A lot of people managing silences with different approaches, but it's always cumbersome.

At Giant Swarm, we manage silences through an operator, we call it [silence-operator](https://github.com/giantswarm/silence-operator/)
We would love to move the implementation to prometheus-operator and provide it to the community. 

**Why do we need it?**
Dynamically manage Silences with prometheus-operator. 

Benefits for us (Giant Swarm):  less components, better upstream support 
**Environment**

* Prometheus Operator version:

N/A

**Anything else we need to know?**:

I would like to implement this.

Any suggestions?:
- Do I need to do it in separated PRs?
- What kind of e2e test shall I integrate?

simonpasquier:
The topic was already discussed in #2398. At the very least we can start a [design document](https://github.com/prometheus-operator/prometheus-operator/tree/main/Documentation/proposals) to discuss the motivation, use cases, possible pitfalls and implementation options.

marioferh:
Hi, I noticed this issue hasn’t had any recent activity. If no one is currently working on it, I’d be happy to take it on.

danielmellado:
Hi Mario, I was already taking a look at this and https://github.com/prometheus-operator/prometheus-operator/issues/6508 but let's collaborate on this ;)

slashpai:
Thanks folks :)

mcbenjemaa:
Please take a look at the PR https://github.com/prometheus-operator/prometheus-operator/pull/5485
for the Design document.

Issue #2398, opened 2019-02-12
[Feature idea/request] Create alertmanager silences via CRD


dharmab:
It would be nice to create silences in alertmanager via a custom resource definition. Some use cases:

- I am a cluster operator who needs to upgrade the etcd cluster. I want to create a silence prior to taking down each etcd node and remove the silence after I have finished upgrading the node.
- I am a developer who needs to update a StatefulSet. I want to create a silence for the duration of the StatefulSet change and remove it when the change is complete.

Is this a good and/or feasible idea? CRDs are easier to deal with than the alertmanager API.

metalmatze:
What would be wrong about talking to the Alertmanager API directly? :thinking: 
I don't think we need a CRD for this.

/cc @mxinden as he is one of the Alertmanager maintainers

dharmab:
Using apiserver is nice because we already have apiserver connected to our identity provider (in my case, Active Directory.) So the user can use their existing identity, which already using 2FA and visible in audit logs. Or in the case of machine accounts, we don't need to create separate machine credentials for Alertmanager as we can re-use a ServiceAccount.

brancz:
I don't think a CRD is necessary for auditing/auth. Just use the [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) in front of the Alertmanager and create an artificial resource then you can already use the auditing/auth features.

Pluies:
As an additional data point, I also think this would be a nice feature. :)

Having a CRD would integrate nicely into the whole Kubernetes ecosystem, e.g. give the possibility to create silences with Helm, quickly list them with `kubectl`, etc.

It also lowers the overall complexity of the system from an operator point of view, by having one interface to interact with the system (k8s) instead of two (k8s + alertmanager api).

This would also be conceptually identical to the way the prometheus-operator defines [Prometheus rules as CRDs](https://github.com/coreos/prometheus-operator/blob/master/example/prometheus-operator-crd/prometheusrule.crd.yaml), which we've found very easy and convenient to work with here at $work.

brancz:
I'm not sure they're identical to Kubernetes APIs, as silences expire and have lots of lifecycle behavior of their own. While I understand the desire, I'm not certain it fits into the "Kubernetes API"-box.

Nexucis:
[...]

PR #5485, opened 2023-04-11
Silence CRD Design Proposal

mcbenjemaa:
## Description

Design document for the Proposal of adding Silence CRD.

  * [#5452](https://github.com/prometheus-operator/prometheus-operator/issues/5452)
  * [#2398](https://github.com/prometheus-operator/prometheus-operator/issues/2398)

## Type of change

## Changelog entry

```release-note

```


Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+
+* This proposal doesn't aim to remove the CR after the expiration date has reached
+* Refactoring of the other CRDs is not in scope for the first version
+
+# How
+
+Creating a new cluster-scope Silence CRD that will act as an interface by adding silences via the Alertmanager API.
+Usage of Silences doesn't exclude the use of the other CRDs, they are not mutually exclusive.
+`Silence` will allow the creation of any silences to Alertmanager, while the other CRDs provide sane defaults. This will allow
+for isolated testing of the new `Silence` CRD.
```
xiu, line 62: I guess this is coming from the ScrapeConfig proposal, this doesn't make a lot of sense in the context of the Silence CRD. There is no overlap with the other CRDs.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+
+A typical Silence resouce should looks like the following:
```
xiu, line 64:
```suggestion
A typical `Silence` resource could look like the following:
```
We shouldn't mandate what it should look like, the implementation will take care of that.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+```yaml
+[...]
+spec:
+  silenceSelector: ...
+  silenceNamespaceSelector: ...
+```
+
+# Alternatives
+
+* Use Alertmanager API, with the pitfalls described earlier
```
xiu, line 108:
```suggestion
* Use Alertmanager API, with the pitfalls described earlier
* Use [silence-operator](https://github.com/giantswarm/silence-operator), which is suboptimal as it requires users to install two operators for the feature
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Audience
+
+* Users who serve Prometheus as a service and want to have an interface in defining silences exposed to developers.
+* Users who want to manage silences the same way as for services running within the Kubernetes cluster
+* Users who want a supported Kubernetes way of silences outside the Kubernetes cluster
+
+# Non-Goals
+
+* This proposal doesn't aim to remove the CR after the expiration date has reached
```
TheoBrigitte, line 54: No operator should ever remove CR. But do you mean that silences won't be expired/removed from Alertmanager ?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+    - name: namespace
+      value: test-ns
+      isRegex: false
+```
+
+The above resource will result in creating a silence in the Alertmanager with label selector,
+`app: test-alertmanager`.
+
+* `selector` used to select which Kubernetes Service to use as an Alertmanager API.
+* `validUntil` to define the expiration of the silence.
```
xiu, line 87: Would it make sense to have silences that stay active so long as the resource exists? For instance, if `validUntil` is undefined, then the Silence stays active. If it is, the `Silence` expires at the date defined.

TheoBrigitte, line 87: I would rather align this with the Silence field `ExpiresAt`.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+  matchers:
+    - name: namespace
+      value: test-ns
+      isRegex: false
+```
+
+The above resource will result in creating a silence in the Alertmanager with label selector,
+`app: test-alertmanager`.
+
+* `selector` used to select which Kubernetes Service to use as an Alertmanager API.
```
xiu, line 86: Shouldn't this be on the Prometheus CRD, which would match a label on the Silence? Same as `PrometheusRule`.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+prometheus-operator doesn't have a way to automate the management of Alertmanager silences. Many users have either been using some internal scripts
+or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
+
+* Users who are using CI/CD jobs to manage silences have reported that this is really cumbersome.
+* Users that use scripts are not in better shape obviously.
+* Users that use a standalone operator that exposes a Silence CRD are in a better situation, some folks have reported that they are using GitOps
+  to fully manage the life cycle of Silences Custom resources, Nevertheless, this comes with a drawback
+  as the team must be able to manage a new component in their stack.
+
+Additionally, Having a new component in the stack and keeping it maintained is not always ideal (said the folks at [Giant Swarm](https://giantswarm.io) the owners of [silence-operator](https://github.com/giantswarm/silence-operator)).
```
xiu, line 25:
```suggestion
Additionally, having a new component in the stack and keeping it maintained is not always ideal (said the folks at [Giant Swarm](https://giantswarm.io) the owners of [silence-operator](https://github.com/giantswarm/silence-operator)).
```

Can you elaborate as to why it's not ideal?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+It's really better to have this feature as part of prometheus-operator and keep it maintained and available for the community.
```
xiu, line 26:
```suggestion
Having the Silence CRD as part of prometheus-operator provides users with a better experience: there is no need to install another operator or build a system on top of it to provide that functionality.
```

xiu (commented):
Exciting! 🥳 Added a few comments, feel free to discard if I misunderstood.

Documentation/proposals/202304-silences.md
mcbenjemaa: Ah Yes

Documentation/proposals/202304-silences.md
mcbenjemaa: Yeah, I'm guessing if we can add a default maximum days if the user has not defined the `validUntil `

Documentation/proposals/202304-silences.md
mcbenjemaa: Yes, I have added both strategies

Documentation/proposals/202304-silences.md
mcbenjemaa: I was highlighting that maybe some users request the feature of automatic cleanup after the silences have expired.
We will not support that for now...

Documentation/proposals/202304-silences.md
mcbenjemaa: done.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+By Adding support for `Silence` CRD in the prometheus-operator, this will make users more flexible in terms of choosing the tool
+to deploy their silences and will free others from managing a standalone component within the stack.
+
+## Pitfalls of the current solution
+
+Using Alertmanager API Directly comes with drawbacks:
+
+* Teams have to build an automation to add silences in a centralized manner
+* There is no input validation, which can lead to an invalid silence configuration
+* Teams need to manage expiration on their own
```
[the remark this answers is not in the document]
mcbenjemaa: The intention here is that the controller will handle the expiration, via the status subresource.

Documentation/proposals/202304-silences.md
mcbenjemaa: I thought there might be multiple alertmanagers in same cluster, 
but I also have a section where I define: 

```
  silenceSelector: ...
  silenceNamespaceSelector: ...
```

in the Alertmanager CRD.

Documentation/proposals/202304-silences.md
mcbenjemaa: we can make the field `expiresAt` required so :)

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+
+* Users who are using CI/CD jobs to manage silences have reported that this is really cumbersome.
+* Users that use scripts are not in better shape obviously.
+* Users that use a standalone operator that exposes a Silence CRD are in a better situation, some folks have reported that they are using GitOps
+  to fully manage the life cycle of Silences Custom resources, Nevertheless, this comes with a drawback
+  as the team must be able to manage a new component in their stack.
+
+Additionally, having a new component in the stack and keeping it maintained is not always ideal (said the folks at [Giant Swarm](https://giantswarm.io) the owners of [silence-operator](https://github.com/giantswarm/silence-operator)),
+because that would require us to fully manage this component as of any managed app (monitoring, logging, alerting etc..).
+Having the Silence CRD as part of prometheus-operator provides users with a better experience: there is no need to install another operator or build a system on top of it to provide that functionality.
```
ArthurSens, line 27: I can see that having 2 components on the stack is more complex than 1, but here we're moving one component by making the other slightly more complex as well, i.e. users still need to monitor Prometheus-operator and Alertmanager to understand if their Silence CRs are behaving as expected.

Of course, I'm saying this without any experience with the silence-operator 😅, it might as well be a super simple component and I'm here making a fuss :P

Could you tell us a bit more about the common failure modes that you've found while using it at Giantswarm?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+
+By Adding support for `Silence` CRD in the prometheus-operator, this will make users more flexible in terms of choosing the tool
+to deploy their silences and will free others from managing a standalone component within the stack.
```
ArthurSens, line 30:
```suggestion
Adding support for `Silence` CRD in the Prometheus-operator will provide more flexibility to users in terms of choosing the tool to deploy their silences and will free users of Giantswarm's silence operator from managing a standalone component within the stack.
```

I'm not sure about this suggestion, to be honest 😅 

"Free others" sounds too vague, but being very direct that this is helping users of Giantswarm home-grown solution sounds like an attack, which is not my intention 😬

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Pitfalls of the current solution
+
+Using Alertmanager API Directly comes with drawbacks:
+
+* Teams have to build an automation to add silences in a centralized manner
+* There is no input validation, which can lead to an invalid silence configuration
+* Teams need to manage expiration on their own
+* If a silence was deleted by accident, it will be permanent.
```
ArthurSens, line 39: I might be talking non-sense here, but wouldn't it be possible to implement a [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/#:~:text=Finalizers%20are%20namespaced%20keys%20that,resources%20the%20deleted%20object%20owned.) here? Making sure the silence is revoked from alertmanager before the resource gets deleted.

Documentation/proposals/202304-silences.md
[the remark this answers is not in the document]
ArthurSens: I'm not sure if I understand the problem here. With a CRD, users still need to set an expiration date, don't they?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Goals
+
+* Provide a way for users to manage Alertmanager silences with the Kubernetes Way.
```
ArthurSens, line 43:
```suggestion
* Provide a way for users to manage Alertmanager silences with Custom Resources (CR).
```

Kubernetes way might sound vague and ambiguous.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+* Make it easier to manage Alertmanager silences via centralised repo with e.g. ArgoCD, Flux...
```
ArthurSens, line 44:
```suggestion
* Enable management of Alertmanager silences via centralised repo with e.g. ArgoCD, Flux...
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+* Provide a way to manage silence expiration with a silence CR.
```
ArthurSens, line 45:
```suggestion
```

The first goal already covers this one, doesn't it?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Audience
+
+* Users who serve Prometheus as a service and want to have an interface in defining silences exposed to developers.
+* Users who want to manage silences the same way as for services running within the Kubernetes cluster
+* Users who want a supported Kubernetes way of silences outside the Kubernetes cluster
```
ArthurSens, line 51:
```suggestion
* Users who want to manage silences the same way as for services running within the Kubernetes cluster.
```

"Kubernetes way" sounds vague and ambiguous. And it is the same audience mentioned above right?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+
+* `selector` used to select which Kubernetes Service to use as an Alertmanager API.
+* `expiresAt` to define the expiration of the silence.
+* `matchers` field corresponds to the Alertmanager silence matchers each of which consists of:
+  - `name` - name of tag on an alert to match
+  - `value` - fixed string or expression to match against the value of the tag named by name above on an alert
+  - `isRegex` - a boolean specifying whether to treat value as a regex (=~) or a fixed string (=)
+  - `isEqual` - a boolean specifying whether to use equal signs (= or =~) or to negate the matcher (!= or !~)
+
+This example doesn't list all the fields that are offered by prometheus. The implementation of all the fields will be
```
ArthurSens, line 92:
```suggestion
This example doesn't list all the fields that are offered by Alertmanager. The implementation of all the fields will be
```

Documentation/proposals/202304-silences.md
ArthurSens: Could we decide on using the selector in only one CRD?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+done in an iterative process and as such, the expectation is not for all of them to be implemented in the first version.
```
ArthurSens, line 93: I might be wrong, but the missing fields are just "Created By" and "Comment", aren't they? Which doesn't seem to offer a big challenge to implement when compared to expiration handling.

Couldn't we just add all fields all at once?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+[...]
+spec:
+  silenceSelector: ...
+  silenceNamespaceSelector: ...
+```
+
+# Alternatives
+
+* Use Alertmanager API, with the pitfalls described earlier
+* Use [silence-operator](https://github.com/giantswarm/silence-operator), which is suboptimal as it requires users to install two operators for the feature
```
ArthurSens, line 107:
```suggestion
* Use Alertmanager API.
* Use [silence-operator](https://github.com/giantswarm/silence-operator).
```

ArthurSens (commented):
Thanks for raising the PR :) It is a good start!

And sorry about a lot of NIT reviews, at first glance, a reader can feel that something is trying to be sold here, instead of just stating facts. 

In a design document, we could provide the reader with just plain information and he/she will be the one making conclusions about how bad the current situation is and how good it can become.

Documentation/proposals/202304-silences.md
TheoBrigitte: Alright then maybe re-phrase this to something like

```suggestion
* This proposal do not cover automatic cleanup of expired Silence CRs
```

Documentation/proposals/202304-silences.md
mcbenjemaa: Yeah, this is not a configuration

mcbenjemaa: it's an API, so we just use this selector

Documentation/proposals/202304-silences.md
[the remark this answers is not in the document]
mcbenjemaa: yes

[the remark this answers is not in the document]
mcbenjemaa: the Alertmanager itself will handle the expiration. 
but, the actual Silence resource will still be deployed in the cluster.

I thought we could implement a cleanup mechanism once the silence has expired.

which means, if a silence is expired it will be deleted from the cluster as well as from Alertmanager

Documentation/proposals/202304-silences.md
mcbenjemaa: this is done

Documentation/proposals/202304-silences.md
mcbenjemaa: we can add them as annotation, or you prefer fields under the spec?

Documentation/proposals/202304-silences.md
mcbenjemaa: Yeah,

mcbenjemaa: @theo

Documentation/proposals/202304-silences.md
mcbenjemaa: We didn't have complex failures or monitoring for this component, we just need to have an alert that triggers if  a certain Silence reconciliation failed.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+# Why
+
+prometheus-operator doesn't have a way to automate the management of Alertmanager silences. Many users have either been using some internal scripts
+or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
+
+* Users who are using CI/CD jobs to manage silences have reported that this is really cumbersome.
+* Users that use scripts are not in better shape obviously.
+* Users that use a standalone operator that exposes a Silence CRD are in a better situation, some folks have reported that they are using GitOps
+  to fully manage the life cycle of Silences Custom resources, Nevertheless, this comes with a drawback
+  as the team must be able to manage a new component in their stack.
```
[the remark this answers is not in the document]
mcbenjemaa: could you check!

Documentation/proposals/202304-silences.md
mcbenjemaa: No, this section explains the drawback of using the API directly. 
in the Silence CR approach, this will be fixed by the control loops that will solve the drift.

Documentation/proposals/202304-silences.md
mcbenjemaa: I think yes

Documentation/proposals/202304-silences.md
QuentinBisson: As @mcbenjemaa said, we have only 2 possible failure scenarios in the silence-operator, either the silence cannot be reconcilied with alertmanager (which almost never happen because we have a small Silence CRD API) or the flux kustomization pulling the silences if failing. 

If this hits into prometheus-operator, the failure modes would roughly be the same in my opinion

Documentation/proposals/202304-silences.md
QuentinBisson: I think one of the idea here is that new users would be able to use this operator without having to think how to manage their silences and existing users would be able to use it but they could also use their existing mechanism if they wish

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,108 @@
...
+* Related Tickets:
+  * [#5452](https://github.com/prometheus-operator/prometheus-operator/issues/5452)
+  * [#2398](https://github.com/prometheus-operator/prometheus-operator/issues/2398)
+* Other docs:
+  * n/a
+
+This document describes the creation of `Silence` Custom Resource Definition that defines Alertmanager silences
+configurations in the Kubernetes way.
+
+# Why
```
QuentinBisson, line 14: I think we are missing a few reasons why we think this is useful:
1 - Multi tenancy: Alertmanager cannot be provided in read-only mode but silences CR can be using kubernetes rbac 
2 - Moving the giantswarm silence operator to prometheus operator would provide better community support than what we can currently offer

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,108 @@
...
+
+Prometheus-operator doesn't have a way to automate the management of Alertmanager silences. Users have either been using internal scripts
+or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
+
+* Users using CI/CD jobs to manage silences have reported that this is cumbersome, since they need to either expose
```
QuentinBisson, line 19: I would emphasize that this is even harder for multi-cluster setups as having flux and the operator running would simplify that (no need to complex CI setup or a job running in every cluster)

Documentation/proposals/202304-silences.md
mcbenjemaa: I added them in the goals section

Documentation/proposals/202304-silences.md
mcbenjemaa: I agree

Documentation/proposals/202304-silences.md
QuentinBisson: I think also we could mention that this can offer a better user experience so they can use the same tools to manage their monitoring pipeline (kyverno for instance, cr validations) whereas if silences are created via jobs or any other mechanism lire direct access to alertmanager, you have to redo this tooling again

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+Using Alertmanager API Directly comes with drawbacks:
+
+* Teams have to build an automation to add silences in a centralized manner
+* There is no input validation, which can lead to an invalid silence configuration
+* Teams need to manage expiration on their own
+* If a silence was deleted by accident, it will be permanent.
+
+# Goals
+
+* Provide a way for users to manage Alertmanager silences with Custom Resources (CR).
```
simonpasquier, line 51: (nit)
```suggestion
* Provide a way for users to manage Alertmanager silences with a Custom Resource Definition (CRD).
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+* Enable management of Alertmanager silences via centralised repo with e.g. ArgoCD, Flux...
+* Multi tenancy: Silences CR can be using kubernetes rbac.
+* Moving the Giant Swarm silence operator to prometheus operator would provide better community support
+
+## Audience
+
+* Users who serve Prometheus as a service and want to have an interface in defining silences exposed to developers.
```
simonpasquier, line 58: (nit)
```suggestion
* Users who use the Prometheus operator to offer monitoring-as-a-service to developers.
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+* Users who want to manage silences the same way as for services running within the Kubernetes cluster.
+* Users who want a supported Kubernetes way of silences outside the Kubernetes cluster
```
simonpasquier, line 60: not sure I understand this one

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Non-Goals
+
+* This proposal do not cover automatic cleanup of expired Silence CRs
```
simonpasquier, line 64: expired silences are eventually removed by Alertmanager

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+* There is no input validation, which can lead to an invalid silence configuration
+* Teams need to manage expiration on their own
+* If a silence was deleted by accident, it will be permanent.
+
+# Goals
+
+* Provide a way for users to manage Alertmanager silences with Custom Resources (CR).
+* Enable management of Alertmanager silences via centralised repo with e.g. ArgoCD, Flux...
+* Multi tenancy: Silences CR can be using kubernetes rbac.
+* Moving the Giant Swarm silence operator to prometheus operator would provide better community support
```
simonpasquier, line 54: and less fragmentation?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+* Users who want a supported Kubernetes way of silences outside the Kubernetes cluster
+
+# Non-Goals
+
+* This proposal do not cover automatic cleanup of expired Silence CRs
+* Refactoring of the other CRDs is not in scope for the first version
+
+# How
+
+Creating a new cluster-scope Silence CRD that will act as an interface by adding silences via the Alertmanager API.
```
simonpasquier, line 69: hmm I would have expected a namespaced CRD?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: Silence
+metadata:
+  name: my-silence
+  labels:
+    test: value
+spec:
+  selector:
+    app: test-alertmanager
+  expiresAt: "2023-06-01"
```
simonpasquier, line 83: Don't we need a datetime here?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+  matchers:
+    - name: namespace
```
simonpasquier, line 85: we should use the existing `Matcher` type from the AlertmanagerConfig CRD.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+      value: test-ns
+      isRegex: false
+```
+
+The above resource will result in creating a silence in the Alertmanager with label selector,
+`app: test-alertmanager`.
```
simonpasquier, line 91: I don't think that it's the correct model. This approach would allow any user to create silences on any Alertmanager in the cluster.
IMHO we need `.spec.silenceSelector` and `.spec.silenceNamespaceSelector` in the Alertmanager CRD.

simonpasquier (commented):
Thanks for putting this together! It helps a lot to understand the use cases :)

Could you a high-level overview of how the silences will be reconciled (or rather how the silence operator does it right now)?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,112 @@
...
+configurations in the Kubernetes way.
+
+# Why
+
+Prometheus-operator doesn't have a way to automate the management of Alertmanager silences. Users have either been using internal scripts
+or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
+
+* Users using CI/CD jobs to manage silences have reported that this is cumbersome, since they need to either expose
+  Alertmanager via an ingress or using a service account to connect to the cluster, and this is even more complicated
+  in a multi-cluster setup.
```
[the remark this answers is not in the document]
mcbenjemaa: > Another use case where this would be useful is when enabling AlertmanagerConfig custom resources to offer self-service notifications.

what does this mean?

Documentation/proposals/202304-silences.md
mcbenjemaa: I think this is kind of highlight the GitOps approach

Documentation/proposals/202304-silences.md
mcbenjemaa: yes, But the CRs will remain in the cluster.

Documentation/proposals/202304-silences.md
mcbenjemaa: Yes :)

Documentation/proposals/202304-silences.md
mcbenjemaa: Yes, sure.

Documentation/proposals/202304-silences.md
mcbenjemaa: I think we need a format supported by the Alertmanager API. 
So, we will accept a string a validate that's a date or datetime.

Documentation/proposals/202304-silences.md
mcbenjemaa: Yes, I will check it

Documentation/proposals/202304-silences.md
mcbenjemaa: From my point of view, I dont think this is a good idea  `.spec.silenceSelector` and `.spec.silenceNamespaceSelector` in the Alertmanager CRD.

As silences are An API resource and it's not a config like `AlertmanagerConfig`.

Having the selector in the Silence CR, make it easier to select which Alertmanager to Use. 

What do you think?

mcbenjemaa: From a security point of view, I guess we can limit the users by RBAC who can Create/Update Silences.

mcbenjemaa: If you emphasize using this approach, then we need to use both selectors in Alertmanager CRD.
and remove this selector.

Documentation/proposals/202304-silences.md
mcbenjemaa: Exactly

Documentation/proposals/202304-silences.md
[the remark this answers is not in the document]
QuentinBisson: In GS, our original use case is that we needed to deploy the same silences in multiple clusters from a single source of truth.

Deploying flux is definitely easier on multiple clusters than deploying some internal tooling to sync silences and this is all done from within the cluster so using the alertmanager is enough.

In a CI way (unless you use something like tekton for sure) you would need to have your CI aware about all the different alertmanager URIs, but also have it on every silence which can be expensive depending on the number of clusters you have. Also, if the ci cannot access an alertmanager for some reason, you would need to rerun the CI again

Documentation/proposals/202304-silences.md
QuentinBisson: Maybe rephase it users using GitOps instead?

Documentation/proposals/202304-silences.md
QuentinBisson: Well removing them once they expire should be easy but if you use flux, you would need to have an automation in place to remove expired silences. Maybe CR removal can be disabled?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+## Audience
+
+* Users who use the Prometheus operator to offer monitoring-as-a-service to developers.
+* Users who want to manage silences the same way as for services running within the Kubernetes cluster.
+* Users who want a supported Kubernetes way of silences outside the Kubernetes cluster
+
+# Non-Goals
+
+* This proposal do not cover automatic cleanup of expired Silence CRs
+* Refactoring of the other CRDs is not in scope for the first version
```
QuentinBisson, line 65: We can add a read-only alertmanager here behing a flag where the silence operator would automaticall remove silences created by hand

Documentation/proposals/202304-silences.md
QuentinBisson: The date is more of a remnant of our internal implementation but adding time definitely makes sense

Documentation/proposals/202304-silences.md
QuentinBisson: I also thinks this makes more sense because silence creators should not be aware of which alertmanager they want to target

Documentation/proposals/202304-silences.md
TheoBrigitte: The implementation could also keep the silence active when this field is not defined. I have no strong opinion on this.

Documentation/proposals/202304-silences.md
xiu: Oh, sorry, I meant the `Alertmanager` CRD in my first comment 😅 

Another question though, how should one go about using the `Silence` CRD with a remote Alertmanager cluster? Or is that explicitely not covered in the proposal?

mcbenjemaa: Not Covered

Documentation/proposals/202304-silences.md
mcbenjemaa: In the first release, we will not cover automatic cleanup, maybe we implement it, if requested

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,113 @@
...
+* Provide a way for users to manage Alertmanager silences with a Custom Resource Definition (CRD).
+* Enable management of Alertmanager silences via centralised repo with e.g. ArgoCD, Flux...
+* Multi tenancy: Silences CR can be using kubernetes rbac.
+* Moving the Giant Swarm silence operator to prometheus operator would provide better community support and less fragmentation.
+
+## Audience
+
+* Users who use the Prometheus operator to offer monitoring-as-a-service to developers.
+* Users who want to manage silences the same way as for services running within the Kubernetes cluster.
+* Users who want a supported Kubernetes way of silences outside the Kubernetes cluster
```
JoaoBraveCoding, line 60: These two sound the same. AFAIC as @QuentinBisson mentioned one of the goals is for users who manage their clusters strictly through GitOps

Documentation/proposals/202304-silences.md
JoaoBraveCoding: This is an interesting problem! I agree with @simonpasquier but I also see the other side where it would be great if the user did not have to care about which AM they are targeting.
For the sake of discussion, if we were to only have `selector`, would it be feasible for the operator to only accept Silences in the same place as the `PrometheusRule` or would this be too limiting?

Documentation/proposals/202304-silences.md
JoaoBraveCoding: We should probably add this to the `Non-goals` for this proposal

Documentation/proposals/202304-silences.md
xiu: Got it, so:
* If a silence has `expiredAt`, it'll just expire at the time defined and the CR will have to be manually removed
* If a silence doesn't have `expiredAt`, it'll be kept active as long as the CR exists and removed as soon as the CR is gone

I'm not sure how do-able the second point is but I think it'd make sense, especially with GitOps.

Is that what we aim for?

Documentation/proposals/202304-silences.md
mcbenjemaa: yes

Documentation/proposals/202304-silences.md
mcbenjemaa: I think so,

mcbenjemaa: However,  how do we handle silence overwrite?

Documentation/proposals/202304-silences.md
mcbenjemaa: So @QuentinBisson what do you suggest to keep here?

Documentation/proposals/202304-silences.md
mcbenjemaa: Any thoughts here?

Documentation/proposals/202304-silences.md
QuentinBisson: Silence operator would update them. The silence id in alertmanager would be kept in the status

Documentation/proposals/202304-silences.md
mcbenjemaa: I have added it, in the goals section.
Please take a look

Documentation/proposals/202304-silences.md
QuentinBisson: I think Arthur suggestions make sense but maybe we do not need to talk specifically about the silence operator:

Adding support for `Silence` CRD in the Prometheus-operator will provide more flexibility to users in terms of choosing the tool to deploy their silences and will free users from managing a standalone component within the stack. 

That would work for both the silence operator or CI jobs or other tools that creative people came up with

Documentation/proposals/202304-silences.md
mcbenjemaa: I will check this

Documentation/proposals/202304-silences.md
xiu: Great! Can you add that to the proposal? I guess in the How section.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,120 @@
...
+  * [#2398](https://github.com/prometheus-operator/prometheus-operator/issues/2398)
+* Other docs:
+  * n/a
+
+This document describes the creation of `Silence` Custom Resource Definition that defines Alertmanager silences
+configurations in the Kubernetes way.
+
+# Why
+
+Prometheus-operator doesn't have a way to automate the management of Alertmanager silences. Users have either been using internal scripts
```
QuentinBisson, line 16:
```suggestion
Prometheus-operator does not have provide a way to automate the management of Alertmanager silences and there is no current standard/best practice way to do that.
Moreover, setting an automation to manage silences has been reported as cumbersome by multiple users.
```

Documentation/proposals/202304-silences.md
mcbenjemaa: We will just go with selectors in the alertmanager for now: 
```yaml
[...]
spec:
  silenceSelector: ...
  silenceNamespaceSelector: ...
```

Documentation/proposals/202304-silences.md
mcbenjemaa: Added!

Documentation/proposals/202304-silences.md
mcbenjemaa: this is done

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,131 @@
...
+  * n/a
+
+This document describes the creation of `Silence` Custom Resource Definition that defines Alertmanager silences
+configurations in the Kubernetes way.
+
+# Why
+
+Prometheus-operator does not have provide a way to automate the management of Alertmanager silences and there is no current standard/best practice way to do that.
+Moreover, setting an automation to manage silences has been reported as cumbersome by multiple users.
+or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
```
ArthurSens, line 18:
```suggestion
Prometheus-operator does not provide a way to automate the management of Alertmanager silences and there is no current standard/best practice way to do that.
Moreover, setting an automation to manage silences has been reported as cumbersome by multiple users.
or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,131 @@
...
+
+* Users using CI/CD jobs to manage silences have reported that this is cumbersome, since they need to either expose
+  Alertmanager via an ingress or using a service account to connect to the cluster, and this is even more complicated
+  in a multi-cluster setup.
+  However, there's folks that uses a Kubernetes Job to adds Silences,
+  In both situation, a CI/CD pipeline is needed in order to automate the execution, this includes managing the secrets and/or deprecation of code
+  In the Kubernetes Job way, teams need to build and deploy a container and monitor its status, which is an additional effort.
+  CI/CD pipeline is not the best option in multi-cluster setups, teams need to manage k8s access, secrets, and configs.
```
ArthurSens, line 26:
```suggestion
  However, there are folks that use a Kubernetes Job to add Silences,
  In both situations, a CI/CD pipeline is needed in order to automate the execution, this includes managing the secrets and/or deprecation of code
  Using Kubernetes Jobs, teams need to build and deploy a container and monitor its status, which is an additional effort.
  CI/CD pipeline is not the best option in multi-cluster setups, teams need to manage k8s access, secrets, and configs.
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,131 @@
...
+
+![CI/CD K8s Job Approach](../img/CICD-k8s-job.png "CI/CD K8s Job Approach")
+
+* Users that use a standalone operator that implements a Silence CRD are in a better situation,
+  for example users have reported that they are using GitOps to fully manage the life cycle of Silences Custom resources,
```
ArthurSens, line 31:
```suggestion
  for example, users have reported that they are using GitOps to fully manage the life cycle of Silences Custom resources,
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,131 @@
...
+  which in fact brings benefits to the whole team because having a history and revisions in Git as well as constancy to manage the silences in a multi-tenant environment.
+
+Additionally, having a new component in the stack and keeping it maintained is not always ideal (said the folks at [Giant Swarm](https://giantswarm.io) the owners of [silence-operator](https://github.com/giantswarm/silence-operator)),
+because that requires to manage an extra component +CRDs on top of the Prometheus operator.
```
ArthurSens, line 35:
```suggestion
because that requires managing an extra component +CRDs on top of the Prometheus operator.
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,131 @@
...
+Having the Silence CRD as part of prometheus-operator provides users with a better experience: there is no need to install another operator or build a system on top of it to provide that functionality.
+
+Adding support for `Silence` CRD in the Prometheus-operator will provide more flexibility to users in terms of choosing the tool to deploy their silences and will free users from managing a standalone component within the stack.
+
+## Pitfalls of the current solution
+
+Using Alertmanager API Directly comes with drawbacks:
+
+* Teams have to build an automation to add silences in a centralized manner
+* There is no input validation, which can lead to an invalid silence configuration
```
ArthurSens, line 45: I might be wrong, but I think that Alertmanager's API denies POST requests with invalid input. (definition can be seen [here](https://github.com/prometheus/alertmanager/blob/9a8d1f976e12b325ec47b84987a78b7845738be6/api/v2/openapi.yaml#L327-L347))

I understand that even the Alertmanager CLI, `amtool`, has [input validation](https://github.com/prometheus/alertmanager/blob/9a8d1f976e12b325ec47b84987a78b7845738be6/cli/silence_add.go#L73-L86) when adding a silence.

Could you elaborate on what kind of input validation that is lacking today?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,131 @@
...
+* Teams need to manage expiration on their own
+* If a silence was deleted by accident, it will be permanent.
```
ArthurSens, line 47: I believe that it isn't possible to delete silences in Alertmanager actually. All we can do is expire the silence, and we can recreate them afterward. (Recreating is just creating a new one with fields pre-defined)

image

Unless the user is using Alertmanager as a stateless container. If Alermanager is deleted, not the silence, then everything that was stored in Alertmanager's DB is gone, including the silences.

ArthurSens (commented):
One thought just came up!

It is not possible to delete Silences in Alertmanager (unless Alertmanager's storage is deleted). The usual workflow is just expiring silence rules.

Let's say that a user deploys a new Silence CR, and it eventually gets expired. The user now wants this same silence active again. Would he/she delete the previous CR and create a new one? Would he/she just change the expiration date at the deployed CR? What's the user story that you're imagining?

Documentation/proposals/202304-silences.md
mcbenjemaa: But In CI, there's no input validation, that's the intention.

Documentation/proposals/202304-silences.md
mcbenjemaa: YEs, the controller will handle the re-creation.

Documentation/proposals/202304-silences.md
mcbenjemaa: Changed

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,120 @@
...
+  * n/a
+
+This document describes the creation of `Silence` Custom Resource Definition that defines Alertmanager silences
+configurations in the Kubernetes way.
+
+# Why
+
+Prometheus-operator doesn't have a way to automate the management of Alertmanager silences. Users have either been using internal scripts
+or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
+
```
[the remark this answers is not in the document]
mcbenjemaa: Added!

mcbenjemaa:
Hey, Don't hesitate to add your reviews

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+    test: value
+spec:
+  expiresAt: "2023-06-01 10:00"
+  matchers:
+    - matchType: "="
+      name: namespace
+      value: test-ns
+```
+
+The above resource will result in creating a silence in the Alertmanager that has the appropriate selector.
```
mrueg, line 96: Will Silence custom resources only be applied on creation or regularly "pushed" to AM's API?
Is there a reconciliation loop (i.e. if there's a stateless AM and the AM pod gets recreated)? What's the suggested interval for it?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+
```
mrueg, line 97: What happens if a Silence CR cannot create a silence in AM? Can you describe certain error classes (e.g. invalid configuration, AM unavailable, etc.) and how they will be handled?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+This example doesn't list all the fields that are offered by Alertmanager. The implementation of all the fields will be
+done in an iterative process and as such, the expectation is not for all of them to be implemented in the first version.
+
+Also, to help select `Silences`, a new field will be added to the Alertmanager CRD:
+
+```yaml
+[...]
+spec:
+  silenceSelector: ...
```
mrueg, line 115: If this is selecting silences by labels, is it better to call it silenceLabelSelector ?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+  silenceNamespaceSelector: ...
+```
+
+Additionally, the silence-controller will handle `Silences` overwrite.
+after successfully created a silence, the silence id will be kept in the status sub-resource:
+
+```yaml
+[...]
+status:
+  silenceID: ...
```
mrueg, line 125: should this also `CreatedAt` `UpdatedAt` fields populated with AMdata, so we know that the resource got added to AM at a certain point in time?

Documentation/proposals/202304-silences.md
mcbenjemaa: yeah, there's reconciliation loop

mcbenjemaa: this will defined in the design document later

Documentation/proposals/202304-silences.md
mcbenjemaa: Yes, the status sub resource, will has the error messages, 
Again, this is will be defined in the design document

Documentation/proposals/202304-silences.md
mcbenjemaa: Yes, this will be handled in the Design document.

mcbenjemaa:
any more reviews?

herewasmike:
Hey,

I didn't like silence-operator that guys from giantswarm created so I ended up writing my own.

Some input here: 

You don't need expiresAt at your CRD. It should rather be an configuration of Silence-operator so it can renew silences for which custom resource is still deployed.


I used KOPF so the code is in python but I think it should be clear

`    if state == "active":
        end = datetime.strptime(endsAt, "%Y-%m-%dT%H:%M:%S.%f%z")
        now = datetime.now(timezone.utc)
        diff = (end - now).total_seconds()
        logger.debug(f"{diff} seconds left before expiration")

        if diff < (RECONCILATION_INTERVAL * 2):
            logger.info(f"Less than 2 intervals left on silence. Will renew")
            await apply_silence(spec, status, patch, logger, kwargs=kwargs)
`

This allows me to ensure that if my operator would go rouge for some reason my silences will be expired and I'll notice it.

My CRD schema looks like this

`      schema:
        openAPIV3Schema:
          type: object
          x-kubernetes-preserve-unknown-fields: true
          properties:
            status:
              type: object
              x-kubernetes-preserve-unknown-fields: true
            spec:
              type: object
              properties:
                comment:
                  type: string
                  description: Silence description to be put in comment
                matchers:
                  description: List of alertmanager label matchers for alert silence
                  items:
                    description: An alertmanager matcher object
                    properties:
                      isEqual:
                        description: Either positive or negative operator
                        type: boolean
                        default: true
                      isRegex:
                        description: Strict or regex operator
                        type: boolean
                        default: false
                      name:
                        description: Label name to use with operator
                        type: string
                      value:
                        description: Label value to use with operator
                        type: string
                    required:
                      - name
                      - value
                    type: object
                  type: array
`

mcbenjemaa:
This is ready for merge.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+Also, to help select `Silences`, a new field will be added to the Alertmanager CRD:
+
+```yaml
+[...]
+spec:
+  silenceSelector: ...
+  silenceNamespaceSelector: ...
+```
+
+Additionally, the silence-controller will handle `Silences` overwrite.
```
xiu, line 119:
```suggestion
Additionally, the silence-controller will handle `Silences` overwrites.
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+after successfully created a silence, the silence id will be kept in the status sub-resource:
```
xiu, line 120:
```suggestion
After having successfully created a silence, the silence id will be kept in the status sub-resource:
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Non-Goals
+
+* This proposal do not cover automatic cleanup of expired Silence CRs
+* Refactoring of the other CRDs is not in scope for the first version
+* Adding silences to a remote Alertmanager cluster
+
+# How
+
+Creating a new namespaced Silence CRD that will act as an interface by adding silences via the Alertmanager API.
```
xiu, line 76:
```suggestion
We will create a new namespaced Silence CRD that will act as an interface by adding silences via the Alertmanager API.
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+* There is no input validation in CI, which can lead to an invalid silence configuration
+* Teams need to manage expiration on their own
+* If a silence was deleted by accident, it will be permanent.
+
+# Goals
+
+* Provide a way for users to manage Alertmanager silences with a Custom Resource Definition (CRD).
+* Enable management of Alertmanager silences via centralised repo with e.g. ArgoCD, Flux...
+* Multi tenancy: Silences CR can be using kubernetes rbac.
+* Moving the Giant Swarm silence operator to prometheus operator would provide better community support and less fragmentation.
```
xiu, line 59:
```suggestion
* Provide better community support and reduce fragmentation as Giant Swarm's silence operator are added to prometheus-operator
```

xiu (commented):
Sorry for being late 😅 I noticed a few nits but all in all it looks good to me!

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+* Other docs:
+  * n/a
+
+This document describes the creation of `Silence` Custom Resource Definition that defines Alertmanager silences.
+
+# Why
+
+Prometheus-operator does not provide a way to automate the management of Alertmanager silences and there is no current standard/best practice way to do that.
+Moreover, setting an automation to manage silences has been reported as cumbersome by multiple users.
+or an additional operator that does the job e.g. [silence-operator](https://github.com/giantswarm/silence-operator).
```
simonpasquier, line 17: this part doesn't read properly with the previous line.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+Having the Silence CRD as part of prometheus-operator provides users with a better experience: there is no need to install another operator or build a system on top of it to provide that functionality.
+
+Adding support for `Silence` CRD in the Prometheus-operator will provide more flexibility to users in terms of choosing the tool to deploy their silences and will free users from managing a standalone component within the stack.
+
+## Pitfalls of the current solution
+
+Using Alertmanager API Directly comes with drawbacks:
+
+* Teams have to build an automation to add silences in a centralized manner
+* There is no input validation in CI, which can lead to an invalid silence configuration
```
simonpasquier, line 50: The API will validate the request though so I'm not sure that it's a strong point.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+* Teams need to manage expiration on their own
+* If a silence was deleted by accident, it will be permanent.
+
+# Goals
+
+* Provide a way for users to manage Alertmanager silences with a Custom Resource Definition (CRD).
+* Enable management of Alertmanager silences via centralised repo with e.g. ArgoCD, Flux...
+* Multi tenancy: Silences CR can be using kubernetes rbac.
```
simonpasquier, line 58: (suggestion)
```suggestion
* Access to the Silence custom resources can be controlled by Kubernetes RBAC authorization.
```

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+
+* Users who use the Prometheus operator to offer monitoring-as-a-service to developers.
+* Users who want to manage silences the same way as for services running within the Kubernetes cluster.
+* Users who want a supported GitOps way of silences outside the Kubernetes cluster
+
+# Non-Goals
+
+* This proposal do not cover automatic cleanup of expired Silence CRs
+* Refactoring of the other CRDs is not in scope for the first version
+* Adding silences to a remote Alertmanager cluster
```
simonpasquier, line 72: not sure to understand this one. Is it about creating silences on an Alertmanager cluster not managed by the operator?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+
+```yaml
+[...]
+spec:
+  silenceSelector: ...
+  silenceNamespaceSelector: ...
+```
+
+Additionally, the silence-controller will handle `Silences` overwrites.
+After having successfully created a silence, the silence id will be kept in the status sub-resource:
```
simonpasquier, line 120: we could have the same silence being selected by different Alertmanager resources so this approach will not work out of the box.

simonpasquier (commented):
Sorry for the lag on this one. I reviewed the latest version. Can you add more details on how the silence resources will be reconciled in the Alertmanager API?

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+  - If a silence has `expiredAt` defined, it'll expire at the time defined and the CR will have to be manually removed.
+  - If a silence doesn't have `expiredAt`, it'll be kept active as long as the CR exists and removed as soon as the CR is gone
+* `matchers` field corresponds to the Alertmanager silence matchers each of which consists of:
+  - `name` - name of tag on an alert to match
+  - `value` - fixed string or expression to match against the value of the tag named by name above on an alert
+  - `matchType` - Match operation available with AlertManager >= v0.22.0 and takes precedence over Regex (deprecated) if non-empty.
+  - `regex` - Whether to match on equality (false) or regular-expression (true). Deprecated as of AlertManager >= v0.22.0 where a user should use MatchType instead.
+
+This example doesn't list all the fields that are offered by Alertmanager. The implementation of all the fields will be
+done in an iterative process and as such, the expectation is not for all of them to be implemented in the first version.
```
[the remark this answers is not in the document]
mcbenjemaa: comment field in the silence?

Documentation/proposals/202304-silences.md
mcbenjemaa: but. we are going to add `silenceSelector` and `silenceNamespaceSelector` to the alertmanager CRD

Documentation/proposals/202304-silences.md
mcbenjemaa: yes, we will not cover adding silences to a remote alertmanager, at least not in the same cluster.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+Additionally, having a new component in the stack and keeping it maintained is not always ideal (said the folks at [Giant Swarm](https://giantswarm.io) the owners of [silence-operator](https://github.com/giantswarm/silence-operator)),
+because that requires managing an extra component +CRDs on top of the Prometheus operator.
+Moreover, this project doesn't have a lot of community support, and it's reached by small amount of users.
+Having the Silence CRD as part of prometheus-operator provides users with a better experience: there is no need to install another operator or build a system on top of it to provide that functionality.
+
+Adding support for `Silence` CRD in the Prometheus-operator will provide more flexibility to users in terms of choosing the tool to deploy their silences and will free users from managing a standalone component within the stack.
+
+## Pitfalls of the current solution
+
+Using Alertmanager API Directly comes with drawbacks:
```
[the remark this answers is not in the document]
mcbenjemaa: sure.

Documentation/proposals/202304-silences.md
mcbenjemaa: This means if you're using CI, it doesn't support validation prior to sending it to the API.
pre-validation is a benefit in this case.

Documentation/proposals/202304-silences.md
```diff
@@ -0,0 +1,136 @@
...
+
+* Teams have to build an automation to add silences in a centralized manner
+* There is no input validation in CI, which can lead to an invalid silence configuration
+* Teams need to manage expiration on their own
+* If a silence was deleted by accident, it will be permanent.
```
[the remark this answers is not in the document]
mcbenjemaa: so we need to delete this point or rephrase it?

Documentation/proposals/202304-silences.md
mcbenjemaa: let me rephrase it.

Documentation/proposals/202304-silences.md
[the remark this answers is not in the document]
herewasmike: Can we add CRD schema to the proposal?
I've posted mine above, I think it would be more transparent for everyone

Documentation/proposals/202304-silences.md
herewasmike: It can actually be easily achieved with optional field in the CRD for alertmanager url/auth/etc.

If it's not provided operator would take default one from config

Documentation/proposals/202304-silences.md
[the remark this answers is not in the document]
xiu: I think we should leave the schema to the implementation. We can definitely put some example uses but they shouldn't mandate how the API will look like later.

smrutimandal:
Can we reopen this PR, please

mcbenjemaa:
For sure)

sebastiangaiser:
@mcbenjemaa @simonpasquier can you please reopen the draft

Issue #5458, opened 2023-03-28
Reduce duplication on `UpdateStatus` of PrometheusAgent and server


ArthurSens:
same remarks here about duplication.

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/5385#discussion_r1140403295_
            
---

The method is pretty much the same for Agent and server, with a small difference when doing type casting. 

We can probably have reduce code duplication

PR #5493, opened 2023-04-12
Reduce duplication on UpdateStatus

mcbenjemaa:
## Description

This PR Adds a function `GetStatus` that is used to fetch the Status subresource for both the Prometheus agent and server. to reduce duplication.

## Type of change

## Changelog entry

```release-note

```

Closes https://github.com/prometheus-operator/prometheus-operator/issues/5458


ArthurSens (commented):
Oh nice, thanks for raising the PR!

It was just a super quick look and haven't seen the code just yet, but I've noticed that you've created a new directory `common`. So far we've been using the upper level `pkg/prometheus` for code that is used by both server and agent. Would it be enough to move the code from `pkg/prometheus/common` to `pkg/prometheus`?

mcbenjemaa:
Yeah, make sense

mcbenjemaa:
Done

pkg/prometheus/operator.go
```diff
@@ -15,21 +15,43 @@
...
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 var prometheusKeyInShardStatefulSet = regexp.MustCompile("^(.+)/prometheus-(.+)-shard-[1-9][0-9]*$")
 var prometheusKeyInStatefulSet = regexp.MustCompile("^(.+)/prometheus-(.+)$")
 
+type Config struct {
```
JoaoBraveCoding, line 43: This can definitely be done on a follow-up, but I have the feeling we should benefit more from having an `Operator` interface that then both agent and server would comply with. Food for thought, would like to know the opinion of others.

More functions will definitely be shared among the two operators and I don't think that a `Config` struct would be the best way to pass function arguments around

JoaoBraveCoding (approved):
lgtm, great job! Just a comment for the future from my side

pkg/prometheus/operator.go
mcbenjemaa: So what’s the best option now?

JoaoBraveCoding: > Having a shared interface can be useful now but I don't think it will stay for long 😬

I'm not sure I fully agree since in this interface we would only have methods that are indeed common to both Prom and PromAgen (in the many deployment models), for instance, we will always want to know the Status of a CR regardless of how it is deployed. Still, this was just a possible idea, I don't want to block this PR :) 

> So what’s the best option now?

I would leave the definitive answer to @simonpasquier. Personally, I think I'm more in favour of having the arguments explicit for now, but as I said, this is a slight preference as I don't think a struct called `Config` is a good attraction in this instance.

ArthurSens: Yeah, sorry for sounding like a blocker here. 

What I meant with different statuses with agent and server is that with, e.g. Daemonset or Sidecars, the `PrometheusStatus` interface won't make much sense to the Agent anymore since we're gonna need to tell the deployment model, the number of replicas might not be that interesting in some cases, etc.

We can definitely worry about it only when the time comes, for sure :)

ArthurSens (approved):

pkg/prometheus/operator.go
mcbenjemaa: @simonpasquier I need an eye here.

mcbenjemaa:
I rebased, this now is ready

mcbenjemaa:
flaky test

ArthurSens:
> flaky test

Indeed, I've re-triggered the job

ArthurSens:
I'll let this open just in case someone else from the team still wants to take a look, otherwise I'll merge in the following days :)

pkg/prometheus/operator.go
```diff
@@ -106,3 +128,100 @@ func ValidateRemoteWriteSpec(spec monitoringv1.RemoteWriteSpec) error {
...
+
+	for shard := range ExpectedStatefulSetShardNames(config.Instance) {
+		ssetName := KeyToStatefulSetKey(config.Instance, config.Key, shard)
+		logger := log.With(logger, "statefulset", ssetName, "shard", shard)
+
+		obj, err := config.SsetInfs.Get(ssetName)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				// Object not yet in the store or already deleted.
+				level.Info(logger).Log("msg", "not found")
```
simonpasquier, line 169: same here

pkg/prometheus/operator.go
```diff
@@ -106,3 +128,100 @@ func ValidateRemoteWriteSpec(spec monitoringv1.RemoteWriteSpec) error {
 
 	return nil
 }
+
+// GetStatus is a helper function that retrieves the status subresource of the object identified by the given
+// key.
+func GetStatus(ctx context.Context, config Config) (*monitoringv1.PrometheusStatus, error) {
```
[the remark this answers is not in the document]
mcbenjemaa: The Config struct has all parameters needed. 
Can you explain what do you suggest?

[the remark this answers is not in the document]
mcbenjemaa: This is done, it's a good point. thank you

mcbenjemaa:
This is now ready for merge.

pkg/prometheus/operator.go
```diff
@@ -106,3 +121,95 @@ func ValidateRemoteWriteSpec(spec monitoringv1.RemoteWriteSpec) error {
 
 	return nil
 }
+
+// Process is a helper function that retrieves the status subresource of the object identified by the given
+// key.
```
JoaoBraveCoding, line 126: nit: WDYT like this?
```suggestion
// Process will determine the Status of a Prometheus resource (server or agent) depending on its current state in the cluster
```

JoaoBraveCoding (commented):
small nit but from my side lgtm!

mcbenjemaa:
Can we merge this? @prometheus-operator-reviewers
/ @simonpasquier

ArthurSens (approved):
Thanks for tackling this one!

xiu (approved):
LGTM

Issue #5495, opened 2023-04-12
Support different PrometheusAgent deployment models


ArthurSens:
Now that an initial version of PrometheusAgent was introduced with the `v1alpha1` API, we can start thinking about extending such API to support different Deployment Patterns.

Here are a few that I can think of:

* StatefulSet (what we have today)
* Deamonset - Each agent is responsible for scraping metrics from their own nodes.
* Namespaced - Each agent is responsible for scraping metrics from their own namespace
* Deployment - For use cases where the Agent WAL isn't important and can be lost during restarts

Some implementations sounds harder than others, but I believe a design document talking about for to start this would be great already 😄

ArthurSens:
The OpenTelemetryCollector operator can probably be a good reference for the design doc

https://opentelemetry.io/docs/collector/deployment/
https://opentelemetry.io/docs/collector/scaling/

JoaoBraveCoding:
Isn't namespaced already possible today? 🤔 If you leverage selectors I think it might be possible, but I might be missing something. 

Also, I think something worth considering in these efforts is the performance impact that some of these modes might have on the k8s api-server, namely the discovery API

ArthurSens:
> Isn't namespaced already possible today? 🤔 If you leverage selectors I think it might be possible, but I might be missing something.

Yeah, you're probably right now that I think about it 🤔 

> Also, I think something worth considering in these efforts is the performance impact that some of these modes might have on the k8s api-server, namely the discovery API

💯

bwplotka:
Thanks @ArthurSens !

We (Google Managed Prometheus Team) would love to contribute a design (and maybe impl) of Daemon Set approach - it has unique pros/cons but it shines when it comes to scaling with the load 💪🏽 

Implemented and available [here](https://github.com/GoogleCloudPlatform/prometheus-engine) for now 🤗

ArthurSens:
> Thanks @ArthurSens !
> 
> We (Google Managed Prometheus Team) would love to contribute a design (and maybe impl) of Daemon Set approach - it has unique pros/cons but it shines when it comes to scaling with the load 💪🏽
> 
> Implemented and available [here](https://github.com/GoogleCloudPlatform/prometheus-engine) for now 🤗

That would awesome! Even if it's just a design doc it already helps a ton! We're currently storing design documents [here](https://github.com/prometheus-operator/prometheus-operator/tree/main/Documentation/designs).

[...]

PR #5497, opened 2023-04-13
Add proposal for deploying PrometheusAgent as sidecar

ArthurSens:
## Description

Related to one of the deployment models listed in #5495, this PR adds a design proposal for running Prometheus Agents as sidecars.

## Type of change

## Changelog entry

```release-note

```


Documentation/designs/agent-deployment-pattern-sidecar.md
```diff
@@ -0,0 +1,116 @@
...
+
+This document proposes a new deployment model where Prometheus-Operator injects Prometheus agents as a sidecar container (and Prometheus config reloader) to pods that needs to be scrapped. With a sidecar, we tackle both problems mentioned above:
+
+> * Load on Kubernetes API won't exist since it's not needed anymore. Prometheus will scrape containers from the same pod through their shared network interface and scrape configuration can be declared via pod annotations.
+> * A sudden cardinality spike will not affect the whole monitoring system. In a worst-case scenario, it will fail a single pod.
+
+A common pattern used with Prometheus's Kubernetes service discovery is the [usage of annotation to declaratively tell Prometheus which endpoints need to be scraped](https://www.acagroup.be/en/blog/auto-discovery-of-kubernetes-endpoint-services-prometheus/). From a [code search at Github](https://github.com/search?q=prometheus.io%2Fscrape%3A+%22true%22&type=code) for `prometheus.io/scrape: "true"`, we can tell that this approach has good adoption already. To not conflict with the already commonly used annotation, we can start with our own, but with a very similar approach.
```
matthiasr, line 28: I kind of see this as the "not using the operator" pattern.

This makes me wonder what it would take to allow configuring the sidecar injection with PodMonitors? In other words, inject the sidecar into any (new?) pod that matches any PodMonitor. In principle this could even extend to ServiceMonitors, by resolving the label matchers on the service.

A big benefit of this would be that it makes the choice of deployment model independent from the choice of configuration. As an observability team, one could keep consuming the same CRs (which may be owned by service teams), while transparently moving between agent deployment models, or even running multiple variants in parallel. The annotations approach would require a large coordinated effort across the organization to make sure the annotations are there and represent the same semantic intentions.

matthiasr: An open question for me would be what to do when a monitor changes after a pod has started. I don't know how much flexibility Kubernetes has around that today. Theoretically this is also a problem with annotations, although in practice it's easier to say "don't do that" when it's so closely tied to the pod.

Documentation/designs/agent-deployment-pattern-sidecar.md
```diff
@@ -0,0 +1,116 @@
...
+
+There is two ways to change Prometheus configuration now, 1) by changing annotation on the pod and 2) by changing the remote-write field in PrometheusAgent CRD. The first one will only trigger a hot reload for the involved pod, but the latter has the pottention to trigger millions of hot reloads, depending on the scale of the cluster. 
+
+While there is no research regarding the config-reloader efficiency, this particular container might become problematic for huge scale environments.
+
+### WAL not optimized for small environments
+
+Prometheus Write-Ahead-log(WAL) is stored as a sequence of numbered files with 128MiB each by default. Which means that, by default, at least 128MiB is needed for running Prometheus Agent if we ignore every other part of Prometheus. Using a sidecar, we're optimizing for horizontal scale and 128MiB might be much more than necessary to store metrics from a single Pod.
+
+### Lack of High-Availability setup
```
matthiasr, line 104: IMO this is fine. A sidecar is tied to the lifecycle of the pod, and comes and goes. If the agent sidecar is broken, the pod is broken. If the node breaks, all sidecars will break, but so will the application container.

Documentation/designs/agent-deployment-pattern-sidecar.md
matthiasr: The conceptual shift from the way you write in the implementation is that I don't see PodMonitor as _configuring Prometheus_, I see it as declaring the intention to monitor a pod. In that sense, it doesn't matter which objects or files need to be changed to fulfill that intention.

Documentation/designs/agent-deployment-pattern-sidecar.md
```diff
@@ -0,0 +1,116 @@
...
+Another problem is that one or more containers can be updated to a problematic version that causes a [Cardinality Spike](https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/). Depending on the proportion of the spike, it is possible that a container could single-handedly crash the monitoring system of the whole cluster.
+
+![Traditional Deployment Pattern](assets/agent-deployment-pattern-sidecar/traditional-deployment-pattern.png "Traditional Deployment Pattern").
+
+## Proposal
+
+This document proposes a new deployment model where Prometheus-Operator injects Prometheus agents as a sidecar container (and Prometheus config reloader) to pods that needs to be scrapped. With a sidecar, we tackle both problems mentioned above:
+
+> * Load on Kubernetes API won't exist since it's not needed anymore. Prometheus will scrape containers from the same pod through their shared network interface and scrape configuration can be declared via pod annotations.
+> * A sudden cardinality spike will not affect the whole monitoring system. In a worst-case scenario, it will fail a single pod.
```
matthiasr, line 26: I think this isn't entirely true. While an agent monitoring only a single pod is less likely to buckle under a cardinality increase, by removing this bottleneck we expose the next layer up to the same increase. That may or may not have a higher overall capacity, but we massively increased the capacity of the layer feeding into it, so we may be exposing it to _more_ cardinality issues.

To illustrate, consider a multi cluster setup with an agent setup per cluster, feeding into a global observability layer. Previously, deploying a cardinality explosion to one cluster might overload the agent there, signalling that there is a problem with this deploy. Now, we create the capacity for this cluster alone to blow up global observability.

How exactly this plays out really depends on the details, so I don't think this model would be _worse_. It opens up a different set of tradeoffs which is generally helpful.

Documentation/designs/agent-deployment-pattern-sidecar.md
xiu: I like the idea of using `PodMonitor` for simple deployments, would there be benefits in using annotations over changing `PodMonitor`? 
I can think of the following: 
* having the same Agent over multiple sets of pods and not an Agent per selector.
* having a platform team provide Agents for different use-cases and users select the ones they need and thus removing the burden of configuring the agent from the users

xiu (commented):
Nice write up! Just a nit and comment.

Documentation/designs/agent-deployment-pattern-sidecar.md
ArthurSens: Oh wow, I don't know how I didn't think of that before. Indeed it sounds much more aligned with the way we do things right now. 

We should be able to find and inject pods from Pod/ServiceMonitorSelectors and even inject blockbox-exproters with ProbeSelectors.

The only one that will probably be odd is the upcoming ScrapeConfigSelector, but we can defer the decision with this one to a future iteraction

ArthurSens:
Proposal updated with a better use of the current existing CRs and fixing a few grammatical errors :)

Documentation/designs/agent-deployment-pattern-sidecar.md
nicolastakashi: Actually the HA of Monitoring for the given application is based on the number of replicas for this application.
I mean if the service is running with 2 replicas you already have Prometheus HA.

matthiasr (commented):
One thing that is not clear to me from the proposal (but maybe documented elsewhere already?) is how the injection itself would work. Presumably this will be a mutating webhook? I think it would be helpful to explore what exactly that would hook into, and what the possible failure modes are. When the operator is failing, should pod admissions fail? How would you avoid the issue of updating the operator itself? If injection is best-effort, how can users learn about unmonitored pods and follow up? Or better, how could this heal itself after the fact?

ArthurSens:
> One thing that is not clear to me from the proposal (but maybe documented elsewhere already?) is how the injection itself would work. Presumably this will be a mutating webhook? I think it would be helpful to explore what exactly that would hook into, and what the possible failure modes are. When the operator is failing, should pod admissions fail? How would you avoid the issue of updating the operator itself? If injection is best-effort, how can users learn about unmonitored pods and follow up? Or better, how could this heal itself after the fact?

It was not documented elsewhere actually. In this first proposal, I was more interested in the users' perspective rather than implementation. I do agree that documenting an implementation path will definitely help the developer's perspective.

Implementation-wise, I was planning on checking how OtelCollector-Operator does this and ~copy~get inspired from them but I haven't had time to check that yet. I do believe that a mutating webhook makes sense. 

I'll try to extend the document with implementation details 🙂

nicolastakashi:
Not that this changes the initial idea, but do we have any idea if there is anyone in the community using this approach on their own?

I say this because we see many movements in the CNCF universe not opting for a sidecar, like many service meshs are now trying to move away from a sidecar.

I think that maybe interesting check if this mode will really solve some corner user case.

simonpasquier:
First of all, thanks for starting the discussion @ArthurSens! I love seeing all the feedback :heart: 
I had started to comment on your proposal but unfortunately my write-up got lost. I can provide some details on how the implementation would work. It would indeed require a mutating webhook. The "good" thing is that we already have a standalone webhook service for validation that could be extended. The caveat like with any webhook service is that you have to ensure proper HA (e.g. deploy at least 2 replicas with anti-affinity rules and PDB): it's doable but not for the faints of heart :)
I also agree that we need to assess how useful it would for the community but let's continue the discussion!

ArthurSens:
Turning into a draft until I add implementation details

ArthurSens:
I've added a few extra lines regarding pod mutation with Admission Webhooks.

I'm not entirely familiar with this part of Kubernetes, so feedback is more than welcome!

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
```diff
@@ -31,6 +38,77 @@ With a visualization:
...
+// ServiceMonitor/PodMonitor/Probe selectors.
+// If it should be injected, it returns the agent responsible for the injection.
+func shoulPodBeInjected(pod corev1.Pod) (monitoringv1alpha1.PrometheusAgent, error){}
+
+// generatePatchesForSidecarInjection generates the patches necessary to inject the agent sidecar into the container
+// The agent parameter is used to configure extra container flags, remote-write configs and alikes
+func generatePatchesForSidecarInjection(raw []bytes, agent monitoringv1alpha1.PrometheusAgent)
+```
+
+The function
```
matthiasr, line 110: This sentence

matthiasr: … seems to be missing

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
```diff
@@ -31,6 +38,77 @@ With a visualization:
...
 
+## Implementation
+
+Following what other operators, e.g. OtelCollector and Istio, are doing already, we'll use [Admission Webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) to mutate pods with the Prometheus Agent sidecar.
+
+The Prometheus-Operator project has implemented an admission webhook already. All that is left to do is add another route for pod mutation.
+
+In `pkg/admission/admission.go`, we'll do the following changes:
+
+*code redacted for improved reaadability*
```
matthiasr, line 49:
```suggestion
*code edited for improved readability*
```

matthiasr: Tbh I don't know if you really need to show code here. The description is clear enough to me.

matthiasr (approved):

matthiasr:
> I contacted the Prometheus mailing list last week but got no response.

I came here from that message 😀

Documentation/designs/agent-deployment-pattern-sidecar.md
matthiasr: My first thought was "eh just ignore it" but my second thought was "hold on this could actually be really useful". The scrape config for a sidecar would be fairly static (mostly targeting localhost) so it would be quite easy to manage and at the same time offer a good amount of flexibility. One thing I am wondering about is how one could provide "pod metadata" to the sidecar when supplying a custom scrape config, but maybe that is sufficiently covered by external labels.

matthiasr: > users select the ones they need and thus removing the burden of configuring the agent from the users

In first approximation, I don't think they should even care at this level. To me, the ideal model is that users provide a PodMonitor that describes how their pod exposes metrics, and the platform team does ✨ magic ✨ to make those metrics appear in the query interface. If the platform team wants to move from a central statefulset to a daemonset to a sidecar, nothing should change to the users.

The world isn't ideal, so sometimes you do need to give a little bit more control – this is still possible by labeling the PodMonitors and restricting the podMonitorSelectors accordingly. You can make this explicit (label `agentModel=sidecar`), or better, intention based. For example, the pod monitor could be labeled `size=small` for pods that are small (where a sidecar would add significant overhead), but the platform team can decide whether to eat that cost dor uniformity, or split this class of PodMonitors out to be covered by a statefulset.

ArthurSens: > My first thought was "eh just ignore it" but my second thought was "hold on this could actually be really useful". The scrape config for a sidecar would be fairly static (mostly targeting localhost) so it would be quite easy to manage and at the same time offer a good amount of flexibility. One thing I am wondering about is how one could provide "pod metadata" to the sidecar when supplying a custom scrape config, but maybe that is sufficiently covered by external labels.

Pod/Service monitors have labelSelector for pod discovery, something that ScrapeConfig doesn't have. If someone is relying entirely on the ScrapeConfig CRD to inject the agent, I believe we'll need to pod annotations that were in the original proposal

Documentation/designs/agent-deployment-pattern-sidecar.md
ArthurSens: I see your point, but that was not what I had in mind in this statement 😅 

The goal of the sidecar injection is to scale horizontally, making sure that a cardinality increase that is capable of OOMing Prometheus will prevent us from retrieving metrics from a single pod, but not from the entire cluster.

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
ArthurSens: whoopsiee 😝

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
ArthurSens: I wanted to add it to the doc because we're not really following the conventional way when handling admission webhooks. We have a container that runs separated from the operator.

I was just thinking that providing a code snippet for whoever will implement this might help :)

ArthurSens:
I've refactored the proposal to follow what is documented in CONTRIBUTING.md, which made me realize that I don't have an action plan yet 😅

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
```diff
@@ -0,0 +1,151 @@
...
+// ServiceMonitor/PodMonitor/Probe selectors.
+// If it should be injected, it returns the agents responsible for the injection.
+func shoulPodBeInjected(pod corev1.Pod) ([]monitoringv1alpha1.PrometheusAgent, error) {}
+
+// generatePatchesForSidecarInjection generates the patches necessary to inject the agent sidecar into the container
+// The agent parameter is used to configure extra container flags, remote-write configs and alikes
+func generatePatchesForSidecarInjection(raw []bytes, agents []monitoringv1alpha1.PrometheusAgent)
+```
+
+### Caveats
```
simonpasquier, line 120: I'd also mention
* Increased permissions for the operator (it can now mutate pods)
* Opportunity to block pod creation when the admission webhook service isn't available

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### Lack of High-Availability setup
+
+With the problem that Prometheus is not optimized for very small environments, injecting 2 sidecars per Pod sounds like a big waste of resources. However, with only 1 sidecar HA Prometheus won't be an option.
+
+With that said, having an HA Prometheus in the traditional deployment pattern seems to be more critical than the sidecar approach. That's because with Prometheus fails in the first approach we lose the monitoring stack for the whole cluster, while with the latter we just lose metrics from a pod.
+
+## Alternatives
+
+A common pattern used with Prometheus's Kubernetes service discovery is the [usage of annotation to declaratively tell Prometheus which endpoints need to be scraped](https://www.acagroup.be/en/blog/auto-discovery-of-kubernetes-endpoint-services-prometheus/). From a [code search at Github](https://github.com/search?q=prometheus.io%2Fscrape%3A+%22true%22&type=code) for `prometheus.io/scrape: "true"`, we can tell that this approach has good adoption already. To not conflict with the already commonly used annotation, we can start with our own, but with a very similar approach.
```
simonpasquier, line 138: I'm not sure to understand how this is an alternative?

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
```diff
@@ -0,0 +1,151 @@
...
+* Implement other deployment patterns, e.g. daemonset.
+
+## How
+
+This document proposes a new deployment model where Prometheus-Operator injects Prometheus agents as a sidecar container (and Prometheus config reloader) to pods that need to be scrapped. With a sidecar, we tackle both problems mentioned above:
+
+> * Load on Kubernetes API won't exist since it's not needed anymore. Prometheus will scrape containers from the same pod through their shared network interface and scrape configuration can be declared via pod annotations.
+> * A sudden cardinality spike will not affect the whole monitoring system. In a worst-case scenario, it will fail a single pod.
+
+The existing PrometheusAgent CRD would be extended with a new field called `mode`, which can be one of two values(for now): [`statefulset`, `sidecar`], with `statefulset` as default. If `mode` is set to `sidecar`, Prometheus-Operator won't deploy any Prometheus agents initially. Instead, it will use `ServiceMonitor`, `PodMonitor`, and `Probes` to find pods that need to be scrapped through then inject Prometheus Agent as a sidecar to the pods found through the label selectors found in the mentioned resources.
```
simonpasquier, line 47: More details on how the operator will know which resources it should reconcile and how they will materialize in the resulting Prometheus config?

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
ArthurSens: Regarding the webhook blocking pod creation, I believe we can choose to Allow pods in case of unavailability

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
ArthurSens: Instead of using ServiceMonitor/PodMonitor to select pods that need to be injected, we could look for annotations within Pods.

I've added extra info to the proposal

Documentation/proposals/202304-agent-deployment-pattern-sidecar.md
ArthurSens: That's a good point, we probably want a different API for the Status, I still need to think about how it should look

ArthurSens:
Commenting just to remove the stale label.

Just for transparency, we still haven't merged this because of a lack of interest from the community so far. It doesn't make sense to invest time and increased complexity if this is not going to be useful to someone

If you stumble across this PR and you think running Prometheus Agent as a sidecar is useful to you/your organization, please let us know :)

PR #5521, opened 2023-04-21
chore: add +optional markers

simonpasquier:
## Description

To follow https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

## Type of change

## Changelog entry

```release-note

```


ArthurSens (approved):
Looking at the link you provided I see
```
They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).
```

Do you know if `Duration` provide a built-in nil value?

JoaoBraveCoding:
>  Do you know if Duration provide a built-in nil value?

IIUC, no since it's a string, so it should be a pointer. But I'm also not sure we can change it 🤔 

Seems like the docs part of the CI is not happy

ArthurSens:
@simonpasquier do you want to merge this one as it is and take a look at adding pointers to strings later? Or should we wait for the pointers?

I'm fine with both, just need to open an issue for the former if that's how we go forward

simonpasquier:
I'll update the PR to have *Duration fields.

simonpasquier:
The PR has been updated to use pointers instead of values.

PR #5532, opened 2023-04-27
docs(storage): add example for getting prometheus res name

apgapg:
## Description

This PR aims at adding an example command to get the actual name of the Prometheus resource. Begineers may find it difficult to understand.

## Type of change

## Changelog entry

```release-note
docs(storage): add example for getting prometheus res name
```

### DIFF
--- a/Documentation/user-guides/storage.md
+++ b/Documentation/user-guides/storage.md
@@ -176,6 +176,7 @@ gp3-csi   true
 ```
 
 Next, update the `spec.paused` field to `true` (to prevent the operator from recreating the StatefulSet) and update the storage request in the `spec.storage` field of the custom resource. Assuming a Prometheus resource named `example` for which you want to increase the storage size to 10Gi:
+(You can get actual name by running `kubectl get prometheus`) 
 
 ```bash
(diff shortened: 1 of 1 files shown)


JoaoBraveCoding (commented):
I appreciate the effort but I'm not entirely sure this addition makes sense in this document. This walkthrough covers a topic which is mainly looked up by a more experienced user base.
Nevertheless, if you feel like this is not clear in our documentation we could maybe introduce something similar in the more beginner-friendly documents we have.

apgapg:
Thank you for replying.

`example` in docs doesnt seems to convey which resource or CRD we are referring too.
Thats why I added the command to let user know which CRD we are talking about

simonpasquier:
I'd recommend to rewrite the commands with environment variable substitution and fail if the variable isn't set. Something like:

```
export PROMETHEUS_NAME=example # Replace with the name of the Prometheus resource that needs to be patched
kubectl patch "prometheus/${PROMETHEUS_NAME:?}" --patch ...
````

apgapg:
@simonpasquier `kubectl patch "prometheus/${PROMETHEUS_NAME:?}" --patch ...` this syntax seems much better as whenever a new comer runs this command, this would auto fail, making him to find out the actual name.

PR #5541, opened 2023-05-03
Support VolumeMounts in ThanosRuler

### LINKED ISSUE #5538: Add volumeMounts to ThanosRulerSpec
**What is missing?**

Currently, we are able to add a volume to a ThanosRuler, via the `volumes` property, and which gets propagated to the generated StatefulSet. However, `ThanosRuler`, unlike with the spec for `Alertmanager`, `Prometheus` and `PrometheusAgent`, does not have a `volumeMounts` property, which means that it is not easily possible to mount the volume.

**Why do we need it?**

My current use case is to facilitate Stateless ruler operation to a secured Thanos Receive instance, which needs client certs and CAs mounting. 
However we should not just support this use case, but any other future use cases which could require a volume to be mounted within a Ruler instance (although it would be nice to have first-class support for this)

**Environment**

* Prometheus Operator version: `v0.58.0`

vCra:
## Description

Supports creating a VolumeMount in ThanosRuler objects
Closes #5538 

## Type of change

## Changelog entry

```release-note
- Add VolumeMounts to ThanosRuler CRD to allow Volumes to be mounted within Ruler
```

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -3122,6 +3122,21 @@ be appended to other volumes that are generated as a result of StorageSpec objec
 </tr>
 <tr>
 <td>
+<code>volumeMounts</code><br/>
+<em>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#volumemount-v1-core">
+[]Kubernetes core/v1.VolumeMount
+</a>
+</em>
+</td>
+<td>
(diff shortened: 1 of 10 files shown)


xiu (approved):
lgtm

JoaoBraveCoding (approved):
lgtm

ArthurSens (approved):
Thanks for the PR :)

I was wondering if we need to check for name colision with some pre-existing volumes, but I believe ThanosRules doesn't have any

PR #5566, opened 2023-05-10
operator: include newline in error messages

tskinn:
common Kubernetes log configurations will not recognize the last log message of pod if it does not include a newline

## Description

I recently wasted some time debugging prometheus-operator constantly crashing while running with `hostNetwork: true`. I wasn't seeing any meaningful logs and the pod was exiting very quickly. It turned out that the port that the operator was attempting to acquire was already in use. I then wondered why this wasn't in the logs as it should be and found that my log viewing tool doesn't report the last text that a pod produces if it doesn't end in a newline.

Also I noted that every other output to stderr appends newlines so it would be good to be consistent as well.

## Type of change

## Changelog entry

```release-note
- Top level exit errors end with newline ensuring the logs don't get lost
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -262,7 +262,7 @@ func Main() int {
 
 	po, err := prometheuscontroller.New(ctx, cfg, log.With(logger, "component", "prometheusoperator"), r)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


JoaoBraveCoding (approved):
Makes sense, lgtm

philipgough (approved):

PR #5567, opened 2023-05-10
feat: add initial events draft

### LINKED ISSUE #3611: Emit Kubernetes events when the operator recognizes invalid resources
**What is missing?**

The operator will reject resources that would trigger invalid configurations of Prometheus, Alertmanager and Thanos Ruler. While we already have [metrics](https://github.com/prometheus-operator/prometheus-operator/blob/2fda6f185b0f83b4aa09e47b6beffcdc2ca40739/pkg/operator/operator.go#L40-L45) in place, it would be useful to emit Kubernetes events.

**Why do we need it?**

It would be easier for the operator admins to troubleshoot rather than looking at the operator's logs.

Reported by @brancz on [Slack](https://kubernetes.slack.com/archives/C01B03QCSMN/p1603267460027600).

FUSAKLA:
resolves #3611 

Hi, so I started looking into the emitting of k8s events 

This is a non-working draft but opened few questions already:
 - :grey_question:  There are two tools libraries [`tools/record`](https://pkg.go.dev/k8s.io/client-go/tools/record) and [`tools/events`](https://pkg.go.dev/k8s.io/client-go/tools/events) and I'm not sure which should be used. It seems that the `events` is "newer" but it is nowhere documented. Unfortunately, there are no examples or code I'd find using the `tools/events`. Everyone seems to be using the `record` or the kubebuilder libs.
 - :grey_question: Do we want to somehow integrate reporting of metrics/logging/wrapping errors and emitting events? Otherwise, it feels a bit redundant (see the diff)
 - :grey_question: Where to emit the events and how many of them to emit
 
 The current code builds, runs, passes tests, but I do not see the events in k8s :/
 
 I'll keep debugging and make it work, but would be happy to start discussion about the details already.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,21 +174,21 @@ image: .hack-operator-image .hack-prometheus-config-reloader-image .hack-admissi
 # Create empty target file, for the sole purpose of recording when this target
 # was last executed via the last-modification timestamp on the file. See
 # https://www.gnu.org/software/make/manual/make.html#Empty-Targets
-	$(CONTAINER_CLI) build --build-arg ARCH=$(ARCH) --build-arg OS=$(GOOS) -t $(IMAGE_OPERATOR):$(TAG) .
+	$(CONTAINER_CLI) build --no-cache --build-arg ARCH=$(ARCH) --build-arg OS=$(GOOS) -t $(IMAGE_OPERATOR):$(TAG) .
 	touch $@
 
 .hack-prometheus-config-reloader-image: cmd/prometheus-config-reloader/Dockerfile prometheus-config-reloader
 # Create empty target file, for the sole purpose of recording when this target
 # was last executed via the last-modification timestamp on the file. See
 # https://www.gnu.org/software/make/manual/make.html#Empty-Targets
-	$(CONTAINER_CLI) build --build-arg ARCH=$(ARCH) --build-arg OS=$(GOOS) -t $(IMAGE_RELOADER):$(TAG) -f cmd/prometheus-config-reloader/Dockerfile .
+	$(CONTAINER_CLI) build --no-cache --build-arg ARCH=$(ARCH) --build-arg OS=$(GOOS) -t $(IMAGE_RELOADER):$(TAG) -f cmd/prometheus-config-reloader/Dockerfile .
 	touch $@
 
 .hack-admission-webhook-image: cmd/admission-webhook/Dockerfile admission-webhook
 # Create empty target file, for the sole purpose of recording when this target
 # was last executed via the last-modification timestamp on the file. See
 # https://www.gnu.org/software/make/manual/make.html#Empty-Targets
-	$(CONTAINER_CLI) build --build-arg ARCH=$(ARCH) --build-arg OS=$(GOOS) -t $(IMAGE_WEBHOOK):$(TAG) -f cmd/admission-webhook/Dockerfile .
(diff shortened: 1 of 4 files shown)


FUSAKLA:
Regarding choosing between `v1.event` used by the `tools/record` and  `events.v1` used by the `tools/events` I just remembered [great talk from Bryan Boreham about his fun with events](https://youtu.be/YI1ZuN-OHNw?t=369) :smile: 

The KEP is here

https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation#implementation-history

I'll give the `tools/record` library a shot if it works with it and if so, I'd stick to it even though it is the "older" but at least documented and used :grimacing:

JoaoBraveCoding:
(Disclaimer I'm not a maintainer)

> There are two tools libraries [tools/record](https://pkg.go.dev/k8s.io/client-go/tools/record) and [tools/events](https://pkg.go.dev/k8s.io/client-go/tools/events) and I'm not sure which should be used. It seems that the events is "newer" but it is nowhere documented. Unfortunately, there are no examples or code I'd find using the tools/events. Everyone seems to be using the record or the kubebuilder libs.

Since sample-controler still uses the package `record` I would go with that. However, looking at this part of the code [1] I would assume the new package behaves in a similar way... I also noticed this operator [2] uses the new package 

> Do we want to somehow integrate reporting of metrics/logging/wrapping errors and emitting events? Otherwise, it feels a bit redundant

My gut feeling tells me that you might be correct in emitting an event every time we encounter a problem in the reconciliation. However, to be safe I would first emit an event whenever we increase the reject in the metric `prometheus_operator_managed_resources`

> Where to emit the events and how many of them to emit

We should be able to see events by doing `kubectl get ev`. I would say that we want to see 1 event every time the operator reconciles a resource.


[1] https://github.com/kubernetes/sample-controller/blob/758d61ecc444e52d050b2857e3a89c0b1a144a3e/controller.go#L102-L106
[2] https://github.com/kubernetes-sigs/cluster-capacity

FUSAKLA:
Thanks for all the input, I'm a bit short of time lately but will try to get to it again, soon :crossed_fingers:

Makefile
```diff
@@ -174,21 +174,21 @@ image: .hack-operator-image .hack-prometheus-config-reloader-image .hack-admissi
 # Create empty target file, for the sole purpose of recording when this target
 # was last executed via the last-modification timestamp on the file. See
 # https://www.gnu.org/software/make/manual/make.html#Empty-Targets
-	$(CONTAINER_CLI) build --build-arg ARCH=$(ARCH) --build-arg OS=$(GOOS) -t $(IMAGE_OPERATOR):$(TAG) .
+	$(CONTAINER_CLI) build --no-cache --build-arg ARCH=$(ARCH) --build-arg OS=$(GOOS) -t $(IMAGE_OPERATOR):$(TAG) .
```
rexagod, line 177: Why the `--no-cache` flag? I'd prefer smaller compile times by caching layers whenever possible.

rexagod:
There's a tradeoff with both `record` and `events` packages, where even though the former is well documented (and older), the latter allows working with the newer `"k8s.io/api/events/v1".Event` types. IMHO I agree with Simon and Joao, and I'd go with the former, since that's been around for a while and has been tried and tested (68x more imports).

ArthurSens:
Superseeded by https://github.com/prometheus-operator/prometheus-operator/pull/6179

PR #5572, opened 2023-05-11
Proposal: Scrape Classes

EronWright:
## Description

Proposal to add scrape classes to facilitate the use of non-safe TLS elements in a scrape configuration, e.g. in the Istio use case.

A draft implementation is available: https://github.com/prometheus-operator/prometheus-operator/pull/5618. Links to significant code changes: [prom api](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-b4a3c944005e67c959c04552108ff3b629b572d7244d586b0558651bd96eacda), [scrape class](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-255c0aca616a1cdf4385c88d1d063a6c86360fedf99f7fc76bb967034e68897cR642-R652), [podmonitor api](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-ed302c709df2a336b5712f4f4442e4972be00571fd5293c61e0f65f752b9e10a), [operator](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-2297eaf6931eb847fb894df894f8f50ac9fb7cc823e440280d5908b5c9939104), [config generator](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-453ce9c721c51bf9591c2b212e9ce50157b9c40d1b211177bf502be268099656), and [tests](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-40b00673aa9acb2e0dd5431cb36eedc1476124b086a3e070f5d6182709f8ea35).

## Type of change

## Changelog entry

```release-note
- Proposal: scrape classes
```

### DIFF
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -43,10 +43,12 @@ to compose the scrape configurations in a Kubernetes way.
 ## How
 
 The proposed solution is to introduce a notion of a *scrape class*, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/).
-A scrape class defines TLS settings (and possibly other settings in future) to be applied to all scrape endpoints of that class.
+A scrape class defines TLS settings (and possibly other settings in future, e.g. non-safe authorization settings) to be applied to all scrape configs of that class.
 
-When defining a probe/podmonitor/servicemonitor, a user may optionally assign a class to each endpoint.
-A default class may be defined by the administrator.
+One scrape class may be designated as the default class, in which case that class is applied to any scrape config that doesn't specify a value for `scrapeClass`.
+
+When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClass` field.
+For monitors, each endpoint may be assigned a scrape class.
 
 Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
 
@@ -61,12 +63,11 @@ of the `Prometheus` spec. An alternative is outlined later, of factoring the cla
 apiVersion: monitoring.coreos.com/v1
 kind: Prometheus
 spec:
-  # optional: set a default scrape class
-  defaultScrapeClass: istio-mtls
-
   # define scrape classes for use by the monitors
+  # one class may be designated as the default class
   scrapeClasses:
     - name: istio-mtls
+      default: true
       tlsConfig:
         caFile: "/etc/istio-certs/root-cert.pem"
         certFile: "/etc/istio-certs/cert-chain.pem"
@@ -124,7 +125,26 @@ spec:
     scrapeClass: istio-mtls
 ```
 
-Deprecate the unsafe TLS settings in `ServiceMonitor`.
+Out-of-scope: deprecation of the unsafe TLS settings in `ServiceMonitor`.
+
+### ScrapeConfig
+
+Allow the user to select a scrape class for the generic scrape configuration.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: ScrapeConfig
+metadata:
+  name: scrape-config
+spec:
+  scrapeClass: istio-mtls
+  staticConfigs:
+    [...]
+  httpSDConfig:
+    [...]
+  fileSDConfig:
+    [...]
+```
 
 ## Test Plan
 
@@ -185,6 +205,7 @@ An open question is whether the resource would be cluster-scoped or namespace-sc
 
 Objections:
 1. Since the file paths are dependent on the volume mounts in the server, this approach may not achieve a meaningful decoupling.
+2. Extra complexity in defining a new CRD.
 
 ### Non-Safe Monitors
 
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -54,7 +54,7 @@ Class names are assumed to be installation-specific. In practice, some common cl
 
 ### Prometheus Resource
 
-It is proposed that the `Prometheus` resource contain a new section for defining scrape classes.
+It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a new section for defining scrape classes.
 
 The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
 of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource.
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -1,0 +1,189 @@
+
+## Scrape Classes
+
+* **Owners:**
+  * `eronwright`
+
+* **Related Tickets:**
+  * https://github.com/prometheus-operator/prometheus-operator/issues/4121
+  * https://github.com/prometheus-operator/prometheus-operator/issues/3922
+
+This proposal aims to introduce the concept of _scrape classes_, to enable users to leverage scrape configuration data 
+that is provided by the administrator. For example, to allow pod monitors and probes to safely use
+unsafe TLS configuration fields (e.g. `keyFile`). 
+
+## Why
+
+Today, there are some scenarios that are difficult to support in a safe way, notably the need to reference 
+a local TLS file in a scrape configuration. For example, to scrape a pod in an Istio mesh with strict mTLS would require that
+the scrape configuration use a TLS keyfile that is provided by the Prometheus server's Istio sidecar. 
+See [Istio documentation](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings) for more details.
+
+Another motivation is to improve feature parity amongst the monitor resources. The `PodMonitor` and `Probe` resources
+aren't at parity with `ServiceMonitor` because the latter allows for unsafe TLS settings.
+
+A minor motivation is to decouple the monitor spec from low-level infrastructure details like TLS certificates.
+
+### Pitfalls of the current solution
+
+The only known solution for use cases where you'd need to use unsafe TLS settings is to use `additionalScrapeConfig`. 
+The downside is obviously the loss of a great feature of the Prometheus Operator. The monitor resources make it possible 
+to compose the scrape configurations in a Kubernetes way.
+
+## Goals
+
+- Allow for the administrator to define a named, reusable scrape configuration snippet, including unsafe elements such as file references.
+- Allow for a user to select a configuration snippet by name in their probe/podmonitor/servicemonitor endpoint configuration.
+- Avoid giving the user the ability to use arbirary files within the Prometheus pod.
+
+### Audience
+
+- Users who serve Prometheus as a service and want to give their customers autonomy in defining monitors.
+- Users who want to scrape pods within an Istio mesh.
+
+## How
+
+The proposed solution is to introduce a notion of a  _scrape class_, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/). 
+A scrape class defines TLS settings (and possibly other settings in future) to be applied to all scrape endpoints of that class. 
+
+When defining a probe/podmonitor/servicemonitor, a user may optionally assign a class to each endpoint. 
+A default class may be defined by the administrator.
+
+Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
+
+### Prometheus Resource
+It is proposed that the `Prometheus` resource contain a new section for defining scrape classes.
+
+_The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
+of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource._
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+spec:
+  # optional: set a default scrape class
+  defaultScrapeClass: istio-mtls
+
+  # define scrape classes for use by the monitors
+  scrapeClasses:
+    - name: istio-mtls
+      tlsConfig:
+        caFile: "/etc/istio-certs/root-cert.pem"
+        certFile: "/etc/istio-certs/cert-chain.pem"
+        keyFile: "/etc/istio-certs/key.pem"
+        insecureSkipVerify: true
+
+  # mount the certs from the istio sidecar (shown here for illustration purposes)
+  volumeMounts:
+    - name: istio-certs
+      mountPath: "/etc/istio-certs/"
+```
+
+### PodMonitor Resource
+Allow the user to select a scrape class for each endpoint.
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+spec:
+  podMetricsEndpoints:
+  - port: http
+    path: /metrics
+    scrapeClass: istio-mtls
+```
+
+The suggested behavior is:
+1. the `tlsConfig` in the associated scrape class is automatically applied to the scrape configuration of the endpoint.
+2. the inline `tlsConfig` (if any) takes precedence over the `tlsConfig` in the scrape class.
+
+### Probe Resource
+Allow the user to select a scrape class for the probe.
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Probe
+spec:
+  scrapeClass: istio-mtls
+```
+
+### ServiceMonitor Resource
+Allow the user to select a scrape class for each endpoint.
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+spec:
+  endpoints:
+  - port: http
+    path: /metrics
+    scrapeClass: istio-mtls
+```
+
+Deprecate the unsafe TLS settings in `ServiceMonitor`.
+
+## Test Plan
+
+1. Regression test; ensure no change to generated configs unless a scrape class is applied.
+2. Optionality of the default scrape class; ensure that it is an optional configuration element.
+3. Acceptance testing in Istio environment; ensure that the solution is effective for a key use case.
+
+## Alternatives
+
+### Global scrape TLS configuration
+An alternative solution would be to apply a default TLS configuration to all monitors.
+
+For example, via a hypothetical field `spec.scrapeTlsConfig`:
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+spec:
+  scrapeTlsConfig:
+    caFile: "/etc/istio-certs/root-cert.pem"
+    certFile: "/etc/istio-certs/cert-chain.pem"
+    keyFile: "/etc/istio-certs/key.pem"
+    insecureSkipVerify: true
+```
+
+Objections:
+1. A singular default configuration may be too inflexible to effectively scrape a diverse set of pods. For example, in Istio, 
+some pods may be in STRICT mode.
+
+### Istio Permissive Mode
+An alternative for the Istio use case is to use PERMISSIVE mode (see [documentation](https://istio.io/latest/docs/concepts/security/#permissive-mode)),
+or to use `exclude` annotations on the pod such that the metrics endpoint bypasses mTLS.
+
+Objections:
+1. This solution is less secure at a transport level unless TLS is implemented at the application layer.
+2. Strict mTLS is the basis for [Istio authorization policy](https://istio.io/latest/docs/reference/config/security/authorization-policy/), e.g. to explicitly allow access from prom to the metrics endpoint.
+
+### ScrapeClass Resource
+A variant of the proposed solution is to introduce a new custom resource for defining scrape classes.
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: ScrapeClass
+metadata:
+  name: istio
+spec:
+  tlsConfig:
+    caFile: "/etc/istio-certs/root-cert.pem"
+    certFile: "/etc/istio-certs/cert-chain.pem"
+    keyFile: "/etc/istio-certs/key.pem"
+    insecureSkipVerify: true
+```
+
+An open question is whether the resource would be cluster-scoped or namespace-scoped.
+
+Objections:
+1. Since the file paths are dependent on the volume mounts in the server, this approach may not achieve a meaningful decoupling.
+
+### Non-Safe Monitors
+Another alternative would be to allow `PodMonitor` and `Probe` to use unsafe TLS settings. 
+
+Objections:
+1. See [explanation](https://github.com/prometheus-operator/prometheus-operator/issues/3922#issuecomment-802899950) for why 
+unsafe settings were disallowed in the first place.
+
+## Action Plan
+
+* [ ] Change the Operator API to define scrape classes and for endpoint to select a scrape class
+* [ ] Update the scrape configuration generator to use the scrape class
+* [ ] Update documentation
+* [ ] Resolve issues #4121, #3922
+
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -22,7 +22,7 @@ See [Istio documentation](https://istio.io/latest/docs/ops/integrations/promethe
 Another motivation is to improve feature parity amongst the monitor resources. The `PodMonitor` and `Probe` resources
 aren't at parity with `ServiceMonitor` because the latter allows for unsafe TLS settings.
 
-A minor motivation is to decouple the monitor spec from low-level infrastructure details like TLS certificates.
+Yet another motivation is to decouple the monitor spec from low-level infrastructure details like TLS certificates.
 
 ### Pitfalls of the current solution
 
@@ -79,6 +79,8 @@ spec:
       mountPath: "/etc/istio-certs/"
 ```
 
+Any object references in the scrape class definition are assumed to refer to objects in the namespace of the `Prometheus` object.
+
 ### PodMonitor Resource
 Allow the user to select a scrape class for each endpoint.
 ```yaml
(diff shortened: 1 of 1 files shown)


slashpai:
Wont the ScrapeConfig https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202212-scrape-config.md address this? We have initial implementation of this available from 0.65 version

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+
+Objections:
+1. Since the file paths are dependent on the volume mounts in the server, this approach may not achieve a meaningful decoupling.
+
+### Non-Safe Monitors
+Another alternative would be to allow `PodMonitor` and `Probe` to use unsafe TLS settings. 
+
+Objections:
+1. See [explanation](https://github.com/prometheus-operator/prometheus-operator/issues/3922#issuecomment-802899950) for why 
+unsafe settings were disallowed in the first place.
```
simonpasquier, line 181: As far as I understand, there isn't a real security concern for TLS parameters because the target never receives the TLS key in clear. But the issue exists for tokens, basic auth, ...

Anyway the current approach used for ServiceMonitor isn't something that we want to generalize because requiring that the owner of the ServiceMonitor knows in advance the location of the client TLS key & cert files in the Prometheus container creates too much coupling.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+spec:
+  scrapeTlsConfig:
+    caFile: "/etc/istio-certs/root-cert.pem"
+    certFile: "/etc/istio-certs/cert-chain.pem"
+    keyFile: "/etc/istio-certs/key.pem"
+    insecureSkipVerify: true
+```
+
+Objections:
+1. A singular default configuration may be too inflexible to effectively scrape a diverse set of pods. For example, in Istio,
```
simonpasquier, line 145: good call!

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+spec:
+  podMetricsEndpoints:
+  - port: http
+    path: /metrics
+    scrapeClass: istio-mtls
+```
+
+The suggested behavior is:
+1. the `tlsConfig` in the associated scrape class is automatically applied to the scrape configuration of the endpoint.
+2. the inline `tlsConfig` (if any) takes precedence over the `tlsConfig` in the scrape class.
```
simonpasquier, line 96: Is the resulting TLS config 1) a merge between  the scrape class and the pod monitor TLS configs or 2) only the TLS config from the pod monitor?

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+
+_The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
+of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource._
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+spec:
+  # optional: set a default scrape class
+  defaultScrapeClass: istio-mtls
```
simonpasquier, line 65: another option would be to move this to the scrapeClass directly (and ensure that only one item is the default).

```
  scrapeClasses:
    - name: istio-mtls
      default: true
      tlsConfig:
        caFile: "/etc/istio-certs/root-cert.pem"
        certFile: "/etc/istio-certs/cert-chain.pem"
        keyFile: "/etc/istio-certs/key.pem"
        insecureSkipVerify: true
    - name: external-ca
      tlsConfig:
        caFile: "/etc/..."
        insecureSkipVerify: false
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Audience
+
+- Users who serve Prometheus as a service and want to give their customers autonomy in defining monitors.
+- Users who want to scrape pods within an Istio mesh.
+
+## How
+
+The proposed solution is to introduce a notion of a  _scrape class_, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/). 
+A scrape class defines TLS settings (and possibly other settings in future) to be applied to all scrape endpoints of that class.
```
simonpasquier, line 47: > and possibly other settings in future

I'm trying to imagine the other settings that could be supported eventually. Proxy parameters might make sense but I'd be wary to include things like bearer tokens. It would be good to clarify what can/can't be considered in the future.

simonpasquier (commented):
Thanks for starting the document, I really like it and it matches what I had in mind! I'm not sold yet to the scrapeClass name but I don't have a better suggestion yet :)

What should be the behavior if a monitor specifies a "scrapeClass" name that doesn't exist? The operator should probably reject the monitor but happy to hear other options.

Documentation/proposals/202305-scrapeclasses.md
xiu: > Proxy parameters might make sense but I'd be wary to include things like bearer tokens.
For security reasons? Could `ScrapeClass` be made to live in the same namespace as `Prometheus` with RBAC left for the operators to configure?

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+The proposed solution is to introduce a notion of a  _scrape class_, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/). 
+A scrape class defines TLS settings (and possibly other settings in future) to be applied to all scrape endpoints of that class. 
+
+When defining a probe/podmonitor/servicemonitor, a user may optionally assign a class to each endpoint. 
+A default class may be defined by the administrator.
+
+Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
+
+### Prometheus Resource
+It is proposed that the `Prometheus` resource contain a new section for defining scrape classes.
```
xiu, line 55: Is there a reason for it not to be its own CRD? That way ScrapeClasses can be managed separately from Prometheus itself, in the case where we have more settings than `tlsConfig` inside.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+    caFile: "/etc/istio-certs/root-cert.pem"
+    certFile: "/etc/istio-certs/cert-chain.pem"
+    keyFile: "/etc/istio-certs/key.pem"
+    insecureSkipVerify: true
+```
+
+An open question is whether the resource would be cluster-scoped or namespace-scoped.
+
+Objections:
+1. Since the file paths are dependent on the volume mounts in the server, this approach may not achieve a meaningful decoupling.
```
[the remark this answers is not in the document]
xiu: should the `ScrapeClass` be limited to just the `tlsConfig` and its files though? I could imagine it holding the `ProberSpec` as well for instance, which could be more dynamic than the Prometheus deployment.

Documentation/proposals/202305-scrapeclasses.md
simonpasquier: For the same reason that we forbid direct file references to tokens in PodMonitors and Probes: if a ScrapeClass can hold a bearer token value, a "rogue" target could specify that ScrapeClass name in their monitor object and get access to the token value when being scraped by Prometheus.

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
simonpasquier: not sure I'm following.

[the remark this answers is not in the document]
xiu: Linking to my other comment for the complete context: https://github.com/prometheus-operator/prometheus-operator/pull/5572#discussion_r1192333598

Documentation/proposals/202305-scrapeclasses.md
EronWright: Can you think of an example where a merge would be meaningful?

I have the impression that `tlsConfig` in the monitor would provide application-specific TLS settings, whereas the scrape class would provide infra-level TLS (e.g. for the Istio case where TLS is transparent).

Documentation/proposals/202305-scrapeclasses.md
EronWright: A counter-argument is that the administrator is explicitly offering the bearer token.  Assumedly the token would be file-based and rotating.

An example that comes to mind is [projected service account tokens](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-bound-service-account-tokens), wherein a OIDC token representing the Prometheus KSA would be used for authentication.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+See [Istio documentation](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings) for more details.
+
+Another motivation is to improve feature parity amongst the monitor resources. The `PodMonitor` and `Probe` resources
+aren't at parity with `ServiceMonitor` because the latter allows for unsafe TLS settings.
+
+A minor motivation is to decouple the monitor spec from low-level infrastructure details like TLS certificates.
+
+### Pitfalls of the current solution
+
+The only known solution for use cases where you'd need to use unsafe TLS settings is to use `additionalScrapeConfig`.
```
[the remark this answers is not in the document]
EronWright: I revised the section to mention `ScrapeConfig`.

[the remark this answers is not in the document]
EronWright: Would you suggest that `ScrapeConfig` be mentioned in this document? Could you provide a snippet to be included?

Documentation/proposals/202305-scrapeclasses.md
EronWright: Are you advocating for the `ScrapeClass` resource as outlined in alternatives?

EronWright:
Thanks all for the feedback and for the quick turnaround! I addressed the feedback and look forward to offering a PR once we take a decision on open questions. Top questions seem to be whether to use inline vs `ScrapeClass` resource, and whether _scrape class_ is a reasonable name.

If a monitor makes reference to a non-existent class, I suppose the monitor could simply be ignored (and ideally have a status condition to that effect) akin to [how a pod behaves](https://kubernetes.io/docs/concepts/containers/runtime-class/#usage) w.r.t. runtime class. Since numerous prometheuses may exist (with different classes) it makes little sense to reject the monitor during admission control.

I added a new paragraph in the `Prometheus` section to resolve an ambiguity:
> Any object references in the scrape class definition are assumed to refer to objects in the namespace of the `Prometheus` object.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,189 @@
...
+
+Today, there are some scenarios that are difficult to support in a safe way, notably the need to reference 
+a local TLS file in a scrape configuration. For example, to scrape a pod in an Istio mesh with strict mTLS would require that
+the scrape configuration use a TLS keyfile that is provided by the Prometheus server's Istio sidecar. 
+See [Istio documentation](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings) for more details.
+
+Another motivation is to improve feature parity amongst the monitor resources. The `PodMonitor` and `Probe` resources
+aren't at parity with `ServiceMonitor` because the latter allows for unsafe TLS settings.
+
+A minor motivation is to decouple the monitor spec from low-level infrastructure details like TLS certificates.
```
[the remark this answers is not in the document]
EronWright: Given that probes have unique configuration elements, I suppose that the concept of a probe class may make sense. One could argue for introducing a `ProbeClass` and to use the term `MonitorClass` rather than `ScrapeClass`.

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
xiu: It could be generic, keeping the `tlsConfig` example in mind:
```yaml
kind: ScrapeConfig
metadata:
  name: scrape-config
spec:
  scrapeClass: istio-tls-etc
  staticConfigs:
    [...]
  httpSDConfig:
    [...]
  fileSDConfig:
    [...]
```

Taking my second comment into account, it could also define the different files from `fileSDConfig` as well as endpoints from `httpSDConfig`. At the moment though I'm only opening the discussion and want to see others' feedback as to extending the `ScrapeClass` to do more than just `tlsConfig` would make sense.

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
JoaoBraveCoding: Just giving my .02 cents I agree with your use case on https://github.com/prometheus-operator/prometheus-operator/pull/5572#discussion_r1192333598 but I'm still not 100% convinced.

As of now the main advantages to SrapeClass being their own CRD that I'm able to see are:
1. Users being able to do `kc get ScrapeClass`  to see the available classes and pick the more desirable one
2. If there are many ScrapeClasses then maintaining a big list in the `Prometheus`  resource might become cumbersome

JoaoBraveCoding (commented):
Really cool proposal! I believe this work might have the potential to improve a lot the user experience depending on how we agree on what could become a field in the ScrapeClass. (not sure if this shouldn't be decided in this proposal or addressed later on)

Personally, I'm just still not sold on the name... I thought of `ConfigClass`. But I would say it depends a lot if the resource will be in line or not. If the resource ends up being a CRD I think I would prefer `ScrapeClass`. Also thought of `ScrapeConfig`, also in a similar vein to Alertmanager what about `PrometheusConfig`? Just bouncing ideas

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
xiu: The only other point I can see is to decouple the Prometheus CRD from the ScrapeClasses in the event of a Probe-like target managed by a different team (where ScrapeClasses could be managed by that same team). Now this might be a use case that's too little to consider.

After reading @EronWright's message on #prometheus-operator-dev (https://kubernetes.slack.com/archives/C01B03QCSMN/p1684516562355139?thread_ts=1684439235.731889&cid=C01B03QCSMN), I think we should be fine starting with inline, and can, if needed, reconsider a CRD at a later stage.
> 1. the basic idea is to separate the administrative inputs from the user inputs, and it seems likely to go beyond TLS, e.g. file-based bearer tokens.
> 2. Inline vs CRD aren’t mutually exclusive, and given the dependency on file mounts, I have a mild preference for inline class definition as a starting point.  If we go for a CRD, then please consider whether it would be namespaced and which namespace (prom or monitor?).
> 3. Is “scrape class” a good name, or would “monitor class” be better?  Are probes a type of monitor?

Documentation/proposals/202305-scrapeclasses.md
xiu: https://github.com/prometheus-operator/prometheus-operator/pull/5572/files#r1199707802

xiu:
We had a chat in the [contributor office hours](https://docs.google.com/document/d/1-fjJmzrwRpKmSPHtXN5u6VZnn39M28KqyQGBEJsqUOk/edit#) today, here are the notes:

> Notes from @EronWright: https://kubernetes.slack.com/archives/C01B03QCSMN/p1684516562355139?thread_ts=1684439235.731889&cid=C01B03QCSMN 
> 1. the basic idea is to separate the administrative inputs from the user inputs, and it seems likely to go beyond TLS, e.g. file-based bearer tokens.
> > * The usage of Probe configuration as part of ScrapeClasses might make sense but will need to be evaluated.
Proxy Settings might make sense as part of this Scrape Class.
> > * [simonpasquier] Anything that is related to configuring the scrape client on the Prometheus side can be bundled in the ScrapeClasses mechanisms. Not sure about default relabeling configs.
> 2. Inline vs CRD aren’t mutually exclusive, and given the dependency on file mounts, I have a mild preference for inline class definition as a starting point.  If we go for a CRD, then please consider whether it would be namespaced and which namespace (prom or monitor?).
> > * We can set a field as "Experimental" on a v1 CRD to not break stability promises
> > * Adding a new CRD would add more complexity, three places to look into to merge the configuration (Prometheus CRD + ScrapeClass CRD + *Monitor/Probe)
> > * Another benefit of a CRD is that we can reuse the same resource on multiple clusters/deployments
> 3. Is “scrape class” a good name, or would “monitor class” be better?  Are probes a type of monitor?
> > Let's polish the proposal first and agree on what we want to do, we'll change the name later.

One question now is: do we want the ScrapeClass to cover more than the scrape client configuration? (e.g. Probe default URLs etc).

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
EronWright: Given the added complexity of introducing a `ScrapeClass` CR, I propose that we start with inline classes (marked as experimental) and revisit later.

EronWright:
Thanks again, I made the following changes:
1. change to how to designate a scrape class as the default via `default: true`.
2. add a section for `ScrapeConfig`.
3. make deprecation of unsafe settings of `ServiceMonitor` to be out of scope.

I opted to leave two important things as-is:
1. kept the "inline" proposal as opposed to introducing a CRD.
2. didn't spell out which settings would be in `ScrapeClass` aside from the TLS settings.

I must admit that I'm tempted to explore adding the `authorization` field to `ScrapeClass`, for the scenario of using a bearer token file.  I use the [Service Account Token Projection](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-bound-service-account-tokens) feature in other projects and could imagine using it here.

xiu:
> @xiu in answer to your question about probe default URLs, I'm no expert but I'd say no, ScrapeClass should be focused on the true scrape settings.

re-reading the proposal today and considering our past discussions, I would agree. 

> I must admit that I'm tempted to explore adding the authorization field to ScrapeClass, for the scenario of using a bearer token file. I use the [Service Account Token Projection](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-bound-service-account-tokens) feature in other projects and could imagine using it here. 
> *Update: the https://github.com/prometheus-operator/prometheus-operator/pull/5618 does cover authorization using bearer token files.*

I think it'd make sense. Thoughts @simonpasquier / @JoaoBraveCoding?

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Audience
+
+- Users who serve Prometheus as a service and want to give their customers autonomy in defining monitors.
+- Users who want to scrape pods within an Istio mesh.
+
+## How
+
+The proposed solution is to introduce a notion of a *scrape class*, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/).
+A scrape class defines TLS settings (and possibly other settings in future, e.g. non-safe authorization settings) to be applied to all scrape configs of that class.
```
simonpasquier, line 46: (nit)
```suggestion
A scrape class defines TLS settings (and possibly other settings in future, e.g. sensitive authorization settings) to be applied to all scrape configs of that class.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+One scrape class may be designated as the default class, in which case that class is applied to any scrape config that doesn't specify a value for `scrapeClass`.
+
+When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClass` field.
+For monitors, each endpoint may be assigned a scrape class.
+
+Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
+
+### Prometheus Resource
+
+It is proposed that the `Prometheus` resource contain a new section for defining scrape classes.
```
simonpasquier, line 57: (nit) I assume that it applies to Prometheus and PrometheusAgent :)

```suggestion
It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a new section for defining scrape classes.
```

Documentation/proposals/202305-scrapeclasses.md
simonpasquier: One example I can think of: Prometheus defines the "default" client TLS certificate and key but the monitor defines the serverName value.

simonpasquier (commented):
Thanks! My last real question is whether we should merge the scrapeclass parameters with the individual monitor parameters.

To provide a more concrete example, let's consider how OpenShift components configure themselves for monitoring (details omitted for clarity).

The Kube API server's [service monitor ](https://github.com/openshift/cluster-kube-apiserver-operator/blob/a1de1b1cc85ce2d41e918d14534e5187b981819b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml#L38)looks like:

````
kind: ServiceMonitor
metadata:
  name: kube-apiserver
  namespace: openshift-kube-apiserver
spec:
...
    port: https
    scheme: https
    tlsConfig:
      caFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
      serverName: kubernetes
      certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
      keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
````

And the [Prometheus](https://github.com/openshift/cluster-monitoring-operator/blob/251585706ae84c79b88f9c14f99746f02a908e87/assets/prometheus-k8s/prometheus.yaml#L1) definition is:

````
kind: Prometheus
metadata:
  name: k8s
  namespace: openshift-monitoring
spec:
...
  secrets:
  - metrics-client-certs
````

While this is working as expected, it creates unnecessary coupling between the Prometheus and ServiceMonitor resources (the latter needs to know the location of the client certificate & key in the Prometheus container while it only wants to tell "do scraping with client TLS certificates").

With the proposal, this could be rewritten this way if we decide to support merging:

```
kind: ServiceMonitor
metadata:
  name: kube-apiserver
  namespace: openshift-kube-apiserver
spec:
...
    port: https
    scheme: https
    tlsConfig:
      serverName: kubernetes # this parameter (at least) is specific to each endpoint.
---
kind: Prometheus
metadata:
  name: k8s
  namespace: openshift-monitoring
spec:
...
  secrets:
  - metrics-client-certs
  scrapeClasses:
  - name: openshift
    default: true
    tlsConfig:
      caFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
      certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
      keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
````

**Review by JoaoBraveCoding — APPROVED** [review]
LGTM! No objections from my side

> last real question is whether we should merge the scrapeclass parameters with the individual monitor parameters. 

I would say yes to this 👍

**Comment by EronWright** [comment]
Thanks @JoaoBraveCoding for the useful feedback and suggestions, all of which I merged into the proposal. I was unaware of the `PrometheusAgent` CRD, and will be sure to cover it in the [implementation PR](https://github.com/prometheus-operator/prometheus-operator/pull/5618).

Seems there's a strong reason to merge the TLS configuration blocks, and will do so in the implementation. The semantics may vary on a case-by-case basis. This strikes me as similar to the semantics of server-side apply, e.g. the `+structType:atomic` and `+listType:atomic` markers. That is, a given block may be granular or be atomic.

@xiu please take a last look, and merge if you're ready, thanks!

**Review by xiu — APPROVED** [review]
LGTM, thanks a lot for the proposal!

One non-blocking thing: do we want to add a note on the fact that we seek merging `tlsConfig` between resources?

**Comment by xiu** [comment]
Adding @simonpasquier 's comment on #prometheus-operator-dev (https://kubernetes.slack.com/archives/C01B03QCSMN/p1687518206070409?thread_ts=1684439235.731889&cid=C01B03QCSMN):
> [@Eron Wright](https://kubernetes.slack.com/team/UTND57ZQF)
> first of all, many thanks for the effort, it's very nice to see the work moving forward :slightly_smiling_face:
> can you update the proposal PR to include the comments on the config merge?

Once done, we should be able to merge 👍

**Comment by nicolastakashi** [comment]
Amazing job here @EronWright 👏🏽 
If I would choose one option, I prefer to have the CR instead of the fields described on the Prometheus CR Spec.
This will give us more flexibility to evolve the `ScrapeClass` without overloading the Prometheus CR which is currently big enough 😄

**Comment by xiu** [comment]
Hi @EronWright, is this still something you want to pursue?

File: Documentation/proposals/202305-scrapeclasses.md [hunk 12]
```diff
@@ -0,0 +1,223 @@
...
+Allow the user to select a scrape class for each endpoint.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+spec:
+  podMetricsEndpoints:
+  - port: http
+    path: /metrics
+    scrapeClass: istio-mtls
```
paulfantom, line 96: Why per endpoint and not per object? I am curious since many `*Class` objects (`runtimeClass`, `storageClass`, `ingressClass`) in kubernetes are applied on the whole child object and not just part of the configuration. Since there is no restriction on how many PodMonitors there can be, shouldn't prometheus-operator follow the same principle as kubernetes?

nicolastakashi: I agree with you @paulfantom

simonpasquier: I missed this comment but I agree that it makes more sense to operate at the spec level than at the endpoint level.

sathieu:
@EronWright What is the status of this PR. Are you still working on it?

xiu:
@sathieu Hi! This issue was discussed today in the contributor office hours and @nicolastakashi will take a stab at finishing it 👍

EronWright:
Thanks @nicolastakashi for getting this over the finish line!

PR #5585, opened 2023-05-16
tests: refactors ScrapeConfig tests in promcfg_test

JoaoBraveCoding:
## Description
This PR refactors new ScrapeConfig tests so that they are split into test cases instead of each having its individual test. This helps maintainability but it also help addition of new tests and debugging possible issues

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -8632,52 +8632,22 @@ scrape_configs:
 	require.Equal(t, expected, string(cfg))
 }
 
-func TestScrapeConfigStaticConfig(t *testing.T) {
-	p := defaultPrometheus()
-
-	cg := mustNewConfigGenerator(t, p)
-	cfg, err := cg.GenerateServerConfiguration(
-		p.Spec.EvaluationInterval,
-		p.Spec.QueryLogFile,
-		p.Spec.RuleSelector,
-		p.Spec.Exemplars,
-		p.Spec.TSDB,
-		p.Spec.Alerting,
-		p.Spec.RemoteRead,
-		nil,
-		nil,
-		nil,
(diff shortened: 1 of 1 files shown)


ArthurSens (approved):
LGTM!

This will be helpful as we extend ScrapeConfig with the remaining fields :)

pkg/prometheus/promcfg_test.go
```diff
@@ -8632,231 +8632,107 @@ scrape_configs:
...
-						{
-							Targets: []monitoringv1alpha1.Target{"http://localhost:9100"},
-							Labels: map[monitoringv1.LabelName]string{
-								"label1": "value1",
-							},
+func TestScrapeConfigSpecConfig(t *testing.T) {
+	refreshInterval := monitoringv1.Duration("5m")
+	for _, tc := range []struct {
+		name        string
+		patchProm   func(*monitoringv1.Prometheus) *monitoringv1.Prometheus
```
JoaoBraveCoding, line 8639: I've added this field to the test struct in case some test cases need to patch the default Prometheus, for instance, `EnforcedNamespaceLabel`. I had this idea when refactoring the Probes tests

xiu (approved):
LGTM, thanks for the refactoring!

pkg/prometheus/promcfg_test.go
JoaoBraveCoding: I've modified it once more to be, no need for an assignment since we pass Prometheus as a pointer
```go 
patchProm   func(*monitoringv1.Prometheus)
```

ArthurSens:
@JoaoBraveCoding should I merge or do you still want to make a few changes?

JoaoBraveCoding:
@ArthurSens from my side you can merge it! I'm happy with it, thanks for checking 🙏

PR #5586, opened 2023-05-17
tests: refactors probe tests in promcfg_test

JoaoBraveCoding:
## Description

In this PR I moved all the tests to Probes to a single function with test cases similar to what was done in https://github.com/prometheus-operator/prometheus-operator/pull/5585. Unfortunately, this change looks rather complex on the diff due to a lot of functions shifting upwards. I would recommend to reviewers to just open the file directly and review the code that way. The only test functions that were changed were the ones related to Probe.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -846,6 +846,9 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	return cfg
 }
 
+// generateProbeConfig builds the prometheus configuration for a probe. This function
+// assumes that it will never receive a probe with empty targets, since the
+// operator filters those in the validation step in SelectProbes().
 func (cg *ConfigGenerator) generateProbeConfig(
 	m *monitoringv1.Probe,
 	apiserverConfig *monitoringv1.APIServerConfig,
@@ -950,7 +953,7 @@ func (cg *ConfigGenerator) generateProbeConfig(
 		xc := labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, m.Spec.Targets.StaticConfig.RelabelConfigs)
 		relabelings = append(relabelings, generateRelabelConfig(xc)...)
 		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
-	} else {
+	} else if m.Spec.Targets.Ingress != nil {
 		// Generate kubernetes_sd_config section for the ingress resources.
 
 		// Filter targets by ingresses selected by the monitor.
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -48,11 +48,6 @@ func defaultPrometheus() *monitoringv1.Prometheus {
 		},
 		Spec: monitoringv1.PrometheusSpec{
 			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
-				ProbeSelector: &metav1.LabelSelector{
-					MatchLabels: map[string]string{
-						"group": "group1",
-					},
-				},
 				Version:        operator.DefaultPrometheusVersion,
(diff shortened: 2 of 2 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -938,7 +938,9 @@ func (cg *ConfigGenerator) generateProbeConfig(
 		xc := labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, m.Spec.Targets.StaticConfig.RelabelConfigs)
 		relabelings = append(relabelings, generateRelabelConfig(xc)...)
 		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
-	} else {
+	}
+
+	if m.Spec.Targets.Ingress != nil {
```
JoaoBraveCoding, line 943: This change was done because of this test case, this test case didn't exist before I added it, I'm happy to remove it and revert this change however, if a user creates a Probe without `Spec.Targets` this will lead to the operator crashing with a Seg Fault:

https://github.com/JoaoBraveCoding/prometheus-operator/blob/320ffa339d41c76c33c4416a939ba58c8238437e/pkg/prometheus/promcfg_test.go#L7853-L7868

ArthurSens: Nice catch! 

For a future PR, we could probably add `// +kubebuilder:validation:Required` markers. I'm not sure if blackbox-exporter works without targets, but, if it does, sounds like a mistake from the user

ArthurSens (commented):
I was just passing by before joining a meeting, I'll take a deeper look later today 😛

ArthurSens (commented):
Ok that was a tough PR to read 😅 

The only thing that got my attention is that we're testing empty probes with empty targets or even a completely empty probeSpec.

Do you believe that's necessary? Are empty Probes even accepted by the CRD validation?

pkg/prometheus/promcfg.go
simonpasquier: IIUC validation markers can't check that at least one of `.spec.targets.static` or `.spec.targets.ingress`  is defined. It should be possible with [CEL](https://kubernetes.io/docs/reference/using-api/cel/) though but it's still [beta](https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/). Anyway, supporting/using CEL is probably a discussion to have in a separate issue.

simonpasquier: In practice, the operator will not select a probe that have nil static and ingress fields.

https://github.com/prometheus-operator/prometheus-operator/blob/a487750b374b7c69fef1c8fee88af74c9cac8eb3/pkg/prometheus/resource_selector.go#L497C31-L500

https://github.com/prometheus-operator/prometheus-operator/blob/a487750b374b7c69fef1c8fee88af74c9cac8eb3/pkg/apis/monitoring/v1/probe_types.go#L113-L119

simonpasquier, line 943: The doc says that static takes precedence over ingress if both are defined.

```suggestion
	} else if m.Spec.Targets.Ingress != nil {
```

JoaoBraveCoding: I've added a comment to the beginning of the function

ArthurSens:
Hey @JoaoBraveCoding, what do you think about breaking this PR down into multiple smaller ones just to make it easier to review? 😅 

Maybe we could have 2 test functions for a while during the migration to the new format?

JoaoBraveCoding:
> Hey @JoaoBraveCoding, what do you think about breaking this PR down into multiple smaller ones just to make it easier to review? 😅
> 
> Maybe we could have 2 test functions for a while during the migration to the new format?

Yes, it might be easier/safer 1st one is already up https://github.com/prometheus-operator/prometheus-operator/pull/5677 once it's merged I'll move on

PR #5591, opened 2023-05-18
[FEATURE] Adding Prometheus Tracing Support

nicolastakashi:
## Description

Adding tracing configuration section to `Prometheus` as requested in the issue #4722

## Type of change

## Changelog entry

```release-note
[FEATURE] Adding Prometheus Tracing Support
```


nicolastakashi:
Testing evidence
image

pkg/prometheus/promcfg.go
```diff
@@ -2223,3 +2235,92 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 	}
 	return cfg
 }
+
+func (cg *ConfigGenerator) generateTracingConfig(tracingConfig *v1.PrometheusTracingConfig) (yaml.MapItem, error) {
+	cfg := yaml.MapSlice{}
+	objMeta := cg.prom.GetObjectMeta()
+
+	if tracingConfig.ClientType != "" {
```
ArthurSens, line 2243: Since you've added a validation with enums in Kubebuilder, there is no need to check clientType here. Kubernetes API will refuse Prometheus resources if clientType is not set to grpc or http

pkg/prometheus/promcfg.go
```diff
@@ -2223,3 +2235,92 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+		}
+
+		cfg = append(cfg, yaml.MapItem{
+			Key:   "headers",
+			Value: headers,
+		})
+	}
+
+	if tracingConfig.Compression != "" {
+		if tracingConfig.Compression != "gzip" {
```
ArthurSens, line 2301: same thing here

pkg/prometheus/promcfg.go
```diff
@@ -2223,3 +2235,92 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+			return yaml.MapItem{}, errors.New("Invalid Compression")
+		}
+
+		cfg = append(cfg, yaml.MapItem{
+			Key:   "compression",
+			Value: tracingConfig.Compression,
+		})
+	}
+
+	if tracingConfig.Timeout != "" {
```
ArthurSens, line 2311: I wonder if we can't set default values with kubebuilder annotation as well, then we would need to make all those nil checks 🤔

ArthurSens (commented):
Thanks for taking this one up!

Overall it looks good already! We just need to take a look at the failing e2e test and a few nits

pkg/prometheus/promcfg.go
nicolastakashi: Changed 😄

pkg/prometheus/promcfg.go
nicolastakashi: Changed 😄

pkg/prometheus/promcfg.go
nicolastakashi: I did it.
Just ensure that we're setting default values same as Prometheus and it helps clean the code a bit.

pkg/prometheus/promcfg_test.go
```diff
@@ -8830,10 +8882,144 @@ scrape_configs:
...
+						MatchLabels: map[string]string{
+							"group": "group1",
+						},
+					},
+					Version:        operator.DefaultPrometheusVersion,
+					ScrapeInterval: "30s",
+				},
+				EvaluationInterval: "30s",
+			},
+		}
```
JoaoBraveCoding, line 8989:
```suggestion
		p := defaultPrometheus()
		p.Spec.TracingConfig = tc.tracingConfig
```

pkg/prometheus/promcfg_test.go
```diff
@@ -8830,10 +8882,144 @@ scrape_configs:
...
+			nil,
+			nil,
+			nil,
+			nil,
+			nil,
+			nil,
+			nil,
+			nil,
+			nil,
+			tc.tracingConfig,
```
JoaoBraveCoding, line 9009:
```suggestion
			p.Spec.TracingConfig,
```

pkg/prometheus/promcfg_test.go
```diff
@@ -8830,10 +8882,144 @@ scrape_configs:
...
+						},
+					},
+					Version:        operator.DefaultPrometheusVersion,
+					ScrapeInterval: "30s",
+				},
+				EvaluationInterval: "30s",
+			},
+		}
+		cg := mustNewConfigGenerator(t, p)
+
```
JoaoBraveCoding, line 8991:
```suggestion
```

pkg/prometheus/promcfg_test.go
```diff
@@ -8830,10 +8882,144 @@ scrape_configs:
...
 	}
 }
+
+func TestTracingConfig(t *testing.T) {
+	boolFalse := false
+	testCases := []struct {
+		tracingConfig  *monitoringv1.PrometheusTracingConfig
+		name           string
+		expectedConfig string
+		shouldSucceed  bool
```
JoaoBraveCoding, line 8900: nit: just to be consistent with other similar code in this file
```suggestion
		expectedErr  bool
```

pkg/prometheus/promcfg_test.go
```diff
@@ -8830,10 +8882,144 @@ scrape_configs:
...
+
+		if tc.shouldSucceed && err != nil {
+			t.Fatalf("expected config generation to not fail during test case '%s'.", tc.name)
+		}
+
+		result := string(cfg)
+		if result != tc.expectedConfig {
+			t.Logf("\n%s", pretty.Compare(tc.expectedConfig, result))
+			t.Fatalf("expected Prometheus configuration and actual configuration do not match on test case '%s'", tc.name)
+		}
```
JoaoBraveCoding, line 9023:
```suggestion
		if tc.shouldSucceed {
			require.NoError(t, err)
		} else {
			require.Error(t, err)
		}
		require.Equals(t, tc.expectedConfig, string(cfg))
```

JoaoBraveCoding (commented):
some comments and nit's

pkg/prometheus/promcfg.go
```diff
@@ -2223,3 +2235,76 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
 	return cfg
 }
+
+func (cg *ConfigGenerator) generateTracingConfig(tracingConfig *v1.PrometheusTracingConfig) (yaml.MapItem, error) {
+	cfg := yaml.MapSlice{}
+	objMeta := cg.prom.GetObjectMeta()
+
+	if tracingConfig.Endpoint == "" {
+		return yaml.MapItem{}, errors.New("Endpoint is required and can't be empty")
+	}
```
ArthurSens, line 2245: Maybe we could add a validation marker on the API making sure this field is required(`+kubebuilder:validation:Required`), that way we could get rid of this if

Ref: https://book.kubebuilder.io/reference/markers/crd-validation.html

ArthurSens: With all the markers, we could probably drop a few unit tests, or replace them with a e2e test maybe 🤔

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -526,6 +526,48 @@ type PrometheusSpec struct {
 	// Defines the runtime reloadable configuration of the timeseries database
 	// (TSDB).
 	TSDB TSDBSpec `json:"tsdb,omitempty"`
+	// TracingConfig configures tracing in Prometheus. This is an experimental feature, it may change in any upcoming release in a breaking way.
+	// +optional
+	TracingConfig *PrometheusTracingConfig `json:"tracingConfig,omitempty"`
+}
+
+type PrometheusTracingConfig struct {
```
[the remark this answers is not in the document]
nicolastakashi: Done mate!
If I'm not wrong only the bool field is required change to be a pointer.

pkg/prometheus/promcfg_test.go
nicolastakashi: Is this still valid?

pkg/prometheus/promcfg.go
nicolastakashi: Actually I just noticed we don't even need this annotation, since it's not tagged with +optional, it is required by default.

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
ArthurSens: I'm still getting used to the concepts, but I believe that `string` does not have built-in nil, and therefore neither has `Duration`. e.g. you can't compare `somestring == nil`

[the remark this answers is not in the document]
nicolastakashi: Duration on that context is also a string, the operator is just creating a type named Duration to ensure expressiveness.
I guess built in nil is like if non values provided it will be a default type value like on maps and slices are empty.

But I might be wrong hehe I'm not 100% sure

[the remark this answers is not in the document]
JoaoBraveCoding: Arthur is correct, since strings don't have a default `nil` value we should declare them as `*string`

pkg/prometheus/promcfg_test.go
```diff
@@ -8837,3 +8889,87 @@ scrape_configs:
...
+  insecure: false
+  headers:
+    custom: header
+  compression: gzip
+  timeout: 10s
+`,
+		},
+	}
+	for _, tc := range testCases {
+		p := defaultPrometheus()
```
JoaoBraveCoding, line 8945: Place test cases inside of `t.Run` which will make it easier to identify which test fails
```suggestion
	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			p := defaultPrometheus()
			.....
		}
	}
```

JoaoBraveCoding (commented):
Some extra comments! Thanks for addressing the previous ones <3

pkg/prometheus/promcfg_test.go
```diff
@@ -2553,10 +2569,6 @@ func TestEnforcedNamespaceLabelOnExcludedPodMonitor(t *testing.T) {
 						"group": "group1",
 					},
 				},
-				TypeMeta: metav1.TypeMeta{
-					APIVersion: monitoring.GroupName + "/" + monitoringv1.Version,
-					Kind:       monitoringv1.PodMonitorsKind,
-				},
```
ArthurSens, line 2559: Is that intentional? 🤔

pkg/prometheus/promcfg_test.go
```diff
@@ -8837,3 +8873,95 @@ scrape_configs:
 
 	}
 }
+
+func TestTracingConfig(t *testing.T) {
+	boolFalse := false
+	samplingFraction := "2"
+	clientType := "grpc"
+	compression := "gzip"
```
ArthurSens, line 8881: Instead of declaring all those variables here, you can use `pointer.Bool(false)`, `pointer.Int(2)`, `pointer.String("grpc")` where needed.

The failing test is probably related to a copy-paste mistake when declaring the variables above 😅

ArthurSens (commented):
Looks like unrelated code was deleted by accident?

pkg/prometheus/promcfg_test.go
ArthurSens: There are other occurrences in the same file

pkg/prometheus/promcfg_test.go
nicolastakashi: Oh!
That's cool.
TIL

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
nicolastakashi: Amazing @JoaoBraveCoding!
Done 😄

pkg/prometheus/promcfg_test.go
nicolastakashi: Ups!
It was by mistake, I fixed, can you check pls?

xiu:
@nicolastakashi your rebase seems to have gone wrong 😅 It squashed into the last commit from main.

nicolastakashi:
@xiu that's true 😅

I just fixed it, tks

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -526,6 +526,48 @@ type PrometheusSpec struct {
...
+	// Client used to export the traces. Options are "http" or "grpc".
+	//+kubebuilder:validation:Enum=http;grpc
+	//+kubebuilder:default:="grpc"
```
simonpasquier, line 537: (nit) I wouldn't use a default as I prefer to delegate the default  value to Prometheus.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -526,6 +526,48 @@ type PrometheusSpec struct {
...
+	// Endpoint to send the traces to. Should be provided in format <host>:<port>.
+	Endpoint string `json:"endpoint"`
+
+	// Sets the probability a given trace will be sampled. Must be a float from 0 through 1.
+	//+kubebuilder:default:="0"
+	// +optional
+	SamplingFraction *string `json:"samplingFraction"`
+
+	// If disabled, the client will use a secure connection.
+	//+kubebuilder:default:=false
```
simonpasquier, line 550: same here

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -526,6 +526,48 @@ type PrometheusSpec struct {
...
+	// +optional
+	Headers map[string]string `json:"headers"`
+
+	// Compression key for supported compression types. Supported compression: gzip
+	//+kubebuilder:validation:Enum=gzip
+	// +optional
+	Compression *string `json:"compression"`
+
+	// Maximum time the exporter will wait for each batch export. Default '10s'
+	//+kubebuilder:default:="10s"
```
simonpasquier, line 564: same thing about default

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -526,6 +526,48 @@ type PrometheusSpec struct {
...
+
+type PrometheusTracingConfig struct {
+	// Client used to export the traces. Options are "http" or "grpc".
+	//+kubebuilder:validation:Enum=http;grpc
+	//+kubebuilder:default:="grpc"
+	// +optional
+	ClientType *string `json:"clientType"`
+
+	// Endpoint to send the traces to. Should be provided in format <host>:<port>.
+	Endpoint string `json:"endpoint"`
```
simonpasquier, line 542: (nit) I see that it was already discussed but in order to be explicit, we could add the `//+required` annotation

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -526,6 +526,48 @@ type PrometheusSpec struct {
...
+
+	// Sets the probability a given trace will be sampled. Must be a float from 0 through 1.
+	//+kubebuilder:default:="0"
+	// +optional
+	SamplingFraction *string `json:"samplingFraction"`
```
simonpasquier, line 547: Maybe it was discussed before but representing float64 in CRDs is "challenging" (https://github.com/kubernetes-sigs/controller-tools/issues/245). Apparently one option could be to use `resource.Quantity` which should bring API validation but I haven't looked deeper into it.

pkg/prometheus/promcfg.go
```diff
@@ -2223,3 +2234,68 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+	cfg = append(cfg, yaml.MapItem{
+		Key:   "endpoint",
+		Value: tracingConfig.Endpoint,
+	})
+
+	parsedFraction, err := strconv.ParseFloat(*tracingConfig.SamplingFraction, 64)
+	if err != nil {
+		return yaml.MapItem{}, err
+	}
+	if parsedFraction > 1 || parsedFraction < 0 {
```
simonpasquier, line 2256: Prometheus is tolerant to out-of-bounds value (e.g. it will assume 0 for values < 0 and 1 for values > 1). We could the same here as a precaution.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -526,6 +526,48 @@ type PrometheusSpec struct {
 	// Defines the runtime reloadable configuration of the timeseries database
 	// (TSDB).
 	TSDB TSDBSpec `json:"tsdb,omitempty"`
+	// TracingConfig configures tracing in Prometheus. This is an experimental feature, it may change in any upcoming release in a breaking way.
+	// +optional
+	TracingConfig *PrometheusTracingConfig `json:"tracingConfig,omitempty"`
```
simonpasquier, line 531: doesn't it also apply for PrometheusAgent?

simonpasquier (commented):
Thanks for the PR!

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: The trace in prometheus is only instrumented for queries. Since it's not possible to query the agent, I don't think it's possible to configure tracing

pkg/prometheus/promcfg.go
nicolastakashi: Should we just remove this validation and let prometheus take care of it @simonpasquier ?

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: @simonpasquier I tried to use Quantity but for some reason (maybe I'm not getting right the API), if I set 0.5 when get the value it returns 0 🥲

pkg/prometheus/promcfg_test.go
```diff
@@ -8837,3 +8886,102 @@ scrape_configs:
...
+  external_labels:
+    prometheus: default/test
+    prometheus_replica: $(POD_NAME)
+scrape_configs: []
+tracing:
+  endpoint: https://otel-collector.default.svc.local:3333
+  insecure: null
+  headers: {}
+  compression: null
+  timeout: null
```
ArthurSens, line 8916: Looks strange to pass down null values in the config... maybe we should only write it down if different from nil?

what do you think

nicolastakashi: I'm good with that!
But we came to the first design we have, many null checkers on our side.

nicolastakashi: @ArthurSens don, can you check?

ArthurSens (approved):
awesome!

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: Oh, apologies!

Let me open an issue to move tracing config to CommonPrometheusFields and make we configure both Server and Agent

PR #5612, opened 2023-05-27
RemoteReadSpec: add followRedirects in order to allow to change the default value

hayk96:
## Description

**What this PR does / why we need it:**
- [PROBLEM] When using `RemoteReadSpec` it's not possible to configure/change the [follow_redirects](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_read) parameter.
- [FIX] This PR adds an ability to modify the default value of the `follow_redirects` in RemoteReadSpec.

## Type of change

## Changelog entry

```release-note
Add `followRedirects` in order to allow to change the default value in `RemoteReadSpec`.
```


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -823,6 +823,9 @@ type RemoteReadSpec struct {
 	TLSConfig *TLSConfig `json:"tlsConfig,omitempty"`
 	// Optional ProxyURL.
 	ProxyURL string `json:"proxyUrl,omitempty"`
+	// Configure whether HTTP requests follow HTTP 3xx redirects.
+	// Requires Prometheus v2.26.0 and above.
+	FollowRedirects bool `json:"followRedirects,omitempty"`
```
xiu, line 828: Can you change the type to be `*bool` and add `//+optional` to it? This will allow us to leave the default value to Prometheus.

Reference: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

xiu (commented):
Thanks for the PR!

Could you add a test case for this as well, potentially in https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg_test.go#L5420?

pkg/apis/monitoring/v1/prometheus_types.go
hayk96: Thanks for the review @xiu,
sure, I've just committed the changes

pkg/prometheus/promcfg_test.go
```diff
@@ -5474,6 +5474,44 @@ scrape_configs: []
...
+    prometheus_replica: $(POD_NAME)
+scrape_configs: []
+remote_read:
+- url: http://example.com
+  remote_timeout: 30s
+  follow_redirects: true
+`,
+		},
+		{
+			version: "v2.26.0",
```
slashpai, line 5499: I think you meant one testcase for version lesser than `2.26`?

pkg/prometheus/promcfg_test.go
```diff
@@ -5474,6 +5474,44 @@ scrape_configs: []
...
+			remoteRead: monitoringv1.RemoteReadSpec{
+				URL:             "http://example.com",
+				FollowRedirects: pointer.Bool(false),
```
ArthurSens, line 5502: Adding the same test but just different booleans might be too much 😅 

They follow the same codebase, so only one already covers everything

hayk96: @ArthurSens agree :slightly_smiling_face: Should we keep it for just one boolean in this case? (e.g. `false`)

ArthurSens: yeah, just picking one is fine!

simonpasquier:
Thanks for the contribution! Note that e2e tests are broken until #5619 gets merged.

xiu:
> Thanks for the contribution! Note that e2e tests are broken until https://github.com/prometheus-operator/prometheus-operator/pull/5619 gets merged.

@hayk96 #5619 has been merged, can you rebase this one? thanks!

simonpasquier:
Thanks! If you're motivated, a similar change could be done for the remote write configuration.

hayk96:
Yes, I'll contribute to that as well. Thank you!

PR #5617, opened 2023-05-30
cmd/prometheus-config-reloader: add SIGTERM handler

simonpasquier:
## Description

When a Prometheus pod is terminated, the container runtime sends a SIGTERM signal [1] to all pod's containers. Prior this change, prometheus-config-reloader would return a non-zero exit code in this case which can lead to the pod being reported as failed for a brief period of time and create additional work on the statefulset controller's side. To be a good citizen, prometheus-config-reloader should handle the signal and return gracefully.

[1] https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/cmd/prometheus-config-reloader/main.go
+++ b/cmd/prometheus-config-reloader/main.go
@@ -22,8 +22,10 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


philipgough (approved):
lgtm

Issue #5536, opened 2023-05-01
Custom annotations to prometheus statefulset


michalschott:
**What is missing?**
Theres no ability to inject custom annotations to statefulsets

**Why do we need it?**
In my case, I'm using [polaris](https://github.com/FairwindsOps/polaris) as a validation webhook, rule exemptions can be defined by adding annotations to objects.

`podMetadata` is not enough, as this needs to be set on controller level

simonpasquier:
Can you provide more details about what you're missing and what you'd expect to see happening?
Are you asking specifically about setting annotations to the statefulset created by the operator?

michalschott:
Hey @simonpasquier, sorry for not being clear.

Yes, annotation needs to be set on STS level managed by operator, not on pods (podMetadata).

simonpasquier:
The operator has a CLI argument (`--labels`) to add labels to all the resources that it manages. Would it help to add a `--annotations` flag?
As of today, you should be able to add the annotations to the statefulset manually and the operator should preserve them.

michalschott:
@simonpasquier Indeed I could add them manually, but for empheral clusters I'd need to do it over and over, maybe as an extra step - long therm this is bot efficient norbscalable, especially with gitops approach.

Guess adding annotations flag, similar to labels would be enough.

michalschott:
@JoaoBraveCoding @simonpasquier I opted in to add `--prometheus-annotations` flag instead, because there might be use case where we'd like not to add the same annotations to AlertManager or Thanos.

PR #5626, opened 2023-06-01
Allow to add custom annotations to prometheus StatefulSet

michalschott:
## Description

Allow to add custom annotations to prometheus StatefulSet. Fixes [5536](https://github.com/prometheus-operator/prometheus-operator/issues/5536)

## Type of change

## Changelog entry

Allow to add custom annotations to prometheus StatefulSet

```release-note
Allow to add custom annotations to prometheus StatefulSet
```


pkg/operator/config.go
```diff
@@ -58,6 +60,55 @@ type ContainerConfig struct {
 	EnableProbes  bool
 }
 
+type Annotations struct {
+	AnnotationsString string
+	AnnotationsMap    map[string]string
+}
+
+// Implement the flag.Value interface
```
michalschott, line 68: These could probably be merged with labels functions with some sort of interfaces / generics. Apologies, not that experienced ;)

simonpasquier (commented):
Thanks for the PR! I'd prefer to keep it consistent with the `--labels` flag (e.g. `--annotations`) and have the  annotations applied to all resources.

michalschott:
> Thanks for the PR! I'd prefer to keep it consistent with the `--labels` flag (e.g. `--annotations`) and have the annotations applied to all resources.

@simonpasquier no problem, actioned.

pkg/operator/config.go
```diff
@@ -58,6 +59,55 @@ type ContainerConfig struct {
...
+func (annotations *Annotations) String() string {
+	return annotations.AnnotationsString
+}
+
+// Merge labels create a new map with labels merged.
```
ArthurSens, line 72:
```suggestion
// Merge annotations create a new map with annotations merged.
```

pkg/operator/config.go
```diff
@@ -58,6 +59,55 @@ type ContainerConfig struct {
...
+			sp := strings.Split(pair, "=")
+			m[sp[0]] = sp[1]
+		}
+	}
+	(*annotations).AnnotationsMap = m
+	(*annotations).AnnotationsString = value
+	return nil
+}
+
+// Returns an arrary with the keys of the label map sorted
```
ArthurSens, line 101:
```suggestion
// Returns an arrary with the keys of the annotation map sorted
```

pkg/operator/config.go
```diff
@@ -58,6 +59,55 @@ type ContainerConfig struct {
...
+func (annotations *Annotations) SortedKeys() []string {
```
ArthurSens, line 102: Looks like SortedKeys isn't used, do we need to keep it?

ArthurSens (commented):
Just some small nits!

simonpasquier (commented):
before merging it, the annotations would need to be propagated to all resources managed by the operator (all statefulsets, configmaps, secrets, services, ...).

pkg/operator/config.go
michalschott: So true, removing it. Blind copy-paste.

michalschott:
> before merging it, the annotations would need to be propagated to all resources managed by the operator (all statefulsets, configmaps, secrets, services, ...).

@simonpasquier figuring out how to propagate to statefulsets was easy enough. Spend some time, but can't figure out - could you point me to the bits which are responsible for svc/cm/secret creation?

simonpasquier:
I would look for places which call `Labels.Merge()`, e.g.

https://github.com/prometheus-operator/prometheus-operator/blob/a7ec5d2e72f597f8411792f6b98bd1cffdcabc18/pkg/prometheus/server/operator.go#L649-L659

https://github.com/prometheus-operator/prometheus-operator/blob/a7ec5d2e72f597f8411792f6b98bd1cffdcabc18/pkg/prometheus/statefulset.go#L109-L123

There are many others...

simonpasquier:
> Shall I squash or smth?

no need to squash, we can do it on merge.

pkg/operator/config.go
```diff
@@ -58,6 +59,45 @@ type ContainerConfig struct {
...
+	for key, value := range annotations.AnnotationsMap {
+		mergedAnnotations[key] = value
+	}
+	return mergedAnnotations
+}
+
+// Set implements the flag.Set interface.
+func (annotations *Annotations) Set(value string) error {
+	m := map[string]string{}
+	if value != "" {
```
sthaha, line 89: nit: if isn't needed.

simonpasquier (approved):
Thanks!

PR #5627, opened 2023-06-01
change: deprecate `.spec.allowOverlappingBlocks`

simonpasquier:
## Description

The CLI argument is deprecated in Prometheus since v2.39.0.

## Type of change

## Changelog entry

```release-note
Deprecate .spec.allowOverlappingBlocks` since it's deprecated in Prometheus >= 2.39.0.
```


ArthurSens (approved):
Do we remove the field once e2e-test doesn't cover 2.39 anymore?

simonpasquier:
The rule would be that we remove the field if/when v2alpha1 happens. But in v1 we can't remove fields from the API, we can just say that it has no effect.

PR #5640, opened 2023-06-04
chore:Moved PrometheusTracingConfig to CommonPrometheusFields

### LINKED ISSUE #5635: Move PrometheusTracingConfig to `CommonPrometheusFields`
AFAICT it is technically possible to configure tracing in agent mode and it's even supported in remote write

https://github.com/prometheus/prometheus/blob/a8772a41782758e8153c494dcd207e770d8421a4/storage/remote/client.go#L204

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/5591#discussion_r1213268189_
            
---

We added support for configuring tracing in Prometheus Server in #5591, but as Simon pointed out, Prometheus Agents can also be configured in the same way.

Let's move the `PrometheusTracingConfig` field to [CommonPrometheusFields](https://github.com/prometheus-operator/prometheus-operator/blob/d9df32355c7b81f8737bd545270144ea606380ea/pkg/apis/monitoring/v1/prometheus_types.go#L56) and add the same configuration to [GenerateAgentConfiguration](https://github.com/prometheus-operator/prometheus-operator/blob/d9df32355c7b81f8737bd545270144ea606380ea/pkg/prometheus/promcfg.go#L2051)

GauravJ3:
## Description

This PR adds support for configuring tracing in Prometheus Server as well as Prometheus Agents

Closes #5635

## Type of change

- [] `FEATURE` (non-breaking change which adds functionality)

## Changelog entry

```release-note
Added `.spec.tracingConfig` field to PrometheusAgent CRD.
```


GauravJ3:
Hey @ArthurSens addressed this issue.. can you review the same ?
https://github.com/prometheus-operator/prometheus-operator/issues/5635

GauravJ3:
[Update] fixed a small bug I made - `added param into cg.GenerateAgentConfiguration`
I'll work on unit test and raise a PR for same in sometime.

GauravJ3:
can you help me how to fix this checks / Generate and format (ubuntu-latest) (pull_request)  I'm using windows machine not sure if this is lint issue or what.

xiu:
hey @GauravJ3, thanks for the PR! You'll need to run `make generate` and commit the generated files (see https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#contribution-flow).

GauravJ3:
Hi @xiu why `checks / Generate and format (ubuntu-latest)`  taking so long like 5min+ ? Is there any way to optimise it ?

pkg/prometheus/promcfg.go
```diff
@@ -2055,6 +2055,7 @@ func (cg *ConfigGenerator) GenerateAgentConfiguration(
 	sCons map[string]*monitoringv1alpha1.ScrapeConfig,
 	store *assets.Store,
 	additionalScrapeConfigs []byte,
+	tracingConfig *monitoringv1.PrometheusTracingConfig,
```
simonpasquier, line 2058: no need to pass a tracingConfig since it's now included in cg.prom.

```suggestion
```

pkg/prometheus/promcfg.go
```diff
@@ -2099,6 +2100,15 @@ func (cg *ConfigGenerator) GenerateAgentConfiguration(
 		cfg = append(cfg, cg.generateRemoteWriteConfig(store))
 	}
 
+	if tracingConfig != nil {
+		tracingcfg, err := cg.generateTracingConfig(tracingConfig)
```
simonpasquier, line 2104: `func (cg *ConfigGenerator) generateTracingConfig(tracingConfig *monitoringv1.PrometheusTracingConfig)`  signature could be changed to `func (cg *ConfigGenerator) generateTracingConfig()` since `cg.prom` holds the tracing config.

GauravJ3: Updated.. Can you check the new code @simonpasquier thanks :)

pkg/prometheus/promcfg.go
```diff
@@ -2238,15 +2247,18 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
 	cfg := yaml.MapSlice{}
 	objMeta := cg.prom.GetObjectMeta()
 
-	cfg = append(cfg, yaml.MapItem{
-		Key:   "endpoint",
-		Value: tracingConfig.Endpoint,
-	})
+	tracingConfig := cg.prom.GetCommonPrometheusFields().TracingConfig
 
+	if len(tracingConfig.Endpoint) > 0 {
```
simonpasquier, line 2256: API validation should already ensure that this isn't an empty string

ArthurSens: Yes, it is fine :)

GauravJ3: @ArthurSens done

pkg/prometheus/promcfg.go
```diff
@@ -511,8 +510,9 @@ func (cg *ConfigGenerator) GenerateServerConfiguration(
 		cfg = append(cfg, cg.generateRemoteReadConfig(remoteRead, store))
 	}
 
-	if tracingConfig != nil {
-		tracingcfg, err := cg.generateTracingConfig(tracingConfig)
+	if cg.prom.GetCommonPrometheusFields().TracingConfig != nil {
```
JoaoBraveCoding, line 513: nit
```suggestion
	if cpf.TracingConfig != nil {
```

pkg/prometheus/promcfg.go
```diff
@@ -2099,6 +2099,15 @@ func (cg *ConfigGenerator) GenerateAgentConfiguration(
 		cfg = append(cfg, cg.generateRemoteWriteConfig(store))
 	}
 
+	if cg.prom.GetCommonPrometheusFields().TracingConfig != nil {
```
JoaoBraveCoding, line 2102: nit
```suggestion
	if cpf.TracingConfig != nil {
```

JoaoBraveCoding (commented):
I'm not a part of the reviewers, but from my POV since this is something at the level of promcfg I don't see it as a requirement to have to test this at the Agent level.

pkg/prometheus/promcfg.go
sthaha, line 2102: just a thought: should this `nil` check be moved to `cg.generateTracingConfig` which then returns nil there by making `cfg := append(cfg, nil) a no op. So the code would read
```
		tracingcfg, err := cg.generateTracingConfig()
		if err != nil {
			return nil, errors.Wrap(err, "generating tracing configuration failed")
		}

		cfg = append(cfg, tracingcfg)
```

GauravJ3: Ahh.. good one, missed that. Thanks 👍

GauravJ3: @sthaha there are multiple test cases which then needed to be re-written, was originally thought of moving inside only

GauravJ3: Yeah sorry not test cases, but you've to return empty yaml.MapSlice{} and an error object , since it's signature is not pointer to ymal.MapSlice{} 

`func (cg *ConfigGenerator) generateTracingConfig() (yaml.MapItem, error) {..}`

GauravJ3: Yes but can change the signature to achieve it .

simonpasquier (commented):
Thanks! I think that @sthaha suggestion is interesting but it can be tackled as a follow-up.

simonpasquier (approved):

PR #5641, opened 2023-06-05
Add label selector for kubelets

hifi:
## Description

We have somewhat special setup where not all nodes are monitored by the operator. We can use selectors with most other things but for kubelets there are no selectors available.

The alternative would be to maintain our own service and endpoints but since this change is quite small I'm hoping it could be incorporated to the operator directly as we already label the nodes differently.

## Type of change

## Changelog entry

```release-note
- Add label selector for kubelets
```


cmd/operator/main.go
```diff
@@ -148,6 +148,7 @@ func init() {
 	flagset.StringVar(&cfg.TLSConfig.KeyFile, "key-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to private TLS certificate file.")
 	flagset.StringVar(&cfg.TLSConfig.CAFile, "ca-file", "", "- NOT RECOMMENDED FOR PRODUCTION - Path to TLS CA file.")
 	flagset.StringVar(&cfg.KubeletObject, "kubelet-service", "", "Service/Endpoints object to write kubelets into in format \"namespace/name\"")
+	flagset.StringVar(&cfg.KubeletSelector, "kubelet-selector", "", "Label selector to filter kubelets.")
```
simonpasquier, line 151:
```suggestion
	flagset.StringVar(&cfg.KubeletSelector, "kubelet-selector", "", "Label selector to filter nodes.")
```

simonpasquier (commented):
This is a legit use case indeed. Just one nit otherwise LGTM!

simonpasquier (approved):
Thanks!

Issue #4606, opened 2022-03-01
Support global settings and templates in the new Alertmanager's `spec.alertmanagerConfiguration` field


simonpasquier:
**What is missing?**

https://github.com/prometheus-operator/prometheus-operator/pull/4220 added the ability to reference an `AlertmanagerConfig` object  to provision the main Alertmanager configuration instead of using a raw secret.
But users can't pass custom templates and global settings as defined by https://prometheus.io/docs/alerting/latest/configuration/#configuration-file.

The extended schema should look like:

```yaml
apiVersion: monitoring.coreos.com/v1
kind: Alertmanager
metadata:
  name: main
spec:
   alertmanagerConfiguration:
     name: main
     global:
       smtpFrom: "" # string
       smtpSmarthost: "" # string
       smtpHello: "" # string
       smtpAuthUsername: "" # string
       smtpAuthIdentity: "" # string
       smtpAuthPassword: # *v1.SecretKeySelector
         key: "smtp_password"
         name: "am-credentials"
       smtpAuthSecret: # *v1.SecretKeySelector
         key: "smtp_secret"
         name: "am-credentials"      
       smtpRequireTls: true # *bool
       slackApiUrl: # *v1.SecretKeySelector
         key: "slack_api"
         name: "am-credentials"      
       victorOpsApiKey: # *v1.SecretKeySelector
         key: "vops_key"
         name: "am-credentials"
       victorOpsApiUrl: "" # string
       pagerDutyUrl: "" # string
       opsGenieApiKey: # *v1.SecretKeySelector
         key: "opsgenie_key"
         name: "am-credentials"
       opsGenieApiUrl: "" # string
       weChatApiSecret: # *v1.SecretKeySelector
         key: "wechat_key"
         name: "am-credentials"
       wechatApiUrl: "" # string
       wechatApiCorpId: "" # string
       resolveTimeout: "" # string (duration)
       httpConfig: {} # we need a new HTTPConfig type, it should be a 1:1 copy of v1alpha1.HTTPConfig
     templates: # list of ConfigMap references (e.g. []v1.ConfigMapKeySelector)
     - key: template1
       name: my-configmap
```

**Why do we need it?**

To have feature parity with the Alertmanager configuration format when using the AlertmanagerConfig CRD for the global config.

**Environment**

* Prometheus Operator version:

    `Insert image tag or Git SHA here`
    
    

**Anything else we need to know?**:

simonpasquier:
[...]

PR #5649, opened 2023-06-07
feat(alertmanager): support smtp config in alertmanagerConfiguration

mtt0:
## Description

#4606 disscussed global config's implemention, this PR support smtp global configs:

```yaml
kind: Alertmanager
metadata:
  name: main
spec:
  alertmanagerConfiguration:
    name: main
    global:
      smtp:
        smartHost: smtp.example.org:587
        authUsername: dev@smtp.example.org
        authPassword:
          name: alertmanager-main
          key: smtp-auth-password
    ...
```

## Type of change

## Changelog entry

```release-note

```


pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +345,37 @@ type AlertmanagerWebSpec struct {
 	Timeout *uint32 `json:"timeout,omitempty"`
 }
 
+// GlobalSMTPConfig configures smtp parameters that are valid in all other configuration contexts.
+// See https://prometheus.io/docs/alerting/latest/configuration/#configuration-file
+type GlobalSMTPConfig struct {
+	// The default SMTP From header field.
+	SMTPFrom string `json:"smtpFrom,omitempty"`
```
simonpasquier, line 352: To follow the [k8s recommendations](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) could you please change all fields from values to pointers?

```suggestion
	// +optional
	SMTPFrom *string `json:"smtpFrom,omitempty"`
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -270,6 +270,8 @@ type AlertmanagerConfiguration struct {
 // AlertmanagerGlobalConfig configures parameters that are valid in all other configuration contexts.
 // See https://prometheus.io/docs/alerting/latest/configuration/#configuration-file
 type AlertmanagerGlobalConfig struct {
+	*GlobalSMTPConfig `json:",inline"`
```
simonpasquier, line 273: I'd prefer to avoid type embedding and suggest to add all SMTP fields directly to the `AlertmanagerGlobalConfig` struct.

pkg/apis/monitoring/v1/alertmanager_types.go
mtt0: 👌

pkg/apis/monitoring/v1/alertmanager_types.go
mtt0: My reasons for using embeds:
1. The embedding method gathers a group of related configurations, which is convenient for writing `convertSMTPConfig`, and is convenient for being called in the `convertGlobalConfig` function, avoiding the function body being too long and making it more readable
2. In addition to automatically generating more codes, the embedded structure seems to have no other disadvantages. If so, I am sorry（QAQ

If you insist, then I would avoid embedding

simonpasquier: Embedding can create subtle hard-to-debug issues (though in practice using it here is probably safe in the long term).

Another option would be to introduce a specific field grouping all SMTP parameters together: 

```
type AlertmanagerGlobalConfig struct {
	SMTPConfig *GlobalSMTPConfig `json:"smtp,omitempty"`
...
}

type GlobalSMTPConfig struct {
	From *string `json:"from,omitempty"`
...
}
```

The manifest would look like:

```
spec:
  alertmanagerConfiguration:
    name: foo
    global:
      smtp:
        from: bar@example.com
        requireTLS: true
```

mtt0: Thanks for letting me know this.

This haved been discussed at [4606#issuecomment-1058087736](https://github.com/prometheus-operator/prometheus-operator/issues/4606#issuecomment-1058087736).

I prefer a hierarchical structure, but it's different from the flat structure of the Alertmanager configuration. While I think it's acceptable and clear, users need to read [prometheus-operator's AlertmanagerGlobalConfig documentation](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.AlertmanagerGlobalConfig) before configuring.

I expect you to have a strong opinion this time 😂, both fine by me. You decide, I'll do it

simonpasquier: :rofl: 

As of today, I'm less convinced that being close to the Alertmanager configuration format is the right choice. In any case, the field names aren't the same (e.g. `smtp_from` vs. `smtpFrom`) so you're not going to be able to copy/paste your original Alertmanager config file into the custom resource. For this reason, I'd go the way suggested in https://github.com/prometheus-operator/prometheus-operator/pull/5649#discussion_r1224432215.

pkg/alertmanager/amcfg.go
```diff
@@ -1299,6 +1306,51 @@ func makeNamespacedString(in string, crKey types.NamespacedName) string {
 	return crKey.Namespace + "/" + crKey.Name + "/" + in
 }
 
+func (cb *configBuilder) convertSMTPConfig(ctx context.Context, out *globalConfig, in monitoringv1.GlobalSMTPConfig, crKey types.NamespacedName) error {
+	if out == nil {
+		return errors.New("globalConfig is required")
+	}
```
ArthurSens, line 1312: 1 line before calling this function we're declaring `out := &globalConfig{}`, therefore `out` will never be nil. There's no need for this if :)

mtt0: Yes, you are right. I thought at first that `convertSMTPConfig` should not be aware of it. But maybe since they are both methods of cb it is ok to remove this check?

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +348,45 @@ type AlertmanagerWebSpec struct {
...
+type GlobalSMTPConfig struct {
+	// The default SMTP From header field.
+	// +optional
+	From *string `json:"from,omitempty"`
+
+	// The default SMTP smarthost used for sending emails, including port number.
+	// Port number usually is 25, or 587 for SMTP over TLS (sometimes referred to as STARTTLS).
+	// Example: smtp.example.org:587
+	// +optional
+	SmartHost *string `json:"smartHost,omitempty"`
```
simonpasquier, line 362: (suggestion) we can add some validations to the field
```
	SmartHost *HostPort `json:"smartHost,omitempty"`

...

// +kubebuilder:validation:Pattern=`^.+:.+$`
type HostPort string
```

The regexp validation isn't 100% proof but it should be good enough for catching obvious mistakes.

The other option is to store host & port in a struct:

```
	SmartHost *NetworkAddress `json:"smartHost,omitempty"`

...

type NetworkAddress struct {
    Host string `json:"host,omitempty"`
    Port string `json:"port,omitempty"`
}
```

mtt0:
It seems like CI is timeout, could you rerun this task ? @simonpasquier 
image

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +347,53 @@ type AlertmanagerWebSpec struct {
...
+	// The default SMTP TLS requirement.
+	// Note that Go does not support unencrypted connections to remote SMTP endpoints.
+	// +optional
+	RequireTLS *bool `json:"requireTLS,omitempty"`
+}
+
+// HostPort represents a "host:port" network address.
+type HostPort struct {
+	// +kubebuilder:validation:MinLength=1
+	Host string `json:"host,omitempty"`
```
simonpasquier, line 392:
```suggestion
	Host string `json:"host"`
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +347,53 @@ type AlertmanagerWebSpec struct {
...
+	// +kubebuilder:validation:MinLength=1
+	Port string `json:"port,omitempty"`
```
simonpasquier, line 394:
```suggestion
	Port string `json:"port"`
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +347,53 @@ type AlertmanagerWebSpec struct {
...
+
+	// The default SMTP TLS requirement.
+	// Note that Go does not support unencrypted connections to remote SMTP endpoints.
+	// +optional
+	RequireTLS *bool `json:"requireTLS,omitempty"`
+}
+
+// HostPort represents a "host:port" network address.
+type HostPort struct {
+	// +kubebuilder:validation:MinLength=1
```
simonpasquier, line 391:
```suggestion
	 // Defines the host's address, it can be a DNS name or a literal IP address.
	// +kubebuilder:validation:MinLength=1
```

simonpasquier (commented):
Just a few nits :)

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +347,53 @@ type AlertmanagerWebSpec struct {
...
+	Host string `json:"host,omitempty"`
+	// +kubebuilder:validation:MinLength=1
```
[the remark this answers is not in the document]
mtt0: Can be port name？

mtt0:
> Just a few nits :)

👌，the commits will be rebased, and the generated files will also be updated

pkg/apis/monitoring/v1/alertmanager_types.go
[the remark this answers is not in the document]
simonpasquier: yes because it uses [net.SplitHostPort()](https://pkg.go.dev/net#SplitHostPort) to parse the host:port string :)
See https://pkg.go.dev/net#Dial too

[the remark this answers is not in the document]
mtt0: Thanks, alreay done

JoaoBraveCoding (approved):
from my side lgtm

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +347,55 @@ type AlertmanagerWebSpec struct {
...
+// GlobalSMTPConfig configures global SMTP parameters.
+// See https://prometheus.io/docs/alerting/latest/configuration/#configuration-file
+type GlobalSMTPConfig struct {
+	// The default SMTP From header field.
+	// +optional
+	From *string `json:"from,omitempty"`
+
+	// The default SMTP smarthost used for sending emails, including port number.
+	// Port number usually is 25, or 587 for SMTP over TLS (sometimes referred to as STARTTLS).
+	// Example: smtp.example.org:587
```
simonpasquier, line 359: (suggestion) since the HostPort struct splits the host and port into independent and mandatory fields, we can shorten the description.

```suggestion
	// The default SMTP smarthost used for sending emails.
```

simonpasquier (commented):
Almost LGTM. I promise that it's my last nit :)

pkg/apis/monitoring/v1/alertmanager_types.go
mtt0: 🤣 OKay Okay, lol

simonpasquier (approved):
Thanks!

Issue #5657, opened 2023-06-09
ScrapeConfig should support limits


dmitriishaburov:
**What is missing?**

Currently there is no way to configure limits (`sample_limit/etc...`) via ScrapeConfig while other CRDs (i.e. ServiceMonitor) supports them. Also `enforced...` limits are not enforced on ScrapeConfig-generated scrapes.

**Why do we need it?**

That's missing feature that may stop from using ScrapeConfig.

**Anything else we need to know?**:

I can take that feature to implement.

PR #5658, opened 2023-06-09
feat: limits support for ScrapeConfig

dmitriishaburov:
## Description

Closes https://github.com/prometheus-operator/prometheus-operator/issues/5657.

- Adds support for setting following fields in ScrapeConfig: `sampleLimit`, `targetLimit`, `labelLimit`, `labelNameLengthLimit` and `labelValueLengthLimit`
- ScrapeConfig now respects `enforcedSampleLimit`, `enforcedTargetLimit`, `enforcedLabelLimit`, `enforcedLabelNameLengthLimit`, `enforcedLabelValueLengthLimit` and `enforcedBodySizeLimit` configured on Prometheus level

> Feel free to work on it. One advice: it might be simpler to review if you send 2 PR: one to take into account the existing enforced* values and another one to implement the limit fields in the ScrapeConfig CRD.

Not sure if that makes sense to split it into 2 PRs, because `AddLimitsToYAML` function excects both enforced and object-level limits. In general excluding autogenerated code PR is quite small.

Skipped unit test since there is general test for limits that test them against ServiceMonitor, but other objects (PodMonitor/Probe) don't have that test.

## Type of change

## Changelog entry

```release-note
- Add scrape limits support to ScrapeConfig CRD
```


pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -101,6 +101,24 @@ type ScrapeConfigSpec struct {
 	// Authorization header to use on every scrape request.
 	// +optional
 	Authorization *v1.SafeAuthorization `json:"authorization,omitempty"`
+	// SampleLimit defines per-scrape limit on number of scraped samples that will be accepted.
+	// +optional
+	SampleLimit uint64 `json:"sampleLimit,omitempty"`
```
xiu, line 106: We'll need those types to be pointer as per https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required since `uint64` doesn't have a `nil` value.

xiu (commented):
Looking good! Can you also add some tests for these new fields? Thanks!

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
dmitriishaburov: `AddLimitsToYAML` expects limit to be `uint64`, not `*uint64` and checks if it's not zero [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L208-L211).
So I assime it's better to remove `+optional` comment instead of modifying existing logic everywhere or dereference these pointers?

simonpasquier: I'd be in favor of updating `AddLimitsToYAML`()

dmitriishaburov: Since that would affect existing ServiceMonitor / PodMonitor / Probe behaviour as they use it as well, that goes above my confidence level in codebase changes :)

Some help would be appreciated in that case

ArthurSens: That's a valid concern @dmitriishaburov, thanks for sharing!

Maybe we could create a PR updating the old fields to pointer first @xiu @simonpasquier? Then a rebase on this one should make it easier

xiu: Good idea, was thinking of a follow up at first but might as well do it beforehand :) Opened #5668 for that (still draft waiting for e2e to pass).

ArthurSens:
@dmitriishaburov, we merged [a PR](https://github.com/prometheus-operator/prometheus-operator/pull/5668) from @xiu that updates `AddLimitsToYaml()` to use string pointers instead of raw strings.

If you rebase your PR on top of main, it should be good to continue :)

dmitriishaburov:
@ArthurSens thanks! I'll rebase and update my PR on Monday and add unit test as well.

dmitriishaburov:
Automatically closed due to sync with upstream.
Changed to *uint64 and added unit test

pkg/prometheus/promcfg.go
```diff
@@ -2187,6 +2187,16 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 
 	cfg = cg.addSafeAuthorizationToYaml(cfg, fmt.Sprintf("scrapeconfig/auth/%s/%s", sc.Namespace, sc.Name), store, sc.Spec.Authorization)
 
+	cfg = cg.AddLimitsToYAML(cfg, sampleLimitKey, sc.Spec.SampleLimit, cpf.EnforcedSampleLimit)
+	cfg = cg.AddLimitsToYAML(cfg, targetLimitKey, sc.Spec.TargetLimit, cpf.EnforcedTargetLimit)
+	cfg = cg.AddLimitsToYAML(cfg, labelLimitKey, sc.Spec.LabelLimit, cpf.EnforcedLabelLimit)
+	cfg = cg.AddLimitsToYAML(cfg, labelNameLengthLimitKey, sc.Spec.LabelNameLengthLimit, cpf.EnforcedLabelNameLengthLimit)
+	cfg = cg.AddLimitsToYAML(cfg, labelValueLengthLimitKey, sc.Spec.LabelValueLengthLimit, cpf.EnforcedLabelValueLengthLimit)
```
JoaoBraveCoding, line 2194: Shouldn't these be preceded by `cg.WithMinimumVersion("2.27.0")`?

JoaoBraveCoding (commented):
Just one comment but apart from that lgtm

pkg/prometheus/promcfg.go
dmitriishaburov: It's handled inside [AddLimitsToYAML](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L181-L221), so there's no need to duplicate

pkg/prometheus/promcfg.go
```diff
@@ -2187,6 +2187,16 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+
+	if cpf.EnforcedBodySizeLimit != "" {
+		cfg = cg.WithMinimumVersion("2.28.0").AppendMapItem(cfg, "body_size_limit", cpf.EnforcedBodySizeLimit)
+	}
```
ArthurSens, line 2198: For consistency, could we add this to AddLimitsToYAML as well?

dmitriishaburov: As I can see `AddLimitsToYAML` takes both CRD (probe, servicemonitor etc) and global Prometheus limits and adds proper limit accordingly.

In `body_size_limit` case there is no CRD-level configuration, so that would be somewhat hacky. Passing `nil` every time as 3rd argument?

pkg/prometheus/promcfg_test.go
```diff
@@ -8980,6 +8980,30 @@ scrape_configs:
 scrape_configs:
 - job_name: scrapeconfig/default/testscrapeconfig1
   scheme: https
+`,
+		},
+		{
```
simonpasquier, line 8985: can you add 2 other tests?
* 1 with the enforced* limits + scrapeconfig limits being defined
* 1 with only the enforced* limits being defined

dmitriishaburov: There are `TestSampleLimits`, `TestTargetLimits`, `TestLabelLimits`, `TestLabelNameLengthLimits` and `TestLabelValueLengthLimits` ([example](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg_test.go#L6319-L6535)) tests that already testing same functionality and many other behaviours of `AddLimitToTAML` function which reused in that PR.

Not sure if that needs to be duplicated specifically for ScrapeConfig.

simonpasquier: hmm ok, I think that we can extend the existing unit tests to cover all combinations (pod monitors, service monitors, probes and scrapeconfigs) but it doesn't have to be in this PR.

simonpasquier (approved):
Thanks! Can you rebase to resolve the conflicts?

dmitriishaburov:
@simonpasquier done

PR #5668, opened 2023-06-13
chore: Move limits to *uint64 and add optional markers

xiu:
## Description

References:
* https://github.com/prometheus-operator/prometheus-operator/pull/5658#discussion_r1227145581
* https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

## Type of change

## Changelog entry

n/a

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -206,15 +206,16 @@ var (
 // AddLimitsToYAML appends the given limit key to the configuration if
 // supported by the Prometheus version.
 func (cg *ConfigGenerator) AddLimitsToYAML(cfg yaml.MapSlice, k limitKey, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
-	if limit == nil && enforcedLimit == nil {
+	finalLimit := getLimit(limit, enforcedLimit)
+	if finalLimit == nil {
 		return cfg
 	}
 
 	if k.minVersion == "" {
-		return cg.AppendMapItem(cfg, k.prometheusField, getLimit(limit, enforcedLimit))
+		return cg.AppendMapItem(cfg, k.prometheusField, finalLimit)
 	}
 
-	return cg.WithMinimumVersion(k.minVersion).AppendMapItem(cfg, k.prometheusField, getLimit(limit, enforcedLimit))
+	return cg.WithMinimumVersion(k.minVersion).AppendMapItem(cfg, k.prometheusField, finalLimit)
 }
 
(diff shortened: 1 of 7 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -1346,14 +1346,14 @@ func generateRunningFilter() yaml.MapSlice {
...
 	if enforced != nil {
-		if user < *enforced && user != 0 || *enforced == 0 {
-			return user
+		if user != nil && *user < *enforced && *user != 0 || *enforced == 0 {
+			return *user
 		}
 		return *enforced
 	}
-	return user
+	return 0
```
simonpasquier, line 1356: wouldn't it be better to return nil and the caller would set the field only if it's not nil?

xiu: @ArthurSens Good point, if we now return a *uint64, we can do that.

xiu: Updated in 17c9d21d9218a6f48a3c5918f735f5ea916a4adc

pkg/prometheus/promcfg.go
```diff
@@ -205,16 +205,21 @@ var (
 
 // AddLimitsToYAML appends the given limit key to the configuration if
 // supported by the Prometheus version.
-func (cg *ConfigGenerator) AddLimitsToYAML(cfg yaml.MapSlice, k limitKey, limit uint64, enforcedLimit *uint64) yaml.MapSlice {
-	if limit == 0 && enforcedLimit == nil {
+func (cg *ConfigGenerator) AddLimitsToYAML(cfg yaml.MapSlice, k limitKey, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
+	if limit == nil && enforcedLimit == nil {
+		return cfg
+	}
```
simonpasquier, line 211: (nit) `getLimit()` already covers this case?

```suggestion
```

xiu: true! updated :)

simonpasquier (approved):
lgtm! I'll defer to @ArthurSens who's taking care of the release if this should merge or wait :)

JoaoBraveCoding (approved):
lgtm

ArthurSens:
> lgtm! I'll defer to @ArthurSens who's taking care of the release if this should merge or wait :)

Somehow I missed this message 😅 

Thanks @xiu for tackling this one, it makes life easier for the new contributor :)

PR #5677, opened 2023-06-14
tests: refactors probe test into generic test function

JoaoBraveCoding:
## Description

Follow up to https://github.com/prometheus-operator/prometheus-operator/pull/5586 as to make the review easier, this PR removed the test `TestProbeStaticTargetsConfigGeneration` and splits this test into 3 separate tests as to test different parts of the code in isolation

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -846,6 +846,9 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	return cfg
 }
 
+// generateProbeConfig builds the prometheus configuration for a probe. This function
+// assumes that it will never receive a probe with empty targets, since the
+// operator filters those in the validation step in SelectProbes().
 func (cg *ConfigGenerator) generateProbeConfig(
 	m *monitoringv1.Probe,
 	apiserverConfig *monitoringv1.APIServerConfig,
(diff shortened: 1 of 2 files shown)


ArthurSens (approved):
Thanks!

pkg/prometheus/promcfg.go
```diff
@@ -950,7 +953,7 @@ func (cg *ConfigGenerator) generateProbeConfig(
 		xc := labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, m.Spec.Targets.StaticConfig.RelabelConfigs)
 		relabelings = append(relabelings, generateRelabelConfig(xc)...)
 		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
-	} else {
+	} else if m.Spec.Targets.Ingress != nil {
```
simonpasquier, line 956: suggestion: I would replace the `if () { ... } else if () {...}` structure by switch/case which I find more readable.

JoaoBraveCoding: I'm not seeing/understand how I can translate this `if else` block into a switch case, can you elaborate a bit?

simonpasquier: sure!

```golang
switch {
case m.Spec.Targets.StaticConfig != nil: // static targets defined
    ...
case m.Spec.Targets.Ingress != nil: // ingress targets defined
    ...
default:
    // The caller ensures that at least one of StaticConfig or Ingress is defined.
}

```

pkg/prometheus/promcfg.go
```diff
@@ -907,7 +910,8 @@ func (cg *ConfigGenerator) generateProbeConfig(
 	}
 	labeler := namespacelabeler.New(cpf.EnforcedNamespaceLabel, cpf.ExcludedFromEnforcement, false)
 
-	if m.Spec.Targets.StaticConfig != nil {
+	switch {
```
simonpasquier, line 913: (nit) emphasize why the order is important?
```suggestion
	// As stated in the CRD documentation, if both StaticConfig and Ingress are defined, the former takes precedence which is why the first case statement checks for m.Spec.Targets.StaticConfig.
	switch {
```

simonpasquier (approved):

PR #5680, opened 2023-06-15
chore: update e2e tests to Kubernetes v1.27.3

philipgough:
## Description

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -2184,7 +2184,6 @@ func testAMWeb(t *testing.T) {
 func testAlertManagerMinReadySeconds(t *testing.T) {
 	// Don't run Alertmanager tests in parallel. See
 	// https://github.com/prometheus/alertmanager/issues/1835 for details.
-	runFeatureGatedTests(t)
 
 	testCtx := framework.NewTestCtx(t)
 	defer testCtx.Cleanup(t)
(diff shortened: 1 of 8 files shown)


test/e2e/alertmanager_test.go
```diff
@@ -2140,7 +2140,6 @@ func testAMWeb(t *testing.T) {
 func testAlertManagerMinReadySeconds(t *testing.T) {
 	// Don't run Alertmanager tests in parallel. See
 	// https://github.com/prometheus/alertmanager/issues/1835 for details.
-	runFeatureGatedTests(t)
```
simonpasquier, line 2143: should we delete the `runFeatureGatedTests()` func and references to `FEATURE_GATED_TESTS` in the GH action workflow as well?

simonpasquier: No strong opinion as long as we document that it's not used for now.

simonpasquier (approved):

test/framework/helpers.go
```diff
@@ -94,7 +94,7 @@ func (f *Framework) WaitForPodsReady(ctx context.Context, namespace string, time
 		for _, p := range pl.Items {
 			isRunningAndReady, err := k8sutil.PodRunningAndReady(p)
 			if err != nil {
-				return false, err
+				return false, nil
```
philipgough, line 97: so from what I can see, it looks like we started to hit a race condition here with Deletions for thanos ruler in the latest version of K8s 

Essentially, Pods were being Completed before this code was executed so it would always exit with an error.

Much of the use of this function checks for 0 replicas so it might be best to split it into something like WaitForListEmpty and check that the ListOptions has no Pods to support that use case.

simonpasquier: Agreed. We could also use foreground deletion in the Delete*AndWaitUntilGone() functions and wait that the top-level resource is gone rather than checking (indirectly) that the pods are missing.

simonpasquier (approved):

PR #5690, opened 2023-06-15
feat: enable config reload via signal

### LINKED ISSUE #3381: Secure Prometheus Config Reload

simonpasquier:
## Description

Requires https://github.com/thanos-io/thanos/pull/6453

Closes #3381

## Type of change

## Changelog entry

```release-note
Add `.spec.reloadStrategy` to the Prometheus and PrometheusAgent CRDs. The value can be `HTTP` (default if not specified) or `ProcessSignal`.
```


pkg/prometheus/agent/statefulset.go
```diff
@@ -188,13 +187,8 @@ func makeStatefulSetSpec(
 		return nil, err
 	}
 
-	webRoutePrefix := "/"
-	if cpf.RoutePrefix != "" {
-		webRoutePrefix = cpf.RoutePrefix
-	}
-
```
ArthurSens, line 195: Nice clean up by moving this to CommonPrometheusFields!

cmd/prometheus-config-reloader/main.go
```diff
@@ -227,3 +253,91 @@ func createOrdinalEnvvar(fromName string) error {
...
+
+	lastReload  time.Time
+	lastSuccess bool
+}
+
+func (prt *prometheusReloadTracker) getConfigStatus(ctx context.Context) (bool, time.Time, error) {
+	r, err := http.NewRequestWithContext(ctx, "GET", prt.baseURL.JoinPath("/api/v1/status/runtimeinfo").String(), nil)
+	if err != nil {
+		return false, time.Time{}, err
+	}
```
ArthurSens, line 270: Nice idea to get reload status even when using signal 👍

cmd/prometheus-config-reloader/main.go
```diff
@@ -227,3 +253,91 @@ func createOrdinalEnvvar(fromName string) error {
...
+		return false, time.Time{}, fmt.Errorf("%s: invalid status code: %d", r.URL.String(), resp.StatusCode)
+	}
+
+	var runtimeInfo = struct {
+		Status string `json:"status"`
+		Data   struct {
+			ReloadConfigSuccess bool      `json:"reloadConfigSuccess"`
+			LastConfigTime      time.Time `json:"lastConfigTime"`
+		} `json:"data"`
+	}{}
```
ArthurSens, line 288: Do we know if prometheus has a history of changing the API response here? I wonder if we'll need to have a test to make sure we don't break when upgrading prometheus versions

ArthurSens (commented):
Looking good 👍 , let's see how the PR upstream evolves

cmd/prometheus-config-reloader/main.go
simonpasquier: Yes otherwise it would be like shooting in the dark. This was the blocking point for me to move this forward.

cmd/prometheus-config-reloader/main.go
simonpasquier: https://prometheus.io/docs/prometheus/latest/stability/

The v1 API should be stable. The API returns these fields since at least late 2019 so I wouldn't check the exact Prometheus version (if it ever happens, the only advice would be to upgrade).

cmd/prometheus-config-reloader/main.go
```diff
@@ -227,3 +253,91 @@ func createOrdinalEnvvar(fromName string) error {
...
+		return false, time.Time{}, err
+	}
+
+	resp, err := prt.client.Do(r)
+	if err != nil {
+		return false, time.Time{}, fmt.Errorf("%s: %w", r.URL.String(), err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode/100 != 2 {
```
nicolastakashi, line 278: TIL
So cleaver.
Better than check if >= 200 <= 299

cmd/prometheus-config-reloader/main.go
```diff
@@ -227,3 +253,91 @@ func createOrdinalEnvvar(fromName string) error {
...
+		// The previous configuration was valid but the current configuration
+		// isn't so there's no need to poll again.
+		if !success && prt.lastSuccess {
+			return fmt.Errorf("configuration reload failed")
+		}
+
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("%w: last configuration status: success=%v, success_timetamp=%v", ctx.Err(), success, last)
+		case <-t.C:
```
nicolastakashi, line 340: Dont you need to reset the timer?

simonpasquier: No it will automatically resend to the channel at the next tick.

nicolastakashi:
LGTM

nicolastakashi (approved):

go.mod
```diff
@@ -83,7 +84,7 @@ require (
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
-	github.com/google/uuid v1.3.1 // indirect
+	github.com/google/uuid v1.3.1
```
mrueg, line 87: should this move up to the direct dependencies then?

simonpasquier: this is being addressed separately in https://github.com/prometheus-operator/prometheus-operator/pull/6114

simonpasquier:
The Thanos PR has merged and this is now ready for a final review.

ArthurSens (approved):
Nice job!

Issue #5455, opened 2023-03-28
Reduce code duplication from `createOrUpdateConfigurationSecret()` 


ArthurSens:
There might be room for reducing the duplication between agent & server? Again this can be a follow-up.

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/5385#discussion_r1140401155_
            
---

More precisely, the configuration from PrometheusAgent is a subset from Prometheus. Parts that are similar can probably be centralized in `pkg/prometheus` and parts that are exclusive for Prometheus Server can stay at `pkg/prometheus/server`

sthaha:
I am happy to take this up if no one is working on. 
cc: @slashpai @simonpasquier

PR #5700, opened 2023-06-19
refactor: move common steps from server and agent to prompkg

sthaha:
## Description

This commit partially address https://github.com/prometheus-operator/prometheus-operator/issues/5455 by refactoring server and agent operator code that updates the prometheus config secrets. It moves the common functionalities to pkg/prometheus (prompkg).

The refactor adds the following functions

Store Related
  * AddRemoteWritesToStore
  * AddRemoteReadToStore
  * AddAPIServerConfigToStore

Operator Related
  * MakeCompressedSecretForPrometheus

GH: https://github.com/prometheus-operator/prometheus-operator/issues/5455

## Type of change

- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.) refactor

### DIFF
--- a/pkg/prometheus/statefulset.go
+++ b/pkg/prometheus/statefulset.go
@@ -15,12 +15,14 @@
 package prometheus
 
 import (
+	"bytes"
 	"fmt"
 	"net/url"
 	"path"
 	"path/filepath"
 	"strings"
 
+	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -125,6 +127,22 @@ func MakeConfigSecret(p monitoringv1.PrometheusInterface, config operator.Config
 			ConfigFilename: {},
 		},
 	}
+}
+
+func MakeCompressedSecretForPrometheus(p monitoringv1.PrometheusInterface, config operator.Config, data []byte) (*v1.Secret, error) {
+
+	s := MakeConfigSecret(p, config)
+	s.ObjectMeta.Annotations = config.Annotations.Merge(map[string]string{
+		"generated": "true",
+	})
+
+	var buf bytes.Buffer
+	if err := operator.GzipConfig(&buf, data); err != nil {
+		return nil, errors.Wrap(err, "failed to gzip config")
+	}
+	s.Data[ConfigFilename] = buf.Bytes()
+
+	return s, nil
 }
 
 func ConfigSecretName(p monitoringv1.PrometheusInterface) string {
--- a/pkg/prometheus/store.go
+++ b/pkg/prometheus/store.go
@@ -1,0 +1,90 @@
+// Copyright 2023 The prometheus-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


sthaha:
I have currently marked the PR as a draft to ensure all tests pass, and that the maintainers can comment on the approach before adding more. 

cc: @simonpasquier

pkg/prometheus/store.go
```diff
@@ -0,0 +1,90 @@
...
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/assets"
+)
+
+func AddRemoteWritesToStore(ctx context.Context, store *assets.Store, namespace string, remotes []monv1.RemoteWriteSpec) error {
```
sthaha, line 27: Unit tests?

pkg/prometheus/statefulset.go
```diff
@@ -127,6 +129,22 @@ func MakeConfigSecret(p monitoringv1.PrometheusInterface, config operator.Config
 	}
 }
 
+func MakeCompressedSecretForPrometheus(p monitoringv1.PrometheusInterface, config operator.Config, data []byte) (*v1.Secret, error) {
```
simonpasquier, line 132: (suggestion) I would 
* Merge `MakeConfigSecret` and `MakeConfigurationSecret` so the resulting function includes the compression of the raw config.
* get rid of `MakeEmptyConfigurationSecret`  as it should be equivalent to calling `MakeConfigurationSecret(p, c.config, nil)` 
```suggestion
func MakeConfigurationSecret(p monitoringv1.PrometheusInterface, config operator.Config, data []byte) (*v1.Secret, error) {
```

pkg/prometheus/store.go
```diff
@@ -0,0 +1,90 @@
...
+	if store == nil && len(remotes) > 0 {
```
simonpasquier, line 28: I wouldn't do defensive programming. If store is nil, we have a bug and it's better to crash.

ArthurSens: Hey, taking this opportunity to learn a bit more about the operator, what exactly are the stores? Why would it be a bug if they are nil?

sthaha: How about we return an error rather than crash? I feel something odd about terminating a program abruptly, especially when we can handle that state gracefully ?

simonpasquier: a nil store means that the developer forgot to pass a valid instance which is always a bug.

simonpasquier: A store is a struct where we keep data referenced by secret or configmap keys such as TLS certificates, bearer tokens, .... If the same secret key is referenced by several resources, the operator will get it once from the Kube API. It also serves as a way to drop resources that reference non-existing data.

pkg/prometheus/statefulset.go
```diff
@@ -127,6 +129,22 @@ func MakeConfigSecret(p monitoringv1.PrometheusInterface, config operator.Config
...
+
+	s := MakeConfigSecret(p, config)
+	s.ObjectMeta.Annotations = config.Annotations.Merge(map[string]string{
+		"generated": "true",
+	})
```
sthaha, line 137: @simonpasquier this part is different to the old code that I moved.

pkg/prometheus/store.go
ArthurSens: While unit tests would be helpful for quick tests during development, we already have e2e tests that wouldn't pass if a regression happens here.

I wouldn't block the PR :)

ArthurSens (approved):

JoaoBraveCoding (approved):
lgtm

Issue #5681, opened 2023-06-15
Report deprecated fields more clearly


simonpasquier:
**What is missing?**

While the operator prints [warning logs](https://github.com/prometheus-operator/prometheus-operator/blob/daf6ac1e25c9ea85137a617f0b80810f8d8dd85a/pkg/prometheus/server/operator.go#L1383-L1412) when it notices usage of deprecated fields in custom resources, it's hard for users to notice it. In particular when they don't have access to the operator's logs (e.g. they may be able to provision Prometheus resources but the operator is managed by another team).

Suggested improvements:
* Report usage of deprecated fields in the Reconciled condition. The condition's status should be `True` but the reason and message fields can convey additional information (such as `reason: DeprecatedFieldsInUse` and `message: Field xxx is deprecated and may be removed in a future version, please use yyy instead.`).
* Report a gauge metric per resource signaling how many deprecated fields are used.
* Create an alert from the metric.

We also need to make sure that all deprecated fields are reported (for instance `prometheusRulesExcludedFromEnforce` is not). 

**Why do we need it?**

Better feedback to users who are referencing fields that are being deprecated.

**Environment**

* Prometheus Operator version:

    `Insert image tag or Git SHA here`
    
    

**Anything else we need to know?**:

simonpasquier:
@JoaoBraveCoding good idea!
we already have an issue about implementing validating webhook for service monitors, pod monitors and probes (#4481) but nothing for Prometheus et al.

rexagod:
I'd like to work on this. 🙂

slashpai:
@rexagod Go ahead :)

rexagod:
@simonpasquier WDYT about using [gatekeeper](https://open-policy-agent.github.io/gatekeeper-library/website/validation/requiredlabels) to validate and mutate objects? It would allow us to largely cut down on writing boilerplate code that can be conveniently specified as [reGo snippets](https://www.openpolicyagent.org/docs/latest/policy-language/), if it's a specific usecase. Otherwise, there's already a collection of [pre-existing `contraints`](https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/general) to choose from that can be reused and don't require us to write any logic ([`contraintTemplate`s](https://open-policy-agent.github.io/gatekeeper-library/website/validation/requiredlabels#template)) whatsoever.

This would ensure we can easily add checks before things are persisted in `etcd` and scale them without a fraction of the effort.

[...]

PR #5706, opened 2023-06-20
feature: prometheus validation webhooks

rexagod:
## Description

prometheus validation webhooks that (currently) only check for deprecated field and warn if that's the case.

Fixes: #5681

## Type of change

## Changelog entry

```release-note
prometheus validation webhooks that (currently) only check for deprecated field and warn if that's the case.
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -434,12 +434,32 @@ func run() int {
 		}
 	}
 
-	validationTriggeredCounter := prometheus.NewCounter(prometheus.CounterOpts{
+	promRuleMutationTiggeredCounter := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_rule_mutation_triggered_total",
+		Help: "Number of times a prometheusRule object triggered mutation",
+	})
+
+	promRuleMutationErrorsCounter := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_rule_mutation_errors_total",
+		Help: "Number of errors that occurred while mutating a prometheusRules object",
+	})
+
+	promValidationTriggeredCounter := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_prometheus_validation_triggered_total",
+		Help: "Number of times a prometheus object triggered validation",
+	})
+
+	promValidationErrorsCounter := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_prometheus_validation_errors_total",
+		Help: "Number of errors that occurred while validating a prometheus object",
+	})
+
+	promRuleValidationTriggeredCounter := prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "prometheus_operator_rule_validation_triggered_total",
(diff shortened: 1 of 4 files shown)


cmd/operator/main.go
```diff
@@ -376,12 +376,22 @@ func Main() int {
...
+	})
+
+	promValidationErrorsCounter := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_operator_rule_validation_errors_total",
+		Help: "Number of errors that occurred while validating a prometheusRules object",
+	})
+
+	promRuleValidationTriggeredCounter := prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "prometheus_operator_rule_validation_triggered_total",
 		Help: "DEPRECATED, removed in v0.57.0: Number of times a prometheusRule object triggered validation",
```
ArthurSens, line 391: Hmmm, looks like we deprecated the metrics itself? From a [codesearch](https://github.com/search?q=repo%3Aprometheus-operator%2Fprometheus-operator+validationErrorsCounter&type=code), looks like we're not increasing anywhere in the code. Is that correct @simonpasquier, @slashpai? What are we using instead?

ArthurSens (commented):
I noticed that you use the same counter for both situations:
* An error occurred while trying to validate
* The resource was successfully validated, and the validation shows usage of deprecated fields

I believe this should be handled in separate metrics :)

cmd/operator/main.go
simonpasquier: the Kubernetes API already exposes metrics about webhooks (e.g. `apiserver_admission_webhook_admission`_duration) so there's no need for the operator to repeat them. I'd be fine dropping the deprecated metrics in the operator (in a separate PR).

pkg/admission/admission.go
```diff
@@ -241,6 +267,52 @@ func (a *Admission) mutatePrometheusRules(ar v1.AdmissionReview) *v1.AdmissionRe
 	return reviewResponse
 }
 
+func (a *Admission) validatePrometheus(ar v1.AdmissionReview) *v1.AdmissionResponse {
```
simonpasquier, line 270: Like we do for validating PrometheusRules and AlertmanagerConfig resources, it would be good to avoid duplicating the validation logic in several places otherwise it's going to be harder to maintain in the long run.

cmd/operator/main.go
simonpasquier: @rexagod can you open another PR to remove the deprecated metrics as a first step? It'd make this PR a bit smaller.

rexagod: > From a [codesearch](https://github.com/search?q=repo%3Aprometheus-operator%2Fprometheus-operator+validationErrorsCounter&type=code), looks like we're not increasing anywhere in the code.

[IIUC, it seems they are.](https://github.com/prometheus-operator/prometheus-operator/pull/6024/files#diff-b51086a9eecc900f44fb39e672c73f66396d84c48d15f5d86c8106cea04a4920L270) I've opened up the PR just in case we still want to work towards deprecating these metrics (or not).

simonpasquier: Deprecated metrics have been dropped in #6024.

PR #5709, opened 2023-06-21
feat: Adding Consul support to ScrapeConfig CRD

### LINKED ISSUE #5629: Add support for consul_sd_config to scrapeConfigs CRD
**What is missing?**
Currently ScrapeConfig CRD doesn't now support consul_sd_config option.

**Why do we need it?**
Since Consul is a tool used for service Discovery, adding this to ScrapeConfig CRD would expand the use cases of this CRD. As currently to use consul for service discovery we need to rely on additionalScrapeConfigs to do so.
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#consul_sd_config

nicolastakashi:
I'm adding support to use Consul on Scrape CRD

## Description

Closes #5629

## Type of change

## Changelog entry

```release-note

```


simonpasquier:
@nicolastakashi I've updated the description to link the issue.

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
 	// +optional
 	TLSConfig *v1.SafeTLSConfig `json:"tlsConfig,omitempty"`
 }
+
+// HTTPSDConfig defines a prometheus HTTP service discovery configuration
```
simonpasquier, line 188:
```suggestion
// ConsulSDConfig defines a Consul service discovery configuration
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+// See https://prometheus.io/docs/prometheus/latest/configuration/configuration/#consul_sd_config
+// +k8s:openapi-gen=true
+type ConsulSDConfig struct {
+	// A valid string consisting of a hostname or IP followed by an optional port number.
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Server string `json:"server,omitempty"`
```
simonpasquier, line 195:
```suggestion
	Server string `json:"server"`
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	Datacenter *string `json:"datacenter,omitempty"`
+	// Namespaces are only supported in Consul Enterprise.
+	// +optional
+	Namespace *string `json:"namespace,omitempty"`
+	// Admin Partitions are only supported in Consul Enterprise.
+	// +optional
+	Partition *string `json:"partition,omitempty"`
+	// HTTP Scheme default "http"
+	// +optional
+	Scheme *string `json:"scheme,omitempty"`
```
simonpasquier, line 210: can we make it an enum of (HTTP, HTTPS)?

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
simonpasquier: We already have the `HostPort` struct for AlertmanagerConfig.
Maybe we can have a new `HostWithOptionalPort` struct? The downside is that we drift from the original Prometheus configuration format but we'll get a better/cleaner validation.

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	// A list of services for which targets are retrieved. If omitted, all services are scraped.
+	// +optional
+	Services []string `json:"services,omitempty"`
+	// An optional list of tags used to filter nodes for a given service. Services must contain all tags in the list.
+	// +optional
+	Tags []string `json:"tags,omitempty"`
+	// The string by which Consul tags are joined into the tag label.
+	// default ","
```
simonpasquier, line 218: (nit) I wouldn't mention the Prometheus default value as it might be confusing and potentially out of sync Maybe this instead?

```suggestion
	// If unset, Prometheus uses its default value.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	// +optional
+	TagSeparator *string `json:"tag_separator,omitempty"`
+	// Node metadata key/value pairs to filter nodes for a given service.
+	// +optional
```
simonpasquier, line 222:
```suggestion
	// +mapType:=atomic
	// +optional
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	// Namespaces are only supported in Consul Enterprise.
+	// +optional
+	Namespace *string `json:"namespace,omitempty"`
+	// Admin Partitions are only supported in Consul Enterprise.
+	// +optional
+	Partition *string `json:"partition,omitempty"`
+	// HTTP Scheme default "http"
+	// +optional
+	Scheme *string `json:"scheme,omitempty"`
+	// A list of services for which targets are retrieved. If omitted, all services are scraped.
```
simonpasquier, line 211:
```suggestion
	// A list of services for which targets are retrieved. If omitted, all services are scraped.
	// +listType:=atomic
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	// +optional
+	Services []string `json:"services,omitempty"`
+	// An optional list of tags used to filter nodes for a given service. Services must contain all tags in the list.
```
simonpasquier, line 214:
```suggestion
	// An optional list of tags used to filter nodes for a given service. Services must contain all tags in the list.
	// +listType:=atomic
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	Tags []string `json:"tags,omitempty"`
+	// The string by which Consul tags are joined into the tag label.
+	// default ","
+	// +optional
+	TagSeparator *string `json:"tag_separator,omitempty"`
+	// Node metadata key/value pairs to filter nodes for a given service.
+	// +optional
+	NodeMeta map[string]string `json:"node_meta,omitempty"`
+	// Allow stale Consul results (see https://www.consul.io/api/features/consistency.html). Will reduce load on Consul.
+	// default true
```
simonpasquier, line 225: same comment about default

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	// +optional
+	AllowStale *bool `json:"allow_stale"`
+	// The time after which the provided names are refreshed.
+	// On large setup it might be a good idea to increase this value because the catalog will change all the time.
+	// default "30s"
```
simonpasquier, line 230: ditto

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	RefreshInterval *v1.Duration `json:"refresh_interval,omitempty"`
```
simonpasquier, line 231:
```suggestion
	// +optional
	RefreshInterval *v1.Duration `json:"refresh_interval,omitempty"`
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+	// default ","
+	// +optional
+	TagSeparator *string `json:"tag_separator,omitempty"`
+	// Node metadata key/value pairs to filter nodes for a given service.
+	// +optional
+	NodeMeta map[string]string `json:"node_meta,omitempty"`
+	// Allow stale Consul results (see https://www.consul.io/api/features/consistency.html). Will reduce load on Consul.
+	// default true
+	// +optional
+	AllowStale *bool `json:"allow_stale"`
```
simonpasquier, line 227:
```suggestion
	AllowStale *bool `json:"allow_stale,omitempty"`
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,85 @@ type HTTPSDConfig struct {
...
+// See https://prometheus.io/docs/prometheus/latest/configuration/configuration/#consul_sd_config
+// +k8s:openapi-gen=true
+type ConsulSDConfig struct {
+	// A valid string consisting of a hostname or IP followed by an optional port number.
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Server string `json:"server,omitempty"`
+	// Consul ACL Token, If not provided it will use the ACL from the local Consul Agent.
+	// +optional
+	Token *corev1.SecretKeySelector `json:"token,omitempty"`
```
simonpasquier, line 198: (suggestion) a Ref suffix makes it more obvious that it's a reference to an external object, not a scalar value.
```suggestion
	TokenRef *corev1.SecretKeySelector `json:"tokenRef,omitempty"`
```

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,173 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 			Value: configs,
 		})
 	}
+
+	//ConsulSDConfig
+	if len(sc.Spec.ConsulSDConfigs) > 0 {
```
simonpasquier, line 2287: The generateScrapeConfig() method starts being quite long. As a follow-up, it would be nice to break it up in smaller pieces.

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,173 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+		configs := make([][]yaml.MapItem, len(sc.Spec.ConsulSDConfigs))
+		for i, config := range sc.Spec.ConsulSDConfigs {
+			configs[i] = cg.addBasicAuthToYaml(configs[i], fmt.Sprintf("scrapeconfig/%s/%s/consulsdconfig/%d", sc.Namespace, sc.Name, i), store, config.BasicAuth)
```
simonpasquier, line 2290: before referencing the assets from the store, we need to validate and load them in `*ResourceSelector.SelectScrapeConfigs()`.

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,173 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+				Value: config.Server,
+			})
+
+			if config.Token != nil {
+				value, err := store.GetKey(context.Background(), sc.Namespace, monitoringv1.SecretOrConfigMap{
+					Secret: config.Token,
+				})
+
+				if err != nil {
+					level.Error(cg.logger).Log("failed to read %s secret %s", config.Token.Name, jobName)
```
simonpasquier, line 2309: this should be checked (and rejected) in `SelectScrapeConfigs()` too.

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,173 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+			if config.ProxyConnectHeader != nil {
+				proxyConnectHeader := make(map[string]string, len(config.ProxyConnectHeader))
+
+				for k, v := range config.ProxyConnectHeader {
+					value, err := store.GetKey(context.Background(), sc.Namespace, monitoringv1.SecretOrConfigMap{
+						Secret: &v,
+					})
+
+					if err != nil {
+						level.Error(cg.logger).Log("failed to read %s secret %s", v.Name, jobName)
```
simonpasquier, line 2419: same here

simonpasquier (commented):
Very nice! I've reviewed the API definition in depth so many comments but nothing being a showstopper.

pkg/prometheus/promcfg.go
nicolastakashi: Can you take a look again @simonpasquier?
I'm not sure if I changed on the right place

pkg/prometheus/promcfg.go
nicolastakashi: I'm assuming if we're rejecting in the `*ResourceSelector.SelectScrapeConfigs()` I don't need to check for errors, since it's getting from in the memory store.

am I wrong?

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
nicolastakashi: I like that @simonpasquier but maybe we should have a kind of shared type where we can reuse this because I do believe other places may leverage those things.

WDYT?

If you think it's a good idea, where should I create this shared type? I'm bit lost yet in terms of where I should create things on the project

pkg/prometheus/promcfg.go
simonpasquier: you're correct :)

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
simonpasquier: If we'll only use this new struct for the Kubernetes SD config, it's fine to put it in `pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go`. If/when it gets reused, we can discuss a new location.

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,90 @@ type HTTPSDConfig struct {
...
+	// +optional
+	Datacenter *string `json:"datacenter,omitempty"`
+	// Namespaces are only supported in Consul Enterprise.
+	// +optional
+	Namespace *string `json:"namespace,omitempty"`
+	// Admin Partitions are only supported in Consul Enterprise.
+	// +optional
+	Partition *string `json:"partition,omitempty"`
+	// HTTP Scheme default "http"
+	// +kubebuilder:validation:Enum=http;https
```
simonpasquier, line 209: Kube API conventions recommend upper-case for acronyms.
```suggestion
	// +kubebuilder:validation:Enum=HTTP;HTTPS
```

pkg/prometheus/resource_selector.go
```diff
@@ -677,6 +677,45 @@ func (rs *ResourceSelector) SelectScrapeConfigs(ctx context.Context, listFn List
...
+			if err = rs.store.AddSafeTLSConfig(ctx, sc.GetNamespace(), config.TLSConfig); err != nil {
+				rejectFn(sc, err)
+				continue
+			}
+
+			tokenRef := monitoringv1.SecretOrConfigMap{
+				Secret: config.TokenRef,
+			}
+
+			if _, err = rs.store.GetKey(ctx, sc.GetNamespace(), tokenRef); err != nil {
```
simonpasquier, line 702:
```suggestion
			if config.TokenRef != nil {
				if _, err = rs.store.GetSecretKey(ctx, sc.GetNamespace(), config.TokenRef); err != nil {
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,90 @@ type HTTPSDConfig struct {
...
+	// that should be excluded from proxying. IP and domain names can
+	// contain port numbers.
+	// +optional
+	NoProxy *string `json:"no_proxy,omitempty"`
+	// Use proxy URL indicated by environment variables (HTTP_PROXY, https_proxy, HTTPs_PROXY, https_proxy, and no_proxy)
+	// If unset, Prometheus uses its default value.
+	// +optional
+	ProxyFromEnvironment *bool `json:"proxy_from_environment,omitempty"`
+	// Specifies headers to send to proxies during CONNECT requests.
+	// +optional
```
simonpasquier, line 260: (nit)
```suggestion
	// +mapType:=atomic
	// +optional
```

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,166 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+
+			configs[i] = append(configs[i], yaml.MapItem{
+				Key:   "server",
+				Value: config.Server,
+			})
+
+			if config.TokenRef != nil {
+				value, _ := store.GetKey(context.Background(), sc.GetNamespace(), monitoringv1.SecretOrConfigMap{
+					Secret: config.TokenRef,
+				})
```
simonpasquier, line 2308: forget my previous comment, we still need to check for err because in this case, the secret data isn't kept in the assets store. Eventually we should extend the store to support any type of data but it should/can be a follow-up.

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,166 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+
+			if config.TagSeparator != nil {
+				configs[i] = append(configs[i], yaml.MapItem{
+					Key:   "tag_separator",
+					Value: config.TagSeparator,
+				})
+			}
+
+			if len(config.NodeMeta) > 0 {
+				configs[i] = append(configs[i], yaml.MapItem{
```
simonpasquier, line 2366: I'm half sure that the keys will be alphabetically sorted in the output YAML but it's worth checking. The same comment for the other map[string]... fields.

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,166 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+					Key:   "proxy_from_environment",
+					Value: config.ProxyFromEnvironment,
+				})
+			}
+
+			if config.ProxyConnectHeader != nil {
+				proxyConnectHeader := make(map[string]string, len(config.ProxyConnectHeader))
+
+				for k, v := range config.ProxyConnectHeader {
+					value, _ := store.GetKey(context.Background(), sc.GetNamespace(), monitoringv1.SecretOrConfigMap{
```
simonpasquier, line 2411: same comment here about err checking and `ctx`.

pkg/prometheus/resource_selector.go
```diff
@@ -677,6 +677,45 @@ func (rs *ResourceSelector) SelectScrapeConfigs(ctx context.Context, listFn List
...
+				rejectFn(sc, err)
+				continue
+			}
+
+			for _, v := range config.ProxyConnectHeader {
+				_, err := rs.store.GetKey(context.Background(), sc.GetNamespace(), monitoringv1.SecretOrConfigMap{
+					Secret: &v,
+				})
+
+				if err != nil {
```
simonpasquier, line 712:
```suggestion
				if _, err := rs.store.GetSecretKey(ctx, sc.GetNamespace(), v); err != nil {
```

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,166 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+				configs[i] = addSafeTLStoYaml(configs[i], sc.GetNamespace(), *config.TLSConfig)
+			}
+
+			configs[i] = append(configs[i], yaml.MapItem{
+				Key:   "server",
+				Value: config.Server,
+			})
+
+			if config.TokenRef != nil {
+				value, _ := store.GetKey(context.Background(), sc.GetNamespace(), monitoringv1.SecretOrConfigMap{
```
simonpasquier, line 2306: let's use `ctx` :)

simonpasquier (commented):
Very cool!

pkg/prometheus/promcfg.go
nicolastakashi: If I'm not wrong we're not propagating the context to the ConfigGenerator, should we do that? @simonpasquier

pkg/prometheus/resource_selector.go
nicolastakashi: I can't pass the value to the GetSecret since this function is waiting for a `monitoringv1.SecretOrConfigMap` and the field type is a `*corev1.SecretKeySelector`

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
nicolastakashi: Out of my curiosity, why we don't need optional here?

simonpasquier: I only suggested to add `+mapType:=atomic` but optional is still needed :)

pkg/prometheus/promcfg.go
simonpasquier: ah my bad! yes I'd recommend to pass the context from the caller and I think that you'll need to modify GenerateServerConfiguration().

pkg/prometheus/resource_selector.go
simonpasquier: hmm [GetSecretKey()](https://github.com/prometheus-operator/prometheus-operator/blob/ec63444c5053123664afea98f1df14e972dbed76/pkg/assets/store.go#L335) expects a `SecretKeySelector`?

pkg/prometheus/promcfg.go
```diff
@@ -2281,6 +2282,176 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+			})
+
+			if config.TokenRef != nil {
+				value, err := store.GetKey(context.Background(), sc.GetNamespace(), monitoringv1.SecretOrConfigMap{
+					Secret: config.TokenRef,
+				})
+
+				if err != nil {
+					level.Error(cg.logger).Log("failed to read %s secret %s", v.Name, jobName)
+					continue
```
simonpasquier, line 2312: It's probably best to return an error: if it's a transient failure, the next reconciliation would succeed. If it's permanent (the secret has been deleted), we want the user to know ASAP.

nicolastakashi: Yup!
Makes sense

pkg/prometheus/resource_selector.go
nicolastakashi: Oh! NVM my bad

pkg/prometheus/promcfg.go
nicolastakashi: @simonpasquier what do you mean by checking? Resort the Map?

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,91 @@ type HTTPSDConfig struct {
...
+
+// ConsulSDConfig defines a Consul service discovery configuration
+// See https://prometheus.io/docs/prometheus/latest/configuration/configuration/#consul_sd_config
+// +k8s:openapi-gen=true
+type ConsulSDConfig struct {
+	// A valid string consisting of a hostname or IP followed by an optional port number.
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Server string `json:"server"`
+	// Consul ACL TokenRef, If not provided it will use the ACL from the local Consul Agent.
```
ArthurSens, line 196:
```suggestion
	// Consul ACL TokenRef, if not provided it will use the ACL from the local Consul Agent.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,91 @@ type HTTPSDConfig struct {
...
+	// +optional
+	TokenRef *corev1.SecretKeySelector `json:"tokenRef,omitempty"`
+	// Consul Datacenter name, If not provided it will use the local Consul Agent Datacenter.
```
ArthurSens, line 199:
```suggestion
	// Consul Datacenter name, if not provided it will use the local Consul Agent Datacenter.
```

pkg/prometheus/promcfg.go
ArthurSens: 99% I understood what simon means here 😅 

Maps are not sorted in Go, if we have multiple keys in a map there's a good chance that generating the configuration is not [deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm#:~:text=In%20computer%20science%2C%20a%20deterministic,the%20same%20sequence%20of%20states.). This can cause config reloads when there is no need

To solve this, we need to sort the map before writing the configuration

ArthurSens (commented):
Really cool to see progress here ❤️ 

Like simon said, I wouldn't block the PR for lack of an e2e-test, but it would be nice to see as a follow-up.

I'd suggest using anything that is able to spin-up containers using Golang, e.g. https://testcontainers.com/

pkg/prometheus/promcfg.go
nicolastakashi: Oh cool!
I'm going to add this 😄

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
nicolastakashi: @simonpasquier During this implementation I was wondering which kind of validation you have in mind, because looking at the host port structure, they are both strings, without any especial validation, since the port could be any port number or port name (when using k8s services) I can't see which kind of validation we could apply.

I'm afraid of this structure didn't provide any extra value for us (Of course we will have a meaningful type but without robust type validation).

WDYT?

pkg/prometheus/promcfg.go
simonpasquier: Thanks @ArthurSens for reading in my thoughts :sweat_smile: 
I _think_ that the YAML library already takes care of consistent ordering but worth checking.

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
simonpasquier: The main advantage from my point of view is that you avoid obvious issues like "foo:" or ":8080" but I agree that it might not be worth the effort for now...
I've looked into how the Consul library parses the configured address and it's more complicated than that:
https://github.com/hashicorp/consul/blob/7689a5ef2df8a92f476aabdb6ebd6e3696f8c39b/api/api.go#L722-L755
Also CRD validation rules can't really support all the possible [cases](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/net/url/url.go;l=215-220) for host values :shrug: 

We can at least validate that when defined (e.g. not nil), the server string isn't empty.

nicolastakashi: Ok!
Makes sense, I'm thinking if not makes sense we use only one structure named Host where we wave an option field named port and apply all the validations against that.

If you think makes sense, we can address it on another issue since my idea is to change this in every place we're using plain strings to handle host values.

WDYT? @simonpasquier

simonpasquier: I agree that it should/can be addressed later. For now let's keep it simple (e.g. the `server` field should be of length > 0). Since the API is v1alpha1, we're still allowed to "break" (a bit) the interface.

pkg/prometheus/promcfg.go
nicolastakashi: Done @ArthurSens @simonpasquier can you take another look?

pkg/prometheus/agent/statefulset_test.go
```diff
@@ -192,7 +193,7 @@ func newLogger() log.Logger {
 
 func makeStatefulSetFromPrometheus(p monitoringv1alpha1.PrometheusAgent) (*appsv1.StatefulSet, error) {
 	logger := newLogger()
-	cg, err := prompkg.NewConfigGenerator(logger, &p, false)
+	cg, err := prompkg.NewConfigGenerator(context.TODO(), logger, &p, false)
```
slashpai, line 196: IIRC we had removed `context.TODO()` in code base a while back, so `context.Background()` would be better here :)

pkg/prometheus/promcfg.go
```diff
@@ -59,10 +60,11 @@ type ConfigGenerator struct {
 	notCompatible          bool
 	prom                   monitoringv1.PrometheusInterface
 	endpointSliceSupported bool
+	ctx                    context.Context
```
slashpai, line 63: Its preferred to pass context as first argument to function than having it part of struct  
https://go.dev/blog/context-and-structs

pkg/prometheus/promcfg_test.go
```diff
@@ -65,10 +67,9 @@ func mustNewConfigGenerator(t *testing.T, p *monitoringv1.Prometheus) *ConfigGen
 	if p == nil {
 		p = &monitoringv1.Prometheus{}
 	}
-
 	logger := level.NewFilter(log.NewLogfmtLogger(os.Stderr), level.AllowWarn())
 
-	cg, err := NewConfigGenerator(log.With(logger, "test", t.Name()), p, false)
+	cg, err := NewConfigGenerator(context.TODO(), log.With(logger, "test", t.Name()), p, false)
```
slashpai, line 72:
```suggestion
	cg, err := NewConfigGenerator(context.Background(), log.With(logger, "test", t.Name()), p, false)
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -437,7 +438,7 @@ func TestStatefulSetVolumeInitial(t *testing.T) {
 
 	logger := newLogger()
 
-	cg, err := prompkg.NewConfigGenerator(logger, &p, false)
+	cg, err := prompkg.NewConfigGenerator(context.TODO(), logger, &p, false)
```
slashpai, line 441:
```suggestion
	cg, err := prompkg.NewConfigGenerator(context.Background(), logger, &p, false)
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -54,7 +55,7 @@ func newLogger() log.Logger {
 func makeStatefulSetFromPrometheus(p monitoringv1.Prometheus) (*appsv1.StatefulSet, error) {
 	logger := newLogger()
 
-	cg, err := prompkg.NewConfigGenerator(logger, &p, false)
+	cg, err := prompkg.NewConfigGenerator(context.TODO(), logger, &p, false)
```
slashpai, line 58:
```suggestion
	cg, err := prompkg.NewConfigGenerator(context.Background(), logger, &p, false)
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -894,7 +895,7 @@ func TestPrometheusDefaultBaseImageFlag(t *testing.T) {
 		},
 	}
 
-	cg, err := prompkg.NewConfigGenerator(logger, &p, false)
+	cg, err := prompkg.NewConfigGenerator(context.TODO(), logger, &p, false)
```
slashpai, line 898:
```suggestion
	cg, err := prompkg.NewConfigGenerator(context.Background(), logger, &p, false)
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -948,7 +949,7 @@ func TestThanosDefaultBaseImageFlag(t *testing.T) {
...
```
slashpai, line 952:
```suggestion
	cg, err := prompkg.NewConfigGenerator(context.Background(), logger, &p, false)
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -1551,7 +1552,7 @@ func TestReplicasConfigurationWithSharding(t *testing.T) {
...
```
slashpai, line 1555:
```suggestion
	cg, err := prompkg.NewConfigGenerator(context.Background(), logger, &p, false)
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -1606,7 +1607,7 @@ func TestSidecarResources(t *testing.T) {
 			Spec: monitoringv1.PrometheusSpec{},
 		}
 
-		cg, err := prompkg.NewConfigGenerator(logger, &p, false)
+		cg, err := prompkg.NewConfigGenerator(context.TODO(), logger, &p, false)
```
slashpai, line 1610:
```suggestion
		cg, err := prompkg.NewConfigGenerator(context.Background(), logger, &p, false)
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -2010,7 +2011,7 @@ func TestConfigReloader(t *testing.T) {
 	logger := newLogger()
 	p := monitoringv1.Prometheus{}
 
-	cg, err := prompkg.NewConfigGenerator(logger, &p, false)
+	cg, err := prompkg.NewConfigGenerator(context.TODO(), logger, &p, false)
```
slashpai, line 2014:
```suggestion
	cg, err := prompkg.NewConfigGenerator(context.Background(), logger, &p, false)
```

pkg/prometheus/promcfg.go
```diff
@@ -84,6 +86,7 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
 		version:                version,
 		prom:                   p,
 		endpointSliceSupported: endpointSliceSupported,
+		ctx:                    ctx,
```
slashpai, line 89: Its preferred to pass context as first argument to function than having it part of struct

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -180,3 +184,91 @@ type HTTPSDConfig struct {
...
+	// If unset, Prometheus uses its default value.
+	// +optional
+	FollowRedirects *bool `json:"follow_redirects,omitempty"`
+	// Whether to enable HTTP2.
+	// If unset, Prometheus uses its default value.
+	// +optional
+	EnableHttp2 *bool `json:"enable_http2,omitempty"`
+	// TLS Config
+	// +optional
+	TLSConfig *v1.SafeTLSConfig `json:"tlsConfig"`
```
slashpai, line 273:
```suggestion
	TLSConfig *v1.SafeTLSConfig `json:"tlsConfig,omitempty"`
```

slashpai (commented):
Great work on this, just a few nits

pkg/prometheus/promcfg.go
nicolastakashi: Cool! Thanks for sharing @slashpai I didn't know this practice.
Can you take a look again?

ArthurSens (approved):
Just one single nit, but everything else looks good!

PR #5671, opened 2023-06-13
Discord integration

Azanul:
## Description
Prometheus Alertmanager brings the functionality of pushing notifications to Discord using webhooks.
Fixes #5251 

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [X] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
CRD extension for notifications on Discord
```

PR #5714, opened 2023-06-23
Hide generated files from Github diff

ArthurSens:
## Description

We've recently enabled a Github app that labels PRs by their diff size. Any PR that touches generated files is almost always labeled as XXL, which is not always true if we ignore generated files.

This PR properly configured generated files to be ignored in the Github diff, so they get labeled correctly. See: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,6 @@
 **/zz_generated.*.go linguist-generated=true
+bundle.yaml linguist-generated=true
+example/prometheus-operator-crd-full/* linguist-generated=true
+example/prometheus-operator-crd/* linguist-generated=true
+example/jsonnet/prometheus-operator/* linguist-generated=true
+Documentation/api.md linguist-generated=true


ArthurSens:
As a consequence, changes to the files added to .gitattributes will be compacted by default(even if it is a one-liner), and will require an extra step if someone wants to read the diff

simonpasquier (commented):
pkg/client/applyconfiguration/* too?

ArthurSens:
Not sure how, but I was looking at [this PR](https://github.com/prometheus-operator/prometheus-operator/pull/5671/files#diff-09c85842dbddee018f59f5687c85e991ab393f6ea10788c82d9226ba2c7caf6f) as reference and it looks like all files under `pkg/client` are already compacted by default?

There is a small diff of 14 lines and it is already compacted 🤔

simonpasquier (approved):
ah ok, nevermid then!

Issue #5708, opened 2023-06-20
Missing reload, after alertmanager template has been changed.


jkroepke:
**What happened?**

Since #3647, it is possible to store template on the alertmanager config secret. I have to upgrade the templates, nothing happends.

**Did you expect to see something different?**

I would expect a config reload.

**How to reproduce it (as minimally and precisely as possible)**:

Follow the guide

https://github.com/prometheus-operator/prometheus-operator/blob/c840b7eaa857fc45f8a9be52cfd8d315f7c88e9f/Documentation/user-guides/alerting.md#using-a-kubernetes-secret

and manage the Altermanager trough an Kubernetes Secret. After apply the config, modify the template inside the Kubernetes secret and reapply the diff. 

**Environment**

* Prometheus Operator version:

  `v0.65.2`

* Kubernetes version information:

    `Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.3", GitCommit:"9e644106593f3f4aa98f8a84b23db5fa378900bd", GitTreeState:"clean", BuildDate:"2023-03-15T13:33:12Z", GoVersion:"go1.19.7", Compiler:"gc", Platform:"linux/arm64"}`
    

* Kubernetes cluster kind:

minikube

l646505418:
hi @simonpasquier ，I am interested in this issue and can  work on it

jkroepke:
feel free to start @l646505418

l646505418:
@jkroepke  Hi, would mind telling me how to reproduce this issue? Because I tried to modify the alertmanage.yml in Kubernetes secrets by dashboard, I can see the alertmanager.yaml under  path: /etc/alertmanager/config/ of pod is updated too

jkroepke:
If you read https://github.com/prometheus-operator/prometheus-operator/blob/c840b7eaa857fc45f8a9be52cfd8d315f7c88e9f/Documentation/user-guides/alerting.md#using-a-kubernetes-secret, you can store templates on the secret, too, e.g.: `template_1.tmpl`

Modify the templates would not result into a reload.

l646505418:
> If you read https://github.com/prometheus-operator/prometheus-operator/blob/c840b7eaa857fc45f8a9be52cfd8d315f7c88e9f/Documentation/user-guides/alerting.md#using-a-kubernetes-secret, you can store templates on the secret, too, e.g.: `template_1.tmpl`
> 
> Modify the templates would not result into a reload.

[...]

PR #5727, opened 2023-07-05
alertmanager: add etc/alertmanager/config

l646505418:
## Description

 add etc/alertmanager/config to watched-dir when init pods to support config reload when template files are modified.
#5708 

## Type of change

```

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -569,7 +569,7 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 		})
 	}
 
-	watchedDirectories := []string{}
+	watchedDirectories := []string{alertmanagerConfigDir}
 	configReloaderVolumeMounts := []v1.VolumeMount{
 		{
 			Name:      alertmanagerConfigVolumeName,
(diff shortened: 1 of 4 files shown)


simonpasquier:
Thanks for the PR! 

It addresses the issue highlighted in #5708 but I think that there are other cases that aren't covered. In particular templates can  be defined via `.spec.alertmanagerConfiguration.templates`  or also passed via `.spec.secrets`/`.spec.configmaps`. Those cases can be addressed in a separate PR though.

simonpasquier (approved):

l646505418:
> Thanks for the PR!
> 
> It addresses the issue highlighted in #5708 but I think that there are other cases that aren't covered. In particular templates can be defined via `.spec.alertmanagerConfiguration.templates` or also passed via `.spec.secrets`/`.spec.configmaps`. Those cases can be addressed in a separate PR though.

For templates that are  passed via `.spec.secrets`/`.spec.configmaps` is already supported by operator. And  for templates that are defined via `.spec.alertmanagerConfiguration.templates`, I think it is also supported by operator(I don't check that).😂

l646505418:
> 🤦 your're 100% correct for configmaps and secrets! for templates, I don't think that it's implemented though.

Hi, I found that when user specify the templates via `.spec.alertmanagerConfiguration.templates` ,the files will be mounted to path `etc/alertmanager/config`, so this PR also resolved the template reload for templates that are defined via `.spec.alertmanagerConfiguration.templates`.

jkroepke:
> for templates, I don't think that it's implemented though.

They are implemented, that was the reason for my issue.

PR #5732, opened 2023-07-07
Run PrometheusVersionUpgrade tests separately

ArthurSens:
## Description

As our compatibility matrix grows, the tests assuring we can upgrade from one Prometheus version to the next are taking up to 1~2h to complete. We aim to accelerate our e2e tests that are required to pass on Pull Requests by moving this test to a separate CI, that runs once a day.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/test-prom-version-upgrade.yaml
+++ b/.github/workflows/test-prom-version-upgrade.yaml
@@ -1,0 +1,50 @@
+name: e2e
+on:
+  schedule:
+    - cron:  '37 15 * * *' # Every day 15:37
+
+jobs:
+  e2e-tests:
+    name: E2E tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Import environment variables from file
+      run: cat ".github/env" >> $GITHUB_ENV
+    - name: Install Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '${{ env.golang-version }}'
+    - name: Build images
+      run: |
+        export SHELL=/bin/bash
+        make build image
+    - name: Start KinD
+      uses: helm/kind-action@v1.7.0
+      with:
+        version: ${{ env.kind-version }}
+        node_image: ${{ env.kind-image }}
+        wait: 300s
+        config: ./test/e2e/kind-conf.yaml
+        cluster_name: e2e
+    - name: Wait for cluster to finish bootstraping
+      run: |
+        kubectl wait --for=condition=Ready pods --all --all-namespaces --timeout=300s
+        kubectl cluster-info
+        kubectl get pods -A
+    - name: Load images
+      run: |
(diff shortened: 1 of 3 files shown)


simonpasquier:
Another option would be to run these compatibility tests post-commit (e.g. after merge) rather than pre-commit. FWIW I've never seen that they detected a regression but I guess it's better to keep them :shrug:

ArthurSens:
> Another option would be to run these compatibility tests post-commit (e.g. after merge) rather than pre-commit. FWIW I've never seen that they detected a regression but I guess it's better to keep them 🤷

Exactly because we've never seen a regression that it sounded not so important to run on every PR. And agreed that it is important, therefore, we're not deleting the test, just running less frequently :)

xiu (approved):
LGTM! It should speed things up :)

xiu:
One thing I just thought of: will this run for every branch or just `main`? If only `main`, should we have it run on the release branches as well (provided there were changes in them)?

nicolastakashi:
> One thing I just thought of: will this run for every branch or just `main`? If only `main`, should we have it run on the release branches as well (provided there were changes in them)?

I don't know if worth run those tests on release branches, since those branches are usually changed when critical security issues or bugs are reported and I'm not sure if any of those conditions will broke anything regarding Prometheus version.

.github/workflows/test-prom-version-upgrade.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Install Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '${{ env.golang-version }}'
+    - name: Build images
+      run: |
+        export SHELL=/bin/bash
+        make build image
+    - name: Start KinD
+      uses: engineerd/setup-kind@v0.5.0
```
simonpasquier, line 23: can you use helm/kind-action instead, similar to #5817?

simonpasquier (commented):
I agree with @nicolastakashi and I'm fine with running the upgrade tests only on the main branch periodically.

simonpasquier (approved):
LGTM

ArthurSens:
Alright, the first run failed 😅 

https://github.com/prometheus-operator/prometheus-operator/actions/runs/6027041323

Not sure why though, I don't see any differences from the original e2e-tests

simonpasquier:
Hmm it seems that the CRDs aren't installed.

ArthurSens:
I basically copy-pasted from the e2e-tests workflow

Rereading the e2e-tests workflow, I also don't see when the CRDs are installed... why does it work there? 🤔

simonpasquier:
CRDs are installed at the start of the e2e tests by `framework.CreateOrUpdatePrometheusOperator()` I believe.

Additional note: the workflows are named the same in the GitHub UI.
image

Issue #5176, opened 2022-11-21
Support _param_target based sharding for Probes


soasme:
**What is missing?**
The current prometheus-operator does not support sharding static config targets defined in a single Probe, because "sharding is done on the contents of the `__address__` target meta-label." but the `__address__` label is set to the Prober.URL, rather than Targets.StaticConfig.Static[*].

**Why do we need it?**
Enable more even distribution of all targets defined in a Probe.

**Environment**

* Prometheus Operator version:

    `0.60.1`
    
    

**Anything else we need to know?**:
Proposed code change: [source promcfg.go L1060-L1063](https://github.com/prometheus-operator/prometheus-operator/blob/v0.60.1/pkg/prometheus/promcfg.go#L1060-L1063), (not tested though):

```diff
func (cg *ConfigGenerator) generateProbeConfig(
	m *v1.Probe,
	apiserverConfig *v1.APIServerConfig,
	store *assets.Store,
	shards int32,
) yaml.MapSlice {
                 ...
		// Add configured relabelings.
		xc := labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, m.Spec.Targets.StaticConfig.RelabelConfigs)
		relabelings = append(relabelings, generateRelabelConfig(xc)...)
+               relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)    
		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
   
```

It seems such a code change sets hashmod and modulus for `_param_target`, rather than `__address__`.

steven05jiang:
@simonpasquier please review the [fix](https://github.com/prometheus-operator/prometheus-operator/pull/5735). thanks!

PR #5735, opened 2023-07-11
support sharding for probe

steven05jiang:
## Description

fix https://github.com/prometheus-operator/prometheus-operator/issues/5176. 
By adding a sharding relabeling rule, prometheus operator will shard targets in the probe across the prometheus cluster.

## Type of change

## Changelog entry

```release-note
support sharding for probe.
```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -957,6 +957,7 @@ func (cg *ConfigGenerator) generateProbeConfig(
 		// Add configured relabelings.
 		xc := labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, m.Spec.Targets.StaticConfig.RelabelConfigs)
 		relabelings = append(relabelings, generateRelabelConfig(xc)...)
+		relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)
 		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
 
 	case m.Spec.Targets.Ingress != nil:
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -909,6 +909,7 @@ func (cg *ConfigGenerator) generateProbeConfig(
 			},
 		}...)
 	}
+	relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)
 	labeler := namespacelabeler.New(cpf.EnforcedNamespaceLabel, cpf.ExcludedFromEnforcement, false)
 
 	// As stated in the CRD documentation, if both StaticConfig and Ingress are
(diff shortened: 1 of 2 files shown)


soasme (commented):
The fix looks good to me.

pkg/prometheus/promcfg.go
```diff
@@ -957,6 +957,7 @@ func (cg *ConfigGenerator) generateProbeConfig(
 		// Add configured relabelings.
 		xc := labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, m.Spec.Targets.StaticConfig.RelabelConfigs)
 		relabelings = append(relabelings, generateRelabelConfig(xc)...)
+		relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)
```
ArthurSens, line 960: Do we want to configure sharing on every occasion? Or only when sharding is enabled in Prometheus?

ArthurSens (commented):
Hello 👋 , sorry for taking this long to review... this one is a bit out of my comfort zone 😅

pkg/prometheus/promcfg.go
steven05jiang: thanks for the suggestion. this is the change [e4ded8d](https://github.com/prometheus-operator/prometheus-operator/pull/5735/commits/e4ded8d7fd91d4d9874cb87a4cb88c1e31f1b2a2)

pkg/prometheus/promcfg.go
```diff
@@ -909,6 +909,7 @@ func (cg *ConfigGenerator) generateProbeConfig(
 			},
 		}...)
 	}
+	relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)
```
simonpasquier, line 912: I'd keep it as the last relabeling config.

simonpasquier: @steven05jiang yes I think that it makes sense to do so.

pkg/prometheus/promcfg.go
```diff
@@ -1053,10 +1052,11 @@ func (cg *ConfigGenerator) generateProbeConfig(
 
 		// Add configured relabelings.
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, m.Spec.Targets.Ingress.RelabelConfigs))...)
-		relabelings = generateAddressShardingRelabelingRulesForProbes(relabelings, shards)
+	}
 
+	if m.Spec.Targets.StaticConfig != nil || m.Spec.Targets.Ingress != nil {
```
simonpasquier, line 1057: I don't think that the if statement is required
```suggestion
```

simonpasquier (commented):
one small nit otherwise lgtm!

pkg/prometheus/promcfg.go
simonpasquier: This is a good remark. The operator already takes care of filtering out probes without static/ingress targets [here](https://github.com/prometheus-operator/prometheus-operator/blob/4e37c62781ce19971726f5a34b073327a3860d1a/pkg/prometheus/resource_selector.go#L497-L500). But even if the check doesn't work, it shouldn't be a problem to include the relabeling configs.

steven05jiang: Sounds good. This is the update ffb2a35

ArthurSens (approved):
good job @steven05jiang and @simonpasquier!

sorry for not being very useful here 😅

simonpasquier (approved):
Thanks!

steven05jiang:
Thanks all for the review!

PR #5743, opened 2023-07-12
Update github.com/thanos-io/thanos

captncraig:
## Description

This is a simple update to github.com/thanos-io/thanos, which is only used in prometheus-config-reloader. The reason for this is a single fix to the reloader package which prevents deadlock for us in certain cases: [this change](https://github.com/thanos-io/thanos/pull/6519)

## Type of change


metalmatze (approved):

ArthurSens:
I wonder if we shouldn't announce this one as a bug fix as well? 🤔

captncraig:
I was a little torn on that. It wouldn't hurt to put it in the changelog, but it is also a little hard to explain why its a problem, and the fix is entirely upstream. By all means, add it to the changelog if you think its worth it.

simonpasquier:
Thanks! This is definitely a bugfix, I'll add a note to the changelog.

PR #5748, opened 2023-07-17
support mount multi PVC templates

l646505418:
## Description

Modify the spec.Storage.VolumeClaimTemplate to spec.Storage.VolumeClaimTemplates to support mount multi PVC templates.

## Type of change

- [`x` ] `FEATURE` (non-breaking change which adds functionality)

## Changelog entry

```
Modify the spec.Storage.VolumeClaimTemplate to spec.Storage.VolumeClaimTemplates to support mount multi PVC templates.

```


l646505418:
Hi team, the only way to accomplish this feature is to modify the CRD, but because I am not very familiar with operator and I use the Windows, it is hard for me to generate the CRDs and some clientset files. Maybe you guys can review and genetra,thanks

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -837,7 +837,7 @@ type StorageSpec struct {
 	// Defines the PVC spec to be used by the Prometheus StatefulSets.
 	// The easiest way to use a volume that cannot be automatically provisioned
 	// is to use a label selector alongside manually created PersistentVolumes.
-	VolumeClaimTemplate EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate,omitempty"`
+	VolumeClaimTemplates []EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate,omitempty"`
```
xiu, line 840: Since Prometheus is a `v1` CRD, we can't remove a field without bumping the version (e.g. `v2`). In this case, we'd have to add a new field but not remove the previous one for compatibility purposes in order to stick to `v1`.

xiu (commented):
Thanks for the contribution! I added one comment.

> Hi team, the only way that I came up with to accomplish this feature is to modify the CRD, but because I am not very familiar with operator and I use the Windows, it is hard for me to generate the CRDs and some clientset files. Maybe you guys can review and genetra,thanks

You can use `make generate` to regenerate CRDs and other client files. See https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#contribution-flow

pkg/apis/monitoring/v1/prometheus_types.go
l646505418: Ok, get it!

l646505418:
> Thanks for the contribution! I added one comment. The PR also contains some alertmanager changes that I'm not sure are intended in this one.
> 
> > Hi team, the only way that I came up with to accomplish this feature is to modify the CRD, but because I am not very familiar with operator and I use the Windows, it is hard for me to generate the CRDs and some clientset files. Maybe you guys can review and genetra,thanks
> 
> You can use `make generate` to regenerate CRDs and other client files. See https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#contribution-flow

Actually I met this issue [https://github.com/kubernetes/code-generator/issues/88](), so it is hard for me to generate CRDs

PR #5750, opened 2023-07-19
pkg/apis: Add params to ScrapeConfig

### LINKED ISSUE #5592: Add params support to the ScrapConfig CRD (federation use case)
ScrapeConfig CRD does not support federation scrapes because it does not allow setting the "http params"

`prometheus-operator:v0.65.1`

We would like to be able to define scraping federation endpoints through the use of a ScrapeConfig CRD. In order to scrape the federation endpoints, we need the ability to include the "desired match[] parameters" or the federation endpoint will return nothing.
https://prometheus.io/docs/prometheus/latest/federation/#configuring-federation

Currently, the spec for ScrapeConfig does not allow specifying HTTP params of scrape configs:
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config
```
# Optional HTTP URL parameters.
params:
  [ [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string): [[<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string), ...] ]
```

Per the Prometheus-Operator API Spec:
https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1alpha1.ScrapeConfigSpec

slashpai:
Fixes #5592

## Description

## Type of change

## Changelog entry

```release-note
Add http params to ScrapeConfig
```

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
+++ b/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
@@ -102,6 +102,10 @@ type ScrapeConfigSpec struct {
 	// HonorLabels chooses the metric's labels on collisions with target labels.
 	// +optional
 	HonorLabels *bool `json:"honorLabels,omitempty"`
+	// Optional HTTP URL parameters
+	// +optional
+	// +mapType:=atomic
+	Params map[string][]string `json:"params,omitempty"`
 	// Configures the protocol scheme used for requests.
 	// If empty, Prometheus uses HTTP by default.
 	// +kubebuilder:validation:Enum=HTTP;HTTPS
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -2222,6 +2222,10 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		cfg = append(cfg, yaml.MapItem{Key: "metrics_path", Value: *sc.Spec.MetricsPath})
 	}
 
+	if len(sc.Spec.Params) > 0 {
+		cfg = append(cfg, yaml.MapItem{Key: "params", Value: stringMapToMapSlice(sc.Spec.Params)})
+	}
+
 	if sc.Spec.RelabelConfigs != nil {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
 		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -15679,6 +15679,18 @@ bool
 </tr>
 <tr>
 <td>
+<code>params</code><br/>
+<em>
+map[string][]string
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>Optional HTTP URL parameters</p>
+</td>
+</tr>
+<tr>
+<td>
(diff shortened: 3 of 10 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -2179,6 +2179,10 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		cfg = append(cfg, yaml.MapItem{Key: "metrics_path", Value: *sc.Spec.MetricsPath})
 	}
 
+	if sc.Spec.Params != nil {
+		cfg = append(cfg, yaml.MapItem{Key: "params", Value: sc.Spec.Params})
```
nicolastakashi, line 2183: Should we sort the params argument by the key to ensure we will always render the config on the same structure?
I did the same [here](https://github.com/prometheus-operator/prometheus-operator/pull/5709#discussion_r1252733778).

nicolastakashi (changes requested):

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -98,6 +98,9 @@ type ScrapeConfigSpec struct {
 	// HonorLabels chooses the metric's labels on collisions with target labels.
 	// +optional
 	HonorLabels *bool `json:"honorLabels,omitempty"`
+	// Optional HTTP URL parameters
+	// +optional
+	Params *map[string][]string `json:"params,omitempty"`
```
simonpasquier, line 103: we don't need a pointer for optional maps and slices.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

```suggestion
	Params map[string][]string `json:"params,omitempty"`
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -98,6 +98,9 @@ type ScrapeConfigSpec struct {
 	// HonorLabels chooses the metric's labels on collisions with target labels.
 	// +optional
 	HonorLabels *bool `json:"honorLabels,omitempty"`
+	// Optional HTTP URL parameters
+	// +optional
```
simonpasquier, line 102:
```suggestion
	// +optional
	// +mapType:=atomic
```

pkg/prometheus/promcfg.go
slashpai: IIUC by default Go prints the map with sorted keys but while iterating over a map, it follows the order of the keys appearing as it is (an example here https://go.dev/play/p/YeF128GxX2e). Since we are just passing the value here do we need to sort here?

simonpasquier: @slashpai you should be able to leverage `stringMapToMapSlice()` after https://github.com/prometheus-operator/prometheus-operator/pull/5770 merges.

pkg/prometheus/promcfg.go
```diff
@@ -2179,6 +2179,10 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		cfg = append(cfg, yaml.MapItem{Key: "metrics_path", Value: *sc.Spec.MetricsPath})
 	}
 
+	if sc.Spec.Params != nil {
```
simonpasquier, line 2182: (nit)
```suggestion
	if len(sc.Spec.Params) > 0 {
```

slashpai:
@simonpasquier done

simonpasquier (approved):

Issue #5713, opened 2023-06-21
Endpoint.BearerTokenSecret must be pointer


setood:
Endpoint.BearerTokenSecret must be pointer:
https://github.com/prometheus-operator/prometheus-operator/blob/c840b7eaa857fc45f8a9be52cfd8d315f7c88e9f/pkg/apis/monitoring/v1/types.go#L396-L399

like in other places:
https://github.com/prometheus-operator/prometheus-operator/blob/c840b7eaa857fc45f8a9be52cfd8d315f7c88e9f/pkg/apis/monitoring/v1beta1/alertmanager_config_types.go#L556

I ran into this problem when I tried to create a servicemonitor with [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder), a BearerTokenSecret added in it and that cannot be removed.

my code is like that:
```go
import monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"

var monitor *monitoringv1.ServiceMonitor
k8sClient.Create(ctx, monitor);
```

heliapb:
Hi @ArthurSens I would like to work on this one, could you please assign to me

slashpai:
@heliapb Go ahead :)

slashpai:
Completed in https://github.com/prometheus-operator/prometheus-operator/pull/5752

PR #5752, opened 2023-07-20
fix minor issue with endpoint bearertoken

heliapb:
## Description

Fixes issue https://github.com/prometheus-operator/prometheus-operator/issues/5713

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/assets/store.go
+++ b/pkg/assets/store.go
@@ -204,12 +204,16 @@ func (s *Store) AddOAuth2(ctx context.Context, ns string, oauth2 *monitoringv1.O
 }
 
 // AddToken processes the given SecretKeySelector and adds the referenced data to the store.
-func (s *Store) addToken(ctx context.Context, ns string, sel v1.SecretKeySelector, key string) error {
+func (s *Store) addToken(ctx context.Context, ns string, sel *v1.SecretKeySelector, key string) error {
+	if sel == nil {
+		return nil
+	}
+
 	if sel.Name == "" {
 		return nil
 	}
 
-	token, err := s.GetSecretKey(ctx, ns, sel)
+	token, err := s.GetSecretKey(ctx, ns, *sel)
 	if err != nil {
 		return errors.Wrap(err, "failed to get token from secret")
 	}
@@ -219,11 +223,7 @@ func (s *Store) addToken(ctx context.Context, ns string, sel v1.SecretKeySelecto
 	return nil
 }
 
-func (s *Store) AddBearerToken(ctx context.Context, ns string, sel v1.SecretKeySelector, key string) error {
-	if sel == (v1.SecretKeySelector{}) {
-		return nil
-	}
-
+func (s *Store) AddBearerToken(ctx context.Context, ns string, sel *v1.SecretKeySelector, key string) error {
 	err := s.addToken(ctx, ns, sel, key)
 	if err != nil {
 		return errors.Wrap(err, "failed to get bearer token")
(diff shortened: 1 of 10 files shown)


slashpai:
@heliapb You need to regenerate after updating type and 
`make generate` command

ArthurSens (commented):
I'm struggling to understand why the e2e-tests failed :( 

Locally I'm getting rate limit errors, not sure if that's related

pkg/assets/store.go
```diff
@@ -219,8 +219,8 @@ func (s *Store) addToken(ctx context.Context, ns string, sel v1.SecretKeySelecto
 	return nil
 }
 
-func (s *Store) AddBearerToken(ctx context.Context, ns string, sel v1.SecretKeySelector, key string) error {
-	err := s.addToken(ctx, ns, sel, key)
+func (s *Store) AddBearerToken(ctx context.Context, ns string, sel *v1.SecretKeySelector, key string) error {
+	err := s.addToken(ctx, ns, *sel, key)
```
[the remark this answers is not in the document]
heliapb: Hi there, thanks for the feedback, changed as suggested. If you might have any other idea about the e2e failing would be much appreciated, as it is my first attempt at contributing to prometheus operator and I'm still a bit lost.

pkg/prometheus/resource_selector.go
```diff
@@ -125,7 +125,7 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
 
 			smKey := fmt.Sprintf("serviceMonitor/%s/%s/%d", sm.GetNamespace(), sm.GetName(), i)
 
-			if err = rs.store.AddBearerToken(ctx, sm.GetNamespace(), endpoint.BearerTokenSecret, smKey); err != nil {
+			if err = rs.store.AddBearerToken(ctx, sm.GetNamespace(), *endpoint.BearerTokenSecret, smKey); err != nil {
```
simonpasquier, line 128: I'd rather update the `AddBearerToken()` method to check for nil parameter and return early.

heliapb: Hi @simonpasquier changed, if you could review!

pkg/assets/store.go
```diff
@@ -220,6 +220,10 @@ func (s *Store) addToken(ctx context.Context, ns string, sel v1.SecretKeySelecto
 }
 
 func (s *Store) AddBearerToken(ctx context.Context, ns string, sel v1.SecretKeySelector, key string) error {
+	if sel == (v1.SecretKeySelector{}) {
+		return nil
+	}
+
```
simonpasquier, line 226: Sorry for being unclear, I meant to change the signature of AddBearerToken() like this:
```suggestion
func (s *Store) AddBearerToken(ctx context.Context, ns string, sel *v1.SecretKeySelector, key string) error {
	if sel == nil {
		return nil
	}

```

heliapb: Hi if I set it that way I get an error `invalid operation: sel == nil (mismatched types "k8s.io/api/core/v1".SecretKeySelector and untyped nil)compiler[MismatchedTypes](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#MismatchedTypes)`

simonpasquier: hmm this commit should work https://github.com/simonpasquier/prometheus-operator/commit/a94c876861773f182f5fd3845ede028d9dc40ffc

heliapb: Well in that change you are chaning the func addToken instead of AddBearerToken. So could you please clarify?

simonpasquier: yes I suggest to change both function signatures.

heliapb: Hi @simonpasquier done, when you have the time pls review

ArthurSens (approved):
LGTM, thanks for the effort!

Issue #5578, opened 2023-05-15
Unable to mount multiple volumes using prometheus statefulset


rajani-konreddy:
**What happened?**

When we are trying to create separate volume for data and backup using Prometheus statefulset.

Prometheus stateful set spec:

What did you expect to see?
Should be able to mount multiple volumes with Statefulset

What did you see instead? Under which circumstances?
It is throwing below an error.

Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(Prometheus.spec.storage): unknown field "volumeClaimTemplates" in com.coreos.monitoring.v1.Prometheus.spec.storage

System information
No response

**Did you expect to see something different?**

**How to reproduce it (as minimally and precisely as possible)**:

volumeMounts:
    - name: prometheus-data
      mountPath: /data
      subPath: prometheus-db
    - name: prometheus-backup
      mountPath: /prometheus
    command:
    - "/opt/myaddition.sh"
  storage:
    volumeClaimTemplates:
    - metadata:
        name: prometheus-data
      spec:
        resources:
          requests:
            storage: {{ .Values.Prometheus.cluster_1_deploy.spec.data.storage }}
        volumeMode: Filesystem
        storageClassName: {{ .Values.Prometheus.cluster_1_deploy.spec.data.storageClassName }}
   - metadata:
        name: prometheus-backup
      spec:
        resources:
          requests:
            storage: {{ .Values.Prometheus.cluster_1_deploy.spec.backup.storage }}
        volumeMode: Filesystem
        storageClassName: {{ .Values.Prometheus.cluster_1_deploy.spec.backup.storageClassName }}

**Environment**

* Prometheus Operator version: v0.64.0

    `Insert image tag or Git SHA here`
    
    

* Kubernetes version information:

    `kubectl version` : 
    
kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:31:21Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #5760, opened 2023-07-20
support mount multiple volumes

l646505418:
## Description

Add voluimeClaimTemplates to Storage.Spec to support mount multiple volumes

#5578 

## Type of change

- [`x` ] `FEATURE` (non-breaking change which adds functionality)


nicolastakashi:
Amazing work here @l646505418!
I'm just wondering if we should have some deprecation process on the `VolumeClaimTemplate`

l646505418:
> 

As @xiu commented before, `Since Prometheus is a v1 CRD, we can't remove a field without bumping the version (e.g. v2). In this case, we'd have to add a new field but not remove the previous one for compatibility purposes in order to stick to v1.`

xiu:
👍 we can update the comment to direct users to use the new field without removing the old one though.

l646505418:
> 👍 we can update the comment to direct users to use the new field without removing the old one though.

Yes, but as I mentioned before, I met some diffculties to run `make generate`, it is hard for me to generate CRDs and other files(eg. zz_generated)

simonpasquier:
For the record, the k8s developer documentation has some recommendations about making a singular field plural:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#making-a-singular-field-plural

I'm not sure how it works in practice for CRDs (vs. k8s native resources).

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -838,6 +838,8 @@ type StorageSpec struct {
 	// The easiest way to use a volume that cannot be automatically provisioned
```
xiu, line 838: This comment should be led by something along the lines of:
```*DEPRECATED*: Please use `VolumeClaimTemplates` which allows for multiple VolumeClaim to be used.```

File: pkg/apis/monitoring/v1/prometheus_types.go [hunk 2]
```diff
@@ -838,6 +838,8 @@ type StorageSpec struct {
...
 	// is to use a label selector alongside manually created PersistentVolumes.
 	VolumeClaimTemplate EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate,omitempty"`
+
```
xiu, line 841: Add a comment explaining what this field does, probably same as the previous one (without the deprecation notice) but plural.

pkg/prometheus/agent/statefulset.go
```diff
@@ -137,6 +137,22 @@ func makeStatefulSet(
 				Ephemeral: ephemeral,
 			},
 		})
+	} else if storageSpec.VolumeClaimTemplates != nil {
+		for _, specVcTemplate := range storageSpec.VolumeClaimTemplates {
```
xiu, line 141: I didn't test it but we might be able to simplify this part by removing the else block on L140 and supporting both in this one. (same comment for the server)

```suggestion
	} else {
	    if storageSpec.VolumeClaimTemplate != nil {
	        storageSpec.VolumeClaimTemplates = append(storageSpec.VolumeClaimTemplates, storageSpec.VolumeClaimTemplate)
	    }

	    for _, specVcTemplate := range storageSpec.VolumeClaimTemplates {
```

What do you think?

xiu (commented):
Thanks for your work on this, @l646505418! Had another look at this one, I added comments where I think we should update that comment.

> Yes, but as I mentioned before, I met some diffculties to run make generate, it is hard for me to generate CRDs and other files(eg. zz_generated)

We'll help with that if we can't find another solution no worries. 

I see we have 3 unit tests failing at the moment:
* `TestStatefulSetVolumeInitial`
* `TestThanosObjectStorage`
* `TestThanosObjectStorageFile`

Once those tests are fixed and comments addressed, we'll make sure to generate assets to see the pipeline go green before merging 👍

l646505418:
> Thanks for your work on this, @l646505418! Had another look at this one, I added comments where I think we should update that comment.
> 
> > Yes, but as I mentioned before, I met some diffculties to run make generate, it is hard for me to generate CRDs and other files(eg. zz_generated)
> 
> We'll help with that if we can't find another solution no worries.
> 
> I see we have 3 unit tests failing at the moment:
> 
> * `TestStatefulSetVolumeInitial`
> * `TestThanosObjectStorage`
> * `TestThanosObjectStorageFile`
> 
> Once those tests are fixed and comments addressed, we'll make sure to generate assets to see the pipeline go green before merging 👍

Hi，I have updated codes, can you guys review again?

pkg/prometheus/server/statefulset.go
```diff
@@ -201,18 +202,25 @@ func makeStatefulSet(
 			},
 		})
 	} else {
-		pvcTemplate := operator.MakeVolumeClaimTemplate(storageSpec.VolumeClaimTemplate)
-		if pvcTemplate.Name == "" {
-			pvcTemplate.Name = prompkg.VolumeName(p)
+		if reflect.DeepEqual(storageSpec.VolumeClaimTemplate, monitoringv1.EmbeddedPersistentVolumeClaim{}) {
+			storageSpec.VolumeClaimTemplates = append([]monitoringv1.EmbeddedPersistentVolumeClaim{storageSpec.VolumeClaimTemplate}, storageSpec.VolumeClaimTemplates...)
```
ArthurSens, line 206: As far as I could understand, you're comparing if VolumeClaimTemplate is equal com an empty object? And only if it is indeed empty, then you're appending it to array of PVClaims?

Seen in [godocs](https://pkg.go.dev/reflect#DeepEqual) of DeeplyEqual:
`Struct values are deeply equal if their corresponding fields, both exported and unexported, are deeply equal.`

```suggestion
		if !reflect.DeepEqual(storageSpec.VolumeClaimTemplate, monitoringv1.EmbeddedPersistentVolumeClaim{}) {
			storageSpec.VolumeClaimTemplates = append([]monitoringv1.EmbeddedPersistentVolumeClaim{storageSpec.VolumeClaimTemplate}, storageSpec.VolumeClaimTemplates...)
```
Looks like we're only missing the negation operator here :P

ArthurSens (commented):
Hey @l646505418 , thanks for working on this. It's looking like we're in a good direction, just need to fix a few things that were caught in out tests.

Ideally, you'd be able to run those tests yourself by running `make test-unit` or `make test-e2e` and there's no need to depend on CI to catch them during development. Most editors also provide a good debugging experience when working with go projects :)

Let us know if you're also having problems running tests!

l646505418:
> 

Oh， your change is right, it is my fault😂
I indeed can run unit test in my local, but I have mentioned before I have some troubles generating CRDs , I am not sure if that will have big impacts.

xiu:
@l646505418 are you able to run `make test-unit` in your branch? We still see 3 failing tests, they shouldn't be related to `make generate`.

l646505418:
> @l646505418 are you able to run `make test-unit` in your branch? We still see 3 failing tests, they shouldn't be related to `make generate`.

Hi ,I modified codes and all unit tests are passed, can you review?

pkg/prometheus/agent/statefulset.go
```diff
@@ -138,18 +139,24 @@ func makeStatefulSet(
...
-		if pvcTemplate.Name == "" {
-			pvcTemplate.Name = prompkg.VolumeName(p)
+		if !reflect.DeepEqual(storageSpec.VolumeClaimTemplate, monitoringv1.EmbeddedPersistentVolumeClaim{}) {
+			storageSpec.VolumeClaimTemplates = append([]monitoringv1.EmbeddedPersistentVolumeClaim{storageSpec.VolumeClaimTemplate}, storageSpec.VolumeClaimTemplates...)
 		}
-		if storageSpec.VolumeClaimTemplate.Spec.AccessModes == nil {
-			pvcTemplate.Spec.AccessModes = []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce}
-		} else {
-			pvcTemplate.Spec.AccessModes = storageSpec.VolumeClaimTemplate.Spec.AccessModes
+		for _, specVcTemplate := range storageSpec.VolumeClaimTemplates {
```
ArthurSens, line 145:
```suggestion
		for _, vcTemplate := range storageSpec.VolumeClaimTemplates {
```

this is just a volumeClaimTemplate, feels confusing to have spec on the variable name 🤔

pkg/prometheus/server/statefulset_test.go
```diff
@@ -228,6 +228,69 @@ func TestStatefulSetPVC(t *testing.T) {
...
+				Storage: &monitoringv1.StorageSpec{
+					VolumeClaimTemplates: pvcTemplates,
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+
+	if !reflect.DeepEqual(2, len(sset.Spec.VolumeClaimTemplates)) {
+		t.Fatal("Error adding PVCs Spec to StatefulSetSpec")
```
ArthurSens, line 279:
```suggestion
		t.Fatalf("Expected %d PVCs, got %d", 2, len(sset.Spec.VolumeClaimTemplates))
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -228,6 +228,69 @@ func TestStatefulSetPVC(t *testing.T) {
...
+		ssetPvc := sset.Spec.VolumeClaimTemplates[0]
+		if !reflect.DeepEqual(*pvc.Spec.StorageClassName, *ssetPvc.Spec.StorageClassName) {
+			t.Fatal("Error adding PVCs Spec to StatefulSetSpec")
+		}
+
+		ssetPvcSecond := sset.Spec.VolumeClaimTemplates[1]
+		if !reflect.DeepEqual(*pvcSecond.Spec.StorageClassName, *ssetPvcSecond.Spec.StorageClassName) {
+			t.Fatal("Error adding PVCs Spec to StatefulSetSpec")
+		}
+	}
```
ArthurSens, line 290: We can make the code more readable by reducing the indentation level here. (removing unnecessary else statement)

```suggestion
	} 
	ssetPvc := sset.Spec.VolumeClaimTemplates[0]
	if !reflect.DeepEqual(*pvc.Spec.StorageClassName, *ssetPvc.Spec.StorageClassName) {
		t.Fatal("Error adding PVCs Spec to StatefulSetSpec")
	}

	ssetPvcSecond := sset.Spec.VolumeClaimTemplates[1]
	if !reflect.DeepEqual(*pvcSecond.Spec.StorageClassName, *ssetPvcSecond.Spec.StorageClassName) {
		t.Fatal("Error adding PVCs Spec to StatefulSetSpec")
	}
```

ArthurSens: Also, why are we comparing only StorageClassName? Shouldn't we compare the whole PVC?

pkg/prometheus/statefulset.go
```diff
@@ -366,14 +369,29 @@ func BuildCommonVolumes(p monitoringv1.PrometheusInterface, tlsAssetSecrets []st
 	return volumes, promVolumeMounts, nil
 }
 
-func VolumeClaimName(p monitoringv1.PrometheusInterface, cpf monitoringv1.CommonPrometheusFields) string {
-	volName := VolumeName(p)
+func VolumeClaimName(p monitoringv1.PrometheusInterface, cpf monitoringv1.CommonPrometheusFields) []string {
+	volNames := []string{}
 	if cpf.Storage != nil {
-		if cpf.Storage.VolumeClaimTemplate.Name != "" {
+		volName := VolumeName(p)
```
ArthurSens, line 375:
```suggestion
```
This value isn't used since we override it on line 378.

If you need to initialize the variable `volName`, you can declare it at the beginning of this function:

```go
func VolumeClaimName(p monitoringv1.PrometheusInterface, cpf monitoringv1.CommonPrometheusFields) []string {
	volNames := []string{}
	var volName string
.
.
.
```

pkg/prometheus/statefulset.go
```diff
@@ -366,14 +369,29 @@ func BuildCommonVolumes(p monitoringv1.PrometheusInterface, tlsAssetSecrets []st
...
+		if cpf.Storage.VolumeClaimTemplates != nil {
+			for _, pvcTemplate := range cpf.Storage.VolumeClaimTemplates {
+				volName = VolumeName(p)
+				if pvcTemplate.Name != "" {
+					volName = pvcTemplate.Name
+				}
+				volNames = append(volNames, volName)
+			}
+		} else if cpf.Storage.VolumeClaimTemplate.Name != "" {
```
ArthurSens, line 384: The only way we get here is if `cpf.Storage.VolumeClaimTemplates` is nil, right? If that's the case, trying to access `cpf.Storage.VolumeClaimTemplate.Name` will throw a nil pointer exception. 

We should do nothing if VolumeClaimTemplates is nil

pkg/prometheus/statefulset.go
```diff
@@ -366,14 +369,29 @@ func BuildCommonVolumes(p monitoringv1.PrometheusInterface, tlsAssetSecrets []st
...
 			volName = cpf.Storage.VolumeClaimTemplate.Name
+			volNames = append(volNames, volName)
+		} else {
+			volNames = append(volNames, volName)
 		}
+	} else {
+		volName := VolumeName(p)
+		volNames = append(volNames, volName)
```
ArthurSens, line 392: Same thing here. If `Storage` is nil, we don't want to append any volume.

pkg/prometheus/statefulset.go
l646505418: So how about when user only define `Storage.VolumeClaimTemplate`?

ArthurSens: Ah sorry, I misread the condition. I thought this line was comparing VolumeClaimTemplate<ins>s</ins>.Name 😅 

You're correct

ArthurSens (approved):
LGTM, thanks!

ArthurSens:
Good find!

I'm reverting the PR and let's discuss the new approach

PR #5577, opened 2023-05-14
Add ScrapeConfigs to ObjectReference

mcbenjemaa:
## Description

- [X] Move `resourceToKind` away from the v1 package.
- [X] Add `scrapeconfigs` as part of the `ObjectReference.Resource` enum.

I have duplicated the consts in `monitoring/resource.go` otherwise it will cause a circular-dependency . 

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [X] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Add ScrapeConfigs to ObjectReference
```

Closes https://github.com/prometheus-operator/prometheus-operator/issues/5544

PR #5772, opened 2023-07-26
chore: Deduplicate CRD constants

xiu:
## Description

This is a follow up from #5577 in which we duplicated consts to add ScrapeConfigs to ObjectReference.

cc @mcbenjemaa 

## Type of change

## Changelog entry

```release-note
Deprecates CRD constants and moves them to `api/monitoring`. This only impacts developers reusing our CRDs in their projects. The constants will be removed in v0.75.
```

### DIFF
--- a/pkg/apis/monitoring/v1/alertmanager_types.go
+++ b/pkg/apis/monitoring/v1/alertmanager_types.go
@@ -21,8 +21,14 @@ import (
 )
 
 const (
-	AlertmanagersKind   = "Alertmanager"
-	AlertmanagerName    = "alertmanagers"
+	// Deprecated: Use github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring
+	// TODO: Remove in v0.75
+	AlertmanagersKind = "Alertmanager"
+	// Deprecated: Use github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring
+	// TODO: Remove in v0.75
+	AlertmanagerName = "alertmanagers"
+	// Deprecated: Use github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring
+	// TODO: Remove in v0.75
 	AlertManagerKindKey = "alertmanager"
 )
 
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -348,11 +348,11 @@ func run() int {
 		kclient,
 		namespaces(cfg.Namespaces.AllowList).asSlice(),
 		monitoringv1alpha1.SchemeGroupVersion,
-		monitoringv1alpha1.ScrapeConfigName,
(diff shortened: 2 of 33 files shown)


ArthurSens:
I'm not sure if this is a safe change for downstream projects 😬.

I see a pattern amongst Kubernetes controllers to declare those types of constants in the same package of their API specs, and it is expected for controllers that consume APIs of other controllers to find those constants as they are declared now.


From a [quick code search](https://github.com/search?q=monitoringv1.PrometheusRuleKind&type=code), I can see several downstream projects consuming those constants. If we move them, it would be a breaking change for everyone 😬

nicolastakashi:
I tend to agree with @ArthurSens!

Although I think the change makes sense, we should plan this as a breaking change and make an gradual rollout of this change, like start this work internally on the operator and communicate to people and give them some time to update it.

simonpasquier:
I'm not feeling too strong on "breaking" other projects as we're still on v0.x. But to avoid being rude, we could maintain the deprecated constants for a couple of releases and add a deprecated comment that would be caught by Go linters.

xiu:
That sounds good to me, I'll update the PR with that.

xiu:
Note that I haven't forgotten about that one, still planning to execute on it 👍

xiu:
@simonpasquier @ArthurSens @nicolastakashi Can you take another look? Thanks!

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -21,8 +21,14 @@ import (
 )
 
 const (
-	AlertmanagersKind   = "Alertmanager"
-	AlertmanagerName    = "alertmanagers"
+	// Deprecated: Use github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring
+	// TODO: Remove in v0.75
```
ArthurSens, line 25:
```suggestion
	// TODO: Remove in v0.75
	// Deprecated: Use github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring instead of github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1. This constant will be removed in v0.75
```

I would just make the deprecation message more clear (see screenshot above)

ArthurSens (approved):
Nice, just tried replacing the module in some downstream projects and I could see the deprecation caught by linters :) 

image

ArthurSens:
@xiu have you had the chance to take a look at the comment above?

pkg/apis/monitoring/v1/alertmanager_types.go
ArthurSens: The main problem I see in the message is this "TODO: Remove in v0.75" is displayed downstream. If we change the order, the TODO shouldn't be displayed anymore

PR #5773, opened 2023-07-27
Feat: Ignore revisionHistoryLimit field in StatefulSet for Prometheus, Alertmanager& ThanosRuler

### LINKED ISSUE #5712: AlertManager Continuous reconcile loop when setting revision history limit
**What happened?**

We use ArgoCD with a fully gitops deployment model. Due to this deployment model we have no need for revisions. In fact we wish to not have them because they clutter the UI for no reason. Many charts and operators do not allow us to set the revision history limit, so we are creating an admission webhook to set the revision history limit for all deployments, daemonsets and statefulsets in our cluster.

However even though prometheus-operator does not provide any way to set the revision history limit (as far as I can find) it does try to enforce the revision history limit for the AlertManager statefulset (possibly other managed resources as well).

This causes a continuous reconciliation and recreation (because revision history limit on sts is immutable) of alertmanager on our environments.

The operator does not allow setting the revision history limit, so it should not care when it's default gets modified by a custom admission controller. It tries to enforce setting it to 10, and I see no reason for it to do that.

Create admission webhook that modifies the statefulset revision history limit before admission, deploy new AlertManager instance. Observe prometheus-operator continously destroying and recreating the statefulset.

**Environment**

    `prometheus-operator:v0.65.2`
    
    

 ```
clientVersion:
    buildDate: "2023-06-14T09:47:38Z"
    compiler: gc
    gitCommit: 25b4e43193bcda6c7328a6d147b1fb73a33f1598
    gitTreeState: clean
    gitVersion: v1.27.3
    goVersion: go1.20.5
    major: "1"
    minor: "27"
    platform: linux/amd64
kustomizeVersion: v5.0.1
serverVersion:
    buildDate: "2023-05-19T00:03:21Z"
    compiler: gc
    gitCommit: b278d3784e784ae3d8e986fe641beb572e08ecc0
    gitTreeState: clean
    gitVersion: v1.26.3
    goVersion: go1.19.7
    major: "1"
    minor: "26"
    platform: linux/amd64
```

* Kubernetes cluster kind:

  `AKS`

* Prometheus Operator Logs:

```
level=info ts=2023-06-21T12:58:47.2969237Z caller=operator.go:736 component=alertmanageroperator key=prometheus/prometheus-stack-kube-prom-alertmanager msg="recreating AlertManager StatefulSet because the update operation wasn't possible" reason="Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden"
```

### PR DESCRIPTION
**Pull request by Azanul**
## Description

Fixes #5712 

## Type of change

## Changelog entry

```release-note
Ignore updates revisionHistoryLimit field in the underlying StatefulSet allowing external services to manage revisionHistoryLimit
```


Azanul:
@rouke-broersma @simonpasquier

rouke-broersma:
My issue is with alertmanager, could you add the same for them?

simonpasquier (commented):
I'm not sure that we need to expose a `revisionHistoryLimit` field in the existing CRDs. IIUC the issue is that the Kubernetes API always returns a default value of 10 in the statefulset spec and if something/someone modifies the value then [createSSetInputHash()](https://github.com/prometheus-operator/prometheus-operator/blob/93b21612db3e6bfe34745295dfc85ee84ba377ac/pkg/prometheus/server/operator.go#L1414) generates a new hash value triggering a statefulset update.
I assume that if `createSSetInputHash` sets `RevisionHistoryLimit` to nil before generating the sts hash, it would solve the issue too.

Azanul:
> if createSSetInputHash sets RevisionHistoryLimit to nil before generating the sts hash, it would solve the issue too.

Purposefully ignoring any changes to `RevisionHistoryLimit`, I think that will work.

simonpasquier:
> @simonpasquier Why does the spec goes as an interface in [Thanos createSSetInputHash](https://github.com/prometheus-operator/prometheus-operator/blob/93b21612db3e6bfe34745295dfc85ee84ba377ac/pkg/thanos/operator.go#L706)?

No concrete reason I believe. https://github.com/prometheus-operator/prometheus-operator/commit/bd658c582c697e256b0c17b329b9d46fdc452eb9#diff-c8f4eb63862907c2cd9e067b2cf70e03f4e134075bd28f507dade553850d4fa3L1627 is where we changed it for Prometheus.
And I'd even recommend to apply the same change to Thanos Ruler as it should help reducing the number of sync loops.

rouke-broersma:
> I'm not sure that we need to expose a `revisionHistoryLimit` field in the existing CRDs. IIUC the issue is that the Kubernetes API always returns a default value of 10 in the statefulset spec and if something/someone modifies the value then [createSSetInputHash()](https://github.com/prometheus-operator/prometheus-operator/blob/93b21612db3e6bfe34745295dfc85ee84ba377ac/pkg/prometheus/server/operator.go#L1414) generates a new hash value triggering a statefulset update.
> I assume that if `createSSetInputHash` sets `RevisionHistoryLimit` to nil before generating the sts hash, it would solve the issue too.

I agree but note that it was your own comment saying that the operator needs to assume full control and that led to this solution instead of just ignoring the field 😅

https://github.com/prometheus-operator/prometheus-operator/issues/5712#issuecomment-1600884504

simonpasquier:
@rouke-broersma yes I sometimes contradicts myself ;)

Azanul:
> Shall I do it in the same PR?

@simonpasquier

simonpasquier (commented):
can you do the same change [here](https://github.com/prometheus-operator/prometheus-operator/blob/2ee89e451de5a27edee6ff0c084673974a1a362c/pkg/alertmanager/operator.go#L818)?

Azanul:
> @Azanul sorry for the late reply. Can you open another PR to update the Thanos Ruler's `createSSetInputHash()` method? I'd prefer to merge it first then you can rebase this PR...

check out https://github.com/prometheus-operator/prometheus-operator/pull/5784

pkg/alertmanager/operator.go
```diff
@@ -821,6 +821,9 @@ func createSSetInputHash(a monitoringv1.Alertmanager, c Config, tlsAssets *opera
 		http2 = a.Spec.Web.WebConfigFileFields.HTTPConfig.HTTP2
 	}
 
+	// Ignoring changes to RevisionHistoryLimit field
```
simonpasquier, line 824: (suggestion) provide a bit more context.
```suggestion
	// The controller should ignore any changes to RevisionHistoryLimit field because 
	// it may be modified by external actors.
	// See https://github.com/prometheus-operator/prometheus-operator/issues/5712
```

pkg/prometheus/agent/operator.go
```diff
@@ -759,6 +759,9 @@ func createSSetInputHash(p monitoringv1alpha1.PrometheusAgent, c operator.Config
 		http2 = p.Spec.Web.WebConfigFileFields.HTTPConfig.HTTP2
 	}
 
+	// Ignoring changes to RevisionHistoryLimit field
```
simonpasquier, line 762: Sorry for being dense in my previous comment: can you update the comment here too?

pkg/prometheus/server/operator.go
```diff
@@ -1417,6 +1417,9 @@ func createSSetInputHash(p monitoringv1.Prometheus, c operator.Config, ruleConfi
 		http2 = p.Spec.Web.WebConfigFileFields.HTTPConfig.HTTP2
 	}
 
+	// Ignoring changes to RevisionHistoryLimit field
```
simonpasquier, line 1420: here too

pkg/thanos/operator.go
```diff
@@ -704,6 +704,10 @@ func (o *Operator) UpdateStatus(ctx context.Context, key string) error {
 }
 
 func createSSetInputHash(tr monitoringv1.ThanosRuler, c Config, ruleConfigMapNames []string, ss appsv1.StatefulSetSpec) (string, error) {
+
+	// Ignoring changes to RevisionHistoryLimit field
```
simonpasquier, line 708: here too

pkg/prometheus/agent/operator.go
Azanul: No worries

PR #5776, opened 2023-07-27
chore: adding golden files support on webconfig tests

nicolastakashi:
## Description

I'm adding golden file supports on `webconfig` package, solves #5741

## Type of change

## Changelog entry

```release-note
adding golden file supports on `webconfig` package
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,7 @@ This is a rough outline of what a contributor's workflow looks like:
 - Make sure your commit messages are in the proper format (see below).
 - Push your changes to a topic branch in your fork of the repository.
 - Make sure the tests pass, and add any new tests as appropriate.
+- If the tests are checking long strings such as yaml, json or any other complex content, ensure you're using golden files.
 - Submit a pull request to the original repository.
 
 Many files (documentation, manifests, ...) in this repository are auto-generated. For instance, `bundle.yaml` is generated from the *Jsonnet* files in `/jsonnet/prometheus-operator`. Before submitting a pull request, make sure that you've executed `make generate` and committed the generated changes.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ This is a rough outline of what a contributor's workflow looks like:
 - Make sure your commit messages are in the proper format (see below).
 - Push your changes to a topic branch in your fork of the repository.
 - Make sure the tests pass, and add any new tests as appropriate.
-- If the tests are checking long strings such as yaml, json or any other complex content, ensure you're using golden files.
+- If the tests are checking long strings such as YAML, JSON or any other complex content, ensure you're using [golden files](https://pkg.go.dev/gotest.tools/v3/golden).
 - Submit a pull request to the original repository.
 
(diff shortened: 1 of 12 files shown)


ArthurSens:
Amazing to see progress here!

One thing that I feel is missing is a Makefile step to update golden files.

 Looking at the [documentation of the library](https://pkg.go.dev/gotest.tools/v3/golden) you've used, if we run `go test -update`, the golden files will be updated automatically. However, running `go test -update` in a package that doesn't have `gotest.tools/v3/golden` fails saying that the flag -update is not expected. 

Let's think of a way of update the golden files without failing everything else 😅

ArthurSens:
I could implement a small shell to get the information:

```bash
#!/bin/bash

dependency="gotest.tools/v3/golden"

# List all packages in the project
packages=$(go list ./...)

# Loop through each package and check if it imports the specific dependency
for pkg in $packages; do
    # Use 'go list' with 'XTestImports' template to get the imports from test binaries
    imports=$(go list -f '{{join .XTestImports "\n"}}' $pkg)
    
    # Check if the dependency is in the imports
    if echo "$imports" | grep -q "$dependency"; then
        # If the dependency is found, print the package path
        echo "$pkg"
    fi
done
```

xiu:
Looks good to me :) Should we add a mention of this all to CONTRIBUTING.md? To guide people into using golden rather than the old way.

nicolastakashi:
Tks @ArthurSens for the shell script, very useful 😄

nicolastakashi:
@xiu I added a entry to the contributing.md but I'm not sure if the information I put is enough?

ArthurSens (approved):
I'm just not sure about the CONTRIBUTING.md entry, I think we need to have a more elaborated section about testing (and that would include explaining the new Makefile step added in this PR)

I often see PRs with simple mistakes that could easily be caught by running `make test-unit` locally. In another PR, we could probably improve testing guidelines, including documentation for running unit tests, how to update golden files, when to use golden files and even a step-by-step guide for our more complicated e2e tests, how to run all, how to run just a subset of tests, etc.

But of course, out of scope for this PR!

scripts/update-golden-files.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
```
simonpasquier, line 2: A short explanation of what the script is meant for would be useful.

CONTRIBUTING.md
```diff
@@ -61,6 +61,7 @@ This is a rough outline of what a contributor's workflow looks like:
 - Make sure your commit messages are in the proper format (see below).
 - Push your changes to a topic branch in your fork of the repository.
 - Make sure the tests pass, and add any new tests as appropriate.
+- If the tests are checking long strings such as yaml, json or any other complex content, ensure you're using golden files.
```
simonpasquier, line 64:
```suggestion
- If the tests are checking long strings such as YAML, JSON or any other complex content, ensure you're using [golden files](https://pkg.go.dev/gotest.tools/v3/golden).
```

simonpasquier (commented):
lgtm, just a few nits!

scripts/update-golden-files.sh
nicolastakashi: Can you check again @simonpasquier wdyt?

simonpasquier (approved):
Thanks!

ArthurSens:
Ah, @nicolastakashi, I just changed the PR description since this change is not relevant for end-users :)

PR #5780, opened 2023-07-31
feat: runAsNonRoot true as default for config-reloader initcontainer

yanehi:
## Description
Besides the securityContext options `allowPrivilegeEscalation: false` and `readOnlyRootFilesystem: true` I also want to make sure that the config-reloader InitContainer runs with `runAsNonRoot: true`.

Unfortunately I haven't found a way to set this via the values of the prometheusOperator. I guess this happens fix at this point in the code (https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/operator/config_reloader.go#L237):

pkg/operator/config_reloader.go
```go
SecurityContext: &v1.SecurityContext{
	AllowPrivilegeEscalation: &boolFalse,
	ReadOnlyRootFilesystem:   &boolTrue,
	Capabilities: &v1.Capabilities{
		Drop: []v1.Capability{"ALL"},
	},
```
## Type of change

## Changelog entry

```release-note
feat: set runAsNonRoot true as default for config-reloader initcontainer
```

### DIFF
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -237,6 +237,7 @@ func CreateConfigReloader(name string, options ...ReloaderOption) v1.Container {
 		SecurityContext: &v1.SecurityContext{
 			AllowPrivilegeEscalation: &boolFalse,
 			ReadOnlyRootFilesystem:   &boolTrue,
+			RunAsNonRoot:             &boolTrue,
 			Capabilities: &v1.Capabilities{
 				Drop: []v1.Capability{"ALL"},
 			},


simonpasquier:
> Unfortunately I haven't found a way to set this via the values of the prometheusOperator.

You can use `spec.containers` to `[patch](https://prometheus-operator.dev/docs/operator/strategic-merge-patch/)` the pod template generated by the operator. It looks like the CI isn't happy with the change though.

yanehi:
> > Unfortunately I haven't found a way to set this via the values of the prometheusOperator.
> 
> You can use `spec.containers` to `[patch](https://prometheus-operator.dev/docs/operator/strategic-merge-patch/)` the pod template generated by the operator. It looks like the CI isn't happy with the change though.

This fixed my problem, thank you very much!

PR #5781, opened 2023-07-31
refactor: Remove operator override on alertmanagerConfig top level co…

### LINKED ISSUE #5019: Duplicated notifications because of the top-level route enforcement
**What did you do?**
I have created AlertmanagerConfig CRD and set top-level route property "continue: false". CRD was correctly picked up by Alertmanager and placed on top of routes but with property "continue: true".

I expected that property won't be enforced. That is causing us problem because after notification is sent (route from CRD) it continues to be evaluated and notifcation is sent because of the default route (route from global Alertmanager). All of ours notifications are now duplicated.
How can I have property "continue: false" or the top level route because this brings much noise in our alerting system.

**Environment**

    Image:      quay.io/prometheus-operator/prometheus-operator:v0.48.1
    
    

**Client Version:** version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T18:03:20Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"darwin/arm64"}

**Server Version:** version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.14-eks-18ef993", GitCommit:"ac73613dfd25370c18cbbbc6bfc65449397b35c7", GitTreeState:"clean", BuildDate:"2022-07-06T18:06:50Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}

iblackman:
…ntinue

Allow users to decide if they want the alert to still be evaluated by the following routes or not. Having the override blocks the possible usage where you have a catch-all route generating duplicate alerts since the route added by an `alertmanagerConfig` already filters to its namespace.

Fixes #5019

## Description

As of now the `prometheus-operator` enforces `continue: true` on the routes generated by `alertmanagerConfig` and by doing so the users don't have any kind of control if the alerts should be evaluated by following roles or not, they will always be evaluated the following. The `prometheus-operator` should allow the users to decide if they want to continue or not, it shouldn't strip this power from the users.

## Your Proposal Title

* **Owners:**
  * @iblackman

* **Related Tickets:**
  * https://github.com/prometheus-operator/prometheus-operator/issues/5019

> TL;DR: This change removes the override done by the `prometheus-operator` on the `continue` value on top-level routes generated by `alertmanagerConfig` empowering the users to decide if they want to continue or not the evaluation of the alerts for the following routes.

## Why

Currently, once you define an `alertmanagerConfig` for example:
```yaml
kind: AlertmanagerConfig
metadata:
  name: example
  namespace: example
spec:
  route:
    continue: false
    groupBy:
    - alertname
    - container
    - pod
    - statefulset
    - replicaset
    - daemonset
    - deployment
    - namespace
    - targetName
    - persistentvolumeclaim
    groupInterval: 10m
    groupWait: 1m
    receiver: pagerduty-example
    repeatInterval: 30m
```

It will result in the following `Alertmanager` configuration:
```yaml
global:
  ...
route:
  ...
  routes:
  - receiver: example/example/pagerduty-example
    group_by:
    - alertname
    - container
    - pod
    - statefulset
    - replicaset
    - daemonset
    - deployment
    - namespace
    - targetName
    - persistentvolumeclaim
    matchers:
    - namespace="example"
    continue: true
    group_wait: 1m
    group_interval: 10m
    repeat_interval: 30m
inhibit_rules:
...
```

You can notice that even though the value for `continue` in the `alertmanagerConfig/example` was set to `false`, its value in the route generated in `Alertmanager` configuration was set to `true`. Overriding what was passed by the user.

### Pitfalls of the current solution

With the current behavior, it is not possible to set `continue: false` in the `alertmanagerConfig` top-level route. Because the operator always overrides that value to `true`.

## Goals

Goals and use cases for the solution as proposed in [How](#how):

* Allow easy collaboration and decision making on design ideas.
* Have a consistent design style that is readable and understandable.
* Have a design style that is concise and covers all the essential information.

## How

Stop with the override done by the operator on the top-level route of an `alertmanagerConfig` respecting the value passed for the attribute `continue`. By removing `r.Continue = true` from `processRoute`.

* How will you test and verify?
  * Tested in a local environment and it worked as expected, taking the value `continue: false` into consideration.
* How will you migrate users, without downtime? How do we solve incompatibilities?
  * I have not found any incompatibilities.
* What open questions are left? 
    * It is possible that people are currently setting the value to `false` but since the operator is overriding it they are not aware of it, and once this change is merged their value will be taken into consideration and it may change the route evaluation for them. 

## Alternatives

If the proposed change isn't acceptable we could create a global level variable to control this behavior of the operator to allow users if they desire to stop the override done by the `prometheus-operator`, where the default value could be to override but at least it would allow people to change it.

## Action Plan

By merging this PR it will already remove the override.

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -19370,8 +19370,7 @@ bool
 <td>
 <em>(Optional)</em>
 <p>Boolean indicating whether an alert should continue matching subsequent
-sibling nodes. It will always be overridden to true for the first-level
-route by the Prometheus operator.</p>
+sibling nodes.</p>
 </td>
 </tr>
 <tr>
@@ -23068,8 +23067,7 @@ bool
 <td>
 <em>(Optional)</em>
 <p>Boolean indicating whether an alert should continue matching subsequent
-sibling nodes. It will always be overridden to true for the first-level
-route by the Prometheus operator.</p>
+sibling nodes.</p>
 </td>
 </tr>
 <tr>
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -4796,8 +4796,7 @@ spec:
                     type: array
                   continue:
                     description: Boolean indicating whether an alert should continue
-                      matching subsequent sibling nodes. It will always be overridden
-                      to true for the first-level route by the Prometheus operator.
+                      matching subsequent sibling nodes.
                     type: boolean
                   groupBy:
                     description: List of labels to group by. Labels must not be repeated
--- a/example/prometheus-operator-crd-full/monitoring.coreos.com_alertmanagerconfigs.yaml
+++ b/example/prometheus-operator-crd-full/monitoring.coreos.com_alertmanagerconfigs.yaml
@@ -4819,8 +4819,7 @@ spec:
                     type: array
                   continue:
                     description: Boolean indicating whether an alert should continue
-                      matching subsequent sibling nodes. It will always be overridden
-                      to true for the first-level route by the Prometheus operator.
+                      matching subsequent sibling nodes.
                     type: boolean
                   groupBy:
                     description: List of labels to group by. Labels must not be repeated
@@ -9533,8 +9532,7 @@ spec:
                     type: array
                   continue:
                     description: Boolean indicating whether an alert should continue
-                      matching subsequent sibling nodes. It will always be overridden
-                      to true for the first-level route by the Prometheus operator.
+                      matching subsequent sibling nodes.
                     type: boolean
                   groupBy:
                     description: List of labels to group by. Labels must not be repeated
--- a/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
+++ b/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
@@ -4796,8 +4796,7 @@ spec:
                     type: array
                   continue:
                     description: Boolean indicating whether an alert should continue
-                      matching subsequent sibling nodes. It will always be overridden
-                      to true for the first-level route by the Prometheus operator.
+                      matching subsequent sibling nodes.
                     type: boolean
                   groupBy:
                     description: List of labels to group by. Labels must not be repeated
--- a/jsonnet/prometheus-operator/alertmanagerconfigs-crd.json
+++ b/jsonnet/prometheus-operator/alertmanagerconfigs-crd.json
@@ -5010,7 +5010,7 @@
                         "type": "array"
                       },
                       "continue": {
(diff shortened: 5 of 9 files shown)


simonpasquier:
The `continue: true` stanza is on purpose. You may have 2 AlertmanagerConfig objects in the same namespace and you want the alerts to be evaluated by both.

ArthurSens:
I agree that completely removing the option to set `continue: false` from users is not ideal and would love to see a more flexible approach :), at the same time simon's concern is completely valid!

> What do you think about having continue: true as default for the top-level route (I just saw that the default on alertmanager config itself is false, [ continue: | default = false ]https://prometheus.io/docs/alerting/latest/configuration/)? And if the continue is set to false on the definition of alertmanagerConfig resource it would accept the value false

Sounds like a good approach, but let's make sure to extend our documentation explaining that if you have multiple AlertmanagerConfig resources being selected by an alertmanager, having `continue: false` will cause routes not to be evaluated (I think we can't guarantee the order which they will be merged)

iblackman:
> Sounds like a good approach, but let's make sure to extend our documentation explaining that if you have multiple AlertmanagerConfig resources being selected by an alertmanager, having `continue: false` will cause routes not to be evaluated (I think we can't guarantee the order which they will be merged)

I totally agree, if there is any change it needs to be reflected in the docs. But I'm having issues making this new approach (default false), I'm not proficient with Go, no matter what I try here I can't seem to get this `default false on top level routes` to work.

simonpasquier:
One option could be that the last AlertmanagerConfig for a given namespace has `continue: true` instead of `continue: false`. 

```
route:
  receiver: "null"
  routes:
  # AlertmanagerConfig amc1 in namespace ns1
  - receiver: ns1/amc1/test1
    group_by:
    - job
    matchers:
    - namespace="ns1"
    continue: true # alerts should continue to the next route because it's part of the same namespace.
  # AlertmanagerConfig amc2 in namespace ns1
  - receiver: ns1/amc2/test2
    group_by:
    - instance
    matchers:
    - namespace="ns1"
    continue: false # alerts for namespace ns1 are routed either by this route or the previous one 
  # AlertmanagerConfig amc1 in namespace ns2
  - receiver: ns2/amc1/test3
    group_by:
    - job
    - instance
    matchers:
    - namespace="ns2"
    continue: false # alerts for namespace ns2 are routed by this route
```

aslafy-z:
Any news on this topic? Thank you

PR #5784, opened 2023-08-01
reduce the number of reconciliations

Azanul:
## Description

Reflecting the same changes as https://github.com/prometheus-operator/prometheus-operator/commit/bd658c582c697e256b0c17b329b9d46fdc452eb9#diff-c8f4eb63862907c2cd9e067b2cf70e03f4e134075bd28f507dade553850d4fa3L1627 in ThanosRuler CRD

## Type of change

## Changelog entry

```release-note
reduce the number of reconciliations on ThanosRuler objects.
```

### DIFF
--- a/pkg/thanos/operator.go
+++ b/pkg/thanos/operator.go
@@ -708,7 +708,6 @@ func createSSetInputHash(tr monitoringv1.ThanosRuler, c Config, ruleConfigMapNam
 		ThanosRulerLabels      map[string]string
 		ThanosRulerAnnotations map[string]string
 		ThanosRulerGeneration  int64
-		ThanosRulerSpec        monitoringv1.ThanosRulerSpec
 		Config                 Config
 		StatefulSetSpec        appsv1.StatefulSetSpec
 		RuleConfigMaps         []string `hash:"set"`
@@ -716,7 +715,6 @@ func createSSetInputHash(tr monitoringv1.ThanosRuler, c Config, ruleConfigMapNam
 		ThanosRulerLabels:      tr.Labels,
 		ThanosRulerAnnotations: tr.Annotations,
 		ThanosRulerGeneration:  tr.Generation,
-		ThanosRulerSpec:        tr.Spec,
 		Config:                 c,
 		StatefulSetSpec:        ss,
 		RuleConfigMaps:         ruleConfigMapNames,
--- a/pkg/thanos/operator.go
+++ b/pkg/thanos/operator.go
@@ -703,22 +703,30 @@ func (o *Operator) UpdateStatus(ctx context.Context, key string) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/thanos/operator.go
```diff
@@ -703,20 +703,28 @@ func (o *Operator) UpdateStatus(ctx context.Context, key string) error {
...
+		ThanosRulerLabels      map[string]string
+		ThanosRulerAnnotations map[string]string
+		ThanosRulerSpec        monitoringv1.ThanosRulerSpec
+		Config                 Config
+		StatefulSetSpec        appsv1.StatefulSetSpec
+		RuleConfigMaps         []string `hash:"set"`
+	}{
+		ThanosRulerLabels:      tr.Labels,
+		ThanosRulerAnnotations: tr.Annotations,
+		ThanosRulerSpec:        tr.Spec,
```
simonpasquier, line 717: we should be able to use tr.Generation instead of the full spec like we do with the Prometheus and Alertmanager controllers.

Azanul: What about http server configuration?

Azanul: And tls `SharedNames`?

simonpasquier: Thanos Ruler doesn't have HTTP server config and doesn't use the SharedSecret struct.

Azanul: noted

Azanul: I've used the `Generation` while creating hash, please check and let me know if anything else needs to be done

pkg/thanos/operator.go
```diff
@@ -703,20 +703,30 @@ func (o *Operator) UpdateStatus(ctx context.Context, key string) error {
...
 	hash, err := hashstructure.Hash(struct {
-		TR monitoringv1.ThanosRuler
-		C  Config
-		S  interface{}
-		R  []string `hash:"set"`
-	}{tr, c, ss, ruleConfigMapNames},
+		ThanosRulerLabels      map[string]string
+		ThanosRulerAnnotations map[string]string
+		ThanosRulerGeneration  int64
+		ThanosRulerSpec        monitoringv1.ThanosRulerSpec
```
simonpasquier, line 711: you can remove the `ThanosRulerSpec ` field (a change in the generation value implies an update of the spec).

Azanul: Prometheus and AlertManager both use Generation as well as spec

simonpasquier: unless I'm mistaken, they use the statefulset's spec not the Prometheus/Alertmanager spec?

Azanul: You're right, I need to have my eyes checked.

Issue #5738, opened 2023-07-11
Implement golden files for test assertion - Admission package


nicolastakashi:
**What is missing?**

As described in the main issue #5724 we should implement the golden files strategy to improve the tests' readability as well as maintainability.

**Why do we need it?**

- Improve tests maintainability
- Improve tests the readability

**Environment**

* Prometheus Operator version: N/A

**Anything else we need to know?**:

ArthurSens:
fixed by https://github.com/prometheus-operator/prometheus-operator/pull/5785

PR #5785, opened 2023-08-01
chore: adding golden files support on admission tests

nicolastakashi:
## Description

I'm adding golden file supports on `admission` package, solves #5738

## Type of change

## Changelog entry

```release-note
adding golden file supports on `admission` package
```

### DIFF
--- a/pkg/admission/admission_test.go
+++ b/pkg/admission/admission_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
+	"gotest.tools/v3/golden"
 	v1 "k8s.io/api/admission/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
@@ -40,7 +41,7 @@ func TestMutateRule(t *testing.T) {
 	ts := server(api().servePrometheusRulesMutate)
 	defer ts.Close()
 
-	resp := sendAdmissionReview(t, ts, goodRulesWithAnnotations)
+	resp := sendAdmissionReview(t, ts, golden.Get(t, "goodRulesWithAnnotations.golden"))
 
 	if len(resp.Response.Patch) == 0 {
 		t.Errorf("Expected a patch to be applied but found none")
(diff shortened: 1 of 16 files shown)


ArthurSens:
While reading the [go docs](https://pkg.go.dev/gotest.tools/v3/golden) of this library, I can see that `golden.Get` does not update golden files even if we pass the -update flag. I believe we need to refactor those tests so we can use, [Assert](https://pkg.go.dev/gotest.tools/v3/golden#Assert), [AssertBytes](https://pkg.go.dev/gotest.tools/v3/golden#AssertBytes), [Bytes](https://pkg.go.dev/gotest.tools/v3/golden#Bytes) or [String](https://pkg.go.dev/gotest.tools/v3/golden#String)

simonpasquier:
> While reading the [go docs](https://pkg.go.dev/gotest.tools/v3/golden) of this library, I can see that `golden.Get` does not update golden files even if we pass the `-update` flag. I believe we need to refactor those tests so we can use, [Assert](https://pkg.go.dev/gotest.tools/v3/golden#Assert), [AssertBytes](https://pkg.go.dev/gotest.tools/v3/golden#AssertBytes), [Bytes](https://pkg.go.dev/gotest.tools/v3/golden#Bytes) or [String](https://pkg.go.dev/gotest.tools/v3/golden#String)

I think that it's ok here because we use golden files for storing input data, not the expected output.

simonpasquier (approved):
LGTM but I'll let the final word to @ArthurSens

ArthurSens (approved):
> I think that it's ok here because we use golden files for storing input data, not the expected output.

Ooooh good point! Could we add a comment about this on the code? I'm 100% sure I'll forget about this eventually 😅

simonpasquier:
Maybe the comment should go to https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#contribution-flow?

nicolastakashi:
> Maybe the comment should go to https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#contribution-flow?

I'm not exactly what comment we should add, but on the fist PR I sent including golden files, we add a line talking about the golden files.

Maybe worth in another PR improve the documentation about testing.

simonpasquier:
@nicolastakashi I was referring to @ArthurSens's remark about adding a comment for `golden.Get()` and indeed suggested to have the comment next to the mention of the golden files in the contributing doc.
Personally I'm fine with the PR in the current state and not sure that we need to add something to the contributing guide. Let's merge it and if Arthur has a good suggestion then we can have a follow-up PR ;)

PR #5795, opened 2023-08-04
feat: adding golden files support on alertmanger tests

nicolastakashi:
## Description

I'm adding golden file supports on `admission` package, solves #5739

## Type of change

## Changelog entry

```release-note
adding golden file supports on `alertmanager` package
```

### DIFF
--- a/pkg/alertmanager/amcfg_test.go
+++ b/pkg/alertmanager/amcfg_test.go
@@ -24,11 +24,11 @@ import (
 
 	"github.com/blang/semver/v4"
 	"github.com/go-kit/log"
-	"github.com/google/go-cmp/cmp"
 	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/timeinterval"
 	"github.com/prometheus/common/model"
 	"github.com/stretchr/testify/require"
+	"gotest.tools/v3/golden"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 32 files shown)


pkg/alertmanager/amcfg_test.go
```diff
@@ -2301,9 +1879,7 @@ templates: []
 			}
 
 			// Verify the generated yaml is as expected
-			if diff := cmp.Diff(tc.expected, string(cfgBytes)); diff != "" {
-				t.Errorf("Unexpected result (-want +got):\n%s", diff)
-			}
+			golden.Assert(t, string(cfgBytes), tc.golden)
```
ArthurSens, line 1882: I feel like the diff is really useful sometimes. Not a hard blocker, but it would be nice to keep it somehow if possible :)

pkg/alertmanager/amcfg_test.go
```diff
@@ -3789,7 +3329,7 @@ templates: []
 
 	for _, tc := range testCase {
 		t.Run(tc.name, func(t *testing.T) {
-			ac, err := alertmanagerConfigFromBytes(tc.rawConf)
+			ac, err := alertmanagerConfigFromBytes(golden.Get(t, tc.golden))
```
ArthurSens, line 3332: We can't use golden.Get(), see comment in https://github.com/prometheus-operator/prometheus-operator/pull/5785#issuecomment-1670193826

pkg/alertmanager/amcfg_test.go
simonpasquier: it seems that gotest.tools/v3/golden uses github.com/google/go-cmp under the hood so it will display the nice diff :)

ArthurSens: Noice! TIL

ArthurSens (approved):
Similar to https://github.com/prometheus-operator/prometheus-operator/pull/5785#pullrequestreview-1571756921 :)

PR #5801, opened 2023-08-09
Remove crd descriptions

skl:
## Description

When trying to apply CRDs with `kubectl`, the following error message is observed:

```
The CustomResourceDefinition "prometheuses.monitoring.coreos.com" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```

Using `maxDescLen` to remove the descriptions from the CRDs significantly reduces the file size, e.g. for `example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml`:

- Before: 580kB
- After: 190kB

## Type of change

## Changelog entry

```release-note
Remove descriptions from CRDs to prevent hitting the kubectl annotation limit.
```


skl:
@slashpai oh I didn't know about that, thanks!

ArthurSens:
Closing since the problem is already solved :)

PR #5805, opened 2023-08-11
pkg/apis: Add metric relabeling to ScrapeConfigs

### LINKED ISSUE #5803: Add metric_relabel_configs to ScrapeConfig CRD
**What is missing?**

Support of `metric_relabel_configs` directive in ScrapeConfig CRD.

Prometheus configuration : https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs

**Why do we need it?**

To allow relabeling on samples before ingestion.

Today the ScrapeConfig CRD only supports relabeling before scraping (`relabelings` property)

**Environment**

* Prometheus Operator version: v0.67.1

rimitchell:
## Description

Brings support for metric relabeling to the ScrapeConfig object. Aligning with the existing functionality of Service, Pod and Probe monitors.

Fixes #5803 

## Type of change

## Changelog entry

ScrapeConfigs: Now supports metric relabeling

```release-note
- Add metricRelabel functionality to ScrapeConfigs
```


pkg/prometheus/promcfg.go
```diff
@@ -2520,6 +2520,8 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		})
 	}
 
+	cfg = append(cfg, yaml.MapItem{Key: "metric_relabel_configs", Value: generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.MetricRelabelConfigs))})
+
```
xiu, line 2524: Most ScrapeConfig unit tests are failing as we're passing a new empty `metric_relabel_config` field to the configuration if it's not set. The test cases you updated added the key but not at the right spot. The following ensures that we're adding the key only if necessary.

```suggestion
      if sc.Spec.MetricRelabelConfigs != nil {
	cfg = append(cfg, yaml.MapItem{Key: "metric_relabel_configs", Value: generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.MetricRelabelConfigs))})
      }
```

pkg/prometheus/promcfg_test.go
```diff
@@ -8879,6 +8879,7 @@ func TestScrapeConfigSpecConfig(t *testing.T) {
   external_labels:
     prometheus: default/test
     prometheus_replica: $(POD_NAME)
+  metric_relabel_configs: []
```
xiu, line 8882: We shouldn't add an empty slice if `spec.MetricRelabelConfigs` wasn't set, see my previous comment. This is true for all the tests were you added this.

xiu (changes requested):
Thanks for this! Added some comments. To regenerate assets, please use `make generate` and commit the results.

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -144,6 +144,8 @@ type ScrapeConfigSpec struct {
 	// Only valid in Prometheus versions 2.27.0 and newer.
 	// +optional
 	LabelValueLengthLimit *uint64 `json:"labelValueLengthLimit,omitempty"`
+	// MetricRelabelConfigs to apply to samples before ingestion.
+	MetricRelabelConfigs []*v1.RelabelConfig `json:"metricRelabelings,omitempty"`
```
nicolastakashi, line 148: Small nit: we should include +optional annotation

pkg/prometheus/promcfg.go
rimitchell: Oops. My bad on missing the generated files. I've added this check and the optional annotation below and have now included the generated files

pkg/prometheus/promcfg.go
```diff
@@ -2520,6 +2520,10 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		})
 	}
 
+	if sc.Spec.MetricRelabelConfigs != nil {
```
nicolastakashi, line 2523: I'd suggest this just to improve a bit readability.
```suggestion
	if sc.Spec.MetricRelabelConfigs != nil {
		relabelConfigValue = generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.MetricRelabelConfigs))
		cfg = append(cfg, yaml.MapItem{
			Key:   "metric_relabel_configs",
			Value: relabelConfigValue,
		})
	}	
```

pkg/prometheus/resource_selector.go
```diff
@@ -691,6 +691,15 @@ func (rs *ResourceSelector) SelectScrapeConfigs(ctx context.Context, listFn List
 			continue
 		}
 
+		for _, rl := range sc.Spec.MetricRelabelConfigs {
```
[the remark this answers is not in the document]
ArthurSens: Option 1 isn't good since we'd skip the rejection function if the `action` field is empty.

I haven't looked deeply into the code, but option 3 could indeed be a change in functionality for other ServiceMonitors, PodMonitors and Probes. Let's keep that in a separate PR, if relevant 😅 

Option 2 looks like a good suggestion, but I can see that we're already using this same pattern every other place we call `validateRelabelConfig`. It would be nice to change those in a separate PR if you feel like it :)

ArthurSens (approved):
LGTM! Thanks for the PR

I'll let others take another look before merging it :)

pkg/prometheus/resource_selector.go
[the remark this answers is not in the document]
nicolastakashi: Yeah!
It's totally fine for me keep as it is and we do the changes we think it's valuable in another PR 💪🏽

simonpasquier (commented):
I suggest to wait for https://github.com/prometheus-operator/prometheus-operator/pull/5841 which is fixing a flaw in the current code regarding resource selection.

pkg/prometheus/resource_selector.go
```diff
@@ -691,6 +691,15 @@ func (rs *ResourceSelector) SelectScrapeConfigs(ctx context.Context, listFn List
...
+			if rl.Action != "" {
+				if err = validateRelabelConfig(rs.p, *rl); err != nil {
+					rejectFn(sc, err)
+					continue
+				}
+			}
+		}
```
ArthurSens, line 701:
```suggestion
		if err = validateRelabelConfigs(rs.p, sc.Spec.MetricRelabelConfigs); err != nil {
			rejectFn(sc, fmt.Errorf("metricRelabelConfigs: %w", err))
			continue
		}
```

ArthurSens (commented):
@rimitchell, could you rebase and make one adjustment?

Simon found and fixed a bug in #5841 that affects your PR 😬

pkg/prometheus/resource_selector.go
rimitchell: Rebased and code updated to use the _validateRelabelConfigs_ function

pkg/prometheus/resource_selector.go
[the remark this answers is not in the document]
rimitchell: Thank you for catching that. I've added a set for the metric relabeling

ArthurSens (approved):
Thanks for your patience through the whole process!

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
nicolastakashi: Very small nit but I'm not sure if we need a pointer here since default value for array in go lang is empty.
wdyt @ArthurSens and @simonpasquier ?

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -144,6 +144,9 @@ type ScrapeConfigSpec struct {
 	// Only valid in Prometheus versions 2.27.0 and newer.
 	// +optional
 	LabelValueLengthLimit *uint64 `json:"labelValueLengthLimit,omitempty"`
+	// MetricRelabelConfigs to apply to samples before ingestion.
+	// +optional
+	MetricRelabelConfigs []*v1.RelabelConfig `json:"metricRelabelings,omitempty"`
```
nicolastakashi, line 149: I'm not sure if we need a pointer here since default value for an array in golang is empty.
wdyt @ArthurSens and @simonpasquier ?

simonpasquier (approved):
Thanks!

PR #5806, opened 2023-08-13
statefulset: remove redundant nil check around Labels and Annotations

Juneezee:
## Description

From the Go docs:

> "If the map is nil, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/Cw_wnqYzemu

## Type of change

## Changelog entry

```release-note
statefulset: remove redundant nil check around Labels and Annotations
```

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -360,15 +360,11 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 		"alertmanager":                 a.Name,
 	}
 	if a.Spec.PodMetadata != nil {
-		if a.Spec.PodMetadata.Labels != nil {
-			for k, v := range a.Spec.PodMetadata.Labels {
-				podLabels[k] = v
-			}
+		for k, v := range a.Spec.PodMetadata.Labels {
+			podLabels[k] = v
 		}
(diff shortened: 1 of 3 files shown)


pkg/alertmanager/statefulset.go
```diff
@@ -360,15 +360,11 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 		"alertmanager":                 a.Name,
 	}
 	if a.Spec.PodMetadata != nil {
-		if a.Spec.PodMetadata.Labels != nil {
-			for k, v := range a.Spec.PodMetadata.Labels {
-				podLabels[k] = v
-			}
+		for k, v := range a.Spec.PodMetadata.Labels {
```
nicolastakashi, line 363: I'm just curious if beyond the readability improvement, this change may have some perf impact ~~(probably not expressive enough)~~

nicolastakashi: Owow amazing @Juneezee !
Thanks for sharing that, it was totally out of curiosity hehe

nicolastakashi (approved):

simonpasquier (approved):
thanks! there's definitely no need for checking nil slices.

PR #5807, opened 2023-08-14
enhancement: deploy operator in single namespace without `ClusterRole`

zhuangzhi-cc:
## Description

We need deploy prometheus operator in single namespace without cluster role rabc. Follow [simonpasquier](https://github.com/simonpasquier)'s suggestion if the --namespaces is a single namespace, operator would running in none clusterrole mode.
eg.  --namespaces=monitoring

## Type of change

## Changelog entry

1. add single namespace to `namespaces` argument
2. Createo `role` rbac instead of   `ClusterRole`
3. deploy prometheus operator, prometheus and application PODs in same namespace.
```release-note

```


pkg/alertmanager/operator.go
```diff
@@ -243,32 +243,34 @@ func (c *Operator) bootstrap(ctx context.Context) error {
...
-		if listwatch.IsAllNamespaces(allowList) {
-			nsResyncPeriod = resyncPeriod
-		}
-		nsInf := cache.NewSharedIndexInformer(
-			o.metrics.NewInstrumentedListerWatcher(
-				listwatch.NewUnprivilegedNamespaceListWatchFromClient(ctx, o.logger, o.kclient.CoreV1().RESTClient(), allowList, o.config.Namespaces.DenyList, fields.Everything()),
-			),
-			&v1.Namespace{}, nsResyncPeriod, cache.Indexers{},
-		)
+	if !operator.IsSingleNamespace(c.config.Namespaces) {
```
simonpasquier, line 246: I think that the test should be more elaborated and dependent on the controller: e.g. Alertmanager can skip the namespace informers if AlertmanagerAllowList and AlertmanagerConfigAllowList have a single item (and different from v1.NamespaceAll).

For Prometheus and Thanos, the checks will be different.

pkg/prometheus/server/operator.go
```diff
@@ -969,11 +978,15 @@ func (c *Operator) handleConfigMapUpdate(old, cur interface{}) {
 }
 
 func (c *Operator) enqueueForPrometheusNamespace(nsName string) {
-	c.enqueueForNamespace(c.nsPromInf.GetStore(), nsName)
+	if !operator.IsSingleNamespace(c.config.Namespaces) {
+		c.enqueueForNamespace(c.nsPromInf.GetStore(), nsName)
```
simonpasquier, line 982: We still need to run the reconciliation. I think that the change should rather be to `enqueueForNamespace()`.

pkg/prometheus/server/operator.go
```diff
@@ -521,7 +530,7 @@ func (c *Operator) Run(ctx context.Context) error {
 
 	c.addHandlers()
 
-	if c.kubeletSyncEnabled {
+	if c.kubeletSyncEnabled && !operator.IsSingleNamespace(c.config.Namespaces) {
```
simonpasquier, line 533: why do we need this?

pkg/prometheus/server/operator.go
zhuangzhi-cc: Yes, it woule be more simplefy, I would change it.

pkg/prometheus/server/operator.go
zhuangzhi-cc: level=error ts=2023-09-01T11:39:50.560491434Z caller=operator.go:626 component=prometheusoperator msg="Syncing nodes into Endpoints object failed" err="listing nodes failed: nodes is forbidden: User \"system:serviceaccount:rio:prometheus-operator\" cannot list resource \"nodes\" in API group \"\" at the cluster scope"
That's the error log if disable this limit.

pkg/alertmanager/operator.go
zhuangzhi-cc: code changed

pkg/prometheus/server/operator.go
zhuangzhi-cc: code changed

simonpasquier (commented):
Thinking more about it, I wonder if it wouldn't be simpler in terms of implementation to add a special "single namespace" lister/watcher. In other words, extend `NewUnprivilegedNamespaceListWatchFromClient()` to avoid the "get namespace" requests when a single namespace (not equal to all namespaces) is given.

Another thing to keep in mind: when running the operator in single namespace mode, the `.spec.*NamespaceSelector` fields become ineffective. When these selectors aren't nil, I wonder if the operator should report it via the "Reconciled=False" status configuration or if it should ignore the namespace selectors? I'd rather lean towards the first option as it's more explicit.

zhuangzhi-cc:
> Thinking more about it, I wonder if it wouldn't be simpler in terms of implementation to add a special "single namespace" lister/watcher. In other words, extend `NewUnprivilegedNamespaceListWatchFromClient()` to avoid the "get namespace" requests when a single namespace (not equal to all namespaces) is given.
> 
> Another thing to keep in mind: when running the operator in single namespace mode, the `.spec.*NamespaceSelector` fields become ineffective. When these selectors aren't nil, I wonder if the operator should report it via the "Reconciled=False" status configuration or if it should ignore the namespace selectors? I'd rather lean towards the first option as it's more explicit.

Do you means create a fake ListerWatcher in  `NewUnprivilegedNamespaceListWatchFromClient`  like:

func NewUnprivilegedNamespaceListWatchFromClient(
	ctx context.Context,
	l log.Logger,
	c cache.Getter,
	allowedNamespaces, deniedNamespaces map[string]struct{},
	fieldSelector fields.Selector,
) cache.ListerWatcher {
......
	if IsAllNamespaces(allowedNamespaces) {
....
       }

      if IsSingleNamespace(allowedNamespaces) {

		listFunc := func(options metav1.ListOptions) (runtime.Object, error) {
			list := &v1.NamespaceList{}
			return list, nil
		}
		watchFunc := func(_ metav1.ListOptions) (watch.Interface, error) {
			// Since the client does not have Watch privileges, do not
			// actually watch anything. Use a watch.FakeWatcher here to
			// implement watch.Interface but not send any events.
			return watch.NewFake(), nil
		}
		return &cache.ListWatch{ListFunc: listFunc, WatchFunc: watchFunc}
     }
.....
}

simonpasquier:
Yes something along those lines. It could even work when there are multiple namespaces in `allowedNamespaces` beccause the controllers only care about the namespace names after all. 

````golang
func NewUnprivilegedNamespaceListWatchFromClient(
ctx context.Context,
l log.Logger,
c cache.Getter,
allowedNamespaces, deniedNamespaces map[string]struct{},
fieldSelector fields.Selector,
) cache.ListerWatcher {
...
if IsAllNamespaces(allowedNamespaces) {
   ....
}


listFunc := func(options metav1.ListOptions) (runtime.Object, error) {
                list := &v1.NamespaceList{}
                for name := range allowedNamespaces {
                        result := &v1.Namespace{}
                        list.Items = append(list.Items, &v1.Namespace{
                                ObjectMeta: metav1.ObjectMeta{Name: name},
                        })
                }
                return list, nil		
	})
}
watchFunc := func(_ metav1.ListOptions) (watch.Interface, error) {
		// Since the client does not have Watch privileges, do not
		// actually watch anything. Use a watch.FakeWatcher here to
		// implement watch.Interface but not send any events.
		return watch.NewFake(), nil
}

return &cache.ListWatch{ListFunc: listFunc, WatchFunc: watchFunc}
)
````

File: pkg/operator/config.go [hunk 4]
```diff
@@ -192,3 +192,23 @@ type Namespaces struct {
 	// Allow list for prometheus/alertmanager custom resources.
 	PrometheusAllowList, AlertmanagerAllowList, AlertmanagerConfigAllowList, ThanosRulerAllowList map[string]struct{}
 }
+
+// Check if this is a single namespace deployment,
+func isSingleNamespace(ns map[string]struct{}) bool {
+	_, ok := ns[v1.NamespaceAll]
+	return !ok && len(ns) == 1
```
simonpasquier, line 199: (nit)
```suggestion
	_, allNamespaces := ns[v1.NamespaceAll]
	return !allNamespaces && len(ns) == 1
```

pkg/prometheus/server/operator.go
```diff
@@ -521,7 +523,7 @@ func (c *Operator) Run(ctx context.Context) error {
 
 	c.addHandlers()
 
-	if c.kubeletSyncEnabled {
+	if c.kubeletSyncEnabled && !operator.IsPrometheusInSingleNamespace(c.config.Namespaces) {
```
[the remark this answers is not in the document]
zhuangzhi-cc: If don't do this check, there would be some errors because of `nodes monitor` need cluserrole like: 

level=error ts=2023-09-01T11:39:50.560491434Z caller=operator.go:626 component=prometheusoperator msg="Syncing nodes into Endpoints object failed" err="listing nodes failed: nodes is forbidden: User "system:serviceaccount:rio:prometheus-operator" cannot list resource "nodes" in API group "" at the cluster scope"

those error logs not impact the program running

pkg/operator/config.go
zhuangzhi-cc: this code has been removed.

simonpasquier:
There's an issue with the `testOperatorNSScope` test (and more precisely the `SingleNS` subtest). The reason is that the namespace selection returns an empty list because the operator doesn't have access to the namespace's labels:

https://github.com/prometheus-operator/prometheus-operator/blob/6d3714e04be52e2e53942772eca964441ac9cfbc/pkg/prometheus/server/rules.go#L174-L182

We have the same "problem" in other places:

https://github.com/prometheus-operator/prometheus-operator/blob/6d3714e04be52e2e53942772eca964441ac9cfbc/pkg/prometheus/resource_selector.go#L78-L91

This is similar to what I said in a previous comment
> Another thing to keep in mind: when running the operator in single namespace mode, the .spec.*NamespaceSelector fields become ineffective. When these selectors aren't nil, I wonder if the operator should report it via the "Reconciled=False" status configuration or if it should ignore the namespace selectors? I'd rather lean towards the first option as it's more explicit.

In retrospect, the controllers should rather default the namespace selectors to nil and report the situation via the `Reconciled` status condition with a dedicated reason and message:

```
  availableReplicas: 1
  conditions:
  - lastTransitionTime: "2023-09-05T12:01:49Z"
    observedGeneration: 5
    status: "True"
    type: Available
  - lastTransitionTime: "2023-09-05T12:01:49Z"
    observedGeneration: 5
    status: "True"
    type: Reconciled
    reason: "NamespaceSelectorsIgnored"
    message: "The resource specifies namespace selectors for the ServiceMonitor, PrometheusRule custom resources but the operator can only select resources from the 'foo' namespace."
````

**Comment by simonpasquier** [comment]
@zhuangzhi-cc thanks for keeping up the PR up-to-date! I'd advise to wait for https://github.com/prometheus-operator/prometheus-operator/pull/5898 which fixes a problem that's a bit different but after it (and with the refactoring), it should be even easier to implement single-namespace watcher.

**Comment by simonpasquier** [comment]
@zhuangzhi-cc now that #5898 and #5934 have merged, it's possible to run the operator watching a single namespace without ClusterRoleBinding. If you create a role binding to the `prometheus-operator` cluster role in the required namespace, it should just work. Does it satisfy your use case?

```
kc create rolebinding -n my_namespace prometheus-operator --clusterrole prometheus-operator --serviceaccount default:prometheus-operator
./operator --namespaces my_namespace --as system:serviceaccount:default:prometheus-operator
```

zhuangzhi-cc:
> @zhuangzhi-cc thanks for keeping up the PR up-to-date! I'd advise to wait for #5898 which fixes a problem that's a bit different but after it (and with the refactoring), it should be even easier to implement single-namespace watcher.

@simonpasquier I see that PR #5898  has alreay merged, is there any idea about single-namespace feature?

simonpasquier:
> @simonpasquier I see that PR #5898 has alreay merged, is there any idea about single-namespace feature?

@zhuangzhi-cc please check my comment right before (https://github.com/prometheus-operator/prometheus-operator/pull/5807#issuecomment-1729782159).

zhuangzhi-cc:
> > @simonpasquier I see that PR #5898 has alreay merged, is there any idea about single-namespace feature?
> 
> @zhuangzhi-cc please check my comment right before ([#5807 (comment)](https://github.com/prometheus-operator/prometheus-operator/pull/5807#issuecomment-1729782159)).

Sorry for my  browser not sync the latest message because I am not connect to VPN :), I would make some test with the latest code. thanks a lot!

zhuangzhi-cc:
this feature has supported at https://github.com/prometheus-operator/prometheus-operator/pull/5898 and https://github.com/prometheus-operator/prometheus-operator/pull/5934

PR #5823, opened 2023-08-20
Get status less frequently to avoid rate limits

### LINKED ISSUE #5556: Test flakiness due to rate limits
It seems like we have a flaky test on the loose at the moment. 

I've seen this test fail in a couple of PRs (can't provide links because the pipeline was re-triggered 😅), but I saw it occur in our default branch as well[[1](https://github.com/prometheus-operator/prometheus-operator/actions/runs/4895979176/jobs/8742210525)].

Opening the issue just so we could discuss a solution. (I'm still finding time to debug the test myself)

ArthurSens:
## Description

I hope it fixes #5556 😅, would need to merge and keep an eye out if we still get rate limits in tests to make sure.

## Type of change

## Changelog entry

```release-note

```


philipgough:
lets merge and see

PR #5841, opened 2023-08-23
fix: reject monitors with invalid relabel configs

simonpasquier:
## Description

I noticed that probes, service monitors, pod monitors and scrape configs with invalid relabel configs were not rejected by the operator. This PR fixes it.

## Type of change

## Changelog entry

```release-note
reject ServiceMonitor, PodMonitor, Probe and ScrapeConfig with invalid relabel configs.
```

### DIFF
--- a/pkg/prometheus/resource_selector.go
+++ b/pkg/prometheus/resource_selector.go
@@ -152,20 +152,14 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
 				break
 			}
 
-			for _, rl := range endpoint.RelabelConfigs {
-				if rl.Action != "" {
-					if err = validateRelabelConfig(rs.p, *rl); err != nil {
-						break
-					}
-				}
+			if err = validateRelabelConfigs(rs.p, endpoint.RelabelConfigs); err != nil {
+				err = fmt.Errorf("relabelConfigs: %w", err)
+				break
 			}
 
-			for _, rl := range endpoint.MetricRelabelConfigs {
-				if rl.Action != "" {
-					if err = validateRelabelConfig(rs.p, *rl); err != nil {
-						break
-					}
-				}
+			if err = validateRelabelConfigs(rs.p, endpoint.MetricRelabelConfigs); err != nil {
+				err = fmt.Errorf("metricRelabelConfigs: %w", err)
+				break
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/prometheus/resource_selector.go
```diff
@@ -225,55 +219,75 @@ func validateScrapeIntervalAndTimeout(p monitoringv1.PrometheusInterface, scrape
...
+
 	version, err := semver.ParseTolerant(promVersion)
 	if err != nil {
 		return errors.Wrap(err, "failed to parse Prometheus version")
 	}
+
 	minimumVersionCaseActions := version.GTE(semver.MustParse("2.36.0"))
 	minimumVersionEqualActions := version.GTE(semver.MustParse("2.41.0"))
+	if rc.Action == "" {
+		rc.Action = string(relabel.Replace)
```
slashpai, line 248: shouldn't this be already taken care since we set default as `replace`?
https://github.com/prometheus-operator/prometheus-operator/blob/78a28753bef7f702c27f16ea38449f599d805301/pkg/apis/monitoring/v1/prometheus_types.go#L1340-L1341

pkg/prometheus/resource_selector_test.go
```diff
@@ -400,77 +415,193 @@ func TestValidateRelabelConfig(t *testing.T) {
...
+				}
+			},
+			selected: false,
 		},
 	} {
-		t.Run(fmt.Sprintf("case %s %s", tc.scenario, tc.proberSpec.URL), func(t *testing.T) {
-			err := validateProberURL(tc.proberSpec.URL)
-			if err != nil && !tc.expectedErr {
-				t.Fatalf("expected no error, got: %v", err)
+		t.Run(fmt.Sprintf(tc.scenario), func(t *testing.T) {
```
ArthurSens, line 570:
```suggestion
		t.Run(tc.scenario, func(t *testing.T) {
```

ArthurSens (commented):
I'm struggling a little bit to understand what/where is the bug that you're fixing here 😬 

Do I understand it correctly that we're allowing actions with a mixture of Lower/Upper case and if we fail to compare with `relabel.<action name>`?

If that's the case, what do you think about passing this validation to kubebuilder annotations?

pkg/prometheus/resource_selector_test.go
ArthurSens: small nit on all tests created

pkg/prometheus/resource_selector.go
simonpasquier: This is for tests + environments that don't have a recent version of the CRD (e.g. one without the default value).

simonpasquier:
> I'm struggling a little bit to understand what/where is the bug that you're fixing here 😬
> 


One issue is here:

https://github.com/prometheus-operator/prometheus-operator/blob/34bb905e3ca90963ffdaa50cbe1924137f9a5ed7/pkg/prometheus/resource_selector.go#L535-L548

If an invalid relabel config is detected, it will be logged but it won't continue from the outer for loop so the probe will be added to the map.

Another issue here:

https://github.com/prometheus-operator/prometheus-operator/blob/34bb905e3ca90963ffdaa50cbe1924137f9a5ed7/pkg/prometheus/resource_selector.go#L155-L172

If L157 returns an error, `err` might be overwritten at L165 and again the service monitor isn't rejected...

TL;DR: write unit tests to assert that the code is doing what it should ;)

> Do I understand it correctly that we're allowing actions with a mixture of Lower/Upper case and then we fail to compare with `relabel.<action name>`?
> 
> If that's the case, what do you think about passing this validation to kubebuilder annotations?

No upper/lower case actions are fine (in fact the Prometheus config supports everything but will always compare to the lower-case version).

pkg/prometheus/resource_selector.go
```diff
@@ -152,20 +152,14 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
 				break
 			}
 
-			for _, rl := range endpoint.RelabelConfigs {
-				if rl.Action != "" {
```
simonpasquier, line 156: I'm not sure why we had this test in the first place. I've modified `validateRelabelConfig()` to default to the replace action if the field is empty.

ArthurSens (approved):

ArthurSens:
> > I'm struggling a little bit to understand what/where is the bug that you're fixing here 😬
> 
> One issue is here:
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/34bb905e3ca90963ffdaa50cbe1924137f9a5ed7/pkg/prometheus/resource_selector.go#L535-L548
> 
> If an invalid relabel config is detected, it will be logged but it won't continue from the outer for loop so the probe will be added to the map.
> 
> Another issue here:
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/34bb905e3ca90963ffdaa50cbe1924137f9a5ed7/pkg/prometheus/resource_selector.go#L155-L172
> 
> If L157 returns an error, `err` might be overwritten at L165 and again the service monitor isn't rejected...
> 
> TL;DR: write unit tests to assert that the code is doing what it should ;)
> 
> > Do I understand it correctly that we're allowing actions with a mixture of Lower/Upper case and then we fail to compare with `relabel.<action name>`?
> > If that's the case, what do you think about passing this validation to kubebuilder annotations?
> 
> No upper/lower case actions are fine (in fact the Prometheus config supports everything but will always compare to the lower-case version).

Ooooh now I get it, we're breaking the inner loop when we wanted to break the outer loop. Thanks for the explanation!

PR #5851, opened 2023-08-25
Checkout repository during dependabot workflow

ArthurSens:
## Description

An attempt to fix https://github.com/prometheus-operator/prometheus-operator/pull/5849

Changes:
* Checkout the repository
* Only commit if there is a diff

## Type of change

## Changelog entry

```release-note

```


.github/workflows/automerge-dependabot.yaml
```diff
@@ -15,13 +15,21 @@ jobs:
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
       - name: Generate Documentation/compatibility.md
         if: contains(steps.metadata.outputs.dependency-names, 'k8s.io/api')
         run: |
           make generate --always-make
-          git add Documentation/compatibility.md
-          git commit -m "Generate Documentation/compatibility.md"
-          git push
+          changedFiles=$(git diff --name-only | wc -l)
```
simonpasquier, line 27: you can do `git diff --exit-code`

ArthurSens: but that would make the workflow fail, wouldn't it?

What I was thinking here was that we don't want to push anything if there aren't any changes to commit (An empty commit has exit code != 0)

.github/workflows/automerge-dependabot.yaml
```diff
@@ -15,13 +15,21 @@ jobs:
...
+          if [[ "$changedFiles" -eq 0 ]]; then
```
simonpasquier, line 28:
```suggestion
          if ! git diff --exit-code; then
```

simonpasquier (approved):

Issue #5783, opened 2023-08-01
AzureAD remote write authentication support


MadsHT:
**What is missing?**

In [Prometheus v.2.45.0](https://github.com/prometheus/prometheus/releases/tag/v2.45.0) the `azuread` authentication method for remote write was introduced from [prometheus/prometheus#11944](https://github.com/prometheus/prometheus/pull/11944) and skimming through the crd and code it does not seem like either of those supports this ned `azuread` property.

**Why do we need it?**

We need it to simplify the process of remote writing metrics to our Azure monitor workspace.

**Environment**

* Prometheus Operator version:

    ![billede](https://github.com/prometheus-operator/prometheus-operator/assets/31923273/9a6a8230-e429-4d61-bfeb-c9be16f1d442)
    
    

**Anything else we need to know?**:

heylongdacoder:
Hi @MadsHT, we need your opinion about this(https://github.com/prometheus-operator/prometheus-operator/pull/5852#discussion_r1308473369). 😄

MadsHT:
@heylongdacoder Sorry for not replying, work and life got in the way, Awesome that you guys looked at this, and got it through! :D

PR #5852, opened 2023-08-26
prometheus: Add AzureAD for remote write

heylongdacoder:
Fixes: #5783

## Description

## Type of change

## Changelog entry

```release-note
Add AzureAD for Prometheus remote write.
```


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1202,6 +1211,22 @@ type Sigv4 struct {
 	RoleArn string `json:"roleArn,omitempty"`
 }
 
+// AzureAD defines the configuration for remote write's azuread parameters.
+// +k8s:openapi-gen=true
+type AzureAD struct {
+	// The Azure Cloud. Options are 'AzurePublic', 'AzureChina', or 'AzureGovernment'. If blank, AzurePublic is used.
+	Cloud string `json:"cloud,omitempty"`
```
nicolastakashi, line 1218: I think would be useful have this annotation `// +kubebuilder:validation:Enum=AzureChina; AzureGovernment; AzurePublic"` and maybe `// +kubebuilder:default:="AzurePublic"`

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1202,6 +1211,22 @@ type Sigv4 struct {
...
+	// ManagedIdentity defines the Azure User-assigned Managed identity.
+	ManagedIdentity *ManagedIdentity `json:"managedIdentity"`
```
nicolastakashi, line 1220: If it might be null worth to add `// +optional` annotation

simonpasquier: It seems to me that ClientID is a mandatory field, hence neither  `managedIdentity` nor `managedIdentity.clientID` can't be optional.

https://github.com/prometheus/prometheus/blob/1b9a53b5eea62e82a65bb65a8139c38dbd8f259e/storage/remote/azuread/azuread.go#L87-L93

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1202,6 +1211,22 @@ type Sigv4 struct {
...
+}
+
+// ManagedIdentity defines the Azure User-assigned Managed identity.
+// +k8s:openapi-gen=true
+type ManagedIdentity struct {
+	// The secret or configmap containing the client id
+	ClientID SecretOrConfigMap `json:"clientId"`
```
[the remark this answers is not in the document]
heylongdacoder: From the prometheus configuration doc(https://prometheus.io/docs/prometheus/latest/configuration/configuration/), the client_id was defined as <string> instead of <secret>, so I think is depends on what user think about this.

pkg/apis/monitoring/v1/prometheus_types.go
heylongdacoder: agreed on this. I also changed the field comment to `// The Azure Cloud. Options are 'AzurePublic', 'AzureChina', or 'AzureGovernment'.` 😄

pkg/apis/monitoring/v1/prometheus_types.go
heylongdacoder: yup the `managedIdentity` part is intended 😄

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
simonpasquier: If it's really not sensitive, I'd go with a string but I'm not sure yet.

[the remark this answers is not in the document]
ArthurSens: Hi @heylongdacoder , I've searched a bit (googled it :P) and noticed that AzureAD's Client ID is not sensitive at all, it is public data. 

Could we simplify here and just use an string instead of SecretOrConfigMap?

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: Is this field really optional?

Could we remove the `// +optional` comment if it isn't, or change the type to `*string` if it really is optional?

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: Could we add the comment `// +required` and remove the pointer then?

pkg/apis/monitoring/v1/prometheus_types.go
heylongdacoder: https://github.com/prometheus/prometheus/blob/main/docs/configuration/configuration.md?plain=1#L3536, yup I think that is optional and I change the type to `*string` 😄

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
heylongdacoder: thanks for the confirmation, I have changed it to string 😄

pkg/apis/monitoring/v1/prometheus_types.go
heylongdacoder: added `// +required` and removed pointer.

simonpasquier (approved):
Thanks!

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
MadsHT: ClientID is usually not considered sensitive information, sorry for my late reply.

PR #5877, opened 2023-09-01
Document server-side apply for CRDs

simonpasquier:
## Description

This is only partially documented in the kube-prometheus README but it comes up often in the issues. Hopefully some people will find it helpful.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/Documentation/troubleshooting.md
+++ b/Documentation/troubleshooting.md
@@ -11,6 +11,25 @@ draft: false
 description: Guide on troubleshooting the Prometheus Operator.
 ---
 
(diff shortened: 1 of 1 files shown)


ArthurSens (approved):
Another option is installing the CRDs with descriptions removed

simonpasquier:
@ArthurSens Sure but I'd rather document one way of working around the issue and I think that server-side apply is the best option (if you're before Kube v1.22, you'd better upgrade at some point :shrug:).

PR #5883, opened 2023-09-04
enhancement: Replace UpdateStatus() for ApplyStatus() to avoid informer errors

marioferh:
## Description

To set the status subresource, the operator make use of ApplyStatus() instead of UpdateStatus()  to avoid informer errors.

## Type of change

## Changelog entry

- replace UpdateStatus() for ApplyStatus() in order to avoid informer errors.

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -54,7 +54,8 @@ import (
 )
 
 const (
-	resyncPeriod = 5 * time.Minute
+	resyncPeriod                   = 5 * time.Minute
+	prometheusOperatorFieldManager = "PrometheusOperator"
 )
 
 var (
@@ -806,16 +807,14 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 
 	availableCondition := stsReporter.Update(a)
 	reconciledCondition := c.reconciliations.GetCondition(key, a.Generation)
-	availableConditionAC := GetAppliedConfigFromCondition(&availableCondition)
-	reconciledConditionAC := GetAppliedConfigFromCondition(&reconciledCondition)
 	a.Status.Paused = a.Spec.Paused
 
 	asac := monitoringv1ac.AlertmanagerStatus()
-	GetAppliedConfigStatusFromStatus(a, asac, availableConditionAC, reconciledConditionAC)
+	GetAppliedConfigStatusFromStatus(a, asac, availableCondition, reconciledCondition)
 
 	aac := monitoringv1ac.Alertmanager(a.Name, a.Namespace).WithStatus(asac)
 
-	if _, err = c.mclient.MonitoringV1().Alertmanagers(a.Namespace).ApplyStatus(ctx, aac, metav1.ApplyOptions{FieldManager: a.Name}); err != nil {
+	if _, err = c.mclient.MonitoringV1().Alertmanagers(a.Namespace).ApplyStatus(ctx, aac, metav1.ApplyOptions{FieldManager: prometheusOperatorFieldManager, Force: true}); err != nil {
 		return errors.Wrap(err, "failed to apply status subresource")
 	}
 
@@ -1815,7 +1814,7 @@ func GetAppliedConfigStatusFromStatus(a *monitoringv1.Alertmanager, asac *monito
 	asac.WithUpdatedReplicas(a.Status.UpdatedReplicas)
 	asac.WithUnavailableReplicas(a.Status.UnavailableReplicas)
 
-	for i := range values {
-		asac.WithConditions(values[i])
+	for _, condition := range conditions {
+		asac.WithConditions(GetAppliedConfigFromCondition(&condition))
 	}
 }
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -44,6 +44,7 @@ import (
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
 	"github.com/prometheus-operator/prometheus-operator/pkg/assets"
+	monitoringv1ac "github.com/prometheus-operator/prometheus-operator/pkg/client/applyconfiguration/monitoring/v1"
 	monitoringclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
 	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
 	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
@@ -805,11 +806,17 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 
 	availableCondition := stsReporter.Update(a)
 	reconciledCondition := c.reconciliations.GetCondition(key, a.Generation)
-	a.Status.Conditions = operator.UpdateConditions(a.Status.Conditions, availableCondition, reconciledCondition)
+	availableConditionAC := GetAppliedConfigFromCondition(&availableCondition)
+	reconciledConditionAC := GetAppliedConfigFromCondition(&reconciledCondition)
 	a.Status.Paused = a.Spec.Paused
 
-	if _, err = c.mclient.MonitoringV1().Alertmanagers(a.Namespace).UpdateStatus(ctx, a, metav1.UpdateOptions{}); err != nil {
-		return errors.Wrap(err, "failed to update status subresource")
+	asac := monitoringv1ac.AlertmanagerStatus()
+	GetAppliedConfigStatusFromStatus(a, asac, availableConditionAC, reconciledConditionAC)
+
+	aac := monitoringv1ac.Alertmanager(a.Name, a.Namespace).WithStatus(asac)
+
+	if _, err = c.mclient.MonitoringV1().Alertmanagers(a.Namespace).ApplyStatus(ctx, aac, metav1.ApplyOptions{FieldManager: a.Name}); err != nil {
+		return errors.Wrap(err, "failed to apply status subresource")
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


pkg/alertmanager/operator.go
```diff
@@ -808,7 +809,8 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	a.Status.Conditions = operator.UpdateConditions(a.Status.Conditions, availableCondition, reconciledCondition)
 	a.Status.Paused = a.Spec.Paused
 
-	if _, err = c.mclient.MonitoringV1().Alertmanagers(a.Namespace).UpdateStatus(ctx, a, metav1.UpdateOptions{}); err != nil {
+	aac := applyconfigurationv1.Alertmanager("altertmanager", a.Namespace)
```
simonpasquier, line 812: you need to build a complete *AlertmanagerStatusApplyConfiguration struct to provide to the ApplyStatus() method.

```suggestion
       asac := applyconfigurationv1.AlertmanagerStatus()
       
       ...
       
       asac.WithPaused(a.Spec.Paused)

	aac := applyconfigurationv1.Alertmanager(a.Name, a.Namespace).WithStatus(asac))
```

pkg/alertmanager/operator.go
```diff
@@ -808,7 +809,8 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
...
+	if _, err = c.mclient.MonitoringV1().Alertmanagers(a.Namespace).ApplyStatus(ctx, aac, metav1.ApplyOptions{}); err != nil {
 		return errors.Wrap(err, "failed to update status subresource")
```
simonpasquier, line 814:
```suggestion
		return errors.Wrap(err, "failed to apply status subresource")
```

pkg/prometheus/agent/operator.go
```diff
@@ -810,7 +811,8 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 		return errors.Wrap(err, "failed to get prometheus agent status")
 	}
 	p.Status = *pStatus
-	if _, err = c.mclient.MonitoringV1alpha1().PrometheusAgents(p.Namespace).UpdateStatus(ctx, p, metav1.UpdateOptions{}); err != nil {
+	paac := v1alpha1.PrometheusAgent("prometheusAgent", p.Namespace)
```
simonpasquier, line 814: Same here. Also IIUC the first argument should be the name of the PrometheusAgent object.

simonpasquier:
To avoid too many back-and-forth with the e2e tests you might want to modify only one controller (e.g. Alertmanager) first.

danielmellado (approved):
getting a bit cryptic with all these 'trac', 'aac' and so but they come from the applyconfiguration changes 🗡️

pkg/prometheus/agent/operator.go
marioferh: ok. Done in other PR

pkg/alertmanager/operator.go
```diff
@@ -1790,3 +1800,14 @@ func ListOptions(name string) metav1.ListOptions {
 func tlsAssetsSecretName(name string) string {
 	return fmt.Sprintf("%s-tls-assets", prefixedName(name))
 }
+
+func GetAppliedConfigFromCondition(c *monitoringv1.Condition) *monitoringv1ac.ConditionApplyConfiguration {
+	return &monitoringv1ac.ConditionApplyConfiguration{
```
simonpasquier, line 1805: (suggestion) let's use the helpers from the apply package.
```suggestion
	return monitoringv1ac.Condition().
		WithType(c.Type).
		WithStatus(c.Status).
		...
```

simonpasquier (commented):
The call to `stsReporter.Update()` at L806 also sets some of the status fields so we need to adapt it.

pkg/alertmanager/operator.go
marioferh: done

pkg/alertmanager/operator.go
```diff
@@ -808,8 +809,17 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	a.Status.Conditions = operator.UpdateConditions(a.Status.Conditions, availableCondition, reconciledCondition)
 	a.Status.Paused = a.Spec.Paused
```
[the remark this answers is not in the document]
marioferh: done

ArthurSens:
👋 -- Just out of curiosity, what errors are we seeing in UpdateStatus() that we're fixing here? :)

marioferh:
> 👋 -- Just out of curiosity, what errors are we seeing in UpdateStatus() that we're fixing here? :)

`	if _, err = c.mclient.MonitoringV1().Alertmanagers(a.Namespace).UpdateStatus(ctx, a, metav1.UpdateOptions{}); err != nil {
		return errors.Wrap(err, "failed to update status subresource")
	}`
	
Trying to update the operator status, the error: "failed to update status subresource" is shown and could be confused.

simonpasquier:
To add more context, we're seeing this type of messages in the OpenShift's prometheus operator logs (I've seen in plain Kind clusters too):

````
level=error ts=2023-09-11T07:11:06.75456597Z caller=klog.go:116 component=k8s_client_runtime func=ErrorDepth msg="status \"prominstancens-allowlist-s0t7xg-2-137e5aff/instance\" failed: faile
d to update prometheus status subresource: Operation cannot be fulfilled on prometheuses.monitoring.coreos.com \"instance\": the object has been modified; please apply your changes to the la
test version and try again"
````

My guess is that since UpdateStatus() reads the Prometheus object from the cache, it may have a slightly outdated version with an old `versionNumber`. While the status eventually converges (e.g. the operator will requeue the status update), it's somehow disturbing for users to see errors in the logs. Using an apply operation of update should get rid of this annoyance.

**Comment by ArthurSens** [comment]
> My guess is that since UpdateStatus() reads the Prometheus object from the cache, it may have a slightly outdated version with an old `versionNumber`. While the status eventually converges (e.g. the operator will requeue the status update), it's somehow disturbing for users to see errors in the logs. Using an apply operation of update should get rid of this annoyance.

Would it be possible to have the Status set to an older version if we use apply, even if for a little bit? 

I don't really see a problem with that, just thinking out loud at the moment, but do you see that as a problem?

**Comment by simonpasquier** [comment]
@ArthurSens my assumption is that the update will retrigger a status update if required. In most cases, it should be a no-op.

File: pkg/alertmanager/operator.go [hunk 6]
```diff
@@ -1790,3 +1797,28 @@ func ListOptions(name string) metav1.ListOptions {
...
+	return monitoringv1ac.Condition().
+		WithType(c.Type).
+		WithStatus(c.Status).
+		WithLastTransitionTime(c.LastTransitionTime).
+		WithReason(c.Reason).
+		WithMessage(c.Message).
+		WithObservedGeneration(c.ObservedGeneration)
+}
+
+func GetAppliedConfigStatusFromStatus(a *monitoringv1.Alertmanager, asac *monitoringv1ac.AlertmanagerStatusApplyConfiguration, values ...*monitoringv1ac.ConditionApplyConfiguration) {
```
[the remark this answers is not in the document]
marioferh: not sure if like this is best option because in GetAppliedConfigStatusFromStatus are monitoringv1ac and monitoringv1 values. Check the new changes please

pkg/alertmanager/operator.go
```diff
@@ -1790,3 +1796,25 @@ func ListOptions(name string) metav1.ListOptions {
...
+
+func GetAppliedConfigStatusFromStatus(a *monitoringv1.Alertmanager, asac *monitoringv1ac.AlertmanagerStatusApplyConfiguration, conditions ...monitoringv1.Condition) {
+	asac.WithPaused(a.Status.Paused)
+	asac.WithReplicas(a.Status.Replicas)
+	asac.WithAvailableReplicas(a.Status.AvailableReplicas)
+	asac.WithUpdatedReplicas(a.Status.UpdatedReplicas)
+	asac.WithUnavailableReplicas(a.Status.UnavailableReplicas)
+
+	for _, condition := range conditions {
+		asac.WithConditions(GetAppliedConfigFromCondition(&condition))
```
simonpasquier, line 1818: (nit) this is what I meant by getting rid of `GetAppliedConfigFromCondition`.
```suggestion
		asac.WithConditions(
			monitoringv1ac.Condition().
			WithType(c.Type).
			WithStatus(c.Status).
			WithLastTransitionTime(c.LastTransitionTime).
			WithReason(c.Reason).
			WithMessage(c.Message).
			WithObservedGeneration(c.ObservedGeneration),
		)		
```

pkg/alertmanager/operator.go
```diff
@@ -1790,3 +1796,25 @@ func ListOptions(name string) metav1.ListOptions {
...
+	return monitoringv1ac.Condition().
+		WithType(c.Type).
+		WithStatus(c.Status).
+		WithLastTransitionTime(c.LastTransitionTime).
+		WithReason(c.Reason).
+		WithMessage(c.Message).
+		WithObservedGeneration(c.ObservedGeneration)
+}
+
+func GetAppliedConfigStatusFromStatus(a *monitoringv1.Alertmanager, asac *monitoringv1ac.AlertmanagerStatusApplyConfiguration, conditions ...monitoringv1.Condition) {
```
simonpasquier, line 1810: Thinking more about it, we can have a single function that takes a monitoringv1.Alertmanager object with an up-to-date status field and returns an AlertmanagerApplyConfiguration object. This would minimize the changes in UpdateStatus().

```golang
func ApplyConfigurationFromAlertmanager(
    a *monitoringv1.Alertmanager
) monitoringv1ac.AlertmanagerApplyConfiguration {
    asac := monitoringv1ac.AlertmanagerStatus().
        WithPaused(a.Status.Paused).
        ...

    ...

    return monitoringv1ac.Alertmanager(a.Name, a.Namespace).WithStatus(asac)
}
```

marioferh: done, I think now is simpler and prettier

pkg/alertmanager/operator.go
```diff
@@ -53,7 +54,8 @@ import (
 )
 
 const (
-	resyncPeriod = 5 * time.Minute
+	resyncPeriod                   = 5 * time.Minute
+	prometheusOperatorFieldManager = "PrometheusOperator"
```
danielmellado, line 58: nit: It doesn't hurt, but if it has only one usage I wouldn't bother creating a var for it. Do you plan to use this in any follow-up PRs? Don't bother changing it back just because of this, just curious.

simonpasquier: The same change will be applied to the Prometheus and Thanos controllers so we'll need a constant in the end.

simonpasquier (commented):
Great! Can you proceed with the other UpdateStatus() functions?

simonpasquier:
Clarified offline with @marioferh. Let's merge this and PRs will follow for the other controllers.

simonpasquier (approved):
Thanks!

PR #5898, opened 2023-09-07
fix: reconcile on namespace changes

### LINKED ISSUE #3847: serviceMonitorNamespaceSelector not working correctly

simonpasquier:
## Description
When the operator was configured to select only a limited number of
namespaces, it would not watch for namespace changes. It means that the
operator may not reconcile when a namespace label is added/removed
(affecting which objects should be selected or not).

This change enables the operator to use a privileged namespace
lister/watcher whenever the service account has the needed permissions.

**IMPORTANT:** it also requires Kubernetes >= 1.22 to be effective but the
operator will degrade to the less efficient implementation if this condition isn't met.

xref #3847 (one more PR required to fix it completely).

## Type of change

## Changelog entry

```release-note
Reconcile monitoring resources on namespace updates.
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -303,6 +303,15 @@ func run() int {
 		return 1
 	}
 
+	kubernetesVersion, err := kclient.Discovery().ServerVersion()
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to request Kubernetes server version", "err", err)
+		cancel()
+		return 1
+	}
+	cfg.KubernetesVersion = *kubernetesVersion
+	level.Info(logger).Log("msg", "connection established", "cluster-version", cfg.KubernetesVersion)
+
 	scrapeConfigSupported, err := checkPrerequisites(
 		ctx,
 		logger,
--- a/pkg/listwatch/listwatch.go
+++ b/pkg/listwatch/listwatch.go
@@ -16,21 +16,27 @@ package listwatch
 
 import (
 	"context"
+	"fmt"
+	"sort"
 	"strings"
 
+	"github.com/blang/semver/v4"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/apimachinery/pkg/watch"
-	"k8s.io/client-go/kubernetes/scheme"
+	authv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
+
+	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
 )
 
-// NewUnprivilegedNamespaceListWatchFromClient mimics
+// NewNamespaceListWatchFromClient mimics
 // cache.NewListWatchFromClient. It allows for the creation of a
 // cache.ListWatch for namespaces from a client that does not have `List`
 // privileges. If the slice of namespaces contains only v1.NamespaceAll, then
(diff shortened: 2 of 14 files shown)


test/e2e/prometheus_instance_namespaces_test.go
```diff
@@ -302,11 +331,11 @@ func testPrometheusInstanceNamespacesAllowList(t *testing.T) {
 		// Remove the selecting label on the "allowed" namespace and check that
 		// the target is removed.
 		// See https://github.com/prometheus-operator/prometheus-operator/issues/3847
-		//if err := testFramework.RemoveLabelsFromNamespace(framework.KubeClient, allowedNs, "monitored"); err != nil {
+		//if err := framework.RemoveLabelsFromNamespace(context.Background(), allowedNs, "monitored"); err != nil {
```
simonpasquier, line 334: this part is still commented because the service account isn't configured with list/watch permissions on namespaces. I have another change in flight that would fix it by implementing a poller based watcher but to keep this PR reviewable, I'll send as a follow-up.

cmd/operator/main.go
```diff
@@ -303,6 +303,15 @@ func run() int {
 		return 1
 	}
 
+	kubernetesVersion, err := kclient.Discovery().ServerVersion()
```
simonpasquier, line 306: Instead of repeating the same check in each controller, let's do it once on startup.

test/e2e/prometheus_instance_namespaces_test.go
simonpasquier: https://github.com/prometheus-operator/prometheus-operator/pull/5925/commits/5ada40db1508fdb128fbd558c52bc10a1f92a33f

pkg/listwatch/listwatch.go
```diff
@@ -44,56 +50,144 @@ import (
...
 		}
+		if metadataNameLabelSupported {
+			// Using a label selector is more efficient but requires Kubernetes 1.22 at least.
+			tweak = func(options *metav1.ListOptions) {
+				TweakByLabel(options, "kubernetes.io/metadata.name", ExcludeFilterType, deniedNamespaces)
+			}
+		}
 
-		return cache.NewFilteredListWatchFromClient(c, "namespaces", metav1.NamespaceAll, tweak)
+		level.Debug(l).Log("msg", "using privileged namespace lister/watcher")
```
ArthurSens, line 108: this debug log seems out of place 🤔

pkg/listwatch/listwatch.go
```diff
@@ -44,56 +50,144 @@ import (
...
-				Resource("namespaces").
-				Name(name).
-				VersionedParams(&options, scheme.ParameterCodec).
-				Do(ctx).
-				Into(result)
-			if apierrors.IsNotFound(err) {
-				level.Info(l).Log("msg", "namespace not found", "namespace", name)
-				continue
-			}
+			result, err := corev1Client.Namespaces().Get(ctx, name, metav1.GetOptions{})
```
ArthurSens, line 164: Should we put this inside an if?

```go
if getAllowed {
  result, err := corev1Client.Namespaces().Get(ctx, name, metav1.GetOptions{})
  .
  .
  .
}
```

pkg/listwatch/listwatch.go
```diff
@@ -44,56 +50,144 @@ import (
...
+	if !getAllowed {
+		err := fmt.Errorf("missing permissions")
+		for _, r := range reasons {
+			err = fmt.Errorf("%w: %w", err, r)
+		}
+
+		level.Warn(l).Log("msg", "the operator lacks required permissions which may result in degraded functionalities", "err", err)
 	}
 
 	listFunc := func(options metav1.ListOptions) (runtime.Object, error) {
```
ArthurSens, line 161:
```suggestion
	listFunc := func(_ metav1.ListOptions) (runtime.Object, error) {
```

Is it intended to not use the options provided here?

ArthurSens (commented):
Sorry for taking this long to review, the subject here is a bit out of my comfort zone 😬 I'm still going through the codebase and trying to understand everything. 

A few doubts came up while reading the PR

pkg/listwatch/listwatch.go
simonpasquier: hmm not sure to understand but I can remove all the "using (un)privileged namespace lister/watcher" debug logs as they were more for me to check the code path.

pkg/listwatch/listwatch.go
simonpasquier: yes it's intended and the PR didn't change anything in that regard.

pkg/listwatch/listwatch.go
simonpasquier: I kept the existing behavior to minimize the changes. The SA might have get permissions on a subset of the configured namespaces but getAllowed will be false.

simonpasquier (commented):
Thanks for the review @ArthurSens. I understand that the change isn't easy to review, fortunately we have quite good e2e test coverage which helped me find bugs in the first iterations :)

simonpasquier:
@ArthurSens in case you missed my earlier comment: https://github.com/prometheus-operator/prometheus-operator/commit/5ada40db1508fdb128fbd558c52bc10a1f92a33f is the next step but I didn't want to increase the scope of this PR even more.

pkg/listwatch/listwatch.go
ArthurSens: It's fine to keep it! I missed the return literally in the next line and thought you left duplicated logs 😅

test/e2e/alertmanager_instance_namespaces_test.go
```diff
@@ -45,18 +45,28 @@ func testAlertmanagerInstanceNamespacesAllNs(t *testing.T) {
...
 
-	am := framework.MakeBasicAlertmanager(nonInstanceNs, "non-instance", 3)
+	am := framework.MakeBasicAlertmanager(nonInstanceNs, "non-instance", 2)
 	_, err = framework.MonClientV1.Alertmanagers(nonInstanceNs).Create(context.Background(), am, metav1.CreateOptions{})
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	am = framework.MakeBasicAlertmanager(instanceNs, "instance", 3)
+	am = framework.MakeBasicAlertmanager(instanceNs, "instance", 2)
```
ArthurSens, line 69: Any special reason to reduce the number of replicas?

I assume it is just because the test is valid for both 3 or 2 replicas and 2 replicas is just faster, but I wanted to ask anyway

ArthurSens (approved):
Alright, LGTM!

Thanks for working on this, reviewing it also helped me understand better what kubebuilder usually abstracts away from other projects I work with :P

ArthurSens:
> @ArthurSens in case you missed my earlier comment: [5ada40d](https://github.com/prometheus-operator/prometheus-operator/commit/5ada40db1508fdb128fbd558c52bc10a1f92a33f) is the next step but I didn't want to increase the scope of this PR even more.

Just saw it and looks 👌 !

test/e2e/alertmanager_instance_namespaces_test.go
simonpasquier: yes, the test doesn't care if it's 3, 2 or 1 replicas. But I'll revert it to avoid unrelated changes. We can have a look at all e2e tests globally and decide whether they really need multi replicas: in most cases, I think that the answer is no and switching to single replicas might save a few resources.

PR #5899, opened 2023-09-07
Feat: Add OwnerReferences field to EmbeddedPersistentVolumeClaim

wallee94:
This change allows adding OwnerReferences to the PersistentVolumeClaims created by the operator, so Kubernetes can remove the PVCs when the owners are deleted.

## Description

I have a use case at work where a Prometheus object is created for every namespace. We expect the operator to keep the PVCs even after the StatefulSet or Prometheus is deleted, as it currently does. Still, we want to be able to set the associated namespace as owner reference so the claims are removed if the namespace is deleted.
In this case, the PVCs live in a different namespace than the referenced.

## Type of change

## Changelog entry

```release-note
feat: Add OwnerReferences field to EmbeddedPersistentVolumeClaim
```


simonpasquier:
Thanks for the PR!

>  In this case, the PVCs live in a different namespace than the referenced.

Not sure to understand, what is "the referenced" here? Another namespace?

wallee94:
My bad. I'm re-reading this, and I forgot to mention an important detail. All our Prometheus objects and PVCs live in the same "monitoring" namespace.
Whenever someone adds a new namespace, we create a Prometheus to monitor it, but it's created in a second "monitoring" namespace. If the new namespace is deleted, the Prometheus and PVC objects are not. That's why we need to set an ownerReference to cascade the deletion.
This change might be related to [this issue](https://github.com/prometheus-operator/prometheus-operator/issues/1897).
It doesn't make the operator responsible for cleaning the PVCs but adds an option to set a reference for cleanup.

simonpasquier:
Thanks for the details, it's more clear now.

> If the new namespace is deleted, the Prometheus and PVC objects are not. That's why we need to set an ownerReference to cascade the deletion.

Don't you set the ownerReference of the Prometheus object to the "new" namespace?

IMHO the resolution of #1897 would be the proper way to handle it. [PersistentVolumeClaim retention](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention) is beta since 1.27, not sure when it's going to GA.

wallee94:
Thanks for the link to the retention doc!

You're right, it's different. For the issue, changing retention would work to remove them when the StatefulSet is deleted. In this case, I want to keep the PVCs as long as the namespace exists

I agree the operator probably shouldn't handle these deletions, and we should rely on the k8s interface, but ownerReference is a k8s interface 😕 

> Don't you set the ownerReference of the Prometheus object to the "new" namespace?

Yep, that's working: deleting the "new" namespace removes the dependent Prometheus but not the PVCs. I meant we need a ownerReference for both

simonpasquier:
Let me recap:
1. User creates a namespace "foo".
1. You create a Prometheus object with a PVC template in the existing "monitoring" namespace. The Prometheus object has a reference to the "foo" namespace.
1. When deleting the "foo" namespace, the Prometheus object is deleted by the garbage collector but not the PVCs.

I can think of several solutions to fix the "PVC doesn't get deleted when":
* Provide the ability to define ownerreferences in the VolumeClaim template (e.g. this PR).
  * Full flexibility but more complicated for the user.
* Copy the Prometheus ownerreferences to the VolumeClaim template (probably opt-in via a new field in the Prometheus CRD).
  * Simpler for the user.
* Support the `persistentVolumeClaimRetentionPolicy` field in the Prometheus CRD.
  * More aligned with the Statefulset semantics and future Kubernetes versions.
  * One issue though is that the operator may have to recreate the statefulset (e.g. when updating immutable fields) so it needs to be clever enough to revert to the Retain policy before doing so.

wallee94:
Yep, that sounds right. 

That's also the issue I see with using `persistentVolumeClaimRetentionPolicy`: the operator recreates the StatefulSet when we change the PVC's storage request. 

Coping the Prometheus ownerreferences sounds good to me. I could work on that change, but it also makes it easier to set a config that could make the user lose their data. For me, setting the ownerreferences in the PVC template makes sense because it's explicit, and it's just passing metadata to the claims, not so different to the fields it already has to set labels or annotations on the PVCs

wallee94:
Got it. Alright! Adding an opt-in field like that one to propagate the `ownerReferences` sounds good.
I can work on it next week. I'm closing this PR and will send a new one.
Thank you @simonpasquier !

PR #5904, opened 2023-09-10
Run tests only if go file changes

ArthurSens:
## Description

Making it easier to contribute to anything that is not go.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,6 +1,11 @@
 name: e2e
 on:
   pull_request:
+    paths:
+      - '**.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - '.github/workflows/e2e.yml'
   push:
     branches:
       - 'release-*'
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -1,6 +1,11 @@
 name: unit
 on:
   pull_request:
+    paths:
+      - '**.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - '.github/workflows/unit.yml'
   push:
     branches:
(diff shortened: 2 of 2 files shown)


ArthurSens:
I'm just not sure if making changes to jsonnet requires us to run tests or not 🤔 Maybe it could break e2e tests?

.github/workflows/e2e.yaml
```diff
@@ -8,6 +8,8 @@ on:
       - 'main'
     tags:
       - 'v*'
+    paths:
+      - '**.go'
```
simonpasquier, line 12: Will a change to the GitHub actions workflow trigger the e2e tests?

ArthurSens: good catch

ArthurSens: It should also be on pull_request, not on push

.github/workflows/e2e.yaml
```diff
@@ -1,6 +1,9 @@
 name: e2e
 on:
   pull_request:
+    paths:
+      - '**.go'
+      - '.github/workflows/e2e.yml'
```
simonpasquier, line 6: also go.{mod,sum} files?

ArthurSens: Only on the root level? Or all go.{mod|sum}?

simonpasquier: probably all files for safety.

ArthurSens:
Ah interesting, the go-related CI is required, making this change impossible 😅

PR #5919, opened 2023-09-19
Add controller-id proposal

danielmellado:
## Description
This PR adds a new proposal for PO regarding the usage of
`controller-id` for handling several instances in different namespaces
or rogue Prometheus' CRD usage.

## Type of change

### DIFF
--- a/Documentation/proposals/202309-controller-id.md
+++ b/Documentation/proposals/202309-controller-id.md
@@ -1,0 +1,79 @@
+# Controller-ID Proposal
+
+* Owners:
+  * [danielmellado](https://github.com/danielmellado)
+* Related Tickets:
+  * [#4281](https://github.com/prometheus-operator/prometheus-operator/issues/4281)
+  * [#4498](https://github.com/prometheus-operator/prometheus-operator/pull/4498)
+* Other Docs:
+  * n/a
+
+# Introduction
+
+This proposal aims to implement a solution to support multiple cluster-level
+Prometheus instances running concurrently without conflicting over the same
+custom resources. This solution isn't limited to the Prometheus resources as
+it'll also be available for `AlertManager`and `ThanosRuler` ones, as well as for
+any pod-based resource that could be added in the future.
+
+This issue can significantly impact use cases where multiple Prometheus operator
+instances run at the same time in the Kubernetes cluster.
+
+# Why
+
+Currently, we encounter issues when different users deploy different instances
+of the Prometheus operator, that will try to reconcile the same resources.
+
+In the worst-case scenario, these operators may not only compete for ownership
+of the CRD resources but also attempt to rewrite or redeploy different versions
+of the CRD, causing disruptions to all pods.
+
+The remediation for this scenario, where users deploy their Prometheus operator
+instances in parallel, involves using one of the many CLI arguments such as
+`--deny-namespaces`, `--namespaces`, `--prometheus-instance-selector`  or
+`prometheus-instance-namespaces`. But this requires cooperation between the
+different parties and there's no way to ensure that a specific monitoring
+resource is managed only by a specific operator instance.
+
+# How
+
+After some research from @machine424, we have identified a potential solution
+already implemented by the
+[zalando/postgres-operator](https://github.com/zalando/postgres-operator). When
+an operator is configured with a specific "controller ID" value, it will only
+reconcile resources that have a matching "controller ID" annotation.
+
+Conversely, if the operator is not configured with a "controller ID," it will
+skip all resources that have a "controller ID" annotation. More details can be
+found in the
+[zalando/postgres-operator documentation](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#operators-with-defined-ownership-of-certain-postgres-clusters).
+
+# Goals
+
+* Guarantee that a custom resource will be managed by a specific Prometheus
+  operator instance.
+
+## Audience
+
+This proposal is relevant to the following audience:
+
+* Users who provide Prometheus as a service and want to run multiple Prometheus
+  operator instances in different namespaces.
+* Users seeking to mitigate the impact of rogue Prometheus instances.
+
+# Non-Goals
+
+* Provide a solution that works with user intervantion. It'll require work from
+  the user deploying the operator and resources. (e.g. if the operator is
+  started without any specific argument, it'll attempt to reconcile all
+  resources in all namespaces).
+
+# Alternatives
+
+Although the initial discussion for this proposal considered adding an owner
+reference within the scope, the
+[ControllerRef](https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/controller-ref.md)
+model does not directly address this problem because the `ControllerRef`model
+solves the problem of controllers that fight over controlled objects due to
+overlapping selectors (e.g. a ReplicaSet fighting with a ReplicationController
+over Pods because both controllers have label selectors that match those Pods)


danielmellado:
@simonpasquier ^^ FYI, thanks!

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,65 @@
...
+* [#4281](https://github.com/prometheus-operator/prometheus-operator/issues/4281)
+* [#4498](https://github.com/prometheus-operator/prometheus-operator/pull/4498)
+
+## Other Docs:
+* n/a
+
+### Introduction
+
+This proposal aims to implement a solution to support multiple cluster-level
+Prometheus instances running concurrently without conflicting over the same CRD.
```
simonpasquier, line 16: (nit) CRD -> custom resources. CRD is a definition of a custom resource (e.g. the schema definition)

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,65 @@
...
+This issue can significantly impact use cases where multiple Prometheus operator
+instances run in parallel.
+
+## Why
+
+Currently, we encounter issues when users or different operators deploy Prometheus
+alongside Prometheus Operator's instance. Even though User-Managed Workloads
+(UWM) exists, some operators or users may deploy Prometheus directly from
+manifests, Helm charts, or other rogue operators, leading to the creation of
+'rogue' Prometheus instances alongside the operator's instance.
```
simonpasquier, line 26: I don't think that Prometheus instances deployed without the operator are an issue. I would simplify the paragraph saying that the core issue is when different users deploy different instances of the Prometheus operator which try to reconcile the same resources.

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,65 @@
...
+In the worst-case scenario, these operators may not only compete for ownership
+of the CRD resources but also attempt to rewrite or redeploy different versions
+of the CRD, causing disruptions to all pods.
+
+The remediation for this scenario, where users deploy their Prometheus operator
+instances in parallel, involves using one of the many CLI arguments, such as
+`--deny-namespaces`, `--namespaces`, `--prometheus-instance-selector`, or
+`prometheus-instance-namespaces`.
+
+Although the initial discussion for this proposal considered adding an owner
```
simonpasquier, line 37: should be listed in an alternatives section.

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,65 @@
...
+reconcile resources that have a matching "controller ID" annotation.
+
+Conversely, if the operator is not configured with a "controller ID," it will
+skip all resources that have a "controller ID" annotation. More details can be
+found in the
+[zalando/postgres-operator documentation](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#operators-with-defined-ownership-of-certain-postgres-clusters).
+
+## Goals
+
+* Add a similar "controller-id" based approach to PO.
```
simonpasquier, line 57: I'd rather phrase it as: "Guarantee that a custom resource will be managed by a specific Prometheus operator instance"

simonpasquier (commented):
Thanks for starting this!

ArthurSens (commented):
Thanks for the simple and clear write-up. Just addressing simon comments and it has my approval!

The implementation sounds simple enough to implement in a single PR, so I don't see a reason to add "Action plan" to the proposal :)

danielmellado:
Thanks @ArthurSens and @simonpasquier . I addressed your comments on the PR, should be good to go then ;)

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,65 @@
...
+* Users who provide Prometheus as a service and want to run multiple Prometheus
+  operator instances in different namespaces.
+* Users seeking to mitigate the impact of rogue Prometheus instances.
+
+# Alternatives
+
+Although the initial discussion for this proposal considered adding an owner
+reference within the scope, the
+[ControllerRef](https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/controller-ref.md)
+model does not directly address this problem.
```
simonpasquier, line 65:
```suggestion
model does not directly address this problem because the `ControllerRef` model solves the problem of controllers that fight over controlled objects due to overlapping selectors (e.g. a ReplicaSet fighting with a ReplicationController over Pods because both controllers have label selectors that match those Pods)
```

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,65 @@
...
+Conversely, if the operator is not configured with a "controller ID," it will
+skip all resources that have a "controller ID" annotation. More details can be
+found in the
+[zalando/postgres-operator documentation](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#operators-with-defined-ownership-of-certain-postgres-clusters).
+
+# Goals
+
+* Guarantee that a custom resource will be managed by a specific Prometheus
+  operator instance.
+
```
simonpasquier, line 51: For non-goals, I'd add
*  it's not going to be an zero-configuration solution (e.g. it still requires work from the user deploying the operator + resource).
* this will be an opt-in feature

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,65 @@
...
+of the Prometheus operator, that will try to reconcile the same resources.
+
+In the worst-case scenario, these operators may not only compete for ownership
+of the CRD resources but also attempt to rewrite or redeploy different versions
+of the CRD, causing disruptions to all pods.
+
+The remediation for this scenario, where users deploy their Prometheus operator
+instances in parallel, involves using one of the many CLI arguments, such as
+`--deny-namespaces`, `--namespaces`, `--prometheus-instance-selector`, or
+`prometheus-instance-namespaces`.
```
simonpasquier, line 32:
```suggestion
`prometheus-instance-namespaces` But this requires cooperation between the different parties and there's no way to ensure that a specific monitoring resource is managed only by a specific operator instance.
```

simonpasquier (commented):
Just a few nits but lgtm overall.

Documentation/proposals/202309-controller-id.md
ArthurSens: You mean "This will not be an opt-in feature"?

danielmellado: I assume that Simon means that users would be able to not trigger the feature? Not sure why they wouldn't like to but just in case they would opt-out the config?

simonpasquier: I mean to say that when you start the operator without any specific argument, it will attempt to reconcile all resources in all namespaces. But I realize that it's similar to the first non-goal.

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,76 @@
...
+  * [#4281](https://github.com/prometheus-operator/prometheus-operator/issues/4281)
+  * [#4498](https://github.com/prometheus-operator/prometheus-operator/pull/4498)
+* Other Docs:
+  * n/a
+
+# Introduction
+
+This proposal aims to implement a solution to support multiple cluster-level
+Prometheus instances running concurrently without conflicting over the same
+custom resources.
```
simonpasquier, line 15: This is probably obvious for all of us but the solution isn't going to be limited to the Prometheus resources and will also be available for Alertmanager and ThanosRuler (and any pod-based resource we may add in the future).

Documentation/proposals/202309-controller-id.md
danielmellado: I'll remove the 2nd and put a clarification on the first one.

Documentation/proposals/202309-controller-id.md
danielmellado: I'll add a note about that, thanks!

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,79 @@
...
+
+This proposal is relevant to the following audience:
+
+* Users who provide Prometheus as a service and want to run multiple Prometheus
+  operator instances in different namespaces.
+* Users seeking to mitigate the impact of rogue Prometheus instances.
+
+# Non-Goals
+
+* It's not going to be a zero-configuration solution, as it'll require work from
```
simonpasquier, line 66: (nit)
```suggestion
* Provide a solution that works with user intervention. It'll require additional configuration from
```

danielmellado:
> 2 final nits and this is good to go!

thanks!

Documentation/proposals/202309-controller-id.md
```diff
@@ -0,0 +1,79 @@
...
+The remediation for this scenario, where users deploy their Prometheus operator
+instances in parallel, involves using one of the many CLI arguments such as
+`--deny-namespaces`, `--namespaces`, `--prometheus-instance-selector`  or
+`prometheus-instance-namespaces`. But this requires cooperation between the
+different parties and there's no way to ensure that a specific monitoring
+resource is managed only by a specific operator instance.
+
+# How
+
+After some research, we have identified a potential solution already
```
simonpasquier, line 40: I forgot that we should give credit to @machine424 for the suggestion :)
```suggestion
After some research from @machine424, we have identified a potential solution already
```

danielmellado: Totally agree, adding him back there, wasn't sure whether it should be on the proposal.

simonpasquier (approved):
lgtm. @ArthurSens if you're fine with it too, you can merge.

PR #5926, opened 2023-09-19
Fix prom upgrade test

ArthurSens:
## Description

Fixing the test that I broke with #5732 😬 

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/test-prom-version-upgrade.yaml
+++ b/.github/workflows/test-prom-version-upgrade.yaml
@@ -1,11 +1,11 @@
-name: e2e
+name: Test Prometheus upgrades
 on:
   schedule:
     - cron:  '37 15 * * *' # Every day 15:37
 
 jobs:
-  e2e-tests:
-    name: E2E tests
+  upgrade-prometheus:
+    name: Upgrade Prometheus
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


ArthurSens:
The last commit needs to be dropped before merging

simonpasquier (approved):
lgtm

ArthurSens:
Hmmmm, still getting rate limited :/

I'll add an extra commit decreasing the frequency we check for ready status

simonpasquier:
The `client rate limiter Wait returned an error: context deadline exceeded` message is a bit misleading: in fact, it tells that the wait function timed out without success.

ArthurSens:
> The `client rate limiter Wait returned an error: context deadline exceeded` message is a bit misleading: in fact, it tells that the wait function timed out without success.

Thanks for the pointer! Indeed the message got me in the wrong direction

After some debugging I've realized that `t.Parallel()` in `testPromVersionMigration` would unblock the deferred function `testCtx.Cleanup(t)`. The cleanup deletes the prometheus-operator, therefore the Prometheus objects would never get reconciled.

Since this test suite only has one single test, I just removed t.Parallel() and now it is working :)

simonpasquier (approved):

Issue #4682, opened 2022-03-25
Get rid of `github.com/pkg/errors` dependency


simonpasquier:
https://github.com/pkg/errors has been archived and is no longer maintained. 

~~At least 2 options are available:~~
* ~~Use a drop-in replacement like https://github.com/emperror/errors.~~
- [ ] Use the [`errors`](https://pkg.go.dev/errors) package from the standard Go library.
- [x] Add a `depguard` rule in the golang-cilint configuration to prevent future use of `github.com/pkg/errors`.

See also https://github.com/thanos-io/thanos/issues/5176

simonpasquier:
Personally I'd be fine with using the standard `errors` package since the code doesn't rely on advanced capabilities of `github.com/pkg/errors` such as stack trace logging.

slashpai:
will do this

PhilipGough:
@slashpai - thanks, its yours :)

slashpai:
Not stale I am working on this

slashpai:
@simonpasquier  I was wondering if we want to consider https://github.com/cockroachdb/errors as replacement to github.com/pkg/errors

chiukapoor:
Hello Team, Is anyone working on this issue? If not, may I work on it?

slashpai:
@chiukapoor please feel free to work on :)

chiukapoor:
@slashpai may you please assign the issue to me

slashpai:
> @slashpai may you please assign the issue to me

Generally commenting on issue is needed to start working on :). I can assign it to you though

simonpasquier:
@chiukapoor please submit as many PRs as you wish. In general, smaller changes are easier to review and faster being merged :)

simonpasquier:
Hmm shouldn't we keep the issue opened until all occurrences of github.com/pkg/errors have been replaced?

ArthurSens:
Oh yes, my mistake

PR #5930, opened 2023-09-20
Enable depguard linter against github.com/pkg/errors

ArthurSens:
## Description

@simonpasquier @slashpai, I was thinking about adding https://github.com/fatih/faillint to our CI to make sure we're not using certain functions/packages. 

This could help us to fix https://github.com/prometheus-operator/prometheus-operator/issues/4682, and to make sure we don't re-introduce unwanted packages in the future

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/alertmanager/validation/v1alpha1/validation.go
+++ b/pkg/alertmanager/validation/v1alpha1/validation.go
@@ -15,12 +15,11 @@
 package v1alpha1
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"regexp"
 	"strings"
-
-	"github.com/pkg/errors"
 
 	"github.com/prometheus-operator/prometheus-operator/pkg/alertmanager/validation"
 	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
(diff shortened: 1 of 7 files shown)


ArthurSens:
Now I just need to fix all those occurrences 😭

simonpasquier:
Can't we use [depguard](https://golangci-lint.run/usage/linters/#depguard) which is available from golangci-lint?

ArthurSens:
> Can't we use [depguard](https://golangci-lint.run/usage/linters/#depguard) which is available from golangci-lint?

Oh you even mentioned that in the issue and I missed 😬...

I guess depguard will catch a lot more problems, since I'm using faillint to catch only Wrap/Wrapf 😬

ArthurSens:
I wonder if there is a way to make this one easier to review. 

Can golangci-lint only warn, but not fail, for a particular linter?

I feel like changing 700+ lines in a single PR is super easy to accidentally break some logs

simonpasquier:
I agree that we need an incremental approach.

> Can golangci-lint only warn, but not fail, for a particular linter?

It looks like we can add  the enforcement  gradually by running it only for specific paths:
https://golangci-lint.run/usage/linters/#depguard

pkg/alertmanager/validation/v1alpha1/validation.go
```diff
@@ -208,7 +207,7 @@ func validateEmailConfig(configs []monitoringv1alpha1.EmailConfig) error {
 		if config.Smarthost != "" {
 			_, _, err := net.SplitHostPort(config.Smarthost)
 			if err != nil {
-				return errors.Wrapf(err, "invalid field 'smarthost': %s", config.Smarthost)
+				return fmt.Errorf("invalid 'smarthost' %s: %w", config.Smarthost, err)
```
ArthurSens, line 210: This is the only logline that I made a significant change. It already had `:`, so I re-arranged it.

(same in v1beta1)

ArthurSens:
Okeee, started with 2 packages only. Will continue in other PRs.

pkg/alertmanager/validation/v1alpha1/validation.go
simonpasquier: this looks good to me.

simonpasquier (approved):

PR #5932, opened 2023-09-21
feat: add TLS and basic authentication for config reloader

### LINKED ISSUE #5836: Add basic auth to prometheus-config-reloader
**What did you do?**
I want to add basic auth to prometheus-config-reloader.
I try to add add basic auth via *--web.config.file* like prometheus, but get "unknown long flag '--web.config.file'"
**Environment**

Operator version is 0.57.0.
configmap-reload version is :v0.5.0
    
    

    `kubectl version`
    

* Kubernetes cluster kind:

* Manifests:

```
```

* Prometheus Operator Logs:

```
```

dongjiang1989:
## Description

Add TLS and basic authentication for config reloader.
```
--web-config-file=<string> 
        Path to configuration file that can enable TLS or authentication. See: https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md
```

fix #5836

## Type of change

## Changelog entry

```release-note
Add TLS and basic authentication for config reloader.
```


dongjiang1989:
@simonpasquier PTAL!
/cc @BrucePeng92

BrucePeng92:
LGTM, Thank you for your work!

cmd/prometheus-config-reloader/main.go
```diff
@@ -91,6 +97,12 @@ func main() {
 
 	versionutil.RegisterIntoKingpinFlags(app)
 
+	err := web.Validate(*webConfig)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Unable to validate web configuration file. err: "+err.Error())
```
ArthurSens, line 102: Could we use `level.Error(logger....` instead?

cmd/prometheus-config-reloader/main.go
```diff
@@ -158,11 +170,15 @@ func main() {
...
-		srv := http.Server{Addr: *listenAddress}
+		srv := &http.Server{}
 
 		g.Add(func() error {
 			level.Info(logger).Log("msg", "Starting web server for metrics", "listen", *listenAddress)
-			return srv.ListenAndServe()
+			return web.ListenAndServe(srv, &web.FlagConfig{
+				WebListenAddresses: &[]string{*listenAddress},
+				WebConfigFile:      webConfig,
+				WebSystemdSocket:   func() *bool { b := false; return &b }(),
```
ArthurSens, line 180: Sorry, I'm not familiar with the expoter toolkit package

What is WebSystemdSocket used for? Why do we always return false?

dongjiang1989: `WebSystemdSocket` use systemd socket activation listeners instead of port listeners, but is only supported on Linux.
So, set the default `*false` here for different OS.

cmd/prometheus-config-reloader/main.go
dongjiang1989: If `webConfig` is configured incorrectly, http.Server will fail to start. Similar to `app.Parse` and `version verification`, `os.Stderr` output is more appropriate。

```
	err := web.Validate(*webConfig)
	if err != nil {
		fmt.Fprintln(os.Stderr, "Unable to validate web configuration file. err: "+err.Error())
		os.Exit(2)
	}

	if _, err := app.Parse(os.Args[1:]); err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(2)
	}

	if versionutil.ShouldPrintVersion() {
		versionutil.Print(os.Stdout, "prometheus-config-reloader")
		os.Exit(0)
	}

	logger, err := logging.NewLogger(*logLevel, *logFormat)
	if err != nil {
		stdlog.Fatal(err)
	}
```

dongjiang1989:
> Sorry for taking this long to take a look at this one, I still haven't found time for a proper review, but so far it's looking good
> 
> 
> 
> One doubt came up:

@ArthurSens PTAL recheck. Thanks

cmd/prometheus-config-reloader/main.go
```diff
@@ -76,6 +77,11 @@ func main() {
 		"address on which to expose metrics (disabled when empty)").
 		String()
 
+	webConfig := app.Flag(
+		"web-config-file",
+		"Path to configuration file that can enable TLS or authentication. See: https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md",
```
simonpasquier, line 82: (nit) might be better to link to https://prometheus.io/docs/prometheus/latest/configuration/https/

cmd/prometheus-config-reloader/main.go
simonpasquier: Hmm we can leave it to nil too, no?

simonpasquier (commented):
Hmm I'm wondering if it should be an setting from the operator arguments or a new field in the Alertmanager/Prometheus/PrometheusAgent CRDs. I'd lean towards the second option because it's something that is under the responsibility of the person deploying the custom resource who doesn't necessarily operate the operator.

simonpasquier:
I'd also recommend splitting the PR into 2:
1. Add the new flag to the config-reloader binary
2. Update the operator to make use of it.

cmd/prometheus-config-reloader/main.go
dongjiang1989: > Hmm we can leave it to nil too, no?
Yes. It can be set to nil.

cmd/prometheus-config-reloader/main.go
dongjiang1989: Thanks you for your review. Done

dongjiang1989:
> I'd also recommend splitting the PR into 2:
> 
> 1. Add the new flag to the config-reloader binary
> 2. Update the operator to make use of it.

Thanks you for your review. Next, I split this PR into 2 PR submissions ✍️

cmd/prometheus-config-reloader/main.go
rexagod, line 102: `level` outputs all strings to `os.Stdout`, currently. Moving to `os.Stderr` may break this assumption for folks who use these logs in one way or the other (eg., redirecting `>` to a file) but don't actively redirect the latter to the former (`2>&1`), since this has been the behavior throughout the codebase. Nonetheless, even if we decide to direct errors to `os.Stderr`, that should be done for all `level.Error` cases as done above to adhere to the standards being introduced here.

To be explicit, I'm -1 on this since this will break levelled logging for us. Folks can still grep all error logs from `os.Stdout` and redirect those to `os.Stderr` if they want to.

rexagod (changes requested):

cmd/prometheus-config-reloader/main.go
simonpasquier: To be fair, we're not really consistent: if an error happens before setting up the log struct, it's logged to stderr.

I don't feel strongly about this but I think that we should be consistent and eventually redirect everything to stdout. So for this PR, I'd use `level.Info(log)`.

dongjiang1989:
split this PR into 2 PR submissions:
1. Add the new flag to the config-reloader: https://github.com/prometheus-operator/prometheus-operator/pull/6011
2. Add the operator crds to make use of config-reloader web-config-file : https://github.com/prometheus-operator/prometheus-operator/pull/6194

so. close this pr frist.

PR #5933, opened 2023-09-21
fix: get IMPERSONATE_USER from current environment

simonpasquier:
## Description

This change allows to run the operator locally as a different user:

```
export IMPERSONATE_USER=system:serviceaccount:default:system
./run/run-external.sh -c
```

## Type of change

_not classifying it as a bug fix because it's not relevant for end users._

## Changelog entry

```release-note

```


philipgough (approved):
lgtm

Issue #5936, opened 2023-09-22
Issues with ScrapeConfigCRD consulSDConfigs


heliapb:
Hello there, I was trying to test moving away from using additionalScrapeConfigs, to replace with ScrapeConfig instead, but got some errors while trying to create the first one on Prometheus Operator. I use consul for service discovery with additionalScrapeConfigs for context, and creating the ScrapeConfig for the service resulted in errors in prometheus-operator, and no service appeared in Prometheus Targets.

```
github.com/prometheus-operator/prometheus-operator/pkg/prometheus.(*ResourceSelector).validateConsulSDConfigs(0xc000316d20, {0x2604410, 0xc000130190}, 0xc000845b80)
	/home/runner/work/prometheus-operator/prometheus-operator/pkg/prometheus/resource_selector.go:756 +0x376
github.com/prometheus-operator/prometheus-operator/pkg/prometheus.(*ResourceSelector).SelectScrapeConfigs(0xc000316d20, {0x2604410, 0xc000130190}, 0xc00111b3b8)
	/home/runner/work/prometheus-operator/prometheus-operator/pkg/prometheus/resource_selector.go:713 +0xc9c
github.com/prometheus-operator/prometheus-operator/pkg/prometheus/server.(*Operator).createOrUpdateConfigurationSecret(0xc000a50000, {0x2604410, 0xc000130190}, 0xc000127100, 0x0?, {0xc0013cc5c0, 0x1, 0x1}, 0xc001300d80)
	/home/runner/work/prometheus-operator/prometheus-operator/pkg/prometheus/server/operator.go:1531 +0x3a5
github.com/prometheus-operator/prometheus-operator/pkg/prometheus/server.(*Operator).sync(0xc000a50000, {0x2604410, 0xc000130190}, {0xc000b9f980, 0x30})
	/home/runner/work/prometheus-operator/prometheus-operator/pkg/prometheus/server/operator.go:1201 +0x6be
github.com/prometheus-operator/prometheus-operator/pkg/prometheus/server.(*Operator).Sync(0xc000a50000, {0x2604410?, 0xc000130190?}, {0xc000b9f980, 0x30})
	/home/runner/work/prometheus-operator/prometheus-operator/pkg/prometheus/server/operator.go:1157 +0x32
github.com/prometheus-operator/prometheus-operator/pkg/operator.(*ResourceReconciler).processNextReconcileItem(0xc000a1d2b0, {0x2604410, 0xc000130190})
[...]

PR #5938, opened 2023-09-22
fix: Update ec2 validate func

slashpai:
If secret reference is nil it will cause operator to crash, hence updating the validate function for ec2 sd config to check
for secret only if provided in config.

Noticed this when https://github.com/prometheus-operator/prometheus-operator/issues/5936 logged. Since this change is not released yet we don't need to put in changelog I guess as bugfix

## Description

## Type of change

## Changelog entry

```release-note

```


simonpasquier (approved):

simonpasquier:
Thanks!

PR #5941, opened 2023-09-25
add: scrapeConfigSelector check

### LINKED ISSUE #5928: Specifying only scrapeConfigSelector breaks the deployment of Prometheus
in Docs there is this statement for `scrapeConfiigSelector`:

> Experimental ScrapeConfigs to be selected for target discovery. An empty label selector matches all objects. A null label selector matches no objects.
> 
> If spec.serviceMonitorSelector, spec.podMonitorSelector, spec.probeSelector and spec.scrapeConfigSelector are null, the Prometheus configuration is unmanaged. The Prometheus operator will ensure that the Prometheus configuration’s Secret exists, but it is the responsibility of the user to provide the raw gzipped Prometheus configuration under the prometheus.yaml.gz key. This behavior is deprecated and will be removed in the next major version of the custom resource definition. It is recommended to use spec.additionalScrapeConfigs instead.

This is currently not true because `spec.scrapeConfigSelector` is not checked in code see here:

https://github.com/prometheus-operator/prometheus-operator/blob/b8d05eb743dd4c9c73681e598d14a92347954025/pkg/prometheus/server/operator.go#L1407

I run today into it because I only use the instance for federation and do not have defined any of "spec.serviceMonitorSelector, spec.podMonitorSelector, spec.probeSelector" but only "spec.scrapeConfigSelector" and the operator does not bring up an instance of prometheus because of this restriction.

nidhey27:
## Description

The current Prometheus Operator implementation checks for the presence of spec.serviceMonitorSelector, spec.podMonitorSelector, and spec.probeSelector to determine if the Prometheus configuration should be treated as unmanaged. However, the documentation specifies that the operator should also check for spec.scrapeConfigSelector. This PR aligns the implementation with the documentation by adding the necessary check for spec.scrapeConfigSelector.

Closes #5928

## Type of change

## Changelog entry

```release-note
Ensure that the operator checks for spec.scrapeConfigSelector when determining if the Prometheus configuration should be treated as unmanaged.
```

### DIFF
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -1404,7 +1404,8 @@ func logDeprecatedFields(logger log.Logger, p *monitoringv1.Prometheus) {
 		}
 	}
 
-	if p.Spec.ServiceMonitorSelector == nil && p.Spec.PodMonitorSelector == nil && p.Spec.ProbeSelector == nil {
+	if p.Spec.ServiceMonitorSelector == nil && p.Spec.PodMonitorSelector == nil && p.Spec.ProbeSelector == nil && p.Spec.ScrapeConfigSelector == nil {
+
 		level.Warn(logger).Log("msg", "neither serviceMonitorSelector nor podMonitorSelector, nor probeSelector specified. Custom configuration is deprecated, use additionalScrapeConfigs instead")
 	}
 }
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
Thanks!

PR #5945, opened 2023-09-25
fix: add --cluster.label to alertmanager

slashpai:
If multiple Alertmanager clusters are deployed on the same cluster, it can happen
that because pod IP addresses are recycled, an Alertmanager instance from cluster B
connects with cluster A.

 `--cluster.label`  flag was introduced in alertmanager `v0.26`, this helps to block
any traffic that is not meant for the cluster.

The value is hardcoded and the value is guaranteed to be unique in a given cluster but
if there's a use case, we can consider a new field in the CRD.

This commit sets default `--cluster.label` flag as namespace/name
when Alertmanager version used is >= v0.26

## Description

## Type of change

## Changelog entry

```release-note
add --cluster.label to alertmanager which prevents alertmanager pod joining cluster not expected
```

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -306,6 +306,17 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 		amArgs = append(amArgs, fmt.Sprintf("--cluster.peer-timeout=%s", a.Spec.ClusterPeerTimeout))
 	}
 
+	// If multiple Alertmanager clusters are deployed on the same cluster, it can happen
+	// that because pod IP addresses are recycled, an Alertmanager instance from cluster B
+	// connects with cluster A.
+	// --cluster.label flag was introduced in alertmanager v0.26, this helps to block
+	// any traffic that is not meant for the cluster.
+	// The value is hardcoded and the value is guaranteed to be unique in a given cluster but
+	// if there's a use case, we can consider a new field in the CRD.
+	if version.GTE(semver.MustParse("0.26.0")) {
+		amArgs = append(amArgs, fmt.Sprintf("--cluster.label=%s/%s", a.Namespace, a.Name))
+	}
+
 	isHTTPS := a.Spec.Web != nil && a.Spec.Web.TLSConfig != nil && version.GTE(semver.MustParse("0.22.0"))
 
 	livenessProbeHandler := v1.ProbeHandler{
--- a/pkg/alertmanager/statefulset_test.go
+++ b/pkg/alertmanager/statefulset_test.go
@@ -1186,6 +1186,48 @@ func TestAutomountServiceAccountToken(t *testing.T) {
 		}
 	}
 }
+
+func TestClusterLabel(t *testing.T) {
+	tt := []struct {
+		scenario                string
+		version                 string
+		expectedClusterLabelArg bool
+	}{{
+		scenario:                "--cluster.label set by default for version >= v0.26.0",
+		version:                 "0.26.0",
+		expectedClusterLabelArg: true,
+	}, {
+		scenario:                "no --cluster.label set for older versions",
+		version:                 "0.25.0",
+		expectedClusterLabelArg: false,
+	}}
+
+	for _, ts := range tt {
+		t.Run(ts.scenario, func(t *testing.T) {
+			a := monitoringv1.Alertmanager{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "alertmanager",
+					Namespace: "monitoring",
+				},
+				Spec: monitoringv1.AlertmanagerSpec{
+					Replicas: toPtr(int32(1)),
+					Version:  ts.version,
+				},
+			}
+
+			ss, err := makeStatefulSetSpec(&a, defaultTestConfig, nil)
+			require.NoError(t, err)
+
+			args := ss.Template.Spec.Containers[0].Args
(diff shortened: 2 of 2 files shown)


simonpasquier:
small nit: this is more a bug fix for me since it can avoid an Alertmanager pod from joining another cluster than expected.

pkg/alertmanager/statefulset_test.go
```diff
@@ -1186,6 +1186,42 @@ func TestAutomountServiceAccountToken(t *testing.T) {
...
+			a.Name = "alertmanager"
+			a.Spec.Version = ts.version
+
+			ss, err := makeStatefulSetSpec(&a, defaultTestConfig, nil)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			args := ss.Template.Spec.Containers[0].Args
+			if got := slices.ContainsFunc(args, containsString(ts.expectClusterLabelArg)); !got {
```
simonpasquier, line 1218: can we use `require.Contains()` or `require.Condition()`?

pkg/alertmanager/statefulset_test.go
```diff
@@ -1186,6 +1186,42 @@ func TestAutomountServiceAccountToken(t *testing.T) {
...
+		scenario              string
+		version               string
+		expectClusterLabelArg string
+	}{{
+		scenario:              "--cluster.label set by default for version >= v0.26.0",
+		version:               "0.26.0",
+		expectClusterLabelArg: "--cluster.label=monitoring/alertmanager",
+	}, {
+		scenario: "no --cluster.label set for older versions",
+		version:  "0.25.0",
```
simonpasquier, line 1201: this doesn't check that the arguments list doesn't contain `--cluster.label`?

pkg/alertmanager/statefulset_test.go
```diff
@@ -1186,6 +1186,42 @@ func TestAutomountServiceAccountToken(t *testing.T) {
...
+	for _, ts := range tt {
+		t.Run(ts.scenario, func(t *testing.T) {
+			a := monitoringv1.Alertmanager{}
+			a.Spec.Replicas = toPtr(int32(1))
+			a.Namespace = "monitoring"
+			a.Name = "alertmanager"
+			a.Spec.Version = ts.version
+
+			ss, err := makeStatefulSetSpec(&a, defaultTestConfig, nil)
+			if err != nil {
```
simonpasquier, line 1213: (nit)
```suggestion
			require.NoError(t, err)
```

pkg/alertmanager/statefulset_test.go
```diff
@@ -1186,6 +1186,42 @@ func TestAutomountServiceAccountToken(t *testing.T) {
...
+		version:               "0.26.0",
+		expectClusterLabelArg: "--cluster.label=monitoring/alertmanager",
+	}, {
+		scenario: "no --cluster.label set for older versions",
+		version:  "0.25.0",
+	}}
+
+	for _, ts := range tt {
+		t.Run(ts.scenario, func(t *testing.T) {
+			a := monitoringv1.Alertmanager{}
```
simonpasquier, line 1206: (nit) easier to read IMHO if you have a full definition of the variable

```suggestion
			a := monitoringv1.Alertmanager{
				ObjectMeta: metav1.ObjectMeta{
					Name: "alertamanager",
					...
			}
```

simonpasquier:
i've found a few references in the Alertmanager issues:
* https://github.com/prometheus/alertmanager/issues/1835 
* https://github.com/prometheus/alertmanager/issues/2250

slashpai:
@simonpasquier  addressed comments

pkg/alertmanager/statefulset.go
```diff
@@ -306,6 +306,13 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 		amArgs = append(amArgs, fmt.Sprintf("--cluster.peer-timeout=%s", a.Spec.ClusterPeerTimeout))
 	}
 
+	// --cluster.label flag was introduced in alertmanager v0.26, this helps to block any traffic that
+	// is not meant for the cluster.
+	// The value is hardcoded but f there's a use case, we can consider a new field in the CRD.
```
simonpasquier, line 311:
```suggestion
	// The value is hardcoded and the value is guaranteed to be unique in a given cluster but if there's a use case, we can consider a new field in the CRD.
```

pkg/alertmanager/statefulset.go
```diff
@@ -306,6 +306,13 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 		amArgs = append(amArgs, fmt.Sprintf("--cluster.peer-timeout=%s", a.Spec.ClusterPeerTimeout))
 	}
 
+	// --cluster.label flag was introduced in alertmanager v0.26, this helps to block any traffic that
```
simonpasquier, line 309: (nit) let's add more context about the issue being fixed.
```suggestion
	// If multiple Alertmanager clusters are deployed on the same cluster, it can happen that because pod IP addresses are recycled, an Alertmanager instance from cluster B connects with cluster A.
	// --cluster.label flag was introduced in alertmanager v0.26 and prevents an instance to connect to other cluster's member unless they share the same label.
```

simonpasquier (commented):
just a few comments to clarify what we want to prevent.

slashpai:
@simonpasquier  updated

simonpasquier (approved):
Thanks!

Issue #5338, opened 2023-02-14
Add OCI label(s) to docker images


james-callahan:
**What is missing?**

The docker image label `org.opencontainers.image.source`

**Why do we need it?**

I'd like to take advantage of renovate's changelog support (https://docs.renovatebot.com/modules/datasource/docker/)

To do so, the recommended way is to add the standard OCI label `org.opencontainers.image.source` to docker images with a link to the github project.
A full list of pre-defined annotations is available at: https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

**Environment**

N/A

simonpasquier:
Sounds good to me :) I guess that the other predefined OCI annotations would be good to have too.
Do you want to send a PR?

ksankeerth:
Hi @simonpasquier,

I'm interested in contributing  to this issue. I could add all the annotations mentioned in https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys. If I explain my approach, We can have  a shell script to extract the information needed for labels and generate docker files.  What do you think? 

Thanks,
Sankeerthan

chiukapoor:
Hello @simonpasquier, May you please assign the issue to me?

In brief, we have to update the Dockerfile with predefined OCI annotations. Do we need to follow an automated way to do so or should we just manually update it?

simonpasquier:
Updating the Dockerfile manually is fine by me.

PR #5946, opened 2023-09-25
Fixing #5338 | Add OCI labels in the Dockerfile

chiukapoor:
## Description

* Fixing #5338
* Add OCI labels in the Dockerfile

## Type of change

## Changelog entry

```release-note
Add OCI labels in the Dockerfile
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,4 +7,9 @@ COPY operator /bin/operator
 # On busybox 'nobody' has uid `65534'
 USER 65534
 
+LABEL org.opencontainers.image.source="https://github.com/prometheus-operator/prometheus-operator" \
+    org.opencontainers.image.url="https://prometheus-operator.dev/" \
+    org.opencontainers.image.documentation="https://prometheus-operator.dev/" \
+    org.opencontainers.image.licenses="Apache-2.0"
+
 ENTRYPOINT ["/bin/operator"]


chiukapoor:
@simonpasquier PTAL

Dockerfile
```diff
@@ -7,4 +7,8 @@ COPY operator /bin/operator
 # On busybox 'nobody' has uid `65534'
 USER 65534
 
+LABEL org.opencontainers.image.source="https://github.com/prometheus-operator/prometheus-operator" \
+    org.opencontainers.image.documentation="https://prometheus-operator.dev/docs/prologue/introduction/" \
```
simonpasquier, line 11: (nit) to avoid HTTP 404 replies if the page moves in the future.  
```suggestion
    org.opencontainers.image.documentation="https://prometheus-operator.dev/" \
```

Dockerfile
```diff
@@ -7,4 +7,8 @@ COPY operator /bin/operator
 # On busybox 'nobody' has uid `65534'
 USER 65534
 
+LABEL org.opencontainers.image.source="https://github.com/prometheus-operator/prometheus-operator" \
```
simonpasquier, line 10: Maybe we should add `org.opencontainers.image.url` too?

Dockerfile
chiukapoor: Updated :+1:

Dockerfile
chiukapoor: Added :+1:

simonpasquier (approved):
Thanks!

PR #5954, opened 2023-10-02
fixes race condition when deleting alertmanager, prometheus and thanosruler instances

mheers:
## Description

A race condition in appears when using [Pulumi](https://www.pulumi.com/) to delete a prometheus instance custom resouce from a Kubernetes cluster.

I've created a demo repo https://github.com/mheers/pulumi-prometheus-operator-race-condition to reproduce the behaviour.

Deleting the CR with pulumi will get the operator stuck in a loop. The operator will try to delete the prometheus instance but will fail with errors exported to https://github.com/mheers/pulumi-prometheus-operator-race-condition/blob/main/logs/operator.log

## Type of change

## Changelog entry

```release-note
Fixes race condition when deleting alertmanager or prometheus instances
```

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -640,12 +640,12 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("failed to set Alertmanager type information: %w", err)
 	}
 
-	if am.Spec.Paused {
+	// Check if the Alertmanager instance is marked for deletion.
+	if c.rr.DeletionInProgress(am) {
 		return nil
 	}
 
-	// Check if the Alertmanager instance is marked for deletion.
-	if !am.ObjectMeta.DeletionTimestamp.IsZero() {
+	if am.Spec.Paused {
 		return nil
 	}
 
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -644,6 +644,11 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return nil
 	}
 
+	// Check if the Alertmanager instance is marked for deletion.
+	if !am.ObjectMeta.DeletionTimestamp.IsZero() {
+		return nil
+	}
+
 	logger := log.With(c.logger, "key", key)
 	logDeprecatedFields(logger, am)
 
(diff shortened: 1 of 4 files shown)


simonpasquier:
Thanks for the PR! For completeness can you share the operator's version where the issue happens?

mheers:
I've tested and experienced it on
- 0e3614cf613d2f652e72066cbfde66f0e77b7842
- 0.68.0
- 0.67.0

pkg/alertmanager/operator.go
```diff
@@ -644,6 +644,11 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return nil
 	}
 
+	// Check if the Alertmanager instance is marked for deletion.
```
simonpasquier, line 647: I'd suggest to get inspiration from the UpdateStatus() method and do the check at the very beginning of the function.

https://github.com/prometheus-operator/prometheus-operator/blob/535cd8cf87670b812a12b4fb500ec92d06689877/pkg/alertmanager/operator.go#L782-L789

simonpasquier (commented):
Ideally we'd need the same for pkg/prometheus/agent/operator.go and pkg/thanos/operator.go.

pkg/alertmanager/operator.go
mheers: Thanks for your comment. I've added thanos and the agent, reused the DeletionInProgress function and moved the checks as high as possible.

pkg/prometheus/agent/operator.go
```diff
@@ -514,6 +514,13 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	}
 
 	logger := log.With(c.logger, "key", key)
+
+	// Check if the Agent instance is marked for deletion.
+	if c.rr.DeletionInProgress(p) {
+		level.Info(logger).Log("msg", "the resource is deleting, not reconciling")
```
simonpasquier, line 520: (nit) `c.rr.DeletionInProgress` already logs (though at debug level)
```suggestion
```

pkg/prometheus/server/operator.go
```diff
@@ -1166,6 +1166,12 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	logger := log.With(c.logger, "key", key)
 	logDeprecatedFields(logger, p)
 
+	// Check if the Prometheus instance is marked for deletion.
+	if c.rr.DeletionInProgress(p) {
+		level.Info(logger).Log("msg", "the resource is deleting, not reconciling")
```
simonpasquier, line 1171: (nit) same
```suggestion
```

PR #5961, opened 2023-10-03
Add proposal for Shard Autoscaling

ArthurSens:
## Description

Add proposal for automated sharding

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/Documentation/proposals/202310-shard-autoscaling.md
+++ b/Documentation/proposals/202310-shard-autoscaling.md
@@ -107,7 +107,7 @@ spec:
 
 Inspired by the [StatefulSet API](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention), when setting `.shardRetentionPolicy.whenScaled` is set to `Delete` the operator would simply delete the underlying StatefulSet on the spot, not caring about the data stored in that particular shard. Although not the safest approach, it is exactly the current behavior, and changing it might surprise a lot of users.
 
-If `.shardRetentionPolicy.whenScaled` is set to `Retain` Prometheus-Operator will keep the shard running for a customizable period of time. By default, the Operator will keep the Shard for the whole retention period, or forever if only size-based retention is defined. Alternatively, the user can choose a customized retention period.
+If `.shardRetentionPolicy.whenScaled` is set to `Retain` Prometheus-Operator will keep the shard running for a customizable period of time. By default, the Operator will keep the Shard for the whole retention period, or forever if only size-based retention is defined. Alternatively, the user can define a custom retention period.
 
 If a retention time is defined, the deletion logic is controlled by an annotation present on the StatefulSet:
 
--- a/Documentation/proposals/202310-shard-autoscaling.md
+++ b/Documentation/proposals/202310-shard-autoscaling.md
@@ -117,7 +117,7 @@ operator.prometheus.io/deletion-timestamp: X
 
 On scale downs, the configuration of all shards would be re-arranged to make sure the Prometheus that was scaled down doesn't scrape any metrics anymore, but it would still be available for queries.
 
-When the current time finally surpasses the one in the annotation, the Operator will finally trigger the deletion of the statefulset associated with the shard.
+When the deletion timestamp set in the annotation is reached, the Operator will trigger the deletion of the statefulset associated with the shard.
 
 ### Scaling up after scaling down
 
--- a/Documentation/proposals/202310-shard-autoscaling.md
+++ b/Documentation/proposals/202310-shard-autoscaling.md
@@ -1,0 +1,165 @@
+# Shard Autoscaling
+
+* Owners:
+  * [Arthur Sens](https://github.com/ArthurSens)
+  * [Nicolas Takashi](https://github.com/nicolastakashi)
+* Related Tickets:
+  * [#4927](https://github.com/prometheus-operator/prometheus-operator/issues/4727)
+  * [#4946](https://github.com/prometheus-operator/prometheus-operator/issues/4946)
+  * [#4967](https://github.com/prometheus-operator/prometheus-operator/issues/4967)
+* Other docs:
+  * n/a
+
+This document aims to address existing issues preventing users from leveraging the Horizontal Pod Autoscaler (HPA) in conjunction with Prometheus sharding.
+
+* [Why](#why)
+* [Pitfalls of the current solution](#pitfalls-of-the-current-solution)
+* [Goals](#goals)
+* [Non-Goals](#non-goals)
+* [Audience](#audience)
+* [How](#how)
+  * [Scale subresource](#scale-subresource)
+  * [Graceful scale-down of Prometheus Servers](#graceful-scale-down-of-prometheus-servers)
+    * [Scaling up after scaling down](#scaling-up-after-scaling-down)
+  * [Graceful scale-down of Prometheus Agents](#graceful-scale-down-of-prometheus-agents)
+* [Scale-down alternatives](#scale-down-alternatives)
+* [Action Plan](#action-plan)
+
+# Why
+
+Managing Prometheus instances with highly variable workloads can be quite challenging. When the number of scraped targets can scale up and down between a few items to thousands over a short period of time (e.g. within a day), operators are forced to opt for the safest scenario, ending up with over-provisioned instances. Additionally, the WAL replay often becomes a problem after a restart as it takes a long time to be replayed into memory. Without available replicas, this leads to unnecessary downtime.
+
+To address these issues, this document outlines a solution to enhance the Prometheus Operator by enabling the usage of Horizontal Pod Autoscalers (HPAs) to scale shards up and down. This improvement aims to simplify the process of scaling Prometheus instances, making it more accessible and efficient for operators.
+
+# Pitfalls of the current solution
+
+Prometheus can scale vertically pretty well, but instances that grow up to hundreds of GiB/TiB in WAL size start face several challenges:
+
+* Eventual cardinality explosions have a big blast radius if a single Prometheus instance is responsible for scraping the majority of monitored applications.
+* Recovering from a crash takes several minutes due to the WAL replay.
+
+While the Prometheus and PrometheusAgent CRDs implement sharding for scrape configurations, there are still pending issues and missing features that prevent adoption.
+
+* Environments where the load is undistributed throughout the day/week, operators need to constantly adjust number of shards or shard size.
+* Scaling down Prometheus Servers causes data loss, as there's no guarantee that stored data is persisted in centralized storage before the container is deleted.
+
+# Goals
+
+* Enable scaling of the Prometheus shards up and down via Horizontal Pod Autoscaler objects.
+
+# Non-Goals
+
+* Implement the autoscaling logic into the Prometheus operator.
+* Automatic balancing of targets to ensure homogeneous load across all Prometheus pods.
+
+# Audience
+
+* Users looking for a shard autoscaling mechanism with a good cost-reliability ratio.
+
+# How
+
+Today, there are a few strategies to measure the load of Prometheus instances.
+
+1. Perhaps you can use [Kubernetes Metrics API](https://kubernetes.io/docs/tasks/debug/debug-cluster/resource-metrics-pipeline/#metrics-api) to measure load in terms of CPU and/or Memory usage.
+
+2. Another strategy might be to measure the amount of samples ingested per second
+
+- Prometheus Agents: `rate(prometheus_agent_samples_appended_total[5m])`
+- Prometheus Servers: `rate(prometheus_tsdb_head_samples_appended_total[5m])`
+
+3. Or even measuring samples queried per second for Prometheus instances under heavy query load.
+
+- Prometheus Servers: `rate(prometheus_engine_query_samples_total[5m])`
+
+These are all data that could be configured as input for different Horizontal Pod Autoscalers, but how does HPAs knows if the amount of shard is bigger/lower than desired?
+
+## Scale subresource
+
+When working with any resource, including CRDs, HPAs depends on a subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). This proposal suggests to implement the scale subresource for the Prometheus and PrometheusAgent CRDs. Instead of working on the "replicas" count, it will operate on the "shards" count because the purpose of scaling up (resp. down) is to distribute the same number of targets across more (resp. less) Prometheus instances.
+
+With only this change, Prometheus Agents can already be horizontally scaled without problems, but for Prometheus Servers it gets a little more complicated.
+
+## Graceful scale-down of Prometheus Servers
+
+Prometheus Servers don't behave like the usual stateless applications. A few cases require Prometheus pods to still be available even after a scale-down event is triggered:
+
+* Thanos proxying queries back to the sidecars to retrieve data that hasn't been uploaded to object storage yet.
+* Prometheus servers with long retention periods still hold data that users might want to query.
+
+If the operator deletes Prometheus pods in excess during a scale-down operation, data loss will be a common problem, so a more complex strategy needs to be used here.
+
+We propose that Prometheus-Operator introduces a new `shardRetentionPolicy` field to the Prometheus CRD:
+
+```yaml
+apiVersion: monitoring.coreos.com
+kind: Prometheus
+metadata:
+  name: example
+spec:
+  shardRetentionPolicy:
+    # Default: Delete
+    whenScaled: Retain|Delete 
+    # if not specified, the operator picks up the retention period if it's defined,
+    # otherwise (e.g. size-based retention) the statefulset lives forever.
+    retain:
+      retentionPeriod: 3d
+```
+
+Inspired by the [StatefulSet API](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention), when setting `.shardRetentionPolicy.whenScaled` is set to `Delete` the operator would simply delete the underlying StatefulSet on the spot, not caring about the data stored in that particular shard. Although not the safest approach, it is exactly the current behavior, and changing it might surprise a lot of users.
+
+If `.shardRetentionPolicy.whenScaled` is set to `Retain` Prometheus-Operator will keep the shard running for a customizable period of time. By default, the Operator will keep the Shard for the whole retention period, or forever if only size-based retention is defined. Alternatively, the user can choose a customized retention period.
+
+If a retention time is defined, the deletion logic is controlled by an annotation present on the StatefulSet:
+
+```yaml
+operator.prometheus.io/deletion-timestamp: X
+```
+
+On scale downs, the configuration of all shards would be re-arranged to make sure the Prometheus that was scaled down doesn't scrape any metrics anymore, but it would still be available for queries.
+
+When the current time finally surpasses the one in the annotation, the Operator will finally trigger the deletion of the statefulset associated with the shard.
+
+### Scaling up after scaling down
+
+If a scale-up event happens while Prometheus Servers that were marked for deletion are still running, the already running Prometheus will be unmarked and the configuration of the shard pool will be reloaded as usual.
+
+We intentionally don't want to spin up new instances while others that are marked for deletion are still running as shard names and containers' environment variables that configure each shard's target group will be too complex to coordinate.
+
+## Graceful scale-down of Prometheus Agents
+
+Prometheus Agents are different than servers since queries are not available in this mode. Their only responsibility is scraping metrics and pushing them via remote-write to a long-term storage backend, making the scale-down experience much easier to handle.
+
+When receiving the SIGTERM signal, the Prometheus Agent should gracefully handle the signal by finishing all remote-write queues before ending the process. Prometheus-Operator, by default, adjusts the [Graceful Termination Period](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) of Prometheus/PrometheusAgent pods to 600s. Ten minutes should be enough for them to flush the remote-write queue, but, if needed, users can redefine Graceful Termination Period using [Strategic Merge Patch](https://prometheus-operator.dev/docs/operator/strategic-merge-patch/).
+
+Since there's no use case for retaining Prometheus Agents, its CRD will not be extended with the `RetentionPolicy` mentioned in [Graceful scale-down of Prometheus Servers](#graceful-scale-down-of-prometheus-servers)
+
+# Scale down Alternatives
+
+## Dump & Backfill
+
+During scale-down, the Prometheus-Operator could read TSDB Blocks from the Prometheus instance being deleted and backfill it into another instance.
+
+***Advantages:***
+* Prometheus can be shut down immediately without data loss.
+
+***Disadvantages:***
+* Loading TSDB Blocks into memory is expensive, requiring Prometheus-Operator to run with big Memory/CPU requests.
+* Complex and hard to coordinate workflow. (Require restarts to reload TSDB blocks, hard to identify possible corruptions)
+
+## Snapshot & Upload on shutdown
+
+During scale-down, Prometheus-Operator could send an HTTP request to [Prometheus' Snapshot endpoint](https://prometheus.io/docs/prometheus/latest/querying/api/#snapshot). Thanos sidecar could be extended to watch the snapshot directory and automatically upload snapshots to Object storage.
+
+There is a related issue open already: https://github.com/thanos-io/thanos/issues/6263
+
+***Advantages:***
+* Prometheus can be shut down immediately without data loss.
+
+***Disadvantages:***
+* Prometheus running without Thanos sidecars won't benefit from this strategy.
+
+# Action Plan
+
+1. Re-implement [Pull Request #4735](https://github.com/prometheus-operator/prometheus-operator/pull/4735), but this time focusing on Shards and making sure we add the [selectorpath to the scale subresource](https://book.kubebuilder.io/reference/generating-crd.html#scale). Enabling horizontal autoscaling for Prometheus Agents.
+
+2. Implement the graceful shutdown strategy of Prometheus servers.
--- a/Documentation/proposals/202310-shard-autoscaling.md
+++ b/Documentation/proposals/202310-shard-autoscaling.md
@@ -75,7 +75,7 @@ These are all data that could be configured as input for different Horizontal Po
 
 ## Scale subresource
 
-When working with any resource, including CRDs, HPAs depends on a subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). This proposal suggests to implement the scale subresource for the Prometheus and PrometheusAgent CRDs. Instead of working on the "replicas" count, it will operate on the "shards" count because the purpose of scaling up (resp. down) is to distribute the same number of targets across more (resp. less) Prometheus instances.
+When working with any resource, including CRDs, HPAs depend on a subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). This proposal suggests to implement the scale subresource for the Prometheus and PrometheusAgent CRDs. Instead of working on the "replicas" count, it will operate on the "shards" count because the purpose of scaling up (resp. down) is to distribute the same number of targets across more (resp. less) Prometheus instances.
 
 With only this change, Prometheus Agents can already be horizontally scaled without problems, but for Prometheus Servers it gets a little more complicated.
 
--- a/Documentation/proposals/202310-shard-autoscaling.md
+++ b/Documentation/proposals/202310-shard-autoscaling.md
@@ -115,7 +115,7 @@ If a retention time is defined, the deletion logic is controlled by an annotatio
 operator.prometheus.io/deletion-timestamp: X
 ```
 
-On scale downs, the configuration of all shards would be re-arranged to make sure the Prometheus that was scaled down doesn't scrape any metrics anymore, but it would still be available for queries.
+On scale downs, the configuration of all shards will be re-arranged to make sure that the "scaled-down" Prometheus pods don't scrape targets anymore but they will still be available for queries.
 
 When the deletion timestamp set in the annotation is reached, the Operator will trigger the deletion of the statefulset associated with the shard.
 
--- a/Documentation/proposals/202310-shard-autoscaling.md
+++ b/Documentation/proposals/202310-shard-autoscaling.md
@@ -71,7 +71,7 @@ Today, there are a few strategies to measure the load of Prometheus instances.
 
 - Prometheus Servers: `rate(prometheus_engine_query_samples_total[5m])`
 
-These are all data that could be configured as input for different Horizontal Pod Autoscalers, but how does HPAs knows if the amount of shard is bigger/lower than desired?
+These are all data that could be configured as input for different Horizontal Pod Autoscalers, but how do HPAs know if the amount of shard is bigger/lower than desired?
 
 ## Scale subresource
 
--- a/Documentation/proposals/202310-shard-autoscaling.md
+++ b/Documentation/proposals/202310-shard-autoscaling.md
@@ -121,7 +121,7 @@ When the deletion timestamp set in the annotation is reached, the Operator will 
 
 ### Scaling up after scaling down
 
-If a scale-up event happens while Prometheus Servers that were marked for deletion are still running, the already running Prometheus will be unmarked and the configuration of the shard pool will be reloaded as usual.
+If a scale-up event happens while the Prometheus pods marked for deletion are still running, the annotation will be removed and the scrape targets will be redistributed to all active shards.
 
 We intentionally don't want to spin up new instances while others that are marked for deletion are still running as shard names and containers' environment variables that configure each shard's target group will be too complex to coordinate.
 


ArthurSens:
@nicolastakashi your review would also be nice

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+  * [#4946](https://github.com/prometheus-operator/prometheus-operator/issues/4946)
+  * [#4967](https://github.com/prometheus-operator/prometheus-operator/issues/4967)
+* Other docs:
+  * n/a
+
+This document aims to solve many small problems to unblock usage of Horizontal Pod Autoscalers with Prometheus sharding mechanism.
+
+# Why
+
+When operating hundreds/thousands of Prometheus instances that have very non-standardized loads, it becomes troublesome to fine-tune all of them. While several horizontal pod autoscalers have shown up in the market, they don't really work with Prometheus because they cause metrics duplication issues. Prometheus-Operator offers a sharding mechanism to solve this problem. Still, they also fall short when samples ingested per second change alongside the load on the monitored systems.
```
nicolastakashi, line 17: I guess worth to mention this don't work because HPA use to scale the number of replicas from a Deployment/Statefulset and this will only increase reliability from Prometheus POV and not balance the load since all replicas will scrape all the same targets.

To solve that issue Prometheus should use a shard strategy, and then we can point to the shard document which explains Prometheus Operator sharding strategy.

wdyt?

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+We propose that Prometheus-Operator introduces a new field to the Prometheus Server CRDs:
+
+```yaml
+apiVersion: monitoring.coreos.com
+kind: Prometheus
+metadata:
+  name: example
+spec:
+  autoscaling:
+    scaleDownStrategy: force|preventDataLoss # Default: force
```
nicolastakashi, line 85: I'd suggest `graceful` instead of `preventDataLoss` and I'm bit torn between `immediate` over `force`.

wdyt?

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+If scaleDownStrategy == force, the operator would simply delete the pod on the spot, not caring about the data stored in that particular shard. Although not the safest approach, it is exactly the current behavior nowadays and changing it might surprise a lot of users.
+
+If scaleDownStrategy == preventDataLoss then Prometheus-Operator would first mark the pod with an annotation:
+
+```yaml
+prometheus.operator.io/safe-deletion-timestamp: X
```
nicolastakashi, line 93: We could get some inspiration on kubectl and have something like `grace-period` this is the name of a existing flag when running kubectl delete pod.

`prometheus.operator.io/grace-period`

wdyt?

Documentation/proposals/202310-automated-sharding.md
ArthurSens: It is mentioned in "Pitfalls of current solution".  Did you miss it or you prefer that this is mentioned in the "Why" section?

Documentation/proposals/202310-automated-sharding.md
ArthurSens: Agreed, sounds better!

Documentation/proposals/202310-automated-sharding.md
ArthurSens: It's a bit different, grace period is a duration, while here we set an explicit Unix timestamp.

I'm still open for a better name, I just think that it should make it clear that it is indeed a timestamp and not a duration

nicolastakashi: Hmmm 🤔 
Why not a duration, i think it would be even more clear.
prometheus.operator.io/grace-period: 2h

Means Prometheus Operator will wait 2h after scale down signal to delete this instance.

No?

ArthurSens: what if Prometheus-Operator restarts? will it wait for another 2 hours?

nicolastakashi: We need in someway keep the state.
Because the timestamp will be different from time to time right.
I mean users will need to update the timestamp.
Or I'm missing something?

ArthurSens: I don't think a state is needed here, the implementation is much more complex.

We can have a Unix timestamp and if current time > annotation then we trigger the deletion 🤔 

Imagine the following situations:

* When setting the annotation:

```go
retention := prom.args["--storage.tsdb.retention.time"]
prom.annotations["prometheus.operator.io/safe-deletion-timestamp"] = time.Now().Add(retention).Unix()
```

* When deciding if it needs to be deleted
```go
deletionTs := time.Parse(prom.annotations["prometheus.operator.io/safe-deletion-timestamp"])

if time.Now() > deletionTs {
  // delete prom
}
```

--- 
Of course the code here is very superficial, but the logic would be something like that

nicolastakashi: Maybe I miss understand this.
Users will not provide that annotation, we're going to do all the automation.
So it will be always calculated by the Operator.

is that right?

ArthurSens: > Maybe I miss understand this. Users will not provide that annotation, we're going to do all the automation. So it will be always calculated by the Operator.
> 
> is that right?

correct!

nicolastakashi: Ok!
I got it.
I could suggest still prometheus.operator.io/grace-period-timestamp 😄

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
+# Automated Sharding
```
nicolastakashi, line 1: I'm thinking about!
Maybe we should switch to Sharding AutoScaling.
Looking to Automated sharding, seems we're going to automate the sharding decision for users.
Like automatically see their instances are reaching a specific size and shard it out-of-box.

wdyt?

Documentation/proposals/202310-automated-sharding.md
nicolastakashi: My feeling when I read the why is we're already introducing the Pitfalls.
Wdyt about this.

> Managing large Prometheus instances with highly variable workloads can be quite challenging. These instances often demand significant computing resources like memory and CPU. In many cases, the required resources exceed what operators have at their disposal. Additionally, these large Prometheus instances can take a long time to become operational after a restart due to the substantial Write-Ahead Log (WAL), which takes time to be replayed in memory.
>
> To address these issues, this document outlines a solution to enhance the Prometheus Operator's Sharding feature by automating horizontal scaling. This improvement aims to simplify the process of scaling Prometheus instances, making it more accessible and efficient for operators.

So we can left the Pitfalls explaining what it's doing currently.

Documentation/proposals/202310-automated-sharding.md
xiu:
```suggestion
operator.prometheus.io/safe-deletion-timestamp: X
```
Regarding the naming, grace-period feels slightly off to me as I would expect that to be a duration. How about "scheduled-deletion-timestamp"? This would mark the intent better imo.

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+
+If scaleDownStrategy == force, the operator would simply delete the pod on the spot, not caring about the data stored in that particular shard. Although not the safest approach, it is exactly the current behavior nowadays and changing it might surprise a lot of users.
+
+If scaleDownStrategy == preventDataLoss then Prometheus-Operator would first mark the pod with an annotation:
+
+```yaml
+prometheus.operator.io/safe-deletion-timestamp: X
+```
+
+Where X is the timestamp of the current retention period.
```
xiu, line 96: If we have a sidecar shipping blocks away, do we want to wait for the prometheus retention period still or do we want to wait for the youngest block at the time of the scale down to be confirmed shipped?

Note: such a feature could be a later improvement if needed, not needed for the first version imo

xiu (commented):
Quick pass, I like it!

Documentation/proposals/202310-automated-sharding.md
nicolastakashi: Having the Sidecar uploading the blocks created by the snapshot feature, we can shutdown the Prometheus straight the way and we don't need to wait the deletion period.

Since it requires a implementation on Thanos I also think this is improving version

Documentation/proposals/202310-automated-sharding.md
nicolastakashi: Seems cool also

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+### Graceful scale down of Prometheus Servers
+
+Prometheus Servers don't behave like the usual stateless applications. A few cases require Prometheus pods to still be available even after a scale down event is triggered:
+
+* Thanos proxying queries back to the sidecars to retrieve data that still wasn't uploaded to object storage.
+* Prometheus servers with long retention periods still hold data that users might want to query.
+
+If we simply delete Prometheus pods during scale-down, data loss will be a common problem, so a more complex scale-down strategy needs to be used here.
+
+We propose that Prometheus-Operator introduces a new field to the Prometheus Server CRDs:
```
simonpasquier, line 76: StatefulSet got a new `persistentVolumeClaimRetentionPolicy` field which is currently alpha (https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention). Maybe our API should get inspiration from it?

````
spec:
  shardRetentionPolicy:
    whenScaled: Delete|Retain
````

File: Documentation/proposals/202310-automated-sharding.md (hunk 3 above)
Comment (simonpasquier) [reply Documentation/proposals/202310-automated-sharding.md:93]: Agreed that it should a point in time in the future and not a period.

(bike-shedding) what about `operator.prometheus.io/deletion-timestamp: 2023-10-05T08:18:09+00:00Z`?

File: Documentation/proposals/202310-automated-sharding.md (hunk 5 above)
Comment (simonpasquier) [inline Documentation/proposals/202310-automated-sharding.md:96]: Does it make sense to allow for user customization?

```
spec:
  shardRetentionPolicy:
    whenScaled: Retain
    # if not specified, the operator picks up the retention period if it's defined,
    # otherwise (e.g. size-based retention) the statefulset lives forever.
    retain:
      retentionPeriod: 3d
```

File: Documentation/proposals/202310-automated-sharding.md [hunk 7]
```diff
@@ -0,0 +1,130 @@
...
+
+```yaml
+prometheus.operator.io/safe-deletion-timestamp: X
+```
+
+Where X is the timestamp of the current retention period.
+
+The configuration of all shard would be re-arranged, to make sure the Prometheus that was scaled down do not scrapes any metrics anymore, but it would still be available for queries.
+
+When the current time finally surpasses the one in the annotation, the Operator will finally trigger its deletion.
```
simonpasquier, line 100:
```suggestion
When the current time finally surpasses the one in the annotation, the Operator will finally trigger the deletion of the statefulset associated with the shard.
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+metadata:
+  name: example
+spec:
+  autoscaling:
+    scaleDownStrategy: force|preventDataLoss # Default: force
+```
+
+If scaleDownStrategy == force, the operator would simply delete the pod on the spot, not caring about the data stored in that particular shard. Although not the safest approach, it is exactly the current behavior nowadays and changing it might surprise a lot of users.
+
+If scaleDownStrategy == preventDataLoss then Prometheus-Operator would first mark the pod with an annotation:
```
simonpasquier, line 90: (nit) the annotation should be on the statefulset?

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+```yaml
+prometheus.operator.io/safe-deletion-timestamp: X
+```
+
+Where X is the timestamp of the current retention period.
+
+The configuration of all shard would be re-arranged, to make sure the Prometheus that was scaled down do not scrapes any metrics anymore, but it would still be available for queries.
+
+When the current time finally surpasses the one in the annotation, the Operator will finally trigger its deletion.
+
```
simonpasquier, line 101: For completeness, can we describe what happens if the Prometheus object is scaled up before hitting the deletion timestamp? E.g. the targets are reallocated to the shard and the operator removes the annotation.

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+### Dump & Backfill
+
+During scale-down, Prometheus-Operator could read TSDB Blocks from the Prometheus instance being deleted and backfill it into another instance.
+
+***Advantages:***
+* Prometheus can be shutdown immediately without data loss.
+
+***Disadvantages:***
+* Loading TSDB Blocks into memory is expensive, requiring Prometheus-Operator to run with big Memory/CPU requests.
```
simonpasquier, line 112: Would the TSDB blocks be loaded by the operator or the Prometheus?

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+# Scale down Alternatives
+
+### Dump & Backfill
+
+During scale-down, Prometheus-Operator could read TSDB Blocks from the Prometheus instance being deleted and backfill it into another instance.
+
+***Advantages:***
+* Prometheus can be shutdown immediately without data loss.
+
+***Disadvantages:***
```
simonpasquier, line 111: Other drawbacks:
* Complex and hard to coordinate workflow (what do if the copy fails? how do you know that no corruption happens? need for restarting the remaining pods).

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+* Loading TSDB Blocks into memory is expensive, requiring Prometheus-Operator to run with big Memory/CPU requests.
+
+### Snapshot & Upload on shutdown
```
simonpasquier, line 114: I think that it might be a viable alternative for folks using Prometheus agent + Thanos as their main metrics storage.

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Scale Status subresource
+
+When working with CRDs, common HPAs (e.g. Keda) depends on a Status subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). Unfortunately, Prometheus-Operator resources still don't implement this field. There was an attempt to implement this in https://github.com/prometheus-operator/prometheus-operator/pull/4735, but was reverted because the PR was scaling replicas instead of shards. Although scaling replicas can help with High-Availability, it is expensive and hard to manage since it duplicates scrapes while not reducing the load on top of Prometheus because all replicas still scrape the same targets. Sharding serves as a better autoscaling strategy since the load is spread into all instances instead of getting duplicated.
+
+With only this change, Prometheus Agents can already be horizontally scaled without problems, but for Prometheus Servers it gets a little more complicated.
```
simonpasquier, line 65: I think that we still need to account for proper shutdown of the agents and ensure that the deletion happens only after the agent has forwarded all metrics to the remote write endpoints. It might be worth having a dedicated section on it. WDYT?

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,130 @@
...
+
+# Why
+
+When operating hundreds/thousands of Prometheus instances that have very non-standardized loads, it becomes troublesome to fine-tune all of them. While several horizontal pod autoscalers have shown up in the market, they don't really work with Prometheus because they cause metrics duplication issues. Prometheus-Operator offers a sharding mechanism to solve this problem. Still, they also fall short when samples ingested per second change alongside the load on the monitored systems.
+
+For teams that manage large scale of Prometheus instances, automated will help them find a good cost-reliability ratio while eliminating manual effort of fine-tuning several Prometheus several times a day.
+
+# Pitfalls of the current solution
+
+Prometheus can scale vertically pretty well, but instances that grow up to hundreds of GiB in WAL size start to get troublesome:
```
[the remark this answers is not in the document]
nicolastakashi: Which can lead us in the future have some kind of Prometheus VPA 😄

Documentation/proposals/202310-automated-sharding.md
nicolastakashi, line 96: What will happen if the node is drained before the deletion-timestamp?
The operator will recreate the pod and ensure this is only be deleted when the deletion-timestamp matches?

Documentation/proposals/202310-automated-sharding.md
nicolastakashi: Prometheus Server with Thanos Sidecar as well no?

simonpasquier: You mean Thanos sidecar with object upload?

Documentation/proposals/202310-automated-sharding.md
simonpasquier: Good point! If the annotation is on the statefulset, it's a bit less problematic but it's still an issue if someone deletes the statefulset. As a first approach, I'd be ok with admitting that a manual deletion of the statefulset voids the automatic shard scaling.

Documentation/proposals/202310-automated-sharding.md
nicolastakashi: Yup

simonpasquier: I'm not familiar enough but you might still want to keep the pods around for the retention period in case you have local alerting?

Documentation/proposals/202310-automated-sharding.md
ArthurSens: I like @simonpasquier's idea of having a default value (retention period) but allowing users to customize the retention to different periods, e.g. 2h.

Also mentioned by simon, I believe we can't assume it is okay to delete after blocks are uploaded because the user can still depend on that data for alerting purposes

Documentation/proposals/202310-automated-sharding.md
ArthurSens: Nice, sounds perfect!

Documentation/proposals/202310-automated-sharding.md
ArthurSens: I believe the operator since it can easily identify all Prometheus endpoints? (it has the RBAC permissions)

Thinking again about this strategy, this could be done in batches to reduce resource consumption 🤔

Documentation/proposals/202310-automated-sharding.md
ArthurSens: I had the impression that Prometheus Agent would still try to finish the remote-write queue when receiving SIGTERM. I'll need to investigate

Documentation/proposals/202310-automated-sharding.md
ArthurSens: Statefulsets won't be deleted on pod eviction, the only way is if someone manually deletes them. If someone does that, that goes beyond our control 😅

ArthurSens:
Thanks everyone for all the reviews so far! I've addressed as many comments as I could, ready for another round :)

Documentation/proposals/202310-automated-sharding.md
xiu: > Also mentioned by simon, I believe we can't assume it is okay to delete after blocks are uploaded because the user can still depend on that data for alerting purposes

Good point :+1:

bwplotka:
Related issue BTW that you will hit: https://github.com/prometheus/prometheus/issues/12941

ArthurSens:
Proposal updated with information about:
* Prometheus Agent scale down strategy
* How do we scale up Prometheus servers after a scale down

ArthurSens:
Is this good to merge? Any doubts or change requests? :)

nicolastakashi (approved):

Documentation/proposals/202310-automated-sharding.md
simonpasquier: I feel that reusing the existing statefulset is more straightforward and less confusing.

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+  * [Arthur Sens](https://github.com/ArthurSens)
+  * [Nicolas Takashi](https://github.com/nicolastakashi)
+* Related Tickets:
+  * [#4927](https://github.com/prometheus-operator/prometheus-operator/issues/4727)
+  * [#4946](https://github.com/prometheus-operator/prometheus-operator/issues/4946)
+  * [#4967](https://github.com/prometheus-operator/prometheus-operator/issues/4967)
+* Other docs:
+  * n/a
+
+This document aims to solve many small problems to unblock the usage of Horizontal Pod Autoscalers with Prometheus sharding mechanism.
```
simonpasquier, line 13: (nit)
```suggestion
This document aims to address existing issues preventing users from leveraging the Horizontal Pod Autoscaler (HPA) in conjunction with Prometheus sharding.
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+    * [Scaling up after scaling down](#scaling-up-after-scaling-down)
+  * [Graceful scale-down of Prometheus Agents](#graceful-scale-down-of-prometheus-agents)
+* [Scale-down alternatives](#scale-down-alternatives)
+* [Action Plan](#action-plan)
+
+# Why
+
+Managing Prometheus instances with highly variable workloads can be quite challenging. When the number of scraped targets can scale up and down between a few items to thousands over a short period of time (e.g. within a day), operators are forced to opt for the safest scenario, ending up with over-provisioned instances. Additionally, the WAL replay often becomes a problem after a restart as it takes a long time to be replayed into memory. Without available replicas, this leads to unnecessary downtime.
+
+To address these issues, this document outlines a solution to enhance the Prometheus Operator by enabling the usage of Horizontal Pod Autoscalers(HPAs) to scale shards up and down. This improvement aims to simplify the process of scaling Prometheus instances, making it more accessible and efficient for operators.
```
simonpasquier, line 32: (nit)
```suggestion
To address these issues, this document outlines a solution to enhance the Prometheus Operator by enabling the usage of Horizontal Pod Autoscalers (HPAs) to scale shards up and down. This improvement aims to simplify the process of scaling Prometheus instances, making it more accessible and efficient for operators.
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+
+# Pitfalls of the current solution
+
+Prometheus can scale vertically pretty well, but instances that grow up to hundreds of GiB/TiB in WAL size start to get troublesome:
```
simonpasquier, line 36:
```suggestion
Prometheus can scale vertically pretty well, but instances that grow up to hundreds of GiB/TiB in WAL size face several challenges:
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+
+* Eventual cardinality explosions have a big blast radius if a single Prometheus instance is responsible for scraping the majority of monitored applications.
+* Recovering from a crash takes several minutes due to the WAL replay.
+
+Meanwhile, another strategy would be to use Horizontal Pod Autoscalers (HPAs) with Prometheus statefulsets by increasing the number of replicas. However, arguably it makes things more complicated:
```
simonpasquier, line 41: (suggestion) do we need to keep this part? It should be obvious that increasing the number of replicas (not shards) doesn't solve the scalability issue. I'd rather see a brief explanation about why the actual sharding doesn't work either.

````
While the Prometheus and PrometheusAgent CRDs implement sharding for scrape configurations, there are still pending issues and missing features that prevent adoption.
````

File: Documentation/proposals/202310-automated-sharding.md [hunk 19]
```diff
@@ -0,0 +1,164 @@
...
+* Implement the autoscaling logic into the Prometheus operator.
+* Automatic balancing of targets to ensure homogeneous load across all Prometheus pods.
+
+# Audience
+
+* Users looking for a shard autoscaling mechanism with a good cost-reliability ratio.
+
+# How
+
+Today, there are a few strategies to measure load on top of Prometheus instances.
```
simonpasquier, line 60: (nit)
```suggestion
Today, there are a few strategies to measure the load of Prometheus instances.
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+2. Another strategy might be to measure the amount of samples ingested per second
+
+- Prometheus Agents: `rate(prometheus_agent_samples_appended_total[5m])`
+- Prometheus Servers: `rate(prometheus_tsdb_head_samples_appended_total[5m])`
+
+3. Or even measuring samples queried per second for Prometheus instances under heavy query load.
+
+- Prometheus Servers: `rate(prometheus_engine_query_samples_total[5m])`
+
+Modern Custom Horizontal Pod Autoscalers (HPAs), such as Keda are already capable of [scaling CRDs](https://keda.sh/docs/2.11/concepts/scaling-deployments/#scaling-of-custom-resources) based on different parameters, such as [CPU, Memory](https://keda.sh/docs/2.11/scalers/cpu/) or even [custom Prometheus metrics](https://keda.sh/docs/2.11/scalers/prometheus/).
```
simonpasquier, line 73: The main point here IMHO is to say that the operator is going to be "autoscaler-agnostic". We don't prescribe that you use HPA or Keda: as long as the scaler supports the scale subresource, it should work.

Documentation/proposals/202310-automated-sharding.md
simonpasquier: Let's start with this assumption. We can adjust later if needed.

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Scale Status subresource
+
+When working with CRDs, common HPAs (e.g. Keda) depends on a Status subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). Unfortunately, Prometheus-Operator resources still don't implement this field. There was an attempt to implement this in https://github.com/prometheus-operator/prometheus-operator/pull/4735 but was reverted because the PR was scaling replicas instead of shards. Although scaling replicas can help with high availability, it is expensive and hard to manage since it duplicates scrapes while not reducing the load on top of Prometheus because all replicas still scrape the same targets. Sharding serves as a better autoscaling strategy since the load is spread into all instances instead of getting duplicated.
```
simonpasquier, line 77: I would omit the history.
```suggestion
When working with CRDs, common HPAs (e.g. Keda) depends on a subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). This proposal suggests to implement the scale subresource for the Prometheus and PrometheusAgent CRDs. Instead of working on the "replicas" count, it will operate on the "shards" count because the purpose of scaling up (resp. down) is to distribute the same number of targets across more (resp. less) Prometheus instances.
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+- Prometheus Agents: `rate(prometheus_agent_samples_appended_total[5m])`
+- Prometheus Servers: `rate(prometheus_tsdb_head_samples_appended_total[5m])`
+
+3. Or even measuring samples queried per second for Prometheus instances under heavy query load.
+
+- Prometheus Servers: `rate(prometheus_engine_query_samples_total[5m])`
+
+Modern Custom Horizontal Pod Autoscalers (HPAs), such as Keda are already capable of [scaling CRDs](https://keda.sh/docs/2.11/concepts/scaling-deployments/#scaling-of-custom-resources) based on different parameters, such as [CPU, Memory](https://keda.sh/docs/2.11/scalers/cpu/) or even [custom Prometheus metrics](https://keda.sh/docs/2.11/scalers/prometheus/).
+
+## Scale Status subresource
```
simonpasquier, line 75:
```suggestion
## Scale subresource
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+With only this change, Prometheus Agents can already be horizontally scaled without problems, but for Prometheus Servers it gets a little more complicated.
+
+## Graceful scale-down of Prometheus Servers
+
+Prometheus Servers don't behave like the usual stateless applications. A few cases require Prometheus pods to still be available even after a scale-down event is triggered:
+
+* Thanos proxying queries back to the sidecars to retrieve data that still wasn't uploaded to object storage.
+* Prometheus servers with long retention periods still hold data that users might want to query.
+
+If we simply delete Prometheus pods during scale-down, data loss will be a common problem, so a more complex scale-down strategy needs to be used here.
```
simonpasquier, line 88: (nit)
```suggestion
If the operator deletes Prometheus pods in excess during a scale-down operation, data loss will be a common problem, so a more complex strategy needs to be used here.
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+    whenScaled: Retain|Delete 
+    # if not specified, the operator picks up the retention period if it's defined,
+    # otherwise (e.g. size-based retention) the statefulset lives forever.
+    retain:
+      retentionPeriod: 3d
+```
+
+Inspired by the [StatefulSet API](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention), when setting `.sharRetentionPolicy.whenScaled` is set to `Delete` the operator would simply delete the underlying StatefulSet on the spot, not caring about the data stored in that particular shard. Although not the safest approach, it is exactly the current behavior, and changing it might surprise a lot of users.
+
+If `.shardRetentionPolicy.whenScaled` is set to `Retain` Prometheus-Operator will keep the shard running for a customizable period of time. By default, the Operator will keep the Shard for the whole retention period, or forever if no retention period is set. Alternatively, the user can choose a customized retention period.
```
simonpasquier, line 109:
```suggestion
If `.shardRetentionPolicy.whenScaled` is set to `Retain` Prometheus-Operator will keep the shard running for a customizable period of time. By default, the Operator will keep the Shard for the whole retention period, or forever if only size-based retention is defined. Alternatively, the user can choose a customized retention period.
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+
+If retention needs to be done, the deletion logic is controlled by an annotation present on the StatefulSet:
```
simonpasquier, line 111:
```suggestion
If a retention time is defined, the deletion logic is controlled by an annotation present on the StatefulSet:
```

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+We intentionally don't want to spin up new instances while others that are marked for deletion are still running as shard names and containers' environment variables that configure each shard's target group will be too complex to coordinate.
+
+## Graceful scale-down of Prometheus Agents
+
+Prometheus Agents are different than servers since queries are not available in this mode. Their only responsibility is scraping metrics and pushing them via remote-write to a long-term storage backend, making the scale-down experience much easier to handle.
+
+When receiving the SIGTERM signal, the Prometheus Agent will gracefully handle the signal by finishing all remote-write queues before ending the process. [When terminating a Pod, Kubernetes sends the SIGTERM signal and, by default, waits 30 seconds for all containers in the pod to finish their processes](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination). Thirty seconds is usually plenty of time to finish the remote-write queues, but the PrometheusAgent CRD will be extended to allow changing the Pod's `terminationGracePeriodSeconds` field so bigger or unstable instances can be shut down without data loss.
```
simonpasquier, line 131: The operator sets a default termination grace period of 600s which should be more than enough. If really needed, users can use strategic merge patch to adjust the value.

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+## Dump & Backfill
+
+During scale-down, the Prometheus-Operator could read TSDB Blocks from the Prometheus instance being deleted and backfill it into another instance.
+
+***Advantages:***
+* Prometheus can be shut down immediately without data loss.
+
+***Disadvantages:***
+* Loading TSDB Blocks into memory is expensive, requiring Prometheus-Operator to run with big Memory/CPU requests.
+* Complex and hard to coordinate workflow. (Require restarts to reload TSDB blocks, hard to identify possible corruptions)
```
simonpasquier, line 146: I'm also not sure how the data would be transferred. Even with persistent storage, it would be intricate to attach the existing PVCs to other pods.

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Snapshot & Upload on shutdown
+
+During scale-down, Prometheus-Operator could send an HTTP request to [Prometheus' Snapshot endpoint](https://prometheus.io/docs/prometheus/latest/querying/api/#snapshot). Thanos sidecar could be extended to watch the snapshot directory and automatically upload snapshots to Object storage.
```
simonpasquier, line 150: that's an interesting approach which could be complementary to the main proposal.

Documentation/proposals/202310-automated-sharding.md
ArthurSens: Good point, I can see that the current wording is too strong towards Keda. I'll reword it

Documentation/proposals/202310-automated-sharding.md
ArthurSens: Yep, that's exactly what we proposed just a few lines below :)

See section "Scaling up after scaling down"

Documentation/proposals/202310-automated-sharding.md
ArthurSens: ah good reminder! I'll rephrase this part

Documentation/proposals/202310-automated-sharding.md
ArthurSens: Totally agree, but for the sake of simplicity I'd prefer to leave this out of the main proposal for now 

After we have the whole proposal working I'd move forward with this idea

ArthurSens:
Thanks for all the reviews everybody, proposal updated again!

Documentation/proposals/202310-automated-sharding.md
```diff
@@ -0,0 +1,165 @@
...
+* Recovering from a crash takes several minutes due to the WAL replay.
+
+While the Prometheus and PrometheusAgent CRDs implement sharding for scrape configurations, there are still pending issues and missing features that prevent adoption.
+
+* Environments where the load is undistributed throughout the day/week, operators need to constantly adjust number of shards or shard size.
+* Scaling down Prometheus Servers causes data loss, as there's no guarantee that stored data is persisted in centralized storage before the container is deleted.
+
+# Goals
+
+* Enable scaling of the Prometheus shards up and down via Horizontal Pod Autoscaler objects.
```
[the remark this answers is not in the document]
ArthurSens: I had the feeling we weren't suggesting which HPA to use, but we're still focused on enabling the HPA use-case 🤔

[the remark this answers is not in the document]
ArthurSens: ah ok, from your comment (https://github.com/prometheus-operator/prometheus-operator/pull/5961#discussion_r1439467960) I think we had different point of views about what HPA means.

From my point of view, anything that is able to scale pods horizontally, is an HPA. From your comment, I think you distinguish the native K8s HPA resource from other tools that are capable of doing the same thing. Is that correct?

I'm fine with updating the proposal making a more clear distinction between them, just wanted to make sure I understand the problem here :)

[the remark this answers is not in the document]
simonpasquier: You're correct. When I see "HPA", I tend to assume the Kubernetes Horizontal Pod Autoscaling controller.

Documentation/proposals/202310-shard-autoscaling.md
```diff
@@ -0,0 +1,165 @@
...
+2. Another strategy might be to measure the amount of samples ingested per second
+
+- Prometheus Agents: `rate(prometheus_agent_samples_appended_total[5m])`
+- Prometheus Servers: `rate(prometheus_tsdb_head_samples_appended_total[5m])`
+
+3. Or even measuring samples queried per second for Prometheus instances under heavy query load.
+
+- Prometheus Servers: `rate(prometheus_engine_query_samples_total[5m])`
+
+These are all data that could be configured as input for different Horizontal Pod Autoscalers, but how does HPAs knows if the amount of shard is bigger/lower than desired?
```
simonpasquier, line 74: (nit)
```suggestion
These are all data that could be configured as input for different Horizontal Pod Autoscalers, but how do HPAs know if the amount of shard is bigger/lower than desired?
```

Documentation/proposals/202310-shard-autoscaling.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Scale subresource
+
+When working with any resource, including CRDs, HPAs depends on a subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). This proposal suggests to implement the scale subresource for the Prometheus and PrometheusAgent CRDs. Instead of working on the "replicas" count, it will operate on the "shards" count because the purpose of scaling up (resp. down) is to distribute the same number of targets across more (resp. less) Prometheus instances.
```
simonpasquier, line 78: (nit)
```suggestion
When working with any resource, including CRDs, HPAs depend on a subresource called [Scale](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource). This proposal suggests to implement the scale subresource for the Prometheus and PrometheusAgent CRDs. Instead of working on the "replicas" count, it will operate on the "shards" count because the purpose of scaling up (resp. down) is to distribute the same number of targets across more (resp. less) Prometheus instances.
```

Documentation/proposals/202310-shard-autoscaling.md
```diff
@@ -0,0 +1,165 @@
...
+    whenScaled: Retain|Delete 
+    # if not specified, the operator picks up the retention period if it's defined,
+    # otherwise (e.g. size-based retention) the statefulset lives forever.
+    retain:
+      retentionPeriod: 3d
+```
+
+Inspired by the [StatefulSet API](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention), when setting `.shardRetentionPolicy.whenScaled` is set to `Delete` the operator would simply delete the underlying StatefulSet on the spot, not caring about the data stored in that particular shard. Although not the safest approach, it is exactly the current behavior, and changing it might surprise a lot of users.
+
+If `.shardRetentionPolicy.whenScaled` is set to `Retain` Prometheus-Operator will keep the shard running for a customizable period of time. By default, the Operator will keep the Shard for the whole retention period, or forever if only size-based retention is defined. Alternatively, the user can choose a customized retention period.
```
simonpasquier, line 110: (nit)
```suggestion
If `.shardRetentionPolicy.whenScaled` is set to `Retain` Prometheus-Operator will keep the shard running for a customizable period of time. By default, the Operator will keep the Shard for the whole retention period, or forever if only size-based retention is defined. Alternatively, the user can define a custom retention period.
```

Documentation/proposals/202310-shard-autoscaling.md
```diff
@@ -0,0 +1,165 @@
...
+
+If a retention time is defined, the deletion logic is controlled by an annotation present on the StatefulSet:
+
+```yaml
+operator.prometheus.io/deletion-timestamp: X
+```
+
+On scale downs, the configuration of all shards would be re-arranged to make sure the Prometheus that was scaled down doesn't scrape any metrics anymore, but it would still be available for queries.
```
simonpasquier, line 118: (nit)
```suggestion
On scale downs, the configuration of all shards will be re-arranged to make sure that the "scaled-down" Prometheus pods don't scrape targets anymore but they will still be available for queries.
```

Documentation/proposals/202310-shard-autoscaling.md
```diff
@@ -0,0 +1,165 @@
...
+
+When the current time finally surpasses the one in the annotation, the Operator will finally trigger the deletion of the statefulset associated with the shard.
```
simonpasquier, line 120:
```suggestion
When the deletion timestamp set in the annotation is reached, the Operator will trigger the deletion of the statefulset associated with the shard.
```

Documentation/proposals/202310-shard-autoscaling.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Scaling up after scaling down
+
+If a scale-up event happens while Prometheus Servers that were marked for deletion are still running, the already running Prometheus will be unmarked and the configuration of the shard pool will be reloaded as usual.
```
simonpasquier, line 124:
```suggestion
If a scale-up event happens while the Prometheus pods marked for deletion are still running, the annotation will be removed and the scrape targets will be redistributed to all active shards.
```

simonpasquier (approved):
Just a few nits that I'm going to commit right away.

PR #5962, opened 2023-10-03
Add scale subresource to Prometheus/PrometheusAgent

### LINKED ISSUE #4727: prometheuses, alertmanagers, thanosrulers CRD should add scale subresources
**What is missing?**
prometheuses, alertmanagers, thanosrulers CRD should add [scale subresources](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource)

**Why do we need it?**

So `kubectl scale ...` would just work and provided documentation stated "When the scale subresource is enabled, the /scale subresource for the custom resource is exposed. The autoscaling/v1.Scale object is sent as the payload for /scale."

**Environment**

    `prometheus-operator/prometheus-operator:v0.53.1`

ArthurSens:
## Description
Fix #4727 

Adds scale subresource to Prometheus and PrometheusAgent. The goal is that it can be used to scale prometheus shards with the command `kubectl scale`, but also by Horizontal Pod Autoscalers (HPAs).

## Type of change

## Changelog entry

```release-note
Implement scale subresource for Prometheus and PrometheusAgent CRDs
```


ArthurSens:
I'm still figuring out what this selector field is used for by HPAs... I've added a label selector but I'm not sure if that's correct.

Another thing I've noticed is that .Status.Replicas is the result of `.spec.Replicas * .spec.Shards`, should it be only `.spec.Shards` now?

ArthurSens:
urrgh, for some reason that I still couldn't figure out yet `kubectl scale` works as expected, but calling `UpdateScale()` from the go client doesn't seem to do anything 🙄

pkg/prometheus/agent/operator.go
```diff
@@ -805,6 +805,7 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 		return errors.Wrap(err, "failed to get prometheus agent status")
 	}
 	p.Status = *pStatus
+	p.Status.Selector = fmt.Sprintf("app.kubernetes.io/instance=%s", p.Name)
```
simonpasquier, line 808: also `app.kubernetes.io/name=prometheus-agent`?

pkg/prometheus/server/operator.go
```diff
@@ -1367,6 +1367,7 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	}
 
 	p.Status = *pStatus
+	p.Status.Selector = fmt.Sprintf("app.kubernetes.io/instance=%s", p.Name)
```
simonpasquier, line 1370: and `app.kubernetes.io/name=prometheus`?

test/framework/prometheus.go
```diff
@@ -319,17 +321,14 @@ func (f *Framework) CreatePrometheusAndWaitUntilReady(ctx context.Context, ns st
...
-		ctx,
-		name,
-		ns,
-		monitoringv1.PrometheusSpec{
-			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
-				Replicas: ptr.To(int32(replicas)),
-			},
-		},
-	)
+func (f *Framework) ScalePrometheusAndWaitUntilReady(ctx context.Context, p *monitoringv1.Prometheus, ns string, shards int32) (*monitoringv1.Prometheus, error) {
```
simonpasquier, line 324: (nit) could use p.Namespace directly?
```suggestion
func (f *Framework) ScalePrometheusAndWaitUntilReady(ctx context.Context, p *monitoringv1.Prometheus, shards int32) (*monitoringv1.Prometheus, error) {
```

ArthurSens:
https://www.loom.com/share/1b7a5f5ae79a466689cd78153d966450

Quick video showing this PR interacting with Keda :)

**Almost** working!!!

ArthurSens:
@simonpasquier, should we tackle the UpdateScale() and GetScale() in this PR too, or leave it for a separate one?

Also, I'm not sure about the whole status struct. I'm a bit confused on how to organize replicas and shards over there

simonpasquier:
> should we tackle the UpdateScale() and GetScale() in this PR too, or leave it for a separate one?

you mean updating the client library to expose UpdateScale() and GetScale()? IMHO it can be a follow-up PR.

> Also, I'm not sure about the whole status struct. I'm a bit confused on how to organize replicas and shards over there

Do you mean whether we need a dedicated struct for shard count and selector or if we add them at the top?

simonpasquier:
:thinking: maybe the first work items should be about implementing the scale subresource without `labelSelectorPath`?
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource

ArthurSens:
> 🤔 maybe the first work items should be about implementing the scale subresource without `labelSelectorPath`? https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource

Isn't it needed to work with HPAs though? Could you elaborate on why you believe we should implement without labelSelectorPath?

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -851,6 +852,22 @@ type PrometheusStatus struct {
 	// +listMapKey=shardID
 	// +optional
 	ShardStatuses []ShardStatus `json:"shardStatuses,omitempty"`
+	// Scale is the Prometheus's scale subresource status.
+	// +optional
+	Scale ScaleSubresourceStatus `json:"scale,omitempty"`
```
simonpasquier, line 857: My preference would be to keep the fields at this level rather than introducing an intermediate struct.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -851,6 +852,22 @@ type PrometheusStatus struct {
...
+}
+
+// ScaleSubresourceStatus defines the observed state of the Prometheus's
+// scale subresource.
+type ScaleSubresourceStatus struct {
+	// The total number non-terminated pods targeted by this
+	// Prometheus resource.
```
simonpasquier, line 864:
```suggestion
	// Shards is the most recently observed number of shards.
```

simonpasquier:
To generate UpdateScale() and GetScale(), adding the following comments before the Prometheus and PrometheusAgent structs (+ regenerate) should be enough?

````
// +genclient:method=GetScale,verb=get,subresource=scale,result=k8s.io/api/autoscaling/v1.Scale
// +genclient:method=UpdateScale,verb=update,subresource=scale,input=k8s.io/api/autoscaling/v1.Scale,result=k8s.io/api/autoscaling/v1.Scale
`````

**Comment by ArthurSens** [comment]
> To generate UpdateScale() and GetScale(), adding the following comments before the Prometheus and PrometheusAgent structs (+ regenerate) should be enough?
> 
> ```
> // +genclient:method=GetScale,verb=get,subresource=scale,result=k8s.io/api/autoscaling/v1.Scale
> // +genclient:method=UpdateScale,verb=update,subresource=scale,input=k8s.io/api/autoscaling/v1.Scale,result=k8s.io/api/autoscaling/v1.Scale
> ```

Yes, it does generate the methods as expected... they just don't do what I expected them to do, which was get and update the number of shards of a Prometheus resource.

I'll leave these ones for another PR once I get back from PTO, hope that's ok :)

**Comment by ArthurSens** [comment]
We got a failure on the OperatorUpgrade e2e test while waiting for Alertmanager to be ready after upgrading the version. I've just run the same test locally and it worked first try 🤔 

I'm retriggering the CI, hopefully it was just some Github Actions instability

**Comment by ArthurSens** [comment]
@simonpasquier does your intuition say anything about the test failure? It is passing locally to me :(

**Comment by simonpasquier** [comment]
Hmm, could it be that the CRD versions aren't up-to-date or that the operator fails to update the status subresource?

**Review by metalmatze — COMMENTED** [review]
This is going to be great! :100:

File: test/framework/prometheusagent.go [hunk 6]
```diff
@@ -161,3 +167,37 @@ func (f *Framework) PatchPrometheusAgent(ctx context.Context, name, ns string, s
 
 	return p, nil
 }
+
+// TODO(arthursens): Use UpdateScale() instead of PatchPrometheusAgentShardsAndWaitUntilReady() once we have a
+// working UpdateScale() implementation.
```
ArthurSens, line 172: note to self: Open issue before merging

ArthurSens:
@prometheus-operator/prometheus-operator-reviewers, this should be ready for review now!

Thanks @adinhodovic for finding the bug ❤️!

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -922,6 +923,12 @@ type PrometheusStatus struct {
 	// +listMapKey=shardID
 	// +optional
 	ShardStatuses []ShardStatus `json:"shardStatuses,omitempty"`
+	// Shards is the most recently observed number of shards.
+	// +optional
```
simonpasquier, line 927: not sure that we need optional here

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -922,6 +923,12 @@ type PrometheusStatus struct {
...
+	Shards int32 `json:"shards,omitempty"`
+	// The selector used to match the pods targeted by this Prometheus resource.
+	// +optional
```
simonpasquier, line 930: same

pkg/prometheus/agent/operator.go
```diff
@@ -808,6 +808,21 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
...
+	}
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: selectorLabels})
+	if err != nil {
+		return fmt.Errorf("failed to create selector for prometheus agent scale status: %w", err)
+	}
+	p.Status.Selector = selector.String()
+	p.Status.Shards = 1
+	if p.Spec.Shards != nil {
+		p.Status.Shards = *p.Spec.Shards
+	}
```
simonpasquier, line 824: (nit)
```suggestion
	p.Status.Shards = ptr.Deref(p.Spec.Shards, 1)
```

pkg/prometheus/agent/operator.go
```diff
@@ -808,6 +808,21 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	}
 	p.Status = *pStatus
 
+	selectorLabels := map[string]string{
+		"app.kubernetes.io/name":       "prometheus-agent",
+		"app.kubernetes.io/managed-by": "prometheus-operator",
+		"app.kubernetes.io/instance":   p.Name,
+	}
```
simonpasquier, line 815: maybe we should have a shared function with `makeStatefulSetSpec()` to return the selector labels?

pkg/prometheus/server/operator.go
```diff
@@ -1161,6 +1161,20 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	}
 
 	p.Status = *pStatus
+	selectorLabels := map[string]string{
```
simonpasquier, line 1164: same comments than for agent.

test/framework/prometheus.go
```diff
@@ -315,6 +315,12 @@ func (f *Framework) CreatePrometheusAndWaitUntilReady(ctx context.Context, ns st
 		return nil, fmt.Errorf("waiting for %v Prometheus instances timed out (%v): %v", p.Spec.Replicas, p.Name, err)
 	}
 
+	// We get the prometheus instance again to make sure we have status updated.
```
simonpasquier, line 318: (nit) does it make sense to modify the signature of `WaitForPrometheusReady()` to return the updated object directly?

simonpasquier (commented):
out of curiosity: the tests failed because they didn't get the latest version of the object?

ArthurSens:
> out of curiosity: the tests failed because they didn't get the latest version of the object?

In pkg/prometheus/{server,agent}/operator.go I was not doing a nil check before updating the Status
```go
	p.Status.Shards = *p.Spec.Shards
```

Prometheus-Operator was restarting due to nilpointers, the alertmanager tests were failing because of that. I was deep down reading alertmanager tests and trying to understand why changing Prometheus API would affect their tests 😆 🤦 

Doing the nilcheck solved the problem
```go
	p.Status.Shards = 1
	if p.Spec.Shards != nil {
		p.Status.Shards = *p.Spec.Shards
	}
```

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: good point, muscle memory 😅

pkg/prometheus/agent/operator.go
ArthurSens: > maybe we should have a shared function with `makeStatefulSetSpec()` to return the selector labels?

I'm not sure I understand what you mean by "shared function with makeStatefulSetSpec"...

could you elaborate?

simonpasquier: I'm referring to this part where the controller defines the label selector for the statefulset's pods:
https://github.com/prometheus-operator/prometheus-operator/blob/a8eb064ec76327afca43638085c82a20f6850689/pkg/prometheus/server/statefulset.go#L347-L354

pkg/prometheus/agent/operator.go
```diff
@@ -808,6 +809,18 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
...
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: selectorLabels})
+	if err != nil {
+		return fmt.Errorf("failed to create selector for prometheus agent scale status: %w", err)
+	}
+	p.Status.Selector = selector.String()
+	p.Status.Shards = ptr.Deref(p.Spec.Shards, 1)
```
simonpasquier, line 822: What happens if the CRD hasn't been updated and doesn't expose the new selector and shards fields? Will the status update operation fail?

ArthurSens: Good catch! It fails with the error:

```
level=error ts=2023-12-20T17:47:18.49449Z caller=klog.go:116 component=k8s_client_runtime func=ErrorDepth msg="status \"monitoring/k8s\" failed: failed to apply prometheus status subresource: failed to create typed patch object (monitoring/k8s; monitoring.coreos.com/v1, Kind=Prometheus): .status.selector: field not declared in schema"
```

ArthurSens: I think we can use the same strategy we use for PrometheusAgent/Scrape config CRDs.

Let's check if the operator has permissions for `prometheuses/scale`. If it does, it's very likely that the new scale fields are present in the CRD and it's safe to update them

pkg/prometheus/agent/operator.go
ArthurSens: addressed in [0b2ecbf](https://github.com/prometheus-operator/prometheus-operator/pull/5962/commits/0b2ecbf142aed30dc423236da5576467b8bc0fbc)

thanks for clarifying

pkg/prometheus/agent/operator.go
simonpasquier: Yes, this looks a sane approach to me.

cmd/operator/main.go
```diff
@@ -261,7 +262,27 @@ func run(fs *flag.FlagSet) int {
 		return 1
 	}
 
-	po, err := prometheuscontroller.New(ctx, restConfig, cfg, log.With(logger, "component", "prometheusoperator"), r, scrapeConfigSupported, canReadStorageClass)
+	scaleSubresourceSupported, err := checkPrerequisites(
```
simonpasquier, line 265: I would use `k8sutil.IsAllowed()` directly?

cmd/operator/main.go
```diff
@@ -261,7 +262,27 @@ func run(fs *flag.FlagSet) int {
...
+		ctx,
+		logger,
+		kclient,
+		cfg.Namespaces.AllowList.Slice(),
```
simonpasquier, line 269: should rather be `cfg.Namespaces.PrometheusAllowList.Slice()`?

cmd/operator/main.go
```diff
@@ -294,9 +315,29 @@ func run(fs *flag.FlagSet) int {
 		return 1
 	}
 
+	agentScaleSubresourceSupported, err := checkPrerequisites(
```
simonpasquier, line 318: same here, we could use k8sutil.IsAllowed() and move it inside the `if prometheusAgentSupported {...}` block.

pkg/prometheus/agent/operator.go
simonpasquier: Hmm, wait a sec. The operator doesn't need any `prometheuses/scale` permissions to run properly?

ArthurSens: Yeah, from my tests here the permissions aren't necessary 🤔 strange...

Not sure how to handle the situation though... is it possible to check something like "is field .status.shards and .status.selector present in CRD"?

simonpasquier: The only way it could work would be to read the CustomResourceDefinition object. The operator's service account doesn't have the permissions but it doesn't seem controversial to allow the service account getting its own CRDs. E.g.

```
rules:
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  resourceNames:
  - prometheuses.monitoring.coreos.com
  - prometheusagents.monitoring.coreos.com
  verbs:
  - get

```

ArthurSens: But requiring more RBAC permissions is also a breaking change, isn't it? If someone is upgrading the RBAC, why not update the CRDs as well?

simonpasquier: Totally, it wouldn't work either until you upgrade the RBAC but at least we'll have that in place for the future.

We could implement a fallback mechanism in UpdateStatus() which would retry applying the change without the new fields being set. This can be removed after 1 or 2 releases.
I'll add metrics and alerting to track status update operations (it was an oversight when implementing the status subresource) so folks can be aware of the issue.

simonpasquier: Also if we're confident that a fallback solution will be in place before the next release, I'm fine merging this PR in the current state.

ArthurSens: Yeah, I can work on this before the next release for sure!

simonpasquier:
See #6185 which adds metrics to the status reconciler.

pkg/prometheus/agent/operator.go
ArthurSens: Just added it, PTAL

ArthurSens:
Is this good to merge or do we want #6197 to get in first? 🤔

jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -85,9 +85,11 @@ function(params) {
           'prometheuses',
           'prometheuses/finalizers',
           'prometheuses/status',
+          'prometheuses/scale',
```
simonpasquier, line 88: I'm still not sure why we need to add "prometheuses/scale" and "prometheusagents/scale" to the operator's default permissions.

pkg/prometheus/agent/operator.go
```diff
@@ -808,8 +809,19 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
...
+	selectorLabels := makeAgentSelectorLabels(p.Name)
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: selectorLabels})
+	if err != nil {
+		return fmt.Errorf("failed to create selector for prometheus agent scale status: %w", err)
+	}
+	p.Status.Selector = selector.String()
+	p.Status.Shards = ptr.Deref(p.Spec.Shards, 1)
+
+	if _, err = c.mclient.MonitoringV1alpha1().PrometheusAgents(p.Namespace).ApplyStatus(ctx, prompkg.ApplyConfigurationFromPrometheusAgent(p, true), metav1.ApplyOptions{FieldManager: operator.PrometheusOperatorFieldManager, Force: true}); err != nil {
+		// Try again, but this time does not update scale subresource.
```
simonpasquier, line 821: Can we log a warning?

pkg/prometheus/server/operator.go
```diff
@@ -1161,9 +1162,19 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	}
 
 	p.Status = *pStatus
+	selectorLabels := makeServerSelectorLabels(p.Name)
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: selectorLabels})
+	if err != nil {
+		return fmt.Errorf("failed to create selector for prometheus scale status: %w", err)
+	}
+	p.Status.Selector = selector.String()
+	p.Status.Shards = ptr.Deref(p.Spec.Shards, 1)
```
simonpasquier, line 1171: nit-picky but we factorize a bit more and have a single function in pkg/prometheus doing it (as long as it gets the selector labels as input).

jsonnet/prometheus-operator/prometheus-operator.libsonnet
ArthurSens: Whoops, that was leftover from the previous commits! Good catch

simonpasquier:
> Is this good to merge or do we want https://github.com/prometheus-operator/prometheus-operator/pull/6197 to get in first? 🤔

I'll leave it to you.

pkg/prometheus/agent/operator.go
ArthurSens: what do you think of the log added?

pkg/prometheus/operator.go
```diff
@@ -284,3 +285,21 @@ func (sr *StatusReporter) Process(ctx context.Context, p monitoringv1.Prometheus
...
+	switch p.(type) {
+	case *monitoringv1.Prometheus:
+		labelMap["app.kubernetes.io/name"] = "prometheus"
+		labelMap["prometheus"] = name
+	case *monitoringv1alpha1.PrometheusAgent:
+		labelMap["app.kubernetes.io/name"] = "prometheus-agent"
+	default:
+	}
+	return labelMap
+}
```
ArthurSens, line 305: Kind of annoying that Prometheus server has an extra label when compared to Prometheus agent 😬 

I've done a type switch here to properly add the individual labels, let me know if you prefer a different strategy

ArthurSens: We could also have an extra method to the Interface itself 🤔

ArthurSens:
> > Is this good to merge or do we want #6197 to get in first? 🤔
> 
> I'll leave it to you.

I'm fighting with the linter in that PR at the moment 😠

pkg/prometheus/operator.go
simonpasquier: I find it simpler to have dedicated functions in pkg/prometheus/server and pkg/prometheus/agent in this case. Moving logic to pkg/apis/monitoring increases the surface of the API package (and can create issues like #6203).

test/framework/prometheus.go
```diff
@@ -398,10 +415,15 @@ func (f *Framework) WaitForPrometheusReady(ctx context.Context, p *monitoringv1.
 		},
 		timeout,
 	); err != nil {
-		return fmt.Errorf("prometheus %v/%v failed to become available: %w", p.Namespace, p.Name, err)
+		return nil, fmt.Errorf("prometheus %v/%v failed to become available: %w", p.Namespace, p.Name, err)
 	}
 
-	return nil
+	p, err := f.MonClientV1.Prometheuses(p.Namespace).Get(ctx, p.Name, metav1.GetOptions{})
```
simonpasquier, line 421: we don't need the extra Get() if we use the `current` variable.

ArthurSens: good find!

pkg/prometheus/agent/operator.go
```diff
@@ -818,8 +819,20 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
...
+	selectorLabels := MakeSelectorLabels(p.Name)
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: selectorLabels})
+	if err != nil {
+		return fmt.Errorf("failed to create selector for prometheus agent scale status: %w", err)
+	}
+	p.Status.Selector = selector.String()
+	p.Status.Shards = ptr.Deref(p.Spec.Shards, 1)
+
+	if _, err = c.mclient.MonitoringV1alpha1().PrometheusAgents(p.Namespace).ApplyStatus(ctx, prompkg.ApplyConfigurationFromPrometheusAgent(p, true), metav1.ApplyOptions{FieldManager: operator.PrometheusOperatorFieldManager, Force: true}); err != nil {
+		level.Warn(c.logger).Log("msg", "failed to apply prometheus status subresource, trying again without scale fields", "err", err)
```
simonpasquier, line 831: (nit) I would use an info level since the operator will handle it nicely.
```suggestion
		level.Info(c.logger).Log("msg", "failed to apply prometheus status subresource, trying again without scale fields", "err", err)
```

pkg/prometheus/agent/operator.go
```diff
@@ -1223,3 +1236,12 @@ func ListOptions(name string) metav1.ListOptions {
 		})).String(),
 	}
 }
+
+func MakeSelectorLabels(name string) map[string]string {
```
simonpasquier, line 1240: (nit) it can be private
```suggestion
func makeSelectorLabels(name string) map[string]string {
```

simonpasquier (approved):
LGTM :ship:

PR #5964, opened 2023-10-04
chore: rework operator + controllers configuration

simonpasquier:
## Description

* Refactored individual controller configuration structs to include only the needed parameters.
* Moved fields away from the operator.Config struct as they were not needed by the individual controllers.
* Parsed label and field selector arguments early, instead of doing it in each controller.

## Type of change

## Changelog entry

```release-note

```


simonpasquier:
Tested with our downstream OCP e2e tests (https://github.com/openshift/prometheus-operator/pull/257) and it passes.

ArthurSens (approved):
Oof, quite a big PR... maybe PRs like this could be broken down in the future somehow?

Maybe one PR for each of the changes you mentioned?
```
Refactored individual controller configuration structs to include only the needed parameters.
Moved fields away from the operator.Config struct as they were not needed by the individual controllers.
Parsed label and field selector arguments early, instead of doing it in each controller.
Refactor and share the web server logic across components
Same for the logging setup.
```

Anyway, thanks for testing this with openshift already :)

simonpasquier:
@ArthurSens yes i realize that the PR should have been broken up into smaller chunks but doing it after the fact wasn't easy. I'll remember for the next time for sure.

Issue #4782, opened 2022-05-11
Shards aware TopologySpreadConstraint for Prometheus


sbeginCoveo:
**What is missing?**

The ability to define TopologySpreadConstraint using the label `operator.prometheus.io/shard` when sharding prometheus

**Why do we need it?**

Right now we can only define a single TSC or anti-affinity for _all_ statefulsets. 

* Prometheus Operator version:

`prometheus-operator/prometheus-operator:v0.53.1`

**Anything else we need to know?**:

With some guidance, I could make this happen. We would need to define the implementation details first. For now I think we would need to update all these components:
- the [Operator's statefulset generation](https://github.com/prometheus-operator/prometheus-operator/blob/9456e316035eadd0e8a472df4eee50787798a803/pkg/prometheus/statefulset.go#L1001) 
- the [prometheus crd](https://github.com/prometheus-operator/prometheus-operator/blob/main/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml#L6578)
- the [helm chart in which its deployed](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L251)

sbeginCoveo:
I'm thinking of something along these lines, to the best of my knowledge:

**User definition**
The user defines a set of "managed TSC" config and custom ones, such as:
```yaml
shardsAwareTSC:
- maxSkew: 2
  topologyKey: topology.kubernetes.io/zone
  whenUnsatisfiable: DoNotSchedule
- maxSkew: 1
  topologyKey: kubernetes.io/hostname
  whenUnsatisfiable: DoNotSchedule

additionnalTSC:
- labelSelector:
    matchLabels:
      app.kubernetes.io/name: prometheus
  maxSkew: 1
  topologyKey: kubernetes.io/hostname
  whenUnsatisfiable: ScheduleAnyway
```

**Generation**
And then each STS has its own set ot TSCs generated based on common prometheus labels, plus the additional TSCs are merged in the array. 

We could even keep the `additionnalTSC` named `topologySpreadConstraints` in the prometheus CRD to avoid creating a breaking change

_Shard 1 STS_
```yaml
topologySpreadConstraints:
- labelSelector:
    matchLabels:
      operator.prometheus.io/name: prometheus
      operator.prometheus.io/shard: "1"
  maxSkew: 2
  topologyKey: topology.kubernetes.io/zone
[...]

PR #5967, opened 2023-10-05
Add `additionalLabels` field to TopologySpreadConstraint

nicolastakashi:
## Description

Add the ability to define TopologySpreadConstraint using other labels available on the Prometheus Pod

Resolve: #4782

## Type of change

## Changelog entry

```release-note
additionalLabels to Prometheus Pod TopologySpreadConstraint 
```


nicolastakashi:
Still missing the unit tests, but I guess it's fine for a first round of review. 🥳

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,157 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
+
+	if len(tp.AdditionalLabelSelectors) > 0 {
+		if tp.LabelSelector == nil {
+			tp.LabelSelector = &metav1.LabelSelector{
+				MatchLabels: make(map[string]string),
+			}
+		}
+
+		for _, label := range tp.AdditionalLabelSelectors {
+			tp.LabelSelector.MatchLabels[label] = podLabels[label]
```
nicolastakashi, line 194: I'm not sure what should be the behavior if users provide a label that dont exists on the podLabels.
Should we fail or just log it?

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,157 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
 	return l.Status
 }
 
+type TopologySpreadConstraint struct {
+	// MaxSkew describes the degree to which pods may be unevenly distributed.
```
nicolastakashi, line 60: I was torn between replicate the TopologySpreadConstraint struct or add only one property named TopologySpreadConstraintAdditionalLabels.

I'm open for suggestions

ArthurSens: Could we inline the original topology struct? Any special reason to replicate the whole struct?

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,157 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
+	//
+	// This is a beta field and requires the MatchLabelKeysInPodTopologySpread feature gate to be enabled (enabled by default).
+	// +listType=atomic
+	// +optional
+	MatchLabelKeys []string `json:"matchLabelKeys,omitempty" protobuf:"bytes,8,opt,name=matchLabelKeys"`
+
+	//+optional
+	// +listType=atomic
+	// Defines what Prometheus Operator managed labels should be added to topologySpreadConstraints[].labelSelector.
+	AdditionalLabelSelectors []string `json:"additionalLabelSelectors,omitempty"`
```
ArthurSens, line 181: Do we want to allow any label? Or just the ones added by Prometheus-Operator?

We could do an enum if it's the latter, with proper validation

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,157 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
+}
+
+func (tp TopologySpreadConstraint) BuildK8sTopologySpreadConstraint(podLabels map[string]string) v1.TopologySpreadConstraint {
+
+	if len(tp.AdditionalLabelSelectors) > 0 {
```
ArthurSens, line 186: []string can be nil, so we also need to check for nil

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: Another question, is it possible to override pre-existing labels here? If it is, is that a problem?

ArthurSens (commented):
I'm still trying to understand this whole Topology object... I have more questions than suggestions at the moment 😬

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: I tried inline but inline ends with something like that:

TopologySpreadConstraint:
    additionalLabelSelector
    TopologySpreadConstraint:
      ...properties_goes_here

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: This could make sense, maybe starting only by shard and name.
wdyt?

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: Not really!
Because there are no default labels, it's filled by the user

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: Yeah, I believe those are the common use cases :)

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: aah ok

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: NVM! I was doing something wrong.

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: I've been looking on other part of codes and looks like it would only be nil if we mark it as a pointer, default array values are empty in go if I'm not wrong.

ArthurSens: Hmmm, I might be seeing things incorrectly, to be honest... I think empty slices when compared to nil will return true, but they won't return nilpointer exception when accessed?

nicolastakashi: Looks like yes!
I wrote a unit test to ensure the behavior as you can see.
https://github.com/prometheus-operator/prometheus-operator/pull/5967/files#diff-e26bfd67af0e336825679ce12693ffeba2d292a7139d14bc20a18ea01af2fb05R212

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +57,35 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
+// +kubebuilder:validation:Enum=name;shard
+type AdditionalLabelSelector string
+
+type TopologySpreadConstraint struct {
+	*v1.TopologySpreadConstraint `json:",inline"`
+
+	//+optional
+	// +listType=atomic
+	// Defines what Prometheus Operator managed labels should be added to labelSelector on the topologySpreadConstraint.
+	AdditionalLabelSelectors []AdditionalLabelSelector `json:"additionalLabelSelectors,omitempty"`
```
simonpasquier, line 69: Re-reading the discussion on the issue, I wonder if the operator's API shouldn't expose an enum value rather than a list?

````
  topologySpreadConstraints:
  - maxSkew: 2
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
    # ByShard => automatically inject label selectors matching all pods in the shard (e.g. name + shard).
    # ByShard => automatically inject label selectors matching all pods of the resource.
   # None => does nothing.
    additionalLabelSelectors: ByShard|ByResource|None
````

But I don't have great ideas for naming the field...

Comment (nicolastakashi) [reply pkg/apis/monitoring/v1/prometheus_types.go:69]: Hmmm 🤔 
I guess it might make the implementation and evolution a bit complex, if for some reason we introduce a new managed label which might make sense add to the topology, we have to think and new combinations.

Using a list of enum as it is we left the combination strategy to the user.

File: pkg/apis/monitoring/v1/prometheus_types.go [hunk 6]
```diff
@@ -56,6 +57,35 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
+}
+
+func (tp TopologySpreadConstraint) BuildK8sTopologySpreadConstraint(podLabels map[string]string) v1.TopologySpreadConstraint {
```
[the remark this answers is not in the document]
nicolastakashi: Changed, is the new place a good place? @simonpasquier

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,18 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
 }
 
+// +kubebuilder:validation:Enum=name;shard
+type AdditionalLabelSelector string
+
+type TopologySpreadConstraint struct {
+	*v1.TopologySpreadConstraint `json:",inline"`
+
+	//+optional
+	// +listType=atomic
```
simonpasquier, line 66:
```suggestion
	// +listType=set
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,18 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
 	return l.Status
 }
 
+// +kubebuilder:validation:Enum=name;shard
```
simonpasquier, line 59: only a suggestion but I feel that name and shard are a bit vague.
```suggestion
// +kubebuilder:validation:Enum=ResourceName;ShardName
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,18 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
+type AdditionalLabelSelector string
+
+type TopologySpreadConstraint struct {
+	*v1.TopologySpreadConstraint `json:",inline"`
```
simonpasquier, line 63: hmm why a pointer?

nicolastakashi: By mistake

nicolastakashi: When I'm not using a pointer I'm getting those errors.

```# github.com/prometheus-operator/prometheus-operator/pkg/client/applyconfiguration/monitoring/v1
pkg/client/applyconfiguration/monitoring/v1/topologyspreadconstraint.go:42:14: cannot use &value (value of type *int32) as int32 value in assignment
pkg/client/applyconfiguration/monitoring/v1/topologyspreadconstraint.go:50:18: cannot use &value (value of type *string) as string value in assignment
pkg/client/applyconfiguration/monitoring/v1/topologyspreadconstraint.go:58:24: cannot use &value (value of type *"k8s.io/api/core/v1".UnsatisfiableConstraintAction) as "k8s.io/api/core/v1".UnsatisfiableConstraintAction value in assignment```

File: pkg/operator/types.go [hunk 10]
```diff
@@ -53,3 +53,25 @@ func MakeHostAliases(input []monitoringv1.HostAlias) []v1.HostAlias {
...
+
+func MakeK8sTopologySpreadConstraint(podLabels map[string]string, tsc monitoringv1.TopologySpreadConstraint) v1.TopologySpreadConstraint {
+	if len(tsc.AdditionalLabelSelectors) > 0 {
+		if tsc.LabelSelector == nil {
+			tsc.LabelSelector = &metav1.LabelSelector{
+				MatchLabels: make(map[string]string),
+			}
+		}
+
+		labelMap := map[monitoringv1.AdditionalLabelSelector]string{
```
nicolastakashi, line 65: @simonpasquier due to [this](https://github.com/prometheus-operator/prometheus-operator/pull/5967#discussion_r1362156621) suggestion, we have to keep a map from the CRD value to Prometheus Operator Labels.

I'm not against the idea, I'm just wondering if is there a better place to have this map and avoid the object creation on every reconcile loop.

nicolastakashi: @simonpasquier friendly ping 😄

simonpasquier: I wouldn't too worried about this but you can have a const instead.

pkg/operator/types.go
```diff
@@ -53,3 +53,25 @@ func MakeHostAliases(input []monitoringv1.HostAlias) []v1.HostAlias {
 
 	return output
 }
+
+func MakeK8sTopologySpreadConstraint(podLabels map[string]string, tsc monitoringv1.TopologySpreadConstraint) v1.TopologySpreadConstraint {
+	if len(tsc.AdditionalLabelSelectors) > 0 {
```
simonpasquier, line 58:
```suggestion
	if len(tsc.AdditionalLabelSelectors) == 0 {
		return *tsc.TopologySpreadConstraint
	}
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,18 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
 	return l.Status
 }
 
+// +kubebuilder:validation:Enum=ResourceName;ShardName
+type AdditionalLabelSelector string
```
simonpasquier, line 60: (nit) we can define the enum values as const and reuse in the controller.
```suggestion
type AdditionalLabelSelector string

const (
    ResourceNameLabelSelector AdditionalLabelSelector = "ResourceName"
    ShardNameLabelSelector AdditionalLabelSelector = "ShardName"
    
)
```

pkg/apis/monitoring/v1/prometheus_types.go
simonpasquier: Hmm I don't see the same thing :thinking:

pkg/operator/types.go
```diff
@@ -53,3 +53,25 @@ func MakeHostAliases(input []monitoringv1.HostAlias) []v1.HostAlias {
...
+			}
+		}
+
+		labelMap := map[monitoringv1.AdditionalLabelSelector]string{
+			"ShardName":    "operator.prometheus.io/shard",
+			"ResourceName": "operator.prometheus.io/name",
+		}
+
+		for _, label := range tsc.AdditionalLabelSelectors {
+			labelName := labelMap[label]
```
simonpasquier, line 71: As a precaution, we need to return an error if the label doesn't match an entry in the map.

pkg/apis/monitoring/v1/prometheus_types.go
simonpasquier: Is there a use case where you would define the following?

````
spec:
  topologySpreadConstraints:
   - additionalLabelSelectors: [ShardName]
     ...
````

With the current proposed API, it would be valid but from the original issue, it doesn't seem to be useful and almost always want `additionalLabelSelectors: [ShardName, ResourceName]` or `additionalLabelSelectors: [ResourceName]`. But I may be missing something.

File: pkg/prometheus/agent/statefulset_test.go [hunk 14]
```diff
@@ -206,3 +208,87 @@ func makeStatefulSetFromPrometheus(p monitoringv1alpha1.PrometheusAgent) (*appsv
 		0,
 		nil)
 }
+
+func TestPodTopologySpreadConstraintWithAdditionalLabels(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+		spec monitoringv1alpha1.PrometheusAgentSpec
```
simonpasquier, line 215: If possible, I would define the expected topology spread constraints in the test struct.
```suggestion
		spec monitoringv1alpha1.PrometheusAgentSpec
		expected []v1.TopologySpreadConstraint
```

simonpasquier (commented):
Thanks and sorry for the late review!

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: After remove the `*` mark and run make generate I got this issue.

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: Actually!
After think a bit more I didn't noticed any usecase.
I just implemented your idea of having ShardNameAndResource and ResourceName only.

pkg/operator/types.go
nicolastakashi: Approach is different now.

pkg/operator/types.go
```diff
@@ -53,3 +53,31 @@ func MakeHostAliases(input []monitoringv1.HostAlias) []v1.HostAlias {
 
 	return output
 }
+
+const ShardLabel = "operator.prometheus.io/shard"
+const ResourceNameLabel = "operator.prometheus.io/name"
```
ArthurSens, line 58: Could we import the constants from the prometheus package instead? 

https://github.com/prometheus-operator/prometheus-operator/blob/79502954f995929c8b3511d533a2e34bb3ee40ca/pkg/prometheus/statefulset.go#L60-L61

pkg/prometheus/agent/statefulset_test.go
```diff
@@ -206,3 +210,146 @@ func makeStatefulSetFromPrometheus(p monitoringv1alpha1.PrometheusAgent) (*appsv
...
+			},
+			tsc: v1.TopologySpreadConstraint{
+				MaxSkew:           1,
+				TopologyKey:       "kubernetes.io/hostname",
+				WhenUnsatisfiable: v1.DoNotSchedule,
+				LabelSelector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"app":                           "prometheus",
+						prompkg.ShardLabelName:          "0",
+						prompkg.PrometheusNameLabelName: "",
```
ArthurSens, line 301: is it a problem to have empty label selectors?

ArthurSens (commented):
Looking good! Just a question

pkg/operator/types.go
nicolastakashi: I tried but we got a cyclical dependency 😢

pkg/prometheus/agent/statefulset_test.go
nicolastakashi: For some reason this is empty on tests, but I guess it will never be empty on real life 🤔

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: @simonpasquier any idea?

simonpasquier: Have you tried to delete the generated Go files first?

pkg/operator/types.go
```diff
@@ -53,3 +53,31 @@ func MakeHostAliases(input []monitoringv1.HostAlias) []v1.HostAlias {
...
+
+func MakeK8sTopologySpreadConstraint(podLabels map[string]string, tsc monitoringv1.TopologySpreadConstraint) v1.TopologySpreadConstraint {
```
simonpasquier, line 60: Don't we need to match also on the `app.kubernetes.io/name` label to differentiate between server and agent pods?

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -56,6 +56,25 @@ func (l *Prometheus) GetStatus() PrometheusStatus {
...
+	// Automatically add a label selector that will select all pods matching the same shard.
+	ShardAndResourceNameLabelSelector AdditionalLabelSelector = "ShardAndResourceName"
+)
+
+type TopologySpreadConstraint struct {
+	*v1.TopologySpreadConstraint `json:",inline"`
+
+	//+optional
+	// Defines what Prometheus Operator managed labels should be added to labelSelector on the topologySpreadConstraint.
+	AdditionalLabelSelectors *AdditionalLabelSelector `json:"additionalLabelSelectors,omitempty"`
```
simonpasquier, line 75: I'm still wondering about the best naming for this field :thinking:

pkg/apis/monitoring/v1/prometheus_types.go
simonpasquier: I think that I managed to workaround the issue by defining a few intermidiate structs:

```go
type TopologySpreadConstraint struct {
       CoreV1TopologySpreadConstraint `json:",inline"`
 
        //+optional
        // Defines what Prometheus Operator managed labels should be added to labelSelector on the topologySpreadConstraint.
        AdditionalLabelSelectors *AdditionalLabelSelector `json:"additionalLabelSelectors,omitempty"`
 }

type CoreV1TopologySpreadConstraint v1.TopologySpreadConstraint
```

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: Me too 😢 
But I ran out of ideas.

pkg/operator/types.go
nicolastakashi: Updated!
Can you check it?

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: ManagedLabelsToLabelSelector?

I don't think we'll find the perfect name here 😅 , `additionalLabelSelectors` is also fine for me

ArthurSens (approved):

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: I'm fine with both!
@simonpasquier any other feedback about this PR?

pkg/prometheus/statefulset.go
```diff
@@ -427,3 +428,29 @@ func BuildPodMetadata(cpf monitoringv1.CommonPrometheusFields, cg *ConfigGenerat
...
+	}
+
+	if tsc.LabelSelector == nil {
+		tsc.LabelSelector = &metav1.LabelSelector{
+			MatchLabels: make(map[string]string),
+		}
+	}
+
+	if *tsc.AdditionalLabelSelectors == monitoringv1.ResourceNameLabelSelector {
+		for key, value := range podLabels {
```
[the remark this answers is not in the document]
nicolastakashi: Sorry!
Could you come again @simonpasquier ?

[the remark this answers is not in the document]
simonpasquier: To be clearer, I assume that this function needs to get the [finalSelectorLabels](https://github.com/prometheus-operator/prometheus-operator/blob/88e86c5caf84dc85338c904e13b0656bf1b56a67/pkg/prometheus/server/statefulset.go#L359) map, not `finalLabels`:
* For `OnShard` it should copy all the keys
* For `OnResource`, it should copy all the keys minus `operator.prometheus.io/shard`.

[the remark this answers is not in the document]
nicolastakashi: Ok!
I guess I got it.
Can you check @simonpasquier ?

pkg/prometheus/agent/statefulset.go
```diff
@@ -391,7 +396,7 @@ func makeStatefulSetSpec(
 				Volumes:                       volumes,
 				Tolerations:                   cpf.Tolerations,
 				Affinity:                      cpf.Affinity,
-				TopologySpreadConstraints:     cpf.TopologySpreadConstraints,
+				TopologySpreadConstraints:     tscs,
```
simonpasquier, line 399: (nit) we could factorize code a bit:
```suggestion
				TopologySpreadConstraints:     prompkg.MakeK8sTopologySpreadConstraints(finalSelectorLabels, cpf.TopologySpreadConstraints)),
```

pkg/prometheus/server/statefulset.go
```diff
@@ -463,6 +463,11 @@ func makeStatefulSetSpec(
 		return nil, fmt.Errorf("failed to merge containers spec: %w", err)
 	}
 
+	// tscs := make([]v1.TopologySpreadConstraint, 0, len(cpf.TopologySpreadConstraints))
+	// for _, tsc := range cpf.TopologySpreadConstraints {
+	// 	tscs = append(tscs, prompkg.MakeK8sTopologySpreadConstraint(finalSelectorLabels, tsc))
+	// }
+
```
simonpasquier, line 470:
```suggestion
```

pkg/prometheus/statefulset.go
```diff
@@ -496,3 +497,35 @@ func ShareProcessNamespace(p monitoringv1.PrometheusInterface) *bool {
 		) == monitoringv1.ProcessSignalReloadStrategyType,
 	)
 }
+
+func MakeK8sTopologySpreadConstraint(selectorLabels map[string]string, tscs []monitoringv1.TopologySpreadConstraint) []v1.TopologySpreadConstraint {
+
+	coreTscs := make([]v1.TopologySpreadConstraint, 0, len(tscs))
+	// for _, tsc := range cpf.TopologySpreadConstraints {
+	// 	tscs = append(tscs, prompkg.MakeK8sTopologySpreadConstraint(finalSelectorLabels, tsc))
+	// }
```
simonpasquier, line 506:
```suggestion
```

pkg/prometheus/statefulset.go
```diff
@@ -496,3 +497,35 @@ func ShareProcessNamespace(p monitoringv1.PrometheusInterface) *bool {
 		) == monitoringv1.ProcessSignalReloadStrategyType,
 	)
 }
+
+func MakeK8sTopologySpreadConstraint(selectorLabels map[string]string, tscs []monitoringv1.TopologySpreadConstraint) []v1.TopologySpreadConstraint {
+
```
simonpasquier, line 502: (nit) fast path
```suggestion
    if len(tscs) == 0 {
         return nil
     }

```

pkg/prometheus/statefulset.go
nicolastakashi: ups

simonpasquier (approved):
Thanks!

Issue #5947, opened 2023-09-26
Support additional relabelings in scrape classes


SuperQ:
**What is missing?**

With the new [scrape classes](https://github.com/prometheus-operator/prometheus-operator/pull/5572) feature, it would be useful to allow additional relabelings that are added after the operator-managed relabelings, but before the relabelings supplied by the scrape target (`ServiceMonitor`, `PodMonitor`, etc)

This is an extension of the request in https://github.com/prometheus-operator/prometheus-operator/issues/3255.

**Why do we need it?**

This is needed in order to allow `Prometheus` admins to extend and enforce relabelings for downstream users of various scrape configs.

Examples:

Always propagate the `deployment` label:
```yaml
kind: Prometheus
spec:
  scrapeClasses:
    - name: deployments
      default: true
      additionalRelabelings:
        - sourceLabels: [__meta_kubernetes_pod_label_deployment]
          targetLabel: deployment
```

Always propagate release information provided by CD tooling:
```yaml
kind: Prometheus
spec:
  scrapeClasses:
    - name: versioning
      default: true
      additionalRelabelings:
        - sourceLabels: [__meta_kubernetes_pod_annotation_example_com_version]
          targetLabel: release_version
```

rexagod:
WIP.

nicolastakashi:
@rexagod Tge ScrapeClasse design is not finished yet.

rexagod:
Ah, I see. I guess I'll wait it out then, thanks for the heads up!

sathieu:
I have a different use-case: We define a tenant as a list of namespaces (or namespace regexps). I want to create a tenant label for each metric.

I need something like:

```yaml
kind: Prometheus
spec:
  scrapeClasses:
    - name: deployments
      default: true
      additionalMetricsRelabelings:
        - sourceLabels: [namespace]
          regex: tenant1-.*
          targetLabel: tenant
          replacement: tenant1
        - sourceLabels: [namespace]
          regex: foo|bar
          targetLabel: tenant
          replacement: foo-bar
```

This metrics relabeling should be enforced **after** scrape target metrics relabeling. Also if the scrape target references an unexisting scrapeClass, the default scrape class should be applied (for security reasons).

simonpasquier:
@sathieu this is an interesting use case but I'm not sure that scrape classes (as they are currently envisioned #6113) are the right option. The design proposal calls out the following non-goal:

[...]

PR #5972, opened 2023-10-06
Add node relabeling to pod and servicemonitors

QuentinBisson:
## Description

This PR enhances the pod and service monitors by automatically relabelling the node label as having the node a pod is deployed could help pinpoint issues happening on one node

## Type of change

## Changelog entry

```release-note
Add node label relabelling to pod and service monitors.
```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -801,6 +801,10 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 			{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_name"}},
 			{Key: "target_label", Value: "pod"},
 		},
+		{
+			{Key: "source_labels", Value: []string{"__meta_kubernetes_pod_node_name"}},
+			{Key: "target_label", Value: "node"},
+		},
 	}...)
 
 	// Relabel targetLabels from Pod onto target.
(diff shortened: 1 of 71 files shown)


paulfantom:
This also increases metrics cardinality since label set can be easily mutated by pod being rescheduled to a different node.

nicolastakashi:
I do agree this is a useful information, but having this automatically might increase users cardinality without their concern.

For me this kind of changing should be user intention and a could solution for that would be the additionalRelabeling on the Scrape Class feature.

https://github.com/prometheus-operator/prometheus-operator/issues/5947

WDYT?

QuentinBisson:
Considering this is as volatile (a change per pod rescheduling) as the
instance label, would this really increase metric cardinality?

Now considering scrape class is still at the proposal level, I would rather
work on adding default relabelings on all pod/service monitors using a
property like pod monitors DefaultRelabelings for now and maybe that could
then be specified at the scrape class level? WDYT?

Le sam. 7 oct. 2023, 13:14, Nicolas Takashi ***@***.***> a
écrit :

QuentinBisson:
This PR shows what I was talking about in my previous comment https://github.com/prometheus-operator/prometheus-operator/pull/5978.

Using this, it would be up to the user to configure it on the Prometheus level

ArthurSens:
Hi Quentin, thanks for your interest in the project! We really appreciate that you're eager to contribute solutions to your problems upstream :)

This particular change, however, we can't accept it 😕. As mentioned by paulfantom and nicolas, this increases cardinality and for sure it would break several installations around the globe.

We'll follow up on the discussion elsewhere

QuentinBisson:
Hi @ArthurSens no hard feelings :)
I proposed [another PR](https://github.com/prometheus-operator/prometheus-operator/pull/5978) which will solve my issue in a more flexible way (while also bringing flexibility to prometheus default relabelings) while not increasing cardinality without a conscious user choice

Issue #3255, opened 2020-05-29
how to add a customize metric relabelling  for all Servicemonitor and PodMonitor


sunhubs:
**What did you do?**
I want to add customize metric relabelling for all the scrape jobs generated by servicemoitor and podmonitor. For example i want to relable [__meta_kubernetes_pod_label_app] to [app].
currently, i have to set for all the ServiceMonitor and PodMonitor yaml file to add:
`
relabelings:

    - sourceLabels: [__meta_kubernetes_pod_label_app]

      targetLabel: app
`
i think it is a bit annoying
**Did you expect to see some different?**
i just want to have a way to reduce this effort, for example a configuration to let the auto generated scrape_config contains the custom relabeling for all ServiceMonitor and PodMonitor as those default ones.
**Environment**
helm:v3.2.1
prometheus:v2.16.0

* Prometheus Operator version:
v0.38.1
    

* Kubernetes version information:
client 1.14.7
server 1.16.8
    

* Kubernetes cluster kind:
   AWS eks
    insert how you created your cluster: kops, bootkube, etc.

* Manifests:
No more need
```
insert manifests relevant to the issue
```

* Prometheus Operator Logs:
No need logs
```
insert Prometheus Operator logs relevant to the issue here
```

**Anything else we need to know?**:
No

brancz:
This doesn’t exist but I think it’s an interesting feature to consider, especially when consistent practices across an organization exist

sunhubs:
Thanks @brancz , but may I know, when will this be released?

brancz:
Whenever someone implements it :)

PR #5978, opened 2023-10-07
Allow extra relabelings to all service and pod monitors when scraped by Prometheus

QuentinBisson:
## Description

Fixes https://github.com/prometheus-operator/prometheus-operator/issues/3255

This PR allows users of the operator to ensure a set of custom relabelings to all their pod/service monitors.

This would supersed this PR https://github.com/prometheus-operator/prometheus-operator/pull/5972

Notable examples would be adding the node label using __meta_kubernetes_pod_node_name but it could be used for other organization purposes such as:

```
- sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
  action: "replace"
  targetLabel: app
```

This feature could in the future be added to Scrape Classes as well

## Type of change

## Changelog entry

```release-note
Allow addition of extra relabelings to all service and pod monitors
```

### DIFF
--- a/pkg/client/applyconfiguration/monitoring/v1/prometheusspec.go
+++ b/pkg/client/applyconfiguration/monitoring/v1/prometheusspec.go
@@ -646,6 +646,19 @@ func (b *PrometheusSpecApplyConfiguration) WithPodTargetLabels(values ...string)
 	return b
 }
 
+// WithExtraRelabelings adds the given value to the ExtraRelabelings field in the declarative configuration
+// and returns the receiver, so that objects can be build by chaining "With" function invocations.
+// If called multiple times, values provided by each call will be appended to the ExtraRelabelings field.
+func (b *PrometheusSpecApplyConfiguration) WithExtraRelabelings(values ...**monitoringv1.RelabelConfig) *PrometheusSpecApplyConfiguration {
+	for i := range values {
+		if values[i] == nil {
+			panic("nil value passed to WithExtraRelabelings")
+		}
+		b.ExtraRelabelings = append(b.ExtraRelabelings, *values[i])
+	}
+	return b
+}
+
 // WithTracingConfig sets the TracingConfig field in the declarative configuration to the given value
 // and returns the receiver, so that objects can be built by chaining "With" function invocations.
 // If called multiple times, the TracingConfig field is set to the value of the last call.
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -2389,6 +2389,21 @@ PodMonitor and ServiceMonitor objects.</p>
 </tr>
 <tr>
 <td>
+<code>extraRelabelings</code><br/>
+<em>
+<a href="#monitoring.coreos.com/v1.RelabelConfig">
+[]RelabelConfig
+</a>
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>ExtraRelabelings allow setting extra default relabelings to all
+PodMonitor and ServiceMonitor objects targeted by Prometheus.</p>
+</td>
+</tr>
+<tr>
+<td>
 <code>tracingConfig</code><br/>
 <em>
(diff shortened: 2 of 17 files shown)


nicolastakashi:
Thanks @QuentinBisson for this PR!
I personally have mixed feelings about this, because I think it should be part of scrape class as we already have proposed.
I'm curious to what the rest of the team thinks about this.

QuentinBisson:
I do think it makes sense to be able to configure this at the scrape class level but what if you want to enable it at an individual prometheus level ? You would need to configure a custom scrape class for this Prometheus ?

Also, as mentionned in my other PR, scrape classes are still at the proposal level where this would fix a real issue of ours right now. 

I would think the addition of scrape classes will make us transition to some kind of v2 in the Prometheus API so maybe that could be addressed then ?

nicolastakashi:
I'm thinking of a more long-term goal, I mean adding a field to the CRD which will be deprecated very soon in favor of a feature that is under design, it might introduce a lot of complexity on the maintenance side, because we will need to introduce deprecations, breaking changes and maintain "duplicated" features.

But I do understand your POV and I'm not totally against this approach, it might be a solution indeed, just wondering if looking for a long-term makes sense since the Scrape Class proposal is almost proposed.

> I think it makes sense to configure this at the scrape class level, but what if you want to enable it at an individual prometheus level? Would you need to configure a custom scrape class for this Prometheus?

Yes, from my understanding the Scrape Classe feature will be placed on the Prometheus CR level so that you can have different configs for different Prometheus.

QuentinBisson:
Hello @nicolastakashi I fully understand your point about operator maintenance and I did not know it was coming so soon.
Thinking about it over the weekend, I still think the Prometheus CR should have this property as I would envision a lot of organisations setting relabelings at the prometheus level for organizationnal stuff and maybe add custom relabelings at the scrape class level.
Also, having a global field at the prometheus level would avoid the need to set the same relabelings across all scrape classes.

For instance, organizations could add the following default at the prometheus level:
- organization label
- app label
- cluster label
- other organizational labels

At the scrape class level, depending on your use case you could:
- Have a scrape class per team with a different relabeling for the team label so teams would be able to just use the scrape class and not care about setting the team label
- There are probably a lot of other labels that would make sense here but the team one was the only one I could think of 😓 

But as you said, maybe let's check if other people have other ideas :)


BTW, i'd be happy to add the code for the relabelings on the scrape class level as well once it's out

QuentinBisson:
@ArthurSens and @paulfantom as you both were on my previous PR, would you mind taking a look at this PR when you have some free time ? I would love your extra input :)

peter-crist:
I'm extremely interested in this as well. I have a single prometheus instance and alerting solely through Grafana. 

If you're not setup as a federated/remote-storage configuration, or alerting from Alertmanager, external labels don't help you and it is quite difficult to get cluster information available in Grafana. The ability to add a specific label to every metric would solve this problem. Setting up this relabel config on service/podmonitors individually is not maintainable.

Additionally, and outside the scope of this issue, Grafana doesn't provide a way of provisioning a global variable that can be used on all dashboards. This means, the solution would have to come from applying the label to the metric itself as opposed to decorating it in Grafana.

This would go a long way in the simple challenge of displaying cluster information alongside our metrics.

pkg/client/applyconfiguration/monitoring/v1/prometheusspec.go
```diff
@@ -646,6 +646,19 @@ func (b *PrometheusSpecApplyConfiguration) WithPodTargetLabels(values ...string)
 	return b
 }
 
+// WithExtraRelabelings adds the given value to the ExtraRelabelings field in the declarative configuration
+// and returns the receiver, so that objects can be build by chaining "With" function invocations.
```
peter-crist, line 650: nit: `so that objects can be built by`

QuentinBisson:
Closing this PR in favor of https://github.com/prometheus-operator/prometheus-operator/pull/6379 in scrape classes

PR #5981, opened 2023-10-09
fix: set verbosity log level only for debug

simonpasquier:
## Description

client-go functions can avoid costly logging evaluations by using this pattern:

```
if klog.V(2).Enabled() { klog.Info("log this") }
```

With this change, the klog verbosity is set to zero (except at debug log level where it is 6) to ensure that the performance optimization is preserved.

## Type of change

## Changelog entry

```release-note
Enable klog verbose level only when the log level is debug.
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -42,7 +42,6 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


simonpasquier:
Example of client-go function using conditional logging:

https://github.com/kubernetes/client-go/blob/6030d798e44096d5ffb58e00802cadd81a74dcf0/rest/request.go#L1205-L1215

ArthurSens (approved):
I'm not entirely sure I understood the problem, but the changes look fine!

PR #5988, opened 2023-10-09
.github/workflows: skip unit and e2e tests for *.md file updates

slashpai:
## Description

**Skip unit test and e2e test for \*.md update**

Ref:

- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,6 +1,8 @@
 name: e2e
 on:
   pull_request:
+    paths-ignore:
+      - '**/*.md'
   push:
     branches:
       - 'release-*'
@@ -8,6 +10,8 @@ on:
       - 'main'
     tags:
       - 'v*'
+    paths-ignore:
+      - '**/*.md'
 jobs:
   e2e-tests:
     name: E2E tests
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -1,6 +1,8 @@
 name: unit
 on:
   pull_request:
+    paths-ignore:
+      - '**/*.md'
   push:
     branches:
       - 'release-*'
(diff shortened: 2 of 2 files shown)


nicolastakashi (approved):

ArthurSens:
We run mdox for format checking on those docs, don't we?

slashpai:
> We run mdox for format checking on those docs, don't we?

Ya you are right just noticed that. I was trying to avoid running all tests for a simple doc change.

I think we might be able to use https://github.com/dorny/paths-filter for running specific job based on filter.

nicolastakashi:
We can move the mdocs for a dedicated workflow and have something like that filtering by path.
https://github.com/coralogix/coralogix-operator/blob/main/.github/workflows/helm-chart-deploy.yml#L8
Or file extension.

slashpai:
I updated the PR to ignore running unit test and e2e tests which consumes lot of time when updates are done to *.md

.github/workflows/e2e.yaml
```diff
@@ -1,13 +1,17 @@
 name: e2e
 on:
   pull_request:
+    paths-ignore:
+      - '**/*.md'
```
slashpai, line 5: ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

slashpai:
@ArthurSens @nicolastakashi @simonpasquier could you please take a look

simonpasquier (approved):

simonpasquier:
Thanks!

PR #5993, opened 2023-10-11
feat: optimize memory usage of secret/configmap informers

simonpasquier:
## Description

The secrets and configmaps informers only need to store the name, namespace and resource version of the objects: the event handlers don't look at the other metadata fields. By stripping the rest of the metadata, we can save a fair share of memory on clusters with lots of secrets/configmaps.

Example with a kind cluster running 16,384 secrets (in 128 namespaces), each secret has 4 annotations of 1,024 chars each.

## Before the change

image

## After this change

image

In all cases, we see a big spike at the start of the operator which is expected because it fetches all data for the informers from the Kubernetes API. Depending on when the GC kicks in, the max allocated memory can remain high for longer:

image

The maximum memory usage with/without the change is similar but once the situation settles, we see a ~80% decrease in memory usage with this PR (from 300MB -> 50MB).

To take full advantage of this optimization, we can also tune the GOGC variable to reclaim the memory faster after startup.

With `GOGC=50`:

image

With `GOGC=30`:

image

I didn't observe significant changes to the CPU usage.

The optimization has been inspired by https://github.com/cert-manager/cert-manager/pull/5966

## Type of change

## Changelog entry

```release-note
Reduce memory usage for clusters with large number of secrets and/or configmaps.
```

### DIFF
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -10,6 +10,7 @@ local defaults = {
     requests: { cpu: '', memory: '' },
   },
   enableReloaderProbes: false,
+  goGC: '30',
   port: 8080,
   resources: {
     limits: { cpu: '200m', memory: '200Mi' },
@@ -168,6 +169,7 @@ function(params) {
         name: 'http',
       }],
       resources: po.config.resources,
+      env: [{ name: 'GOGC', value: po.config.goGC }],
       securityContext: {
         allowPrivilegeEscalation: false,
         readOnlyRootFilesystem: true,
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -42063,6 +42063,9 @@ spec:
       - args:
         - --kubelet-service=kube-system/kubelet
         - --prometheus-config-reloader=quay.io/prometheus-operator/prometheus-config-reloader:v0.69.1
+        env:
+        - name: GOGC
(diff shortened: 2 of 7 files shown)


jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -10,6 +10,7 @@ local defaults = {
     requests: { cpu: '', memory: '' },
   },
   enableReloaderProbes: false,
+  goGC: '30',
```
philipgough, line 13: As per https://pkg.go.dev/runtime the default is 100. Should we retain the default value here while allowing it to be configurable or is the idea that we try (and push users to) to take advantage of the optimisation?

philipgough (approved):
One question on the default setting but other than that lgtm. Awesome improvement.

jsonnet/prometheus-operator/prometheus-operator.libsonnet
simonpasquier: The latter. My observation is that the memory usage increases a lot initially when the operator populates its local cache but it's transient because most of the data is discarded by the transform func and reclaiming more aggressively than the default is beneficial in this case.

philipgough: Yeah makes sense. Feels pretty low risk so all good with me :)

nicolastakashi (approved):

PR #5998, opened 2023-10-12
Feat: Add field PropagateOwnerReferences to PersistentVolumeClaim

wallee94:
This change adds a bool field `propagateOwnerReferences` in `volumeClaimTemplate`. If set to true, the operator will copy the ownerReferences set in a StatefulSet to the PersistenVolumeClaims created by it.

## Description

Different solution for the problem described in https://github.com/prometheus-operator/prometheus-operator/pull/5899#issuecomment-1717541223

## Type of change

## Changelog entry

```release-note
feat: Add PropagateOwnerReferences field to EmbeddedPersistentVolumeClaim
```

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -7143,6 +7143,19 @@ Kubernetes core/v1.PersistentVolumeClaimStatus
 <p><em>Deprecated: this field is never set.</em></p>
 </td>
 </tr>
+<tr>
+<td>
+<code>propagateOwnerReferences</code><br/>
+<em>
+bool
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>Controls whether the ownerReferences in the StatefulSet metadata should be set as ownerReferences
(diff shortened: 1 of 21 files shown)


simonpasquier (commented):
I thought that you wanted to copy the Alertmanager/Prometheus/... owners to the PVC metadata, not the statefulset.

wallee94:
Right. I wanted to set the reference `namespace < PVC`. But working on it, it made sense to me to use `namespace < Prometheus < PVC`, to make the operator set the ownerReferences to the resource that created it, like in the StatefulSet.

I think both are similar. I can change it back to copy Prometheus owners, or change the field name.

wallee94:
I think it does. The only potential issue I had in mind was resizing volumes, as the operator recreates the sts, but I just found out that "--cascade=orphan" preserves the PVCs even if the retention policy is set to Delete. So I think that works as long as we follow the steps in the docs
https://prometheus-operator.dev/docs/operator/storage/#resizing-volumes

Thank you @simonpasquier! I'm closing the PR.

Issue #5905, opened 2023-09-11
Alertmanager pod can't be created when more than 1 template is specified in `.spec.alertmanagerConfiguration.templates`


dntosas:
**What happened?**
AlertManager its trying to mount templates volume twice
```
Warning   FailedCreate       statefulset/alertmanager-prometheus-operator-alertmanager   create Pod alertmanager-prometheus-operator-alertmanager-0 in StatefulSet alertmanager-prometheus-operator-alertmanager failed error: Pod "alertmanager-prometheus-operator-alertmanager-0" is invalid: spec.containers[0].volumeMounts[6].mountPath: Invalid value: "/etc/alertmanager/templates": must be unique
```

When applied this block of config:
```
    configuration:
      name: global
      global:
        resolveTimeout: 5m
        slackApiUrl:
          name: alertmanager-slack-legacy-webhook
          key: webhookUrl
      templates:
        - configMap:
            name: alertmanager-templates
            key: slack-default.tmpl
```
**Did you expect to see something different?**
Same behavior as before.

**How to reproduce it (as minimally and precisely as possible)**:

**Environment**

* Prometheus Operator version:
0.68

* Kubernetes version information:
1.26

* Kubernetes cluster kind:

    insert how you created your cluster: kops, bootkube, etc.

* Manifests:

```
insert manifests relevant to the issue
```

* Prometheus Operator Logs:

```
insert Prometheus Operator logs relevant to the issue here
```

**Anything else we need to know?**:

slashpai:
Was this working correctly in `v0.67`?

If possible can you post operator logs as well?

dntosas:
for some magic reason, deleted and re-created the resource, and everything played as expected 🤷 

re-produced the same behavior in 15 instances

simonpasquier:
I can reproduce the issue with the following spec:

```
spec:
  alertmanagerConfiguration:
    name: config-example
    templates:
    - configMap:
        key: key1
        name: template1
    - configMap:
        key: key1
        name: template2

```

simonpasquier:
Currently the operator generates the following pod template snippet in the statefulset:

```
   volumes:
     ...
      - name: notification-templates
        projected:
          defaultMode: 420
          sources:
          - configMap:
              items:
              - key: key1
                path: key1
              name: template1
          - configMap:
              items:
              - key: key1
                path: key1
              name: template2
```

[...]

PR #6006, opened 2023-10-14
fix: handle duplicate sources

rexagod:
## Description

Skip appending sources with the same key for ConfigMaps and Secrets in the AlertmanagerConfiguration.

Fixes: #5905

## Type of change

### DIFF
--- a/pkg/alertmanager/statefulset.go
+++ b/pkg/alertmanager/statefulset.go
@@ -534,12 +534,11 @@ func makeStatefulSetSpec(logger log.Logger, a *monitoringv1.Alertmanager, config
 	amCfg := a.Spec.AlertmanagerConfiguration
 	if amCfg != nil && len(amCfg.Templates) > 0 {
 		sources := []v1.VolumeProjection{}
-		configmapKeys := sets.Set[string]{}
-		secretKeys := sets.Set[string]{}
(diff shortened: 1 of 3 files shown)


pkg/alertmanager/statefulset.go
```diff
@@ -531,8 +534,14 @@ func makeStatefulSetSpec(a *monitoringv1.Alertmanager, config Config, tlsAssetSe
 	amCfg := a.Spec.AlertmanagerConfiguration
 	if amCfg != nil && len(amCfg.Templates) > 0 {
 		sources := []v1.VolumeProjection{}
+		configmapKeys := sets.Set[string]{}
+		secretKeys := sets.Set[string]{}
```
ArthurSens, line 538: If a secret and a configmap have the same name, I believe we'll also have conflicting paths, is that correct?

Don't we need to use the same set for both?

ArthurSens (commented):
Thanks for the PR!

Could we also have a unit test to make sure we don't mount duplicated paths?

pkg/alertmanager/statefulset.go
rexagod: Hmm, that makes sense. Patched!

ArthurSens (approved):
Thanks!

PR #6007, opened 2023-10-14
fix: validate route matchers

rexagod:
## Description

Validate route matchers for AlertmanagerConfigurations, so they don't crash the main instance.

Fixes: #5163

## Type of change


pkg/alertmanager/validation/v1alpha1/validation.go
```diff
@@ -374,6 +375,14 @@ func validateAlertManagerRoutes(r *monitoringv1alpha1.Route, receivers, muteTime
 		}
 	}
 
+	// Validate matchers, so they don't end up crashing the main instance: https://github.com/prometheus/alertmanager/blob/v0.26.0/config/coordinator.go#L124.
+	for _, matcher := range r.Matchers {
```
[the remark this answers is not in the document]
simonpasquier: yes correct.

simonpasquier (commented):
I realize that there are more fields referring to label names that would need the same validation (e.g. InhibitRule, GroupBy)

simonpasquier:
Also we have Validate() methods in pkg/apis/monitoring/{v1beta1,v1alpha1} packages but they don't seem to be used anywhere :thinking: 

https://github.com/prometheus-operator/prometheus-operator/blob/75e3dd982046bf237da50c4e696d628ac1c1046d/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L950-L965

pkg/alertmanager/validation/v1alpha1/validation.go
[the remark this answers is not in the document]
rexagod: ACK, so it's safe to say that `v1` versions will always be available alongside `alpha` and/or `beta` (since otherwise, the implicit version upgrade won't take place, and this functionality will be left out)? I'm assuming that's the case and moving this to `v1` based on the conversation above. 👍🏼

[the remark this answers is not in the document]
rexagod: I see a lot of code in `alpha` and `beta` has overlapping logic as well. What if we merged the two under `v1` and used interface parameters, getting rid of them entirely?

rexagod:
> Also we have Validate() methods in pkg/apis/monitoring/{v1beta1,v1alpha1} packages but they don't seem to be used anywhere 🤔

IIUC they seem to be used in `checkInhibitRules`.

https://github.com/prometheus-operator/prometheus-operator/blob/175887b844dd5a8818cc84704585933fdbefc385/pkg/alertmanager/operator.go#L1649-L1655

pkg/alertmanager/validation/v1alpha1/validation.go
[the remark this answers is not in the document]
nicolastakashi: I'm not against at all, but I'd suggest to do this on a dedicated PR.

rexagod:
> I realize that there are more fields referring to label names that would need the same validation (e.g. InhibitRule, GroupBy).

Pushed a commit for `checkInhibitRule` (which uses `Validate` internally). For `GroupBy`, I saw that this was a `[]string` type (instead of `[]Matcher`). Are all `GroupBy` values `Matcher` values too? If so, should I change its type to `[]Matcher` in order to leverage the additional validation support?

simonpasquier:
> Are all `GroupBy` values `Matcher` values too? If so, should I change its type to `[]Matcher` in order to leverage the additional validation support?

No because `...` is a valid GroupBy value but not a valid label name.

pkg/admission/admission.go
```diff
@@ -293,7 +293,7 @@ func (a *Admission) validateAlertmanagerConfig(ar v1.AdmissionReview) *v1.Admiss
 		msg := "invalid config"
 		level.Debug(a.logger).Log("msg", msg, "content", string(ar.Request.Object.Raw))
 		level.Info(a.logger).Log("msg", msg, "err", err)
-		return toAdmissionResponseFailure("AlertmanagerConfig is invalid", alertManagerConfigResource, []error{err})
+		return toAdmissionResponseFailure(fmt.Sprintf("AlertmanagerConfig is invalid: %v", err), alertManagerConfigResource, []error{err})
```
simonpasquier, line 296: hmm why repeat the error in the message?

pkg/apis/monitoring/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.28.3
 	k8s.io/apimachinery v0.28.3
 	sigs.k8s.io/controller-runtime v0.16.3
+	github.com/prometheus/alertmanager v0.26.0
```
simonpasquier, line 10: i wouldn't make this module depend on `github.com/prometheus/alertmanager`. It should have as few dependencies as possible so folks can import it easily.

To do so, I'd create a dedicated func in `pkg/alertmanager/validation/validation.go`  that would check label names.

test/e2e/alertmanager_test.go
```diff
@@ -938,9 +939,10 @@ func testAlertmanagerConfigVersions(t *testing.T) {
 			Route: &monitoringv1alpha1.Route{
 				Receiver: "webhook",
 				Matchers: []monitoringv1alpha1.Matcher{{
-					Name:  "job",
-					Value: "webapp.+",
-					Regex: true,
+					Name:      "job",
+					MatchType: monitoringv1alpha1.MatchType(labels.MatchRegexp.String()),
```
simonpasquier, line 943: I wouldn't change the test: it verifies that the "legacy" API (e.g. when matchType is empty) works as expected.

simonpasquier:
I forgot to mention that we can fix in steps: it's fine to have this PR only looking after the route matchers.

pkg/admission/admission.go
rexagod: I wasn't able to see the exact reason why the E2E tests were failing earlier, including the error here made it easier to pinpoint the issue. I'll remove these.

rexagod:
There's a considerable amount of code that had to be duplicated to make `ParseMatchers` functionality available throughout `v1alpha1` and `v1beta1` while not depending upstream for the same.

I imagine https://github.com/prometheus-operator/prometheus-operator/issues/6023 will work towards extracting APIs that depend on overlapping behaviors in each of these versions into `v1`, albeit at the expense of parameter type safety, but we should be able to circumvent that to degree by specifying the expected arg types in the function comment, or just plain out using generics.

pkg/apis/monitoring/go.mod
rexagod: It seems a considerable amount of code that had to be duplicated to make `ParseMatcher` functionality available throughout `v1alpha1` and `v1beta1` while not depending upstream for the same.

I imagine https://github.com/prometheus-operator/prometheus-operator/issues/6023 will work towards extracting APIs that depend on overlapping behaviors in each of these versions into `v1` (a dedicated function), albeit at the expense of parameter type safety, but we should be able to circumvent that to degree by specifying the expected arg types in the function comment, or just plain out using generics.

Should I defer the merge effort to the same aforementioned PR or include that change in this one?

simonpasquier: I'll prefer a separate PR for #6023. And as you said, it's probably better to do it before this one.

rexagod:
Blocked by #6023.

rexagod:
👋🏼 Sorry for the delay here, but could this be reopened? I've pushed some changes to this branch to resume the effort.

Issue #6033, opened 2023-10-20
Add support for AWS sigv4 auth to Alertmanager endpoints


simonpasquier:
**What is missing?**

https://github.com/prometheus/prometheus/pull/12774 will be available in Prometheus v2.48.0. The Prometheus CRD should be extented to support this new auth method.

**Why do we need it?**

feature parity.

**Environment**

* Prometheus Operator version:

    `Insert image tag or Git SHA here`
    
    

**Anything else we need to know?**:

heylongdacoder:
I will give it a try 😄

PR #6036, opened 2023-10-22
prometheus: Add AWS sigv4 auth to Alertmanager endpoints

heylongdacoder:
## Description

Fixes: #6033

## Type of change

## Changelog entry

```release-note
Add AWS sigv4 auth to Prometheus's Alertmanager endpoints
```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -122,6 +122,31 @@ func ValidateRemoteWriteSpec(spec monitoringv1.RemoteWriteSpec) error {
 	return nil
 }
 
+func ValidateAlertmanagerEndpoints(am monitoringv1.AlertmanagerEndpoints) error {
+	var nonNilFields []string
+
+	if am.BearerTokenFile != "" {
+		nonNilFields = append(nonNilFields, fmt.Sprintf("%q", "bearerTokenFile"))
+	}
+
+	for k, v := range map[string]interface{}{
+		"basicAuth":     am.BasicAuth,
+		"authorization": am.Authorization,
+		"sigv4":         am.Sigv4,
+	} {
+		if reflect.ValueOf(v).IsNil() {
+			continue
+		}
+		nonNilFields = append(nonNilFields, fmt.Sprintf("%q", k))
+	}
+
+	if len(nonNilFields) > 1 {
+		return fmt.Errorf("%s can't be set at the same time, at most one of them must be defined", strings.Join(nonNilFields, " and "))
+	}
+
+	return nil
+}
+
 // Process will determine the Status of a Prometheus resource (server or agent) depending on its current state in the cluster
 func (sr *StatusReporter) Process(ctx context.Context, p monitoringv1.PrometheusInterface, key string) (*monitoringv1.PrometheusStatus, error) {
 
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -339,6 +339,35 @@ func (cg *ConfigGenerator) addBasicAuthToYaml(cfg yaml.MapSlice,
 	return cg.WithKeyVals("component", strings.Split(assetStoreKey, "/")[0]).AppendMapItem(cfg, "basic_auth", authCfg)
 }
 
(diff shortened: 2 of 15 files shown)


pkg/prometheus/operator.go
```diff
@@ -122,6 +122,27 @@ func ValidateRemoteWriteSpec(spec monitoringv1.RemoteWriteSpec) error {
 	return nil
 }
 
+func ValidateAlertmanagerEndpoints(am monitoringv1.AlertmanagerEndpoints) error {
```
nicolastakashi, line 125: I think this function may fits better under the AM [validation.go](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/alertmanager/validation/validation.go)

pkg/prometheus/server/operator.go
```diff
@@ -1518,13 +1518,8 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *mon
 	}
 
 	if p.Spec.Alerting != nil {
-		for i, am := range p.Spec.Alerting.Alertmanagers {
-			if err := store.AddBasicAuth(ctx, p.GetNamespace(), am.BasicAuth, fmt.Sprintf("alertmanager/auth/%d", i)); err != nil {
```
nicolastakashi, line 1522: I don't get the reason why we move this logic from this place to the `pkg/prometheus/store.go`

pkg/prometheus/operator.go
heylongdacoder: Hi @nicolastakashi, thanks for the review! I've checked the file, but it appears that the file is more related to the Alertmanager's configuration. What I'm adding here is the Prometheus' AlertmanagerEndpoints configuration, and I noticed there is another 'ValidateRemoteWriteSpec' function in this 'prometheus/operator.go' file as well. In my opinion, it makes more sense to place this function in the 'prometheus' folder? 😄

pkg/prometheus/server/operator.go
heylongdacoder: I noticed some similar functions in the `pkg/prometheus/store.go`, such as `AddRemoteWritesToStore`, so I moved it to this file to standardize it. 😄

heylongdacoder:
e2e test failed due to the operator crash. Made this changes to fix it (https://github.com/prometheus-operator/prometheus-operator/compare/f6711256e47695ef95a77276e4f5a8983f82a221..8cc54344f9195680a176298b9ea33e3e3cbdb952) 🙏

pkg/prometheus/promcfg.go
```diff
@@ -1848,27 +1879,7 @@ func (cg *ConfigGenerator) generateRemoteWriteConfig(
...
-			if spec.Sigv4.Profile != "" {
-				sigV4 = append(sigV4, yaml.MapItem{Key: "profile", Value: spec.Sigv4.Profile})
-			}
-			if spec.Sigv4.RoleArn != "" {
-				sigV4 = append(sigV4, yaml.MapItem{Key: "role_arn", Value: spec.Sigv4.RoleArn})
-			}
-
-			cfg = cg.WithMinimumVersion("2.26.0").AppendMapItem(cfg, "sigv4", sigV4)
-		}
+		cfg = cg.WithMinimumVersion("2.26.0").addSigv4ToYaml(cfg, fmt.Sprintf("remoteWrite/%d", i), store, spec.Sigv4)
```
ArthurSens, line 1882: Copy-paste mistake with remoteWrite? 😅

pkg/prometheus/server/operator.go
ArthurSens: Looks good to me :)

Indeed a lot of asset store code is already present in the store package

ArthurSens (commented):
Small nit

pkg/prometheus/promcfg.go
heylongdacoder: Hi @ArthurSens, thanks for the review! Nope, it is not a mistake. Since we can now configure `sigv4` in both the `alertmanagerEndpoints` and `remoteWrite` configurations, I have added a new `addSigv4ToYaml` function and updated this section to use this new function as well. 😄

Issue #3480, opened 2020-09-09
"targets" field in staticConfig in Probe custom resource is wrongly called "static"


lilic:
Seems like it might have been a typo, but the API field is called `static` instead of `targets`. In the [description](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#probetargetstaticconfig), we seem to call it targets:
> Targets is a list of URLs to probe using the configured prober.

Should we break API to do a fix for this as it seems like a bug to me, despite it being a breaking change?

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#static_config

lilic:
cc @prometheus-operator/prometheus-operator-reviewers

s-urbaniak:
As an after thought I believe we should have declared the Probe CRDs as suggested in https://github.com/prometheus-operator/prometheus-operator/issues/3439 using v1alpha1 most likely.

I think, in this very case, I would be fine with breaking behavior, but we should take care that new CRDs follow a more gentle versioning approach, especially https://github.com/prometheus-operator/prometheus-operator/pull/3451.

brancz:
IIRC there was some functionality in CRDs to rename fields, if so we can use that to rename. It is indeed unfortunate, but we did find small inconsistencies like this even after 1.5 years of v1alpha1 in the ServiceMonitor and Prometheus resources, so I wouldn't beat ourselves up about it too much :)

lilic:
@brancz I will do some research around that, thanks for the pointer!

lilic:
Asked and Jordan answered:

> you can do that by defining multiple versions of your custom resource, and creating a custom conversion webhook to convert instances back and forth between versions. See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions and https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion

Don't think this works for us here. At least I don't see it working.

Agreed with @s-urbaniak we should not define new APIs as v1 from now onwards.

brancz:
I would do something a bit simpler without the webhook. Let's just introduce a new field that is actually the name that we want, deprecate the `static` field, and use both for some time. Eventually, we can stop using the value from the `static` field.

lilic:
Yes agreed, my thinking as well!

lilic:
[...]

PR #6047, opened 2023-10-26
enhancement: `s/staticConfig_targets/staticConfig_hosts`

rexagod:
## Description

Deprecate Probe `targets` in favor of Probe `hosts` to address the inconsistency between the currently exported `Targets` field and its corresponding JSON representation, "static".

Fixes: #3480 

## Type of change

## Changelog entry

```release-note
- Introduce a `hosts` field in `staticConfig`, that will replace `targets` within the same schema section, in an effort to make the exported field match the JSON representation (earlier: `Targets,static", now: "Hosts,hosts").

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -937,6 +937,10 @@ func (cg *ConfigGenerator) generateProbeConfig(
 	switch {
 	case m.Spec.Targets.StaticConfig != nil:
 		// Generate static_config section.
+		// Hosts takes precedence over Targets, since the latter is deprecated.
+		if len(m.Spec.Targets.StaticConfig.Hosts) > 0 {
+			m.Spec.Targets.StaticConfig.Targets = m.Spec.Targets.StaticConfig.Hosts
+		}
 		staticConfig := yaml.MapSlice{
 			{Key: "targets", Value: m.Spec.Targets.StaticConfig.Targets},
 		}
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -9032,6 +9032,18 @@ More info: <a href="https://prometheus.io/docs/prometheus/latest/configuration/c
 </em>
 </td>
 <td>
+<p>The list of hosts to probe.
+Deprecated: Use <code>hosts</code> instead.</p>
+</td>
+</tr>
+<tr>
+<td>
+<code>hosts</code><br/>
+<em>
+[]string
+</em>
+</td>
+<td>
 <p>The list of hosts to probe.</p>
 </td>
 </tr>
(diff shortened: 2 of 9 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -937,6 +937,10 @@ func (cg *ConfigGenerator) generateProbeConfig(
 	switch {
 	case m.Spec.Targets.StaticConfig != nil:
 		// Generate static_config section.
+		// Hosts takes precedence over Targets, since the latter is deprecated.
+		if m.Spec.Targets.StaticConfig.Hosts != nil {
+			m.Spec.Targets.StaticConfig.Targets = m.Spec.Targets.StaticConfig.Hosts
+		}
```
ArthurSens, line 943: Not sure how we handled similar deprecations in the past (if we did at all), but any reason why not doing something like below?
```suggestion
			m.Spec.Targets.StaticConfig.Targets = append(m.Spec.Targets.StaticConfig.Targets, ...m.Spec.Targets.StaticConfig.Hosts)
```

ArthurSens (commented):
Thanks for tackling such an old issue :)

One question just out of curiosity, not really a change request

pkg/prometheus/promcfg.go
rexagod: Since we are encouraging folks to use the newer `hosts` over the older `static` in this patch, I thought it'd make sense for the former to take precedence when both are specified. I pointed this out in the change log above as well to make it explicit for the users, but LMK if that doesn't make sense, and I'll make the change. :)

simonpasquier: We do something similar with `port` and `targetPort` for the endpoints:

https://github.com/prometheus-operator/prometheus-operator/blob/63e15b7eb4897653d2462f8f41826743393a1f89/pkg/prometheus/promcfg.go#L794-L803

pkg/prometheus/promcfg.go
```diff
@@ -937,6 +937,10 @@ func (cg *ConfigGenerator) generateProbeConfig(
 	switch {
 	case m.Spec.Targets.StaticConfig != nil:
 		// Generate static_config section.
+		// Hosts takes precedence over Targets, since the latter is deprecated.
+		if m.Spec.Targets.StaticConfig.Hosts != nil {
```
simonpasquier, line 941:
```suggestion
		if len(m.Spec.Targets.StaticConfig.Hosts) > 0 {
```

simonpasquier (commented):
Can you add a test to pkg/prometheus/promcfg_test.go? It would also be good add a warning log reporting the use of deprecated field.

rexagod:
👋🏼 Sorry for the delay here, but could this be reopened? I've pushed some changes to this branch to resume the effort.

simonpasquier:
@rexagod apparently you can't reopen a PR after a force-push without some manual fix-up.

https://stackoverflow.com/a/75471397

PR #6050, opened 2023-10-30
enhancement: include operator version in CRD annotations

rexagod:
## Description

Include operator version in CRD annotations, PTAL https://github.com/prometheus-operator/prometheus-operator/pull/4498 for details.

Fixes: #4344.

***

This patch builds on top of @paulfantom's PR by (a) rebasing it against `main` and (b) adding support for the corresponding GNU `sed`-based behavior in MacOS using BSD `sed`.

***

## Type of change

## Changelog entry

```release-note
- Include operator version in CRD annotations ("prometheus-operator.dev/version").
```

### DIFF
--- a/scripts/generate/append-operator-version.sh
+++ b/scripts/generate/append-operator-version.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  find example/prometheus-operator-crd/ -name '*.yaml' -exec sed -i '' -e "/^    controller-gen.kubebuilder.io.version.*/a\\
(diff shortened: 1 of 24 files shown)


scripts/generate/append-operator-version.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
nicolastakashi, line 3: I'd suggest use the label as we have on the managed resources: `operator.prometheus.io`
WDYT?

rexagod: Hmm, makes sense! 👍🏼

ArthurSens (approved):
LGTM!

Issue #5798, opened 2023-08-09
Paramaterize group for CRs to use customize prometheus operator


rashmichandrashekar:
I am trying to run a customized version of opentelemetry-operator(https://github.com/open-telemetry/opentelemetry-operator) which in turn uses prometheus-operator. I want to use CRDs (pod and service monitor) in a different group (other than "monitoring.coreos.com"), so that it doesnt interfere in an environment which already has prometheus operator running. I see it is currently hardcoded per version here - https://github.com/prometheus-operator/prometheus-operator/blob/4e37c62781ce19971726f5a34b073327a3860d1a/pkg/apis/monitoring/register.go#L18

Is it possbile to pass this in as a parameter to get CRs for the CRDs defined in a different group?
Currently I see that the only way to do this is to create another version and register it with a different group.

Thanks in advance!

Environment - AKS

simonpasquier:
AFAIK this isn't feasible with the CRD tooling. The only way I'm aware of is to fork the repository and replace all occurrences of "monitoring.coreos.com" by your custom API group.

vishiy:
can it be parametrized ? It would be ideal to parametereize it and default to current (monitoring.coreos.com)

simonpasquier:
I don't know of any way to do it with the current tools that we use for CRD generation (e.g. `controller-gen`).

rashmichandrashekar:
@simonpasquier - I have created a PR to fix this issue. Please take a look and lmk if you have any feedback. Thanks!
https://github.com/prometheus-operator/prometheus-operator/pull/6055

PR #6055, opened 2023-10-30
Paramaterize group for CRs to use prometheus operator with opentelemetry-operator

rashmichandrashekar:
## Description
Paramaterize api group for v1 apis/monitoring pkg so that CRs can be detected in api group other than default 'monitoring.coreos.com' 
Fixes this issue  - https://github.com/prometheus-operator/prometheus-operator/issues/5798

## Type of change

- [] `ENHANCEMENT` (non-breaking change which improves existing functionality)

## Changelog entry

```release-note
Paramaterize api group for v1 apis/monitoring pkg so that CRs can be detected in api group other than default 'monitoring.coreos.com' 
```


simonpasquier:
There are many other places in the code base that rely on `monitoring.GroupName`. I'm quite sure that the change isn't enough.

rashmichandrashekar:
> There are many other places in the code base that rely on `monitoring.GroupName`. I'm quite sure that the change isn't enough.



> There are many other places in the code base that rely on `monitoring.GroupName`. I'm quite sure that the change isn't enough.

Thanks @simonpasquier. I had only made the changes relevant to the otel operator relying on this. I have updated all the references, pls take a look.

simonpasquier:
Personally, I'm not super keen on maintaining extra code that isn't relevant for the operator itself. I would find it easier to have a forked version of the code base which updates only the group name.

cc @prometheus-operator/prometheus-operator-reviewers for inputs.

rashmichandrashekar:
@simonpasquier - Supporting custom api group might still be relevant to the operator since customers can run 2 instances side by side to use CRDs and CRs created in different groups without conflicting with each other, so that, uninstalling one instance of the operator doesnt affect the other instance and so on. One such instance is if we were to support an offering for managed prometheus operator, we wouldn't want that to interfere with OSS CRDs/CRs for installation/deletion and CR detection.

Forking this makes maintainability and upgrades very hard. We cannot go mod the package version and will have to manually take these changes every time we update. We want to keep the dependency on prometheus-operator as clean as possible for maintenance and upgrades. The code changes in this PR are also back compatible and introduces no breaking change. Hope this helps! If there is anything else to make this maintainability easier, I am happy to contribute. Please let me know.

rashmichandrashekar:
> IMHO forking and replacing all occurrences of `monitoring.coreos.com` by `foobar.example.com` isn't really complicated (see https://github.com/rhobs/obo-prometheus-operator/). The main advantage from my POV is that it's much more explicit (e.g. 1 binary = 1 API group) and reduces the burden on the maintainers team in terms of support.
> 
> Also from what I can tell, the proposed change isn't enough to run the operator with a different API group name.

Thanks @simonpasquier. I agree forking and replacing is not complicated. But it makes builds and dependency upgrades harder. Since this change is backward compatible, it shouldnt cause any breaking changes. 

Also, I have tested this out by importing the prometheus operator package and made sure it picks up the resources in the custom group well. Is there a better way to test this? Perhaps, I can add tests to this repo? Also, is there a way to test this change out using e2e tests. They seem to pass just fine, since they are checking for existing behavior.

simonpasquier:
> Also, I have tested this out by importing the prometheus operator package and made sure it picks up the resources in the custom group well. 

We're probably not talking about the same thing. What do you mean by "importing the prometheus operator package and made sure it pickes up the resources in the custom group"?

rashmichandrashekar:
> > Also, I have tested this out by importing the prometheus operator package and made sure it picks up the resources in the custom group well.
> 
> We're probably not talking about the same thing. What do you mean by "importing the prometheus operator package and made sure it pickes up the resources in the custom group"?

The changes I have made to the v1 package will work when it is imported and the resources are manipulated in the updated group. May be you meant something else? When you say the changes might not enough, I am wondering if there is a way for me to test the scenario out which you might think it would break. Please lmk. Thanks!

simonpasquier:
Ok you're talking from the use case of a client project importing the `github.com/prometheus-operator/prometheus-operator/pkg/apis` Go module and that wants to manage the operator's objects but from a different API group?
My main worry is that someone assumes that running the operator with the `PROMETHEUS_OPERATOR_V1_CUSTOM_GROUP=monitoring.example.com` env variable would magically make the operator reconcile objects from the `monitoring.example.com` group.
It would help if you had some example code showing how you want to use it.

rashmichandrashekar:
> Ok you're talking from the use case of a client project importing the `github.com/prometheus-operator/prometheus-operator/pkg/apis` Go module and that wants to manage the operator's objects but from a different API group? My main worry is that someone assumes that running the operator with the `PROMETHEUS_OPERATOR_V1_CUSTOM_GROUP=monitoring.example.com` env variable would magically make the operator reconcile objects from the `monitoring.example.com` group. It would help if you had some example code showing how you want to use it.

Thanks @simonpasquier. Yes, the use case is - the prometheus-operator v1 package is being imported in [opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator/blob/4baf9c06639956668a4cd3adbeb65f16cc6cc501/cmd/otel-allocator/watcher/promOperator.go#L26). This looks for CRs in the group "monitoring.coreos.com". With the opentelemetry-operator I would instead like to look for CRs in a different group. I am working on an implementation where we have a managed offering of the opentelemetry-operator which looks for CRs in a custom group so that it doesnt interfere with the OSS installation of prometheus-operator. Meaning that if a customer were to run the managed offering and the OSS prometheus operator side side by side we dont want one picking up CRs created for the other and vice versa (same applies for deletion).
Since we are depending on numerous OSS components, it will be hard for us to keep a copy of the repo and take upgrades and make changes every time. A much cleaner option would be to import these components as go modules so that we can upgrade to newer versions easily and tweak the behavior using environment variables specific to our environment .
Would it help if I added some doc (and comment) around this, so that users don't get confused or mistake this for something that this is not intend to do? I can add some examples too.

simonpasquier:
I've been thinking up about it. One less-intrusive option would be to inject a different group name at build time via the `go tool link` options:

https://github.com/prometheus-operator/prometheus-operator/compare/main...simonpasquier:prometheus-operator:custom-tag-for-register

ArthurSens:
Hey 👋 -- Sorry for the lag here everybody!

I'm aligned with Simon's point of view. I'm also concerned that Prometheus-operator offering tooling to modify the API Group name will confuse users about what could be done with it and how far we could offer support to people trying to use different group names.

I'm voicing this here to in case opinions from other maintainers are needed :)

> I've been thinking up about it. One less-intrusive option would be to inject a different group name at build time via the `go tool link` options:
> 
> [main...simonpasquier:prometheus-operator:custom-tag-for-register](https://github.com/prometheus-operator/prometheus-operator/compare/main...simonpasquier:prometheus-operator:custom-tag-for-register)

If this approach works, changing from `const` to `var` sounds like a pretty simple change! No blockers for this one :)

rashmichandrashekar:
> Hey 👋 -- Sorry for the lag here everybody!
> 
> I'm aligned with Simon's point of view. I'm also concerned that Prometheus-operator offering tooling to modify the API Group name will confuse users about what could be done with it and how far we could offer support to people trying to use different group names.
> 
> I'm voicing this here just in case opinions from other maintainers are needed :)
> 
> > I've been thinking up about it. One less-intrusive option would be to inject a different group name at build time via the `go tool link` options:
> > [main...simonpasquier:prometheus-operator:custom-tag-for-register](https://github.com/prometheus-operator/prometheus-operator/compare/main...simonpasquier:prometheus-operator:custom-tag-for-register)
> 
> If this approach works, changing from `const` to `var` sounds like a pretty simple change! No blockers for this one :)

Apologies for the delay in response. I was investigating if this solution works for us and it indeed does. Thanks so much for all the help @simonpasquier and @ArthurSens.

rashmichandrashekar:
> Hey 👋 -- Sorry for the lag here everybody!
> 
> I'm aligned with Simon's point of view. I'm also concerned that Prometheus-operator offering tooling to modify the API Group name will confuse users about what could be done with it and how far we could offer support to people trying to use different group names.
> 
> I'm voicing this here just in case opinions from other maintainers are needed :)
> 
> > I've been thinking up about it. One less-intrusive option would be to inject a different group name at build time via the `go tool link` options:
> > [main...simonpasquier:prometheus-operator:custom-tag-for-register](https://github.com/prometheus-operator/prometheus-operator/compare/main...simonpasquier:prometheus-operator:custom-tag-for-register)
> 
> If this approach works, changing from `const` to `var` sounds like a pretty simple change! No blockers for this one :)

Apologies for the delay in response. I was investigating if this solution works for us and it indeed does. Thanks so much @simonpasquier  for all the help. Do you want me to just update this PR to change const to var?

simonpasquier:
@rashmichandrashekar please do it! This isn't controversial at all but please a comment in the code explaining why it's a var and not a const.

rashmichandrashekar:
> @rashmichandrashekar please do it! This isn't controversial at all but please a comment in the code explaining why it's a var and not a const.

Thanks @simonpasquier. I have updated it. Please take a look and lmk if you need me to add anything else in the comment.

ArthurSens (approved):
:shipit:

pkg/apis/monitoring/register.go
```diff
@@ -14,6 +14,9 @@
 
 package monitoring
 
-const (
+// GroupName is set to var instead of const, since this provides the ability for clients importing the module -
+// github.com/prometheus-operator/prometheus-operator/pkg/apis to manage the operator's objects in a different
+// API group
```
ArthurSens, line 19:
```suggestion
// GroupName is set to `var` instead of `const`, since this provides the ability for clients importing the module -
// github.com/prometheus-operator/prometheus-operator/pkg/apis to manage the operator's objects in a different
// API group
//
// Use `ldflags` in the client side, e.g.:
// go run -ldflags="-s -X github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring.GroupName=monitoring.example.com" ./example/client/.
```

ArthurSens (approved):
Not a blocker, but we could give an usage example. I don't think many people know about this option

pkg/apis/monitoring/register.go
rashmichandrashekar: Sure @ArthurSens. Are you suggesting that I add the file with the example too or just the comment?

ArthurSens: Just the comment is good enough!

rashmichandrashekar: Thanks @ArthurSens. I dont have the permissions to merge. Could you please merge this?

rashmichandrashekar: Following up. @ArthurSens @simonpasquier  - Could you pls help get this PR merged? Thanks!

ArthurSens:
Thanks again!

PR #6057, opened 2023-10-31
Ignore invalid CRDs and pick up valid CRDs without failing and add validation for relabeling and metricRelabeling fields for pod/svc monitors

rashmichandrashekar:
## Description

When a podmonitor/svcmonitor CRD has an erraneous field for regex in the relabeling/metric relabeling fields, the crd creation succeeds since it can be a valid string but not a valid regex( ex- .*( ). This causes any subsequent valid CRDs being created from getting picked up. Also, the current generatePodMonitorConfig/generateServiceMonitorConfig doesnt do any validation for relabeling/metricRelabeling. Can leverage the **validateRelabelConfig** method here - https://github.com/prometheus-operator/prometheus-operator/blob/dfc150aa362f0260060979d0b53a524efb974168/pkg/prometheus/resource_selector.go#L236

## Type of change

## Changelog entry
```release-note
Add validation for relabeling/metricsRelabeling for pod/service monitor and continue picking up valid CRDs and ignore invalid CRDs
```

### DIFF
--- a/pkg/apis/monitoring/v1/register.go
+++ b/pkg/apis/monitoring/v1/register.go
@@ -20,20 +20,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring"
-	"os"
 )
 
-var PackageGroupName = func() string {
-	group := monitoring.GroupName
-	customGroupV1 := os.Getenv("PROMETHEUS_OPERATOR_V1_CUSTOM_GROUP")
-	if customGroupV1 != "" {
-		group = customGroupV1
-	}
-	return group
-}()
-
 // SchemeGroupVersion is the group version used to register these objects
-var SchemeGroupVersion = schema.GroupVersion{Group: PackageGroupName, Version: Version}
+var SchemeGroupVersion = schema.GroupVersion{Group: monitoring.GroupName, Version: Version}
 
 // Resource takes an unqualified resource and returns a Group qualified GroupResource
 func Resource(resource string) schema.GroupResource {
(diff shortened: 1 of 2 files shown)


slashpai:
Relabel configs are validated while selecting servicemonitors/podmonitors eg: https://github.com/prometheus-operator/prometheus-operator/blob/dfc150aa362f0260060979d0b53a524efb974168/pkg/prometheus/resource_selector.go#L155-L163. 

This is not checked while generating since we will be rejecting once that are failing the validation instead of failing the reconciliation due to one/few wrong servicemonitors.

simonpasquier:
What @slashpai says: at this stage, invalid configs have already been rejected.

rashmichandrashekar:
@slashpai @simonpasquier - Thanks for taking a look. Is it okay to expose this method [validateRelabelConfig](https://github.com/prometheus-operator/prometheus-operator/blob/5f1a151a48d3ec71230c09e0ab4575b0a26560a9/pkg/prometheus/resource_selector.go#L237) so that others consuming this package, who dont take the same route as through the resource selector can reuse the validation and can have it consistent with the prom-operator?
I can make a PR for that.

simonpasquier:
Can you create a new issue in this repository to discuss the general approach with all @prometheus-operator/prometheus-operator-reviewers?
While you have one specific request now, I bet that the topic will come up again in the future and it's be preferred that we get to a general agreement across all maintainers regarding code reuse by other projects.

rashmichandrashekar:
> Can you create a new issue in this repository to discuss the general approach with all @prometheus-operator/prometheus-operator-reviewers? While you have one specific request now, I bet that the topic will come up again in the future and it's be preferred that we get to a general agreement across all maintainers regarding code reuse by other projects.

@simonpasquier - Created this issue - https://github.com/prometheus-operator/prometheus-operator/issues/6076. Hope this helps.

PR #6064, opened 2023-11-06
fix: remove verbose logging admission-webhook

slashpai:
After disabling HTTP2 connections by default in #6028 started seeing verbose logging in admission webhook pods like:

`ts=2023-11-06T01:50:34.601774794Z caller=stdlib.go:105 caller=server.go:3215 msg="http: superfluous response.WriteHeader call from main.newSrv.func1 (main.go:173)" `

Go doc says If WriteHeader is not called explicitly, the first call to Write will trigger an implicit
WriteHeader(http.StatusOK).

This fix is to change the order of the method invocation so log doesn't complain about this

## Description

## Type of change

## Changelog entry

```release-note
fix: remove verbose logging admission-webhook
```


simonpasquier (approved):
Thanks!

PR #6067, opened 2023-11-07
[WIP] feat: Support service names for individual Prometheus Instances

### LINKED ISSUE #6060: Prometheus Operators incorrectly removes gRPC port from shared prometheus-operated Service resource
**What happened?**
We deploy hundreds of Prometheus instances leveraging the Prometheus Operator, the majority of which are deployed along with a thanos sidecar. While investigating a separate issue, we made a change to remove `/spec/thanos` from one of the Prometheus instances in the cluster and did not modify the rest (20+) Prometheus instances which did specify `/spec/thanos` in the definition. This change ended up modifying the shared `prometheus-operated` Service resource to remove the gRPC port altogether, which resulted in all of our Prometheus instances being disconnected from our query pipeline. 

Given that there were several Prometheus instances that still had `/spec/thanos` defined, the gRPC port should have continued to exist on the shared Service resource. 

- Setup multiple Prometheus instances with `/spec/thanos` defined and ensure they are synced and running. 
- Remove `/spec/thanos` from one of the Prometheus instances. 

**Environment**

    `v0.63.0`
    `v1.25.7`

This bug likely happens because the reconciliation of the the Shared Service resource is happening in isolation: 
https://github.com/prometheus-operator/prometheus-operator/blob/ddfea8919b8d489b016bdbd59481a77602cd76eb/pkg/prometheus/server/statefulset.go#L82 

This is fine for Stateful Sets since those resources are unique per instances - however because the Service resource is shared across all Prometheus instances, you need to inspect all of them to determine if there is any Thanos dependency.

nidhey27:
## Description
This PR introduces the capability for each Prometheus instance managed by the Prometheus Operator to specify a `.spec.serviceName` in its configuration. This change allows individual instances to have their own dedicated Service resources. The motivation behind this feature is to provide modularity and stability by ensuring that changes to the Service configuration of one Prometheus instance do not affect other instances. This enhancement is particularly useful in environments where multiple Prometheus instances with different Thanos sidecar configurations are deployed.

Closes #6060 

## Type of change

## Changelog entry

```release-note
Added support for individual Service resources per Prometheus instance via .spec.serviceName, allowing for greater isolation and independent management of Prometheus instances within a cluster.
```

### DIFF
--- a/pkg/apis/monitoring/v1/prometheus_types.go
+++ b/pkg/apis/monitoring/v1/prometheus_types.go
@@ -607,6 +607,9 @@ type CommonPrometheusFields struct {
 	//
 	// +optional
 	KeepDroppedTargets *uint64 `json:"keepDroppedTargets,omitempty"`
+	// ServiceName is the name of the Service resource created for the PrometheusAgent instance.
+	// +optional
+	ServiceName string `json:"serviceName,omitempty"`
 }
 
 // +genclient
@@ -861,6 +864,9 @@ type PrometheusStatus struct {
 	// +listMapKey=shardID
 	// +optional
 	ShardStatuses []ShardStatus `json:"shardStatuses,omitempty"`
+	// ServiceName is the name of the Service resource associated with the Prometheus
+	// instance.
+	ServiceName string `json:"serviceName,omitempty"`
 }
 
 // AlertingSpec defines parameters for alerting configuration of Prometheus servers.
--- a/pkg/prometheus/agent/statefulset.go
+++ b/pkg/prometheus/agent/statefulset.go
@@ -429,9 +429,14 @@ func makeStatefulSetService(p *monitoringv1alpha1.PrometheusAgent, config operat
 		p.Spec.PortName = prompkg.DefaultPortName
 	}
 
+	serviceName := p.Spec.ServiceName
+	if serviceName == "" {
+		serviceName = governingServiceName
+	}
+
 	svc := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: governingServiceName,
+			Name: serviceName,
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					Name:       p.GetName(),
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -257,5 +257,12 @@ func (sr *StatusReporter) Process(ctx context.Context, p monitoringv1.Prometheus
 		sr.Reconciliations.GetCondition(key, p.GetObjectMeta().GetGeneration()),
 	)
 
+	if commonFields.ServiceName != "" {
+		pStatus.ServiceName = commonFields.ServiceName
+	} else {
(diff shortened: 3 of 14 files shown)


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -607,6 +607,9 @@ type CommonPrometheusFields struct {
 	//
 	// +optional
 	KeepDroppedTargets *uint64 `json:"keepDroppedTargets,omitempty"`
+	// ServiceName is the name of the Service resource created for the PrometheusAgent instance.
+	// +optional
+	ServiceName string `json:"serviceName,omitempty"`
```
nicolastakashi, line 612: (nit)
```suggestion
	ServiceName *string `json:"serviceName,omitempty"`
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -861,6 +864,9 @@ type PrometheusStatus struct {
 	// +listMapKey=shardID
 	// +optional
 	ShardStatuses []ShardStatus `json:"shardStatuses,omitempty"`
+	// ServiceName is the name of the Service resource associated with the Prometheus
+	// instance.
+	ServiceName string `json:"serviceName,omitempty"`
```
nicolastakashi, line 869:
```suggestion
	ServiceName *string `json:"serviceName,omitempty"`
```

pkg/prometheus/agent/statefulset.go
```diff
@@ -429,9 +429,14 @@ func makeStatefulSetService(p *monitoringv1alpha1.PrometheusAgent, config operat
 		p.Spec.PortName = prompkg.DefaultPortName
 	}
 
+	serviceName := p.Spec.ServiceName
+	if serviceName == "" {
```
nicolastakashi, line 433:
```suggestion
	if serviceName == nil {
```

pkg/prometheus/operator.go
```diff
@@ -257,5 +257,12 @@ func (sr *StatusReporter) Process(ctx context.Context, p monitoringv1.Prometheus
 		sr.Reconciliations.GetCondition(key, p.GetObjectMeta().GetGeneration()),
 	)
 
+	if commonFields.ServiceName != "" {
+		pStatus.ServiceName = commonFields.ServiceName
+	} else {
+		const governingServiceName = "prometheus-agent-operated"
```
nicolastakashi, line 263:
```suggestion
		pStatus.ServiceName = "prometheus-agent-operated"
```

simonpasquier (commented):
After more thinking, I'm not sure that exposing a serviceName field is the right approach because it will still fall short if multiple Prometheus objects configure the same service name or if we have 1 Prometheus object and 1 PrometheusAgent object with the same name.

Instead it might be more useful to add a service field to the CRDs:

```
kind: Prometheus
metadata:
  name: example
spec:
  service: {}
```

If the field isn't nil then the operator creates/manages the `prometheus-example` service. Initially we can only support the ClusterIP service type.

PR #6068, opened 2023-11-07
feat:  support the operator make use of  the config-reloader tls and basic authentication with crds

dongjiang1989:
## Description

ref: https://github.com/prometheus-operator/prometheus-operator/pull/5932#issuecomment-1761545280

Support the operator make use of  the config-reloader tls and basic authentication with crds 

## Type of change

## Changelog entry

```release-note
Support the operator make use of  the config-reloader tls and basic authentication with crds
```


dongjiang1989:
https://github.com/prometheus-operator/prometheus-operator/pull/5932#issuecomment-1761545280

@simonpasquier  PTAL

pkg/prometheus/agent/statefulset.go
```diff
@@ -236,6 +255,24 @@ func makeStatefulSetSpec(
 		webConfigGenerator.Warn("web.config.file")
 	}
 
+	// Mount web config and web TLS credentials as volumes for config-reloader.
+	// We always mount the web config file for versions greater than 0.69.0.
+	if cfgReloaderVersion.GTE(semver.MustParse("0.69.0")) && cpf.WebConfigReloader != nil {
```
nicolastakashi, line 260: Just wondering if we can abstract this code in a func and reuse between agent and server.

dongjiang1989: Thank you for your review. But abstracting this code into a func will have too many parameters. Probably not worth it 🤔️

simonpasquier (commented):
I'm wondering what's the correct API for this feature. Adding a `.spec.webConfigReloader` field increases the API surface and I'm not 100% sure that it's required. Maybe it's good enough to say "reuse the Prometheus web config for the reloader"?

dongjiang1989:
> I'm wondering what's the correct API for this feature. Adding a `.spec.webConfigReloader` field increases the API surface and I'm not 100% sure that it's required. Maybe it's good enough to say "reuse the Prometheus web config for the reloader"?

Thanks for your review. 
Adding`.spec.webConfigReloader` allows for more flexible configuration of config-reloader. 
But it can also be reused the Prometheus/AlertManager web config and add a `.spec.enableWebConfigReloader` flag field to control whether to enable this function.

Which method do you think is most suitable?

pkg/prometheus/agent/statefulset.go
nicolastakashi: I'm kinda Ok having some parameters on the func signature.
If we think it's to much we can always create an struct and send it to the func.

WDYT?

dongjiang1989:
Closing. 
Add the operator crds to make use of config-reloader web-config-file : https://github.com/prometheus-operator/prometheus-operator/pull/6194

PR #6030, opened 2023-10-19
feat: emit events for invalid configurations

rexagod:
## Description

Emit events when the controller rejects a resource, owing to an invalid configuration.

Fixes: #3611

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [x] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Emit events when the controller rejects a resource, owing to an invalid configuration. These are:
* AlertmanagerConfig
* PodMonitor
* Probe
* PrometheusRule
* ScrapeConfig
* ServiceMonitor
```

Issue #3922, opened 2021-03-17
Allow file references in PodMonitorEndpointTLSConfig


akhalyavka:
**What is missing?**
PodMonitorEndpointTLSConfig doesn't allow to reference certificates and key files.

**Why do we need it?**
This is causing us some trouble with Istio. As explained in [istio docs](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings), sidecar proxy can't deal with scraping requests because they are destined for pod's IP, so in order to scrape metrics over TLS sidecar should be configured to save certs and key to a volume and job's TLS config should reference them. While it is possible to achieve with ServiceMonitor, it is not with PodMonitor. Is there any good reason for this disparity?

* Prometheus Operator version:

    `v0.43.2`

simonpasquier:
It's on purpose that PodMonitors (as well as Probes) can't reference files that reside on the Prometheus pod's filesystem. It avoids exposing sensitive materials to the scraped targets inadvertently. ServiceMonitors support this because they predate PodMonitors and we can't change the API.

akhalyavka:
@simonpasquier Is it reasonable to assume that this functionality in ServiceMonitor will be deprecated in the future? Could it be added to PodMonitors with the flag to opt-out of this behavior (like `ArbitraryFSAccessThroughSMsConfig`)?

ThiruNandagopal:
It would be great if we have this feature in podmonitor also similar to servicemonitor as we don't need to follow 2 different approach for same feature.

mblaschke:
Would be nice to have this feature also for PodMonitor to support fetching metrics with Istio

EronWright:
I have a suggestion for a fix, feedback welcome.
https://github.com/prometheus-operator/prometheus-operator/issues/4121#issuecomment-1542976008

abohne:
Any updates on this?

taisph:
Is this issue also why a tlsConfig with files in a scrape class is not applied to PodMonitors? The same scrape class works with ServiceMonitors.

simonpasquier:
ScrapeClass is the answer to the original issue.

PR #6113, opened 2023-11-22
chore: adding scrape classe

nicolastakashi:
## Description

Taking over #5572 

Proposal to add scrape classes to facilitate the use of non-safe TLS elements in a scrape configuration, e.g. in the Istio use case.

A draft implementation is available: https://github.com/prometheus-operator/prometheus-operator/pull/5618. Links to significant code changes: [prom api](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-b4a3c944005e67c959c04552108ff3b629b572d7244d586b0558651bd96eacda), [scrape class](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-255c0aca616a1cdf4385c88d1d063a6c86360fedf99f7fc76bb967034e68897cR642-R652), [podmonitor api](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-ed302c709df2a336b5712f4f4442e4972be00571fd5293c61e0f65f752b9e10a), [operator](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-2297eaf6931eb847fb894df894f8f50ac9fb7cc823e440280d5908b5c9939104), [config generator](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-453ce9c721c51bf9591c2b212e9ce50157b9c40d1b211177bf502be268099656), and [tests](https://github.com/prometheus-operator/prometheus-operator/pull/5618/files#diff-40b00673aa9acb2e0dd5431cb36eedc1476124b086a3e070f5d6182709f8ea35).

## Type of change

## Changelog entry

```release-note
Adding Scrape Classe Proposal
```

### DIFF
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -58,7 +58,7 @@ It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a n
 The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
 of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource.
 
-Define a scrape classes for use by the monitors, one class may be designated as the default class.
+One (and only one) scrape class may be designated as the default class.
 
 When a resource defines several default scrape classes, it should fail the reconciliation.
 
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -84,7 +84,7 @@ Any object references in the scrape class definition are assumed to refer to obj
 
 ### PodMonitor Resource
 
-Allow the user to select a scrape class for each endpoint.
+Allow the user to select a scrape class which applies to all endpoints.
 
 ```yaml
 apiVersion: monitoring.coreos.com/v1
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -44,7 +44,7 @@ A scrape class defines TLS settings (and possibly other settings in future, e.g.
 One scrape class may be designated as the default class, in which case that class is applied to any scrape config that doesn't specify a value for `scrapeClass`.
 
 When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClass` field.
-For monitors, each endpoint may be assigned a scrape class.
+When there's a match, a scrape class is assigned to all the endpoints.
 
 Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
 
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -74,6 +74,10 @@ spec:
   volumeMounts:
     - name: istio-certs
       mountPath: "/etc/istio-certs/"
+  volumes:
+    - name: istio-certs
+      secret:
+        secretName: istio-certs
 ```
 
 Any object references in the scrape class definition are assumed to refer to objects in the namespace of the `Prometheus` object.
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -1,0 +1,217 @@
+## Scrape Classes
+
+* **Owners:**
+  * `nicolastakashi`
+  * `eronwright`
+
+* **Related Tickets:**
+  * https://github.com/prometheus-operator/prometheus-operator/issues/4121
+  * https://github.com/prometheus-operator/prometheus-operator/issues/3922
+  * https://github.com/prometheus-operator/prometheus-operator/issues/5947
+  * https://github.com/prometheus-operator/prometheus-operator/issues/5948
+
+This proposal introduces the concept of *scrape classes*, enabling users to utilize scrape configuration data provided by the administrator, through scrape objects such as PodMonitor, Probe, ServiceMonitor and ScrapeConfig.
+
+## Why
+
+Sometimes Prometheus administrators needs to provide default configurations for scrape objects, such as the definition of TLS certificates, when running Prometheus to scrape pods in an Istio mesh with strict mTLS as described in [Istio documentation](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings).
+
+Another motivation is to improve feature parity amongst the monitor resources. The `PodMonitor` and `Probe` resources aren't at parity with `ServiceMonitor` because the latter allows for unsafe TLS settings.
+
+### Pitfalls of the current solution
+
+The only known solution for use cases where you'd need to use unsafe TLS settings is to use `additionalScrapeConfig`.
+The downside is obviously the loss of a great feature of the Prometheus Operator. The monitor resources make it possible
+to compose the scrape configurations in a Kubernetes way.
+
+## Goals
+
+- Allow for the administrator to define a named, reusable scrape configuration snippet, including unsafe elements such as file references.
+- Allow for a user to select a configuration snippet by name in their probe/podmonitor/servicemonitor endpoint configuration.
+- Avoid giving the user the ability to use arbirary files within the Prometheus pod.
+
+### Audience
+
+- Users who serve Prometheus as a service and want to give their customers autonomy in defining monitors, but want to provide a default configuration for scraping.
+
+## How
+
+The proposed solution is to introduce a notion of a *scrape class*, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/).
+A scrape class defines TLS settings (and possibly other settings in future, e.g. sensitive authorization settings) to be applied to all scrape configs of that class.
+
+One scrape class may be designated as the default class, in which case that class is applied to any scrape config that doesn't specify a value for `scrapeClass`.
+
+When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClass` field.
+For monitors, each endpoint may be assigned a scrape class.
+
+Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
+
+### Prometheus Resource
+
+It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a new section for defining scrape classes.
+
+The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
+of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+spec:
+  # define scrape classes for use by the monitors
+  # one class may be designated as the default class
+  scrapeClasses:
+    - name: istio-mtls
+      default: true
+      tlsConfig:
+        caFile: "/etc/istio-certs/root-cert.pem"
+        certFile: "/etc/istio-certs/cert-chain.pem"
+        keyFile: "/etc/istio-certs/key.pem"
+        insecureSkipVerify: true
+
+  # mount the certs from the istio sidecar (shown here for illustration purposes)
+  volumeMounts:
+    - name: istio-certs
+      mountPath: "/etc/istio-certs/"
+```
+
+Any object references in the scrape class definition are assumed to refer to objects in the namespace of the `Prometheus` object.
+
+### PodMonitor Resource
+
+Allow the user to select a scrape class for each endpoint.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+spec:
+  scrapeClass: istio-mtls
+  podMetricsEndpoints:
+  - port: http
+    path: /metrics
+```
+
+The proposed behavior is:
+1. the `tlsConfig` in the associated scrape class is automatically applied to the scrape configuration of the endpoint.
+2. the inline `tlsConfig` (if any) takes precedence over the `tlsConfig` in the scrape class.
+
+### Probe Resource
+
+Allow the user to select a scrape class for the probe.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Probe
+spec:
+  scrapeClass: istio-mtls
+```
+
+### ServiceMonitor Resource
+
+Allow the user to select a scrape class for each endpoint.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+spec:
+  scrapeClass: istio-mtls
+  endpoints:
+  - port: http
+    path: /metrics
+```
+
+Out-of-scope: deprecation of the unsafe TLS settings in `ServiceMonitor`.
+
+### ScrapeConfig
+
+Allow the user to select a scrape class for the generic scrape configuration.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: ScrapeConfig
+metadata:
+  name: scrape-config
+spec:
+  scrapeClass: istio-mtls
+  staticConfigs:
+    [...]
+  httpSDConfig:
+    [...]
+  fileSDConfig:
+    [...]
+```
+
+## Test Plan
+
+1. Regression test; ensure no change to generated configs unless a scrape class is applied.
+2. Optionality of the default scrape class; ensure that it is an optional configuration element.
+3. Acceptance testing in Istio environment; ensure that the solution is effective for a key use case.
+
+## Alternatives
+
+### Global scrape TLS configuration
+
+An alternative solution would be to apply a default TLS configuration to all monitors.
+
+For example, via a hypothetical field `spec.scrapeTlsConfig`:
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+spec:
+  scrapeTlsConfig:
+    caFile: "/etc/istio-certs/root-cert.pem"
+    certFile: "/etc/istio-certs/cert-chain.pem"
+    keyFile: "/etc/istio-certs/key.pem"
+    insecureSkipVerify: true
+```
+
+Objections:
+1. A singular default configuration may be too inflexible to effectively scrape a diverse set of pods. For example, in Istio,
+   some pods may be in STRICT mode.
+
+### Istio Permissive Mode
+
+An alternative for the Istio use case is to use PERMISSIVE mode (see [documentation](https://istio.io/latest/docs/concepts/security/#permissive-mode)),
+or to use `exclude` annotations on the pod such that the metrics endpoint bypasses mTLS.
+
+Objections:
+1. This solution is less secure at a transport level unless TLS is implemented at the application layer.
+2. Strict mTLS is the basis for [Istio authorization policy](https://istio.io/latest/docs/reference/config/security/authorization-policy/), e.g. to explicitly allow access from prom to the metrics endpoint.
+
+### ScrapeClass Resource
+
+A variant of the proposed solution is to introduce a new custom resource for defining scrape classes.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: ScrapeClass
+metadata:
+  name: istio-mtls
+spec:
+  tlsConfig:
+    caFile: "/etc/istio-certs/root-cert.pem"
+    certFile: "/etc/istio-certs/cert-chain.pem"
+    keyFile: "/etc/istio-certs/key.pem"
+    insecureSkipVerify: true
+```
+
+An open question is whether the resource would be cluster-scoped or namespace-scoped.
+
+Objections:
+1. Since the file paths are dependent on the volume mounts in the server, this approach may not achieve a meaningful decoupling.
+2. Extra complexity in defining a new CRD.
+
+### Non-Safe Monitors
+
+Another alternative would be to allow `PodMonitor` and `Probe` to use unsafe TLS settings.
+
+Objections:
+1. See [explanation](https://github.com/prometheus-operator/prometheus-operator/issues/3922#issuecomment-802899950) for why
+   unsafe settings were disallowed in the first place.
+
+## Action Plan
+
+* [ ] Change the Operator API to define scrape classes and for endpoint to select a scrape class
+* [ ] Update the scrape configuration generator to use the scrape class
+* [ ] Update documentation
+* [ ] Resolve issues #4121, #3922
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -155,7 +155,7 @@ If the monitor resource specifies a scrape class name that isn't defined in the 
 
 This behavior is consistent with the behavior of monitor resources referencing a non-existing secret for bearer token authentication.
 
-To ensure users will have proper information about the error, the operator will emit an event with the error message on the monitor resource and also update the status of the monitor resource with the error message.
+To ensure users will have proper information about the error, the operator may (in the future) emit an event with the error message on the monitor resource and also update the status of the monitor resource with the error message.
 
 ## Test Plan
 
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -32,7 +32,7 @@ to compose the scrape configurations in a Kubernetes way.
 
 ## Non-Goals
 
-- Allow Prometheus owners to override the configuration defined in the monitoring resources (at least in the first iteration).
+- Provide a way for Prometheus administrators to enforce/override settings on scrape configurations.
 
 - Deprecation of the unsafe TLS settings in `ServiceMonitor`.
 
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -29,6 +29,7 @@ to compose the scrape configurations in a Kubernetes way.
 - Allow for the administrator to define a named, reusable scrape configuration snippet, including unsafe elements such as file references.
 - Allow for a user to select a configuration snippet by name in their probe/podmonitor/servicemonitor endpoint configuration.
 - Avoid giving the user the ability to use arbirary files within the Prometheus pod.
+
 ## Non-Goals
 
 - Allow Prometheus owners to override the configuration defined in the monitoring resources (at least in the first iteration).
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -21,8 +21,8 @@ Another motivation is to improve feature parity amongst the monitor resources. T
 ### Pitfalls of the current solution
 
 The only known solution for use cases where you'd need to use unsafe TLS settings is to use `additionalScrapeConfig`.
-The downside is obviously the loss of a great feature of the Prometheus Operator. The monitor resources make it possible
-to compose the scrape configurations in a Kubernetes way.
+
+The downside is the loss of integration provided by Prometheus Operator through monitor resources to compose the scrape configurations in a Kubernetes way.
 
 ## Goals
 
@@ -64,6 +64,7 @@ kind: Prometheus
 spec:
   # define scrape classes for use by the monitors
   # one class may be designated as the default class
+  # when a resource defines several default scrape classes, it should fail the reconciliation.
   scrapeClasses:
     - name: istio-mtls
       default: true
@@ -78,9 +79,9 @@ spec:
     - name: istio-certs
       mountPath: "/etc/istio-certs/"
   volumes:
-    - name: istio-certs
-      secret:
-        secretName: istio-certs
+  - emptyDir:
+      medium: Memory
+    name: istio-certs
 ```
 
 Any object references in the scrape class definition are assumed to refer to objects in the namespace of the `Prometheus` object.
--- a/Documentation/proposals/202305-scrapeclasses.md
+++ b/Documentation/proposals/202305-scrapeclasses.md
@@ -218,7 +218,7 @@ Objections:
 
 ## Action Plan
 
-* [ ] Change the Operator API to define scrape classes and for endpoint to select a scrape class
+* [ ] Change the Operator API to define scrape classes and for scrape resources to select a scrape class
 * [ ] Update the scrape configuration generator to use the scrape class
 * [ ] Update documentation
 * [ ] Resolve issues #4121, #3922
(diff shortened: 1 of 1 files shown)


nicolastakashi:
cc @ArthurSens @xiu @simonpasquier 

PS: Having more context about this and the other features we have on the pipeline to this Scrape Classe feature, I'd vote to have a dedicated CRD because I will explode Prometheus/Agent config.

Another use case is having this as dedicated CRD we can leverage same scrape config without duplications on users might have agent and servers on the same env.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,217 @@
...
+The only known solution for use cases where you'd need to use unsafe TLS settings is to use `additionalScrapeConfig`.
+The downside is obviously the loss of a great feature of the Prometheus Operator. The monitor resources make it possible
+to compose the scrape configurations in a Kubernetes way.
+
+## Goals
+
+- Allow for the administrator to define a named, reusable scrape configuration snippet, including unsafe elements such as file references.
+- Allow for a user to select a configuration snippet by name in their probe/podmonitor/servicemonitor endpoint configuration.
+- Avoid giving the user the ability to use arbirary files within the Prometheus pod.
+
```
simonpasquier, line 32:
```suggestion
## Non-Goals

- Allow Prometheus owners to override the configuration defined in the monitoring resources (at least in the first iteration).
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,217 @@
...
+## How
+
+The proposed solution is to introduce a notion of a *scrape class*, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/).
+A scrape class defines TLS settings (and possibly other settings in future, e.g. sensitive authorization settings) to be applied to all scrape configs of that class.
+
+One scrape class may be designated as the default class, in which case that class is applied to any scrape config that doesn't specify a value for `scrapeClass`.
+
+When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClass` field.
+For monitors, each endpoint may be assigned a scrape class.
```
simonpasquier, line 45:
```suggestion
When there's a match, a scrape class is assigned to all the endpoints.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,217 @@
...
+      tlsConfig:
+        caFile: "/etc/istio-certs/root-cert.pem"
+        certFile: "/etc/istio-certs/cert-chain.pem"
+        keyFile: "/etc/istio-certs/key.pem"
+        insecureSkipVerify: true
+
+  # mount the certs from the istio sidecar (shown here for illustration purposes)
+  volumeMounts:
+    - name: istio-certs
+      mountPath: "/etc/istio-certs/"
```
simonpasquier, line 74:
```suggestion
      mountPath: "/etc/istio-certs/"
  volumes:
    - name: istio-certs
      secret:
        secretName: istio-certs
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,217 @@
...
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+spec:
+  scrapeClass: istio-mtls
+  endpoints:
+  - port: http
+    path: /metrics
+```
+
+Out-of-scope: deprecation of the unsafe TLS settings in `ServiceMonitor`.
```
simonpasquier, line 122: can it be moved to the non-goals section?

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,217 @@
...
+spec:
+  scrapeClass: istio-mtls
+  podMetricsEndpoints:
+  - port: http
+    path: /metrics
+```
+
+The proposed behavior is:
+1. the `tlsConfig` in the associated scrape class is automatically applied to the scrape configuration of the endpoint.
+2. the inline `tlsConfig` (if any) takes precedence over the `tlsConfig` in the scrape class.
```
simonpasquier, line 95: What are we saying?
1. When a tlsConfig field in the pod monitor's endpoint isn't defined, the operator picks up the corresponding value from the scrape class (merge approach).
2. The tlsConfig fields from the scrape class are applied only when the endpoint's tlsConfig is nil.

The latter is simpler but less flexible (e.g. we may want to apply the same CA file to all pod monitors but leave some freedom on the other other TLS params).

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,217 @@
...
+  volumeMounts:
+    - name: istio-certs
+      mountPath: "/etc/istio-certs/"
+```
+
+Any object references in the scrape class definition are assumed to refer to objects in the namespace of the `Prometheus` object.
+
+### PodMonitor Resource
+
+Allow the user to select a scrape class for each endpoint.
```
simonpasquier, line 81:
```suggestion
Allow the user to select a scrape class which applies to all endpoints.
```

simonpasquier (commented):
Thanks for pursuing this!

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+  * https://github.com/prometheus-operator/prometheus-operator/issues/5947
+  * https://github.com/prometheus-operator/prometheus-operator/issues/5948
+
+This proposal introduces the concept of *scrape classes*, enabling users to utilize scrape configuration data provided by the administrator, through scrape objects such as PodMonitor, Probe, ServiceMonitor and ScrapeConfig.
+
+## Why
+
+Sometimes Prometheus administrators needs to provide default configurations for scrape objects, such as the definition of TLS certificates, when running Prometheus to scrape pods in an Istio mesh with strict mTLS as described in [Istio documentation](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings).
+
+Another motivation is to improve feature parity amongst the monitor resources. The `PodMonitor` and `Probe` resources aren't at parity with `ServiceMonitor` because the latter allows for unsafe TLS settings.
```
ArthurSens, line 19: This motivation sounds a bit weird, we can improve feature parity by allowing unsafe TLS on CRD that are still lacking 🤔

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+The only known solution for use cases where you'd need to use unsafe TLS settings is to use `additionalScrapeConfig`.
+The downside is obviously the loss of a great feature of the Prometheus Operator. The monitor resources make it possible
+to compose the scrape configurations in a Kubernetes way.
+
+## Goals
+
+- Allow for the administrator to define a named, reusable scrape configuration snippet, including unsafe elements such as file references.
+- Allow for a user to select a configuration snippet by name in their probe/podmonitor/servicemonitor endpoint configuration.
+- Avoid giving the user the ability to use arbirary files within the Prometheus pod.
+## Non-Goals
```
ArthurSens, line 32:
```suggestion
- Avoid giving the user the ability to use arbirary files within the Prometheus pod.

## Non-Goals
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+
+- Allow Prometheus owners to override the configuration defined in the monitoring resources (at least in the first iteration).
+### Audience
```
ArthurSens, line 35:
```suggestion
- Allow Prometheus owners to override the configuration defined in the monitoring resources (at least in the first iteration).

### Audience
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+    certFile: "/etc/istio-certs/cert-chain.pem"
+    keyFile: "/etc/istio-certs/key.pem"
+    insecureSkipVerify: true
+```
+
+An open question is whether the resource would be cluster-scoped or namespace-scoped.
+
+Objections:
+1. Since the file paths are dependent on the volume mounts in the server, this approach may not achieve a meaningful decoupling.
+2. Extra complexity in defining a new CRD.
```
ArthurSens, line 208: I don't see a problem with volumeMounts here...

The operator could inject volume and volume mounts into the statefulset if needed, I don't see why using a CRD ould force the user to manually set volumes

ArthurSens (commented):
I'm inclining a bit to the CRD alternative, but there are a few unknowns there too...

* Would we add a scrapeClassSelector field to Prometheus? To `.*Monitor`? All of them?
* What would we do if scrapeClassSelector selects more than one ScrapeClass? How would we merge them?


If we do it inline in the Prometheus CRD, we won't have this multiple ScrapeClass problem though 🤔

nicolastakashi:
> I'm inclining a bit to the CRD alternative, but there are a few unknowns there too...
> 
> * Would we add a scrapeClassSelector field to Prometheus? To `.*Monitor`? All of them?
> * What would we do if scrapeClassSelector selects more than one ScrapeClass? How would we merge them?
> 
> If we do it inline in the Prometheus CRD, we won't have this multiple ScrapeClass problem though 🤔

These are really good questions. 😄 

> * Would we add a scrapeClassSelector field to Prometheus? To `.*Monitor`? All of them?

I would say it keeps the same behavior as the original proposal, and the `.*Monitor` should specify which `scrape class` it wants to use. 

This is the actual behavior on Ingress Controllers as we can see [here](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource), so I believe we can mimic this.

> * What would we do if scrapeClassSelector selects more than one ScrapeClass? How would we merge them?

The idea is implement the same existing behavior on Ingress and Ingress Class.
You can have many ingress classes on the cluster, but the ingress object can only use one by defining the `ingressClassName` on the Ingress object.

That said, we can have many Scrape Classes on the cluster, but the `.*Monitor` can use only one by defining the field scrapeClassName to its spec.

Does it make sense?

Documentation/proposals/202305-scrapeclasses.md
nicolastakashi: I think it might be related.
https://github.com/prometheus-operator/prometheus-operator/issues/3922
It shouldn't exists even on Service Monitors from what I understood.
Maybe I might be wrong 😅

Documentation/proposals/202305-scrapeclasses.md
nicolastakashi: I think it could be quite complex.
Because the `Scrape Class` will not have any definition of the source of information.
Given the `tlsConfig` it's just pointing to which path the certificate is defined, I can't see Prometheus being able to mount volumes only based on that information and I'm not sure if adding a reference to volumes on the `Scrape Class` makes sense.

That said!
I can agree we have two touch points here and not 100% of decoupling, but it is what it is

ArthurSens:
Aaaaaaah ok, I get it! Yes, makes sense... so scrapeclass would a cluster scoped CRD?

Documentation/proposals/202305-scrapeclasses.md
nicolastakashi: The first one is what we're saying.
Btw @simonpasquier I've updated the description, wdyt?

xiu (commented):
Did a first pass, thanks for taking this one over!

xiu:
> Aaaaaaah ok, I get it! Yes, makes sense... so scrapeclass would be a cluster scoped CRD?

Why would it need to be cluster scoped? Is it to reuse a scrapeclass on multiple clusters? If so, we could use a selector like we do for the monitors and have each prometheus define how to select them. If we need multiple layers though (e.g. global scrape classes and local scrape classes), we might need multiple selectors.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+
+The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
+of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+spec:
+  # define scrape classes for use by the monitors
+  # one class may be designated as the default class
```
simonpasquier, line 66: (nit) When a resource defines several default scrape classes, it should fail the reconciliation.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+
+## How
+
+The proposed solution is to introduce a notion of a *scrape class*, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/).
+A scrape class defines TLS settings (and possibly other settings in future, e.g. sensitive authorization settings) to be applied to all scrape configs of that class.
+
+One scrape class may be designated as the default class, in which case that class is applied to any scrape config that doesn't specify a value for `scrapeClass`.
+
+When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClass` field.
+When there's a match, a scrape class is assigned to all the endpoints.
```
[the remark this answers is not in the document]
xiu: What then happens if a scrape class is removed? Do we remove the Monitor resources?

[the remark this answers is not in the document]
simonpasquier: @xiu yes, this should be the case.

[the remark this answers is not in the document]
nicolastakashi: I made a couple of tests with Ingress and Ingress Class since the proposal has been changed to work close to this concept.

I tested using NGINX Ingress Controller because it's what I'm most familiar with and I noticed the following behavior.

> What happens when scrapeClassName doesn't match a scrape class of the Prometheus object? The explicit option from my POV is to reject the monitoring object.

When I tried to create an Ingress object referencing an ingressClass that didn't exist I was able to create the manifest but I saw error logs on the ingress pods.

```
I1129 08:56:05.747184       7 store.go:403] "Ignoring ingress because of error while validating ingress class" ingress=" default/ingress-with-auth" error=" no object matching key \"nginx\" in local store"
W1129 08:56:08.715138       7 controller.go:331] ignoring ingress ingress-with-auth in default based on annotation: no object matching key "nginx" in local store
I1129 08:56:08.715190       7 main.go:107] "successfully validated configuration, accepting" ingress="default/ingress-with-auth"
I1129 08:56:08.719779       7 store.go:436] "Ignoring ingress because of error while validating ingress class" ingress=" default/ingress-with-auth" error=" no object matching key \"nginx\" in local store"
```

I don't think this is the Best UX and I think we should reject the object is the class didn't exist, probable through admission webhook to provide a better experience. 

> What then happens if a scrape class is removed? Do we remove the Monitor resources?
I was able to delete the ingress class but when the controller reconciled the ingress object I got the same error as above.

Maybe we could add some finalizers to the scrape class to avoid this issue, tbh I'm not sure about the best strategy here.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,225 @@
...
+## Why
+
+Sometimes Prometheus administrators needs to provide default configurations for scrape objects, such as the definition of TLS certificates, when running Prometheus to scrape pods in an Istio mesh with strict mTLS as described in [Istio documentation](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings).
+
+Another motivation is to improve feature parity amongst the monitor resources. The `PodMonitor` and `Probe` resources aren't at parity with `ServiceMonitor` because the latter allows for unsafe TLS settings.
+
+### Pitfalls of the current solution
+
+The only known solution for use cases where you'd need to use unsafe TLS settings is to use `additionalScrapeConfig`.
+The downside is obviously the loss of a great feature of the Prometheus Operator. The monitor resources make it possible
```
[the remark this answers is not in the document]
nicolastakashi: Yes, integration is a better word and feature on that case.
Updated can you review?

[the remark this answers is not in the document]
xiu: LGTM 👍

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
simonpasquier: > Maybe we could add some finalizers to the scrape class

Since the scrape class definition is part of the Prometheus CR, I'm not sure how finalizers would help.

To be clear, I'm fine with the operator rejecting a monitoring object that references a non-existing scrape class name. We have https://github.com/prometheus-operator/prometheus-operator/issues/4481 for implementing admission webhook for monitoring resources + https://github.com/prometheus-operator/prometheus-operator/pull/6030 for emitting Kubernetes events on rejection.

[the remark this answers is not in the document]
simonpasquier: >  I'm more in favor of dedicated CR tbh looking forward it will be easier to maintain.

Reading the proposal, I assumed that the ScrapeClass CRD wasn't the chosen option. I'm not convinced that a ScrapeClass CRD is simpler than defining the scrape classes inside the Prometheus resource.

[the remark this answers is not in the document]
nicolastakashi: At the moment I copied the proposal as is from the original PR.
I have no doubts of having Scrape Classe at Prometheus Level is easier to implement and I do agree it provide a better DX while handling volumes for example.

But in the other hand it's limiting if you want to leverage same scrape classe between Agent and Server and also it may increase Prometheus Object a lot, especially after features such as default relabel configs be added to the Scrape Class.

Another thing to consider is having a Scrape Class CRD makes the Operator more close to the Ingress implementation which is something well know by the  k8s users.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,223 @@
...
+When there's a match, a scrape class is assigned to all the endpoints.
+
+Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
+
+### Prometheus Resource
+
+It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a new section for defining scrape classes.
+
+The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
+of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource.
```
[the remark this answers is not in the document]
ArthurSens: Well, I can see that new RBAC rules will be needed if going with the new CRD approach 🤔

Documentation/proposals/202305-scrapeclasses.md
ArthurSens: If we're defining TLS configuration in scrapeClasses, is there are reason to not specify the secret references here as well?

I couldn't understand why adding volume references in scrape classes wouldn't make sense 🤔

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
simonpasquier: Speaking of similarity with IngressClass, should a ScrapeClass CRD be namespace or cluster scoped?

* if namespace scoped, reusing a scrape class for different Prometheus/PrometheusAgent objects becomes limited. 
* if cluster scoped, a ScrapeClass is probably going to match a small number of Prometheus/PrometheusAgent object, due to the volume mounts.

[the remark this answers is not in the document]
nicolastakashi: Yes!
I see your point the volume mounts couples the scrape class to the prometheus objects indeed and this makes a bit difficult to reuse cross prometheus instances. The solution I see for this is to have volumes definitions on the scrape class, but I'm not convinced about this one.

Ofc we can move forward with the ScrapeClass being defined on the Prometheus object and in the further if we think it makes more sense move to a dedicated CRD we can plan a work towards the new requirements/design and update this doc.

WDYT?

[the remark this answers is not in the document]
nicolastakashi: Cool 😄 
So just to summarise, let's keep the behaviour we described above.

- The Operator will not allow multiple default scrape classes
- If the endpoint selects a scrape class that didn't exists the Operator will reject the resource.
- If the delete the scrape class in use the Operator will fail the prometheus reconcile.

RIght?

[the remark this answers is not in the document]
ArthurSens: Sounds correct!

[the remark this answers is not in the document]
simonpasquier: > If the delete the scrape class in use the Operator will fail the prometheus reconcile.

IMHO it should behave the same as if the service monitor specifies a scrape class that doesn't exist

1. The Prometheus object is created with a scrape class named `foo`.
1. Service monitor A is created with `scrapeClassName: foo` -> the service monitor is selected by the Prometheus controller.
2. The Prometheus object is updated resulting in the scrape class `foo` being removed or renamed.
3. Service monitor A is rejected on the next reconciliation.

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,224 @@
...
+- Deprecation of the unsafe TLS settings in `ServiceMonitor`.
+
+### Audience
+
+- Users who serve Prometheus as a service and want to give their customers autonomy in defining monitors, but want to provide a default configuration for scraping.
+
+## How
+
+The proposed solution is to introduce a notion of a *scrape class*, akin to a Kubernetes [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/).
+A scrape class defines TLS settings (and possibly other settings in future, e.g. sensitive authorization settings) to be applied to all scrape configs of that class.
```
simonpasquier, line 45:
```suggestion
A scrape class defines TLS settings (and possibly other settings in future, e.g. sensitive authorization settings) to be applied to all scrape resources (ServiceMonitor, PodMonitor, Probe and ScrapeConfig) of that class.
```

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
ArthurSens: Hmmmm, sounds like a difficult thing to debug in case that happens.

We probably want to make sure we have logs/metrics/events that allow troubleshooting, also docs!

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,224 @@
...
+
+One scrape class may be designated as the default class, in which case that class is applied to any scrape config that doesn't specify a value for `scrapeClass`.
+
+When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClass` field.
```
simonpasquier, line 49: (nit)
```suggestion
When defining a podmonitor/servicemonitor/probe/scrapeconfig, a user may assign a scrape class via the `scrapeClassName field.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,224 @@
...
+When there's a match, a scrape class is assigned to all the endpoints.
+
+Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
+
+### Prometheus Resource
+
+It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a new section for defining scrape classes.
```
simonpasquier, line 56:
```suggestion
It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a new field for defining scrape classes.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,224 @@
...
+The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
+of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+spec:
+  # define scrape classes for use by the monitors
+  # one class may be designated as the default class
+  # when a resource defines several default scrape classes, it should fail the reconciliation.
```
simonpasquier, line 67: (nit) can you move L66-67 outside the code block to make them more visible?

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,224 @@
...
+  podMetricsEndpoints:
+  - port: http
+    path: /metrics
+```
+
+If the `Monitor` resource has a `tlsConfig` field defined, the Operator will use a merge strategy to combine the `tlsConfig` fields from the PodMonitor object with the `tlsConfig` fields of the scrape class, the `tlsConfig` fields in the `PodMonitor` resource take precedence.
+
+### Probe Resource
+
+Allow the user to select a scrape class for the probe.
```
simonpasquier, line 107:
```suggestion
Allow the user to select a scrape class for the prober service.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,224 @@
...
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+spec:
+  scrapeClassName: istio-mtls
+  podMetricsEndpoints:
+  - port: http
+    path: /metrics
+```
+
+If the `Monitor` resource has a `tlsConfig` field defined, the Operator will use a merge strategy to combine the `tlsConfig` fields from the PodMonitor object with the `tlsConfig` fields of the scrape class, the `tlsConfig` fields in the `PodMonitor` resource take precedence.
```
simonpasquier, line 103:
```suggestion
If the `Monitor` resource has a `tlsConfig` field defined, the Operator will use a merge strategy to combine the `tlsConfig` fields from the PodMonitor object with the `tlsConfig` fields of the scrape class, the `tlsConfig` fields in the `PodMonitor` resource take precedence.

If the resource specifies a scrape class name that isn't defined in the Prometheus/PrometheusAgent object, then the scrape resource is dropped by the operator.

If the resource doesn't specify a scrape class name and the Prometheus/PrometheusAgent object defines a default scrape class, the operator will act as if the PodMonitor resource specified the default scrape class name.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,224 @@
...
+
+Another alternative would be to allow `PodMonitor` and `Probe` to use unsafe TLS settings.
+
+Objections:
+1. See [explanation](https://github.com/prometheus-operator/prometheus-operator/issues/3922#issuecomment-802899950) for why
+   unsafe settings were disallowed in the first place.
+
+## Action Plan
+
+* [ ] Change the Operator API to define scrape classes and for endpoint to select a scrape class
```
simonpasquier, line 221: (nit)
```suggestion
* [ ] Change the Operator API to define scrape classes and for scrape resources to select a scrape class
```

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
simonpasquier: I agree but it's not very different from a ServiceMonitor referring to a non-existing secret key for bearer token. #3611 and #3335 should help.

simonpasquier (commented):
LGTM minus my last 2 small comments.

Documentation/proposals/202305-scrapeclasses.md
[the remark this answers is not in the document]
nicolastakashi: That's make total sense!
I've added this information to the doc cc @simonpasquier @ArthurSens

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,235 @@
...
+Class names are assumed to be installation-specific. In practice, some common class names like `istio-mtls` are likely to emerge.
+
+### Prometheus Resource
+
+It is proposed that the `Prometheus` and `PrometheusAgent` resources contain a new field for defining scrape classes.
+
+The rationale for defining scrape classes inline is that, in practice, the TLS file paths are closely related to the `volumeMounts`
+of the `Prometheus` spec. An alternative is outlined later, of factoring the class definitions into a separate resource.
+
+Define a scrape classes for use by the monitors, one class may be designated as the default class.
```
simonpasquier, line 61: (nit) not sure that it reads properly
```suggestion
One (and only one) scrape class may be designated as the default class.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,235 @@
...
+    [...]
+  fileSDConfig:
+    [...]
+```
+
+If the monitor resource specifies a scrape class name that isn't defined in the Prometheus/PrometheusAgent object, then the scrape resource is rejected by the operator.
+
+This behavior is consistent with the behavior of monitor resources referencing a non-existing secret for bearer token authentication.
+
+To ensure users will have proper information about the error, the operator will emit an event with the error message on the monitor resource and also update the status of the monitor resource with the error message.
```
simonpasquier, line 158: This isn't the case right now 

```suggestion
To ensure users will have proper information about the error, the operator may (in the future) emit an event with the error message on the monitor resource and also update the status of the monitor resource with the error message.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,235 @@
...
+  scrapeClassName: istio-mtls
+  staticConfigs:
+    [...]
+  httpSDConfig:
+    [...]
+  fileSDConfig:
+    [...]
+```
+
+If the monitor resource specifies a scrape class name that isn't defined in the Prometheus/PrometheusAgent object, then the scrape resource is rejected by the operator.
```
simonpasquier, line 154: (nit) it repeats a bit what is written  for PodMonitor. Maybe reorganize like this?

````
### Monitoring resources

If the monitor resource specifies a scrape class name ...

#### PodMonitor

< insert YAML example >

### Probe

Allow the user to select a scrape class for the prober service.

< insert YAML example >

...
````

File: Documentation/proposals/202305-scrapeclasses.md [hunk 25]
```diff
@@ -0,0 +1,235 @@
...
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Probe
+spec:
+  scrapeClassName: istio-mtls
+```
+
+### ServiceMonitor Resource
+
+Allow the user to select a scrape class for each endpoint.
```
simonpasquier, line 123:
```suggestion
Allow the user to select a scrape class for all endpoints.
```

Documentation/proposals/202305-scrapeclasses.md
```diff
@@ -0,0 +1,235 @@
...
+spec:
+  scrapeClassName: istio-mtls
+  endpoints:
+  - port: http
+    path: /metrics
+```
+
+### ScrapeConfig
+
+Allow the user to select a scrape class for the generic scrape configuration.
```
simonpasquier, line 137:
```suggestion
Allow the user to select a scrape class for the whole scrape configuration.
```

simonpasquier (approved):
LGTM, thanks!

ArthurSens (approved):
Thanks @nicolastakashi for taking this over the finish line!

PR #6127, opened 2023-11-28
chore: use apiserver controller to reload certs

simonpasquier:
## Description

This commit removes the dependency on
`github.com/brancz/kube-rbac-proxy` which was only needed to reload the TLS certificate and key files dynamically.

Instead the operator and admission webhook server use the `DynamicServingCertificateController` from `k8s.io/apiserver` which is a proven solution.

## Type of change

## Changelog entry

```release-note

```


simonpasquier:
I've also verified with the OpenShift e2e tests which make use of the TLS configuration for the admission webhook and it passed (https://github.com/openshift/prometheus-operator/pull/261).

philipgough (approved):
lgtm

PR #6133, opened 2023-11-29
jsonnet: Create alert for config-reloader apply errors

aeber:
## Description

If a syntactically correct servicemonitor is deployed that contains some other error (such as missing variable expansion as described here: https://github.com/prometheus-operator/prometheus-operator/issues/3644), there is currently no alert to detect this.
In this case `reloader_last_reload_successful` is not set to `0` and therefore the `ConfigReloaderSidecarErrors` alert does not apply.

## Type of change

## Changelog entry

```release-note
Create alert for config-reloader apply errors
```

### DIFF
--- a/example/mixin/alerts.yaml
+++ b/example/mixin/alerts.yaml
@@ -82,7 +82,7 @@ groups:
       description: |-
         Errors encountered while the {{$labels.pod}} config-reloader sidecar attempts to apply config in {{$labels.namespace}} namespace.
         As a result, configuration for service running in {{$labels.pod}} may be stale and cannot be updated anymore.
-      summary: config-reloader sidecar was unable to apply new config for 5m
+      summary: config-reloader sidecar was unable to apply new config
     expr: |
       increase(reloader_config_apply_operations_failed_total{namespace=~".+"}[5m]) > 0
     for: 5m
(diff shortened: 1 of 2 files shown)


example/mixin/alerts.yaml
```diff
@@ -77,3 +77,14 @@ groups:
     for: 10m
     labels:
       severity: warning
+  - alert: ConfigReloaderSidecarApplyErrors
+    annotations:
+      description: |-
+        Errors encountered while the {{$labels.pod}} config-reloader sidecar attempts to apply config in {{$labels.namespace}} namespace.
+        As a result, configuration for service running in {{$labels.pod}} may be stale and cannot be updated anymore.
+      summary: config-reloader sidecar was unable to apply new config for 5m
```
paulfantom, line 85:
```suggestion
      summary: config-reloader sidecar was unable to apply new config
```

Just nitpicking, but if someone overrides `FOR` statement in alert, this summary won't be correct anymore.

paulfantom:
I am actually curious now why `reloader_last_reload_successful` is not set to `0` when this happens. Especially since reload wasn't fully successful. :thinking:

aeber:
> Yes I agree that we might want to fix the reload metrics rather than adding another alert.

I would be very happy with that too. Unfortunately, I currently lack the time to debug this further.

PR #6135, opened 2023-11-30
feat(scrapeConfigs): Add sharding to scrapeConfigs

### LINKED ISSUE #5637: ScrapeConfig objects don't work properly when Prometheus sharding is used
`generateScrapeConfig` gets the shard number as an input parameter but it doesn't make use of it. As a result, targets from ScrapeConfigs are scraped by all shards while they should be distributed.
 
https://github.com/prometheus-operator/prometheus-operator/blob/d3c37ea81bf3afcd7030819b054f65780e69a237/pkg/prometheus/promcfg.go#L2135-L2140

adinhodovic:
## Description

Fixes #5637.

## Type of change

## Changelog entry

```release-note
feat(scrapeConfigs): Add sharding to scrapeConfigs
```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -540,7 +540,7 @@ func (cg *ConfigGenerator) GenerateServerConfiguration(
 	scrapeConfigs = cg.appendServiceMonitorConfigs(scrapeConfigs, sMons, apiserverConfig, store, shards)
 	scrapeConfigs = cg.appendPodMonitorConfigs(scrapeConfigs, pMons, apiserverConfig, store, shards)
 	scrapeConfigs = cg.appendProbeConfigs(scrapeConfigs, probes, apiserverConfig, store, shards)
-	scrapeConfigs, err := cg.appendScrapeConfigs(ctx, scrapeConfigs, sCons, store)
+	scrapeConfigs, err := cg.appendScrapeConfigs(ctx, scrapeConfigs, sCons, store, shards)
 	if err != nil {
 		return nil, fmt.Errorf("generate scrape configs: %w", err)
 	}
@@ -1444,6 +1444,18 @@ func generateAddressShardingRelabelingRules(relabelings []yaml.MapSlice, shards 
 	return generateAddressShardingRelabelingRulesWithSourceLabel(relabelings, shards, "__address__")
 }
 
+func (cg *ConfigGenerator) generateAddressShardingRelabelingRulesIfMissing(relabelings []yaml.MapSlice, shards int32) []yaml.MapSlice {
+	for i, relabeling := range relabelings {
+		for _, relabelItem := range relabeling {
+			if relabelItem.Key == "action" && relabelItem.Value == "hashmod" {
+				level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "idx", i)
+				return relabelings
+			}
+		}
+	}
+	return generateAddressShardingRelabelingRules(relabelings, shards)
+}
+
 func generateAddressShardingRelabelingRulesForProbes(relabelings []yaml.MapSlice, shards int32) []yaml.MapSlice {
 	return generateAddressShardingRelabelingRulesWithSourceLabel(relabelings, shards, "__param_target")
 }
@@ -1680,7 +1692,6 @@ func (cg *ConfigGenerator) generateAdditionalScrapeConfigs(
 		var addlScrapeConfig yaml.MapSlice
 		var relabelings []yaml.MapSlice
 		var otherConfigItems []yaml.MapItem
-		var alreadyHasShardLabel = false
 
 		for _, mapItem := range mapSlice {
 			if mapItem.Key != "relabel_configs" {
@@ -1696,18 +1707,10 @@ func (cg *ConfigGenerator) generateAdditionalScrapeConfigs(
 				if !ok {
 					return nil, fmt.Errorf("error parsing relabel config: %w", err)
 				}
-				for _, relabelItem := range relabeling {
-					if relabelItem.Key == "action" && relabelItem.Value == "hashmod" {
-						level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "relabeling", relabeling)
-						alreadyHasShardLabel = true
-					}
-				}
 				relabelings = append(relabelings, relabeling)
 			}
 		}
-		if !alreadyHasShardLabel {
-			relabelings = generateAddressShardingRelabelingRules(relabelings, shards)
-		}
+		relabelings = cg.generateAddressShardingRelabelingRulesIfMissing(relabelings, shards)
 		addlScrapeConfig = append(addlScrapeConfig, otherConfigItems...)
 		addlScrapeConfig = append(addlScrapeConfig, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
 		addlScrapeConfigs = append(addlScrapeConfigs, addlScrapeConfig)
@@ -2224,7 +2227,7 @@ func (cg *ConfigGenerator) GenerateAgentConfiguration(
 	scrapeConfigs = cg.appendServiceMonitorConfigs(scrapeConfigs, sMons, apiserverConfig, store, shards)
 	scrapeConfigs = cg.appendPodMonitorConfigs(scrapeConfigs, pMons, apiserverConfig, store, shards)
 	scrapeConfigs = cg.appendProbeConfigs(scrapeConfigs, probes, apiserverConfig, store, shards)
-	scrapeConfigs, err := cg.appendScrapeConfigs(ctx, scrapeConfigs, sCons, store)
+	scrapeConfigs, err := cg.appendScrapeConfigs(ctx, scrapeConfigs, sCons, store, shards)
 	if err != nil {
 		return nil, fmt.Errorf("generate scrape configs: %w", err)
 	}
@@ -2259,7 +2262,8 @@ func (cg *ConfigGenerator) appendScrapeConfigs(
 	ctx context.Context,
 	slices []yaml.MapSlice,
 	scrapeConfigs map[string]*monitoringv1alpha1.ScrapeConfig,
-	store *assets.Store) ([]yaml.MapSlice, error) {
+	store *assets.Store,
+	shards int32) ([]yaml.MapSlice, error) {
 	scrapeConfigIdentifiers := make([]string, len(scrapeConfigs))
 	i := 0
 	for k := range scrapeConfigs {
@@ -2272,7 +2276,7 @@ func (cg *ConfigGenerator) appendScrapeConfigs(
 
 	for _, identifier := range scrapeConfigIdentifiers {
 		cfgGenerator := cg.WithKeyVals("scrapeconfig", identifier)
-		scrapeConfig, err := cfgGenerator.generateScrapeConfig(ctx, scrapeConfigs[identifier], store)
+		scrapeConfig, err := cfgGenerator.generateScrapeConfig(ctx, scrapeConfigs[identifier], store, shards)
 
 		if err != nil {
 			return slices, err
@@ -2288,6 +2292,7 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 	ctx context.Context,
 	sc *monitoringv1alpha1.ScrapeConfig,
 	store *assets.Store,
+	shards int32,
 ) (yaml.MapSlice, error) {
 	jobName := fmt.Sprintf("scrapeConfig/%s/%s", sc.Namespace, sc.Name)
 	cfg := yaml.MapSlice{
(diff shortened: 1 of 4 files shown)


simonpasquier (commented):
Since ScrapeConfigs offer a low-level interface to the Prometheus configuration, I'd expect that users want to override the automated sharding in some cases. Which means that we may need something similar to what you implemented in 5c8245927a which checks the presence of hashmod action. WDYT?

pkg/prometheus/promcfg.go
```diff
@@ -2317,6 +2319,24 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 
 	if sc.Spec.RelabelConfigs != nil {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
+	}
+
+	if shards != 1 {
```
simonpasquier, line 2324: I wouldn't make a special case for shard == 1

pkg/prometheus/promcfg.go
```diff
@@ -2317,6 +2319,24 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+		alreadyHasShardLabel := false
+		if sc.Spec.RelabelConfigs != nil {
+			for _, relabelConfig := range sc.Spec.RelabelConfigs {
+				if relabelConfig.Action == "hashmod" {
+					level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "relabeling", relabelConfig)
+					alreadyHasShardLabel = true
```
simonpasquier, line 2330:
```suggestion
					alreadyHasShardLabel = true
					break
```

pkg/prometheus/promcfg.go
```diff
@@ -2317,6 +2319,24 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 
 	if sc.Spec.RelabelConfigs != nil {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
+	}
+
+	if shards != 1 {
+		alreadyHasShardLabel := false
+		if sc.Spec.RelabelConfigs != nil {
```
simonpasquier, line 2326: range works with nil slices too.
```suggestion
```

pkg/prometheus/promcfg.go
```diff
@@ -2317,6 +2319,24 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 
 	if sc.Spec.RelabelConfigs != nil {
```
simonpasquier, line 2320: (nit) not part of this change but checking the length of the slice would be more idiomatic.
```suggestion
	if len(sc.Spec.RelabelConfigs) > 0 {
```

pkg/prometheus/promcfg.go
```diff
@@ -2317,6 +2319,24 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+					alreadyHasShardLabel = true
+				}
+			}
+		}
+		if !alreadyHasShardLabel {
+			relabelings = generateAddressShardingRelabelingRules(relabelings, shards)
+		}
+	}
+
+	if sc.Spec.RelabelConfigs != nil || shards != 1 {
```
simonpasquier, line 2339: do we need a condition here? At this point, `relabelings` has at least 1 item.

simonpasquier (commented):
Thanks! Just a few nits.

pkg/prometheus/promcfg.go
```diff
@@ -2317,6 +2319,24 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 
 	if sc.Spec.RelabelConfigs != nil {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
+	}
+
+	if shards != 1 {
+		alreadyHasShardLabel := false
+		if sc.Spec.RelabelConfigs != nil {
+			for _, relabelConfig := range sc.Spec.RelabelConfigs {
+				if relabelConfig.Action == "hashmod" {
```
leviliangtw, line 2328: Hello, I have a suggestion to make the code more readable. 
Here you check the action of "hashmod" as the indicator of "has Shard Label or not"
Is it possible to change the target label "__tmp_hash"  to something like "__tmp_shard_hash"?
Then, you can check the target label "__tmp_shard_hash" and make sure that this is of the purpose for sharding. 
Also, it could also avoid misjudgment if somebody just add their own "hashmod" action for other purpose.

simonpasquier: @leviliangtw I'm not sure to understand your point about the "__tmp_shard_hash" label?

I'm not aware of other use cases for the "hashmod" action than target sharding so I'm fine with the current approach but we might want to convey the nuance in the docs (e.g. "if the relabelConfigs list contains an item with `action: hashmod` then the operator will not apply the automatic target sharding.").

pkg/prometheus/promcfg.go
adinhodovic: Had a similar approach to https://github.com/adinhodovic/prometheus-operator/blob/a65089818bb12c1d573650554b74d9fcc11e9fb5/pkg/prometheus/promcfg.go#L1668-L1670.

Fine with removing it, however the changeset is noisy due to all sharding relabel configs added to each scrape config as you can see in the commit.

pkg/prometheus/promcfg.go
adinhodovic: No, not if we add sharding relabels for non-sharded setups. Fixed.

pkg/prometheus/promcfg.go
```diff
@@ -2318,11 +2320,23 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+	for _, relabelConfig := range sc.Spec.RelabelConfigs {
+		if relabelConfig.Action == "hashmod" {
+			level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "relabeling", relabelConfig)
+			alreadyHasShardLabel = true
+			break
+		}
+	}
+	if !alreadyHasShardLabel {
+		relabelings = generateAddressShardingRelabelingRules(relabelings, shards)
+	}
```
ArthurSens, line 2337: I think we want to wrap this inside a conditional `if shards > 1`

Seems wrong to add shard-related relabelings even though Prometheus isn't sharded 🤔

pkg/prometheus/promcfg.go
ArthurSens: I think you removed the conditional you had above, which means we need the condition back 😅 

IMHO we should keep the conditional before we get here, i.e. on line 2326

pkg/prometheus/promcfg.go
simonpasquier: The initial reasoning was probably that the extra branch makes the code a bit less readable for little gain.

pkg/prometheus/promcfg.go
```diff
@@ -2318,11 +2320,23 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
-	if sc.Spec.RelabelConfigs != nil {
+	if len(sc.Spec.RelabelConfigs) > 0 {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
-		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
 	}
 
+	alreadyHasShardLabel := false
+	for _, relabelConfig := range sc.Spec.RelabelConfigs {
+		if relabelConfig.Action == "hashmod" {
+			level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "relabeling", relabelConfig)
```
simonpasquier, line 2330: (nit) RelabelConfig doesn't implement the Stringer interface so it won't log cleanly.
```suggestion
			level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "idx", i)
```

pkg/prometheus/promcfg.go
```diff
@@ -2318,11 +2320,23 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
 	}
 
-	if sc.Spec.RelabelConfigs != nil {
+	if len(sc.Spec.RelabelConfigs) > 0 {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
-		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
 	}
 
+	alreadyHasShardLabel := false
+	for _, relabelConfig := range sc.Spec.RelabelConfigs {
```
simonpasquier, line 2328:
```suggestion
	for i, relabelConfig := range sc.Spec.RelabelConfigs {
```

pkg/prometheus/promcfg.go
ArthurSens: I imagine that Prometheus will need to perform extra CPU work for nothing if we add a no-op relabeling rule. I don't know exactly how much of an impact this could have, but I can imagine a few CPU cycles going to waste for very large-scale Prometheus :P

adinhodovic: I've reverted the change and the skipping non-sharded configs will be done again now. I agree on not adding a relabelling rule, might be confusing for users that do not shard as well.

pkg/prometheus/promcfg.go
```diff
@@ -2318,8 +2320,25 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+				alreadyHasShardLabel = true
+				break
+			}
+		}
+		if !alreadyHasShardLabel {
+			relabelings = generateAddressShardingRelabelingRules(relabelings, shards)
+		}
+	}
+
+	if len(sc.Spec.RelabelConfigs) > 0 || shards != 1 {
```
simonpasquier, line 2341:
```suggestion
	if len(relabelings) > 0 {
```

pkg/prometheus/promcfg.go
```diff
@@ -2318,8 +2320,25 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
 	}
 
-	if sc.Spec.RelabelConfigs != nil {
+	if len(sc.Spec.RelabelConfigs) > 0 {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
+	}
+
+	if shards != 1 {
+		alreadyHasShardLabel := false
+		for i, relabelConfig := range sc.Spec.RelabelConfigs {
```
simonpasquier, line 2329: (nit)
```suggestion
		for i, relabelConfig := range relabelconfigs {
```

pkg/prometheus/promcfg.go
```diff
@@ -2318,8 +2320,25 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+			if relabelConfig.Action == "hashmod" {
+				level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "idx", i)
+				alreadyHasShardLabel = true
+				break
+			}
+		}
+		if !alreadyHasShardLabel {
+			relabelings = generateAddressShardingRelabelingRules(relabelings, shards)
```
simonpasquier, line 2337: (suggestion) we could also have a new `generateAddressShardingRelabelingRulesIfMissing` function (sorry for the long name, feel free to use something shorter!). It would make code a bit more readable IMHO.

ArthurSens (approved):
LGTM! Sorry for the whole ping-pong with the 1 shard situation 😬

pkg/prometheus/promcfg.go
```diff
@@ -1444,6 +1444,22 @@ func generateAddressShardingRelabelingRules(relabelings []yaml.MapSlice, shards
...
+			if relabelItem.Key == "action" && relabelItem.Value == "hashmod" {
+				level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "idx", i)
+				alreadyHasShardLabel = true
+			}
+		}
+	}
+	if !alreadyHasShardLabel {
+		return generateAddressShardingRelabelingRules(relabelings, shards)
+	}
+	return relabelings
```
simonpasquier, line 1460: just a suggestion, feel free to ignore
```suggestion
	for i, relabeling := range relabelings {
		for _, relabelItem := range relabeling {
			if relabelItem.Key == "action" && relabelItem.Value == "hashmod" {
				level.Debug(cg.logger).Log("msg", "found existing hashmod relabeling rule, skipping", "idx", i)
				return generateAddressShardingRelabelingRules(relabelings, shards)
			}
		}
	}

	return relabelings
```

adinhodovic: cleaner :+1:

simonpasquier (approved):
Thanks!

PR #6137, opened 2023-12-01
chore: Adding StartupProbeTimeoutSeconds to Prometheus CRD

### LINKED ISSUE #4730: Configurable startup/liveness/readiness Probe
**What is missing?**
I would like to have possibility to change startup/liveness/readiness probes parameters in statefulset via `Prometheus` CRD.

**Why do we need it?**
Currently my prometheus stucked in restart loop (due to long WAL-replay) and k8s is killing container before it gets ready. I would like to extend startupProbe.

**Environment**

    `v0.55.0`
    
    

NA

nicolastakashi:
## Description

Adding StartupProbeTimeoutSeconds to the Prometheus CRD to allow users set a custom StartupProbe configuration

Resolves #4730

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that applies._

- [x] `CHANGE` (fix or feature that would cause existing functionality not to work as expected)

## Changelog entry

```release-note
Adding StartupProbeTimeoutSeconds to Prometheus CRD 
```


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -621,6 +621,10 @@ type CommonPrometheusFields struct {
 	// If not specified, the configuration is reloaded using the /-/reload HTTP endpoint.
 	// +optional
 	ReloadStrategy *ReloadStrategyType `json:"reloadStrategy,omitempty"`
+
+	// Defines how long to wait for starting liveness and readiness probes.
+	// +optional
+	StartupProbeTimeoutSeconds *int32 `json:"startupProbeTimeoutSeconds,omitempty"`
```
simonpasquier, line 627: `...TimeoutSeconds` is confusing because the v1.Probe struct has a `TimeoutSeconds` field. Maybe this?
```suggestion
	MaximumStartupDurationSeconds *int32 `json:"maximumStartupDurationSeconds,omitempty"`
```

nicolastakashi:
> How do we deal with rounding? E.g. what happens when the user says "max duration = 159s"? I'd be inclined to say that we round up if necessary.

Good catch!
I agree round it up 👍🏽

nicolastakashi:
> > How do we deal with rounding? E.g. what happens when the user says "max duration = 159s"? I'd be inclined to say that we round up if necessary.
> 
> Good catch! I agree round it up 👍🏽

@simonpasquier TIL something new hehe.
When we perform division with two integer operands, the result is also an integer and any fractional part is truncated (rounded down). 

We have this info also available on Go docs [here](https://arc.net/l/quote/gselxkdv) it's called truncated division.

I also added some unit tests to ensure the behavior

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -621,6 +621,11 @@ type CommonPrometheusFields struct {
 	// If not specified, the configuration is reloaded using the /-/reload HTTP endpoint.
 	// +optional
 	ReloadStrategy *ReloadStrategyType `json:"reloadStrategy,omitempty"`
+
+	// Defines the maximum time that the `prometheus` container's startup probe will wait before being considered failed.
+	// If unset, it will be equal to 600 seconds (= 15 minutes).
+	// +optional
+	MaximumStartupDurationSeconds *int32 `json:"maximumStartupDurationSeconds,omitempty"`
```
simonpasquier, line 628: A few more edge cases to consider:
* what if we have a zero value?
* what if we have a negative value?

For simplicity, we could consider that it's equivalent to the value being nil. Or consider that 0 disables the probe?

nicolastakashi: I prefer keep the default behavior if use set to zero we can use the default probe we have.

nicolastakashi: I'm wondering if we should have a min value of 60 for example, to ensure it will never try to divide by 0

simonpasquier: Hmm not sure to follow. If we define the increase "step" to be 10s then any value here between 1 and 10 will be translated to 10 (and [11,20) -> 20, ...).

nicolastakashi: Sorry!
I don't get you.
Could you come again?

pkg/prometheus/statefulset.go
```diff
@@ -496,3 +496,15 @@ func ShareProcessNamespace(p monitoringv1.PrometheusInterface) *bool {
 		) == monitoringv1.ProcessSignalReloadStrategyType,
 	)
 }
+
+func GetStatupProbePeriodSecondsAndFailureThreshold(cfp monitoringv1.CommonPrometheusFields) (int32, int32) {
+	var startupPeriodSeconds int32 = 15
+	var startupFailureThreshold int32 = 60
+
+	if cfp.MaximumStartupDurationSeconds != nil && *cfp.MaximumStartupDurationSeconds > 0 {
```
simonpasquier, line 504: (nit)
```suggestion
	if ptr.Deref(cfp.MaximumStartupDurationSeconds, 0) > 0 {
```

pkg/prometheus/server/statefulset_test.go
```diff
@@ -2906,3 +2906,63 @@ func TestPersistentVolumeClaimRetentionPolicy(t *testing.T) {
 		t.Fatalf("expected persistentVolumeClaimDeletePolicy.WhenScaled to be %s but got %s", appsv1.DeletePersistentVolumeClaimRetentionPolicyType, sset.Spec.PersistentVolumeClaimRetentionPolicy.WhenScaled)
 	}
 }
+
+func TestStartupProbeTimeoutSeconds(t *testing.T) {
+	tests := []struct {
+		maximumStartupDurationSeconds   *int32
+		expectedStartupPeriodSeconds    int32
+		expectedStartupFailureThreshold int32
+	}{
```
simonpasquier, line 2915: I would add a test case with `maximumStartupDurationSeconds: 1`.

pkg/prometheus/server/statefulset_test.go
```diff
@@ -2906,3 +2906,63 @@ func TestPersistentVolumeClaimRetentionPolicy(t *testing.T) {
...
+		},
+		{
+			maximumStartupDurationSeconds:   ptr.To(int32(1200)),
+			expectedStartupPeriodSeconds:    20,
+			expectedStartupFailureThreshold: 60,
+		},
+		{
+			maximumStartupDurationSeconds:   ptr.To(int32(129)),
+			expectedStartupPeriodSeconds:    2,
+			expectedStartupFailureThreshold: 64,
```
simonpasquier, line 2944: I'd expect expectedStartupPeriodSeconds * expectedStartupFailureThreshold to be > 129.

pkg/prometheus/server/statefulset_test.go
```diff
@@ -2906,3 +2906,63 @@ func TestPersistentVolumeClaimRetentionPolicy(t *testing.T) {
...
+		},
+		{
+			maximumStartupDurationSeconds:   ptr.To(int32(322)),
```
simonpasquier, line 2947: same here

pkg/prometheus/server/statefulset_test.go
nicolastakashi: Actually it will not be, because go round down integer divisions
129 / 60 (converting to second) = 2 (instead of 2.15 because of truncation division)

simonpasquier: yes I understand and I'd prefer if the value was rounded up to the next integer: as a user if I specify "maximumStartupDurationSeconds: 170", I don't expect the probe to fail after only 120s.

pkg/prometheus/statefulset.go
```diff
@@ -496,3 +496,15 @@ func ShareProcessNamespace(p monitoringv1.PrometheusInterface) *bool {
 		) == monitoringv1.ProcessSignalReloadStrategyType,
 	)
 }
+
+func GetStatupProbePeriodSecondsAndFailureThreshold(cfp monitoringv1.CommonPrometheusFields) (int32, int32) {
+	var startupPeriodSeconds int32 = 15
+	var startupFailureThreshold int32 = 60
+
+	if ptr.Deref(cfp.MaximumStartupDurationSeconds, 0) > 0 {
+		startupPeriodSeconds = *cfp.MaximumStartupDurationSeconds / 60
```
simonpasquier, line 505: Maybe we could have a rough heuristic to determine the "correct" period (e.g. we don't want to check 1) too few if the duration is small or 2) too often if the duration is large)?
* If max duration <= 60s then period = 10s
* If max duration > 60s then period = 20s

Or we just keep the default 15s and avoid the complexity :smile:

pkg/prometheus/statefulset.go
```diff
@@ -496,3 +496,15 @@ func ShareProcessNamespace(p monitoringv1.PrometheusInterface) *bool {
...
+		startupFailureThreshold = *cfp.MaximumStartupDurationSeconds / startupPeriodSeconds
```
simonpasquier, line 506:
```suggestion
		startupFailureThreshold = int(math.Ceil(float64(*cfp.MaximumStartupDurationSeconds)/startupPeriodSeconds)) * startupPeriodSeconds
```

pkg/prometheus/server/statefulset_test.go
simonpasquier: It should convert to the default period seconds.

pkg/prometheus/statefulset.go
nicolastakashi: I'd say if the max duration < 60s we should use default other wise we're fine to use the max duration defined by the user.

pkg/prometheus/statefulset.go
```diff
@@ -496,3 +497,17 @@ func ShareProcessNamespace(p monitoringv1.PrometheusInterface) *bool {
...
 	)
 }
+
+func GetStatupProbePeriodSecondsAndFailureThreshold(cfp monitoringv1.CommonPrometheusFields) (int32, int32) {
+	var startupPeriodSeconds float64 = 15
+	var startupFailureThreshold float64 = 60
+
+	maximumStartupDurationSeconds := float64(ptr.Deref(cfp.MaximumStartupDurationSeconds, 0))
+
+	if maximumStartupDurationSeconds > 0 && maximumStartupDurationSeconds >= 60 {
```
simonpasquier, line 507:
```suggestion
	if maximumStartupDurationSeconds >= 60 {
```

pkg/prometheus/agent/statefulset_test.go
```diff
@@ -186,6 +187,75 @@ func TestWALCompression(t *testing.T) {
 	}
 }
 
+func TestStartupProbeTimeoutSeconds(t *testing.T) {
```
simonpasquier, line 190: (nit) I would move the exhaustive tests to the pkg/prometheus package and have only basic tests here to avoid duplication (e.g. check results with a nil `maximumStartupDurationSeconds` and with a positive `maximumStartupDurationSeconds`).

pkg/apis/monitoring/v1/prometheus_types.go
simonpasquier, line 628:
```suggestion
	// +kubebuilder:validation:Minimum=60
	MaximumStartupDurationSeconds *int32 `json:"maximumStartupDurationSeconds,omitempty"`
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -621,6 +621,11 @@ type CommonPrometheusFields struct {
 	// If not specified, the configuration is reloaded using the /-/reload HTTP endpoint.
 	// +optional
 	ReloadStrategy *ReloadStrategyType `json:"reloadStrategy,omitempty"`
+
+	// Defines the maximum time that the `prometheus` container's startup probe will wait before being considered failed.
```
simonpasquier, line 625: (nit)
```suggestion
	// Defines the maximum time that the `prometheus` container's startup probe will wait before being considered failed. The startup probe will return success after the WAL replay is complete. 
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -621,6 +621,11 @@ type CommonPrometheusFields struct {
...
+	// If unset, it will be equal to 600 seconds (= 15 minutes).
```
simonpasquier, line 626:
```suggestion
	// If set, the value should be greater than 60 (seconds). Otherwise it will be equal to 600 seconds (15 minutes).
```

pkg/prometheus/agent/statefulset_test.go
simonpasquier: Yes to cover all the edge cases. Then in pkg/prometheus/server and pkg/prometheus/agent I would only test with 1) a nil `maximumStartupDurationSeconds` and 2) a non-nil `maximumStartupDurationSeconds` to verify that the values are properly translated.

pkg/prometheus/statefulset_test.go
```diff
@@ -0,0 +1,72 @@
...
+			expectedStartupFailureThreshold: 60,
+		},
+		{
+			maximumStartupDurationSeconds:   ptr.To(int32(1)),
+			expectedStartupPeriodSeconds:    15,
+			expectedStartupFailureThreshold: 60,
+		},
+		{
+			maximumStartupDurationSeconds:   ptr.To(int32(60)),
+			expectedStartupPeriodSeconds:    1,
```
simonpasquier, line 34: one potential problem I see here is that the period value might be too low (e.g. the probe needs to return a valid response under 1s).

nicolastakashi:
@simonpasquier any idea why this linter is failing?

simonpasquier:
LGTM (minus the linter issue).

pkg/prometheus/statefulset_test.go
```diff
@@ -0,0 +1,97 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package prometheus
+
+import (
+	"testing"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
```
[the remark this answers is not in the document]
nicolastakashi: Oh!
:(
`make fix-golang`

simonpasquier (approved):
LGTM

nicolastakashi:
@simonpasquier just rebased, I think we're fine to go wdyt?

wojciechka:
Just wondering - are there any plans to release a version of prometheus-operator with this change included?

Hitting an issue with very long startup and wondering if I should wait for a release including `maximumStartupDurationSeconds` or look for alternate approaches for now.

simonpasquier:
The next release is planned for Jan 10th (see https://github.com/prometheus-operator/prometheus-operator/blob/main/RELEASE.md).

simonpasquier:
In the meantime, check https://prometheus-operator.dev/docs/operator/strategic-merge-patch/#merging-patch-for-prometheus

wojciechka:
@simonpasquier Thank you very much for the information and link.

Went with the patching of `startupProbe` and may switch the approach to `maximumStartupDurationSeconds` in the future.

PR #6148, opened 2023-12-07
documentation: mention additionalArgs

jan--f:
## Description

For lists strategic merge patch can have surprising results where a list is replaced instead of added to. The operator offers workarounds for that, which should be explicitly mentioned.

## Type of change

### DIFF
--- a/Documentation/user-guides/strategic-merge-patch.md
+++ b/Documentation/user-guides/strategic-merge-patch.md
@@ -105,7 +105,7 @@ spec:
 The following manifest injects an additional CLI argument in the default
 Prometheus argument list. Note the use of `.spec.additionalArgs` in this
 example.
-Using `.spec.containers[*].args` directly would instead overwrite the containers
+Using `.spec.containers[*].args` directly would instead overwrite the container's
 `args` list completely, including the default arguments.
 
 ```yaml
--- a/Documentation/user-guides/strategic-merge-patch.md
+++ b/Documentation/user-guides/strategic-merge-patch.md
@@ -100,7 +100,7 @@ spec:
     - "3600"
 ```
 
-## How to inject additional CLI arguments into the prometheus pods
+## How to inject additional CLI arguments into the prometheus container
 
 The following manifest injects an additional CLI argument in the default
 Prometheus argument list. Note the use of `.spec.additionalArgs` in this
--- a/Documentation/user-guides/strategic-merge-patch.md
+++ b/Documentation/user-guides/strategic-merge-patch.md
@@ -99,3 +99,22 @@ spec:
     - sleep
     - "3600"
 ```
+
+## How to inject additional CLI arguments into the prometheus pods
+
+The following manifest injects an additional CLI argument in the default
+Prometheus argument list. Note the use of `.spec.additionalArgs` in this
+example.
+Using `.spec.containers[*].args` directly would instead overwrite the containers
+`args` list completely, including the default arguments.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: additional-arguments
+spec:
+  additionalArgs:
+  - name: "scrape.timestamp-tolerance"
+    value: "15ms"
+```


jan--f:
Not sure if an experimental flag is acceptable for a doc section. It was tricky to find something more benign that isn;t covered by a CRD field yet.

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
     - sleep
     - "3600"
 ```
+
+## How to inject additional cli arguments into the prometheus pods
+
+The following manifest injects an additional cli argument in the default
+Prometheus argument list. Not the use of `additionalArgs` in this example.
+Using `args` directly would overwrite the args list completely, including the
```
machine424, line 107:
```suggestion
Using `.spec.containers[*].args` instead would overwrite the container's args list completely, including the
```

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
     - sleep
     - "3600"
 ```
+
+## How to inject additional cli arguments into the prometheus pods
+
+The following manifest injects an additional cli argument in the default
+Prometheus argument list. Not the use of `additionalArgs` in this example.
```
machine424, line 106:
```suggestion
Prometheus argument list. Note the use of `.spec.additionalArgs` in this example.
```

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
...
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: additional-arguments
+spec:
+  containers:
+  - name: prometheus
+    additionalArgs:
```
machine424, line 118: I thought it was directly under `spec`?

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
...
+    - '--scrape.timestamp-tolerance=15ms'
```
machine424, line 119: maybe just `--arg=value`?

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
     - sleep
     - "3600"
 ```
+
+## How to inject additional cli arguments into the prometheus pods
```
simonpasquier, line 103: (nit)
```suggestion
## How to inject additional CLI arguments into the prometheus container
```

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
...
+
+The following manifest injects an additional cli argument in the default
```
simonpasquier, line 105:
```suggestion
The following manifest injects an additional CLI argument in the default
```

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,23 @@ spec:
...
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: additional-arguments
+spec:
+  containers:
+  - name: prometheus
+    additionalArgs:
+  - "--scrape.timestamp-tolerance=15ms"
```
simonpasquier, line 120:
```suggestion
  additionalArgs:
  - name: "scrape.timestamp-tolerance"
  - value: "15ms"
```

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
     - sleep
     - "3600"
 ```
+
+## How to inject additional CLI arguments into the prometheus pods
```
ArthurSens, line 103:
```suggestion
## How to inject additional CLI arguments into the prometheus container
```

ArthurSens (commented):
Besides that LGTM!

Documentation/user-guides/strategic-merge-patch.md
```diff
@@ -99,3 +99,22 @@ spec:
     - sleep
     - "3600"
 ```
+
+## How to inject additional CLI arguments into the prometheus container
+
+The following manifest injects an additional CLI argument in the default
+Prometheus argument list. Note the use of `.spec.additionalArgs` in this
+example.
+Using `.spec.containers[*].args` directly would instead overwrite the containers
```
simonpasquier, line 108:
```suggestion
Using `.spec.containers[*].args` directly would instead overwrite the container's
```

simonpasquier (approved):

ArthurSens:
Automerge won't work because it's only touching .md file 😬

nicolastakashi:
> Sorry for deviating from the subject of this PR; it reminded me of a subject we [discussed recently during our office hours](https://docs.google.com/document/d/1-fjJmzrwRpKmSPHtXN5u6VZnn39M28KqyQGBEJsqUOk/edit#heading=h.w6vs0qd5jbgf).
> 
> We wanted to improve documentation for the strategic merge patch, while also working on updating the behavior here to not completely override cmdArgs or environment variables.
> 
> @simonpasquier @nicolastakashi, seeing that we have a field for additional cmd args, what do you think about just adding a field of additional environment variables and leaving the merge patch as it is? I'm also concerned that changing the behavior of the merge patch would require Prometheus API 2.0

Sounds reasonable to me TBH

simonpasquier:
> seeing that we have a field for additional cmd args, what do you think about just adding a field of additional environment variables and leaving the merge patch as it is? I'm also concerned that changing the behavior of the merge patch would require Prometheus API 2.0

I wouldn't necessarily consider this to be a breaking change if the CRD struct doesn't change but yes it might be disturbing and trigger the opposite request (e.g. "I want to overwrite all env variables and it's not possible anymore").

In principle, I'm ok for a `.spec.additionalEnv` field but the API is going to be more complicated than for args because you can get the envvar value from different ways (e.g. static, field ref, secret ref, ...).

simonpasquier:
In fact I tested and environment variables are "properly" merged in the resulting statefulset. This is because the `Env` field in the `v1.Container` struct has the right `patchMergeKey` and `patchStrategy` annotations. I believe that it will make a good example for the docs.

ArthurSens:
> In fact I tested and environment variables are "properly" merged in the resulting statefulset. This is because the `Env` field in the `v1.Container` struct has the right `patchMergeKey` and `patchStrategy` annotations. I believe that it will make a good example for the docs.

Oh, that's news to me 😮 

Does it means that the $SHARD env is not overwritten when doing a strategic merge patch?

simonpasquier:
Indeed

PR #6151, opened 2023-12-09
update BUILD_BRANCH for consistency with other Prometheus projects

jetsaredim:
## Description
Most of the other Prometheus project repos use the `git rev-parse --abbrev-ref HEAD` for the branch info for non-github builds.  The `git branch --show-current` option is only available on newer versions of git which aren't available in a ton of build environments.

## Type of change

## Changelog entry

Update BUILD_BRANCH for non-github environments to a more sensible and consistent git command data source.

```
- Update BUILD_BRANCH for non-github environments to a more sensible and consistent git command data source
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ BUILD_DATE=$(shell date +"%Y%m%d-%T")
 # source: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
 ifndef GITHUB_ACTIONS
 	BUILD_USER?=$(USER)
-	BUILD_BRANCH?=$(shell git branch --show-current)
+	BUILD_BRANCH?=$(shell git rev-parse --abbrev-ref HEAD)
 	BUILD_REVISION?=$(shell git rev-parse --short HEAD)
 else
 	BUILD_USER=Action-Run-ID-$(GITHUB_RUN_ID)


simonpasquier:
It looks like `--show-current` has been added in Git 1.22, more than 3 years ago:
https://github.blog/2019-06-07-highlights-from-git-2-22/#tidbits

I'm curious to hear more details about which build systems you're talking about.

jetsaredim:
> It looks like `--show-current` has been added in Git 1.22, more than 3 years ago: https://github.blog/2019-06-07-highlights-from-git-2-22/#tidbits
> 
> I'm curious to hear more details about which build systems you're talking about.

There are still some legacy/enterprise build environments which are based on RHEL7.  Even the latest version of that OS had it's package major versions frozen for enterprise compatibility many years ago.  As such, the newest version of git packaged for it is 1.8.3 which doesn't include this option.  Also, as I mentioned if you look across other kubernetes/prometheus projects which use don't use promu they tend toward using the option I suggested.  I guess either way it doesn't make a huge difference since its easy enough to override due to the ?=.  Just figured it might other people's lives easier like it did mine.

Makefile
```diff
@@ -59,7 +59,7 @@ BUILD_DATE=$(shell date +"%Y%m%d-%T")
 # source: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
 ifndef GITHUB_ACTIONS
 	BUILD_USER?=$(USER)
-	BUILD_BRANCH?=$(shell git branch --show-current)
+	BUILD_BRANCH?=$(shell git rev-parse --abbrev-ref HEAD)
```
simonpasquier, line 62: Thanks for the details! IIUC this should work and fallback to your suggestion in case Git doesn't support the newer option:
```suggestion
	BUILD_BRANCH?=$(shell git branch --show-current || git rev-parse --abbrev-ref HEAD)
```

jetsaredim: I don't really have a good way to test this change in my environment.  I'll see if I can come up with something but if you have tested it yourself then I don't see why this couldn't work.  My only concern was the "else" side of the `||` might need a `shell` too but I'm not sure.

simonpasquier: I tested and it worked in my env.

PR #6153, opened 2023-12-10
Spec JobName for ScrapeConfig CRD

### LINKED ISSUE #6147: Add JobName for ScrapeConfig CRD
**What happened?**
As far as I can see the CRD of ScrapeConfig is not showing the full support for all scrape configs.
By example: the authorization data source and the job name.

This is a scrape config how I can set as it is:
```
- job_name: kubernetes-nodes-cadvisor
  honor_timestamps: true
  scrape_interval: 10s
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: https
  authorization:
    type: Bearer
    credentials_file: /var/run/secrets/kubernetes.io/serviceaccount/token
  tls_config:
    ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
    insecure_skip_verify: true
  follow_redirects: true
  enable_http2: true
```

Looking at the authorization, the crd is only allowing a credentials reading from a secret. Check the current 0.70.0 release:
```
              authorization:
                description: Authorization header to use on every scrape request.
                properties:
                  credentials:
                    description: Selects a key of a Secret in the namespace that contains
                      the credentials for authentication.
                    properties:
                      key:
                        description: The key of the secret to select from.  Must be
                          a valid secret key.
                        type: string
                      name:
                        description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
                          TODO: Add other useful fields. apiVersion, kind, uid?'
                        type: string
                      optional:
                        description: Specify whether the Secret or its key must be
                          defined
                        type: boolean
                    required:
                    - key
                    type: object
                    x-kubernetes-map-type: atomic
                  type:
                    description: "Defines the authentication type. The value is case-insensitive.
                      \n \"Basic\" is not a supported value. \n Default: \"Bearer\""
                    type: string
                type: object
```

So a manifest like this is not valid:
```
kind: ScrapeConfig
metadata:
  name: kubernetes-nodes-cadvisor
  namespace: default
spec:
  honorTimestamps: true
  authorization:
    type: Bearer
    credentialsFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```

Additionally, there is no option to specify a name for the job / the scrape_job.
This generates the label-value to `scrapeConfig/<namespace>/<metadata.name>` which some applications cannot use (expecting only `<metadata.name>`).

**Environment**

kube-prom-stack: helm 55.0.0 → prometheus 0.70.0
prometheus-operator-crds: helm 8.0.1 → prometheus 0.70.

    `kubectl version`
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.4", GitCommit:"fa3d7990104d7c1f16943a67f11b154b71f6a132", GitTreeState:"clean", BuildDate:"2023-07-19T12:20:54Z", GoVersion:"go1.20.6", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.4+k3s1", GitCommit:"36645e7311e9bdbbf2adb79ecd8bd68556bc86f6", GitTreeState:"clean", BuildDate:"2023-07-28T09:46:05Z", GoVersion:"go1.20.6", Compiler:"gc", Platform:"linux/arm64"}

* Kubernetes cluster kind:

    insert how you created your cluster: Rancher & k3s

I figured the relabelling in the scrapeConfig to a specific value for `scrape_job` isn't working. If you have a different solution this would easy the issue for the moment ;)

```
    - sourceLabels: ['scrape_job']
      regex: 'scrapeConfig\/namespace\/(.*)'
      targetLabel: 'scrape_job'
      replacement: '$1'
      action: 'replace'
```

→ goal is to have the last part of the regex to be the new value. This would be equivalent to the `metadata.name` value which is my desire.

Thanks for the help guys!

janluak:
prom's scrapeConfig supports providing a specific job-name; scrapeConfigs from CRD no longer bound to `scrapeConfig/<namespace>/<metadata.name>`

Fixes #6147

## Description

Scrape configs from prometheus are supporting a lot more config options compared to the crds aiming to provide the same scrape-config.
With this MR the spec <JobName> is added to the scrape-config-crd. This change makes it possible to specify a name for the scrape-job; otherwise the name defaults to `scrapeConfig/<namespace>/<metadata.name>`

It fixes #6147.

## Type of change

## Changelog entry

```release-note
feat: support `JobName` spec in ScrapeConfig CRD -> allowing to specify a name for the scrape job instead of default
```


simonpasquier (commented):
We still need to prevent duplicate job names in the final configuration file otherwise Prometheus will fail to load it.
I wonder if the `jobName` value shouldn't be translated into `metricRelabelConfigs`?

janluak:
@simonpasquier is there already existing code for the other crds checking the singularity? Then I'd reuse it...


I tried to write a scrape config with explicitly trying to relabel but I didn't get the scrape_job value to change to the desired name...
If it's the more desired way do you have a relabel config that works? Then I implement it instead...

pkg/prometheus/promcfg.go
```diff
@@ -2278,7 +2278,13 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 	sc *monitoringv1alpha1.ScrapeConfig,
 	store *assets.Store,
 ) (yaml.MapSlice, error) {
-	jobName := fmt.Sprintf("scrapeConfig/%s/%s", sc.Namespace, sc.Name)
+	isJobNameSet := sc.Spec.JobName != ""
+	jobName := sc.Spec.JobName
+
+	if !isJobNameSet {
+		jobName = fmt.Sprintf("scrapeConfig/%s/%s", sc.Namespace, sc.Name)
+	}
```
ArthurSens, line 2286:
```suggestion
	jobName := fmt.Sprintf("scrapeConfig/%s/%s", sc.Namespace, sc.Name)
	if sc.Spec.JobName != nil {
		jobName = sc.Spec.JobName
	}
```

We can save one memory allocation while improving readability like this :)

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -96,6 +96,8 @@ func (l *ScrapeConfigList) DeepCopyObject() runtime.Object {
 // ScrapeConfigSpec is a specification of the desired configuration for a scrape configuration.
 // +k8s:openapi-gen=true
 type ScrapeConfigSpec struct {
+	// The job name assigned to scraped metrics by default.
+	JobName string `json:"jobName,omitempty"`
```
ArthurSens, line 100:
```suggestion
	//+optional
	JobName *string `json:"jobName,omitempty"`
```

We want to follow K8s API conventions for optional fields
(adding optional markers and using pointers)

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

ArthurSens:
> @simonpasquier is there already existing code for the other crds checking the singularity? Then I'd reuse it...

Since k8s doesn't allow the same resource kind, with the same name in the same namespace, we never needed to check that. The string formed is always unique; we never needed to check up until now 🤔 .

simonpasquier (commented):
Regarding the implementation, I wouldn't change the value of `job_name` but rather add a relabeling config to inject the `job` label when `.spec.jobName` is set. This way we wouldn't need to check for `job_name` uniqueness.

janluak:
> Regarding the implementation, I wouldn't change the value of `job_name` but rather add a relabeling config to inject the `job` label when `.spec.jobName` is set. This way we wouldn't need to check for `job_name` uniqueness.

sorry for the late reply, got caught up in my project...

Do you have a working relabeling config?
Prior to this MR I tried quite some variants like the one below but after applying it the new label never showed in prometheus. That's why I figured it cannot be overwritten and the `jobName` had to change.

```yaml
  relabelings:
    - targetLabels: ["scrapeJob"]
      replacement: "myName"
      action: replace
```

Do you have a working relabel config? Then I am happy to implement it :)

simonpasquier:
In Prometheus config format, it would be:

```
  relabelings:
    - target_label: "job"
      replacement: "myName"
      action: replace
```

pkg/prometheus/promcfg.go
janluak: thanks! I reverted the changes due to using the relabel_config now

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
janluak: added, thanks

janluak:
So I played around again with the relabel config and implemented it now with the relabel config.

Still, for some reason I cannot get my head around this issue:
The relabeling worked for two jobs but not for the third one.
image
image
image
image

First image shows the grafana selection available.
The other three the configs from prometheus config. I would think they are working the same but for some reason the third one is not visible when using the relabel config. When I tested formerly with the fixed setting the name it worked...

Any ideas?

pkg/prometheus/promcfg.go
```diff
@@ -2290,6 +2291,15 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 	relabelings := initRelabelings()
 	labeler := namespacelabeler.New(cpf.EnforcedNamespaceLabel, cpf.ExcludedFromEnforcement, false)
 
+	if sc.Spec.JobName != "" {
+		relabelings = append(relabelings, yaml.MapSlice{
+			{Key: "target_label", Value: "scrapeJob"},
```
simonpasquier, line 2296:
```suggestion
			{Key: "target_label", Value: "job"},
```

pkg/prometheus/promcfg.go
```diff
@@ -2290,6 +2291,15 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+			{Key: "action", Value: "replace"},
+			{Key: "replacement", Value: sc.Spec.JobName},
+		})
+		cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
```
simonpasquier, line 2300:
```suggestion
```

pkg/prometheus/promcfg_test.go
```diff
@@ -4959,6 +4959,13 @@ func TestScrapeConfigSpecConfig(t *testing.T) {
 			scSpec: monitoringv1alpha1.ScrapeConfigSpec{},
 			golden: "ScrapeConfigSpecConfig_Empty.golden",
 		},
+		{
+			name: "explicit_job_name",
```
simonpasquier, line 4963: we need another test with a ScrapeConfig spec that defines a non-empty `relabelings` list.

pkg/prometheus/promcfg.go
janluak: Could you explain why? I just tried it and now the selector from linkerd-viz can't find the metrics → they are expected at {scrapeJob="linkerd-*"} which isn't fulfilled anymore after this change...

pkg/prometheus/promcfg.go
janluak: If I remove this line (just copied from the `if sc.Spec.RelabelConfigs != nil {` statement) the relabel_config isn't applied anymore. Is there a different trick I missed?

pkg/prometheus/promcfg_test.go
janluak: does commit #ae5a4ad8 suit you? It's not the most elegant solution but works ...

pkg/prometheus/promcfg.go
simonpasquier: The `job_name` field from the Prometheus configuration translates to the `job` label. I'm not sure why linkerd-viz would use `scrapeJob` instead.

pkg/prometheus/promcfg_test.go
simonpasquier: yes, perfect!

pkg/prometheus/promcfg.go
janluak: hm, what about the label `scrapeJob=scrapeConfig/<namespace>/<metadata.name>`?

If config `job_name` translates to label `job`, would we need another relabel_config here?

simonpasquier: Again I'm not sure where this `scrapeJob` label comes from but it's nothing conventional. If you really need, you can still inject it with another relabel config (e.g. copy from the `job` label).

janluak: oh, sorry, I thought it was the default for `scrapeJob`.

Well, if not I will have to stick with the relabelling on linkerd-viz. I guess I will debug on their side as well why they use it.

I fixed it with #7a2b39b5

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
simonpasquier: @janluak can you apply Arthur's comment? Thanks!

pkg/prometheus/promcfg.go
```diff
@@ -2320,6 +2329,9 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 
 	if sc.Spec.RelabelConfigs != nil {
 		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
+	}
+
+	if sc.Spec.JobName != "" || sc.Spec.RelabelConfigs != nil {
```
simonpasquier, line 2334:
```suggestion
	if len(relabelings) > 0 {
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
janluak: Sorry, I am a bit confused: the code already contains the `//+optional` comment:
image
What did I miss? (see code also [here](https://github.com/janluak/prometheus-operator/blob/feature/crd-scrapeConfig-with-jobName/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go))

I committed the change myself tough to have the commit message correct: I didn't know how to stay in the structure if using the GitHub-UI to do the commit...

pkg/prometheus/promcfg.go
janluak: I created two branches:
* feature/crd-scrapeConfig-with-jobName-relabelling-0
* feature/crd-scrapeConfig-with-jobName-relabelling-1

The digit in the end is the same as the comparison used in line 2334 for.

The issue: the suggested refactoring is not equal. 
The variable relabelings is already of `length=1` when initialized.
→ Setting the relabeling to `>0` leaves to a ton of changes of other files not relevant to this feature.
→ Setting the relabeling to `>1` requires only one change to the feature-specific test-file with empty relabel config. 

I'd think the second option is better continuing the current behavior.
Yet, I am familiar enough with the overall behavior to evaluate if the overall change with option one is more desirable...

Just let me know which way you prefer and I merge the one into this merge request branch.
And Merry Christmas ;)

simonpasquier: I see your concern but I think that the current implementation is prone to error. I've submitted #6201 to "fix" it.

ArthurSens (approved):
Nice solution that you two found!

My initial idea was some to iterate over all selected *Monitor resources and make sure we don't have a matching job name. The relabelings strategy sounds much simpler :)

I think this is good to merge once #6201 gets in, and we add the missing pointer here

simonpasquier:
@janluak you'd need to rebase on top of main to resolve the conflict + address @ArthurSens comments.

pkg/prometheus/promcfg.go
janluak: Thanks for your fix. With your refactoring the question of >0 or >1 is obsolete...

janluak:
> Nice solution that you two found!
> 
> My initial idea was to iterate over all selected *Monitor resources and make sure we don't have a matching job name. The relabelings strategy sounds much simpler :)
> 
> I think this is good to merge once #6201 gets in, and we add the missing pointer here

Thanks for the feedback :)

fixed with #ef52ca90

janluak:
> @janluak you'd need to rebase on top of main to resolve the conflict + address @ArthurSens comments.

@simonpasquier thanks for your fix :)

I rebased and implemented the type change as recommended. I hope I didn't screw up the rebase since the label got changed...

ArthurSens:
Looks like the rebase was done incorrectly 😅, we don't want to merge 37 unrelated commits

ArthurSens:
Hey @janluak , have you noticed the commits that were added by accident here? Do you need some help with the rebase?

janluak:
Hi Arthur,Sorry, I was quite busy the last days…Yet, I don’t really have an answer how I revert a rebase. I was thinking of renaming the branch and creating a new one with the same name and then redo the commits but that seems to be quite nasty… other option would be to close this MR in favor of a new one.Do you know the trick how to revert the rebase and then do it correctly?Thanks, JanAm 19.01.2024 um 04:29 schrieb Arthur Silva Sens ***@***.***>:﻿
Hey @janluak , have you noticed the commits that were added by accident here? Do you need some help with the rebase?

ArthurSens:
What I usually do is use the git rebase command in interactive mode: `git rebase -i HEAD~<number of commits I want to rebase>`

I believe in the case of this PR you'd work with 37 commits, and you would drop all commits that aren't related to this PR

---

Opening another pull request with only the changes you want also works! Whatever is best for you :)

mviswanathsai:
Thanks for the effort @janluak! Do you have the time to fix the rebase?

Issue #6159, opened 2023-12-12
The error was:   in "<unicode string>", line 5668, column 29


Sispheor:
When deploying via Ansible using the module K8S we get an error:
```
The error was:   in "<unicode string>", line 5668, column 29
```

This is due to the missing quote in some defintion:
```yaml
enum:
  - '!='
  - =
  - =~
  - '!~'
```

The fix consist into adding quote
```yaml
enum:
  - '!='
  - '='
  - '=~'
  - '!~'
```

simonpasquier:
Thanks for the report! Unfortunately there's nothing we can do about it. The generated YAML is valid wrt to the YAML spec and the issue is with the pyaml library (see https://github.com/yaml/pyyaml/issues/89). More details in https://github.com/prometheus-operator/prometheus-operator/pull/4897.

Sispheor:
Thanks. I close this one then.

PR #6160, opened 2023-12-12
Fix unicode error due to missing quote in enum declaration

Sispheor:
## Description

Buf described here #6159

## Type of change

## Changelog entry

```release-note
- Fix unicode string error when deploying from ansible K8S module
```

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -69,8 +69,8 @@ spec:
                               if non-empty.
                             enum:
                             - '!='
-                            - =
-                            - =~
+                            - '='
+                            - '=~'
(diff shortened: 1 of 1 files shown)


ArthurSens:
Hey @Sispheor, thanks for the PR and interest in contributing a fix! However, this is not something that can be fixed here. We use tools for CRDs/Bundle generation (that's why CI is failing) and the output is a valid yaml file.

As mentioned in the related issue, please get in touch with the Python YAML parsing library

Sispheor:
Ok thanks!

PR #6178, opened 2023-12-19
feat: add support for remaining fields in Kubernetes SD

### LINKED ISSUE #6087: Support other fields available in KubernetesSDConfigs
**What is missing?**

We didn't cover all fields in initial PR for KubernetesSDConfigs https://github.com/prometheus-operator/prometheus-operator/pull/5675, this request is to support remaining  fields in https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config 

**Why do we need it?**

To complete the feature

**Environment**

slashpai:
Fixes #6087

## Description

## Type of change

## Changelog entry

```release-note
feat: add support for remaining fields in Kubernetes SD in ScrapeConfig
```


slashpai:
@simonpasquier This should be good for review :)

pkg/prometheus/promcfg.go
```diff
@@ -2499,11 +2499,79 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+			configs[i] = append(configs[i], yaml.MapItem{
+				Key:   "role",
+				Value: strings.ToLower(string(config.Role)),
+			})
+
+			configs[i] = cg.addBasicAuthToYaml(configs[i], assetStoreKey, store, config.BasicAuth)
+			configs[i] = cg.addSafeAuthorizationToYaml(configs[i], fmt.Sprintf("scrapeconfig/auth/%s/%s/kubernetessdconfig/%d", sc.GetNamespace(), sc.GetName(), i), store, config.Authorization)
+			configs[i] = cg.addOAuth2ToYaml(configs[i], config.Oauth2, store.OAuth2Assets, assetStoreKey)
+
+			if config.ProxyConfig != nil {
```
simonpasquier, line 2520: there's a lot of repetition with existing code. We need a utility function that translates a ProxyConfig field into Prometheus YAML config.

pkg/prometheus/promcfg.go
```diff
@@ -2499,11 +2499,79 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+					Value: config.EnableHttp2,
+				})
+			}
+
+			if config.TLSConfig != nil {
+				configs[i] = addSafeTLStoYaml(configs[i], sc.GetNamespace(), *config.TLSConfig)
+			}
+
+			if config.Namespaces != nil {
+				configs[i] = append(configs[i], yaml.MapItem{
```
simonpasquier, line 2571: I'd rather iterate over every field than assuming an identical serialization between the CRD and Prometheus configuration.

pkg/prometheus/promcfg.go
```diff
@@ -2530,6 +2598,17 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 					Value: selectors,
 				})
 			}
+
+			if config.AttachMetadata != nil {
+				switch config.Role {
```
simonpasquier, line 2603:
```suggestion
				switch strings.ToLower(config.Role) {
```

pkg/prometheus/promcfg.go
```diff
@@ -2530,6 +2598,17 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+				case "pod", "Pod":
+					configs[i] = cg.WithMinimumVersion("2.35.0").AppendMapItem(configs[i], "attach_metadata", config.AttachMetadata)
+				case "endpoints", "Endpoints", "endpointslice", "EndpointSlice":
+					configs[i] = cg.WithMinimumVersion("2.37.0").AppendMapItem(configs[i], "attach_metadata", config.AttachMetadata)
+				default:
+					level.Warn(cg.logger).Log("msg", fmt.Sprintf("ignoring attachMetadata not supported by Prometheus for role: %s", config.Role))
```
simonpasquier, line 2609: should we reject the resource in the first place?

slashpai: I think its ok to warn than rejecting the resource. We do similar for other version specific newer fields.

pkg/prometheus/promcfg.go
slashpai: sure, I can do that in follow up PR since ProxyConfig is used in a few SDConfigs as well

slashpai:
@simonpasquier addressed comments. could you please review again?

pkg/prometheus/promcfg.go
simonpasquier: I'm on the fence too. Prometheus wouldn't fail either if you configure node's metadata with a service role so a warning is probably fine.
For clarity in the CRD description, it would be better to duplicate the struct from pkg/api/monitoring/v1 and state in the v1alpha1 comment that it's only valid for Pod, Endpoint and Endpointslice roles.

pkg/prometheus/resource_selector.go
```diff
@@ -766,8 +766,38 @@ func (rs *ResourceSelector) SelectScrapeConfigs(ctx context.Context, listFn List
...
+			return fmt.Errorf("[%d]: %w", i, err)
+		}
+
+		if config.ProxyConfig != nil {
+			if err := validateProxyConfig(ctx, config.ProxyConfig, rs.store, sc.GetNamespace()); err != nil {
+				return fmt.Errorf("[%d]: %w", i, err)
+			}
+		}
+
+		if config.APIServer != nil && config.Namespaces != nil {
```
simonpasquier, line 795:
```suggestion
		if config.APIServer != nil && len(config.Namespaces) > 0 {
```

slashpai: config.Namespaces isn't array so we wouldn't be able to use `len()`

slashpai:
@simonpasquier addressed comments

pkg/prometheus/promcfg.go
```diff
@@ -2498,11 +2498,89 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+
+			if config.TLSConfig != nil {
+				configs[i] = addSafeTLStoYaml(configs[i], sc.GetNamespace(), *config.TLSConfig)
+			}
+
+			if config.Namespaces != nil {
+				namespaces := []yaml.MapItem{
+					{
+						Key:   "own_namespace",
+						Value: config.Namespaces.IncludeOwnNamespace,
```
simonpasquier, line 2573: we need to test IncludeOwnNamespace against nil

slashpai: I added two more tests for this

simonpasquier: I mean that when IncludeOwnNamespace it shouldn't appear in the generated config.

simonpasquier (approved):

PR #6194, opened 2023-12-28
feat: support the operator make use of the config-reloader tls and basic auth with prometheus/alertmanager webConfigFile

dongjiang1989:
…thentication

## Description

ref: #6068 #5932 

Support the operator make use of the config-reloader tls and basic authentication with prometheus/alertmanager webConfigFile

## Type of change

## Changelog entry

```release-note
support the operator make use of the config-reloader tls and basic auth with prometheus/alertmanager webConfigFile
```


dongjiang1989:
@simonpasquier  IIUC, reuse the Prometheus/Alertmanager web config for the reloader. PTAL

simonpasquier (commented):
Thanks! Got one question regarding the change in pkg/operator/defaults.go

pkg/operator/defaults.go
```diff
@@ -31,6 +31,8 @@ const (
 	DefaultThanosBaseImage = "quay.io/thanos/thanos"
 	// DefaultThanosImage is a default image pulling address for the Thanos long-term prometheus storage collector
 	DefaultThanosImage = DefaultThanosBaseImage + ":" + DefaultThanosVersion
+	// DefaultConfigReloaderVersion is a default image tag for the prometheus-operator config-reloader
+	DefaultConfigReloaderVersion = "v0.69.0"
```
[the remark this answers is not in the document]
dongjiang1989: Hmmm... Be defined but not used.
Removed it

pkg/prometheus/agent/statefulset.go
```diff
@@ -207,9 +220,11 @@ func makeStatefulSetSpec(
 		if err != nil {
 			return nil, err
 		}
+		configReloaderWebConfigFile = confArg.Value
 		promArgs = append(promArgs, confArg)
 		volumes = append(volumes, configVol...)
 		promVolumeMounts = append(promVolumeMounts, configMount...)
+		configReloaderVolumeMounts = append(configReloaderVolumeMounts, configMount...)
```
simonpasquier, line 227: We could make an argument that adding the flag in all cases introduces a risk of breaking users not deploying the latest config-reloader tag. It would be safer to add the flag only when the web field isn't nil and we can remove the if condition in a few releases.

```suggestion
		// To avoid breaking users deploying an old version of the config-reloader image.
		// TODO: remove the if condition after v0.72.0.
		if cpf.Web != nil {
			configReloaderWebConfigFile = confArg.Value
			configReloaderVolumeMounts = append(configReloaderVolumeMounts, configMount...)
		}
```

dongjiang1989: Thanks for your review. 

Got it. But `cpf.Web != nil` only determines whether `Prometheus's web` supports it, not whether the `config-reloader` supports it.

```
		if c.ReloaderConfig.Image.Version >= 0.69.0 {
		     // xxx
		}
```
Because config-reloader v0.69.0 already supports this feature.

dongjiang1989: OR Add flag for Prometheus/alertmanager CRD?

simonpasquier: My intention was only to avoid breaking users that don't specify web config for their Prometheus/Alertmanager. For others, they still need to make sure that they use a recent-enough config-reloader image.

pkg/prometheus/server/statefulset.go
```diff
@@ -285,6 +298,13 @@ func makeStatefulSetSpec(
 		promArgs = append(promArgs, confArg)
 		volumes = append(volumes, configVol...)
 		promVolumeMounts = append(promVolumeMounts, configMount...)
+
+		// To avoid breaking users deploying an old version of the config-reloader image.
+		// TODO: remove the if condition after v0.72.0.
+		if prompkg.ConfigReloaderWebConfigFileSupported(c) {
```
dongjiang1989, line 304: Add config-reloader image version check.

pkg/prometheus/agent/statefulset.go
dongjiang1989: @simonpasquier Add config-reloader image version check. PTAL

simonpasquier (commented):
I wouldn't try to deduce the binary version from the image tag: it won't work for folks that rebuild the images or use SHA1 refs instead of tags.
The safest approach from my POV is like I described earlier: ony pass the `--web-config-file` flag when the Alertmanager/Prometheus resource defines a non-nil `.spec.web` field.

dongjiang1989:
> I wouldn't try to deduce the binary version from the image tag: it won't work for folks that rebuild the images or use SHA1 refs instead of tags. The safest approach from my POV is like I described earlier: ony pass the `--web-config-file` flag when the Alertmanager/Prometheus resource defines a non-nil `.spec.web` field.

Thanks for your review. Got it. Fixed.

.github/workflows/checks.yaml
```diff
@@ -58,6 +58,7 @@ jobs:
         with:
           version: v1.55.2
           args: --timeout 10m0s
+          skip-pkg-cache: true
```
simonpasquier, line 61: it seems unrelated to the PR. Can you revert?

simonpasquier (commented):
LGTM otherwise

.github/workflows/checks.yaml
dongjiang1989: Thanks. Fixed it.

simonpasquier (approved):
Thanks!

PR #6197, opened 2023-12-31
Add UpdateScale and GetScale methods

ArthurSens:
## Description

Addresses comment https://github.com/prometheus-operator/prometheus-operator/pull/5962#discussion_r1430123922

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -828,7 +828,7 @@ func testPromCreateDeleteCluster(t *testing.T) {
 	}
 }
 
-func testPromScaleUpDownCluster(t *testing.T) {
+func testPromScaleUpDownReplicas(t *testing.T) {
 	t.Parallel()
 	testCtx := framework.NewTestCtx(t)
 	defer testCtx.Cleanup(t)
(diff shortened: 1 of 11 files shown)


simonpasquier:
The test fails :thinking: 
```
prometheus_test.go:5082: failed to scale Prometheus allns-y-prometheusstatusscale-s6je32-0-9b070934/test: the name of the object (test based on URL) was undeterminable: name must be provided
````

ArthurSens:
> The test fails 🤔
> 
> ```
> prometheus_test.go:5082: failed to scale Prometheus allns-y-prometheusstatusscale-s6je32-0-9b070934/test: the name of the object (test based on URL) was undeterminable: name must be provided
> ```

Interesting... the generated GetScale() method returns an error if .spec.shards is nil 🤔... not sure how to work around that

simonpasquier:
Hmm, I guess we could have a default value of 1 for `.spec.shards` but it might break some workflows (e.g. reconcilers doing a raw diff between the desired spec and the current spec may trip over the API returning a shard value of 1 instead of null). What happens if the e2e tests set the shard value explicitly to 1?
We could add the default value for the PrometheusAgent CRD though since it's v1alpha1.

ArthurSens:
> Hmm, I guess we could have a default value of 1 for `.spec.shards` but it might break some workflows (e.g. reconcilers doing a raw diff between the desired spec and the current spec may trip over the API returning a shard value of 1 instead of null). What happens if the e2e tests set the shard value explicitly to 1? 

They pass :)

> We could add the default value for the PrometheusAgent CRD though since it's v1alpha1.

That good, but since both CRDs use the same status/scale spec, we probably want to change both 😬

simonpasquier (approved):

test/e2e/prometheus_test.go
simonpasquier, line 1: (nit) should we rename ` testPromScaleUpDownCluster` to `testPrometheusScaleUpDownReplicas` for clarity?

simonpasquier (approved):

Issue #4967, opened 2022-08-12
Graceful scale down of shards


v1jayr:
**What is missing?**
With sharded prometheus, reducing the number of shards results in immediate removal of shards. There is no native mechanism to gracefully prevent loss of metrics held by those shards. Ideally, scale down is a 2 step process where a subset of shards are available for query without adding new metric data (without scraping) and later as part of separate update to operator they are fully removed.  

**Why do we need it?**
Scale down of shards should be just as convenient as scale up. Right now, scale down means either loosing data held by those shards. We should be able to scale down without risk or data loss.

**Environment**
Kubernetes

* Prometheus Operator version:
    `v0.53.1`

**Anything else we need to know?**:

slashpai:
Currently scaling down [shards](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/shards-and-replicas.md) will not reshard data onto remaining instances. It must be manually moved. 

I think we need to figure out the best approach if we need to implement resharding. We will discuss this in next office hours

simonpasquier:
The request makes sense but I'm not yet sure about how it would work in practice.

Keeping in-excess shards around looks good to me but when to remove the shards is more complicated IMHO. If time-based retention time is configured, the operator could determine that the shard can be removed after (retention time period). For storage-based retention, it should probably be a user decision. We need also to reflect this in the object's status somehow.

Happy to hear more ideas about the feature UX :)

nicolastakashi:
@slashpai do still have to manually reshard data when scaling up or down the number of shards?
If so, this can impact in this issue #4946

simonpasquier:
My take is that the scale down of shards is related but not limited to the autoscaling use case. It's also a problem if you scale up/down the Prometheus resource manually.

One idea could be that when scaling down the number of shards, the operator doesn't delete the statefulsets in excess but keep them around for "some time" (*). The operator could set a custom annotation (e.g. `prometheus-operator.dev/deletionTimestamp: 2023-09-22T09:51:33+00:00`) on the statefulset:  when the deadline has been reached, the operator would delete the statefulset shard. Users could force the deletion themselves by removing the annotation.

(*) it's "easy" to deduce a deletion deadline with time-based retention but not really possible for storage-based retention. There's also the case of Prometheus agent where ideally you want to remove the statefulset once all data has been remote written.

simonpasquier:
[...]

PR #6198, opened 2023-12-31
Add ShardRetentionPolicy to Prometheus API

ArthurSens:
## Description

Partially addresses #4967, by implementing half of what was proposed [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#graceful-scale-down-of-prometheus-servers).

When Scaling down Prometheus servers, there are good chances of data loss because the Operator is not aware if the data stored in a Prometheus server is being backed up by a Long-Term-Storage solution. Users who want to avoid such situations might want to tell the operator to keep the instance alive for a while before safe deletion.

This PR introduces the new field `ShardRetentionPolicy` to the Prometheus CRD to address the problem mentioned above and in the linked proposal.

For the sake of keeping PRs small, in this PR I'm only worrying about pod retention when a retention period is set. Follow-up PRs will be opened with the goals:
* Scale up will re-use already existing pod that was marked for future deletion
* Scale up/down will correctly redistribute targets.
* Retention duration is not taken into consideration right now. If ShardRetentionPolicy is set to `Retain`, we'll keep the shard alive forever.

## Type of change

## Changelog entry

```release-note

```
_(No release notes yet since I don't want to advertise a half implemented feature 😬 )_


ArthurSens:
@simonpasquier @nicolastakashi, I'm finally taking this up again :)

I'm trying things out and I think there are many things in the proposal that will require big chunks of code changes, e.g.:
* Not delete a statefulset if it's within the retention period
* Not create a new statefulset if there's another shard already available (just remove the deletion annotation)
* Update scrape configuration from active/inactive shards

I wonder if we want to work on a separate branch and merge to main only once the whole feature is ready 🤔

simonpasquier:
> I wonder if we want to work on a separate branch to make reviews easier and only merge to main only when the whole feature is ready 🤔

I'm not a fan of feature branches: they are usually a pain to work with for the contributor. One alternative would be to have it behind a feature flag. We don't have such mechanism now but I feel that it would be beneficial in the long run. Having said that, I'm not sure if feature flags would be practical for this change?

simonpasquier:
> The goal is to break this PR into smaller ones to make it easier to review, and not to release it experimentally.

It could be both ;)

We don't have a good way to signal to users that a particular feature in a v1 CRD is experimental. Up to now, the only indication would be "it's marked as experimental in the docs" but it isn't great. Maybe I'm wrong but I feel that this feature might need some cycles to be considered stable.
It's not the first time that we've pondered about having feature flags in the operator, maybe it's time to go ahead?

ArthurSens:
>  Maybe I'm wrong but I feel that this feature might need some cycles to be considered stable. It's not the first time that we've pondered about having feature flags in the operator, maybe it's time to go ahead?

Yeah, good point :) 

I'll open a PR for feature flags

simonpasquier:
Before jumping ahead of ourselves, I've added the discussion to the next office hours.

ArthurSens:
Alright @prometheus-operator/prometheus-operator-reviewers, PR ready for review :)

I also have a doubt, should we include a changelog entry even if it's behind feature-gate and only half implemented? 🤔

pkg/apis/monitoring/v1/types.go
```diff
@@ -41,6 +43,70 @@ type ByteSize string
...
+	"ms": {7, uint64(time.Millisecond)},
+	"s":  {6, uint64(time.Second)},
+	"m":  {5, uint64(time.Minute)},
+	"h":  {4, uint64(time.Hour)},
+	"d":  {3, uint64(24 * time.Hour)},
+	"w":  {2, uint64(7 * 24 * time.Hour)},
+	"y":  {1, uint64(365 * 24 * time.Hour)},
+}
+
+func (d Duration) AsTimeDuration() (time.Duration, error) {
```
simonpasquier, line 63: IIUC this is a copy from `github.com/prometheus/common/model`. Can't we call `model.ParseDuration()` in the controller?

I'm not a fan of additional functions to the api packages as it's easy to pull external dependencies by accident (https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#no-functions).

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -798,6 +798,17 @@ type PrometheusSpec struct {
 	// Maximum number of bytes used by the Prometheus data.
 	RetentionSize ByteSize `json:"retentionSize,omitempty"`
 
+	// ShardRetentionPolicy defines the retention policy for the Shard's statefulset.
```
simonpasquier, line 801:
```suggestion
	// ShardRetentionPolicy defines the retention policy for the Prometheus shards.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -798,6 +798,17 @@ type PrometheusSpec struct {
...
+	// ShardRetentionPolicy is an experimental feature, only enabled when the feature gate
+	// `ShardRetentionPolicies` is enabled.
```
simonpasquier, line 803: We can also take inspiration from https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages and categorize gates as alpha, beta and GA.

```suggestion
	// (Alpha) Using this field requires the ShardRetentionPolicies feature gate to be enabled.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -798,6 +798,17 @@ type PrometheusSpec struct {
...
+	//
+	// The final goals for this feature can be seen at https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#graceful-scale-down-of-prometheus-servers,
+	// however, the feature is not yet fully implemented. Currently, scale up after scale
```
simonpasquier, line 806: I would list all limitation like

````
Limitations:
* Scale up after ...
* Target redistribution ...
````

File: pkg/apis/monitoring/v1/prometheus_types.go [hunk 5]
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
 	TSDB TSDBSpec `json:"tsdb,omitempty"`
 }
 
+type WhenScaledRetentionType string
+
+var (
+	WhenScaledRetentionTypeRetain WhenScaledRetentionType = "retain"
```
simonpasquier, line 931:
```suggestion
	WhenScaledRetentionTypeRetain WhenScaledRetentionType = "Retain"
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
...
+	WhenScaledRetentionTypeDelete WhenScaledRetentionType = "delete"
```
simonpasquier, line 932:
```suggestion
	WhenScaledRetentionTypeDelete WhenScaledRetentionType = "Delete"
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
...
+	// WhenScaled defines the retention policy when the Prometheus Server is scaled-down.
+	// When set to retain, the statefulset will only stop scraping targets and will retain the data.
+	// The remaining shards' scrape configuration will be updated to scrape the targets of the scaled-down shard.
+	//
+	// For backwards compatibility, 'delete' is the default value. In a future v2 API, 'retain' will be the default value.
+	//
+	// Default: "delete"
+	// +kubebuilder:validation:Enum=retain;delete
+	// +required
+	WhenScaled string `json:"whenScaled"`
```
simonpasquier, line 945:
```suggestion
	WhenScaled WhenScaledRetentionType `json:"whenScaled"`
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
...
+	WhenScaledRetentionTypeRetain WhenScaledRetentionType = "retain"
+	WhenScaledRetentionTypeDelete WhenScaledRetentionType = "delete"
+)
+
+type ShardRetentionPolicy struct {
+	// WhenScaled defines the retention policy when the Prometheus Server is scaled-down.
+	// When set to retain, the statefulset will only stop scraping targets and will retain the data.
+	// The remaining shards' scrape configuration will be updated to scrape the targets of the scaled-down shard.
+	//
+	// For backwards compatibility, 'delete' is the default value. In a future v2 API, 'retain' will be the default value.
```
simonpasquier, line 940: I wouldn't make any promise/assumption regarding the v2 API.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
...
 
+type WhenScaledRetentionType string
+
+var (
+	WhenScaledRetentionTypeRetain WhenScaledRetentionType = "retain"
+	WhenScaledRetentionTypeDelete WhenScaledRetentionType = "delete"
+)
+
+type ShardRetentionPolicy struct {
+	// WhenScaled defines the retention policy when the Prometheus Server is scaled-down.
```
simonpasquier, line 936: Should it be named `WhenScaleDown`  if it only applies to scale down events?

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
...
+	// When set to retain, the statefulset will only stop scraping targets and will retain the data.
```
simonpasquier, line 937: "shard" rather than "statefulset" to avoid leaking implementation details?

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
...
+	// The remaining shards' scrape configuration will be updated to scrape the targets of the scaled-down shard.
```
simonpasquier, line 938:
```suggestion
	// The targets scraped by the scaled-down shard(s) are redistributed to the remaining shards.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,34 @@ type PrometheusSpec struct {
...
+	//
+	// For backwards compatibility, 'delete' is the default value. In a future v2 API, 'retain' will be the default value.
+	//
+	// Default: "delete"
+	// +kubebuilder:validation:Enum=retain;delete
+	// +required
+	WhenScaled string `json:"whenScaled"`
+
+	// Retention defines the retention period of the Prometheus Server.
```
simonpasquier, line 947: The "retention" term here is confusing with the actual Prometheus retention settings. Not sure if we can find something more explicit (`deleteAfter`, `keepFor`, `timeToLive`?).

To keep things simple, Maybe we can skip this field in this PR?

simonpasquier (commented):
My review was mostly focused on the API.

Question about the e2e tests: should we have a separate e2e job specific for feature gated functionality? If the existing e2e jobs don't enable any feature gates, we can assert that we don't have regression due to bad feature gating. We can also make the "feature gate" job not required in a first time (at least until we are confident about the stability).
It's also of interest for the DaemonSet feature (cc @haanhvu).

pkg/apis/monitoring/v1/types.go
ArthurSens: Actually I wasn't aware of this function in common/model 😮 helps a lot!

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: Good idea :)

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: good point!

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: > To keep things simple, Maybe we can skip this field in this PR?

To make it easier to review? Yeah, I don't see any problems with that :)

simonpasquier: Yes I was mostly concerned with the review load :)

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -798,6 +798,17 @@ type PrometheusSpec struct {
...
+	// ShardRetentionPolicy defines the retention policy for the Prometheus shards.
+	// (Alpha) Using this field requires the ShardRetentionPolicies feature gate to be enabled.
+	//
+	// The final goals for this feature can be seen at https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#graceful-scale-down-of-prometheus-servers,
+	// however, the feature is not yet fully implemented and has known limitations:
+	// * Scale up after scale down will not re-use the previously marked-for-deletion shards.
+	// * Target redistribution doesn't happen on scale up/down.
+	//
+	// +optional
+	ShardRetentionPolicy *ShardRetentionPolicy `json:"shardRetentionPolicy,omitempty"`
```
nicolastakashi, line 810: Wdyt about not exposing this fields at all on the CRD and assume default values when the feature is enabled.

This is mainly what I've suggested in one office hours, we can assume default values for the feature, since we're underd development and we want to do increamental PRs.

So we might think better about the API and when the feature is ready to be exposed we can introduce new fields on the CRs wdyt?

simonpasquier (commented):
I was working on the side to add a gauge metric reporting all feature gates and I reworked a bit the way the operator manages the gates. In particular I believe that the gates should be part of the global operator config, a (nice) side effect being that new gates are automatically available to the controllers. 

If you want to have a look at https://github.com/prometheus-operator/prometheus-operator/pull/6655. IMHO it would avoid more boilerplate code in this PR to introduce the retention policy feature gate in the Prometheus controller.

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: I'm a bit confused how we could implement both the retain and delete APIs with this strategy, since we plan to have multiple behaviors on retain 😬

ArthurSens:
I'm gonna work on a new PR implementing the e2e-tests for feature gated features.

I'll come back to this one afterwards :)

ArthurSens:
Ok, the PR is blocked until we get #6665 in, but I believe all other comments are addressed

ArthurSens:
This should be ready for another round of reviews :)

ArthurSens:
Anything I could do to help with the reviews? :)

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: Sorry!
I completely missed this comment.
What I mean is, at lest in the development phases do not expose this field in the API until we reach the desired state, once we have the feature working behind the scenes we can come back and decide how we want to expose this to the API.

I particularly see two advantages.
1. We reduce the changes of people try to use it while is under development.
2. We can post pone changes to the CRD API and think it better once we have the feature working as expected.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -991,6 +1003,24 @@ type PrometheusSpec struct {
...
+type WhenScaledDownRetentionType string
+
+var (
+	WhenScaledDownRetentionTypeRetain WhenScaledDownRetentionType = "Retain"
+	WhenScaledDownRetentionTypeDelete WhenScaledDownRetentionType = "Delete"
+)
+
+type ShardRetentionPolicy struct {
+	// WhenScaledDown defines the retention policy when the Prometheus Server is scaled-down.
+	// When set to retain, the shard will only stop scraping targets and will retain the data.
```
simonpasquier, line 1015: (suggestion)
```suggestion
	// When set to `Retain`, the operator will keep the pods from the scaled-down shard(s). The pods won't scrape targets anymore but they can still be queried.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -914,6 +925,32 @@ type PrometheusSpec struct {
...
+	// The targets scraped by the scaled-down shard(s) are redistributed to the remaining shards.
+	//
+	// Default: "Delete"
```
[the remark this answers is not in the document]
simonpasquier: If the field is required, there's no default.

pkg/prometheus/server/operator.go
```diff
@@ -972,6 +986,45 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	}
+	retentionAsTime, err := time.Parse(time.RFC3339, retentionAnnotation)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse retention annotation %q: %w", deletionTimestampAnnotation, err)
+	}
+	if c.now().After(retentionAsTime) {
+		return true, nil
+	}
+
+	// We should never get here, but just in case we do, we return false for safety.
```
simonpasquier, line 1014: you mean that because getRetentionDate() returns a very far away timestamp, we always return before? If yes, I'd remove the comment here to avoid a mismatch when we implement proper cleanup.

pkg/prometheus/server/operator.go
```diff
@@ -972,6 +986,45 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+
+	srp := p.Spec.ShardRetentionPolicy
+	if srp == nil || srp.WhenScaledDown == monitoringv1.WhenScaledDownRetentionTypeDelete {
+		return true, nil
+	}
+
+	retentionAnnotation, ok := sts.Annotations[deletionTimestampAnnotation]
+	if !ok {
+		// This statefulset was not marked for deletion yet. Let's just add the annotation and return.
+		sts.Annotations[deletionTimestampAnnotation] = c.getRetentionDate(p)
```
simonpasquier, line 1003: the statefulset is never updated though? Should this PR only deal with retain vs. delete and we address the retention timestamp in a follow-up?

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens: I'm still a bit confused 😬 How can we use the values from the API during reconciliation, but not expose it in the CRD?

pkg/prometheus/server/operator.go
ArthurSens: Yeah good point, I think I misjudged how this function was used when I was making the scope of this PR smaller.

I'll delete this and re-add once we deal with timestamps

ArthurSens:
> While reviewing, I wonder what should happen for Prometheus rules. Should they be deployed to scaled-down shards? My intuition says no (for instance, alerting rules usin `absent()` could fire) but I'm not 100% sure.

Good question, I don't think we've covered that in the proposal. I think we shouldn't have alerting rules in those prometheus as well.

nicolastakashi:
> > While reviewing, I wonder what should happen for Prometheus rules. Should they be deployed to scaled-down shards? My intuition says no (for instance, alerting rules usin `absent()` could fire) but I'm not 100% sure.
> 
> Good question, I don't think we've covered that in the proposal. I think we shouldn't have alerting rules in those prometheus as well.

I think we don't have any issue by lefting PrometheusRules, since the targets will be removed rules will be evaluted an will produce no results.

simonpasquier:
> > > While reviewing, I wonder what should happen for Prometheus rules. Should they be deployed to scaled-down shards? My intuition says no (for instance, alerting rules usin `absent()` could fire) but I'm not 100% sure.
> > 
> > 
> > Good question, I don't think we've covered that in the proposal. I think we shouldn't have alerting rules in those prometheus as well.
> 
> I think we don't have any issue by lefting PrometheusRules, since the targets will be removed rules will be evaluted an will produce no results.

Except if you have rules using absent() as I mentioned above.

ArthurSens:
So... is anything missing in this PR? Besides the current conflicts? I'll try to solve them today or tomorrow :)

simonpasquier (commented):
I have some comments on the API definition but I tested locally and it worked fine. One thing I've noted is that the status subresoure shows (as expected) only 1 replica when you scale down from 2 to 1 with policy=retain:

```
status:                                                                                        
  availableReplicas: 1                                                                                                                                                                        
[...]
  replicas: 1
  shardStatuses:                                                                               
  - availableReplicas: 1
    replicas: 1
    shardID: "0"
    unavailableReplicas: 0
    updatedReplicas: 1
  shards: 1
  unavailableReplicas: 0
  updatedReplicas: 1
````

This isn't for this PR but I wonder if we should reflect the "retained" shards in the shardStatuses field?

File: pkg/apis/monitoring/v1/prometheus_types.go [hunk 18]
```diff
@@ -875,6 +875,18 @@ type PrometheusSpec struct {
 	// Maximum number of bytes used by the Prometheus data.
 	RetentionSize ByteSize `json:"retentionSize,omitempty"`
 
+	// ShardRetentionPolicy defines the retention policy for the Prometheus shards.
+	// (Alpha) Using this field requires the ShardRetentionPolicies feature gate to be enabled.
+	//
+	// The final goals for this feature can be seen at https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#graceful-scale-down-of-prometheus-servers,
+	// however, the feature is not yet fully implemented and has known limitations:
```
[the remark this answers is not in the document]
ArthurSens: I feel like the link to the proposal gives extra context for the known limitations below 😬

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -875,6 +875,18 @@ type PrometheusSpec struct {
...
+	// * Scale up after scale down will not re-use the previously marked-for-deletion shards.
```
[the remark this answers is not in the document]
ArthurSens: In the proposal, we say that, in case a scale up happens after a scale down that left shards undelete, we would "turn on" the previous shard instead of creating a third one.

See: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#scaling-up-after-scaling-down

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -875,6 +875,18 @@ type PrometheusSpec struct {
...
+	// * Target redistribution doesn't happen on scale up/down.
```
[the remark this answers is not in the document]
ArthurSens: Yeah I think I went ahead of myself and documented as if the whole feature was already implemented 😬

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -991,6 +1003,24 @@ type PrometheusSpec struct {
...
+
+type ShardRetentionPolicy struct {
+	// WhenScaledDown defines the retention policy when the Prometheus Server is scaled-down.
+	// When set to `Retain`, the operator will keep the pods from the scaled-down shard(s). The pods won't scrape targets anymore but they can still be queried.
+	// The targets scraped by the scaled-down shard(s) are redistributed to the remaining shards.
+	//
+	// Default: "Delete"
+	// +kubebuilder:validation:Enum=Retain;Delete
+	// +required
+	WhenScaledDown WhenScaledDownRetentionType `json:"whenScaledDown"`
```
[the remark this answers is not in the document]
ArthurSens: That was the [original idea](https://github.com/prometheus-operator/prometheus-operator/pull/6198#discussion_r1621882946) hahaha, I don't remember if I looked at the statefulset implementation or if it was just by chance. 

Now that the statefulset implementation was mentioned, I think it's better to follow their direction :) I'll update the PR

ArthurSens:
> I have some comments on the API definition but I tested locally and it worked fine. One thing I've noted is that the status subresoure shows (as expected) only 1 replica when you scale down from 2 to 1 with policy=retain:
> 
> ```
> status:                                                                                        
>   availableReplicas: 1                                                                                                                                                                        
> [...]
>   replicas: 1
>   shardStatuses:                                                                               
>   - availableReplicas: 1
>     replicas: 1
>     shardID: "0"
>     unavailableReplicas: 0
>     updatedReplicas: 1
>   shards: 1
>   unavailableReplicas: 0
>   updatedReplicas: 1
> ```
> 
> This isn't for this PR but I wonder if we should reflect the "retained" shards in the shardStatuses field?

Good question! The shard number needs to be updated with the correct number of active shards, to keep HPAs working. But keeping shards that were marked for deletion somewhere in the status seems valuable for debugging!

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
simonpasquier: Right and it wouldn't be the case with your PR?

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
simonpasquier: I agree, the link to the proposal is valuable.

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
ArthurSens: Hmmm, I haven't tested that actually 🤔 

I haven't done anything specifically for that to work, so I assumed it just wasn't working like that. I'll try to test it out later today and see what happens

simonpasquier:
superseded by #7274

PR #6199, opened 2024-01-02
feat: adding scrape class

nicolastakashi:
## Description

Adding Scrape Class support

## Type of change

## Changelog entry

```release-note
feat: adding scrape class
```


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1750,3 +1756,22 @@ type AuthorizationValidationError struct {
 func (e *AuthorizationValidationError) Error() string {
 	return e.err
 }
+
+type ScrapeClass struct {
+	// Name of the scrape class, e.g. "istio-mtls".
```
ArthurSens, line 1761:
```suggestion
	// Name of the scrape class.
```

I know the istio-mtls is a common use case, but I don't see a reason for keeping it as example

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1750,3 +1756,22 @@ type AuthorizationValidationError struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Name string `json:"name"`
+
+	// Default designates a scrape class as the default class.
+	// +optional
+	Default *bool `json:"default,omitempty"`
```
ArthurSens, line 1768: What happens if multiple scrape classes have `default: true`?

ArthurSens, line 1768: What happens if the user wants to create scrapeclasses, but not set a default?

ArthurSens (commented):
Awesome to see this!

I've made a very brief review, without really trying things out myself yet. I'll take a deeper look tomorrow!

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: We will fail the resource reconciliation: https://github.com/prometheus-operator/prometheus-operator/pull/6199/files#diff-453ce9c721c51bf9591c2b212e9ce50157b9c40d1b211177bf502be268099656R113

nicolastakashi: Nothing!
Scrape Class will be available for use but will not be added in any monitor resource

ArthurSens (commented):
I wonder if we couldn't have a `CommonScrapeConfiguration` struct that is inlined in multiple CRD specs, similar to `CommonPrometheusFields` that is shared between Prometheus and PrometheusAgent.

The ScrapeClass definition would look like this:

```go
type ScrapeClass struct {
	// Name of the scrape class.
	// +required
	Name string `json:"name"`

	// Default designates a scrape class as the default class.
	// +optional
	Default *bool `json:"default,omitempty"`

	CommonScrapeConfiguration `json:",inline"`
}

type CommonScrapeConfiguration struct {
	// TLSConfig section for scrapes.
	// +optional

	// Authorization section for scrapes.
	// +optional
	Authorization *Authorization `json:"authorization,omitempty"`
}
```

PodMonitor, ServiceMonitor and Probes would also use this new inlined struct, and adding new fields to ScrapeClasses becomes just a matter of moving fields to CommonScrapeConfiguration

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
+	testCases := []struct {
+		name        string
+		scrapeClass *monitoringv1.ScrapeClass
+		golden      string
+		expectedErr bool
+	}{
+		{
+			name:        "Service Monitor without Scrape Class",
+			golden:      "serviceMonitorWithoutScrapeClass.golden",
+			scrapeClass: nil,
```
ArthurSens, line 6459: this line is unnecesary

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
+				Endpoints: []monitoringv1.Endpoint{
+					{
+						Port:     "web",
+						Interval: "30s",
+					},
+				},
+			},
+		}
+
+		if tc.scrapeClass != nil {
```
ArthurSens, line 6522: would it be a problem if we don't do a nil check here? My expectation that a nil scrapeclass would simply not generate anything

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
+			},
+			nil,
+			nil,
+			nil,
+			&assets.Store{},
+			nil,
+			nil,
+			nil,
+			nil,
+		)
```
ArthurSens, line 6551:
```suggestion
		cfg, err := cg.GenerateServerConfiguration(
			context.Background(),
			p.Spec.EvaluationInterval,
			p.Spec.QueryLogFile,
			p.Spec.RuleSelector,
			p.Spec.Exemplars,
			p.Spec.TSDB,
			p.Spec.Alerting,
			p.Spec.RemoteRead,
			map[string]*monitoringv1.ServiceMonitor{"monitor": &monitor},
			map[string]*monitoringv1.PodMonitor{"monitor": &monitor},
			map[string]*monitoringv1.Probes{"monitor": &monitor},
			map[string]*monitoringv1alpha.ScrapeConfig{"monitor": &monitor},
			&assets.Store{},
			nil,
			nil,
			nil,
			nil,
		)
```

What do you think about doing a single test for all four, since the code for all 4 tests are 99% identical?

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
+					CredentialsFile: "/etc/prometheus/secrets/default",
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		p := defaultPrometheus()
+
+		monitor := monitoringv1.PodMonitor{
```
ArthurSens, line 6614: similar comment... `defaultPodMonitor()`

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
+					CredentialsFile: "/etc/prometheus/secrets/default",
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		p := defaultPrometheus()
+
+		monitor := monitoringv1.Probe{
```
ArthurSens, line 6729: `defaultProbe()`

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
+					CredentialsFile: "/etc/prometheus/secrets/default",
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		p := defaultPrometheus()
+
+		monitor := monitoringv1alpha1.ScrapeConfig{
```
ArthurSens, line 6857: `defaultScrapeConfig()`

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
 	}
 }
+
+func TestServiceMonitorScrapeClass(t *testing.T) {
+	// test table for scrape class
+	testCases := []struct {
+		name        string
+		scrapeClass *monitoringv1.ScrapeClass
+		golden      string
+		expectedErr bool
```
ArthurSens, line 6454: expectedErr is false in all test cases... do we really need it? Maybe you forgot a test case where an error was expected? 😛

ArthurSens (commented):
500 hundred line changes for this test sounds a bit too much, some suggestions to make it more readable:

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -650,6 +650,12 @@ type CommonPrometheusFields struct {
 	// +optional
 	// +kubebuilder:validation:Minimum=60
 	MaximumStartupDurationSeconds *int32 `json:"maximumStartupDurationSeconds,omitempty"`
+
+	// EXPERIMENTAL List of scrape classes to expose to monitors and other scrape configs.
+	// This is experimental feature and might change in the future.
+	// +listType=map
+	// +listMapKey=name
+	ScrapeClasses []ScrapeClass `json:"scrapeClasses,omitempty"`
```
ArthurSens, line 658:
```suggestion
	ScrapeClasses []*ScrapeClass `json:"scrapeClasses,omitempty"`
```

ScrapeClass is an object that we expect to grow, a lot! If we work with value receivers, instead of pointers, I suspect that reconciling Prometheus instances with a large amount of ScrapeClasses could increase memory allocations

I explain in the comments below, but a good thread on Twitter also explains the difference of working with [Pointers vs Value receivers](https://twitter.com/MattJamesBoyle/status/1742822532372136433).

pkg/prometheus/promcfg.go
```diff
@@ -81,14 +85,37 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
 		version:                version,
 		prom:                   p,
 		endpointSliceSupported: endpointSliceSupported,
+		scrapeClasses:          scrapeClasses,
+		defaultScrapeClass:     defaultScrapeClass,
 	}, nil
 }
 
+func getScrapeClassConfig(cpf monitoringv1.CommonPrometheusFields) (map[string]monitoringv1.ScrapeClass, string, error) {
+	scrapeClasses := map[string]monitoringv1.ScrapeClass{}
```
ArthurSens, line 104: We can start the improvement by pre-allocating the map size
```suggestion
	scrapeClasses := make(map[string]*monitoringv1.ScrapeClass{}, len(cpf.ScrapeClasses))
```

pkg/prometheus/promcfg.go
```diff
@@ -81,14 +85,37 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
+	defaultScrapeClass := ""
+	for _, scrapeClass := range cpf.ScrapeClasses {
+		scrapeClasses[scrapeClass.Name] = scrapeClass
```
ArthurSens, line 107: In this line, we're copying the whole ScrapeClass object inside the map, since it isn't a pointer, and we do it once for each ScrapeClass. We could just copy the pointer address instead

At the moment ScrapeClass only has 2 configuration options, TLS and Authorization, but once we add relabelings... 💀

ArthurSens (commented):
You can call this early optimization because I haven't benchmarked it yet, but I suspect we can avoid some memory allocations here

pkg/prometheus/promcfg.go
nicolastakashi: Good point 👍🏽

pkg/prometheus/promcfg_test.go
nicolastakashi: I added a test case to ensure default behavior, if not scrap class is provided we will not add it to the Prometheus Object and as a consequence Monitors should not have the scrape class on their spec as well.

The other way I see this is make the scrape property on the table test struct definition a array of scrape class and than add an empty array.

no strong opinion over here, if you think is better I can change it 😄

pkg/prometheus/promcfg_test.go
nicolastakashi: I'd prefer have dedicated tests for each since IMHO this is more easy to maintain and evolve, but no strong opinion again.

I'm not sure if we will not have many checks on the code since there's no common spec against the monitors specs.

pkg/prometheus/promcfg_test.go
```diff
@@ -6444,3 +6444,500 @@ func TestTracingConfig(t *testing.T) {
...
+					},
+				},
+				Endpoints: []monitoringv1.Endpoint{
+					{
+						Port:     "web",
+						Interval: "30s",
+					},
+				},
+			},
+		}
```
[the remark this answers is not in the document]
nicolastakashi: We can, but I'm wondering what will be the other use cases for that?
Do you think we will change the other tests that is only using only one service monitor to this function?
Because on this current context I can't see value other than redeuce line of codes on the test func.

pkg/prometheus/promcfg_test.go
nicolastakashi: Ups!
Removed 😅

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: I totally agree, btw thanks for sharing the thread 🥳

pkg/prometheus/promcfg.go
nicolastakashi: For sure 😄

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1750,3 +1756,22 @@ type AuthorizationValidationError struct {
...
 	return e.err
 }
+
+type ScrapeClass struct {
+	// Name of the scrape class.
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Name string `json:"name"`
+
+	// Default designates a scrape class as the default class.
```
simonpasquier, line 1766:
```suggestion
	// Default indicates that the scrape applies to all scrape objects that don't configure an explicit scrape class name.
	//
	// Only one scrape class can be set as default.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1750,3 +1756,22 @@ type AuthorizationValidationError struct {
...
+	// +optional
+	Default *bool `json:"default,omitempty"`
+
+	// TLSConfig section for scrapes.
+	// +optional
+	TLSConfig *TLSConfig `json:"tlsConfig,omitempty"`
+
+	// Authorization section for scrapes.
+	// +optional
+	Authorization *Authorization `json:"authorization,omitempty"`
```
simonpasquier, line 1776: To keep it simple, I'd implement TLS configuration only (esp. because it was the main use case driving this request).

pkg/prometheus/store.go
```diff
@@ -102,3 +102,15 @@ func AddAPIServerConfigToStore(ctx context.Context, store *assets.Store, namespa
 	}
 	return nil
 }
+
+func AddScrapeClassConfigToStore(ctx context.Context, store *assets.Store, namespace string, scrapeClasses []*monv1.ScrapeClass) error {
+	for i, scrapeClass := range scrapeClasses {
+		if err := store.AddTLSConfig(ctx, namespace, scrapeClass.TLSConfig); err != nil {
+			return fmt.Errorf("scrape class %d: %w", i, err)
```
simonpasquier, line 109: the name should be unique
```suggestion
			return fmt.Errorf("scrape class %q: %w", scrapeClass.Name, err)
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -650,6 +650,12 @@ type CommonPrometheusFields struct {
 	// +optional
 	// +kubebuilder:validation:Minimum=60
 	MaximumStartupDurationSeconds *int32 `json:"maximumStartupDurationSeconds,omitempty"`
+
+	// EXPERIMENTAL List of scrape classes to expose to monitors and other scrape configs.
+	// This is experimental feature and might change in the future.
+	// +listType=map
+	// +listMapKey=name
+	ScrapeClasses []*ScrapeClass `json:"scrapeClasses,omitempty"`
```
simonpasquier, line 658: no need for a pointer?
```suggestion
	ScrapeClasses []ScrapeClass `json:"scrapeClasses,omitempty"`
```

pkg/apis/monitoring/v1/podmonitor_types.go
```diff
@@ -120,6 +120,10 @@ type PodMonitorSpec struct {
 	//
 	// +optional
 	AttachMetadata *AttachMetadata `json:"attachMetadata,omitempty"`
+
+	// The scrape class to apply.
+	// +optional
+	ScrapeClass *string `json:"scrapeClass,omitempty"`
```
simonpasquier, line 126:
```suggestion
        // +kubebuilder:validation:MinLength=1
	ScrapeClass *string `json:"scrapeClass,omitempty"`
```

pkg/prometheus/resource_selector.go
```diff
@@ -121,6 +121,10 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
 	for namespaceAndName, sm := range serviceMonitors {
 		var err error
 
+		if sm.Spec.ScrapeClass != nil {
+			err = validateScrapeClass(rs.p, *sm.Spec.ScrapeClass)
```
simonpasquier, line 125: (nit)
```suggestion
		err = validateScrapeClass(rs.p, ptr.Deref(sm.Spec.ScrapeClass, ""))
```

pkg/prometheus/promcfg.go
```diff
@@ -781,6 +815,8 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	}
 	if ep.TLSConfig != nil {
 		cfg = addSafeTLStoYaml(cfg, m.Namespace, ep.TLSConfig.SafeTLSConfig)
+	} else if scrapeClass != nil {
```
simonpasquier, line 818: the proposal stated that the TLS configs would be merged with the monitor object's fields taking precedence.

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: @simonpasquier this was introduced based on a suggestion provided by @ArthurSens: https://github.com/prometheus-operator/prometheus-operator/pull/6199#discussion_r1460690072

wdyt?

simonpasquier: The link talks about methods with pointer vs. value receivers which is a bit different than here.
If the API allows `[]*ScrapeClasses`, this would be a valid spec though useless:
```
spec:
  scrapeClasses:
  - null
  - name: foo
    tlsConfig: {...}
  - null
````

In this case, I'd rather favor the API consistency than optimization.

Comment (ArthurSens) [reply pkg/apis/monitoring/v1/prometheus_types.go:658]: No problems from my side, as mentioned in the review that was early optimization :)

File: pkg/prometheus/promcfg_test.go (hunk 13 above)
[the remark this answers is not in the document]
Comment (nicolastakashi) [reply pkg/prometheus/promcfg_test.go]: @ArthurSens changed 😄

File: pkg/prometheus/promcfg_test.go (hunk 4 above)
Comment (nicolastakashi) [reply pkg/prometheus/promcfg_test.go]: @ArthurSens changed 😄

File: pkg/prometheus/promcfg_test.go (hunk 5 above)
Comment (nicolastakashi) [reply pkg/prometheus/promcfg_test.go]: @ArthurSens changed 😄

File: pkg/prometheus/resource_selector.go [hunk 21]
```diff
@@ -513,6 +535,12 @@ func (rs *ResourceSelector) SelectProbes(ctx context.Context, listFn ListAllByNa
 			rs.eventRecorder.Eventf(probe, v1.EventTypeWarning, operator.InvalidConfigurationEvent, "Probe %s was rejected due to invalid configuration: %v", probe.GetName(), err)
 		}
 
+		if probe.Spec.ScrapeClass != nil {
+			err = validateScrapeClass(rs.p, *probe.Spec.ScrapeClass)
+			rejectFn(probe, err)
+			continue
+		}
```
ArthurSens, line 542:
```suggestion
		if err = validateScrapeClass(rs.p, ptr.Deref(probe.Spec.ScrapeClass, "")); err != nil {
			rejectFn(probe, err)
			continue
		}
```

pkg/prometheus/resource_selector.go
```diff
@@ -676,6 +704,12 @@ func (rs *ResourceSelector) SelectScrapeConfigs(ctx context.Context, listFn List
 			rs.eventRecorder.Eventf(sc, v1.EventTypeWarning, operator.InvalidConfigurationEvent, "ScrapeConfig %s was rejected due to invalid configuration: %v", sc.GetName(), err)
 		}
 
+		if sc.Spec.ScrapeClass != nil {
+			err = validateScrapeClass(rs.p, *sc.Spec.ScrapeClass)
+			rejectFn(sc, err)
+			continue
+		}
```
ArthurSens, line 711:
```suggestion
		if err = validateScrapeClass(rs.p, ptr.Deref(sc.Spec.ScrapeClass, "")); err != nil {
			rejectFn(sc, err)
			continue
		}
```

pkg/prometheus/resource_selector.go
ArthurSens: I have the impression that the current code always rejects the scrapeClass 🤔

pkg/prometheus/resource_selector_test.go
```diff
@@ -421,9 +421,10 @@ func TestValidateRelabelConfig(t *testing.T) {
 
 func TestSelectProbes(t *testing.T) {
```
ArthurSens, line 422: Could we test that the monitor resource gets selected with a valid scrapeClass?

ArthurSens:
It would also be nice to test the behavior of ScrapeClass+a monitor resource with some scrape configuration. Do we want to merge the config? Override completely?

pkg/prometheus/store.go
```diff
@@ -102,3 +102,12 @@ func AddAPIServerConfigToStore(ctx context.Context, store *assets.Store, namespa
 	}
 	return nil
 }
+
+func AddScrapeClassConfigToStore(ctx context.Context, store *assets.Store, namespace string, scrapeClasses []monv1.ScrapeClass) error {
```
simonpasquier, line 106: (nit)
```suggestion
func AddScrapeClassesToStore(ctx context.Context, store *assets.Store, namespace string, scrapeClasses []monv1.ScrapeClass) error {
```

pkg/prometheus/promcfg.go
```diff
@@ -795,7 +888,9 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	assetKey := fmt.Sprintf("podMonitor/%s/%s/%d", m.Namespace, m.Name, i)
 	cfg = cg.addOAuth2ToYaml(cfg, ep.OAuth2, store.OAuth2Assets, assetKey)
 
-	cfg = cg.addSafeAuthorizationToYaml(cfg, fmt.Sprintf("podMonitor/auth/%s/%s/%d", m.Namespace, m.Name, i), store, ep.Authorization)
+	if ep.Authorization != nil {
```
simonpasquier, line 891: addSafeAuthorizationToYaml() supports nil Authorization.

pkg/prometheus/promcfg.go
```diff
@@ -1248,7 +1351,14 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 	assetKey := fmt.Sprintf("serviceMonitor/%s/%s/%d", m.Namespace, m.Name, i)
 	cfg = cg.addOAuth2ToYaml(cfg, ep.OAuth2, store.OAuth2Assets, assetKey)
 
-	cfg = addTLStoYaml(cfg, m.Namespace, ep.TLSConfig)
+	if ep.TLSConfig != nil {
+		cfg = addTLStoYaml(cfg, m.Namespace, ep.TLSConfig)
+	}
+
+	if scrapeClass != nil {
+		cfg = addScrapeClassTLSConfigToYaml(cfg, m.Namespace, scrapeClass)
```
simonpasquier, line 1359: IMHO it would be better/simpler to merge `ep.TLSConfig` with the selected scrape class before and then call `addTLStoYaml` on the result.

nicolastakashi: I was working on that.
But we have differences between service monitor and other monitors.
Service monitor allows a TLSConfig while the other allow only a SafeTLS, and using this approach will lead us with two different code one to merge TLS and other to merge SafeTLS config.

I was thinking about keep this implementation since it's more generic and work as expected, wdyt?

pkg/prometheus/resource_selector.go
```diff
@@ -121,6 +121,10 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
 	for namespaceAndName, sm := range serviceMonitors {
 		var err error
 
+		if sm.Spec.ScrapeClass != nil {
```
simonpasquier, line 124: you'd to move this after the for loop otherwise the err might be swallowed.

pkg/prometheus/resource_selector.go
```diff
@@ -323,6 +327,20 @@ func validateRelabelConfig(p monitoringv1.PrometheusInterface, rc monitoringv1.R
 	return nil
 }
 
+func validateScrapeClass(p monitoringv1.PrometheusInterface, sc string) error {
+	if sc == "" {
+		return nil
+	}
```
simonpasquier, line 333: (nit)
```suggestion
func validateScrapeClass(p monitoringv1.PrometheusInterface, sc *string) error {
	if ptr.Deref(sc, "") == "" {
		return nil
	}
```

pkg/prometheus/promcfg.go
```diff
@@ -81,14 +85,37 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
+		defaultScrapeClass:     defaultScrapeClass,
 	}, nil
 }
 
+func getScrapeClassConfig(cpf monitoringv1.CommonPrometheusFields) (map[string]*monitoringv1.ScrapeClass, string, error) {
+	scrapeClasses := make(map[string]*monitoringv1.ScrapeClass, len(cpf.ScrapeClasses))
+	defaultScrapeClass := ""
+	for _, scrapeClass := range cpf.ScrapeClasses {
+		scrapeClasses[scrapeClass.Name] = &scrapeClass
+		if scrapeClass.Default != nil && *scrapeClass.Default {
```
simonpasquier, line 108: (nit)
```suggestion
		if ptr.Deref(scrapeClass.Default, false) {
```

pkg/prometheus/promcfg.go
```diff
@@ -348,6 +381,34 @@ func addTLStoYaml(cfg yaml.MapSlice, namespace string, tls *monitoringv1.TLSConf
 	return cfg
 }
 
+func mergeTLSConfigAndScrapeClassTLSConfig(tlsConfig *monitoringv1.TLSConfig, scrapeClass *monitoringv1.ScrapeClass) *monitoringv1.TLSConfig {
```
simonpasquier, line 384: (suggestion) it could be a ConfigGenerator method:
```suggestion
func (cg *ConfigGenerator) MergeTLSConfigWithScrapeClass(tlsConfig *monitoringv1.TLSConfig, scrapeClassName string) *monitoringv1.TLSConfig {
    if scrapeClassName == "" {
        scrapeClassName = cg.defaultScrapeClassName
    }
    
    scrapeClass, found := cg.scrapeClasses[scrapeClassName]
    if !found {
        return tlsConfig
    }
    
    ...
```

pkg/prometheus/promcfg.go
```diff
@@ -61,6 +61,8 @@ type ConfigGenerator struct {
 	notCompatible          bool
 	prom                   monitoringv1.PrometheusInterface
 	endpointSliceSupported bool
+	scrapeClasses          map[string]*monitoringv1.ScrapeClass
+	defaultScrapeClass     string
```
simonpasquier, line 65:
```suggestion
	defaultScrapeClassName     string
```

pkg/prometheus/server/operator.go
```diff
@@ -1348,6 +1348,10 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *mon
 		}
 	}
 
+	if err := prompkg.AddScrapeClassesToStore(ctx, store, p.GetNamespace(), p.Spec.ScrapeClasses); err != nil {
+		return err
```
simonpasquier, line 1352:
```suggestion
		return fmt.Errorf("failed to process scrape classes: %w", err)
```

pkg/prometheus/resource_selector.go
```diff
@@ -377,6 +395,10 @@ func (rs *ResourceSelector) SelectPodMonitors(ctx context.Context, listFn ListAl
 	for namespaceAndName, pm := range podMonitors {
 		var err error
 
+		if pm.Spec.ScrapeClass != nil {
```
simonpasquier, line 398: same remark here

pkg/apis/monitoring/v1/podmonitor_types.go
```diff
@@ -131,6 +131,10 @@ type PodMonitorSpec struct {
...
```
simonpasquier, line 137: can we add `+kubebuilder:validation:MinLength=1`?

pkg/prometheus/resource_selector.go
nicolastakashi: Can you review @simonpasquier ?

pkg/prometheus/promcfg.go
```diff
@@ -348,6 +381,46 @@ func addTLStoYaml(cfg yaml.MapSlice, namespace string, tls *monitoringv1.TLSConf
 	return cfg
 }
 
+func (cg *ConfigGenerator) MergeTLSConfigWithScrapeClass(tlsConfig *monitoringv1.TLSConfig, scrapeClass *monitoringv1.ScrapeClass) *monitoringv1.TLSConfig {
```
ArthurSens, line 384: Could we have a unit test for this function?

I'm afraid this will become quite big as we add more fields to ScrapeClass in the future, a test will help us keep this working :)

ArthurSens (approved):
2 non blocking comments

Thanks for taking this to the finish line ❤️

pkg/prometheus/resource_selector.go
```diff
@@ -120,65 +120,79 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
...
+		}
 
 		for i, endpoint := range sm.Spec.Endpoints {
 			// If denied by Prometheus spec, filter out all service monitors that access
 			// the file system.
 			if cpf.ArbitraryFSAccessThroughSMs.Deny {
 				if err = testForArbitraryFSAccess(endpoint); err != nil {
-					break
+					rejectFn(sm, err)
+					continue
```
simonpasquier, line 141: IIUC it introduces a bug: if the service monitor contains 1 invalid endpoint followed by 1 valid endpoint, it will still be selected.

nicolastakashi: Thanks @simonpasquier https://github.com/prometheus-operator/prometheus-operator/pull/6339

PR #6200, opened 2024-01-03
chore: refactor ShardedSecret

simonpasquier:
## Description

This commit reorganizes the code makinguse of `SharedSecret` to reduce duplication.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/operator/sharded_secret.go
+++ b/pkg/operator/sharded_secret.go
@@ -22,8 +22,10 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 
+	"github.com/prometheus-operator/prometheus-operator/pkg/assets"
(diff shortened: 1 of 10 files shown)


pkg/operator/sharded_secret.go
```diff
@@ -32,123 +34,138 @@ import (
...
+
+// Append adds a new key + data pair.
+// If the key already exists, data gets overwritten.
+func (s *ShardedSecret) Append(k fmt.Stringer, v Byter) {
+	s.appendKeyValue(k.String(), v.Bytes())
+}
+
+func (s *ShardedSecret) appendKeyValue(key string, data []byte) {
 	s.data[key] = data
 }
```
ArthurSens, line 66: looks like a shallow function 🤔 

Could we rewrite the function above to:
```go
// Append adds a new key + data pair.
// If the key already exists, data gets overwritten.
func (s *ShardedSecret) Append(k fmt.Stringer, v Byter) {
	s.data[k.String()] = v.Bytes()
}
```
?

ArthurSens (commented):
Just one question

pkg/operator/sharded_secret.go
simonpasquier: You're correct. My rationale was that the tests use `appendKeyValue()` so it had to exist anyway but I see your point. I'll fix it.

ArthurSens (approved):

PR #6201, opened 2024-01-03
chore: refactor generateScrapeConfig()

simonpasquier:
## Description

From a comment while reviewing #6153.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -2329,11 +2329,6 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		cfg = append(cfg, yaml.MapItem{Key: "scrape_timeout", Value: *sc.Spec.ScrapeTimeout})
 	}
 
-	if sc.Spec.RelabelConfigs != nil {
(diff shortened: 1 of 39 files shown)


philipgough (approved):
lgtm

pkg/prometheus/promcfg.go
```diff
@@ -3096,6 +3091,14 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		cfg = append(cfg, yaml.MapItem{Key: "metric_relabel_configs", Value: generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.MetricRelabelConfigs))})
 	}
 
+	if sc.Spec.RelabelConfigs != nil {
+		relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(sc.TypeMeta, sc.ObjectMeta, sc.Spec.RelabelConfigs))...)
+	}
+
+	if len(relabelings) > 0 {
```
ArthurSens, line 3098: Since `relabelings` is never 0, it is initialized with `initRelabelings()`, would it be fine to not even check the array length?

simonpasquier: I thought about that and went with the defensive option. But Prometheus wouldn't mind if `relabelings: null` so let's go with the simpler option.

PR #6283, opened 2024-02-02
feat: add version check for thanos. keep_firing_for now available

DeamonMV:
## Description

Added version check for thanos in the context of keep_firing_for. starting from thanos v0.34.0 keep_firing_for is available.

## Type of change

## Verification

## Changelog entry

```
feat: add version check for thanos. keep_firing_for now available
```

### DIFF
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -107,6 +107,7 @@ func (prs *PrometheusRuleSelector) sanitizePrometheusRulesSpec(promRuleSpec moni
 	component := "Prometheus"
 
 	if prs.ruleFormat == ThanosFormat {
+		minVersionKeepFiringFor = semver.MustParse("0.34.0")
 		minVersionLimits = semver.MustParse("0.24.0")
 		component = "Thanos"
 	}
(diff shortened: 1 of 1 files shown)


slashpai:
You need to add test for supported and unsupported version in https://github.com/prometheus-operator/prometheus-operator/blob/8f0897523bb3fa4155fe4549d05e5d671f108f70/pkg/operator/rules_test.go#L41

simonpasquier:
There's also no need to keep the existing switch/case.

DeamonMV:
@slashpai @simonpasquier 

I have pushed changes, could you guys, please, check if it's legitimately\correct?

pkg/operator/rules_test.go
```diff
@@ -243,12 +243,25 @@ func shouldDropKeepFiringForThanos(t *testing.T) {
 			},
 		}},
 	}
-
-	thanosVersion, _ := semver.ParseTolerant(DefaultThanosVersion)
-	pr := newRuleSelectorForConfigGeneration(ThanosFormat, thanosVersion)
-	content, _ := pr.generateRulesConfiguration(rules)
-	if strings.Contains(content, "keep_firing_for") {
-		t.Fatalf("expected `keep_firing_for` not to be present in PrometheusRule")
+	testCases := []struct {
```
simonpasquier, line 246: The test doesn't work as expected: the second sub-test should return a snippet with `keep_firing_for`. It's because generateRulesConfiguration() mutates the rules struct.
Also the `shouldDropKeepFiringForThanos` test function needs to be renamed or a new test function should be added.

DeamonMV: I hope I got your point. 
Changes were pushed.

pkg/operator/rules_test.go
```diff
@@ -227,7 +228,7 @@ func shouldAcceptRuleWithKeepFiringForPrometheus(t *testing.T) {
 	}
 }
 
-func shouldDropKeepFiringForThanos(t *testing.T) {
+func shouldDropRuleFiringForThanos(t *testing.T) {
```
simonpasquier, line 231: Not a big deal but you could have kept the original name?

simonpasquier (approved):
Thanks!

pkg/operator/rules_test.go
DeamonMV: I made this by example of Prometheus - [Accept and Drop](https://github.com/prometheus-operator/prometheus-operator/pull/6283/files/843d1c954b7127814897b6774daeec4066246237#diff-1b88b230701ac430ad9d7d1396568e0e2023aa33739acf2d9c7d604c82af35acR40).
And, I guess, because of now it's possible to use and that option can be Accepted and should be Dropped in case of version of thanos is "old".

DeamonMV:
I guess it would be good to squash commits, but I was to late :)

simonpasquier:
No issue, I've squashed on merge :)

PR #6290, opened 2024-02-07
feat: NoProxy

### LINKED ISSUE #6245: Support no_proxy for Alertmanager
### Component(s)

AlertManager, AlertManagerConfig

Alertmanager has multiple configuration options for [http_config](https://prometheus.io/docs/alerting/latest/configuration/#http_config), among them:

```
# Optional proxy URL.
[ proxy_url: [<string>](https://prometheus.io/docs/alerting/latest/configuration/#string) ]
# Comma-separated string that can contain IPs, CIDR notation, domain names
# that should be excluded from proxying. IP and domain names can
# contain port numbers.
[ no_proxy: [<string>](https://prometheus.io/docs/alerting/latest/configuration/#string) ]
```

Currently operator is aware about proxy_url only -> https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.HTTPConfig

I received an error, when trying to configure no_proxy:

`{"caller":"klog.go:116","component":"k8s_client_runtime","func":"ErrorDepth","level":"error","msg":"sync \"monitoring/k8s-kube-prometheus-stack-alertmanager\" failed: provision alertmanager configuration: failed to initialize from secret: yaml: unmarshal errors:\n  line 3: field no_proxy not found in type alertmanager.httpClientConfig","ts":"2024-01-16T14:33:36.504542825Z"}
`

It would be great, if operator will support no_proxy as well.

There is also a possibility for Alertmanager to take http_config values from environment variables:

```
# Use proxy URL indicated by environment variables (HTTP_PROXY, https_proxy, HTTPs_PROXY, https_proxy, and no_proxy)
[ proxy_from_environment: [<boolean>](https://prometheus.io/docs/alerting/latest/configuration/#boolean) | default: false ]
```

But operator is not aware of proxy_from_environment neither.

## Environment
Prometheus-Operator Version: v0.70.0

alexandersperling:
* Add missing option for no_proxy

## Description

* Add a new type for http_config to allow passing of `no_proxy` value

TBH: I'm not sure, if this is everything thats needed to make this option available.

fixes #6245

## Type of change

## Changelog entry

```release-note
* Add possibility to configure the `no_proxy` option for alertmanager 
```


pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -432,6 +432,10 @@ type HTTPConfig struct {
 	// FollowRedirects specifies whether the client should follow HTTP 3xx redirects.
 	// +optional
 	FollowRedirects *bool `json:"followRedirects,omitempty"`
+	// Comma-separated string that can contain IPs, CIDR notation, domain names
+	// that should be excluded from proxying. IP and domain names can
+	// contain port numbers.
+	NoProxy string `json:"NoProxy,omitempty"`
```
nicolastakashi, line 438: Can you please add // +optional

simonpasquier:
```suggestion
	// +optional
	NoProxy *string `json:"NoProxy,omitempty"`
```

simonpasquier (commented):
The field needs to be propagated to the generated Alertmanager configuration.

pkg/apis/monitoring/v1/alertmanager_types.go
alexandersperling: I added the optional flag

alexandersperling:
> The field needs to be propagated to the generated Alertmanager configuration.

Is this something I have to trigger via make or do I miss some code for that here?

simonpasquier:
@alexandersperling you need to update this method to propagate the CRD field to the generated config but you'll also need to add the `noProxy` field to AlertmanagerConfig CRD:

https://github.com/prometheus-operator/prometheus-operator/blob/7c9524285ca745266223a19371d008c0279d08b2/pkg/alertmanager/amcfg.go#L1491

You need also to perform version checking here and discard the field if the Alertmanager version doesn't support it:

https://github.com/prometheus-operator/prometheus-operator/blob/7c9524285ca745266223a19371d008c0279d08b2/pkg/alertmanager/amcfg.go#L1694

alexandersperling:
@simonpasquier got some time and took a look into this again

could you check if this is fine now or if I still miss something?

ArthurSens (commented):
It's taking a good direction :) 

Could we also add a test in `pkg/alertmanager/amcfg_test.go` to verify the field gets populated and make sure we don't break the behavior in future PRs?

ArthurSens:
Also, I'm not an expert in contributions to Alertmanager CRDs, but do we need to add this field to the v1beta1 API?

PR #6319, opened 2024-02-22
Controller id implementation to avoid errors with multiple operators

marioferh:
Add controller-id flag for handling several instances in different namespaces and avoid issues with multiple operators.

This solution available for `Prometheus` `AlertManager`and `ThanosRuler` instances, as well as for
any pod-based resource that could be added in the future.


pkg/prometheus/agent/statefulset.go
```diff
@@ -45,6 +45,7 @@ func makeStatefulSet(
 	inputHash string,
 	shard int32,
 	tlsSecrets *operator.ShardedSecret,
+	controllerID string,
```
simonpasquier, line 48: I don't think that we need to propagate the controller ID to the statefulset and pods. Let's keep it simple for now.

pkg/prometheus/server/operator.go
```diff
@@ -1435,3 +1446,14 @@ func makeSelectorLabels(name string) map[string]string {
 		"prometheus":                    name,
 	}
 }
+
+// isOwnershipWrong returns false if the controller is the "owner" of the prometheus
+// or ownership has not been setted yet
+// Whether it's owner is determined by the value of label 'controllerID'
+// annotation. If the value matches the controllerID then it owns it.
+func (c *Operator) isOwnershipWrong(sset *appsv1.StatefulSet) bool {
+	if owner, ok := sset.ObjectMeta.Labels[prompkg.ControllerIDName]; ok {
```
simonpasquier, line 1455: the design proposal (https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202309-controller-id.md#how) suggested an annotation and I think that it's more appropriate than a label.

simonpasquier (commented):
I suggest to leverage the existing `ResourceReconciler` struct to implement the same behavior for all controllers in a single way.
In practice:
* ResourceReconciler gets a new "controller id" field. 
* `EnqueueForReconciliation()` and `EnqueueForStatus()` are called on pod/service monitor changes so they should check for the controller id annotation.
* `OnAdd()`, `OnUpdate()` and `OnDelete()` are called on Prometheus, Alertmanager, ... + statefulset changes and should perform the same check.

https://github.com/prometheus-operator/prometheus-operator/blob/f32c2c3075591d4feb63d88f60d30bfcc0e115e7/pkg/operator/resource_reconciler.go#L358-L366

https://github.com/prometheus-operator/prometheus-operator/blob/f32c2c3075591d4feb63d88f60d30bfcc0e115e7/pkg/operator/resource_reconciler.go#L232-L302

pkg/prometheus/server/operator.go
marioferh: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ 
`You can use either labels or annotations to attach metadata to Kubernetes objects. Labels can be used to select objects and to find collections of objects that satisfy certain conditions. In contrast, annotations are not used to identify and select objects.  `

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +471,14 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
 	}
 	return ns, nil
 }
+
+// isOwnershipWrong returns false if the controller is the "owner" of the prometheus
+// or ownership has not been setted yet
+// Whether it's owner is determined by the value of label 'controllerID'
+// annotation. If the value matches the controllerID then it owns it.
+func (rr *ResourceReconciler) isOwnershipWrong(labels map[string]string) bool {
```
machine424, line 479: I think it's more readable to use positive assertions: `isOwner` or `hasOwnership`

And we also need to consider the case where the the operator doesn't set `-controller-id`, in that case it shouldn't touch CRs that have a `controllerID` annotation. 

of course, if the operator doesn't set `-controller-id`, it should take care of CRs without `controllerID` annotation...

pkg/prometheus/server/operator.go
```diff
@@ -959,6 +964,11 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	}
 
 	p := pobj.(*monitoringv1.Prometheus)
+	operator.UpdateObject(
```
machine424, line 967: I don't think the operator should enforce the annotations on CRs, it's up to the user to provide CR's with or without the annotation

pkg/prometheus/server/operator.go
machine424: yes, if we're not going to need labels' features, an annotation is sufficient.

But we can go further, use a label, and add it as selector to all the lists and watches for Prometheus/AM etc., but this may imply more changes...


(maybe we should add a prefix to the annotation/label for more clarity)

pkg/operator/resource_reconciler.go
```diff
@@ -357,12 +361,22 @@ func (rr *ResourceReconciler) onStatefulSetDelete(ss *appsv1.StatefulSet) {
 
 // EnqueueForReconciliation asks for reconciling the object.
 func (rr *ResourceReconciler) EnqueueForReconciliation(obj metav1.Object) {
-	rr.reconcileQ.Add(obj.GetNamespace() + "/" + obj.GetName())
+	if rr.isOwnershipWrong(obj.GetLabels()) {
```
simonpasquier, line 364: It should work on annotations, not labels.

pkg/operator/resource_reconciler.go
simonpasquier, line 479: (suggestion) let's pass the object metadata
```suggestion
func (rr *ResourceReconciler) isManagedBy(o metav1.Object) bool {
```

marioferh: My idea is to force set a rand controller id, because if not, we can have exactly the same issue than before. 
hasOwnership() is not valid because if could be false and we need to enqueue the object in any case.
There is two scenarios:
- label is empty, and we need to set up
- label is correct and match operator controller id.

pkg/prometheus/server/operator.go
danielmellado: I assume that would also depend on the implementation details, if we aim to create a random controller-id for any new one then this should be kept as-is.

pkg/operator/resource_reconciler.go
machine424: I'm afraid a "rand controller id" would be more complex: we'll need to persist that on restarts and not knowing the ID in advance would make configuring the CRs to be managed by that operator more difficult...

I think that well chosen controller ID can prevent collisions.

The same overlap may happen with Service->Pods e.g. I can easily expose a Pod to 2 service by mistake  but that doesn't prevent users to chose good labels to prevent that...

danielmellado: I assume that what Mario wants to point is the case where you may find out that there's more than one operator without controller-id, which would lead us to the initial situation again.

machine424: If we implement this:

```
And we also need to consider the case where the the operator doesn't set -controller-id, in that case it shouldn't touch CRs that have a controllerID annotation.
```

users will have a way to "isolate/protect" their CRs.

simonpasquier: The initial plan wasn't to solve all edge cases: if 2 operators start without controller ids then we're back to the original problem as they might fight for the same resource.

The main goal is that we can "bind" objects to a specific operator when multiple operators run on the same cluster. It's the responsibility of the person deploying the Prometheus object to annotate it with the expected controller id. In the case of OCP, it means that we will start the openshift-monitoring operator with `--controller-id=foo` and we will annotate the k8sPrometheus object with `operator.prometheus.io/controller-id: foo`.

pkg/prometheus/agent/statefulset.go
marioferh: done

pkg/operator/resource_reconciler.go
marioferh: done

pkg/prometheus/server/operator.go
marioferh: right, done

pkg/operator/resource_reconciler.go
```diff
@@ -357,12 +386,26 @@ func (rr *ResourceReconciler) onStatefulSetDelete(ss *appsv1.StatefulSet) {
...
+	level.Debug(rr.logger).Log(
+		"msg", "mariofar onadd",
+		"object", fmt.Sprintf("%s/%s", obj.GetNamespace(), obj.GetName()),
+		"object", fmt.Sprintf("%s/%s", obj.GetGenerateName(), rr.resourceKind),
+	)
+
+	if rr.hasOwnership(obj) {
+		rr.reconcileQ.Add(obj.GetNamespace() + "/" + obj.GetName())
+	} else {
+		level.Error(rr.logger).Log("msg", "%s object has been created by another prometheus operator, controllerID does not match, please be careful deploying multiple operators", rr.resourceKind)
```
simonpasquier, line 398: If we want a log message, it needs to be at debug level. Also moving to hasOwnership() is probably better.

marioferh: done

pkg/operator/resource_reconciler.go
```diff
@@ -295,10 +310,17 @@ func (rr *ResourceReconciler) OnDelete(obj interface{}) {
 		return
 	}
 
-	level.Debug(rr.logger).Log("msg", fmt.Sprintf("%s deleted", rr.resourceKind), "key", key)
-	rr.metrics.TriggerByCounter(rr.resourceKind, DeleteEvent).Inc()
+	objMeta, err := meta.Accessor(obj)
+	if err != nil {
+		level.Error(rr.logger).Log("err", fmt.Sprintf("failed to get object meta: %s", err), "key", key)
```
simonpasquier, line 315: It isn't an error since the informer can return a cache.DeletedFinalStateUnknown object.
```suggestion
		return
```

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +489,16 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
 	}
 	return ns, nil
 }
+
+// hasOwnership returns true if the controller is the "owner" of the object.
+// Whether it's owner is determined by the value of 'controllerID'
+// annotation. If the value matches the controllerID then it owns it, or if the
+// controllerID is "" and there's no annotation set.
```
simonpasquier, line 496:
```suggestion
// controllerID is "" and there's no annotation set or the annotation value is empty.
```

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +489,16 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
...
+func (rr *ResourceReconciler) hasOwnership(obj metav1.Object) bool {
+	if owner, ok := obj.GetAnnotations()[controllerIDLabel]; ok {
+		return owner == rr.controllerID
+	}
+
+	level.Debug(rr.logger).Log("msg", "%s object has been created by another prometheus operator, controllerID does not match, please be careful deploying multiple operators", rr.resourceKind)
```
simonpasquier, line 502: the code needs to be rearranged to always log the debug message (currently it only logs if the annotation is missing).
```suggestion
	level.Debug(rr.logger).Log("msg", "skipping object not managed by the controller", "object", fmt.Sprintf("%s/%s", o.GetNamespace(), o.GetName()), "object_controller_id", obj.GetAnnotations()[controllerIDLabel], "controller_id", rr.controllerID)
```

pkg/operator/resource_reconciler.go
```diff
@@ -295,10 +316,19 @@ func (rr *ResourceReconciler) OnDelete(obj interface{}) {
 		return
 	}
 
-	level.Debug(rr.logger).Log("msg", fmt.Sprintf("%s deleted", rr.resourceKind), "key", key)
-	rr.metrics.TriggerByCounter(rr.resourceKind, DeleteEvent).Inc()
+	objMeta, err := meta.Accessor(obj)
```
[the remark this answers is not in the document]
simonpasquier: If the object is of type cache.DeletedFinalStateUnknown then the reconciler can't (and shouldn't) enqueue it for reconciliation. But in this case, meta.Accessor() will return an error that should be skipped silently.

pkg/operator/resource_reconciler.go
```diff
@@ -269,18 +282,20 @@ func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
 		level.Error(rr.logger).Log("err", fmt.Sprintf("failed to get object meta: %s", err), "key", key)
 	}
 
-	if rr.DeletionInProgress(mCur) {
-		return
-	}
+	if rr.hasOwnership(mCur) {
```
simonpasquier, line 285: (nit)
```suggestion
	if !rr.hasOwnership(mCur) {
		return
	}
```

pkg/operator/resource_reconciler.go
```diff
@@ -241,10 +247,17 @@ func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
...
 	}
 
-	level.Debug(rr.logger).Log("msg", fmt.Sprintf("%s added", rr.resourceKind), "key", key)
-	rr.metrics.TriggerByCounter(rr.resourceKind, AddEvent).Inc()
+	objMeta, err := meta.Accessor(obj)
+	if err != nil {
+		level.Error(rr.logger).Log("err", fmt.Sprintf("failed to get object meta: %s", err), "key", key)
+	}
+
+	if rr.hasOwnership(objMeta) {
```
simonpasquier, line 255: (nit)
```suggestion
	if !rr.hasOwnership(objMeta) {
		return
	}
```

pkg/operator/resource_reconciler.go
```diff
@@ -241,10 +247,17 @@ func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
 		return
 	}
 
-	level.Debug(rr.logger).Log("msg", fmt.Sprintf("%s added", rr.resourceKind), "key", key)
-	rr.metrics.TriggerByCounter(rr.resourceKind, AddEvent).Inc()
+	objMeta, err := meta.Accessor(obj)
+	if err != nil {
+		level.Error(rr.logger).Log("err", fmt.Sprintf("failed to get object meta: %s", err), "key", key)
```
simonpasquier, line 252:
```suggestion
		return
```

pkg/operator/resource_reconciler.go
```diff
@@ -295,10 +312,17 @@ func (rr *ResourceReconciler) OnDelete(obj interface{}) {
...
 
-	level.Debug(rr.logger).Log("msg", fmt.Sprintf("%s deleted", rr.resourceKind), "key", key)
-	rr.metrics.TriggerByCounter(rr.resourceKind, DeleteEvent).Inc()
+	objMeta, err := meta.Accessor(obj)
+	if err != nil {
+		return
+	}
 
-	rr.reconcileQ.Add(key)
+	if rr.hasOwnership(objMeta) {
```
simonpasquier, line 320: (nit) Let keep the same flow and exit if the condition isn't met, it also helps to review ;)
```suggestion
	if rr.hasOwnership(objMeta) {
		return
	}
```

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +491,16 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
...
+// Whether it's owner is determined by the value of 'controllerID'
+// annotation. If the value matches the controllerID then it owns it, or if the
+// controllerID is "" and there's no annotation set or the annotation value is empty.
+func (rr *ResourceReconciler) hasOwnership(obj metav1.Object) bool {
+	if owner, ok := obj.GetAnnotations()[controllerIDLabel]; ok {
+		return owner == rr.controllerID
+	}
+
+	level.Debug(rr.logger).Log("msg", "skipping object not managed by the controller", "object", fmt.Sprintf("%s/%s", obj.GetNamespace(), obj.GetName()), "object_controller_id", obj.GetAnnotations()[controllerIDLabel], "controller_id", rr.controllerID)
+	return false
```
simonpasquier, line 505: (suggestion) just in case, it's better to guard against nil map.

```suggestion
	var controllerID string
	if obj.GetAnnotations() != nil {
		controllerID = obj.GetAnnotations()[controllerIDLabel]
	}
	if controllerID != rr.controllerID {
		level.Debug(rr.logger).Log("msg", "skipping object not managed by the controller", "object", fmt.Sprintf("%s/%s", obj.GetNamespace(), obj.GetName()), "object_controller_id", controllerID, "controller_id", rr.controllerID)
		return false
	}

	return true
```

pkg/operator/resource_reconciler.go
```diff
@@ -85,10 +85,13 @@ type ResourceReconciler struct {
...
 
 	g errgroup.Group
+
+	controllerID string
 }
 
 var (
-	_ = cache.ResourceEventHandler(&ResourceReconciler{})
+	_                 = cache.ResourceEventHandler(&ResourceReconciler{})
+	controllerIDLabel = "operator.prometheus.io/controller-id"
```
simonpasquier, line 94: (nit) can be a const

pkg/operator/resource_reconciler.go
marioferh: done

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +502,22 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
...
+// controllerID is "" and there's no annotation set or the annotation value is empty.
+func (rr *ResourceReconciler) hasOwnership(obj metav1.Object) bool {
+	var controllerID string
+
+	if obj.GetAnnotations() != nil {
+		controllerID = obj.GetAnnotations()[controllerIDLabel]
+	}
+
+	if controllerID != rr.controllerID {
+		level.Debug(rr.logger).Log("msg", "skipping object not managed by the controller", "object", fmt.Sprintf("%s/%s", obj.GetNamespace(), obj.GetName()), "object_controller_id", controllerID, "controller_id", rr.controllerID)
```
simonpasquier, line 518: (nit)
```suggestion
		level.Debug(rr.logger).Log("msg", "skipping object not managed by the controller", "object", fmt.Sprintf("%s/%s", obj.GetNamespace(), obj.GetName()), "object_id", controllerID, "controller_id", rr.controllerID)
```

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +502,22 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
 	}
 	return ns, nil
 }
+
+// hasOwnership returns true if the controller is the "owner" of the object.
+// Whether it's owner is determined by the value of 'controllerID'
+// annotation. If the value matches the controllerID then it owns it, or if the
+// controllerID is "" and there's no annotation set or the annotation value is empty.
+func (rr *ResourceReconciler) hasOwnership(obj metav1.Object) bool {
```
simonpasquier, line 510: (nit)
```suggestion
func (rr *ResourceReconciler) isManagedByController(obj metav1.Object) bool {
```

pkg/operator/resource_reconciler.go
```diff
@@ -357,11 +388,25 @@ func (rr *ResourceReconciler) onStatefulSetDelete(ss *appsv1.StatefulSet) {
 
 // EnqueueForReconciliation asks for reconciling the object.
 func (rr *ResourceReconciler) EnqueueForReconciliation(obj metav1.Object) {
+	level.Debug(rr.logger).Log(
+		"msg", "mariofar onadd",
+		"object", fmt.Sprintf("%s/%s", obj.GetNamespace(), obj.GetName()),
+		"object", fmt.Sprintf("%s/%s", obj.GetGenerateName(), rr.resourceKind),
+	)
```
simonpasquier, line 395: should be removed?

cmd/operator/main.go
```diff
@@ -139,6 +139,7 @@ func parseFlags(fs *flag.FlagSet) {
 	fs.StringVar(&cfg.AlertmanagerDefaultBaseImage, "alertmanager-default-base-image", operator.DefaultAlertmanagerBaseImage, "Alertmanager default base image (path without tag/version)")
 	fs.StringVar(&cfg.PrometheusDefaultBaseImage, "prometheus-default-base-image", operator.DefaultPrometheusBaseImage, "Prometheus default base image (path without tag/version)")
 	fs.StringVar(&cfg.ThanosDefaultBaseImage, "thanos-default-base-image", operator.DefaultThanosBaseImage, "Thanos default base image (path without tag/version)")
+	fs.StringVar(&cfg.ControllerID, "controller-id", "", "Field for handling multiple operators accessing the same resource")
```
simonpasquier, line 142:
```suggestion
	fs.StringVar(&cfg.ControllerID, "controller-id", "", "Value used by the operator to filter Alertmanager, Prometheus, PrometheusAgent and ThanosRuler objects that it should reconcile. If the value isn't empty, the operator only reconciles objects with an `operator.prometheus.io/controller-id` annotation of the same value. Otherwise the operator reconciles all objects without the annotation or with an empty annotation value.")
```

pkg/operator/resource_reconciler.go
```diff
@@ -85,19 +85,26 @@ type ResourceReconciler struct {
...
+
+	controllerID string
 }
 
 var (
 	_ = cache.ResourceEventHandler(&ResourceReconciler{})
 )
 
+const (
+	controllerIDLabel = "operator.prometheus.io/controller-id"
```
simonpasquier, line 97:
```suggestion
	controllerIDAnnotation = "operator.prometheus.io/controller-id"
```

simonpasquier (commented):
A few recommendations regarding e2e tests:
* start with 1 CRD (e.g. Prometheus) and verify that the test fails if `isManagedByController()` is buggy (e.g. always return true) then make the test pass. After that we can generalize to the other CRDs.
* the test scenario should exercise the original issue and verify that 2 operators can watching the same namespaces can run concurrently when they use different controller-id values.
* The sub-test can live in `TestAllNS()` which already deploys an operator watching all namespaces.
  * deploy another prometheus operator also watching for all namespaces but with the controller-id being set (can use `AdditionalArgs`  in https://github.com/prometheus-operator/prometheus-operator/blob/9faf3b2c5fed4fcc11c7a0f2e5af1f950ec3861f/test/framework/framework.go#L197-L207).
  * create 1 Prometheus object with the controller-id set and wait for it to be ready.
  * create 1 Prometheus object without controller-id and wait for it to be ready.

test/e2e/controllerid_test.go
```diff
@@ -0,0 +1,149 @@
...
+	defer testCtx.Cleanup(t)
+	ns := framework.CreateNamespace(context.Background(), t, testCtx)
+	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, ns)
+
+	name := "test"
+	p := framework.MakeBasicPrometheus(ns, name, name, 1)
+	p.Annotations["operator.prometheus.io/controller-id"] = "88"
+
+	_, err := framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, p)
+	if err == nil {
```
simonpasquier, line 34: Each test will have to wait for 5m which would make the tests duration longer than needed.
If we have a successful test running 2 operators in parallel managing separate objects (e.g. "testControllerMultipleOperators"), it will assert the same behavior.

test/framework/framework.go
```diff
@@ -216,6 +216,7 @@ func (f *Framework) CreateOrUpdatePrometheusOperator(
 	createResourceAdmissionHooks,
 	createClusterRoleBindings,
 	createScrapeConfigCrd bool,
+	additionalArgs []string,
```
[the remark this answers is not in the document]
marioferh: you are right, fix it

test/e2e/controllerid_test.go
marioferh: right

test/e2e/main_test.go
```diff
@@ -412,6 +412,59 @@ func TestOperatorUpgrade(t *testing.T) {
...
+	// Create operator-2 in a new ns and set controller-id.
+	addArgs := []string{"--controller-id=42"}
+	finalizers, err = framework.CreateOrUpdatePrometheusOperatorWithOpts(context.Background(),
+		operatorFramework.PrometheusOperatorOpts{
+			Namespace:              ns,
+			AllowedNamespaces:      nil,
+			DeniedNamespaces:       nil,
+			PrometheusNamespaces:   nil,
+			AlertmanagerNamespaces: nil,
+			EnableAdmissionWebhook: true,
```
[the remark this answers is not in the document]
marioferh: done, thanks

test/e2e/main_test.go
```diff
@@ -412,6 +412,59 @@ func TestOperatorUpgrade(t *testing.T) {
...
+		operatorFramework.PrometheusOperatorOpts{
+			Namespace:              ns,
+			AllowedNamespaces:      nil,
+			DeniedNamespaces:       nil,
+			PrometheusNamespaces:   nil,
+			AlertmanagerNamespaces: nil,
+			EnableAdmissionWebhook: false,
+			ClusterRoleBindings:    true,
+			EnableScrapeConfigs:    true,
+			AdditionalArgs:         addArgs,
```
simonpasquier, line 447: (nit) we don't even need the intermediate variable
```suggestion
			AdditionalArgs:         []string{"--controller-id=42"},
```

test/e2e/main_test.go
```diff
@@ -412,6 +412,59 @@ func TestOperatorUpgrade(t *testing.T) {
...
+
+	for _, f := range finalizers {
+		testCtx.AddFinalizerFn(f)
+	}
+
+	testFuncs := map[string]func(t *testing.T){
+		"MultipleOperatorsPrometheusServer": testMultipleOperatorsPrometheusServer,
+		"MultipleOperatorsPrometheusAgent":  testMultipleOperatorsPrometheusAgent,
+		"MultipleOperatorsAlertManager":     testMultipleOperatorsAlertManager,
+		"MultipleOperatorsThanosRuler":      testMultipleOperatorsThanosRuler,
```
simonpasquier, line 461: (nit) we can shorten a bit
```suggestion
		"PrometheusServer": testMultipleOperatorsPrometheusServer,
		"PrometheusAgent":  testMultipleOperatorsPrometheusAgent,
		"Alertmanager":     testMultipleOperatorsAlertManager,
		"ThanosRuler":      testMultipleOperatorsThanosRuler,
```

test/e2e/main_test.go
```diff
@@ -412,6 +412,59 @@ func TestOperatorUpgrade(t *testing.T) {
 	}
 }
 
+// TestIsManagedByController test prometheus operator managing object with correct ControlerID.
+func TestIsManagedByController(t *testing.T) {
```
simonpasquier, line 416:
```suggestion
func TestMultipleOperators(t *testing.T) {
```

test/e2e/controllerid_test.go
```diff
@@ -0,0 +1,108 @@
...
+	name := "test-op-1"
+	p := framework.MakeBasicPrometheus(ns, name, name, 1)
+	_, err := framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, p)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	name = "test-op-2"
+	p = framework.MakeBasicPrometheus(ns, name, name, 1)
+	p.Annotations["operator.prometheus.io/controller-id"] = "42"
```
simonpasquier, line 37: (suggestion) let's create the 2 objects simultaneously, they will be reconciled in parallel and the test should run faster. Same comment for all other tests.

test/e2e/main_test.go
marioferh: done

test/e2e/main_test.go
marioferh: done

test/e2e/main_test.go
marioferh: done

test/e2e/controllerid_test.go
marioferh: how can I do it simultaneously? thanks

test/e2e/main_test.go
```diff
@@ -199,6 +199,7 @@ func TestAllNS(t *testing.T) {
 	t.Run("x", testAllNSAlertmanager)
 	t.Run("y", testAllNSPrometheus)
 	t.Run("z", testAllNSThanosRuler)
+	t.Run("TestIsManagerByController", testMultipleOperators(testCtx))
```
simonpasquier, line 202:
```suggestion
	t.Run("multipleOperators", testMultipleOperators(testCtx))
```

test/e2e/main_test.go
```diff
@@ -452,3 +453,42 @@ func testServerTLS(ctx context.Context, namespace string) func(t *testing.T) {
 		}
 	}
 }
+
+// TestIsManagedByController test prometheus operator managing object with correct ControlerID.
+func testMultipleOperators(testCtx *operatorFramework.TestCtx) func(t *testing.T) {
+	return func(t *testing.T) {
+		skipPrometheusTests(t)
```
simonpasquier, line 460: rather this?
```suggestion
		skipPrometheusAllNSTests(t)
```

test/e2e/main_test.go
```diff
@@ -452,3 +453,42 @@ func testServerTLS(ctx context.Context, namespace string) func(t *testing.T) {
...
+		ns := framework.CreateNamespace(context.Background(), t, testCtx)
+		// Create operator-2 in a new ns and set controller-id.
+		finalizers, err := framework.CreateOrUpdatePrometheusOperatorWithOpts(context.Background(),
+			operatorFramework.PrometheusOperatorOpts{
+				Namespace:              ns,
+				AllowedNamespaces:      nil,
+				DeniedNamespaces:       nil,
+				PrometheusNamespaces:   nil,
+				AlertmanagerNamespaces: nil,
+				EnableAdmissionWebhook: false,
```
simonpasquier, line 471: could be removed since they are zero-values
```suggestion
```

test/e2e/controllerid_test.go
simonpasquier: I meant
1. create test-op-1
2. create test-op-2
3. wait for test-op-1 to be ready
4. wait for test-op-2 to be ready

Hopefully when we reach 4, the check shouldn't take long. But it requires to use `Create` +  `WaitForPrometheusReady` instead of  `CreatePrometheusAndWaitUntilReady`.

simonpasquier, line 37: can you declare a const for the controller id so we don't have to duplicate?

simonpasquier (commented):
Can you try adding t.Parallel() to all  `testMultipleOperators`* tests?

test/e2e/controllerid_test.go
marioferh: With t.Parallel() failed because firs operator is deleted

marioferh: done

test/e2e/main_test.go
```diff
@@ -452,3 +453,37 @@ func testServerTLS(ctx context.Context, namespace string) func(t *testing.T) {
...
+		skipPrometheusAllNSTests(t)
+
+		ns := framework.CreateNamespace(context.Background(), t, testCtx)
+		// Create operator-2 in a new ns and set controller-id.
+		finalizers, err := framework.CreateOrUpdatePrometheusOperatorWithOpts(context.Background(),
+			operatorFramework.PrometheusOperatorOpts{
+				Namespace:           ns,
+				ClusterRoleBindings: true,
+				EnableScrapeConfigs: true,
+				AdditionalArgs:      []string{"--controller-id=42"},
```
simonpasquier, line 469: can you use `testControllerID` instead?

simonpasquier (commented):
Just to make sure that the e2e test assumptions are valid, can you push a commit that modifies the `controllerIDAnnotation` const in pkg/operator/resource_reconciler.go so that we assert that the tests are failing if the controller-id logic is broken? Once we know that the tests fail, you can revert it.

test/e2e/main_test.go
marioferh: sure, let's check it

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +496,24 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
...
+func (rr *ResourceReconciler) isManagedByController(obj metav1.Object) bool {
+	var controllerID string
+
+	if obj.GetAnnotations() != nil {
+		controllerID = obj.GetAnnotations()[controllerIDAnnotation]
+	}
+
+	fmt.Println("mariofar annotations controllerID: ", controllerID)
+	fmt.Println("mariofar rr controllerID: ", rr.controllerID)
+
```
simonpasquier, line 512:
```suggestion
```

simonpasquier (approved):

danielmellado:
@simonpasquier not sure about the squash status after enable/disable but I'd prefer for all these commits to be squashed previous merging

marioferh:
> @simonpasquier not sure about the squash status after enable/disable but I'd prefer for all these commits to be squashed previous merging

sure I can squash them all before merge

test/e2e/main_test.go
marioferh: Ok, now the issue is, after change: 
`controllerIDAnnotation = "operator.prometheus.io/wrong-label"`
controllerID of the object is empty and controllerID of controller if empty, then object is created :S

How could we fix it?

simonpasquier:
I can squash on merge to avoid another run of the CI.

test/e2e/main_test.go
simonpasquier: :facepalm: of course, sorry for my wrong suggestion. So I don't thing we need to be paranoid. Please revert that part and we can merge.

pkg/operator/resource_reconciler.go
```diff
@@ -85,19 +85,26 @@ type ResourceReconciler struct {
...
+	controllerID string
 }
 
 var (
 	_ = cache.ResourceEventHandler(&ResourceReconciler{})
 )
 
+const (
+	controllerIDAnnotation = "operator.prometheus.io/wrong-label"
```
simonpasquier, line 97: to be reverted

marioferh:
> I can squash on merge to avoid another run of the CI.

Yep, I have done one that should fail with different ID and after that we can merged. Thanks Simon

simonpasquier (approved):
Thanks!

pkg/operator/resource_reconciler.go
```diff
@@ -457,3 +496,21 @@ func ListMatchingNamespaces(selector labels.Selector, nsInf cache.SharedIndexInf
 	}
 	return ns, nil
 }
+
+// isManagedByController returns true if the controller is the "owner" of the object.
```
machine424, line 500: maybe change owner->manager/own->manage

pkg/operator/config.go
```diff
@@ -59,6 +59,9 @@ type Config struct {
 	AlertmanagerSelector    LabelSelector
 	ThanosRulerSelector     LabelSelector
 	SecretListWatchSelector FieldSelector
+
+	// Controller id for pod ownership
```
machine424, line 63: This comment is stale/misleading I think we can just remove it.

Documentation/operator.md
```diff
@@ -43,6 +43,8 @@ Usage of ./operator:
     	Config Reloader memory limits. Value "0" disables it and causes no limit to be configured. (default 50Mi)
   -config-reloader-memory-request value
     	Config Reloader memory requests. Value "0" disables it and causes no request to be configured. (default 50Mi)
+  -controller-id operator.prometheus.io/controller-id
```
machine424, line 46: nit: Was it intentional to add `operator.prometheus.io/controller-id` in here? I think `string` + the description are sufficient?

machine424 (commented):
just some nits.
Well done guys :)

PR #6341, opened 2024-02-28
WIP: Split Select*() functions

### LINKED ISSUE #5420: Investigate and split SelectServiceMonitors(and related) function if possible
Just making sure we don't forget :)

This issue is about investigating if we improve readability of those functions

              I feel like breaking this method into two separate ones would make more sense, making sure they do what they are named after 😅

With that said, I think this goes beyond what this PR is intended to change, so I'll just add the comment for now :)

_Originally posted by @ArthurSens in https://github.com/prometheus-operator/prometheus-operator/pull/5392#discussion_r1138554606_

haanhvu:
## Description

The Select*() functions in resource selector are currently too long and do more than just select. This splits those functions into smaller functions for readability.

Fixes #5420

## Type of change

## Verification

No tests broken

## Changelog entry

Split Select*() functions for readability

### DIFF
--- a/pkg/prometheus/resource_selector.go
+++ b/pkg/prometheus/resource_selector.go
@@ -73,14 +73,23 @@ func NewResourceSelector(l log.Logger, p monitoringv1.PrometheusInterface, store
 func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn ListAllByNamespaceFn) (map[string]*monitoringv1.ServiceMonitor, error) {
 	cpf := rs.p.GetCommonPrometheusFields()
 	objMeta := rs.p.GetObjectMeta()
-	namespaces := []string{}
-	// Selectors (<namespace>/<name>) might overlap. Deduplicate them along the keyFunc.
-	serviceMonitors := make(map[string]*monitoringv1.ServiceMonitor)
-
-	servMonSelector, err := metav1.LabelSelectorAsSelector(cpf.ServiceMonitorSelector)
+	namespaces, err := rs.getNamespaces(cpf, objMeta)
 	if err != nil {
 		return nil, err
 	}
+
+	serviceMonitors, err := rs.getServiceMonitorsFromNameSpaces(namespaces, listFn, cpf, objMeta)
+	if err != nil {
+		return nil, err
+	}
+
+	res := rs.filterServiceMonitors(ctx, serviceMonitors, cpf, objMeta)
+
+	return res, nil
+}
+
+func (rs *ResourceSelector) getNamespaces(cpf monitoringv1.CommonPrometheusFields, objMeta metav1.Object) ([]string, error) {
+	namespaces := []string{}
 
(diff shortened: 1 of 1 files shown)


haanhvu:
@ArthurSens @simonpasquier I started with splitting `SelectServiceMonitors()`. Could you take a look to see if this split makes sense? If yes I'll continue with other Select function.

simonpasquier (commented):
I'm not really convinced that the decomposition of the Select* functions into 2 smaller sub-functions helps with readability/maintainability. IMHO more functions makes it harder to follow. 
It might be also because I'm used to the existing code base.

ArthurSens:
Appreciate the initiative!

I think we're going in the wrong direction here, I agree with Simon that this approach doesn't help much and looking at the code it might not be as easy as it looks 🤔 

I was thinking that we could extract some functionality from the Select.* functions and make generic enough that can be re-used by ServiceMonitor, PodMonitor and other related resources.

For example, the code below:

https://github.com/prometheus-operator/prometheus-operator/blob/f8c181b974cb83c92308b732ed24e9b75c4f9171/pkg/prometheus/agent/operator.go#L700-L713

Would become something like this:

```go
	smons, err := resourceSelector.SelectServiceMonitors(ctx, c.smonInfs.ListAllByNamespace)
	if err != nil {
		return fmt.Errorf("selecting ServiceMonitors failed: %w", err)
	}

	pmons, err := resourceSelector.SelectPodMonitors(ctx, c.pmonInfs.ListAllByNamespace)
	if err != nil {
		return fmt.Errorf("selecting PodMonitors failed: %w", err)
	}

	bmons, err := resourceSelector.SelectProbes(ctx, c.probeInfs.ListAllByNamespace)
	if err != nil {
		return fmt.Errorf("selecting Probes failed: %w", err)
	}

	resourceSelector.RemoveInvalidResources(smons, pmons, bmons)
	resourceSelector.AddAssetsToStore(smons, pmons, bmons)
```

or maybe even

```go
	smons, pmons, bmons, err := resourceSelector.SelectServiceMonitors(ctx, c.smonInfs.ListAllByNamespace)
	if err != nil {
		return fmt.Errorf("selecting resources failed: %w", err)
	}
	
	resourceSelector.RemoveInvalidResources(smons, pmons, bmons)
	resourceSelector.AddAssetsToStore(smons, pmons, bmons)
```

The problem I see today is that `Select...` does more than just select resources. I think code is more readable is function name correctly represent what they do

haanhvu:
Yeah I see what you mean now. I'll investigate more to see if this's possible.

simonpasquier:
> The problem I see today is that Select... does more than just select resources.

I don't agree 100%: they do select *and filter* resources. I'm also not convinced that `resourceSelector.RemoveInvalidResources(smons, pmons, bmons)` is simpler to maintain over time.

ArthurSens:
fair point! If we don't have consensus on what to do, let's leave it that way :)

Issue #6310, opened 2024-02-15
config-reloader takes several minutes to reload successfully after an invalid additionalScrapeConfigs is fixed


FloMedja:
### Is there an existing issue for this?

- [X] I have searched the existing issues

## Description

when a user specify a `additionalScrapeConfigs` in the prometheus configuration, If the initial configuration is invalid the `config-reloader` didn't reload when after the configuration is fixed. the config-reloader get stucked.

## Steps to Reproduce

-  create a secret that contains an invalid scrape config. The configuration need to be properly formatted `yaml` but not loadable by prometheus. This is an exemple of an invalid scrape config:

```yaml
- job_name: opencost
  honor_labels: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  dns_sd_configs:
    - names:
        - opencost-cost-analyzer
  type: A     # <----------------  we have a  bad indentation here
  port: 9003 # <---------------  we have a  bad indentation here
```
-  create a prometheus configuration with an `additionalScrapeConfigs` spec pointing to the previously created secret
-  the logs of the `config-reloader` shows that the configuration is invalid and can't be applied
-  fix the invalid scrape config. This is the fix of the previous exemple of scrape config:
```yaml
- job_name: opencost
  honor_labels: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  dns_sd_configs:
    - names:
        - opencost-cost-analyzer
      type: A     # <---------------- indentation fixed
      port: 9003 # <---------------   indentation fixed
```
-  the logs of the `config-reloader` are still showing that the configuration is invalid and can't be applied
- 
## Expected Result

The `config-reloader` load the valid scrape configs after a invalid scrape configs is initially pushed

## Actual Result

the `config-reloader` is stucked after a invalid scrape config is initially pushed, even after this invalid scrape configs is fixed

### Prometheus Operator Version

```shell
v0.70.0
```

### Kubernetes Version

```yaml
v1.26.8
```

### Kubernetes Cluster Type

minikube

### How did you deploy Prometheus-Operator?

helm chart:prometheus-community/kube-prometheus-stack

### prometheus-operator log output

```shell
the prometheus-operator didn't log any particular nessage related to the issue.
```

### Anything else?

Manually trigger the reloading refresh the `config-reloader`  and the scrape config is updated

[...]

PR #6348, opened 2024-02-29
fix: monitoring in config reloader (#6310)

yp969803:
## Description

fixes: #6310 
config-reloader does not successfully reload after an invalid additionalScrapeConfigs is fixed

## Type of change

## Verification
make test

## Changelog entry

```
enabled monitoring in config reloader
```


yp969803:
@simonpasquier can you review the pr

simonpasquier:
I don't understand how it's going to fix the issue: the init container should only run once and exit.

yp969803:
Should'nt it be monitoring any changes in config map in prometheus? Why there is ReloadOnce option if the the container can only be created once

simonpasquier:
the reloader config init reloader synthesizes the on-disk configuration before the Prometheus container starts which is why it runs only once.

yp969803:
@simonpasquier got it, finding some another solution for the issue

Issue #6360, opened 2024-03-02
Add `ttl` to `alertmanagerconfigs` resource


onedr0p:
### Component(s)

AlertManagerConfig

### What is missing? Please describe.

A new field was added to the `pushover_configs` called `ttl`, it would be nice if the operator supported this new field

Refs:

- https://github.com/prometheus/alertmanager/pull/3474
- https://prometheus.io/docs/alerting/latest/configuration/#pushover_config

## Environment
Kubernetes Version: v1.29.2
Prometheus-Operator Version: 0.71.2

slashpai:
IIUC this needs alertmanager version 0.27.0

slashpai:
Do you want to contribute to the change?

aerosouund:
Hello @slashpai 
I would be interested in taking this, in case @onedr0p isn't already working on it

onedr0p:
Go for it, thanks @aerosouund !

slashpai:
@aerosouund sure go ahead :)

slashpai:
Here is a example PR https://github.com/prometheus-operator/prometheus-operator/pull/5886 how to add fields in alertmanagerconfigs

- [ ] Add to https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/alertmanager/types.go
- [ ] Add to https://github.com/prometheus-operator/prometheus-operator/tree/main/pkg/apis/monitoring/v1alpha1 and https://github.com/prometheus-operator/prometheus-operator/tree/main/pkg/apis/monitoring/v1beta1

Incase if you have any questions please feel free to ask

aerosouund:
Hello @slashpai 

I have created a PR that adds the field, although i'd like to ask you where to add the functionality that uses this field.
i didn't find that in the prometheus operator there's an implementation of the `Notify` interface.

where do you think this functionality belongs ? or where do i start looking ? 
Thanks in advance

onedr0p:
@aerosouund Alertmanager will be able to pick up the config and use, see the PR that added the functionality to AM here

https://github.com/prometheus/alertmanager/pull/3474

slashpai:
You need to add the code to include this change in pkg/alertmanager/amcfg.go, unit test to pkg/alertmanager/amcfg_test.go

See the example PR I linked in previous comment

afzal442:
hi @aerosouund just check in! 😄 Are you still following up with your PR? I wanted to take this up.

[...]

PR #6371, opened 2024-03-06
Alertmanager ttl

aerosouund:
## Description

Add TTL field to alertmanager config
solves: https://github.com/prometheus-operator/prometheus-operator/issues/6360

## Type of change

## Verification

## Changelog entry

```release-note
add ttl to alertmanagerConfig
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/prometheus-operator/prometheus-operator
 
-go 1.22
+go 1.22.0
 
 require (
 	github.com/alecthomas/kingpin/v2 v2.4.0
--- a/pkg/alertmanager/types.go
+++ b/pkg/alertmanager/types.go
@@ -29,6 +29,7 @@ import (
 // https://github.com/prometheus/alertmanager/issues/1985
 type alertmanagerConfig struct {
 	Global            *globalConfig   `yaml:"global,omitempty" json:"global,omitempty"`
+	TTL 			  duration        `yaml:"ttl,omitempty" json:"ttl,omitempty"`
 	Route             *route          `yaml:"route,omitempty" json:"route,omitempty"`
 	InhibitRules      []*inhibitRule  `yaml:"inhibit_rules,omitempty" json:"inhibit_rules,omitempty"`
 	Receivers         []*receiver     `yaml:"receivers,omitempty" json:"receivers,omitempty"`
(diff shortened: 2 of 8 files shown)


yp969803:
@aerosouund  you have added the field that is good,
Some reviews:
-  Run "go fmt -w ." after the completion of code
- Run "make generate"
- Merge all the commit into one commit which should follow a certain pattern you can see previous commits

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus-operator/prometheus-operator
 
-go 1.22
+go 1.22.0
```
ArthurSens, line 3: Was this change intentional?

pkg/alertmanager/types.go
```diff
@@ -29,6 +29,7 @@ import (
 // https://github.com/prometheus/alertmanager/issues/1985
 type alertmanagerConfig struct {
 	Global            *globalConfig   `yaml:"global,omitempty" json:"global,omitempty"`
+	TTL 			  duration        `yaml:"ttl,omitempty" json:"ttl,omitempty"`
```
ArthurSens, line 32: We have a linting problem here, you can fix it by running `make format` :)

pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go
```diff
@@ -783,6 +784,9 @@ type PushoverConfig struct {
 	// Notification title.
 	// +optional
 	Title string `json:"title,omitempty"`
+	// time to live for the alert notification
+	// +optional
+	TTL time.Duration `json:"ttl,omitempty"`
```
ArthurSens, line 789:
```suggestion
	TTL *Duration `json:"ttl,omitempty"`
```

We have a [custom Duration type](https://github.com/prometheus-operator/prometheus-operator/blob/6e94864406d265774c1572b187e939f3ea5b8a3e/pkg/apis/monitoring/v1/types.go#L41), which has some useful validations on the API level. Could we use this one instead?

Also, we follow k8s API conventions as much as possible, which means we want to use [pointers for optional fields](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required).

ArthurSens (commented):
Thanks for starting the PR!

The changes made to the API are almost there, we also want to incorporate the changes you've made to the API to the alertmanager packages, i.e. we want to write this configuration in Alertmanager's config files if it is set.

@slashpai shared some directions here: https://github.com/prometheus-operator/prometheus-operator/issues/6360#issuecomment-1983838148

PR #6377, opened 2024-03-07
WIP: Add a flag to set node address priority

### LINKED ISSUE #3247: Node address priority can lead to unreachable targets.
**What is missing?**
Currently prometheus operator uses following logic to determine ip address for a node (which will be used later on for target discovery like kubelet or node-exporter)

```go
func nodeAddress(node v1.Node) (string, map[v1.NodeAddressType][]string, error) {
	m := map[v1.NodeAddressType][]string{}
	for _, a := range node.Status.Addresses {
		m[a.Type] = append(m[a.Type], a.Address)
	}

	if addresses, ok := m[v1.NodeInternalIP]; ok {
		return addresses[0], m, nil
	}
	if addresses, ok := m[v1.NodeExternalIP]; ok {
		return addresses[0], m, nil
	}
	return "", m, fmt.Errorf("host address unknown")
}
```
In my environment I have some nodes which are behind the nat, so they have NodeInternalIP set to a private range, and public IP set as NodeExternalIP. 

Those internal ips are not routable from the rest of the cluster. So when they come up in the target list for kubeadm and node exporters, prometheus cannot scrape them because the target is not routable. 

Ideally we should support for a node to be able to give more priority for their external address (for example through annotation)

**Why do we need it?**

Without such fix it's impossible to scrape nodes which have unroutable internal ip (but they are part of the cluster and connected through flannel for example).

haanhvu:
Currently internal node addresses are prioritized over external addresses by default. This adds a flag to allow users to set node address priority (internal/external). This is helpful for use cases where node internal addresses exist but are not routable.

Fixes #3247

I'm adding tests now. Will remove the WIP title when tests are fully added. Pushed implementation here for review.


cmd/operator/main.go
```diff
@@ -108,6 +108,8 @@ var (
 	// Parameters for the kubelet endpoints controller.
 	kubeletObject   string
 	kubeletSelector operator.LabelSelector
+
+	nodeAddressPriority string
```
simonpasquier, line 112: I'd recommend using a custom type instead of string (e.g. `type NodeAddresPriority string`) and implement the flag.Set interface for it. This way we can validate the flag value before starting everything.

cmd/operator/main.go
```diff
@@ -125,6 +126,7 @@ func parseFlags(fs *flag.FlagSet) {
 
 	fs.StringVar(&kubeletObject, "kubelet-service", "", "Service/Endpoints object to write kubelets into in format \"namespace/name\"")
 	fs.Var(&kubeletSelector, "kubelet-selector", "Label selector to filter nodes.")
+	fs.Var(&nodeAddressPriority, "node-address-priority", "Node address priority. Either internal or external.")
```
simonpasquier, line 129: maybe prefix with `kubelet-` to make it obvious that it affects the kubelet controller config.
```suggestion
	fs.Var(&nodeAddressPriority, "kubelet-node-address-priority", "Node address priority. Either 'internal' or 'external'. Default: 'internal'.")
```

pkg/operator/config.go
```diff
@@ -276,6 +276,24 @@ func (ls *LabelSelector) Set(value string) error {
 	return nil
 }
 
+type NodeAddressPriority string
+
+// String implements the flag.Value interface.
+func (p *NodeAddressPriority) String() string {
+	if p == nil {
```
simonpasquier, line 283:
```suggestion
	if p == nil || *p == "" {
```

cmd/operator/main.go
```diff
@@ -125,6 +126,7 @@ func parseFlags(fs *flag.FlagSet) {
 
 	fs.StringVar(&kubeletObject, "kubelet-service", "", "Service/Endpoints object to write kubelets into in format \"namespace/name\"")
 	fs.Var(&kubeletSelector, "kubelet-selector", "Label selector to filter nodes.")
+	fs.Var(&nodeAddressPriority, "kubelet-node-address-priority", "Node address priority used by kubelet. Either 'internal' or 'external'. Default: 'internal'")
```
simonpasquier, line 129: (nit)
```suggestion
	fs.Var(&nodeAddressPriority, "kubelet-node-address-priority", "Node address priority used by kubelet. Either 'internal' or 'external'. Default: 'internal'.")
```

simonpasquier (commented):
Could you add some tests in `pkg/kubelet/controller_test.go` and verify that getNodeAddresses() returns the correct result depending on internal/external setting?

mviswanathsai:
You can run `make --always-make format generate && git diff --exit-code` to fix the formatting issues.

haanhvu:
> You can run `make --always-make format generate && git diff --exit-code` to fix the formatting issues.

I'll do that later when code & tests are complete. This PR's still in WIP.

haanhvu:
@simonpasquier Sorry for taking some time to push (I didn't have resources to push in the last 2 weeks).

Pls help review if there're any problems with the tests, or we need more tests.

simonpasquier (approved):
Thanks!

PR #6379, opened 2024-03-08
Add extra relabelings to scrape classes

QuentinBisson:
Description

Fixes https://github.com/prometheus-operator/prometheus-operator/issues/3255

This PR allows users of the operator to ensure a set of custom relabelings to all their pod/service monitors.

This superseds this PR https://github.com/prometheus-operator/prometheus-operator/pull/5978

Notable examples would be adding the node label using __meta_kubernetes_pod_node_name but it could be used for other organization purposes such as:
```
- sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
  action: "replace"
  targetLabel: app
```

## Type of change

## Verification

## Changelog entry

```release-note
Allow addition of extra relabelings to all scrape classes
```


QuentinBisson:
@nicolastakashi and @ArthurSens now that scrape classes are finally out, could you check this one? :)

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1795,4 +1795,10 @@ type ScrapeClass struct {
 	// TLSConfig section for scrapes.
 	// +optional
 	TLSConfig *TLSConfig `json:"tlsConfig,omitempty"`
+
+	// ExtraRelabelings allow setting extra default relabelings to all
```
simonpasquier, line 1799: It should be obvious that it applies to the target relabeling rules and when the scrape class' relabelings are applied compared to the monitor object's relabelings.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1795,4 +1795,10 @@ type ScrapeClass struct {
...
+	// scrape objects.
+	//
+	// +optional
+	ExtraRelabelings []*RelabelConfig `json:"extraRelabelings,omitempty"`
```
simonpasquier, line 1803: I'm not sure why we went with []*RelabelConfig in the first place for service/pod monitors but I'd rather use []RelabelConfig.

simonpasquier (commented):
Thanks for kicking this!

pkg/apis/monitoring/v1/prometheus_types.go
ArthurSens, line 1803:
```suggestion
	Relabelings []*RelabelConfig `json:"relabelings,omitempty"`
```

I'm still not sure about my suggestion, but I think "Extra" might be redundant with the concept of scrape classes. What do you think?

pkg/prometheus/promcfg.go
```diff
@@ -421,6 +421,21 @@ func (cg *ConfigGenerator) MergeTLSConfigWithScrapeClass(tlsConfig *monitoringv1
...
+
+	if scrapeClass != nil {
+		scrapeClassName = scrapeClass.Name
+	}
+
+	scrapeClass, found := cg.scrapeClasses[scrapeClassName]
+
+	if !found {
+		return nil
+	}
```
ArthurSens, line 435: Non-blocking!

This part is duplicated with the implementation of `MergeTLSConfigWithScrapeClass` and will probably repeat itself for every other field we add to ScrapeClasses. It would be nice if we could find a way of removing this duplication 🤔

pkg/prometheus/promcfg.go
```diff
@@ -1020,6 +1035,12 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 		})
 	}
 
+	// Add user extra relabelings if specified.
+	extraRelabelings := cg.GetScrapeClassExtraRelabelings(scrapeClass)
+	if len(extraRelabelings) > 0 {
+		relabelings = append(relabelings, generateRelabelConfig(extraRelabelings)...)
+	}
+
```
ArthurSens, line 1043: I wonder if we should log or warn through admission webhooks if we have clashing relabeling rules in the ScrapeClass and the underlying resource.

I guess we can start with a log and iterate with validating webhooks later?

ArthurSens (commented):
Thanks for getting back to this!

I've made a few comments but I think we're pretty close already :)

pkg/apis/monitoring/v1/prometheus_types.go
QuentinBisson: The issue with this is that the generator function requires an array of pointers:
`func generateRelabelConfig(rc []*monitoringv1.RelabelConfig) []yaml.MapSlice`

QuentinBisson: That's fine with me :)

pkg/prometheus/promcfg.go
QuentinBisson: So I just noticed that we use this https://github.com/prometheus-operator/prometheus-operator/blob/e8d020e1d23a966432584839c3ac98aac9e18b42/pkg/prometheus/promcfg.go#L837 at the top of all configured resources and it returns nil if there is no scrape class so I simplified this PR's code and maybe the same should be done for the existing fields

pkg/prometheus/promcfg.go
QuentinBisson: I think this might then become an issue for use case where it actually make sense to override some custom fields though (not that I have one in mind)

pkg/prometheus/promcfg.go
```diff
@@ -421,6 +421,21 @@ func (cg *ConfigGenerator) MergeTLSConfigWithScrapeClass(tlsConfig *monitoringv1
 	return tlsConfig
 }
 
+func (cg *ConfigGenerator) GetScrapeClassExtraRelabelings(scrapeClass *monitoringv1.ScrapeClass) []*monitoringv1.RelabelConfig {
```
nicolastakashi, line 424:
```suggestion
func (cg *ConfigGenerator) GetScrapeClassRelabelings(scrapeClass *monitoringv1.ScrapeClass) []*monitoringv1.RelabelConfig {
```

pkg/apis/monitoring/v1/prometheus_types.go
simonpasquier: Right, let's keep it like that for now and tackle my remark in another PR.

pkg/prometheus/promcfg.go
simonpasquier: I agree that a "merging" strategy would be hard to get right. My expectation would be that the scrape class relabelings apply first then the service/pod monitor's relabelings. The scrape class relabelings may delete a discovered label that the monitor relabelings rely on but I feel like it's a corner case and not something that the operator could/should prevent.

simonpasquier (commented):
We need to call `validateRelabelConfigs()` on the scrape class relabelings and fail the reconciliation if invalid.

https://github.com/prometheus-operator/prometheus-operator/blob/4030eb75f6c76cf8c6ca0cf530e2ed9eb3c3c872/pkg/prometheus/resource_selector.go#L244-L256

pkg/prometheus/resource_selector.go
```diff
@@ -346,6 +346,11 @@ func validateScrapeClass(p monitoringv1.PrometheusInterface, sc *string) error {
 		if c.Name == *sc {
 			return nil
 		}
+		if c.Relabelings != nil {
+			if err := validateRelabelConfigs(p, c.Relabelings); err != nil {
+				return fmt.Errorf("scrapeClass %q relabelings: %w", *sc, err)
+			}
+		}
```
simonpasquier, line 353: Ideally it would need to happen before (e.g. at the beginning of the reconciliation) because it's a problem with the Prometheus/PrometheusAgent spec not with the monitoring object.

pkg/prometheus/resource_selector.go
```diff
@@ -469,7 +474,7 @@ func (rs *ResourceSelector) SelectPodMonitors(ctx context.Context, listFn ListAl
 		}
 
 		if err = validateScrapeClass(rs.p, pm.Spec.ScrapeClassName); err != nil {
-			rejectFn(pm, err)
+			rejectFn(pm, fmt.Errorf("scrapeClass: %w", err))
```
simonpasquier, line 477: I don't think that wrapping the error is required here.

pkg/prometheus/resource_selector.go
QuentinBisson: Oh makes sensé but I could not find where to put the validation. Could you give me a hint ?

pkg/prometheus/resource_selector.go
QuentinBisson: In that case should I remove it from thé servicemonitor validation ?

pkg/prometheus/resource_selector.go
QuentinBisson: I think I found a nice place in the `NewConfigGenerator` function. Can you check the new version?

pkg/prometheus/resource_selector.go
QuentinBisson: I removed it for all 4 places

pkg/apis/monitoring/v1beta1/validation.go
```diff
@@ -183,12 +183,12 @@ func (yr YearRange) Parse() (*ParsedRange, error) {
 
 	start, err := strconv.Atoi(startStr)
 	if err != nil {
-		fmt.Errorf("start year cannot be %s parsed: %w", startStr, err)
+		return nil, fmt.Errorf("start year cannot be %s parsed: %w", startStr, err)
```
QuentinBisson, line 186: I'm not sure why those were ignored, maybe I should not have fixed them in this PR

pkg/prometheus/promcfg.go
```diff
@@ -100,11 +100,17 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
 }
 
-func getScrapeClassConfig(cpf monitoringv1.CommonPrometheusFields) (map[string]*monitoringv1.ScrapeClass, string, error) {
+func getScrapeClassConfig(p monitoringv1.PrometheusInterface, cpf monitoringv1.CommonPrometheusFields) (map[string]*monitoringv1.ScrapeClass, string, error) {
 	scrapeClasses := make(map[string]*monitoringv1.ScrapeClass, len(cpf.ScrapeClasses))
 	defaultScrapeClass := ""
 	for _, scrapeClass := range cpf.ScrapeClasses {
 		scrapeClasses[scrapeClass.Name] = &scrapeClass
+		// Validate all scrape class relabelings are correct.
+		if scrapeClass.Relabelings != nil {
```
QuentinBisson, line 109: @simonpasquier This is where I thought it would be best but I'm not sure about the error message. Do you have any suggestions?

pkg/prometheus/resource_selector.go
```diff
@@ -120,7 +120,7 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
 	res := make(map[string]*monitoringv1.ServiceMonitor, len(serviceMonitors))
 	for namespaceAndName, sm := range serviceMonitors {
 		var err error
-		rejectFn := func(_ *monitoringv1.ServiceMonitor, err error) {
```
QuentinBisson, line 123: So I'm not sure why there this was changed to `_` as it was not when I initially did my first PR. It was named `servicemonitor` first and the issue is that the sm variable defined line 121 was used instead so I renamed it to `sm`. 

I'm not sure there is a reason to even have the servicemonitor paramater in the reject function as the code always use the loop value above but if there is, the function name should shadow the loop variable and if there is none, maybe the variable should just be removed in all resources (probes, podmonitors ...).

What do you think?

pkg/prometheus/resource_selector.go
```diff
@@ -192,7 +192,7 @@ func (rs *ResourceSelector) SelectServiceMonitors(ctx context.Context, listFn Li
 		}
 
 		if err = validateScrapeClass(rs.p, sm.Spec.ScrapeClassName); err != nil {
-			rejectFn(sm, fmt.Errorf("scrapeClass: %w", err))
+			rejectFn(sm, err)
```
QuentinBisson, line 195: I initially aligned the 3 other resources with this one but @simonpasquier told me the error wrap was not needed so let's align servicemonitor to the others instead

pkg/client/applyconfiguration/monitoring/v1/scrapeclass.go
```diff
@@ -53,3 +58,16 @@ func (b *ScrapeClassApplyConfiguration) WithTLSConfig(value *TLSConfigApplyConfi
 	b.TLSConfig = value
 	return b
 }
+
+// WithRelabelings adds the given value to the Relabelings field in the declarative configuration
+// and returns the receiver, so that objects can be build by chaining "With" function invocations.
```
QuentinBisson, line 63: I'm trying to have this `can be build` typo fixed in kubernetes upstream but let's see https://github.com/kubernetes/kubernetes/pull/123844 :p

QuentinBisson:
@simonpasquier the e2e test failures looks to be a fluke. Is there a way to retest?

pkg/prometheus/promcfg.go
```diff
@@ -101,11 +101,17 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
 	}, nil
 }
 
-func getScrapeClassConfig(cpf monitoringv1.CommonPrometheusFields) (map[string]*monitoringv1.ScrapeClass, string, error) {
+func getScrapeClassConfig(p monitoringv1.PrometheusInterface, cpf monitoringv1.CommonPrometheusFields) (map[string]*monitoringv1.ScrapeClass, string, error) {
```
simonpasquier, line 104: (nit) monitoringv1.PrometheusInterface gives access to monitoringv1.CommonPrometheusFields
```suggestion
func getScrapeClassConfig(p monitoringv1.PrometheusInterface) (map[string]*monitoringv1.ScrapeClass, string, error) {
```

pkg/prometheus/promcfg.go
simonpasquier: Looks good to me
(nit) validateRelabelConfigs should work with nil slices too.
```suggestion
```

pkg/prometheus/promcfg.go
```diff
@@ -1021,6 +1027,11 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 		})
 	}
 
+	// Add scrape class relabelings if there is any.
+	if scrapeClass != nil && len(scrapeClass.Relabelings) > 0 {
```
simonpasquier, line 1031: (nit) same here, we should be able to call generateRelabelConfig() with nil slice.
```suggestion
	if scrapeClass != nil {
```

pkg/prometheus/resource_selector.go
simonpasquier: sgtm

pkg/prometheus/resource_selector.go
simonpasquier: it might a side effect of #6384?

simonpasquier:
Thanks! This is almost there. Could you also a few more unit tests, esp. making sure that a default scrape class gets applied?

QuentinBisson:
I hope it's better with the new tests :)

nicolastakashi:
Amazing progress 👏🏽 
It looks great to me.

nicolastakashi (approved):

simonpasquier (approved):
Great!

PR #6389, opened 2024-03-13
Dynamic ThanosRuler governing service name

verejoel:
## Description

Currently it is not possible to run multiple instances of the ThanosRuler, as the service name is hardcoded. This change makes the generated service name dynamic, based on the name of the ThanosRuler resource.

## Type of change

## Verification

Tests pass.

## Changelog entry

```release-note
- ThanosRuler governing service name now includes the ThanosRuler resource name.
```

### DIFF
--- a/pkg/thanos/statefulset.go
+++ b/pkg/thanos/statefulset.go
@@ -38,7 +38,6 @@ const (
 	rulesDir                  = "/etc/thanos/rules"
 	configDir                 = "/etc/thanos/config"
 	storageDir                = "/thanos/data"
-	governingServiceName      = "thanos-ruler-operated"
 	defaultPortName           = "web"
 	defaultRetention          = "24h"
 	defaultEvaluationInterval = "15s"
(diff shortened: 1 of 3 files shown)


verejoel:
The only question I have is if the case where `tr.Name` is empty needs to be handled better? We could fallback to the existing `governingServiceName` as a default value if `name` is a zero string. But I see that this case is not handled for `volumeName` function.

simonpasquier:
Thanks for submitting the PR! We already have https://github.com/prometheus-operator/prometheus-operator/issues/6060 (issue) and https://github.com/prometheus-operator/prometheus-operator/pull/6067 (stale PR). The latter should allow configuration of the service while keeping the current behavior (since changing it might break some users).

simonpasquier:
> The only question I have is if the case where tr.Name is empty needs to be handled better?

A ThanosRuler resource has to have a name.
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

Could you update the PR to implement the option 1 described in https://github.com/prometheus-operator/prometheus-operator/issues/6060#issuecomment-2009195972?

PR #6390, opened 2024-03-13
Fix: ScrapeConfigs Selection Issue Across Different Namespaces

mviswanathsai:
## Description

The ScrapeConfigs added in a different namespace from the Prometheus CR that selects them are not being picked up. We suspect it might be because of how the `enqueueForNamespace()` function works. This PR aims to prove that this is indeed the issue and consequentially also fix it.

## Type of change

- [ x] `BUGFIX` (non-breaking change which fixes an issue)

## Verification
Tests added to `main_test.go`

## Changelog entry

```release-note
Fix ScrapeConfig not picked up bug
```

### DIFF
--- a/test/e2e/scrapeconfig_test.go
+++ b/test/e2e/scrapeconfig_test.go
@@ -249,8 +249,8 @@ func testScrapeConfigLifecycleInDifferentNS(t *testing.T) {
 	)
 	require.NoError(t, err)
 
-    // Make a prometheus object in promns which will select any ScrapeConfig resource with
-    // "role": "scrapeconfig" and/or "kubernetes.io/metadata.name": "<scns>"
+	// Make a prometheus object in promns which will select any ScrapeConfig resource with
+	// "role": "scrapeconfig" and/or "kubernetes.io/metadata.name": "<scns>"
 	p := framework.MakeBasicPrometheus(promns, "prom", scns, 1)
 	p.Spec.ScrapeConfigSelector = &metav1.LabelSelector{
 		MatchLabels: map[string]string{
--- a/test/e2e/scrapeconfig_test.go
+++ b/test/e2e/scrapeconfig_test.go
@@ -235,11 +235,12 @@ func testScrapeConfigLifecycleInDifferentNS(t *testing.T) {
 	// The ns where the scrapeConfig will reside
 	scns := framework.CreateNamespace(context.Background(), t, testCtx)
 	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, promns)
+	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, scns)
 
 	_, err := framework.CreateOrUpdatePrometheusOperator(
 		context.Background(),
 		promns,
-		[]string{promns},
+		[]string{scns},
 		nil,
 		[]string{promns},
 		nil,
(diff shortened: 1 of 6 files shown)


mviswanathsai:
Requesting review @simonpasquier @slashpai @ArthurSens. The test has been added and it does indeed fail, though I am still not 100% sure of the test. Please do look into it and let me know!

test/e2e/scrapeconfig_test.go
```diff
@@ -224,6 +224,87 @@ func testScrapeConfigLifecycle(t *testing.T) {
...
+		nil,
+		false,
+		true, // clusterrole
+		true,
+	)
+	require.NoError(t, err)
+
+	// Make a prometheus object in promns which will select any ScrapeConfig resource with
+	// "role": "scrapeconfig" and/or "kubernetes.io/metadata.name": "<scns>"
+	p := framework.MakeBasicPrometheus(promns, "prom", scns, 1)
```
simonpasquier, line 254: (nit) I would create a very thin Prometheus object with all other selectors being nil.

test/e2e/scrapeconfig_test.go
```diff
@@ -224,6 +224,87 @@ func testScrapeConfigLifecycle(t *testing.T) {
...
+		},
+	}
+	_, err = framework.CreatePrometheusAndWaitUntilReady(context.Background(), promns, p)
+	require.NoError(t, err)
+
+	// 1. Create a ScrapeConfig in scns and check that its targets appear in Prometheus
+	sc := framework.MakeBasicScrapeConfig(scns, "scrape-config")
+	sc.Spec.StaticConfigs = []monitoringv1alpha1.StaticConfig{
+		{
+			Targets: []monitoringv1alpha1.Target{"target1:9090", "target2:9090"},
```
simonpasquier, line 272: these targets will never be up?

simonpasquier (commented):
Good start!

test/e2e/scrapeconfig_test.go
simonpasquier: ah my bad, `WaitForActiveTargets()` doesn't look at the target health.

mviswanathsai: Indeed, it's just to see the targets showing up in the Prometheus resource.

mviswanathsai:
I am still not sure about the test. I ran
```make 

for n in "prometheus-operator" "prometheus-config-reloader" "admission-webhook"; do kind load docker-image "quay.io/prometheus-operator/$n:$(git rev-parse --short HEAD)"; done;````

as suggested by JP, but the e2e test still failed. I am not able to zero down on whether it is the fix or if it might be something with where the governing svc is created or something else about the test itself.  So I have made the push to the branch as suggested by Simon to see what happens of the tests.

File: test/e2e/scrapeconfig_test.go [hunk 3]
```diff
@@ -224,6 +224,87 @@ func testScrapeConfigLifecycle(t *testing.T) {
...
+	// The ns where the prometheus CR will reside
+	promns := framework.CreateNamespace(context.Background(), t, testCtx)
+	// The ns where the scrapeConfig will reside
+	scns := framework.CreateNamespace(context.Background(), t, testCtx)
+	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, promns)
+
+	_, err := framework.CreateOrUpdatePrometheusOperator(
+		context.Background(),
+		promns,
+		[]string{promns},
```
simonpasquier, line 242: I believe that the operator doesn't watch for the scrape config namespace.
```suggestion
		[]string{scns},
```

mviswanathsai: I suspected that too, but this too does not yield the expected result.

mviswanathsai: However, I just changed a few things now (including this suggestion), and the test is passing. 
Let me quickly remove/add a few parameters and see what made the tests fail earlier.

mviswanathsai: Well, this is weird.
The test passed, I then changed around a few label selectors, removed them all together and ran the test and it failed. Now, reverting back to the condition where it passed, the test again failed.

mviswanathsai: Findings:
1. The tests failed after letting Operator check the scns because I was looking for targets in scns in my local code.
2. The test failed when I set the label selector to nil value.
3. Setting both the right label selectors for labels and namespace passed the test.

mviswanathsai:
I am not sure why this test is failing. It seems to be passing locally. 
```
        --- PASS: TestAllNS/y/PromRulesExceedingConfigMapLimit (92.24s)
```

mviswanathsai:
@simonpasquier @slashpai, All the tests are passing locally. I cannot see why it is failing in the CI.

slashpai:
Likely flaky trying to re-run this, will see if it still fails

simonpasquier (approved):
Thanks!

PR #6410, opened 2024-03-19
Add extra context to warnings

QuentinBisson:
## Description

This PR adds more context to the deprecation warnings to know which resource (pod and service monitor) is using deprecated values as the current message is not really helpful to track the resource with issues.

This PR also adds some extra logs for other deprecations.

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -121,7 +121,7 @@ func getScrapeClassConfig(cpf monitoringv1.CommonPrometheusFields) (map[string]*
 // logger.
 func (cg *ConfigGenerator) WithKeyVals(keyvals ...interface{}) *ConfigGenerator {
 	return &ConfigGenerator{
-		logger:                 log.WithSuffix(cg.logger, keyvals),
+		logger:                 log.WithSuffix(cg.logger, keyvals...),
 		version:                cg.version,
 		notCompatible:          cg.notCompatible,
 		prom:                   cg.prom,
@@ -873,7 +873,7 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 
 	//nolint:staticcheck // Ignore SA1019 this field is marked as deprecated.
 	if ep.BearerTokenSecret.Name != "" {
-		level.Warn(cg.logger).Log("msg", "'bearerTokenSecret' is deprecated, use 'authorization' instead.", "podmonitor", m.Name, "namespace", m.Namespace)
+		level.Warn(cg.logger).Log("msg", "'bearerTokenSecret' is deprecated, use 'authorization' instead.")
 		if s, ok := store.TokenAssets[fmt.Sprintf("podMonitor/%s/%s/%d", m.Namespace, m.Name, i)]; ok {
 			cfg = append(cfg, yaml.MapItem{Key: "bearer_token", Value: s})
 		}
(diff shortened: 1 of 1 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -873,6 +873,7 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 
 	//nolint:staticcheck // Ignore SA1019 this field is marked as deprecated.
 	if ep.BearerTokenSecret.Name != "" {
+		level.Warn(cg.logger).Log("msg", "'bearerTokenSecret' is deprecated, use 'authorization' instead.", "podmonitor", m.Name, "namespace", m.Namespace)
```
simonpasquier, line 876: the logger should already have the `pod_monitor="xxx/xxx"` keyval.
```suggestion
		level.Warn(cg.logger).Log("msg", "'bearerTokenSecret' is deprecated, use 'authorization' instead.")
```

simonpasquier (commented):
Good catch! There might a bit more places where we have deprecated fields to log (.e.g. probes)?

pkg/prometheus/promcfg.go
QuentinBisson: Oh, maybe it can be added as a keyval to the logger at the top of the function instead, i'll check how.
I will also add a warning to the  other deprecations

QuentinBisson: Oh yes you're right that it should but it's not showing in my logs for some reason

QuentinBisson: So the weird part is that when i use logfmt, I have the following logs without the resource information

```
level=warn ts=2024-03-20T20:26:27.127226961Z caller=promcfg.go:819 component=prometheus-controller msg="'targetPort' is deprecated, use 'port' instead." version=v2.50.1
level=warn ts=2024-03-20T20:26:27.127415876Z caller=promcfg.go:819 component=prometheus-controller msg="'targetPort' is deprecated, use 'port' instead." version=v2.50.1
```

But when I use json is it there:

```
{"[pod_monitor crossplane/crossplane]":"(MISSING)","caller":"promcfg.go:819","component":"prometheus-controller","level":"warn","msg":"'targetPort' is deprecated, use 'port' instead.","ts":"2024-03-20T20:25:44.813625012Z","version":"v2.50.1"}
```

QuentinBisson: Oh i found the issue

QuentinBisson: I ran this little script:

```go
func WithKeyVals(logger log.Logger, keyvals ...interface{}) log.Logger {
	return log.WithSuffix(logger, keyvals...)
}

func main() {
	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
	logger = WithKeyVals(logger, "caller", log.DefaultCaller)
	level.Info(logger).Log("msg", "hello")
	level.Warn(logger).Log("msg", "hello")
}
```

And we do not add the spread operator (...) after keyvals in the WithKeyVals function, the we pass the keyval function parameter as an array hence this in json: "[pod_monitor crossplane/crossplane]":"(MISSING)"

pkg/prometheus/promcfg.go
```diff
@@ -121,7 +121,7 @@ func getScrapeClassConfig(cpf monitoringv1.CommonPrometheusFields) (map[string]*
 // logger.
 func (cg *ConfigGenerator) WithKeyVals(keyvals ...interface{}) *ConfigGenerator {
 	return &ConfigGenerator{
-		logger:                 log.WithSuffix(cg.logger, keyvals),
+		logger:                 log.WithSuffix(cg.logger, keyvals...),
```
QuentinBisson, line 124: @simonpasquier this is the fix to make sure the pod_monitor key is in the logs

pkg/prometheus/promcfg.go
```diff
@@ -1525,10 +1528,10 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 			{Key: "target_label", Value: "endpoint"},
 			{Key: "replacement", Value: ep.Port},
 		})
-	} else if ep.TargetPort != nil && ep.TargetPort.String() != "" { //nolint:staticcheck // Ignore SA1019 this field is marked as deprecated.
+	} else if ep.TargetPort != nil && ep.TargetPort.String() != "" {
```
QuentinBisson, line 1531: This is actually not deprectated as per the doc page

simonpasquier (approved):
Thanks!

PR #6386, opened 2024-03-12
[WIP] feat: add support for Docker SD in ScrapeConfig 

mviswanathsai:
Moved to #6421

PR #6421, opened 2024-03-22
feat: add support for Docker SD in ScrapeConfig 

mviswanathsai:
## Description
This PR aims to add Docker service discovery config support to the `ScrapeConfig` Alpha CRD.

## Type of change

## Verification

I have written tests under `pkg/prometheus/promfg_test.go` for verifying the working of DockerSD config.

## Changelog entry

```release-note
Add support for Docker SD in ScrapeConfig
```


slashpai:
Thanks for addressing comments from https://github.com/prometheus-operator/prometheus-operator/pull/6386

Overall it looks good to me but let @simonpasquier also take a look before merging :)

slashpai (approved):

slashpai:
@simonpasquier can you review too?

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -675,3 +684,50 @@ type DigitalOceanSDConfig struct {
...
+// Docker SD configurations allow retrieving scrape targets from Docker Engine hosts.
+// This SD discovers "containers" and will create a target for each network IP and
+// port the container is configured to expose.
+// See https://prometheus.io/docs/prometheus/latest/configuration/configuration/#docker_sd_config
+// +k8s:openapi-gen=true
+type DockerSDConfig struct {
+	// Address of the docker daemon
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Host string `json:"host"`
```
nicolastakashi, line 697: What do you think about remove [this part of the code](https://github.com/prometheus-operator/prometheus-operator/pull/6421/files#diff-aca3d056c83da1ddc448c8c40960960d7ed00ef61f8d5f08a40e165564d74dcdR1046) in favor of a kubebuilder annotation: `+kubebuilder:validation:Pattern`?

slashpai: I thought about kubebuilder validation but I wasn't sure if we can have the right regex prometheus accepts since prometheus is using url.Parse https://github.com/prometheus/prometheus/blob/main/discovery/moby/docker.go#L110-L112

url.Parse looks to be abstracted way for a complex regex validation?

nicolastakashi: I see!
Yes docker host could be could be like that: `unix:///var/run/docker.sock` or other kind of dns such as http://

mviswanathsai: Would we need to do the validation here? Or should we let it be handled at the url.Parse level JP mentioned at the Promtheus level?

nicolastakashi: I'm fine using this only on the url.Parse especially because the regex pattern could be very complex.

nicolastakashi (approved):

pkg/prometheus/promcfg_test.go
```diff
@@ -6803,6 +6803,222 @@ func TestScrapeConfigSpecConfigWithDigitalOceanSD(t *testing.T) {
...
+func TestScrapeConfigSpecConfigWithDockerSDConfig(t *testing.T) {
+	for _, tc := range []struct {
+		name   string
+		scSpec monitoringv1alpha1.ScrapeConfigSpec
+		golden string
+	}{
+		{
+			name: "docker_sd_config_with_filters",
+			scSpec: monitoringv1alpha1.ScrapeConfigSpec{
+				DockerSDConfigs: []monitoringv1alpha1.DockerSDConfig{
```
ArthurSens, line 6815: Whoops, forgot to select the lines I was referring to :P

ArthurSens:
Ah, looks like we have many PRs open related to service discovery in scrape configs 😬

Could you also rebase the PR?

pkg/prometheus/promcfg_test.go
mviswanathsai: OAuth2.0 and Authorization are supposed to be mutually exclusive. So I think to reduce it to 2 test cases, from 4. Would that be fine?

ArthurSens: Oh right! Hmmmm, I'm not sure... it doesn't hurt that much to keep the 4 test cases, I was just nitpicking

mviswanathsai: I did feel like having a separate test case for host_name was not really necessary. So I brought it down to three, one for each of the mutually exclusive OAuth, Auth and BasicAuth fields.

ArthurSens (approved):
I'll let @slashpai handle the merge since she is coordinating the 0.73 release. I'm not sure if there's extra coordination going on between other PRs that touch service discoveries

slashpai:
Yes this one should be good to go was next on queue for SDs

slashpai:
@mviswanathsai I remember seeing a test with Docker filter when reviewing earlier not sure if that was removed. I think a test for that would be good to see if it is configured correctly

pkg/prometheus/promcfg_test.go
```diff
@@ -6803,6 +6803,256 @@ func TestScrapeConfigSpecConfigWithDigitalOceanSD(t *testing.T) {
...
+									Key: "proxy-header",
+								},
+							},
+						},
+						FollowRedirects:    ptr.To(true),
+						EnableHTTP2:        ptr.To(true),
+						Port:               ptr.To(9100),
+						RefreshInterval:    ptr.To(monitoringv1.Duration("30s")),
+						HostNetworkingHost: ptr.To("localhost"),
+						Filters: &[]monitoringv1alpha1.DockerFilter{
```
mviswanathsai, line 6844: We still include filters, see here.

slashpai:
Thanks!

PR #5063, opened 2022-10-03
pkg/apis: Support empty relabelConfig separator

slashpai:
Fixes #5003

Signed-off-by: Jayapriya Pai <slashpai9@gmail.com>

## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [x] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note
Allow empty separator in relabelConfig
```

PR #6425, opened 2024-03-23
Allow empty separator in relabel config

### LINKED ISSUE #5003: Concatenating labels (using empty 'separator') in relabelings not possible
**What happened?**

I have a ServiceMonitor with relabelings on its endpoints, and one of the relabeling configs must concatenate two labels. Its relabeling config is defined like this:

```yaml
    - sourceLabels: [label1, label2]
      separator: ""
      action: replace
      targetLabel: custom_label
```

This is exactly the same configuration as I have in a previous Prometheus setup without the prometheus operator, (having this in `prometheus.yml` instead of via a `ServiceMonitor`). This rule works in the other setup correctly and concatenates these labels.

It appears that the empty separator is not working, as it gets replaced by the default value ";". If I use any other separator in that spot, it does work with the provided value, so I'm certain the config is actually being used. Just when I use the empty string (both with single or with double quotes), the value is ignored and replaced by the default `;`.

This means it currently is not possible to concatenate two labels.

Yes, using an empty string as separator should just concatenate the input without any separator.

Create a ServiceMonitor with endpoints contain a relabelings config, with the step I have above (of course with the labels you want to input&output).

**Environment**

    `quay.io/prometheus-operator/prometheus-operator:v0.58.0` (used via the kube-prometheus helm chart)

```bash
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.2", GitCommit:"f66044f4361b9f1f96f0053dd46cb7dce5e990a8", GitTreeState:"clean", BuildDate:"2022-06-15T14:22:29Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.11-gke.400", GitCommit:"b4e1ab06be827438def8aee0021791b413a0961d", GitTreeState:"clean", BuildDate:"2022-06-24T09:27:38Z", GoVersion:"go1.16.15b7", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:

GKE

* Manifests: /

* Prometheus Operator Logs: /

I know some Go, but don't know the workings of the Prometheus operator. But a quick search on `separator` in the repo shows the `RelabelConfig` struct having json tags `omitempty` (Source code: https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/types.go#L1093). And an empty string is considered an empty value, so I suspect that is the culprit here; it does not allow you to intentionally provide the empty string as separator because it gets discarded as missing, and replaced by the default value later on.

haanhvu:
## Description

Fixes #5003

Currently we don't allow empty separator in relabel config. If users specify an empty separator, it will be turned into the default `;` from Prometheus side. Since Prometheus makes use of empty separator, Prometheus operator should do that too.

## Type of change

## Verification
No tests broken. New tests passed.

## Changelog entry

Allow empty seperator in relabel config

## Question
Is this considered a breaking change?


haanhvu:
@slashpai Could you review?

Basically, I followed your PR. Just did 2 changes:

- Change `//+kubebuilder:default=;` to `//+kubebuilder:default=";"`. Without `""` it would generate array type instead of string type, as you could see in [your PR](https://github.com/prometheus-operator/prometheus-operator/pull/5063/files#diff-39f2cbbc0137b438d3cb17c26f83a5a864eae805ed655fd2fa6a7c59010ba612R264).
- Fixed tests

Pls also reactivate CI. I suspect e2e might have been flaky.

slashpai:
Thanks for taking this up.
Did you test with a local cluster if changes are working?

haanhvu:
> Thanks for taking this up. Did you test with a local cluster if changes are working?

I ran with `./scripts/run-external.sh -c` and it ran successfully. But you can see, e2e are failing. At first I thought it was flaky but now looked deeper into the code I see the failure relates to my change.

I think I need to look deeper into how this is handled in Prometheus codebase. From my experience label-related changes are very bug-prone. Don't want to reinvent the wheel and create subtle bugs.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1497,7 +1497,8 @@ type RelabelConfig struct {
 	SourceLabels []LabelName `json:"sourceLabels,omitempty"`
 
 	// Separator is the string between concatenated SourceLabels.
-	Separator string `json:"separator,omitempty"`
+	//+kubebuilder:default=";"
+	Separator string `json:"separator"`
```
simonpasquier, line 1501: The "issue" with adding a default value is that the API response will change to return `separator: ";"`  if the field isn't set and it may disrupt some clients (reconcilers for instance).

Another option would be to change the field type to be a string pointer:
* if the pointer is nil, don't set the `separator` field in the generated Prometheus config (and let Prometheus use its own default).
* if not nil, pass the value as-is.

The risk with this option is for users that have `separator: ""` in their YAML manifests today as it will change the behavior for them.

```suggestion
	Separator *string `json:"separator"`
```

haanhvu: Yes I agree this is a breaking change. But since this change is on par with Prometheus's behavior, I think it's worth doing.

haanhvu: I thought of your suggested approach too. Will try this. I think it is less risky than the current (default `;` approach) because there might be less users setting `separator: ""` than users not setting it at all.

simonpasquier: Agreed, I don't think that `separator: ""` is widely used in the wild, especially knowing that as of today, it's the same as not setting it. If we follow that path, we's till need to mark it as a CHANGE in the changelog for safety.

haanhvu:
@slashpai @simonpasquier All tests are passed now. Do you think we need to add tests for `""` are correctly used (not ignored or replaced by `;`)? If yes, where should I add those tests?

simonpasquier:
I would extend `TestAdditionalAlertRelabelConfigs` to have a relabel config rule setting `separator: ""`.

pkg/prometheus/resource_selector.go
```diff
@@ -313,8 +313,8 @@ func validateRelabelConfig(p monitoringv1.PrometheusInterface, rc monitoringv1.R
 		if !(rc.Regex == "" || rc.Regex == relabel.DefaultRelabelConfig.Regex.String()) ||
 			!(rc.Modulus == uint64(0) ||
 				rc.Modulus == relabel.DefaultRelabelConfig.Modulus) ||
-			!(rc.Separator == "" ||
-				rc.Separator == relabel.DefaultRelabelConfig.Separator) ||
+			!(rc.Separator == nil ||
+				rc.Separator == &relabel.DefaultRelabelConfig.Separator) ||
```
mviswanathsai, line 317: The addresses will always be different, I think you should dereference rc.Separator instead of getting the address of the default value.

pkg/prometheus/resource_selector.go
```diff
@@ -327,8 +327,8 @@ func validateRelabelConfig(p monitoringv1.PrometheusInterface, rc monitoringv1.R
 				rc.TargetLabel == relabel.DefaultRelabelConfig.TargetLabel) ||
 			!(rc.Modulus == uint64(0) ||
 				rc.Modulus == relabel.DefaultRelabelConfig.Modulus) ||
-			!(rc.Separator == "" ||
-				rc.Separator == relabel.DefaultRelabelConfig.Separator) ||
+			!(rc.Separator == nil ||
+				rc.Separator == &relabel.DefaultRelabelConfig.Separator) ||
```
mviswanathsai, line 331: Same here. Check this [playground](https://go.dev/play/p/Ke8OiBUO9f8) out.

pkg/prometheus/resource_selector.go
haanhvu: Thanks. I know what you mean. I was just too focused on the nil separator, since that's the breaking change. Will fix this in next commit.

test/e2e/prometheus_test.go
```diff
@@ -4657,7 +4659,7 @@ func testRelabelConfigCRDValidation(t *testing.T) {
 			relabelConfigs: []*monitoringv1.RelabelConfig{
 				{
 					SourceLabels: []monitoringv1.LabelName{"__address__"},
-					Separator:    ",",
+					Separator:    &sep1,
```
mviswanathsai, line 4662: (nit) you can use ptr.To instead.

test/e2e/prometheus_test.go
```diff
@@ -4668,7 +4670,7 @@ func testRelabelConfigCRDValidation(t *testing.T) {
 			scenario: "empty-separator",
 			relabelConfigs: []*monitoringv1.RelabelConfig{
 				{
-					Separator: "",
+					Separator: &sep2,
```
mviswanathsai, line 4673: Same here.

test/e2e/prometheus_test.go
haanhvu: Agree, using pointer package from k8s would be better.

haanhvu: I'll fix this nit tomorrow when I have more time.

ArthurSens (approved):
Just some nitpicking about file names[[1](https://chat.openai.com/share/4c2c2140-394b-4f57-a202-b5b3234c3bc3)][[2](https://ieftimov.com/posts/testing-in-go-golden-files/#:~:text=First%2C%20golden%20files%20are%20coupled%20to%20the%20test%20cases%20(a.k.a.%20the%20inputs).)]

pkg/prometheus/testdata/AdditionalAlertRelabelConfigs.golden
```diff
@@ -1,3 +1,7 @@
 - action: drop
   source_labels: [__meta_kubernetes_node_name]
-  regex: spot-(.+)
\ No newline at end of file
+  regex: spot-(.+)
+- sourceLabels: [label1, label2]
+  separator: ""
+  action: replace
+  targetLabel: custom_label
```
ArthurSens, line 7: It took me forever to understand why such a simple change (changing separator to pointer), would change so much a golden file, e.g. source labels, action, etc.

That was until I realized that this file is not really a golden file by definition, but an input! We should change the suffix of this file to something else, e.g. `.input`

simonpasquier: :+1: btw the config format is wrong since it expects Prometheus snake-case format.

pkg/prometheus/resource_selector_test.go
```diff
@@ -392,7 +392,7 @@ func TestValidateRelabelConfig(t *testing.T) {
 			relabelConfig: monitoringv1.RelabelConfig{
 				SourceLabels: []monitoringv1.LabelName{"__tmp_port"},
 				TargetLabel:  "__port1",
-				Separator:    relabel.DefaultRelabelConfig.Separator,
+				Separator:    &relabel.DefaultRelabelConfig.Separator,
```
simonpasquier, line 395: (nit)
```suggestion
				Separator:    ptr.To(relabel.DefaultRelabelConfig.Separator),
```

mviswanathsai:
You might want to squash these commits into one (or a few) once you are done with all the changes. You might already be on it, just a reminder!

haanhvu:
> You might want to squash these commits into one (or a few) once you are done with all the changes. You might already be on it, just a reminder!

I've never done this in any of my merged PRs here. Is there any reason to do that in this PR?

mviswanathsai:
It's not really a necessity per se, but having one point in the final commit tree which tells you what exactly this commit did, than to have 10 or so commits that took to achieve it would be more desirable.
Ofcourse, in the PR itself there's nothing 'wrong" with having as many commits.

mviswanathsai:
For example, "fix e2e failure" is meaningful within the context of this PR. But once merged to main, it would become rather ambiguous.

haanhvu:
> For example, "fix e2e failure" is meaningful within the context of this PR. But once merged to main, it would become rather ambiguous.

Yes many commits may not be meaningful out of PR. But no matter how many commits are in the PR, in the end there's only one merged commit right? Like this: https://github.com/prometheus-operator/prometheus-operator/commit/128acfc449deb7481e4d48b6b915a791aa9fc13f

Of course if that's really necessary I'll do it. But I have seen many PRs existing with several commits, even from maintainers. So currently I don't see any values of doing it here. If I miss something else then please specify.

slashpai:
IMHO for best practices its good to squash commits with meaningful message. Incase not doing GitHub has option squash and merge but we would still need to update comments while doing that.

haanhvu:
> IMHO for best practices its good to squash commits with meaningful message. Incase not doing GitHub has option squash and merge but we would still need to update comments while doing that.

Yes, please help me squash when you merge. The first commit message has everything you need to know about this PR. I always follow your format [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#format-of-the-commit-message).

Personally, I prefer to keep PR commits as they are. In case I do several approaches in one PR, that PR merged, then I need to revert the PR and come back to a previous approach, commits history can help me easily find back that approach in my PR.

pkg/prometheus/testdata/AdditionalAlertRelabelConfigs.golden
```diff
@@ -1,3 +1,7 @@
 - action: drop
   source_labels: [__meta_kubernetes_node_name]
-  regex: spot-(.+)
\ No newline at end of file
+  regex: spot-(.+)
+- sourceLabels: [label-a, label-b]
+  separator: ""
+  action: replace
+  targetLabel: custom-label
```
simonpasquier, line 7:
```suggestion
- source_labels: [label-a, label-b]
  separator: ""
  action: replace
  target_label: custom-label
```

haanhvu: I thought the input file followed the operator's format. And the output (expected) file followed Prometheus's format. But from your suggestion, it means both follow Prometheus's format?

simonpasquier: correct, it's because the test is about additionalScrapeConfigs which hold native Prometheus configs.

simonpasquier (approved):

pkg/prometheus/testdata/AdditionalAlertRelabelConfigs.golden
haanhvu: Hmm, then do you think this test can verify our fix? I thought we needed a test regarding the operator's format? 🤔 Like, before the operator's format doesn't support `""` but now it does?

haanhvu:
CI didn't run so I squashed commits to reactivate it. But it still doesn't run... Will wait then.

slashpai:
@simonpasquier Is it good to merge?

pkg/prometheus/testdata/AdditionalAlertRelabelConfigs.golden
mviswanathsai: The operator appends the additional config you specify by name (I think?) in the final generated config. In this, we process the input config (user defined). I think what you're trying to test is that the generated config which contains the appended additional config still contains the empty field instead of omitting it entirely.

PR #6426, opened 2024-03-23
Add e2e test on secret update for PrometheusAgent

### LINKED ISSUE #6414: Add e2e test on secret update for PrometheusAgent
Ideally we'd need an e2e test to verify this but it's a bit more involved. I also think that we should reduce code duplication since all controllers have similar patterns for configmap/secret handlers but again it doesn't need to be in this be in this PR.

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/6412#pullrequestreview-1948360281_

jamshidi799:
It fixes #6414

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/test/e2e/prometheusagent_test.go
+++ b/test/e2e/prometheusagent_test.go
@@ -29,6 +29,7 @@ import (
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
+	testFramework "github.com/prometheus-operator/prometheus-operator/test/framework"
 )
 
 func testCreatePrometheusAgent(t *testing.T) {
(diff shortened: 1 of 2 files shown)


test/e2e/prometheusagent_test.go
```diff
@@ -177,3 +178,116 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+			Selector: map[string]string{
+				"group": name,
+			},
+		},
+	}
+	if finalizerFn, err := framework.CreateOrUpdateServiceAndWaitUntilReady(context.Background(), ns, svc); err != nil {
+		t.Fatal(err)
+	} else {
+		testCtx.AddFinalizerFn(finalizerFn)
+	}
```
ArthurSens, line 234:
```suggestion
	if finalizerFn, err := framework.CreateOrUpdateServiceAndWaitUntilReady(context.Background(), ns, svc); err != nil {
		t.Fatal(err)
	} 
	testCtx.AddFinalizerFn(finalizerFn)
```

ArthurSens: the `else` isn't needed

test/e2e/prometheusagent_test.go
```diff
@@ -177,3 +178,116 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+
+	if _, err := framework.MonClientV1.ServiceMonitors(ns).Create(context.Background(), sm, metav1.CreateOptions{}); err != nil {
+		t.Fatal("Creating ServiceMonitor failed: ", err)
+	}
+
+	pAgent := framework.MakeBasicPrometheusAgent(ns, name, name, 1)
+	pAgent.Spec.ServiceMonitorNamespaceSelector = &metav1.LabelSelector{
+		MatchLabels: matchLabels,
+	}
+	pAgent.Spec.ScrapeInterval = "1s"
```
ArthurSens, line 272: Could you explain the motivation for `1s` scrape interval?

test/e2e/prometheusagent_test.go
```diff
@@ -177,3 +178,116 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+	secret.Data["bearertoken"] = []byte("abc")
+	if _, err := framework.KubeClient.CoreV1().Secrets(ns).Update(context.Background(), secret, metav1.UpdateOptions{}); err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(time.Minute * 2)
+
+	if err := framework.WaitForHealthyTargets(context.Background(), ns, "prometheus-agent-operated", 1); err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
jamshidi799: The idea is first the secret is wrong(it's an empty string while the application expecting the `abc` token) and in this way, the Prometheus target could not be healthy. after getting that confirmation, the bearer token value is fixed correctly(`abc`). So if the Prometheus operator reconciles the secret update correctly, the Prometheus targets should become healthy.

In addition to that, I use two minutes of sleep to ensure that those updates are reflected in the Prometheus config. if we remove it, I think the test will fail

simonpasquier:
Since we're looking at end-to-end tests, I would tackle it from the end-user's standpoint: "I updated the credentials used to connect to my target but the update was never propagated to Prometheus and scrapes started to fail".

In practice:
* deploy a target which requires a valid bearer token
* configure a monitor referring to a secret containing an invalid bearer token
* wait for Prometheus to report the target as down
* update the secret content with the good token
* wait for Prometheus to report the target as up

test/e2e/prometheusagent_test.go
jamshidi799: It reduces the overall time that the test takes to pass

test/e2e/prometheusagent_test.go
```diff
@@ -177,3 +178,116 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+	if err := framework.WaitForHealthyTargets(context.Background(), ns, "prometheus-agent-operated", 1); err == nil {
+		t.Fatal("All targets should be down")
+	}
+
+	secret.Data["bearertoken"] = []byte("abc")
+	if _, err := framework.KubeClient.CoreV1().Secrets(ns).Update(context.Background(), secret, metav1.UpdateOptions{}); err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(time.Minute * 2)
```
ArthurSens, line 287: I believe `WaitForHealthyTargets` already waits until it finds the healthy targets. What do you think about removing this sleep then?

test/e2e/prometheusagent_test.go
```diff
@@ -177,3 +178,116 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+		MatchLabels: matchLabels,
+	}
+	_, err = framework.CreatePrometheusAgentAndWaitUntilReady(ctx, ns, pAgent)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := framework.WaitForHealthyTargets(context.Background(), ns, "prometheus-agent-operated", 1); err == nil {
+		t.Fatal("All targets should be down")
+	}
```
ArthurSens, line 280: If all targets should be down, why do we look for 1 healthy target?

ArthurSens (commented):
I'm struggling a bit to understand the correctness of the test. Is `abc` a valid bearertoken? Could we add comments to the code explaining the test a little bit?

Issue #6361, opened 2024-03-03
init-config-reloader cannot run in restricted PSS


rajan123456:
### Is there an existing issue for this?

- [X] I have searched the existing issues

## Description

https://github.com/bitnami/charts/issues/23606

I am using the bitnami distribution of the kube-prometheus helm chart and deploying it in a Kubernetes namespace which enforces restricted PSS. The init-config-reloader initContainers of both alertmanager and prometheus statefulSets are unable to be admitted to the cluster as they do not set the securityContext properly.

## Steps to Reproduce

1. Create a namespace:
```
kubectl create ns ns-prometheus
```
2. Label the namespace to enforce restricted PSS:
```
kubectl label --overwrite ns ns-prometheus   pod-security.kubernetes.io/enforce=restricted   pod-security.kubernetes.io/enforce-version=v1.29
```
3. Install the helm chart:
```
helm upgrade --install my-prometheus bitnami/kube-prometheus -n ns-prometheus
```
4. Verify status of statefulsets

## Expected Result

The statefulSet pods should be admitted to the namespace enforcing restricted PSS

## Actual Result

Pods are rejected:

Prometheus StatefulSet
```
Events:
  Type     Reason        Age                    From                    Message
  ----     ------        ----                   ----                    -------
  Warning  FailedCreate  2m23s (x16 over 5m7s)  statefulset-controller  create Pod prometheus-my-prometheus-kube-prometh-prometheus-0 in StatefulSet prometheus-my-prometheus-kube-prometh-prometheus failed error: pods "prometheus-my-prometheus-kube-prometh-prometheus-0" is forbidden: violates PodSecurity "restricted:v1.29": runAsNonRoot != true (pod or container "init-config-reloader" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "init-config-reloader" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
  ```
  
  AlertManager StatefulSet
  ```
  Events:
  Type     Reason        Age                     From                    Message
  ----     ------        ----                    ----                    -------
[...]

PR #6469, opened 2024-04-03
WIP: Issue #6361

yp969803:
## Description

fixing: #6361

configuring containers for namespace PSS label

## Type of change

```release-note

```


yp969803:
@simonpasquier can you review, should i proceed with similar approach, or something to be added

simonpasquier:
I assume that the init containers should work fine if the operator sets the security context to match the restricted policy. Can you try this first?

simonpasquier:
To reformulate what we've discussed during the office hours, the idea would be to update the container security context to match the restricted policy:

https://github.com/prometheus-operator/prometheus-operator/blob/60d1fbe725e72b9d9dbbf0d902405323018a0d32/pkg/operator/config_reloader.go#L302-L309

In particular the [original issue](https://github.com/bitnami/charts/issues/23606) reported that the pod creation failed with `forbidden: violates PodSecurity "restricted:v1.29": runAsNonRoot != true (pod or container "init-config-reloader" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "init-config-reloader" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")`.

yp969803:
@simonpasquier can you review the code, why the test is failing, i think have handled the nil pointer effectivly.

simonpasquier:
@yp969803 I don't think that we need the complexity of reading the `pod-security.kubernetes.io/enforce` label at all. As I commented [here](https://github.com/prometheus-operator/prometheus-operator/pull/6469#issuecomment-2255807304), I would rather tweak the settings of the init container to comply with the restricted policy in all cases as you're doing in 
image

yp969803:
@simonpasquier shouldnt we do this for all the containers ? as if init-containers are failing all other containers might also fail,

simonpasquier:
let's start with the init container.

yp969803:
@simonpasquier getting 
```
         	            	prometheus operatorupgrade-operatorupgrade-si71rv-0-8c65f3c4/operator-upgrade failed to become available: client rate limiter Wait returned an error: context deadline exceeded: context deadline exceeded
 ```
 i have only added restricted policies for configloader and init loader

**Comment by simonpasquier** [comment]
I'd suggest that you test the change locally following https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#local-development

**Comment by yp969803** [comment]
@simonpasquier i have tested in local init container, config-loader and prometheus container all three failed

**Comment by yp969803** [comment]
```
kind: Prometheus
metadata:
  name: prometheus-example
  namespace: monitoring
spec:
  replicas: 1
  securityContext:
    runAsUser: 1000
```
also i have to add "runAsUser: 1000" as nobody user didnt work in runAsNonRoot

**Comment by yp969803** [comment]
@simonpasquier i have added stuffs for prometheus container and things are working locally can you the review the work, so that i can also do changes for alertmanager and thanos containers

**Comment by yp969803** [comment]
@simonpasquier can u review the work

File: pkg/alertmanager/statefulset.go [hunk 1]
```diff
@@ -666,6 +666,19 @@ func makeStatefulSetSpec(logger log.Logger, a *monitoringv1.Alertmanager, config
 		alertmanagerURIScheme = "https"
 	}
 
+	securityContext := &v1.SecurityContext{
+		ReadOnlyRootFilesystem:   ptr.To(true),
+		AllowPrivilegeEscalation: ptr.To(false),
+		RunAsNonRoot:             ptr.To(true),
+		RunAsUser:                ptr.To(int64(1000)),
```
simonpasquier, line 673: This would be a breaking change for some users.

simonpasquier (commented):
I feel that we're getting a bit ahead of ourselves. Before implementing new e2e tests, we should check that tightening the security context for config reloaders only don't break the existing tests. I wouldn't change anything for the other containers yet.

pkg/alertmanager/statefulset.go
yp969803: @simonpasquier previously i have checked  all e2e tests for cofig-reloader, they are working fine. then i have updated the stuffs for other containers

yp969803:
```
 		
ReadOnlyRootFilesystem:   ptr.To(true),
		AllowPrivilegeEscalation: ptr.To(false),
		RunAsNonRoot:             ptr.To(true),
		RunAsUser:                ptr.To(int64(1000)),
```
@simonpasquier can you explain why this will result in break

yp969803: iirc the images of prometheus, thanios-ruler, and alertmanager allows runasNonRoot and they didnt have userId specified

simonpasquier: people might use custom images (e.g. not from upstream).

yp969803: @simonpasquier, it is necessary to specify userId, for the feild RunAsNonRoot=true, how we can get the userID in the user uses custom images

yp969803: or we can ask user to specify userID while the user is specifying custom image
```
	// +optional
	Image *string `json:"image,omitempty"`
        UserID *string
 ```

yp969803:
this update requires, discussions with the maintainers, currently marking this pr as draft, will reopen it after discussions with maintainers.

simonpasquier:
since we have a documented procedure (thanks to your #7160 PR), I would say that we don't this change for now. We could revisit later if needed.

Issue #6466, opened 2024-04-03
Add proxyURL validation for the AlertmanagerConfig CRD.


simonpasquier:
We'll need a similar fix for the AlertmanagerConfig CRD.

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/6464#pullrequestreview-1976154712_

PR #6481, opened 2024-04-05
feat:validate proxyURL for alertManagerCfg (#6466)

yp969803:
## Description

fixes: #6466
added validation for proxyURL for alertManagerCfg

## Type of change

## Verification
written unit test
## Changelog entry

```
validate proxyURL for alertManagerCfg

```

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/validation.go
+++ b/pkg/apis/monitoring/v1alpha1/validation.go
@@ -17,6 +17,7 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
@@ -57,6 +58,12 @@ func (hc *HTTPConfig) Validate() error {
 
 	if hc.TLSConfig != nil {
 		if err := hc.TLSConfig.Validate(); err != nil {
+			return err
+		}
+	}
+
+	if hc.ProxyURL != "" {
+		if _, err := url.Parse(hc.ProxyURL); err != nil {
 			return err
 		}
 	}
--- a/pkg/apis/monitoring/v1alpha1/validation_test.go
+++ b/pkg/apis/monitoring/v1alpha1/validation_test.go
@@ -17,6 +17,9 @@ package v1alpha1
 import (
 	"reflect"
 	"testing"
+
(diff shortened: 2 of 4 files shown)


pkg/alertmanager/amcfg.go
```diff
@@ -1724,6 +1724,11 @@ func (hc *httpClientConfig) sanitize(amVersion semver.Version, logger log.Logger
 		return err
 	}
 
+	if hc.ProxyURL != "" {
+		_, err := url.Parse(hc.ProxyURL)
```
afzal442, line 1728: taking the idea of validation on proxyurl obj of `oauth2` type would be better. 
https://github.com/prometheus-operator/prometheus-operator/blob/4664990a5e2f15bd1ae408d54bd77cfd90c0c7c6/pkg/alertmanager/amcfg.go#L1783-L1785

cc @simonpasquier

nicolastakashi: Sorry @afzal442 don't follow what you mean 😄 
The code snippet you shared is only about log message.
Could you clarify?

afzal442: Aha, I see. I didn't look into that much. Nice caught. Parsing was missing. I think only one extra condition to be met here is `!amVersion.GTE(semver.MustParse("0.25.0"))`. 🤔

simonpasquier (commented):
we need also to add a check to convertHTTPConfig()

yp969803:
@simonpasquier  done the above change

simonpasquier:
Sorry reading again the code, I suggest that you add the validation [here](https://github.com/prometheus-operator/prometheus-operator/blob/35340a6c4a97a9e6899f3d0319f6c47f781c2062/pkg/apis/monitoring/v1alpha1/validation.go#L25) and [here](https://github.com/prometheus-operator/prometheus-operator/blob/35340a6c4a97a9e6899f3d0319f6c47f781c2062/pkg/apis/monitoring/v1beta1/validation.go#L25).

simonpasquier:
Here is a short version to get you started: https://github.com/simonpasquier/prometheus-operator/commit/400c2c51a507d9b15e61d4692d161d68f750d3ae

Issue #6284, opened 2024-02-05
Duplicate ports warning


sebhoss:
## Description

I recently updated prometheus-operator and saw the following line in the logs of the operator:

```
level=warn ts=2024-02-05T06:00:35.226955216Z caller=klog.go:106 component=k8s_client_runtime func=Warning msg="spec.template.spec.containers[1].ports[0]: duplicate port definition with spec.template.spec.initContainers[0].ports[0]"
```

I wasn't sure which pod actually declared duplicate ports, so after a short search I noticed that all prometheus pods now have both an initContainer (init-config-reloader) as well as a regular container (config-reloader) running on port 8080. Since I'm not the one who declared those ports/containers, I'm not sure whether I should do something about this warning.

## Steps to Reproduce

- Deploy latest prometheus-operator
- Deploy minimal `Prometheus` resource
- Watch logs of operator

## Expected Result

No warning should appear or at least the warning should tell me what to do.

## Actual Result

Warning confuses me & I want to go home

### Prometheus Operator Version

```shell
v0.71.2
```

### Kubernetes Version

```yaml
v1.28.6
```

### Kubernetes Cluster Type

kubeadm

### How did you deploy Prometheus-Operator?

yaml manifests

### Manifests

```yaml
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
metadata:
  labels:
    app.kubernetes.io/component: prometheus
    app.kubernetes.io/instance: autoscaling
    app.kubernetes.io/name: prometheus
    app.kubernetes.io/part-of: monitoring
  name: autoscaling
  namespace: prometheus
spec:
  additionalAlertManagerConfigs:
    key: prometheus-configuration
    name: alertmanager-configuration
    optional: false
  enableFeatures:
    - auto-gomaxprocs
  enableAdminAPI: true
  enforcedSampleLimit: 100
  evaluationInterval: 30s
  image: harbor.infra.run/dockerhub/prom/prometheus:v2.49.1
  imagePullSecrets:
    - name: harbor-infra-run-pull-secret
  nodeSelector:
    kubernetes.io/os: linux
  podMetadata:
    labels:
      app.kubernetes.io/component: prometheus
      app.kubernetes.io/instance: autoscaling
      app.kubernetes.io/name: prometheus
      app.kubernetes.io/part-of: monitoring
  podMonitorNamespaceSelector:
    matchLabels:
      autoscaling.prometheus.kube.infra.run/discovery: enabled
[...]

Issue #3468, opened 2020-09-03
Override rulefiles


janotav:
When prometheus-operator creates the final configuration for prometheus it puts prometheusrules into paths looking like:

/etc/prometheus/rules/prometheus-release-operator-prometheus-rulefiles-0/<**rule_metadata_namespace>-<rule_metadata_name>.yaml**

Is it possible to override this pattern? Can you please point me to where the relevant code resides?

My use-case is probably very specific and might be outside the prometheus-operator scope and I might have to resort to my own implementation, but I would like to know what are the options here.

Thanks.

janotav:
After some more digging I discovered that my use-case is kind of generalization of _prometheus.prometheusSpec.enforcedNamespaceLabel_

This can be used to "tag" the objects (including rules) with a label that will receive the name of the enclosing namespace.

In my scenario each namespace has its owner, but single owner can own multiple namespaces. The owner of a namespace is defined by a well-known label on the namespace:

```
kind: Namespace
metadata:
  name: namespace-name
  labels:
    owner: team-name
```
What I would like to do is to enforce the value of the well-known label (called _owner_ in this case) that represents ownership rather than the simple namespace name.

This would in turn allow me to route alerts from all namespaces to its owning team.

janotav:
I managed to to come to a reasonable solution for my scenario.

1. I patched prometheus-operator rule filename pattern to use "--" instead of "-" 
2. this allows me to reliably detect namespace from the rule_group label using regex (because in our case namespace name may contain "-" but must not contain "--")
3. this in turn allows me to "join" on the prometheusrule namespace in recording rule
4. finally alert is created on top of this recorded rule

As a result the owner of the namespace who owns the failing prometheusrule receives the alert. Small caveat is that due to the rule indirection, the alert takes a little longer to resolve, but that is negligible.

I am closing this because I assume that this use-case is probably too specific and the wider community would probably not benefit from the ability to override the delimiter in the rule filenames.

PR #6482, opened 2024-04-06
feat(rules): when naming rules files use clear namespace separation

diranged:
## Description

Ref: https://github.com/prometheus-operator/prometheus-operator/issues/3468

I ran into the same issue that @janotav did with [this comment](https://github.com/prometheus-operator/prometheus-operator/issues/3468#issuecomment-687038724). I want to be able to identify metrics from the Thanos Ruler around rule execution and failures and narrow them down to the `Namespace` and `PrometheusRule` that they come from. The problem is the current naming pattern of `$namespace-$rulename-$uid` doesn't account for resources that have `-`'s in the names... ie, the following rule creates a file named `my-namespace-some-rule-thing-xxx-xxx-xxx-xxx-xxx`:

```yaml
kind: PrometheusRule
metadata:
  name: some-rule-thing
  namespace: my-namespace
  resourceVersion: "1146"
  uid: xxx-xxx-xxx-xxx-xxx
```

The string `my-namespace-some-rule-thing-xxx-xxx-xxx-xxx-xxx` cannot be parsed through [metric relabelings](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs) because you cannot definitively pull out the `namespace` and `name` fields from the filename.

## Type of change

## Verification
Due to https://github.com/prometheus-operator/prometheus-operator/issues/6284 - I was unable to fully test this locally.

## Changelog entry

```release-note
Use "--" to separate namespace, resource name and UID in Prometheus Rules filenames
```

### DIFF
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -175,7 +175,7 @@ func (prs *PrometheusRuleSelector) Select(namespaces []string) (map[string]strin
 				return
 			}
 
-			promRules[fmt.Sprintf("%v--%v--%v.yaml", promRule.Namespace, promRule.Name, promRule.UID)] = promRule
+			promRules[fmt.Sprintf("%v,%v,%v.yaml", promRule.Namespace, promRule.Name, promRule.UID)] = promRule
 		})
 		if err != nil {
 			return nil, 0, fmt.Errorf("failed to list prometheus rules in namespace %s: %w", ns, err)
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -175,7 +175,7 @@ func (prs *PrometheusRuleSelector) Select(namespaces []string) (map[string]strin
 				return
 			}
 
-			promRules[fmt.Sprintf("%v-%v-%v.yaml", promRule.Namespace, promRule.Name, promRule.UID)] = promRule
+			promRules[fmt.Sprintf("%v--%v--%v.yaml", promRule.Namespace, promRule.Name, promRule.UID)] = promRule
 		})
 		if err != nil {
 			return nil, 0, fmt.Errorf("failed to list prometheus rules in namespace %s: %w", ns, err)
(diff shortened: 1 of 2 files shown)


diranged:
I see that these tests (https://github.com/prometheus-operator/prometheus-operator/actions/runs/8577200973/job/23509431752?pr=6482) failed - but I can't see how they relate to my change. Could they be flakey?

diranged:
@simonpasquier This is 3 weeks old now - thoughts on what we can do here to move this along?

simonpasquier:
I'm not 100% fan of the `--` separator, could we use `,` to have a single char separator?

diranged:
I don't _really_ care what the separator is ... but I've just seen `--` used in other places. I can't say I am a fan of a `,` in a filename.. either way though, it does actually work. If you want me to use a `,` that's fine with me I guess,..

simonpasquier (dismissed):
Thanks!

diranged:
@simonpasquier So what do we do next to get this merged?

slashpai:
e2e test are failing, looks like you would need to update https://github.com/prometheus-operator/prometheus-operator/blob/53c9ed628dae20d3e3df2cb20fbd7ddba37b2581/test/e2e/prometheus_test.go#L1736 as well

simonpasquier:
I tried the PR and understand why the e2e tests fail:

````
level=error [...] ConfigMap \"prometheus-prometheus-rulefiles-0\" is invalid: data[default,prometheus-example-rules,96018099-a62f-4470-99fa-042bc53b3361.yaml]: Invalid value: \"default,prometheus-example-rules,96018099-a62f-4470-99fa-042bc53b3361.yaml\": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')"
````

simonpasquier:
Can we simply use `_` as a the separator? AFAICT it's a forbidden character for name and namespace so it shouldn't lead to ambiguity.

Issue #4241, opened 2021-09-02
Alertmanager mTLS config


TamasNeumer:
**What is missing?**

AlertManager has recently introduced configuration options for mTLS:
- https://github.com/prometheus/alertmanager/commit/ff85bec45bdce3d296a229adb47e7d29415f8a05
- https://github.com/prometheus/alertmanager/blob/main/docs/https.md

**Why do we need it?**

To be able to configure AlertManager gossip to use mTLS.
Would be nice in general to be able to add flags to the AlertManager container itself.

**Anything else we need to know?**:
I think [in the alertmanager statefulset generator](https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/alertmanager/statefulset.go#L226-L278) we don't have any option for this.

TamasNeumer:
@simonpasquier @brancz 

I'm happy to help and implement the feature, however some guidance would be helpful.

- As far as I see an additional field needs to be added to the `AlertmanagerSpec` API which should point to a `ConfigMap`. The `ConfigMap` is mounted as a volume.
- Then it's the user's responsibility to deploy the `ConfigMap` with the desired content
- We then add an additional (Second?) reloader which targets the the `ConfigMap`?

Thanks!

simonpasquier:
It would be more friendly to expose the different settings in the Alertmanager CRD rather than forcing the user to craft the config manually. You may want to have a look at what exists for web TLS config (https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#webtlsconfig).
cc @fpetkovski who worked on the implementation for web TLS.

mviswanathsai:
Is this still relevant?

simonpasquier:
yes

mviswanathsai:
Taking this up!

mviswanathsai:
I need to find what the command line flags are, the link that the OP provided don't cut it for me. @simonpasquier could you help me here? I want to get the PR out.

mviswanathsai:
https://github.com/prometheus/alertmanager#high-availability Would this be the place to look for a wider perspective?

simonpasquier:
@mviswanathsai I would advise that we add `additionalArgs` to the Alertmanager CRD like we have for Prometheus, PrometheusAgent and ThanosRuler. It will provide flexibility at a low cost (I'm not against adding the fields that you've mentioned but most of them are probably niche use cases).

PR #4593, opened 2022-02-23
[WIP] add MTLS config for alert manager

raptorsun:
## Description

This PR adds MTLS config to AlertManager, allowing user to configure gossip between alert managers using mTLS for authentication.
This resolves [issue 4241](https://github.com/prometheus-operator/prometheus-operator/issues/4241).

The work is half done only, I will complete the PR later :) 

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [x] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #6489, opened 2024-04-09
[WIP] Feat: Add clusterTLSConfigs to the Alertmanager CRD

mviswanathsai:
## Description

This PR adds mTLS config to AlertManager, allowing user to configure gossip between alert managers using mTLS for authentication.
Closes: #4241

## Type of change

- [ x] `FEATURE` (non-breaking change which adds functionality)

## Verification

## Changelog entry

```
add mTLS configuration to alertmanager CRD
```


mviswanathsai:
Questions:
1. What about concurrency and timeouts, that we have in `AlertmanagerWebConfig`? Would that apply here?
2. I am still not sure about the naming, feedback is much needed.
3. What about "de-duplication" logic in #4593? would we need to have different naming conventions for server and client tls creds? clarity is needed in this area. 
4. Would like suggestions if any, for the code yet to be written. For example, I am writing the code in alertmanager/statefulset.go to include the volumes, mounts and args for the clusterTLSConfig (mTLSConfig), on the condition that the version of prometheus is >=0.27.0. This is yet to be pushed.

mviswanathsai:
Requesting review: @simonpasquier @ArthurSens @nicolastakashi @slashpai. 

Some concepts are still not clear to me, especially stuff that is not directly related to the code itself but the cluster tls config in general. I think the review will help a lot. Kindly bear with me, if I am not up to speed!

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -240,6 +240,10 @@ type AlertmanagerSpec struct {
 	HostAliases []HostAlias `json:"hostAliases,omitempty"`
 	// Defines the web command line flags when starting Alertmanager.
 	Web *AlertmanagerWebSpec `json:"web,omitempty"`
+
+	// Defines the cluster tls configuration for Alertmanager mTLS gossip
```
simonpasquier, line 244:
```suggestion
	// Configures the mutual TLS configuration for the Alertmanager cluster's gossip protocol.
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -240,6 +240,10 @@ type AlertmanagerSpec struct {
...
+	//+optional
+	ClusterTLSConfig *AlertmanagerMTLSConfig `json:"clusterTLSConfigs,omitempty"`
```
simonpasquier, line 246:
```suggestion
	ClusterTLSConfig *AlertmanagerTLSConfig `json:"clusterTLSConfig,omitempty"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -379,6 +379,160 @@ func (c *WebTLSConfig) Validate() error {
 	return nil
 }
 
+// MTLSConfig fields defines the contents for the --cluster.tls-config flag
```
simonpasquier, line 382: It can all live in pkg/apis/monitoring/v1/alertmanager_types.go

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -369,6 +373,11 @@ type AlertmanagerWebSpec struct {
 	Timeout *uint32 `json:"timeout,omitempty"`
 }
 
+// TODO: What other fields do we need?
+type AlertmanagerMTLSConfig struct {
+	MTLSConfigFields `json:",inline"`
```
simonpasquier, line 378: I'm not sure why we would need an inline field.

```suggestion
	// Server-side configuration for mutual TLS.
	Server *ServerTLSConfig `json:"server"`
	// Client-side configuration for mutual TLS.
	Client *ClientTLSConfig `json:"client"`
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -379,6 +379,160 @@ func (c *WebTLSConfig) Validate() error {
...
+	}
+
+	if c.TLSKeySecret == (v1.SecretKeySelector{}) {
+		return &ServerTLSConfigError{"invalid server TLS config - keySecret: key must be defined"}
+	}
+
+	return nil
+}
+
+type ClientTLSConfig struct {
```
simonpasquier, line 488: Do we need a dedicated struct? Can't we use SafeTLSconfig?

pkg/apis/monitoring/v1/types.go
```diff
@@ -379,6 +379,160 @@ func (c *WebTLSConfig) Validate() error {
...
+	} else {
+		return &MTLSConfigError{err: "client TLS config is required"}
+	}
+
+	return nil
+
+}
+
+// ServerTLSConfig is Alertmanager's server side mutual TLS config.
+type ServerTLSConfig struct {
```
simonpasquier, line 423: Can't we reuse WebTLSConfig as it seems like a 1:1 mapping?

pkg/apis/monitoring/v1/alertmanager_types.go
simonpasquier: We should also reflect that it's an experimental feature.
Also worth documenting that it requires Alertmanager >= 0.24.0.

pkg/alertmanager/statefulset.go
```diff
@@ -669,6 +670,30 @@ func makeStatefulSetSpec(logger log.Logger, a *monitoringv1.Alertmanager, config
 		}
 	}
 
+	// Mount web config and web TLS credentials as volumes.
+	// We always mount the web config file for versions greater than 0.22.0.
+	// With this we avoid redeploying alertmanager when reconfiguring between
+	// HTTP and HTTPS and vice-versa.
+	if version.GTE(semver.MustParse("0.27.0")) {
```
simonpasquier, line 677: 0.22.0

pkg/alertmanager/statefulset.go
```diff
@@ -669,6 +670,30 @@ func makeStatefulSetSpec(logger log.Logger, a *monitoringv1.Alertmanager, config
 		}
 	}
 
+	// Mount web config and web TLS credentials as volumes.
+	// We always mount the web config file for versions greater than 0.22.0.
+	// With this we avoid redeploying alertmanager when reconfiguring between
+	// HTTP and HTTPS and vice-versa.
```
simonpasquier, line 676: I don't think that the comment applies here because an empty mTLS config file isn't allowed. So we'll have to redeploy the pods if the spec changes from no-TLS to TLS (and vice-versa).

pkg/alertmanager/statefulset.go
mviswanathsai: I am confused, is it 0.22.0 or 0.24.0?

pkg/alertmanager/statefulset.go
mviswanathsai: Right, so we cannot mount an empty volume like we do with web tls. So here, we only mount the config and redeploy the pod depending on whether there is mTLS config or not.

pkg/alertmanager/statefulset.go
simonpasquier: yes 0.24.0

pkg/alertmanager/statefulset.go
mviswanathsai: // If the am version >= 0.24.0, create a new secret for the mTLS configuration.
	// The secret volume is always created and mounted, even if the secret is empty.
	// But, empty volume cannot be passed as the path in the argument for the cluster TLS configuration.
	// So, we only pass the argument `--cluster.tls-config` if the ClusterTLSConfig field is not nil.

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -369,6 +376,40 @@ type AlertmanagerWebSpec struct {
...
+		if err := c.Server.Validate(); err != nil {
+			return fmt.Errorf("am mtls: server config, %w", err)
+		}
+	} else {
+		return fmt.Errorf("am mtls: server config is required")
+	}
+	if c.Client != nil {
+		if err := c.Client.Validate(); err != nil {
+			return fmt.Errorf("am mtls: client config, %w", err)
+		}
```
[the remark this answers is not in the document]
mviswanathsai: I have added some checks. However, I do not quite understand why this does not apply for web TLS. Could you briefly touch upon that?

mviswanathsai:
Background:
When configuring TLS credentials for both the server and the clients, there was an issue where the same names or parameters could be assigned to both, leading to duplication. To handle this, a "de-duplication" logic was implemented. This logic ensured that secrets with the same name were not added more than once to the list of parameters.

Problem with De-duplication:
The "de-duplication" mechanism seems to assume that the secrets with the same name, regardless of being part of the client's tls creds or the server's are the same. I am unable to verify if this is right or not. So, to avoid any confusions, I adopted the following alternative approach.

Alternative Approach:
To avoid potential issues with de-duplication, a new approach was adopted. Unique suffixes are now assigned to the TLS credentials for the server and the clients. This ensures that each set of credentials is distinct, eliminating the risk of duplication without the need for additional logic. This logic can be seen in action [here](https://github.com/prometheus-operator/prometheus-operator/pull/6489/files#diff-1e05d85dffd824a4db6711f75ff178e5f9155f09345d5475ef65d3f02fc8c371R85).

simonpasquier:
Can you rebase on top of main to get #6612?

pkg/alertmanager/statefulset.go
```diff
@@ -675,6 +676,35 @@ func makeStatefulSetSpec(logger log.Logger, a *monitoringv1.Alertmanager, config
...
+	// The secret volume is always created and mounted, even if the secret is empty.
+	// But, empty volume cannot be passed as the path in the argument for the cluster TLS configuration.
+	// So, we only pass the argument `--cluster.tls-config` if the ClusterTLSConfig field is not nil.
+	if version.GTE(semver.MustParse("0.24.0")) {
+		var mtlsFields *monitoringv1.AlertmanagerTLSConfigFields
+		var isEmpty bool
+		if a.Spec.ClusterTLSConfig != nil {
+			isEmpty = false
+			mtlsFields = a.Spec.ClusterTLSConfig
+		}
```
simonpasquier, line 689:
```suggestion
		clusterTLS := a.Spec.ClusterTLSConfig
```

pkg/alertmanager/statefulset.go
```diff
@@ -675,6 +676,35 @@ func makeStatefulSetSpec(logger log.Logger, a *monitoringv1.Alertmanager, config
...
+
+		confArg, configVol, configMount, err := mtlsConfig.GetMountParameters()
+		if err != nil {
+			return nil, err
+		}
+		volumes = append(volumes, configVol...)
+		amVolumeMounts = append(amVolumeMounts, configMount...)
+
+		// Append the argument only if the volume contents is not nil.
+		if !isEmpty {
```
simonpasquier, line 703:
```suggestion
		if clusterTLS != nil {
```

pkg/webconfig/mtls_config.go
simonpasquier, line 1: For now the mTLS part is specific to Alertmanager, unlike the web TLS config which is common across controllers. I'd rather move the code to pkg/alertmanager for now. Though the goal should also be to reuse existing functionality for the TLS server config rather than duplicating the code.

pkg/alertmanager/statefulset.go
mviswanathsai: Woah! How did I find the most roundabout way to do this :)

pkg/webconfig/mtls_config.go
mviswanathsai: I have moved it to pkg/alertmanager and also made changes to tls_credentials.go where we needed public functions. I am a little unsure over `NewTLSCredentials` function which was removed as part of #6612.

mviswanathsai:
I reckon we also need to add tests, I will get to it once we are happy with the current changes.

mviswanathsai:
Currently caught up with other things, will get back to this PR once things higher on my priority list are done!

PR #6492, opened 2024-04-09
Add extra metric relabelings to scrape classes

sathieu:
## Description

See https://github.com/prometheus-operator/prometheus-operator/issues/5947#issuecomment-1845611166

Inspired by #6379.

## Type of change

## Verification

## Changelog entry

```release-note
Add extra metric relabelings to scrape classes
```


simonpasquier (commented):
The main issue for me is that the scrape class metrics relabeling happens after the scrape object's metrics relabeling. Other scrape class fields behave differently and can be overridden by scrape objects. I suppose that you have a specific requirement in mind but the inconsistency brings a UX problem.

sathieu:
@simonpasquier Yes. This is described in the doc, but if this is not enough we can use two configurations: `metricRelabelingsBefore` and `metricRelabelingsAfter` (I'm not good at naming)., or `metricRelabelings` and `metricRelabelingsOverride`, ....

simonpasquier:
Can you share more details about why you need scrape class relabeling to happen after service/pod monitor relabelings rather than before?

sathieu:
Yes.

As described [in the issue](https://github.com/prometheus-operator/prometheus-operator/issues/5947#issuecomment-1845611166), we have multi-tenant clusters, and each tenant is defined by a list of namespaces or a list of namespace regexps. We want the metrics relabeling to happen after the namespace label enforcement, to avoid a tenant injecting metrics into another tenant.

If scrape class metric relabelings happen before scrape resource metric relabelings, a malicious actor can easily change the `tenant` label.

simonpasquier:
> As described https://github.com/prometheus-operator/prometheus-operator/issues/5947#issuecomment-1845611166, we have multi-tenant clusters, and each tenant is defined by a list of namespaces or a list of namespace regexps. We want the metrics relabeling to happen after the namespace label enforcement, to avoid a tenant injecting metrics into another tenant.

To be clear: you're already setting `.spec.enforcedNamespaceLabel: "namespace"`? 
If yes, how do you deal with recording/alerting rules? Do you expect to enforce the tenant label there too?

simonpasquier:
> I forgot about those, I also need the tenant label for recording/alerting rules. 

:thinking: I don't see an obvious solution for rules since relabeling can't be applied here...

> Maybe a new `.spec.enforceMetricRelabelings` in the prometheus CRD would be more appropriate (please note that I don't want `excludedFromEnforcement` to be applied here).

I suppose that the latter is because you deploy generic rules generating namespace-aware metrics/alerts and you still want the metrics/alerts to have the right tenant?

There might still be other use cases that need to skip "enforced" metric relabelings for some monitor/rule resources though.

simonpasquier:
Nevertheless this PR (with scrape class' relabeling happening before monitor's metric relabeling) is worth the effort even if it's not addressing @sathieu requirements.

sathieu:
I've updated the PR to :

```go
	// The Operator adds the scrape class metric relabelings defined here.
	// Then the Operator adds the target-specific metric relabelings defined in the scrape object.
	// Then the Operator adds namespace enforcement relabeling rule.
```

I'll handle my usecase in another PR.

ArthurSens (commented):
Hi @sathieu, sorry for taking so long to give you an answer here, I saw the failing tests and assumed the PR was not ready. I took a deeper look today and I _think_ the failing tests are flaky ones. I've re-triggered them to be sure... let's see

sathieu:
Thanks @ArthurSens I've added your suggestion. I also had to rebase to resolve conflicts. Back to you!

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1840,4 +1840,15 @@ type ScrapeClass struct {
 	//
 	// +optional
 	Relabelings []RelabelConfig `json:"relabelings,omitempty"`
+
+	// MetricRelabelings configures the relabeling rules to apply to all samples before injection.
```
nicolastakashi, line 1844: Do you mean ingestion?

sathieu: Yes :syringe: . Fixed, thanks.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1840,4 +1840,15 @@ type ScrapeClass struct {
 	//
 	// +optional
 	Relabelings []RelabelConfig `json:"relabelings,omitempty"`
+
+	// MetricRelabelings configures the relabeling rules to apply to all samples before ingestion.
+	//
+	// The Operator adds the scrape class metric relabelings defined here.
+	// Then the Operator adds the target-specific metric relabelings defined in the scrape object.
```
ArthurSens, line 1847:
```suggestion
	// Then the Operator adds the target-specific metric relabelings defined in ServiceMonitors, PodMonitors, Probes and ScrapeConfigs.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1840,4 +1840,15 @@ type ScrapeClass struct {
...
+	// Then the Operator adds namespace enforcement relabeling rule.
```
ArthurSens, line 1848:
```suggestion
	// Then the Operator adds namespace enforcement relabeling rule, specified in '.spec.enforcedNamespaceLabel'.
```

pkg/prometheus/promcfg_test.go
```diff
@@ -8783,3 +8783,205 @@ func TestPodMonitorWithNonDefaultScrapeClassRelabelings(t *testing.T) {
 	require.NoError(t, err)
 	golden.Assert(t, string(cfg), "podMonitorObjectWithNonDefaultScrapeClassWithRelabelings.golden")
 }
+
+func TestServiceMonitorWithDefaultScrapeClassMetricRelabelings(t *testing.T) {
```
ArthurSens, line 8787: Awesome tests!

We could improve code readability here if we reduce code duplication using [table tests](https://dev.to/boncheff/table-driven-unit-tests-in-go-407b)

ArthurSens (dismissed):
Awesome, thanks for tackling this!

I've made a few nitpick comments, but if you want to focus on something I'd say improving test readability is what could receive some love right now.

Approving the PR because this seems to work as expected

pkg/prometheus/promcfg_test.go
```diff
@@ -8783,3 +8783,205 @@ func TestPodMonitorWithNonDefaultScrapeClassRelabelings(t *testing.T) {
...
+		nil,
+		nil,
+		&assets.StoreBuilder{},
+		nil,
+		nil,
+		nil,
+		nil,
+	)
+	require.NoError(t, err)
+	golden.Assert(t, string(cfg), "serviceMonitorObjectWithDefaultScrapeClassWithMetricRelabelings.golden")
```
ArthurSens, line 8844: What I mean with code duplication is that this part is basically identical in all tests. We could have a single function with different inputs instead

```go
testCases := []struct{
  scrapeClass monitoringv1.ScrapeClass
  podMonitors map[string]monitoringv1.Podmonitor // can be nil
  serviceMonitors map[string]monitoringv1.ServiceMonitor // can be nil
  probes map[string]monitoringv1.Probe // can be nil
  scrapeConfigs map[string]monitoringv1alpha1.ScrapeConfig // can be nil
  goldenFile string  
}
```
You can put the implementation of the test inside a loop and asser that your inputs results are equal to the appropriate golden file

nicolastakashi (dismissed):

sathieu:
@nicolastakashi I've addressed your comments, rebased (there was a conflict because of 642d9cd3fe343251edd3219eeb9cd84c8a8a6856) and force-pushed. Back to you :basketball: !

ArthurSens (approved):
Perfect, thanks @sathieu !

Issue #6270, opened 2024-01-26
Allow turning off automountServiceAccountToken for Prometheus SA


oscarbm1998:
### Component(s)

Prometheus, Prometheus-Operator

### What is missing? Please describe.

We are currently running a kubernetes cluster which uses the kube-prom-stack for monitoring. For security reasons, we are currently turning off automounting serviceAccountTokens by default in all of our deployments including kube-prom-stack and if the Deployment really requires that, we put the corresponding mount manually into the requiring containers again. Unfortunately, this is not possible for the prometheus system, as there is no option to provide this as part of the config values for prometheus. Therefore the prometehus CRD has to be updated to provide the option to set automountServiceAccountToken: False and the prometheus-operator has to be updated to not default it to True. 

I created a github issue for this over in the prometheus community helm charts section which is linked here: https://github.com/prometheus-community/helm-charts/issues/4148

There it was responed by a user that they could update the CRD for us, but the operator would default the prometheus argument AutoMountServiceToken to true anyways. Therefore he said that i also had to create a github issue here, so the prometeheus operator could be updated to not auto default prometheus to AutoMountServiceToken. Link to that respond chain here: https://github.com/prometheus-community/helm-charts/pull/4170

## Environment
Kubernetes Version: 1.27.7
Prometheus-Operator Version: v0.70.0

oscarbm1998:
The security concers comes from general research on the issue from the internet + recommendations from Azure Defender to disable it as can be seen here:

"Kubernetes clusters should disable automounting API credentials" - Disable automounting API credentials to prevent a potentially compromised Pod resource to run API commands against Kubernetes clusters. For more information, see https://aka.ms/kubepolicydoc.

We are already mounting the service account manually to other parts of the kube-prom-stack such as the AlertManager which has the option of turning the automountServiceAccountToken: False and therefore works. However for prometheus the Prometheus Operator always defaults the automountServiceAccountToken to True even if we want it to be false as can be seen here: https://github.com/prometheus-community/helm-charts/pull/4170#discussion_r1462370431. Therefore we want the option to disable it in the operator.

simonpasquier:
The aka.ms link doesn't work for me but https://kyverno.io/policies/other/restrict-automount-sa-token/restrict-automount-sa-token/ is another stance on the topic. So yes, I agree that it makes sense to have the option.

[...]

PR #6497, opened 2024-04-10
fix: make automountServiceAccountToken configurable

stefan-caraiman:
## Description

Fixes https://github.com/prometheus-operator/prometheus-operator/issues/6270

## Type of change

## Verification

## Changelog entry

```release-note
Make `automountServiceAccountToken` configurable for the Prometheus Operator StatefulSet
```

### DIFF
--- a/pkg/apis/monitoring/v1/prometheus_types.go
+++ b/pkg/apis/monitoring/v1/prometheus_types.go
@@ -356,6 +356,14 @@ type CommonPrometheusFields struct {
 	// Prometheus Pods.
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
+	// AutomountServiceAccountToken indicates whether a service account token should be automatically mounted in the pod.
+	// If the field isn't set, the operator mounts the service account token by default.
+	//
+	// **Warning:** be aware that by default, Prometheus requires the service account token for Kubernetes service discovery.
+	// It is possible to use strategic merge patch to project the service account token into the 'prometheus' container.
+	// +optional
+	AutomountServiceAccountToken *bool `json:"automountServiceAccountToken,omitempty"`
+
 	// Secrets is a list of Secrets in the same namespace as the Prometheus
 	// object, which shall be mounted into the Prometheus Pods.
 	// Each Secret is added to the StatefulSet definition as a volume named `secret-<secret-name>`.
--- a/pkg/prometheus/agent/statefulset.go
+++ b/pkg/prometheus/agent/statefulset.go
@@ -370,7 +370,7 @@ func makeStatefulSetSpec(
 				InitContainers:               initContainers,
 				SecurityContext:              cpf.SecurityContext,
 				ServiceAccountName:           cpf.ServiceAccountName,
-				AutomountServiceAccountToken: ptr.To(true),
+				AutomountServiceAccountToken: ptr.To(ptr.Deref(cpf.AutomountServiceAccountToken, true)),
 				NodeSelector:                 cpf.NodeSelector,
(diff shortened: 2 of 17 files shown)


pkg/prometheus/agent/statefulset.go
```diff
@@ -370,7 +370,7 @@ func makeStatefulSetSpec(
 				InitContainers:               initContainers,
 				SecurityContext:              cpf.SecurityContext,
 				ServiceAccountName:           cpf.ServiceAccountName,
-				AutomountServiceAccountToken: ptr.To(true),
+				AutomountServiceAccountToken: cpf.AutomountServiceAccountToken,
```
simonpasquier, line 373: I expect that we preserve the existing behavior when the field isn't defined.
```suggestion
				AutomountServiceAccountToken: ptr.Deref(cpf.AutomountServiceAccountToken, true),
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -356,6 +356,11 @@ type CommonPrometheusFields struct {
 	// Prometheus Pods.
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
+	// AutomountServiceAccountToken indicates whether a service account token should be automatically mounted in the pod.
+	// If the service account has `automountServiceAccountToken: true`, set the field to `false` to opt out of automounting API credentials.
```
simonpasquier, line 360: There should be a **strong indication** that it can impact Kubernetes service discovery if you don't inject the bearer token manually.

```suggestion
	// If the field isn't set, the operator mounts the service account token by default.
	//
	// **Warning:** be aware that by default, Prometheus requires the service account token for Kubernetes service discovery.
	// It is possible to use strategic merge patch to project the service account token into the 'prometheus' container.
```

pkg/prometheus/agent/statefulset.go
stefan-caraiman: not sure if it works 😅  will result in the following error (not that familiar with Go) :

```
cannot use ptr.Deref(cpf.AutomountServiceAccountToken, true) (value of type bool) as *bool value in struct literalcompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)
```

Got inspired by the spec setup from Alertmanager https://github.com/prometheus-operator/prometheus-operator/blob/e3f9f9269baf61d344c176a3ddff767d7c425e6c/pkg/alertmanager/statefulset.go#L783  where if omitted it defaults to true. Let me know if it still needs patching 😄

stefan-caraiman: @simonpasquier thank you for the suggestion, patched ✅

stefan-caraiman:
@simonpasquier let me know if you want me to update anything else 👍

simonpasquier (commented):
Ideally I'd like to see new unit tests in `pkg/prometheus/server/statefulset_test.go` and `pkg/prometheus/agent/statefulset_test.go`.
If it's too much burden for you, I can add them for you but we'd have to wait a bit.

stefan-caraiman:
> Ideally I'd like to see new unit tests in `pkg/prometheus/server/statefulset_test.go` and `pkg/prometheus/agent/statefulset_test.go`. If it's too much burden for you, I can add them for you but we'd have to wait a bit.

unit tests added 👍  tried following the same pattern as with other unit tests, hopefully it makes sense

simonpasquier:
You'd need to run `make --always-make format generate` and commit the changes too.

stefan-caraiman:
done ✅

PR #6505, opened 2024-04-12
Consider draft PRs electable for staleness

ArthurSens:
## Description

I was going through the list of open PRs today and noticed that we have many abandoned draft PRs where the majority have merge conflicts.

I was confused since we have workflows closing stale issues and PRs, until I noticed that we have an exception for drafts in our workflow.

I'm creating this PR to start a discussion about whether we want to close stale draft PRs as well. 
(I'm in favor)

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -15,7 +15,7 @@ jobs:
           days-before-stale: 60
           days-before-issue-close: 120
           days-before-pr-close: 180
-          exempt-issue-labels: 'kind/feature,help wanted,kind/bug,kind/documentation'
+          exempt-issue-labels: 'kind/feature,help wanted,kind/bug,kind/documentation,needs-triage'
           stale-issue-label: 'stale'
           stale-pr-label: 'stale'
           operations-per-run: 500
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
It seems fair to close draft PRs.

simonpasquier:
> Additionally, we also have this configuration on the workflow
> 
> ```yaml
> exempt-issue-labels: 'kind/feature,help wanted,kind/bug,kind/documentation'
> ```
> 
> We recently changed our issue template to always add one of the labels above. This means that all recent issues will never be stale. 🤔

When triaging, we should always evaluate whether the original label makes sense. For instance, a reported bug might be a support question and we should change the label to `kind/support`. The purpose of the `exempt-issue-labels` rule is for issues that have been recognized as valid issues to never go stale.
And we should probably add needs-triage to the list.

.github/workflows/stale.yaml
```diff
@@ -18,5 +18,4 @@ jobs:
           exempt-issue-labels: 'kind/feature,help wanted,kind/bug,kind/documentation'
```
slashpai, line 18:
```suggestion
          exempt-issue-labels: 'kind/feature,help wanted,kind/bug,kind/documentation,needs-triage'
```

simonpasquier (approved):

simonpasquier (approved):

Issue #6548, opened 2024-04-25
kubelet endpoint contains IP addresses of nodes with Ready condition Unknown


tstringer-fn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

## Description

We have found that the kubelet controller does not make any node condition checks prior to adding their addresses to the kubelet endpoint. Normally, this is not an issue as it ends in the endpoint target being marked as "down" and scraping fails because the node is not ready.

A much larger symptom occurs when there is an IP address that is reused from a down/NotReady node in the cluster (we have seen this exact scenario in GKE). For instance, node1 can be `NotReady` with IP address 1.2.3.4 and the underlying provisioner creates a new node, node2, reusing IP address 1.2.3.4.

During this scenario, because kubelet controller doesn't check for node status it will add two endpoint addresses with the same IP address. An example of subsets might look like this:

```
  - ip: 1.2.3.4
    targetRef:
      kind: Node
      name: node1
      uid: uid1
  - ip: 1.2.3.4
    targetRef:
      kind: Node
      name: node2
      uid: uid2
```

In this case, node1 is down (NotReady) and node2 is provisioned with the same IP address. Currently, prometheus-operator will add both node1 and node2 to the kubelet endpoint as you can see above. Given this, both of these subsets will be scraped and they will both succeed because node2 responds successfully from both scrapes to 1.2.3.4. The time series will have labels based off of their metadata. If metric1 is scraped from node2, it will be scraped twice and the only difference between the time series is the node label.

This is an instance where we get wrong and duplicate data.

## Steps to Reproduce

Unfortunately, it is difficult to reproduce the issue. You'd have to set a node as `NotReady` (which isn't impossible) but then have a node provisioner create another node with the same IP address as the `NotReady` node.

## Expected Result

Prometheus operator should not add a `NotReady` node's IP address to the kubelet endpoints.

## Actual Result

Prometheus operator adds a `NotReady` node's IP address to the kubelet endpoint.

### Prometheus Operator Version

```shell
v0.66.0
```

### Kubernetes Version

```yaml
clientVersion:
  buildDate: "2023-06-14T09:56:58Z"
  compiler: gc
  gitCommit: 11902a838028edef305dfe2f96be929bc4d114d8
  gitTreeState: clean
  gitVersion: v1.26.6
  goVersion: go1.19.10
  major: "1"
  minor: "26"
  platform: darwin/arm64
kustomizeVersion: v4.5.7
serverVersion:
[...]

PR #6549, opened 2024-04-25
Only add node IPs to kubelet endpoint if they have a known Ready condition

tstringer-fn:
## Description

We have found that the kubelet controller does not make any node condition checks prior to adding their addresses to the kubelet endpoint. Normally, this is not an issue as it ends in the endpoint target being marked as "down" and scraping fails because the node is not ready.

A much larger symptom occurs when there is an IP address that is reused from a down/NotReady node in the cluster (we have seen this exact scenario in GKE). For instance, node1 can be `NotReady` with IP address 1.2.3.4 and the underlying provisioner creates a new node, node2, reusing IP address 1.2.3.4.

During this scenario, because kubelet controller doesn't check for node status it will add two endpoint addresses with the same IP address. An example of subsets might look like this:

```
  - ip: 1.2.3.4
    targetRef:
      kind: Node
      name: node1
      uid: uid1
  - ip: 1.2.3.4
    targetRef:
      kind: Node
      name: node2
      uid: uid2
```

In this case, node1 is down (NotReady) and node2 is provisioned with the same IP address. Currently, prometheus-operator will add both node1 and node2 to the kubelet endpoint as you can see above. Given this, both of these subsets will be scraped and they will both succeed because node2 responds successfully from both scrapes to 1.2.3.4. The time series will have labels based off of their metadata. If metric1 is scraped from node2, it will be scraped twice and the only difference between the time series is the node label.

This is an instance where we get wrong and duplicate data.

## Type of change

## Verification

To verify this change I did made a slight modification to [this kind-with-registry.sh](https://kind.sigs.k8s.io/docs/user/local-registry/) script, and now it creates 3 local kind nodes.

<details>
<summary>kind-with-registry.sh (modified)</summary>

```bash
#!/bin/sh
set -o errexit

# 1. Create registry container unless it already exists
reg_name='kind-registry'
reg_port='5001'
if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
  docker run \
    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --network bridge --name "${reg_name}" \
    registry:2
fi

# 2. Create kind cluster with containerd registry config dir enabled
# TODO: kind will eventually enable this by default and this patch will
# be unnecessary.
#
# See:
# https://github.com/kubernetes-sigs/kind/issues/2875
# https://github.com/containerd/containerd/blob/main/docs/cri/config.md#registry-configuration
# See: https://github.com/containerd/containerd/blob/main/docs/hosts.md
cat <<EOF | kind create cluster --config=-
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
  - role: worker
  - role: worker
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri".registry]
    config_path = "/etc/containerd/certs.d"
EOF

# 3. Add the registry config to the nodes
#
# This is necessary because localhost resolves to loopback addresses that are
# network-namespace local.
# In other words: localhost in the container is not localhost on the host.
#
# We want a consistent name that works from both ends, so we tell containerd to
# alias localhost:${reg_port} to the registry container when pulling images
REGISTRY_DIR="/etc/containerd/certs.d/localhost:${reg_port}"
for node in $(kind get nodes); do
  docker exec "${node}" mkdir -p "${REGISTRY_DIR}"
  cat <<EOF | docker exec -i "${node}" cp /dev/stdin "${REGISTRY_DIR}/hosts.toml"
[host."http://${reg_name}:5000"]
EOF
done

# 4. Connect the registry to the cluster network if not already connected
# This allows kind to bootstrap the network but ensures they're on the same network
if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")" = 'null' ]; then
  docker network connect "kind" "${reg_name}"
fi

# 5. Document the local registry
# https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: local-registry-hosting
  namespace: kube-public
data:
  localRegistryHosting.v1: |
    host: "localhost:${reg_port}"
    help: "https://kind.sigs.k8s.io/docs/user/local-registry/"
EOF
```
</details>

The verification script then does the following:

1. Build and push my changes to a local registry: `TAG=latest IMAGE_OPERATOR=localhost:5001/prometheus-operator/prometheus-operator make image && docker push localhost:5001/prometheus-operator/prometheus-operator:latest`
2. Run the above `kind-with-registry.sh` script to create a three node cluster.
3. Stop kubelet on `kind-worker2` node: `docker exec kind-worker2 /bin/bash -c "systemctl stop kubelet"`
4. Wait for `kind-worker2` node to stop being ready:
```bash
while true; do
	echo "Waiting for not Ready status"
	STATUS=$(kubectl get no kind-worker2 -o json | jq '.status.conditions[] | select(.type == "Ready") | .status' -r)
	echo "Node status: $STATUS"
	if [[ "$STATUS" != "True" ]]; then
		echo "✅ NotReady status found"
		kubectl get no kind-worker2
		break
	fi
done
```
5. Install kube-prometheus-stack helm chart but specify the new local image:
```bash
helm upgrade \
	--install \
	kps \
	--set prometheusOperator.image.registry=localhost:5001 \
	--set prometheusOperator.image.repostory=prometheus-operator \
	--set prometheusOperator.image.tag=latest \
	--set prometheusOperator.image.pullPolicy=Always \
	prometheus-community/kube-prometheus-stack
```
6. Port-forward the prometheus service: `kubectl port-forward svc/kps-kube-prometheus-stack-prometheus 9090`
7. Scrape prometheus kubelet targets: `curl -s 'localhost:9090/api/v1/targets?scrapePool=serviceMonitor%2Fdefault%2Fkps-kube-prometheus-stack-kubelet%2F0' | jq '.data.activeTargets[] | \"\(.scrapeUrl): \(.health)\"' -r`

Before the fix, we can see the node is down:

```bash
kubectl get no -o wide
NAME                 STATUS     ROLES           AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                         KERNEL-VERSION                        CONTAINER-RUNTIME
kind-control-plane   Ready      control-plane   2m58s   v1.29.2   192.168.228.4   <none>        Debian GNU/Linux 12 (bookworm)   6.7.11-orbstack-00143-ge6b82e26cd22   containerd://1.7.13
kind-worker          Ready      <none>          2m36s   v1.29.2   192.168.228.5   <none>        Debian GNU/Linux 12 (bookworm)   6.7.11-orbstack-00143-ge6b82e26cd22   containerd://1.7.13
kind-worker2         NotReady   <none>          2m34s   v1.29.2   192.168.228.2   <none>        Debian GNU/Linux 12 (bookworm)   6.7.11-orbstack-00143-ge6b82e26cd22   containerd://1.7.13
```

And we can see that this endpoint IP is being scraped, and down.

```
curl -s 'localhost:9090/api/v1/targets?scrapePool=serviceMonitor%2Fdefault%2Fkps-kube-prometheus-stack-kubelet%2F0' | jq '.data.activeTargets[] | "\(.scrapeUrl): \(.health)"' -r
https://192.168.228.4:10250/metrics: up
https://192.168.228.5:10250/metrics: up
https://192.168.228.2:10250/metrics: down
```

After the fix, we can still see the node is down:

```bash
kubectl get no -o wide
NAME                 STATUS     ROLES           AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                         KERNEL-VERSION                        CONTAINER-RUNTIME
kind-control-plane   Ready      control-plane   7m37s   v1.29.2   192.168.228.4   <none>        Debian GNU/Linux 12 (bookworm)   6.7.11-orbstack-00143-ge6b82e26cd22   containerd://1.7.13
kind-worker          Ready      <none>          7m15s   v1.29.2   192.168.228.2   <none>        Debian GNU/Linux 12 (bookworm)   6.7.11-orbstack-00143-ge6b82e26cd22   containerd://1.7.13
kind-worker2         NotReady   <none>          7m13s   v1.29.2   192.168.228.5   <none>        Debian GNU/Linux 12 (bookworm)   6.7.11-orbstack-00143-ge6b82e26cd22   containerd://1.7.13
```

But now we can see that the NotReady node is not added to the kubelet endpoint and not being scraped.

```bash
curl -s 'localhost:9090/api/v1/targets?scrapePool=serviceMonitor%2Fdefault%2Fkps-kube-prometheus-stack-kubelet%2F0' | jq '.data.activeTargets[] | "\(.scrapeUrl): \(.health)"' -r
https://192.168.228.4:10250/metrics: up
https://192.168.228.2:10250/metrics: up
```

## Changelog entry

```release-note
Change prometheus-operator behavior to only add Ready nodes' endpoint addresses to the kubelet endpoint for scraping.
```

### DIFF
--- a/pkg/kubelet/controller.go
+++ b/pkg/kubelet/controller.go
@@ -155,6 +155,19 @@ func (c *Controller) nodeAddress(node v1.Node) (string, map[v1.NodeAddressType][
 	return "", m, fmt.Errorf("host address unknown")
 }
 
+// nodeReadyConditionKnown checks the node for a known Ready condition. If the
+// condition is Unknown then that node's kubelet has not recently sent any node
+// status, so we should not add this node to the kubelet endpoint and scrape
+// it.
+func nodeReadyConditionKnown(node v1.Node) bool {
+	for _, c := range node.Status.Conditions {
+		if c.Type == v1.NodeReady && c.Status != v1.ConditionUnknown {
+			return true
+		}
+	}
+	return false
+}
+
 func (c *Controller) getNodeAddresses(nodes *v1.NodeList) ([]v1.EndpointAddress, []error) {
 	addresses := make([]v1.EndpointAddress, 0)
 	errs := make([]error, 0)
@@ -163,6 +176,12 @@ func (c *Controller) getNodeAddresses(nodes *v1.NodeList) ([]v1.EndpointAddress,
 		address, _, err := c.nodeAddress(n)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("failed to determine hostname for node (%s): %w", n.Name, err))
+			continue
+		}
+		if !nodeReadyConditionKnown(n) {
+			if c.logger != nil {
+				level.Info(c.logger).Log("msg", "Node Ready condition is Unknown", "node", n.GetName())
+			}
 			continue
 		}
 		addresses = append(addresses, v1.EndpointAddress{
--- a/pkg/kubelet/controller.go
+++ b/pkg/kubelet/controller.go
@@ -169,9 +169,10 @@ func nodeReadyConditionKnown(node v1.Node) bool {
 }
 
 func (c *Controller) getNodeAddresses(nodes *v1.NodeList) ([]v1.EndpointAddress, []error) {
+	addresses := make([]v1.EndpointAddress, 0)
 	errs := make([]error, 0)
-	readyKnownNodes := make(map[string]v1.EndpointAddress)
-	readyUnknownNodes := make(map[string]v1.EndpointAddress)
+	readyKnownNodes := make(map[string]string)
+	readyUnknownNodes := make(map[string]string)
 
 	for _, n := range nodes.Items {
 		address, _, err := c.nodeAddress(n)
@@ -179,22 +180,7 @@ func (c *Controller) getNodeAddresses(nodes *v1.NodeList) ([]v1.EndpointAddress,
 			errs = append(errs, fmt.Errorf("failed to determine hostname for node (%s): %w", n.Name, err))
 			continue
 		}
-		if !nodeReadyConditionKnown(n) {
-			if c.logger != nil {
-				level.Info(c.logger).Log("msg", "Node Ready condition is Unknown", "node", n.GetName())
-			}
-			readyUnknownNodes[address] = v1.EndpointAddress{
-				IP: address,
-				TargetRef: &v1.ObjectReference{
-					Kind:       "Node",
-					Name:       n.Name,
-					UID:        n.UID,
-					APIVersion: n.APIVersion,
-				},
-			}
-			continue
-		}
-		readyKnownNodes[address] = v1.EndpointAddress{
+		addresses = append(addresses, v1.EndpointAddress{
 			IP: address,
 			TargetRef: &v1.ObjectReference{
 				Kind:       "Node",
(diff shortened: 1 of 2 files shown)


tstringer-fn:
This closes #6548.

simonpasquier:
This would be a change. To reduce the impact, I'd prefer to skip nodes which have unknown status *and* for which another node reports with the same IP address.

tstringer-fn:
@simonpasquier, thanks for the feedback! I've made the changes to reflect that. It's worth noting that because this new implementation uses a map, the original tests required sorting for their assertions. It doesn't appear as though the order of IP addresses is important outside of the tests, so I changed that behavior to no longer fail on order. But please let me know if that was a faulty assumption.

Thanks again!

simonpasquier (commented):
I'd rather preserve the order of the list to avoid unnecessary updates to the endpoints object.

tstringer-fn:
Thanks for the feedback! I've added logic that preserves order.

simonpasquier (approved):
Thanks!

PR #6567, opened 2024-05-06
fix: alertmanager: Add TLS-assets for email and webhook receivers to assets secret when using alertmanagerConfiguration

name212:
## Description

When we are using alertmanagerConfiguration email and webhook receivers from this config which contain TLS setting (TLS-assets), they do not add in internal store and therese assets do not save to TLS-assets secret which will mounted to alert manager pod.

## Type of change

## Verification

## Changelog entry

Add TLS-assets for email and webhook receivers to assets secret when using alertmanagerConfiguration

```release-note

```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -1055,6 +1055,9 @@ func (cb *configBuilder) convertEmailConfig(ctx context.Context, in monitoringv1
 	}
 
 	if in.TLSConfig != nil {
+		if err := cb.store.AddSafeTLSConfig(ctx, crKey.Namespace, in.TLSConfig); err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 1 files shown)


simonpasquier:
Are you using it for the global AlertmanagerConfig? Otherwise the operator should already load the TLS assets into the store.

name212:
>Are you using it for the global AlertmanagerConfig

Yes.

simonpasquier:
My recommendation would be to call `checkAlertmanagerConfigResource()` on the global AlertmanagerConfig object which we don't IIUC. E.g. before calling `initializeFromAlertmanagerConfig()`


https://github.com/prometheus-operator/prometheus-operator/blob/48e98584b57b72a4a10951414b1c0b3acf043426/pkg/alertmanager/operator.go#L861-L871

Issue #6560, opened 2024-05-02
Allow setting env vars for config-reloader created for Prometheus Kind


xbglowx:
### Component(s)

Prometheus

### What is missing? Please describe.

The ability to pass env vars to the config-reloader that is created by the operator that is for Prometheus Kind.

I originally created an issue under the helm chart, but I think a better place for it would be here instead? https://github.com/prometheus-community/helm-charts/issues/3982

### Describe alternatives you've considered.

I was able to use Kustomize to patch the config-reloader container that is part of the Operator Pod, since that is a resource that gets created directly from the helm chart, but can't do the same for the config-reloader that gets created by Prometheus kind.

## Environment
Kubernetes Version: 1.27.12
Prometheus-Operator Version: v0.71.2

simonpasquier:
it should be possible using strategic merge patch: https://prometheus-operator.dev/docs/operator/strategic-merge-patch/#how-to-inject-an-environment-variable-in-an-existing-container

simonpasquier:
But if the request is "I want the containers to pick up the correct GOMAXPROCS based on the CPU resource limits", I'd be open to integrate with https://github.com/uber-go/automaxprocs.

xbglowx:
> it should be possible using strategic merge patch: https://prometheus-operator.dev/docs/operator/strategic-merge-patch/#how-to-inject-an-environment-variable-in-an-existing-container

Thanks and sorry that I totally missed that option.

xbglowx:
> But if the request is "I want the containers to pick up the correct GOMAXPROCS based on the CPU resource limits", I'd be open to integrate with https://github.com/uber-go/automaxprocs.

Yes, this would be a nice to have, so that one doesn't have to worry about setting GOMAXPROCS in the first place.

simonpasquier:
Closd by #6576

PR #6491, opened 2024-04-09
Add structure for feature flags

ArthurSens:
## Description

As discussed in the last office hours, we want to start to provide a few features behind feature flags. This PR introduces the concept, but still no feature flags are available.

They will be added as need in following PRs.

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [X] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification

Please check the [Prometheus-Operator testing guidelines](../TESTING.md) for recommendations about automated tests.

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #6576, opened 2024-05-09
feat(env):  auto set GOMAXPROCS by go.uber.org/automaxprocs

dongjiang1989:
## Description

Auto set GOMAXPROCS. 
add dependency on [go.uber.org/automaxprocs](https://github.com/uber-go/automaxprocs)

ref: #6560 

## Type of change

## Verification

## Changelog entry

```release-note
auto set GOMAXPROCS by go.uber.org/automaxprocs
```


nicolastakashi:
Amazing @dongjiang1989 🥳

cmd/operator/main.go
```diff
@@ -183,6 +184,10 @@ func run(fs *flag.FlagSet) int {
 		stdlog.Fatal(err)
 	}
 
+	if _, err := maxprocs.Set(); err != nil {
```
nicolastakashi, line 187: We can provide logger to the maxprocs.set as well, for reference: https://github.com/prometheus/prometheus/blob/d699dc3c7706944aafa56682ede765398f925ef0/cmd/prometheus/main.go#L724

xbglowx:
Thanks for working on this @dongjiang1989

cmd/operator/main.go
dongjiang1989: Thanks. @nicolastakashi 
Fixed. Please re-check

cmd/operator/main.go
```diff
@@ -183,6 +185,13 @@ func run(fs *flag.FlagSet) int {
 		stdlog.Fatal(err)
 	}
 
+	l := func(format string, a ...interface{}) {
```
nicolastakashi, line 188: I think it should be behind a feature flag, wdyt?

cmd/prometheus-config-reloader/main.go
```diff
@@ -122,6 +123,13 @@ func main() {
 		stdlog.Fatal(err)
 	}
 
+	l := func(format string, a ...interface{}) {
```
nicolastakashi, line 126: Same thing behind a feature flag, wdyt?

cmd/operator/main.go
ArthurSens: We still don't have feature flags tho 😅 

But I agree that this needs to be configurable, e.g. what's the side-effect of having gomaxprocs auto enabled during development? If I run the operator in my machine, will it take over all cores?

dongjiang1989: Read the `CGroup` value in the container environment to identify the `CPU quota` of the container, calculate the actual number of cores, and automatically set the number of `GOMAXPROCS` threads.

cmd/prometheus-config-reloader/main.go
dongjiang1989: Thanks. Add `enable-feature` flag done.
Please re-check. @nicolastakashi

cmd/operator/main.go
nicolastakashi: Well!
I think it's a fair and simple feature flag implementation.
I'm ok with that, because it's also the same approach we have in Prometheus.
Let's see what the other maintainers think about it.

ArthurSens: I have another PR open for a few weeks implementing feature-flags as well, but following Kubernetes feature-flags approach. https://github.com/prometheus-operator/prometheus-operator/pull/6491

Differently from Prometheus' approach, K8s allows us maintainers to graduate feature flags slowly while still letting users disable feature flags that we might enable by default in the future.

In my opinion, this gives us the flexibility to enable features when we feel it has achieved enough maturity, while also providing safe-guards to users if they break something in production

simonpasquier:
I'd be fine using automaxprocs by default without a feature flag. Unlike Prometheus, the operator (almost) always runs on Kubernetes and we should provide the best experience for folks setting CPU limits.

> e.g. what's the side-effect of having gomaxprocs auto enabled during development? If I run the operator in my machine, will it take over all cores?

the same as today: the Go runtime will set GOMAXPROCS to the number of cores available from your laptop.

nicolastakashi:
> I'd be fine using automaxprocs by default without a feature flag. Unlike Prometheus, the operator (almost) always runs on Kubernetes and we should provide the best experience for folks setting CPU limits.
> 
> > e.g. what's the side-effect of having gomaxprocs auto enabled during development? If I run the operator in my machine, will it take over all cores?
> 
> the same as today: the Go runtime will set GOMAXPROCS to the number of cores available from your laptop.

Yes, it makes sense.
I was just curious about the `almost` 👀

simonpasquier:
"almost" = except when you do local development

dongjiang1989:
> I'd be fine using automaxprocs by default without a feature flag. Unlike Prometheus, the operator (almost) always runs on Kubernetes and we should provide the best experience for folks setting CPU limits.
> 
> > e.g. what's the side-effect of having gomaxprocs auto enabled during development? If I run the operator in my machine, will it take over all cores?
> 
> the same as today: the Go runtime will set GOMAXPROCS to the number of cores available from your laptop.

Thanks. Got it. Will I revert this feature flag?

simonpasquier (approved):
LGTM. I've tested on my local env with different limit values (including none) and it works fine from what I can tell.

ArthurSens (approved):
> the same as today: the Go runtime will set GOMAXPROCS to the number of cores available from your laptop.

I had a wrong understanding of this configuration, I don't know how I thought that but I had the impression it was configuring how many CPU cores it would take priority over other processes in my machine 😅

simonpasquier:
thanks!

PR #6589, opened 2024-05-14
[CHORE] adding sample limit to scrape classe

nicolastakashi:
## Description

When managing sample limits for different targets, scrape class can support default config for different group of targets with different sample limits.

## Type of change

## Verification

## Changelog entry

```release-note
scrape class sample limit 
```


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1848,4 +1848,10 @@ type ScrapeClass struct {
 	//
 	// +optional
 	Relabelings []RelabelConfig `json:"relabelings,omitempty"`
+
+	// SampleLimit defines per-scrape limit on number of scraped samples that will be accepted.
+	// Only valid in Prometheus versions 2.45.0 and newer.
+	//
+	// +optional
```
ArthurSens, line 1855: Should we explain the order in which the configuration is generated, similar to what we do with the other ScrapeClass fields?

We have quite a few options to limits now:
* Limits set in .*Monitor objects
* Limits in ScrapeClass, set in Prometheus
* Enforced Limites, set in Prometheus

The order could get very confusing for beginners 😬

ArthurSens (commented):
Sorry, haven't had the time to a very detailed review... just a quick walkthrough

simonpasquier:
I'm worried that it makes it too complicated to understand which limits are applied eventually. IIUC the use case for a scrape class limit is that an admin wants to apply a sane default value when a scrape object doesn't specify a limit and keep the object's limit when defined (even if greater than the default limit). Is this correct?

I wonder if such feature shouldn't be delegated to policy engines like [Kyverno](https://kyverno.io/docs/writing-policies/mutate/)?

pkg/apis/monitoring/v1/prometheus_types.go
nicolastakashi: @ArthurSens can you check it back please?

simonpasquier:
> Current State: Prometheus has two properties, sampleLimit and enforcedSampleLimit, which are mutually exclusive. If both are set, enforcedSampleLimit takes precedence.

Just to be clear: as of today, the enforced limit only takes precedence when it's less than the scrape object's limit.

simonpasquier:
> Scenario: An admin wants to set both soft and hard limits for the sample limits in Prometheus.

sorry I don't understand this use case.

nicolastakashi:
> > Scenario: An admin wants to set both soft and hard limits for the sample limits in Prometheus.
> 
> sorry I don't understand this use case.

Ok let me try different!
Imagine you have a service monitor with a sample limit of 2k by default and a enforcedSampleLimit at 4k.
The enforcedSampleLimit will only take precedence over the sample limit defined on the service monitor if the limit on the service monitor were greater than 4k.

The one defined on the service monitor is a soft limit since the service monitor object can increase that value up to the enforcedSampleLimit.

The scrape class in this context acts a default value for sample limit in case the service monitor owner didn't define any.

Does it looks better?

simonpasquier:
Thanks it clarifies a lot!

> The scrape class in this context acts a default value for sample limit in case the service monitor owner didn't define any.

Could it be solved if we consider that when both sampleLimit and enforcedSampleLimit are specified, we take the min if the scrape object has no limit itself?

Given
````
kind: Prometheus
spec:
  sampleLimit: 1000
  enforcedSampleLimit: 2000
````

* For a ServiceMonitor with no sampleLimit defined => 1000.
* For a ServiceMonitor with a sampleLimits of 500 => 500.
* For a ServiceMonitor with a sampleLimits of 1500 => 1500.
* For a ServiceMonitor with a sampleLimits of 2500 => 2000.

WDYT?

nicolastakashi:
> sampleLimit: 1000
>   enforcedSampleLimit: 2000

I thought the same, but the enforced sample limit is taking precedence over the sample limit defined on the prometheus limits.

I think sampleLimit from Prometheus is being configured as a global sample limit and not on the monitor object level @simonpasquier

simonpasquier:
Reading the code again, I think that we could improve the generated config and leverage the fact that the global `sample_limit` is always applied if no limit is set (instead of setting the value on all scrape configs).

Going back to my examples:

* For a ServiceMonitor A with no sampleLimit defined.
* For a ServiceMonitor B with a sampleLimits of 500.
* For a ServiceMonitor C with a sampleLimits of 1500.
* For a ServiceMonitor D with a sampleLimits of 2500.

With a global sampleLimit = 1000 and enforcedSampleLimit = 2000, we should generate:

````
global:
  sample_limit: 1000 # applies to all monitors without an explicit limit
scrape_configs:
- job_name: "A" # sample_limit: 1000 because of global
- job_name: "B"
  sample_limit: 500
- job_name: "C"
  sample_limit: 1500
- job_name: "D"
  sample_limit: 2000 # clamped down by the enforced limit
````

With a global sampleLimit = 1000 and no enforcedSampleLimit, we should generate:

````
global:
  sample_limit: 1000 # applies to all monitors without an explicit limit
scrape_configs:
- job_name: "A" # sample_limit: 1000 because of global
- job_name: "B"
  sample_limit: 500
- job_name: "C"
  sample_limit: 1500
- job_name: "D"
  sample_limit: 2500
````

With no global sampleLimit and enforcedSampleLimit = 2000, we should generate:

````
global:
  sample_limit: 2000 # applies to all monitors without an explicit limit
scrape_configs:
- job_name: "A" # sample_limit: 2000 because of global
- job_name: "B"
  sample_limit: 500
- job_name: "C"
  sample_limit: 1500
- job_name: "D"
  sample_limit: 2000
````

nicolastakashi:
> Reading the code again, I think that we could improve the generated config and leverage the fact that the global `sample_limit` is always applied if no limit is set (instead of setting the value on all scrape configs).
> 
> Going back to my examples:
> 
> * For a ServiceMonitor A with no sampleLimit defined.
> * For a ServiceMonitor B with a sampleLimits of 500.
> * For a ServiceMonitor C with a sampleLimits of 1500.
> * For a ServiceMonitor D with a sampleLimits of 2500.
> 
> With a global sampleLimit = 1000 and enforcedSampleLimit = 2000, we should generate:
> 
> ```
> global:
>   sample_limit: 1000 # applies to all monitors without an explicit limit
> scrape_configs:
> - job_name: "A" # sample_limit: 1000 because of global
> - job_name: "B"
>   sample_limit: 500
> - job_name: "C"
>   sample_limit: 1500
> - job_name: "D"
>   sample_limit: 2000 # clamped down by the enforced limit
> ```
> 
> With a global sampleLimit = 1000 and no enforcedSampleLimit, we should generate:
> 
> ```
> global:
>   sample_limit: 1000 # applies to all monitors without an explicit limit
> scrape_configs:
> - job_name: "A" # sample_limit: 1000 because of global
> - job_name: "B"
>   sample_limit: 500
> - job_name: "C"
>   sample_limit: 1500
> - job_name: "D"
>   sample_limit: 2500
> ```
> 
> With no global sampleLimit and enforcedSampleLimit = 2000, we should generate:
> 
> ```
> global:
>   sample_limit: 2000 # applies to all monitors without an explicit limit
> scrape_configs:
> - job_name: "A" # sample_limit: 2000 because of global
> - job_name: "B"
>   sample_limit: 500
> - job_name: "C"
>   sample_limit: 1500
> - job_name: "D"
>   sample_limit: 2000
> ```

@simonpasquier this works fine for my use case, and I'll open another PR doing this implementation, but this will not solve the use case where a Prometheus Admin would like to set different default sample limits for different group targets.

Do you think this PR is still valid?

ArthurSens:
Yeah, I see ScrapeClasses as a great ally for Platform teams that use Prometheus-Operator to offer Prometheus as a Service.

What I envision the most is using ScrapeClasses to automatically add security configuration, but also to offer "Scrape Tiers", where consumers of these Prometheus as a Service could choose their appropriate tiers while negotiating budgets with the Platform Team.

We have a few examples out there, e.g. [Cloudflare establishes basic limits to all scrape configurations](https://blog.cloudflare.com/how-cloudflare-runs-prometheus-at-scale#:~:text=Basic%20limits,your%20metrics%20more.) and allow teams to manually override them. The problem here is that this approach requires consumers of this API to understand Prometheus' limits and can easily become a barrier.

An much simpler abstraction would be to allow Platform teams to set limits in scrape classes and just offer tiers like:
* Basic scrape tier
* Medium scrape tier
* High scrape tier

simonpasquier:
I'm definitely not against adding limits to scrape classes but as stated in the Cloudflare article, global limits would probably work for > 90% users.
I'm also not sure how a Prometheus admin would prevent a team from referencing a more generous scrape class than what they're supposed to? But maybe it can be considered a feature :)

nicolastakashi:
@simonpasquier @ArthurSens PTAL

nicolastakashi:
@simonpasquier @ArthurSens anything I can do to help review it?

PR #6591, opened 2024-05-15
feat(env): Add automatic memory limit handling

dongjiang1989:
## Description

add auto  GOMEMLIMIT setting

## Type of change

## Verification

## Changelog entry

```release-note
add auto  GOMEMLIMIT setting
```


nicolastakashi:
LGTM 🥳 
Just a simple question, on the PR about MAXPROC I think you also added to config reloader, should we do the same?

dongjiang1989:
> LGTM 🥳 Just a simple question, on the PR about MAXPROC I think you also added to config reloader, should we do the same?

Thanks. Done added.

simonpasquier (commented):
Unlike the gomaxprocs change, I wonder if it shouldn't be an opt-in feature? My rationale is that GOMAXPROCS is always set to "some" value while GOMEMLIMIT is unset by default (e.g. `math.MaxInt64`).

Also should we really consider the system memory if no k8s limits are set? More food for thoughts from https://www.ardanlabs.com/blog/2024/02/kubernetes-memory-limits-go.html:

> If you’re not going to use K8s memory limits, then don’t do anything with GOMEMLIMIT. The Go runtime is really good at finding the sweet spot for your memory requirements.

simonpasquier:
To follow-up with my last comment, I'd suggest that we treat <= 0 values as "don't set GOMEMLIMIT" and > 1 values as 1.
We can set the default to 0, thus making it an opt-in feature for now (we can still change it later). Also I'd only take into account the cgroup memory limit in the calculation and I would highlight in the arg description that you have to set memory limits to benefit from the functionality. What do you think?

cmd/operator/main.go
```diff
@@ -164,6 +167,9 @@ func parseFlags(fs *flag.FlagSet) {
 	fs.Var(&cfg.ThanosRulerSelector, "thanos-ruler-instance-selector", "Label selector to filter ThanosRuler Custom Resources to watch.")
 	fs.Var(&cfg.SecretListWatchSelector, "secret-field-selector", "Field selector to filter Secrets to watch")
 
+	// Auto GOMEMLIMIT Ratio
+	fs.Float64Var(&memlimitRatio, "auto-gomemlimit-ratio", 0.9, "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory")
```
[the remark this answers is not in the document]
dongjiang1989: Thanks. Done.

cmd/operator/main.go
```diff
@@ -184,6 +190,24 @@ func run(fs *flag.FlagSet) int {
...
+		memlimit.WithRatio(memlimitRatio),
+		memlimit.WithProvider(
+			memlimit.ApplyFallback(
+				memlimit.FromCgroup,
+				memlimit.FromSystem,
+			),
+		),
+	); err != nil {
+		level.Warn(logger).Log("component", "automemlimit", "msg", "Failed to set GOMEMLIMIT automatically", "err", err)
+	}
```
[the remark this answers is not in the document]
dongjiang1989: Thanks. Done.

dongjiang1989:
> To follow-up with my last comment, I'd suggest that we treat <= 0 values as "don't set GOMEMLIMIT" and > 1 values as 1. We can set the default to 0, thus making it an opt-in feature for now (we can still change it later). Also I'd only take into account the cgroup memory limit in the calculation and I would highlight in the arg description that you have to set memory limits to benefit from the functionality. What do you think?

Thank. Got it. @simonpasquier ,  PTAL

cmd/operator/main.go
```diff
@@ -169,8 +173,11 @@ func parseFlags(fs *flag.FlagSet) {
 	fs.Var(&cfg.ThanosRulerSelector, "thanos-ruler-instance-selector", "Label selector to filter ThanosRuler Custom Resources to watch.")
 	fs.Var(&cfg.SecretListWatchSelector, "secret-field-selector", "Field selector to filter Secrets to watch")
 
+	// Auto GOMEMLIMIT Ratio
+	fs.Float64Var(&memlimitRatio, "auto-gomemlimit-ratio", 0.9, "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory. The value should be less than 0.0 setting to 0.0 and greater than 1.0 setting to 1.0. Default: 0.9")
+
 	featureGates = k8sflag.NewMapStringBool(ptr.To(make(map[string]bool)))
-	fs.Var(featureGates, "feature-gates", "Feature gates are a set of key=value pairs that describe Prometheus-Operator features. At the moment there are no feature gates available.")
+	fs.Var(featureGates, "feature-gates", "Feature gates are a set of key=value pairs that describe Prometheus-Operator features. Valid options: auto-gomemlimit.")
```
ArthurSens, line 180: The comment line below shows how we intend to show available feature gates.

What we want to do is add default values [here](https://github.com/prometheus-operator/prometheus-operator/blob/4c1706c4cf97c9e2db07635b3e517c1e0a949ec7/pkg/operator/feature_gates.go#L25-L26), then uncomment the line below (line 181 in this PR) and delete line 180.

cmd/operator/main.go
```diff
@@ -202,6 +209,28 @@ func run(fs *flag.FlagSet) int {
 		return 1
 	}
 
+	if ok := operator.HasFeatureGate("auto-gomemlimit"); ok {
```
ArthurSens, line 212:
```suggestion
	if featureGates.Map["auto-gomemlimit"] {
```

pkg/operator/feature_gates.go
```diff
@@ -61,3 +61,7 @@ func mapToString(m map[string]bool) string {
 	}
 	return strings.Join(s, ",")
 }
+
+func HasFeatureGate(feat string) bool {
+	return defaultFeatureGates[feat]
+}
```
ArthurSens, line 67: This isn't necessary, see comments above :)

cmd/operator/main.go
```diff
@@ -202,6 +209,28 @@ func run(fs *flag.FlagSet) int {
...
+			memlimit.WithRatio(memlimitRatio),
+			memlimit.WithProvider(
+				memlimit.ApplyFallback(
+					memlimit.FromCgroup,
+					memlimit.FromSystem,
+				),
+			),
+		); err != nil {
+			level.Warn(logger).Log("component", "automemlimit", "msg", "Failed to set GOMEMLIMIT automatically", "err", err)
+		} else {
```
ArthurSens, line 229: Are we adding automatic gomemlimit to other binaries as well, e.g. config-reloader and admission-webhook? If yes, would it make sense to implement similarly to what Simon did [here](https://github.com/prometheus-operator/prometheus-operator/pull/6599/files)?

We could avoid code duplication that way.

ArthurSens: If we're not adding automatic gomemlimit to all binaries, then ignore this comment 😅

cmd/operator/main.go
```diff
@@ -169,8 +173,11 @@ func parseFlags(fs *flag.FlagSet) {
 	fs.Var(&cfg.ThanosRulerSelector, "thanos-ruler-instance-selector", "Label selector to filter ThanosRuler Custom Resources to watch.")
 	fs.Var(&cfg.SecretListWatchSelector, "secret-field-selector", "Field selector to filter Secrets to watch")
 
+	// Auto GOMEMLIMIT Ratio
+	fs.Float64Var(&memlimitRatio, "auto-gomemlimit-ratio", 0.9, "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory. The value should be less than 0.0 setting to 0.0 and greater than 1.0 setting to 1.0. Default: 0.9")
```
ArthurSens, line 177: I'm having a really hard time understanding what is being said here 😬 

```
The value should be less than 0.0 setting to 0.0 and greater than 1.0 setting to 1.0
```

Could you clarify?

ArthurSens: If we're putting this behind a feature-gate, I think it makes sense to mention that here as well

pkg/operator/feature_gates.go
nicolastakashi: I don't have a hard opinion on that, but I do think it's more idiomatic, I'd would choose one I'll choose the HasFeatureGate 😅 

But as you guys prefer.

cmd/operator/main.go
dongjiang1989: Thank for your review.  Change to `(*featureGates.Map)["auto-gomemlimit"]`

cmd/operator/main.go
dongjiang1989: Thanks. Got it.  PTAL

cmd/operator/main.go
dongjiang1989: > I'm having a really hard time understanding what is being said here 😬
> 
> ```
> The value should be less than 0.0 setting to 0.0 and greater than 1.0 setting to 1.0
> ```
> 
> Could you clarify?

Change it to 
```
The value should be greater than 0.0 and less than 1.0. Default: 0.9
```

cmd/operator/main.go
dongjiang1989: Thank for your review. Done it.

pkg/operator/feature_gates.go
dongjiang1989: Thanks. Got it.  Please re-check

cmd/prometheus-config-reloader/main.go
```diff
@@ -138,6 +142,7 @@ func main() {
 	level.Info(logger).Log("msg", "Starting prometheus-config-reloader", "version", version.Info())
 	level.Info(logger).Log("build_context", version.BuildContext())
 	goruntime.SetMaxProcs(logger)
+	goruntime.SetMemLimit(logger, *memlimitRatio)
```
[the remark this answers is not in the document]
simonpasquier: Good remark. I was dense in my previous comment but my initial idea was that we default the `auto-gomemlimit-ratio` argument to 0, effectively disabling auto memory limit for all users. So we won't need a feature gate and could make it work for all binaries without additional change. WDYT?

simonpasquier (commented):
Thanks! Can you make the same change for the admission webhook too? I'd also advise to factorize the argument parsing in internal/goruntime to avoid duplication.

cmd/prometheus-config-reloader/main.go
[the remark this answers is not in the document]
ArthurSens: Yeah, this makes more sense to me to be honest :) 

Default to 0 (disabled), no feature gates

pkg/operator/feature_gates_test.go
```diff
@@ -0,0 +1,45 @@
...
+	require.GreaterOrEqual(t, len(m), 1)
+}
+
+func TestMapToString(t *testing.T) {
+	m := mapToString(defaultFeatureGates)
+
+	require.True(t, strings.Contains(m, "auto-gomemlimit=true"))
+}
+
+func TestValidateFeatureGates(t *testing.T) {
```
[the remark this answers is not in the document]
dongjiang1989: Just testing for `TestValidateFeatureGatesWithNotSupportFeature`

cmd/prometheus-config-reloader/main.go
```diff
@@ -69,10 +71,12 @@ func main() {
 	retryInterval := app.Flag("retry-interval", "how long the reloader waits before retrying in case the endpoint returned an error").Default(defaultRetryInterval.String()).Duration()
 	reloadTimeout := app.Flag("reload-timeout", "how long the reloader waits for a response from the reload URL").Default(defaultReloadTimeout.String()).Duration()
 
+	memlimitRatio := app.Flag("auto-gomemlimit-ratio", "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory").Default(defaultGOMemlimitRatio).Float64()
+
 	watchedDir := app.Flag("watched-dir", "directory to watch non-recursively").Strings()
 
 	reloadMethod := app.Flag("reload-method", "method used to reload the configuration").Default(httpReloadMethod).Enum(httpReloadMethod, signalReloadMethod)
-	processName := app.Flag("process-executable-name", "executable name used to match the process when using the signal reload method").Default("prometheus").String()
+	processName := app.Flag("process-executable-name", "executable name used to match the process when using the signal reload method").String()
```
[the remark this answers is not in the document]
dongjiang1989: Sorry. Miss it.  Revert it back

cmd/prometheus-config-reloader/main.go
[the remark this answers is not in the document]
dongjiang1989: Thanks. Setting Default to 0.

dongjiang1989:
Apply the same modification to the admission webhook as was done. Setting `auto-gomemlimit-ratio` to 0.

cmd/operator/main.go
```diff
@@ -169,10 +173,11 @@ func parseFlags(fs *flag.FlagSet) {
 	fs.Var(&cfg.ThanosRulerSelector, "thanos-ruler-instance-selector", "Label selector to filter ThanosRuler Custom Resources to watch.")
 	fs.Var(&cfg.SecretListWatchSelector, "secret-field-selector", "Field selector to filter Secrets to watch")
 
+	// Auto GOMEMLIMIT Ratio
+	fs.Float64Var(&memlimitRatio, "auto-gomemlimit-ratio", defaultMemlimitRatio, "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory. The value should be greater than 0.0 and less than 1.0. Default: 0.9")
```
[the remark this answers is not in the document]
dongjiang1989: Thanks. Done

pkg/operator/feature_gates.go
```diff
@@ -23,7 +23,9 @@ import (
 )
 
 // At the moment, the are no feature gates available.
-var defaultFeatureGates = map[string]bool{}
+var defaultFeatureGates = map[string]bool{
+	"auto-gomemlimit": true,
```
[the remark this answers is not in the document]
dongjiang1989: Thanks. Done.

internal/goruntime/memory.go
```diff
@@ -0,0 +1,48 @@
...
+			memlimit.WithProvider(
+				memlimit.ApplyFallback(
+					memlimit.FromCgroup,
+					memlimit.FromSystem,
+				),
+			),
+		); err != nil {
+			level.Warn(logger).Log("component", "automemlimit", "msg", "Failed to set GOMEMLIMIT automatically", "err", err)
+		}
+	}
```
simonpasquier, line 47: (suggestion) we probably don't need to log anything if memlimit==0.
```suggestion
	// the memlimitRatio argument to 0, effectively disabling auto memory limit for all users.
	if memlimitRatio == 0.0 {
		return
	}
	
	if _, err := memlimit.SetGoMemLimitWithOpts(
		memlimit.WithRatio(memlimitRatio),
		memlimit.WithProvider(
			memlimit.ApplyFallback(
				memlimit.FromCgroup,
				memlimit.FromSystem,
			),
		),
	); err != nil {
		level.Warn(logger).Log("component", "automemlimit", "msg", "Failed to set GOMEMLIMIT automatically", "err", err)
	}
	
	level.Info(logger).Log("GOMEMLIMIT set to %d", debug.SetMemoryLimit(-1))
```

simonpasquier (commented):
pkg/operator/feature_gates*.go aren't need anymore, right? Also the default isn't 0 for all binaries.

cmd/operator/main.go
```diff
@@ -98,6 +98,8 @@ func checkPrerequisites(
 const (
 	defaultReloaderCPU    = "10m"
 	defaultReloaderMemory = "50Mi"
+
+	defaultMemlimitRatio = 0.9
```
ArthurSens, line 102:
```suggestion
	defaultMemlimitRatio = 0.0
```

pkg/operator/feature_gates.go
```diff
@@ -23,7 +23,9 @@ import (
 )
 
 // At the moment, the are no feature gates available.
-var defaultFeatureGates = map[string]bool{}
+var defaultFeatureGates = map[string]bool{
+	"AutoGoMemLimit": false,
+}
```
ArthurSens, line 28:
```suggestion
var defaultFeatureGates = map[string]bool{}
```

ArthurSens (commented):
Sorry for all the bumps here @dongjiang1989! After all the discussions on the PR, I think we're aligned on not using a feature-gate at all, but just disable the feature if `-auto-gomemlimit-ratio` is not set.

This means that we'd need to revert all changes made that are related to feature-gates. Thanks for the patience 🙂

dongjiang1989:
> Sorry for all the bumps here @dongjiang1989! After all the discussions on the PR, I think we're aligned on not using a feature-gate at all, but just disable the feature if `-auto-gomemlimit-ratio` is not set.
> 
> This means that we'd need to revert all changes made that are related to feature-gates. Thanks for the patience 🙂

Thanks. Got it.

pkg/operator/feature_gates.go
dongjiang1989: Thanks. Remove it

cmd/operator/main.go
dongjiang1989: Got it. Setting defaultMemlimitRatio default to 0.0 (disabled)

simonpasquier (approved):
thanks!

Issue #5079, opened 2022-10-10
Implement validations using validation-rules feature of custom resource definitions


slashpai:
**What is missing?**

Some of the validations like where two fields cannot be set at the same time can be validated using [validation rules](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) (eg: one cannot set alerting and recording rule at the same time). They are still in beta, so we may look into implementing that once its GA

**Why do we need it?**

For covering more validations

**Environment**

* Prometheus Operator version:

    `Insert image tag or Git SHA here`
    
    

NA

**Anything else we need to know?**:

This issue was created as part of reviewing another [PR](https://github.com/prometheus-operator/prometheus-operator/pull/5072#discussion_r989007674)

slashpai:
validation rules are promoted to GA in 1.29 https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules. 

We should try to move possible validations to using this since not all users may configure admission webhook and we can reduce some complexity around it as well. We would still need admission webhook for Prometheus Rules

PR #6600, opened 2024-05-16
Proposal for DaemonSet deployment of Prometheus Agent

haanhvu:
Proposal for DaemonSet deployment of Prometheus Agent

### DIFF
--- a/Documentation/proposals/202405-agent-daemonset.md
+++ b/Documentation/proposals/202405-agent-daemonset.md
@@ -125,3 +125,8 @@ For the test, we will have unit tests covering new logic, and integration tests 
 * Prometheus Agent DaemonSet selects correctly the pods on the same node.
 
 We’ll also need a new user guide explaining how to use this new mode.
+
+
+## 8. Follow-ups
+
+After the goals of this proposal have been met, we’ll handle what’s stated in the Non-goals section.
--- a/Documentation/proposals/202405-agent-daemonset.md
+++ b/Documentation/proposals/202405-agent-daemonset.md
@@ -9,7 +9,6 @@
 
 This proposal is about designing and implementing the deployment of Prometheus Agent as DaemonSet. Currently, Prometheus Agent can only be deployed as StatefulSet, which could be considered as “cluster-wide” strategy, meaning one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster. Though this works well for many use cases, some use cases may indeed prefer a “node-specific” strategy (DaemonSet), where Prometheus Agent pods scale with the nodes and only scrape the metrics from the targets located on the same node.
 
-
 ## 1. Why
 
 When deploying Prometheus Agent in Kubernetes, two of the biggest users’ concerns are load distribution and scalability.
@@ -22,7 +21,6 @@ DaemonSet deployment is especially more suitable for Prometheus Agent than Prome
 
 This deployment mode has been implemented and proven in the Google Cloud Kubernetes Engine (GKE) with [Google Cloud Managed Service for Prometheus (GMP)'s operator](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we can learn from their cases and collaborate for shared improvements.
 
-
 ## 2. Pitfalls of the current solution
 
 The key pitfalls of managing load distribution and scalability with the current StatefulSet deployment are:
@@ -31,13 +29,11 @@ The key pitfalls of managing load distribution and scalability with the current 
 
 This is not to say that DaemonSet is superior to StatefulSet. StatefulSet also has its own advantages, such as easier storage handling. This is to say that DaemonSet can solve some of the existing problems in StatefulSet, and vice versa. So DaemonSet is a good deployment option to implement besides StatefulSet. Users can choose to use one or both of them, depending on their cases.
 
-
 ## 3. Audience
 
 Users with use cases where scraped load is very large or hard to estimate and/or scalability is hard to predict, so they need a simple way to manage load distribution and scalability. They may apply DaemonSet deployment on the whole cluster or only some nodes according to their needs.
 
 An example of audience is expressed [here](https://github.com/prometheus-operator/prometheus-operator/issues/5495#issuecomment-1519812510).
-
 
 ## 4. Goals
 
@@ -48,7 +44,6 @@ In specific, the MVP will need to:
 * Allow users to set the priority of Prometheus Agent pod compared to other pods on the same node, if desired.
 * Allow each Prometheus Agent pod to only scrape from the pods from PodMonitor that run on the same node.
 
-
 ## 5. Non-Goals
 
 This proposal only aims at Prometheus Agent, not Prometheus.
@@ -58,7 +53,6 @@ Other non-goals are the features that are not easy to implement and require more
 * Storage: We will need to spend time studying more about the WAL, different storage solutions provided by Kubernetes, and how to gracefully handle storage in different cases of crashes. For example, there’s an [issue in Prometheus](https://github.com/prometheus/prometheus/issues/8809) showing that samples may be lost if remote write didn’t flush cleanly. We’ll investigate these further after the MVP.
 
 In the MVP version, we will not allow users to directly switch from a live StatefulSet to DaemonSet deployment. Reasons are explained in the CRD subsection in the How section.
-
 
 ## 6. How
 
@@ -81,6 +75,7 @@ We will add a new `mode` field that accepts either `StatefulSet` or `DaemonSet`,
 ### 6.2. Node detecting:
 
 As pointed out in [Danny from GMP’s talk](https://www.youtube.com/watch?v=yk2aaAyxgKw), to make Prometheus Agent DaemonSet know which node it’s on, we can use [Kubernetes’ downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/). In `config-reloader` container, we can mount the node name as an environment variable like this:
+
 ```
 containers:
 - name: config-reloader
@@ -95,6 +90,7 @@ containers:
 ### 6.3. Targets filtering for pods (PodMonitor support):
 
 To filter pod targets, Danny’s talk has pointed out the best option is to use field selector:
+
 ```
 kubernetes_sd_configs:
 - role: pod
@@ -102,25 +98,23 @@ kubernetes_sd_configs:
   - role: pod
     field: spec.nodeName=$NODE_NAME
 ```
+
 We'll go with this option, because it filters targets right at discovery time, and also because Kubernetes API server watch cache indexes pods by node name (as we can see in [Kubernetes codebase](https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.0/pkg/registry/core/pod/storage/storage.go#L91)).
 
 We've also considered using relabel config that filters pods by `__meta_kubernetes_pod_node_name` label. However, we didn't choose to go with this option because it filters pods only after discovering all the pods from PodMonitor, which increases load on Kubernetes API server.
-
 
 ## 7. Action Plan
 
 For the implementation, we’ll do what we detailed in the How section. The common logics between StatefulSet and DaemonSet modes will be extracted into one place. We will have a separate `daemonset.go` for the separate logic of the DaemonSet mode.
 
 For the test, we will have unit tests covering new logic, and integration tests covering the basic user cases, which are:
-* Prometheus Agent DaemonSet cannot be deployed when a live Prometheus Agent StatefulSet is there.
+* Users cannot switch directly from StatefulSet to DaemonSet.
 * Prometheus Agent DaemonSet is created/deleted successfully.
 * Prometheus Agent DaemonSet is installed on the right nodes.
-* Prometheus Agent DaemonSet scales up/down following the scale of nodes.
 * Prometheus Agent DaemonSet selects correctly the pods from PodMonitor on the same node.
-Currently we only set up a Kind cluster of one node for integration tests. Since the test cases for DaemonSet deployment requires at least two nodes, we will need to modify the Kind cluster config for that.
+  Currently we only set up a Kind cluster of one node for integration tests. Since the test cases for DaemonSet deployment requires at least two nodes, we will need to modify the Kind cluster config for that.
 
 We’ll also need a new user guide explaining how to use this new mode.
-
 
 ## 8. Follow-ups
 
--- a/Documentation/proposals/202405-agent-daemonset.md
+++ b/Documentation/proposals/202405-agent-daemonset.md
@@ -12,33 +12,27 @@ This proposal is about designing and implementing the deployment of Prometheus A
 
 ## 1. Why
 
-When deploying Prometheus Agent in Kubernetes, three of the biggest users’ concerns are: load distribution, scalability, and security.
+When deploying Prometheus Agent in Kubernetes, two of the biggest users’ concerns are load distribution and scalability.
 
-DaemonSet deployment solves all these three concerns:
-* Load distribution: Each Prometheus Agent pod will only scrape the targets located on the same node. Even though the targets on some nodes may produce more metrics than other nodes, the load distribution would be reliable enough.
-* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically added in the nodes that meet user-defined restrictions (if any).
-* Security: Since the scraped targets are local to the Prometheus Agent pod (on the same node), the scope of security problems is reduced to each node.
+DaemonSet deployment is a good solution for these:
+* Load distribution: Each Prometheus Agent pod will only scrape the targets located on the same node. Even though the targets on some nodes may produce more metrics than other nodes, the load distribution is reliable enough. This has been proven in [Google Cloud Managed Service for Prometheus (GMP)'s operator](https://github.com/GoogleCloudPlatform/prometheus-engine/) which follows a similar approach.
+* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically added in those nodes. Users can also select which set of nodes they want to deploy Prometheus Agent and the priority of Prometheus Agent pods compared to other pods on the same node.
 
-DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since the storage requirement for the Agent mode is pretty light compared to a fully functional Prometheus server.
+DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since the Agent mode requires (currently around 20 to 30%) less memory, does not produce TSDB blocks on disks, and naturally blocks querying APIs.
 
-This deployment mode has been implemented in [Google Cloud Managed Service for Prometheus (GMP)](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we have an implementation example to learn from. Also, since this deployment has been tested and proven with GMP’s userbase, we can count on that it can solve a large enough number of use cases.
+This deployment mode has been implemented and proven in the Google Cloud Kubernetes Engine (GKE) with [Google Cloud Managed Service for Prometheus (GMP)'s operator](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we can learn on their cases and collaborate on shared improvements together.
 
 
 ## 2. Pitfalls of the current solution
 
-The current (StatefulSet) deployment brings along the corresponding pitfalls:
-* Load management & scalability: Since one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster, users would need to calculate/estimate the load and scalability of the whole cluster to decide on replicas and sharding strategies. Estimating cluster-wide load and scalability is a much harder task than estimating node-wide load and scalability. Even though they can use helping tools like Horizontal Pod Autoscaler (HPA), that's still additional complexity.
-* Security: Similarly, cluster-wide security is a much bigger problem than node-wide security.
+The key pitfall of managing load distribution and scalability with the current StatefulSet deployment is they are done on the cluster scope. In other words, since one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster, users need to calculate/estimate the load and scalability of the whole cluster to decide on replicas and sharding strategies. Estimating cluster-wide load and scalability is a much harder task than estimating node-wide load and scalability. Even though users can use helping tools like Horizontal Pod Autoscaler (HPA), that's still additional complexity.
 
-At the moment, the proposed DaemonSet deployment mode doesn’t mean to be a replacement for the current StatefulSet mode. It’s actually a solution for use cases where StatefulSet may not be the best choice. We’ll keep the StatefulSet mode as long as there’s user need for it.
+We're not saying that DaemonSet is superior to StatefulSet. In fact, StatefulSet has its own advantages, such as easier storage handling. What we're trying to say is DaemonSet can solve some of the existing problems in StatefulSet, and vice versa. So DaemonSet is a good deployment option to implement besides StatefulSet.
 
 
 ## 3. Audience
 
-Users with use cases where:
-* Scraped load is very large or hard to estimate.
-* Scalability is hard to predict.
-* Security is a big concern.
+Users with use cases where scraped load is very large or hard to estimate and/or scalability is hard to predict, so they need an easy way to manage the load distribution and scalability.
 
 
 ## 4. Goals
@@ -47,7 +41,8 @@ Provide an MVP version of the DaemonSet deployment of Prometheus Agent to the Au
 In specific, the MVP will need to:
 * Allow users to deploy one Prometheus Agent pod per node.
 * Allow users to restrict which set of nodes they want to deploy Prometheus Agent, if desired.
-* Allow each Prometheus Agent pod to scrape from the pods on the same node (PodMonitor support).
+* Allow users to set the priority of Prometheus Agent pod compared to other pods on the same node, if desired.
+* Allow each Prometheus Agent pod to only scrape from the pods from PodMonitor that run on the same node as the Agent.
 
 
 ## 5. Non-Goals
@@ -121,7 +116,7 @@ For the test, we will have unit tests covering new logic, and integration tests 
 * Prometheus Agent DaemonSet is created/deleted successfully.
 * Prometheus Agent DaemonSet is installed on the right nodes.
 * Prometheus Agent DaemonSet scales up/down following the scale of nodes.
-* Prometheus Agent DaemonSet selects correctly the pods on the same node.
+* Prometheus Agent DaemonSet selects correctly the pods from PodMonitor on the same node.
 Currently we only set up a Kind cluster of one node for integration tests. Since the test cases for DaemonSet deployment requires at least two nodes, we will need to modify the Kind cluster config for that.
 
 We’ll also need a new user guide explaining how to use this new mode.
--- a/Documentation/proposals/202405-agent-daemonset.md
+++ b/Documentation/proposals/202405-agent-daemonset.md
@@ -21,13 +21,13 @@ DaemonSet deployment solves all these three concerns:
 
 DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since the storage requirement for the Agent mode is pretty light compared to a fully functional Prometheus server.
 
-This deployment mode has been implemented in [Google Cloud Managed Service for Prometheus (GMP)](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we have an implementation example to learn from. Also, since this deployment has been tested and proved with GMP’s userbase, we can count on that it can solve a large enough number of use cases.
+This deployment mode has been implemented in [Google Cloud Managed Service for Prometheus (GMP)](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we have an implementation example to learn from. Also, since this deployment has been tested and proven with GMP’s userbase, we can count on that it can solve a large enough number of use cases.
 
 
 ## 2. Pitfalls of the current solution
 
-The current (StatefulSet) deployment brings long the corresponding pitfalls:
-* Load management & scalability: Since one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster, users would need to calculate/estimate the load and scalability of the whole cluster to decide on replicas and sharding strategies. Estimating cluster-wide load and scalability is a much harder task than estimating node-wide load and scalability.
+The current (StatefulSet) deployment brings along the corresponding pitfalls:
+* Load management & scalability: Since one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster, users would need to calculate/estimate the load and scalability of the whole cluster to decide on replicas and sharding strategies. Estimating cluster-wide load and scalability is a much harder task than estimating node-wide load and scalability. Even though they can use helping tools like Horizontal Pod Autoscaler (HPA), that's still additional complexity.
 * Security: Similarly, cluster-wide security is a much bigger problem than node-wide security.
 
 At the moment, the proposed DaemonSet deployment mode doesn’t mean to be a replacement for the current StatefulSet mode. It’s actually a solution for use cases where StatefulSet may not be the best choice. We’ll keep the StatefulSet mode as long as there’s user need for it.
@@ -39,16 +39,15 @@ Users with use cases where:
 * Scraped load is very large or hard to estimate.
 * Scalability is hard to predict.
 * Security is a big concern.
-* They want to collect node system metrics (e.g. kubelet, node exporter).
 
 
 ## 4. Goals
 
-Provide an MVP version of the DaemonSet deployment of Prometheus Agent to the audience.
+Provide an MVP version of the DaemonSet deployment of Prometheus Agent to the Audience.
 In specific, the MVP will need to:
-* Allow users to deploy one Prometheus Agent pod per node
-* Allow users to restrict on which set of nodes they want to deploy Prometheus Agent
-* Allow each Prometheus Agent pod to scrape from the pods on the same node (PodMonitor support)
+* Allow users to deploy one Prometheus Agent pod per node.
+* Allow users to restrict which set of nodes they want to deploy Prometheus Agent, if desired.
+* Allow each Prometheus Agent pod to scrape from the pods on the same node (PodMonitor support).
 
 
 ## 5. Non-Goals
@@ -72,21 +71,21 @@ The current [Prometheus Agent CRD](https://prometheus-operator.dev/docs/operator
 * Replica
 * Shard
 * Storage
-We can add a new bool field `daemonset`, which is defaulted to `false`. If the DaemonSet deployment is activated (`daemonset: true`), all the fields except the unrelated ones listed above will be used for the DaemonSet mode.
+
+We can add a new `mode` field that accepts either `StatefulSet` or `DaemonSet`, with `StatefulSet` being the default. If the DaemonSet mode is activated (`mode: DaemonSet`), all the unrelated fields listed above will not be accepted. In the MVP, we will simply fail the reconciliation if any of those fields are set. Following up, we will leverage validation rules with [Kubernetes' Common Expression Language (CEL)](https://kubernetes.io/docs/reference/using-api/cel/).
 
 ### 6.2. Node detecting:
 
-As pointed out in [Danny from GMP’s talk](https://www.youtube.com/watch?v=yk2aaAyxgKw), to make Prometheus Agent DaemonSet know which node it’s on, we can use [Kubernetes’ downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/). In `prometheus` or `config-reloader` container, we can mount the node name as an environment variable like this:
-containers:
+As pointed out in [Danny from GMP’s talk](https://www.youtube.com/watch?v=yk2aaAyxgKw), to make Prometheus Agent DaemonSet know which node it’s on, we can use [Kubernetes’ downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/). In `config-reloader` container, we can mount the node name as an environment variable like this:
 ```
- - name: prometheus
-   image: prom/prometheus
-   env:
-   - name: NODE_NAME
-     valueFrom:
-     	fieldRef:
-     		apiVersion: v1
-     		fieldPath: spec.nodeName
+containers:
+- name: config-reloader
+  env:
+  - name: NODE_NAME
+    valueFrom:
+      fieldRef:
+        apiVersion: v1
+        fieldPath: spec.nodeName
 ```
 
 ### 6.3. Targets filtering for pods (PodMonitor support):
@@ -95,21 +94,21 @@ To filter targets, Danny’s talk has pointed out two options.
 The first option is to use relabel config. For example, to filter pods we could generate:
 ```
 relabel_configs:
- - source_labels: [__meta_kubernetes_pod_node_name]
-   separator: ;
-   regex: $NODE_NAME
-   replacement: $1
-   action: keep
+- source_labels: [__meta_kubernetes_pod_node_name]
+  separator: ;
+  regex: $NODE_NAME
+  replacement: $1
+  action: keep
 ```
 This could be considered as “last-mile filtering”. In other words, we watch for all the changes, and then filter the changes that match.
 
 The second option is to use field selector. For example, to select pods on a specific node, we can do:
 ```
 kubernetes_sd_configs:
-   - role: pod
- 	selectors:
- 	- role: pod
-   	  field: spec.nodeName=$NODE_NAME
+- role: pod
+  selectors:
+  - role: pod
+    field: spec.nodeName=$NODE_NAME
 ```
 This option performs better, because we filter targets right at discovery time, and also because Kubernetes API server watch cache indexes pods by node name (as we can see in [Kubernetes codebase](https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.0/pkg/registry/core/pod/storage/storage.go#L91)). So we’ll go with this option.
 
@@ -123,10 +122,12 @@ For the test, we will have unit tests covering new logic, and integration tests 
 * Prometheus Agent DaemonSet is installed on the right nodes.
 * Prometheus Agent DaemonSet scales up/down following the scale of nodes.
 * Prometheus Agent DaemonSet selects correctly the pods on the same node.
+Currently we only set up a Kind cluster of one node for integration tests. Since the test cases for DaemonSet deployment requires at least two nodes, we will need to modify the Kind cluster config for that.
 
 We’ll also need a new user guide explaining how to use this new mode.
 
 
 ## 8. Follow-ups
 
-After the goals of this proposal have been met, we’ll handle what’s stated in the Non-goals section.
+After the Goals of this proposal have been met, we’ll handle what’s stated in the Non-goals section.
+We will also work on enhancements, such as leveraging validation rules with [Kubernetes' Common Expression Language (CEL)](https://kubernetes.io/docs/reference/using-api/cel/) for the fields in Prometheus Agent CRD for DaemonSet mode.
--- a/Documentation/proposals/202405-agent-daemonset.md
+++ b/Documentation/proposals/202405-agent-daemonset.md
@@ -7,7 +7,7 @@
 * Other docs:
   * n/a
 
-This proposal is about designing and implementing the deployment of Prometheus Agent as DaemonSet. Currently, Prometheus Agent can only be deployed as StatefulSet, which could be considered as “cluster-wide” strategy, meaning one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster. Though this works well for many use cases, some use cases may indeed prefer a “node-specific” strategy (DaemonSet), where each Prometheus Agent only scrapes the metrics from the targets located on the same node.
+This proposal is about designing and implementing the deployment of Prometheus Agent as DaemonSet. Currently, Prometheus Agent can only be deployed as StatefulSet, which could be considered as “cluster-wide” strategy, meaning one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster. Though this works well for many use cases, some use cases may indeed prefer a “node-specific” strategy (DaemonSet), where Prometheus Agent pods scale with the nodes and only scrape the metrics from the targets located on the same node.
 
 
 ## 1. Why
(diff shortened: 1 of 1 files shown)


Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+   	  field: spec.nodeName=$NODE_NAME
+```
+This option performs better, because we filter targets right at discovery time, and also because Kubernetes API server watch cache indexes pods by node name (as we can see in [Kubernetes codebase](https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.0/pkg/registry/core/pod/storage/storage.go#L91)). So we’ll go with this option.
+
+
+## 7. Action Plan
+
+For the implementation, we’ll do what we detailed in the How section. The common logics between StatefulSet and DaemonSet modes will be extracted into one place. We will have a separate `daemonset.go` for the separate logic of the DaemonSet mode.
+
+For the test, we will have unit tests covering new logic, and integration tests covering the basic user cases, which are:
```
simonpasquier, line 121: Right now the e2e tests are based on the [default kind config](https://github.com/prometheus-operator/prometheus-operator/blob/main/test/e2e/kind-conf.yaml) which deploys only one node. It isn't quite enough to test the daemonset scenarios.
I suggest to modify the configuration and deploy a 2 nodes cluster instead (provided that it doesn't break the existing e2e tests).

https://kind.sigs.k8s.io/docs/user/configuration/#nodes

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+
+The second option is to use field selector. For example, to select pods on a specific node, we can do:
+```
+kubernetes_sd_configs:
+   - role: pod
+ 	selectors:
+ 	- role: pod
+   	  field: spec.nodeName=$NODE_NAME
+```
+This option performs better, because we filter targets right at discovery time, and also because Kubernetes API server watch cache indexes pods by node name (as we can see in [Kubernetes codebase](https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.0/pkg/registry/core/pod/storage/storage.go#L91)). So we’ll go with this option.
```
simonpasquier, line 114: :+1:

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+Currently, we already have a Prometheus Agent CRD that supports StatefulSet deployment. We’ll add new field(s) in this CRD to enable DaemonSet deployment.
+
+The reason for enhancing existing CRD (instead of introducing a new CRD) is it would take less time to finish the MVP. We’ll let users experiment with the MVP, and in case users report a separate CRD is needed, we’ll separate the logic of DaemonSet deployment into a new CRD later.
+
+The current [Prometheus Agent CRD](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1alpha1.PrometheusAgent) already has sufficient fields for the DaemonSet deployment. The DaemonSet deployment can use all the existing fields in the CRD except the ones related to:
+* Selectors for service, probe, ScrapeConfig
+* Replica
+* Shard
+* Storage
+We can add a new bool field `daemonset`, which is defaulted to `false`. If the DaemonSet deployment is activated (`daemonset: true`), all the fields except the unrelated ones listed above will be used for the DaemonSet mode.
```
simonpasquier, line 75: I'd suggest to use an enum field instead of a boolean, just in case other deployment modes appear.

````
spec:
  mode: DaemonSet|StatefulSet # assumes StatefulSet if unspecified.
  daemonSet: # if needed later, we can isolate daemonset-specific fields (like updateStrategy).
     ...
````

File: Documentation/proposals/202405-agent-daemonset.md [hunk 4]
```diff
@@ -0,0 +1,132 @@
...
+
+Currently, we already have a Prometheus Agent CRD that supports StatefulSet deployment. We’ll add new field(s) in this CRD to enable DaemonSet deployment.
+
+The reason for enhancing existing CRD (instead of introducing a new CRD) is it would take less time to finish the MVP. We’ll let users experiment with the MVP, and in case users report a separate CRD is needed, we’ll separate the logic of DaemonSet deployment into a new CRD later.
+
+The current [Prometheus Agent CRD](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1alpha1.PrometheusAgent) already has sufficient fields for the DaemonSet deployment. The DaemonSet deployment can use all the existing fields in the CRD except the ones related to:
+* Selectors for service, probe, ScrapeConfig
+* Replica
+* Shard
+* Storage
```
simonpasquier, line 74: @ArthurSens mentioned that we may be able to leverage [CEL](https://kubernetes.io/docs/reference/using-api/cel/) to prevent users setting unsupported fields in daemonset mode. But we also agreed that it's not required for the MVP. As a first approach, the operator can fail the reconciliation in such cases though.

simonpasquier (commented):
This is looking great!

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+
+DaemonSet deployment solves all these three concerns:
+* Load distribution: Each Prometheus Agent pod will only scrape the targets located on the same node. Even though the targets on some nodes may produce more metrics than other nodes, the load distribution would be reliable enough.
+* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically added in the nodes that meet user-defined restrictions (if any).
+* Security: Since the scraped targets are local to the Prometheus Agent pod (on the same node), the scope of security problems is reduced to each node.
+
+DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since the storage requirement for the Agent mode is pretty light compared to a fully functional Prometheus server.
+
+This deployment mode has been implemented in [Google Cloud Managed Service for Prometheus (GMP)](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we have an implementation example to learn from. Also, since this deployment has been tested and proved with GMP’s userbase, we can count on that it can solve a large enough number of use cases.
```
ArthurSens, line 24:
```suggestion
This deployment mode has been implemented in [Google Cloud Managed Service for Prometheus (GMP)](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we have an implementation example to learn from. Also, since this deployment has been tested and proven with GMP’s userbase, we can count on that it can solve a large enough number of use cases.
```

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## 2. Pitfalls of the current solution
+
+The current (StatefulSet) deployment brings long the corresponding pitfalls:
```
ArthurSens, line 29:
```suggestion
The current (StatefulSet) deployment brings along the corresponding pitfalls:
```

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+* Load management & scalability: Since one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster, users would need to calculate/estimate the load and scalability of the whole cluster to decide on replicas and sharding strategies. Estimating cluster-wide load and scalability is a much harder task than estimating node-wide load and scalability.
```
ArthurSens, line 30: We could mention that HPA is an option for up/downscale of Prometheus agents in statefulset mode, but that's extra complexity that could be avoided.

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+* Security: Similarly, cluster-wide security is a much bigger problem than node-wide security.
```
ArthurSens, line 31: Should we provide examples of how cluster-wide attacks can be avoided with daemonsets?

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+At the moment, the proposed DaemonSet deployment mode doesn’t mean to be a replacement for the current StatefulSet mode. It’s actually a solution for use cases where StatefulSet may not be the best choice. We’ll keep the StatefulSet mode as long as there’s user need for it.
+
+
+## 3. Audience
+
+Users with use cases where:
+* Scraped load is very large or hard to estimate.
+* Scalability is hard to predict.
+* Security is a big concern.
+* They want to collect node system metrics (e.g. kubelet, node exporter).
```
ArthurSens, line 42: Here it sounds like we can't collect kubelet/node exporter metrics with statefulset, which isn't true 🤔

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## 4. Goals
+
+Provide an MVP version of the DaemonSet deployment of Prometheus Agent to the audience.
+In specific, the MVP will need to:
+* Allow users to deploy one Prometheus Agent pod per node
+* Allow users to restrict on which set of nodes they want to deploy Prometheus Agent
+* Allow each Prometheus Agent pod to scrape from the pods on the same node (PodMonitor support)
```
ArthurSens, line 51:
```suggestion
* Allow users to deploy one Prometheus Agent pod per node.
* Allow users to restrict which set of nodes they want to deploy Prometheus Agent, if desired.
* Allow each Prometheus Agent pod to scrape from the pods on the same node (PodMonitor support).
```

Documentation/proposals/202405-agent-daemonset.md
ArthurSens: It would be nice to mention this in an `Alternatives` section, where we acknowledge different approaches but explain why we decided to move forward with the proposed approach.

Documentation/proposals/202405-agent-daemonset.md
ArthurSens: That sounds perfect! ☝️

Documentation/proposals/202405-agent-daemonset.md
ArthurSens: Could we mention only the option that we choose to move forward here? The other approach could be mentioned in an `Alternatives` section with an explanation of why we decided to ignore it.

Documentation/proposals/202405-agent-daemonset.md
ArthurSens: ☝️ 

Let's mention that in the proposal :)

ArthurSens (commented):
Awesome start!

I have a few small comments but the proposal already looks pretty good!

I think we could already start refactoring our codebase to extract common configuration that will be used by both statefulset and daemonset modes :)

Documentation/proposals/202405-agent-daemonset.md
haanhvu: Actually I'm thinking of removing Security from Why and Pitfalls of current solution.

First reason is AFAIU StatefulSet and DaemonSet face different kinds of security issues. We can hardly say one is better than the other. In DaemonSet, Prometheus Agent pod knows all secrets and share them with all other pods (GMP already has a proposal to solve this, but it's not implemented yet). In StatefulSet, the scope of security is larger (at cluster scope) and we have to deal with network issues.

Second reason is security is not the key reason we choose to implement DaemonSet.

You agree to remove it from our consideration?

@simonpasquier @kakkoyun

Documentation/proposals/202405-agent-daemonset.md
haanhvu: This doesn't seem like an alternative, but an enhancement. AFAIU alternatives mean what we considered, but decided not do do. Since we may do it after the MVP, should we place it in `Follow-ups` instead?

Documentation/proposals/202405-agent-daemonset.md
haanhvu: If we have alternatives for the general approach (covering the whole How section), I'll put them in a separate `Alternatives` section.

But here, it's just an alternative for a part of How. Do you think we need a whole new section for it? I think keeping them at one place/section makes it easier for readers to compare, while keeping them at two sections make readers scroll back and forth to see which part of How this Alternative refers to => harder to compare 🤔

Documentation/proposals/202405-agent-daemonset.md
haanhvu: Yeah we can do it with `kubernetesSDConfigs` in `ScrapeConfig`? If there's no advantage of solving this use case with DaemonSet, I'll remove it then.

ArthurSens: We can also create use ServiceMonitor/PodMonitor, just need to adjust labels between services/pods :) 

Yeah, I'd remove this part

haanhvu: > Yeah we can do it with `kubernetesSDConfigs` in `ScrapeConfig`?

I meant we can do this in the StatefulSet mode. This is the go-to solution in StatefulSet right?

Documentation/proposals/202405-agent-daemonset.md
simonpasquier: I'd be ok briefly mentioning that using relabel configs would be less efficient.

haanhvu: > I'd be ok briefly mentioning that using relabel configs would be less efficient.

You mean we should have a separate Alternatives section or compare them together in one section?

haanhvu:
@kakkoyun @simonpasquier @ArthurSens I resolved the reviews, also left comments for the reviews not resolved yet. Pls take another review^^

@bwplotka @pintohutch If you have time, do you mind taking a look at this too? ^^

bwplotka:
Nice, will check tomorrow latest 🤞🤞🤞🤞

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+
+DaemonSet deployment solves all these three concerns:
+* Load distribution: Each Prometheus Agent pod will only scrape the targets located on the same node. Even though the targets on some nodes may produce more metrics than other nodes, the load distribution would be reliable enough.
+* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically added in the nodes that meet user-defined restrictions (if any).
+* Security: Since the scraped targets are local to the Prometheus Agent pod (on the same node), the scope of security problems is reduced to each node.
+
+DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since the storage requirement for the Agent mode is pretty light compared to a fully functional Prometheus server.
+
+This deployment mode has been implemented in [Google Cloud Managed Service for Prometheus (GMP)](https://github.com/GoogleCloudPlatform/prometheus-engine/), so we have an implementation example to learn from. Also, since this deployment has been tested and proven with GMP’s userbase, we can count on that it can solve a large enough number of use cases.
```
ArthurSens, line 24: I have a feeling that we're going with a tone that may sound incorrect. It's true that Daemonset solves problems that are present with the Statefulset approach but at the cost of introducing new problems. It's not about Daemonset being better than Statefulset, but what problems we're trying to solve and what trade-offs users are willing to bear.

--- 

> Even though the targets on some nodes may produce more metrics than other nodes, the load distribution would be reliable enough.

* If the user's goal is to monitor only K8s master nodes, this is incorrect. Agents running on master nodes would be hot spots, while all others would remain idle. 
* If the user has metric producer applications that they can't afford to lose insight during cluster scale-ups/down, daemonsets could be a dangerous choice since storage is an unsolved problem.

> Since the scraped targets are local to the Prometheus Agent pod (on the same node), the scope of security problems is reduced to each node.

But it also introduces a new security flaw. Sensitive information is provided through PodMonitors to allow Prometheus to scrape pods in all nodes and those secrets are mounted to all Prometheus Agents, even though only a few might need them.

---

What do you think about this? Is daemonset set really a better option when compared to statefulset? Or is it just a game of trade-offs?

Maybe we could acknowledge that although Statefulset has been a good option for specific scenarios, there are other scenarios where daemonset shines, e.g. when the user cares more about simple scalability and not so much about durability.

Documentation/proposals/202405-agent-daemonset.md
ArthurSens: Given the comment above, I'm in favor of removing or at least re-writing the security concerns. I don't think we should be saying that one is better than the other, but acknowledging how they are different.

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+* Prometheus Agent DaemonSet scales up/down following the scale of nodes.
+* Prometheus Agent DaemonSet selects correctly the pods on the same node.
+Currently we only set up a Kind cluster of one node for integration tests. Since the test cases for DaemonSet deployment requires at least two nodes, we will need to modify the Kind cluster config for that.
+
+We’ll also need a new user guide explaining how to use this new mode.
+
+
+## 8. Follow-ups
+
+After the Goals of this proposal have been met, we’ll handle what’s stated in the Non-goals section.
```
ArthurSens, line 132: I'm not sure if we'll be able to solve them, to be honest. We might be over-confident when saying that we'll solve all non-goals 😅 

What do you think about saying that different design docs will be written if we find ways to solve the non-goals?

ArthurSens (commented):
Really good work 🥳 

There are a few things that we could clarify, but, as mentioned before, looks good enough to start working on the codebase already :)

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+  * [#5495](https://github.com/prometheus-operator/prometheus-operator/issues/5495)
+* Other docs:
+  * n/a
+
+This proposal is about designing and implementing the deployment of Prometheus Agent as DaemonSet. Currently, Prometheus Agent can only be deployed as StatefulSet, which could be considered as “cluster-wide” strategy, meaning one or several high-availability Prometheus Agents are responsible for scraping metrics of the whole cluster. Though this works well for many use cases, some use cases may indeed prefer a “node-specific” strategy (DaemonSet), where each Prometheus Agent only scrapes the metrics from the targets located on the same node.
+
+
+## 1. Why
+
+When deploying Prometheus Agent in Kubernetes, three of the biggest users’ concerns are: load distribution, scalability, and security.
```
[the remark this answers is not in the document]
haanhvu: I didn't cite it from a source. I read users' blogs, browsed the issues in our repo, and from my experience of setting up some nodes for benchmark pipelines for Jaeger in last year GSoC to form this general observation.

There're of course other concerns though, like cost^^ But I don't think DaemonSet helps anything with reducing cost, so didn't state it here.

Documentation/proposals/202405-agent-daemonset.md
haanhvu: Yeah, I'm not saying that we'll "solve" the Non-goals. I say we'll "handle what’s stated" in Non-goals. In Non-goals I stated "The non-goals are the features that are not easy to implement and require more investigation. We will need to investigate whether there are actual user needs for them, if yes, then how to best implement them". No promises here that we'll solve (all) them.

I know these non-goals are trickier and requires more investigation. But do you think it's too pessimistic to think we'll never solve them? ^^ We may not solve them in this year GSoC but we could do it later, e.g. in next year GSoC right?

> What do you think about saying that different design docs will be written if we find ways to solve the non-goals?
I thought this's obvious. They're non-goals, so clearly this proposal is not for them. We even stated them in Follow-ups. But I'll make it more clear for the readers then.

Documentation/proposals/202405-agent-daemonset.md
haanhvu: Yeah I never thought that one is better than the other in general. It's just better in some cases and some aspects. I'll try to make it more clear.

Documentation/proposals/202405-agent-daemonset.md
haanhvu: @simonpasquier @kakkoyun @ArthurSens Could you have a look at the current PrometheusAgent CRD to see if there're any other fields not suitable for DaemonSet, besides the ones I listed here?

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+* Allow each Prometheus Agent pod to scrape from the pods on the same node (PodMonitor support).
+
+
+## 5. Non-Goals
+
+The non-goals are the features that are not easy to implement and require more investigation. We will need to investigate whether there are actual user needs for them, if yes, then how to best implement them. We’ll handle these after the MVP.
+* ServiceMonitor support: There's a performance issue regarding this feature. Since each Prometheus Agent running on a node requires one watch, making all Prometheus Agent pods watch all endpoints will put a huge stress on Kubernetes API server. This is the main reason why GMP hasn’t supported this, even though there are user needs stated in some issues ([#362](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/362), [#192](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/192)). However, as discussed with Danny from GMP [here](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/192#issuecomment-2028850846), ServiceMonitor support based on EndpointSlice seems like a viable approach. We’ll investigate this further after the MVP.
```
kakkoyun, line 56: This is a non-goal but it'd be nice to attack this if we ever finish our planned goals before the program ends.

kakkoyun (dismissed):
This LGTM. I'll have another look at it but this shouldn't be a blocker.

One additional thing I'd like to mention is Grafana Agent. We should check how did they approach it (they already support several deployment approaches). What consideration did they make? We can even try to reach out to people and ask about the tread-offs.

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+## 1. Why
+
+When deploying Prometheus Agent in Kubernetes, three of the biggest users’ concerns are: load distribution, scalability, and security.
+
+DaemonSet deployment solves all these three concerns:
+* Load distribution: Each Prometheus Agent pod will only scrape the targets located on the same node. Even though the targets on some nodes may produce more metrics than other nodes, the load distribution would be reliable enough.
+* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically added in the nodes that meet user-defined restrictions (if any).
+* Security: Since the scraped targets are local to the Prometheus Agent pod (on the same node), the scope of security problems is reduced to each node.
+
+DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since the storage requirement for the Agent mode is pretty light compared to a fully functional Prometheus server.
```
bwplotka, line 22: To be specific: It's currently "only" around 10-20% less memory hungry and does not produce TSDB blocks on disks. Similar CPU wise. I would say that's the main reason for using agent.

More important why agent is better is because trying to query across the nodes, even with Thanos querier on top, is not ideal, because nodes can go up and down and you loose all the data. Agent naturally blocks querying APIs making it easier to scale and prohibits difficult to serve traffic.

```suggestion
DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since agent naturally blocks the querying APIs and forces users to remote write to remote backends. Additionally agent uses (currently) up to 10-20% less memory compared to Prometheus server that produces blocks etc.
```

Documentation/proposals/202405-agent-daemonset.md
bwplotka, line 24: Proposed a bit better wording (: 

```suggestion
This deployment mode has been implemented and proven in the Google Cloud Kubernetes Engine (GKE) with [Google Cloud Managed Service for Prometheus (GMP)](https://github.com/GoogleCloudPlatform/prometheus-engine/) operator, so we can learn on their cases and collaborate on shared improvements together.
```

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+* Security is a big concern.
+
+
+## 4. Goals
+
+Provide an MVP version of the DaemonSet deployment of Prometheus Agent to the Audience.
+In specific, the MVP will need to:
+* Allow users to deploy one Prometheus Agent pod per node.
+* Allow users to restrict which set of nodes they want to deploy Prometheus Agent, if desired.
+* Allow each Prometheus Agent pod to scrape from the pods on the same node (PodMonitor support).
```
bwplotka, line 50:
```suggestion
* Prometheus Agent pod will only scrape the pods from PodMonitor, which are running on the same node as the agent.
```

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+* ServiceMonitor support: There's a performance issue regarding this feature. Since each Prometheus Agent running on a node requires one watch, making all Prometheus Agent pods watch all endpoints will put a huge stress on Kubernetes API server. This is the main reason why GMP hasn’t supported this, even though there are user needs stated in some issues ([#362](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/362), [#192](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/192)). However, as discussed with Danny from GMP [here](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/192#issuecomment-2028850846), ServiceMonitor support based on EndpointSlice seems like a viable approach. We’ll investigate this further after the MVP.
+* Storage: We will need to spend time studying more about the WAL, different storage solutions provided by Kubernetes, and how to gracefully handle storage in different cases of crashes. For example, there’s an [issue in Prometheus](https://github.com/prometheus/prometheus/issues/8809) showing that samples may be lost if remote write didn’t flush cleanly. We’ll investigate these further after the MVP.
```
bwplotka, line 57: What about mixed deployment cases? Daemon set vs others? Would it be part of goal or non goal?

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+
+The reason for enhancing existing CRD (instead of introducing a new CRD) is it would take less time to finish the MVP. We’ll let users experiment with the MVP, and in case users report a separate CRD is needed, we’ll separate the logic of DaemonSet deployment into a new CRD later.
+
+The current [Prometheus Agent CRD](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1alpha1.PrometheusAgent) already has sufficient fields for the DaemonSet deployment. The DaemonSet deployment can use all the existing fields in the CRD except the ones related to:
+* Selectors for service, probe, ScrapeConfig
+* Replica
+* Shard
+* Storage
+
+We can add a new `mode` field that accepts either `StatefulSet` or `DaemonSet`, with `StatefulSet` being the default. If the DaemonSet mode is activated (`mode: DaemonSet`), all the unrelated fields listed above will not be accepted. In the MVP, we will simply fail the reconciliation if any of those fields are set. Following up, we will leverage validation rules with [Kubernetes' Common Expression Language (CEL)](https://kubernetes.io/docs/reference/using-api/cel/).
```
bwplotka, line 75: Would "mode = daemonSet" also set Prometheus mode to agent?

bwplotka (approved):
Looks good generally! Suggested some wording changes, but the plan sounds good! Great work and amazing to see this work moving forward 💪🏽 

I think the main challenge here is to make sure to not confuse users with too many/complex configuration. Making it as easy to use as possible, discover configuration pieces and debugging when something is misconfigured. Not sure what can replace a fresh CRDs honestly, but maybe more focused docs/guides on this mode would be good enough! 🤗 

(Disclaimer: I work for Google Cloud Managed Prometheus Team)

Documentation/proposals/202405-agent-daemonset.md
haanhvu: Yes. This is PrometheusAgent CRD. There's a separate Prometheus CRD for the server mode.

Documentation/proposals/202405-agent-daemonset.md
haanhvu: @simonpasquier once mentioned mixed modes cases. Do we need to do additional things if we want to enable mixed modes?

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+## 1. Why
+
+When deploying Prometheus Agent in Kubernetes, three of the biggest users’ concerns are: load distribution, scalability, and security.
+
+DaemonSet deployment solves all these three concerns:
+* Load distribution: Each Prometheus Agent pod will only scrape the targets located on the same node. Even though the targets on some nodes may produce more metrics than other nodes, the load distribution would be reliable enough.
+* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically added in the nodes that meet user-defined restrictions (if any).
+* Security: Since the scraped targets are local to the Prometheus Agent pod (on the same node), the scope of security problems is reduced to each node.
```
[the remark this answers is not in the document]
pintohutch: > Since the scraped targets are local to the Prometheus Agent pod (on the same node), the scope of security problems is reduced to each node.

Can you give an example? I would think it would actually go the other way.

What if there are exploitations in Prometheus? Wouldn't that compound the security problems in the cluster by the number of nodes (i.e. an attacker now can exploit on any node in the cluster since the container is everywhere).

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,133 @@
...
+Currently, we already have a Prometheus Agent CRD that supports StatefulSet deployment. We’ll add new field(s) in this CRD to enable DaemonSet deployment.
+
+The reason for enhancing existing CRD (instead of introducing a new CRD) is it would take less time to finish the MVP. We’ll let users experiment with the MVP, and in case users report a separate CRD is needed, we’ll separate the logic of DaemonSet deployment into a new CRD later.
+
+The current [Prometheus Agent CRD](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1alpha1.PrometheusAgent) already has sufficient fields for the DaemonSet deployment. The DaemonSet deployment can use all the existing fields in the CRD except the ones related to:
+* Selectors for service, probe, ScrapeConfig
+* Replica
+* Shard
+* Storage
+
```
pintohutch, line 74: > In the MVP, we will simply fail the reconciliation if any of those fields are set.

Would that potentially break scraping if a user were to switch from `mode: StatefulSet` to `mode: DaemonSet`?

Documentation/proposals/202405-agent-daemonset.md
bwplotka: Not sure, what matters is what intention your have here when testing/designing features. It feels mixed modes is a goal then?

Documentation/proposals/202405-agent-daemonset.md
bwplotka: Ah, good point I missed that, thanks!

Documentation/proposals/202405-agent-daemonset.md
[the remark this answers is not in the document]
haanhvu: I was just thinking that the scope of security could be "isolated" to each node, because there's no inter-node communication (between Prometheus and the scrape targets). But after more digging I realized this was too naive view. Scratched it here (and in next commit): https://github.com/prometheus-operator/prometheus-operator/pull/6600#discussion_r1604720536

Documentation/proposals/202405-agent-daemonset.md
haanhvu: If users want to switch from StatefulSet to DaemonSet, they would have to unset the unsupported fields (if they had set them in StatefulSet). Besides documentation, do you have any ideas on how to make this switch smoother?

I discussed with @simonpasquier and @ArthurSens about whether we should simply log, or completely fail the reconciliation when unsupported fields are set. We concluded that having a log might not be enough, because users might neglect it and keep thinking that the unsupported fields would work. Do you have any ideas on this?

pintohutch: Dare I say - a failing webhook on the CRD?

ArthurSens: Ah, very good point! It didn't occur to me that switch from statefulset to daemonset in a live object.

We might need CEL earlier than we thought? from my understanding, it works as an admission webhook

Documentation/proposals/202405-agent-daemonset.md
haanhvu: Removed it. @simonpasquier let us know if you have different opinions.

Documentation/proposals/202405-agent-daemonset.md
simonpasquier: What do you mean by mixed deployment goals @bwplotka?
What I mentioned before is that someone could very well deploy "statefulset" Prometheus/PrometheusAgent resources alongside "daemonset" PrometheusAgent resources.

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,128 @@
...
+* Storage: We will need to spend time studying more about the WAL, different storage solutions provided by Kubernetes, and how to gracefully handle storage in different cases of crashes. For example, there’s an [issue in Prometheus](https://github.com/prometheus/prometheus/issues/8809) showing that samples may be lost if remote write didn’t flush cleanly. We’ll investigate these further after the MVP.
+* Secret management: We’ll need to figure out how to make each Prometheus Agent pod only watch the secrets from the pods on the same node. This is for both enhancing security and reducing load on Kubernetes API server. However, this is not easy to do. So we’ll handle this after the MVP.
+
+
+## 6. How
+
+### 6.1. CRD:
+
+Currently, we already have a Prometheus Agent CRD that supports StatefulSet deployment. We’ll add new field(s) in this CRD to enable DaemonSet deployment.
```
simonpasquier, line 60: (nit) s/Prometheus Agent/PrometheusAgent/
```suggestion
Currently, we already have a PrometheusAgent CRD that supports StatefulSet deployment. We’ll add new field(s) in this CRD to enable DaemonSet deployment.
```

Documentation/proposals/202405-agent-daemonset.md
simonpasquier: Looks good to me.

Documentation/proposals/202405-agent-daemonset.md
simonpasquier: For a first approach, I'd consider that failing the reconciliation is good enough. If the validation can be modeled with CEL, I'd prefer to go this way (a validating webhook is more complex to manage).

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,128 @@
...
+
+Provide an MVP version of the DaemonSet deployment of Prometheus Agent to the Audience.
+In specific, the MVP will need to:
+* Allow users to deploy one Prometheus Agent pod per node.
+* Allow users to restrict which set of nodes they want to deploy Prometheus Agent, if desired.
+* Allow users to set the priority of Prometheus Agent pod compared to other pods on the same node, if desired.
+* Allow each Prometheus Agent pod to only scrape from the pods from PodMonitor that run on the same node as the Agent.
+
+
+## 5. Non-Goals
```
simonpasquier, line 48: Maybe add here that the proposal only covers the PrometheusAgent CRD because this is where we see most of the use cases. I'm not 100% sure that folks won't ask for a daemonset Prometheus CRD and maybe there are valid scenarios for it but I'd rather experiment first with the agent.

Documentation/proposals/202405-agent-daemonset.md
pintohutch: CEL is nice because you don't need a separate webhook service to configure and route to.

Alternatively, you could add some new logic in go code to the [admission webhook](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/webhook.md). This is a nice option if CEL is insufficient for what you want to check (e.g. you have some Prometheus library-based validation or something that cannot be codified easily in CEL).

Documentation/proposals/202405-agent-daemonset.md
simonpasquier: No need for a separate section. Something along those lines for posterity:

```
### 6.3. Targets filtering for pods (PodMonitor support):

We will use field selectors to filter the pods running on the node. For example, to select pods on a specific node, we can do:

...

We've also evaluated using relabeling configuration but it was deemed less efficient since the agent would retrieve more data from the API server.
```

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,128 @@
...
+* Prometheus Agent DaemonSet scales up/down following the scale of nodes.
+* Prometheus Agent DaemonSet selects correctly the pods from PodMonitor on the same node.
+Currently we only set up a Kind cluster of one node for integration tests. Since the test cases for DaemonSet deployment requires at least two nodes, we will need to modify the Kind cluster config for that.
+
+We’ll also need a new user guide explaining how to use this new mode.
+
+
+## 8. Follow-ups
+
+After the Goals of this proposal have been met, we’ll handle what’s stated in the Non-goals section.
```
simonpasquier, line 127: Suggestion to come to an agreement between @ArthurSens and @haanhvu :)

```suggestion
After the Goals of this proposal have been met, we’ll re-evaluate the Non-goals section and see if some of them should/can be addressed.
```

Documentation/proposals/202405-agent-daemonset.md
simonpasquier: FWIW we have a "dedicated" CEL issue https://github.com/prometheus-operator/prometheus-operator/issues/5079

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,121 @@
...
+
+
+## 1. Why
+
+When deploying Prometheus Agent in Kubernetes, two of the biggest users’ concerns are load distribution and scalability.
+
+DaemonSet deployment is a good solution for these:
+* Load distribution: Each Prometheus Agent pod will only scrape the targets located on the same node. Even though the targets on some nodes may produce more metrics than other nodes, the load distribution is reliable enough. This has been proven in [Google Cloud Managed Service for Prometheus (GMP)'s operator](https://github.com/GoogleCloudPlatform/prometheus-engine/) which follows a similar approach.
+* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically added in those nodes. Users can also select which set of nodes they want to deploy Prometheus Agent and the priority of Prometheus Agent pods compared to other pods on the same node.
```
simonpasquier, line 19: (nit)
```suggestion
* Automatic scalability: When new nodes are added to the cluster, new Prometheus Agent pods will be automatically deployed to those nodes. Users can also select which set of nodes they want to deploy Prometheus Agent and the priority of Prometheus Agent pods compared to other pods on the same node.
```

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,121 @@
...
+
+DaemonSet deployment is especially more suitable for Prometheus Agent than Prometheus, since the Agent mode is more lightweight. In specific, it requires (currently around 20 to 30%) less memory, does not produce TSDB blocks on disks, and naturally blocks querying APIs.
```
simonpasquier, line 21: It is kinda implicit but I'd emphasize that Prometheus Agent with daemonset is also best suited for a collect-and-forward approach: when you push all metrics to a central datastore which can scale horizontally itself, you don't have to care about query and rule distribution which is the main issue when sharding Prometheus servers.

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,121 @@
...
+* Storage: We will need to spend time studying more about the WAL, different storage solutions provided by Kubernetes, and how to gracefully handle storage in different cases of crashes. For example, there’s an [issue in Prometheus](https://github.com/prometheus/prometheus/issues/8809) showing that samples may be lost if remote write didn’t flush cleanly. We’ll investigate these further after the MVP.
+* Secret management: We’ll need to figure out how to make each Prometheus Agent pod only watch the secrets from the pods on the same node. This is for both enhancing security and reducing load on Kubernetes API server. However, this is not easy to do. So we’ll handle this after the MVP.
```
simonpasquier, line 55: I think that there's some confusion here: the Agent doesn't have to watch secrets?

simonpasquier (commented):
Virtual approval from my side :)

Documentation/proposals/202405-agent-daemonset.md
bwplotka: > What I mentioned before is that someone could very well deploy "statefulset" Prometheus/PrometheusAgent resources alongside "daemonset" PrometheusAgent resources.

Exactly that - mixed deployment. I mean we should put that in goals section to keep allowing those cases 👍🏽

Documentation/proposals/202405-agent-daemonset.md
pintohutch: > you don't have to care about query and rule distribution which is the main issue when sharding Prometheus servers.

Good callout. With recording/alerting rules are disabled in agent mode. how should the operator handle reconciling existing `PrometheusRule` resources?

It looks like Prometheus will [error](https://github.com/prometheus/prometheus/blob/1081e336a0c4004b6f817c70f0ff4f3810db80fa/config/config.go#L124) on config reloading if they are present.

Apologies if this is too much into the details. I haven't reviewed a PO proposal before and I am unfamiliar with how much is covered at this level 😄

ArthurSens: Currently, changes to PrometheusRule don't trigger reconciliations of PrometheusAgent CRDs 🙂  (We have separate CRDs for Prometheus and PrometheuAgents)

pintohutch: Ahhh - I didn't realize that. Thanks for clarifying!

Documentation/proposals/202405-agent-daemonset.md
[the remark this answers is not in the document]
haanhvu: > In some way it helps a lot with the cost, because it scales with the load, so you don't need to keep buffy Prometheus servers when your cluster is scaled back

Yeah I mentioned automatic scaleup, but forgot to mention automatic scaledown. I'll add this to Scalability section then. We don't have any proof of cost so I wouldn't mention cost here. But automatic scaledown implicitly refer to cost optimization.

Documentation/proposals/202405-agent-daemonset.md
haanhvu: I described this in Pitfalls of the current solution and Audience sections.

haanhvu: I described this in Pitfalls of the current solution and Audience sections in the last commit. But I'm thinking whether I should put it in Goals too.

@simonpasquier @ArthurSens @kakkoyun Should we clarify in How that we won't allow switching from a live StatefulSet to DaemonSet => if users are using StatefulSet and want to deploy mixed mode, they will have to first deleting the StatefulSet object, then deploy DaemonSet, then deploy StatefulSet again?

AFAIU Goals/Non-goals are what guide the How section. If we need to clarify this in How then maybe we need to add mixed deployment in Goals then.

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+## 4. Goals
+
+Provide an MVP version of the DaemonSet deployment of Prometheus Agent to the Audience.
+In specific, the MVP will need to:
+* Allow users to deploy one Prometheus Agent pod per node.
+* Allow users to restrict which set of nodes they want to deploy Prometheus Agent, if desired.
+* Allow users to set the priority of Prometheus Agent pod compared to other pods on the same node, if desired.
+* Allow each Prometheus Agent pod to only scrape from the pods from PodMonitor that run on the same node.
```
simonpasquier, line 49:
```suggestion
* Configure each Prometheus Agent pod to scrape metrics from the pods running on the same node via PodMonitor resources.
```

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,129 @@
...
+* Storage: We will need to spend time studying more about the WAL, different storage solutions provided by Kubernetes, and how to gracefully handle storage in different cases of crashes. For example, there’s an [issue in Prometheus](https://github.com/prometheus/prometheus/issues/8809) showing that samples may be lost if remote write didn’t flush cleanly. We’ll investigate these further after the MVP.
+
+In the MVP version, we will not allow users to directly switch from a live StatefulSet to DaemonSet deployment. Reasons are explained in the CRD subsection in the How section.
+
+
+## 6. How
+
+The MVP version of DaemonSet deployment will be put behind a feature gate.
```
simonpasquier, line 65:
```suggestion
The MVP version of DaemonSet deployment will be put behind a feature gate until we have enough confidence that the functionality is ready for testing by the upstream community at large. It will allow us to develop the feature in lock steps while maintaining the stability of the next releases.
```

simonpasquier (commented):
LGTM. There are a few MarkDown issues reported by the linter that need fixing though.

haanhvu:
> LGTM. There are a few MarkDown issues reported by the linter that need fixing though.

Yeah, do you want to merge it now, or leave it open for a little while as we discussed?

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,129 @@
...
+
+We've also considered using relabel config that filters pods by `__meta_kubernetes_pod_node_name` label. However, we didn't choose to go with this option because it filters pods only after discovering all the pods from PodMonitor, which increases load on Kubernetes API server.
+
+
+## 7. Action Plan
+
+For the implementation, we’ll do what we detailed in the How section. The common logics between StatefulSet and DaemonSet modes will be extracted into one place. We will have a separate `daemonset.go` for the separate logic of the DaemonSet mode.
+
+For the test, we will have unit tests covering new logic, and integration tests covering the basic user cases, which are:
+* Prometheus Agent DaemonSet cannot be deployed when a similar Prometheus Agent StatefulSet resource is there.
```
ArthurSens, line 115: Hmmm, my understanding was that we wanted to allow statefulset and deamonset to run together, as long as they are deployed as different resources.

Have we changed our mind recently?

haanhvu: Yes we allow different resources of StatefulSet and DaemonSet together. I stated here that we don't allow _**similar**_ StatefulSet and DaemonSet resources. Similar means scraping same set of targets. Is that correct understanding? If yes, I'll clarify it in the proposal.

ArthurSens: Could you explain why we won't allow them to scrape the same targets?

haanhvu: As we discussed, in MVP, we won't allow directly switching from a live StatefulSet to DaemonSet. AFAIU we would check if the DaemonSet resource has the same set of targets with a live StatefulSet. If yes, we'll fail the DaemonSet. Is that correct understanding?

And as @simonpasquier described in https://github.com/prometheus-operator/prometheus-operator/pull/6600#discussion_r1617338651, seems like we won't allow StatefulSet and DaemonSet to scrape same set of targets even after MVP?

ArthurSens: Yeah, agreed that initially we won't allow switching from DaemonSet to StatefulSet (and vice-versa), but my understanding from the current wording is that we won't allow separate agents (one sts and another daemonset) to scrape the same targets.

I don't remember us agreeing on the latter. If someone wants sts and daemonset scraping the same targets, as long as they are deployed independently, why would we block it?

haanhvu: Ok agree that we never discussed about blocking this. What @simonpasquier described in https://github.com/prometheus-operator/prometheus-operator/pull/6600#discussion_r1617338651 is just his mixed deployment case.

So back to my question in https://github.com/prometheus-operator/prometheus-operator/pull/6600#discussion_r1615270423: In case user is using StatefulSet, and they want to deploy an additional DaemonSet scraping same set of targets, they would first have to delete the existing StatefulSet, then deploy DaemonSet, then deploy StatefulSet again. Is that correct? 🤔

simonpasquier: We've discussed offline and clarified that we don't want to limit users. It will stay the same as today: it's ok to have 2 Prometheus resources scraping the same set of targets.

Documentation/proposals/202405-agent-daemonset.md
```diff
@@ -0,0 +1,129 @@
...
+* Prometheus Agent DaemonSet is created/deleted successfully.
+* Prometheus Agent DaemonSet is installed on the right nodes.
+* Prometheus Agent DaemonSet scales up/down following the scale of nodes.
```
simonpasquier, line 118: After discussing offline, we realized that it would mean testing the DaemonSet functionality offered by Kubernetes which doesn't add any value to the operator validation.

haanhvu:
@ArthurSens @simonpasquier @kakkoyun I resolved all the comments. We have left this open for a while. Since there're no new reviews, I think we can merge this now.

kakkoyun (approved):
LGTM. 

I'll go ahead and merge it now. We can always send subsequent PRs if we change decisions.

haanhvu:
The auto-merge doesn't work, because the tests in CI don't run (seems like they only run in code-related PRs?)

ArthurSens:
> The auto-merge doesn't work, probably because the tests in CI don't run (seems like they only run in code-related PRs?)

Correct 😅

Issue #6478, opened 2024-04-05
Reject PrometheusRule objects that generate > 512KiB configmap


simonpasquier:
@simonpasquier the title of the issue was generic decently generic, I'm happy to open a new issue as I realize my proposed question is not directly related.

---

The error is obvious:

> `level=error ts=2024-03-21T13:51:21.598484221Z caller=klog.go:116 component=k8s_client_runtime func=ErrorDepth msg="sync \"monitoring/kube-prometheus-stack-prometheus\" failed: failed to make rules ConfigMaps: rule file 'monitoring-kube-prometheus-stack-metrics-catalog-rules-aws-836e95d3-92a2-493e-bee9-6a33bb2cc55b.yaml' is too large for a single Kubernetes ConfigMap"`

In this case, the ruleset that the CRD is trying to create simply exceeds the 1MB limit for ConfigMaps.  So I know the problem, but I'm hoping we can move this error to prior to the CRD accepting the change and thus failing the deployment.  Otherwise, I don't know things are wrong until my attempt at looking at data showcases that I'm missing rules.

_Originally posted by @jtslear in https://github.com/prometheus-operator/prometheus-operator/issues/5085#issuecomment-2012376322_

simonpasquier:
@jtslear IIUC correctly your problem, you've got a PrometheusRule which by itself is above the 1MB limit?

PR #1591, opened 2018-07-06
test/e2e: Decrease config map data limit 

mxinden:
This should help debugging flakes seen among others in https://github.com/coreos/prometheus-operator/pull/1586.

PR #6606, opened 2024-05-18
feat: added validation for prometheusRule size (#6478)

yp969803:
## Description
fixes: #6478
Reject PrometheusRule objects that generate > 1MB configmap

## Type of change

## Verification
added unit test
## Changelog entry

```feat: added validation for prometheusRule size

```


yp969803:
@simonpasquier can u review the pr

simonpasquier (commented):
In fact the original issue is about the internal hardcoded limit of the operator which is 512KiB (1MiB/2):
https://github.com/prometheus-operator/prometheus-operator/blob/7e18b825953f02161e605a6206773d04c1742b80/pkg/prometheus/server/rules.go#L36-L40

IMHO the first things to do would be
* better advertise this limit in the API documentation.
* have a descriptive error message stating the operator limit and the actual value.

I'd also rather see the check in `ValidateRule()` since it would help users that don't run the admission webhook.

Eventually we may want to make the 512MB limit configurable and/or increase the default limit. Unfortunately https://github.com/prometheus-operator/prometheus-operator/pull/1591 has little details on why 50% of 1MiB was picked up.

yp969803:
@simonpasquier done the above changes

yp969803:
@simonpasquier @ArthurSens the pr is ready to merge.

simonpasquier (commented):
I would remove the size check from each operator because it becomes redundant with ValidateRule(). The (minor) change would be a generated rule file could be slightly above the limit in case the namespace label enforcer is enabled but it should be safe (the 512KB limit is arbitrary anyway). https://github.com/prometheus-operator/prometheus-operator/blob/0cf01e54859638560651a3f2438465528aed73b6/pkg/prometheus/server/rules.go#L206-L214

yp969803:

pkg/operator/rules.go
```diff
@@ -158,6 +164,13 @@ func ValidateRule(promRuleSpec monitoringv1.PrometheusRuleSpec) []error {
 	if err != nil {
 		return []error{fmt.Errorf("failed to marshal content: %w", err)}
 	}
+
+	// Check if the size of prometheusRule exceeds 512KB (1048576 Bytes).
```
simonpasquier, line 168:
```suggestion
	// Check if the serialized rules exceed our internal limit.
```

pkg/prometheus/server/rules.go
```diff
@@ -206,7 +200,7 @@ func (c *Operator) selectRuleNamespaces(p *monitoringv1.Prometheus) ([]string, e
 func makeRulesConfigMaps(p *monitoringv1.Prometheus, ruleFiles map[string]string, opts ...operator.ObjectOption) ([]v1.ConfigMap, error) {
 	//check if none of the rule files is too large for a single ConfigMap
 	for filename, file := range ruleFiles {
-		if len(file) > maxConfigMapDataSize {
+		if len(file) > operator.MaxConfigMapDataSize {
```
simonpasquier, line 203: the block shouldn't be needed anymore since the operator should have called ValidateRule() before. 
```suggestion
```

simonpasquier (commented):
Ideally we'd need an e2e test but it can be added later.

yp969803:

pkg/prometheus/server/rules_test.go
```diff
@@ -19,15 +19,14 @@ import (
...
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
 )
 
 func TestMakeRulesConfigMaps(t *testing.T) {
 	t.Run("ShouldReturnAtLeastOneConfigMap", shouldReturnAtLeastOneConfigMap)
-	t.Run("ShouldErrorOnTooLargeRuleFile", shouldErrorOnTooLargeRuleFile)
```
simonpasquier, line 30: this test should be moved to pkg/operator/rules_test.go

pkg/thanos/rules.go
```diff
@@ -207,7 +201,7 @@ func (o *Operator) selectRuleNamespaces(p *monitoringv1.ThanosRuler) ([]string,
 func makeRulesConfigMaps(t *monitoringv1.ThanosRuler, ruleFiles map[string]string, opts ...operator.ObjectOption) ([]v1.ConfigMap, error) {
 	//check if none of the rule files is too large for a single ConfigMap
 	for filename, file := range ruleFiles {
-		if len(file) > maxConfigMapDataSize {
+		if len(file) > operator.MaxConfigMapDataSize {
```
simonpasquier, line 204: same remark than before: `promRuleSelector.Select()` will prevent "too big" rules from reaching here so we can remove this check.

simonpasquier (approved):
Thanks!

PR #6608, opened 2024-05-20
[CHORE] considering global limits over enforced

nicolastakashi:
## Description

Considering global limits over enforced limits when no user limits are set.

## Type of change

## Verification

## Changelog entry

```release-note
considering global limits over enforced limits when no user limits are set.
```


nicolastakashi:
Leveraging global limits instead of enforced limits if no user limit is set as we discussed at #6589

simonpasquier:
Another option would be 
1. Always set `.spec.sampleLimit` to `.spec.enforcedSampleLimit` when it the former is nil/zero and the latter > 0.
2. Don't set per-scrape sample limit in the generated config if the value is nil/zero (and rely on the global limit being always set).

nicolastakashi:
> Another option would be
> 
> 1. Always set `.spec.sampleLimit` to `.spec.enforcedSampleLimit` when it the former is nil/zero and the latter > 0.
> 2. Don't set per-scrape sample limit in the generated config if the value is nil/zero (and rely on the global limit being always set).

I'm Ok with that!
Is it ok to have the same behaviour for all limits?

simonpasquier:
>  Is it ok to have the same behaviour for all limits?

Yes it definitely makes sense to be consistent here. Though I assume that users are mostly concerned about sample limits.

nicolastakashi:
@simonpasquier PTAL 🙏🏽

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2256,39 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
...
-	if cpf.KeepDroppedTargets != nil {
-		slice = cg.WithMinimumVersion("2.47.0").AppendMapItem(slice, "keep_dropped_targets", *cpf.KeepDroppedTargets)
-	}
+
+	slice = cg.appendGlobalLimits(slice, "2.45.0", "sample_limit", cpf.SampleLimit, cpf.EnforcedSampleLimit)
+	slice = cg.appendGlobalLimits(slice, "2.45.0", "target_limit", cpf.TargetLimit, cpf.EnforcedTargetLimit)
+	slice = cg.appendGlobalLimits(slice, "2.45.0", "label_limit", cpf.LabelLimit, cpf.EnforcedLabelLimit)
+	slice = cg.appendGlobalLimits(slice, "2.45.0", "label_name_length_limit", cpf.LabelNameLengthLimit, cpf.EnforcedLabelNameLengthLimit)
+	slice = cg.appendGlobalLimits(slice, "2.45.0", "label_value_length_limit", cpf.LabelValueLengthLimit, cpf.EnforcedLabelValueLengthLimit)
+	slice = cg.appendGlobalLimits(slice, "2.47.0", "keep_dropped_targets", cpf.KeepDroppedTargets, cpf.EnforcedKeepDroppedTargets)
```
ArthurSens, line 2291: Nice! Looks a lot cleaner :)

pkg/prometheus/testdata/KeepDroppedTargets.golden
```diff
@@ -4,6 +4,7 @@ global:
   external_labels:
     prometheus: default/test
     prometheus_replica: $(POD_NAME)
+  keep_dropped_targets: 1000
```
ArthurSens, line 7: Was this change intentional? Do we want to make the same changes to `keep_dropped_targets`?

ArthurSens (commented):
It generally looks good to me, but almost all golden files might need to be renamed.

I think it's also worth updating API docs for all enforced fields, clarifying the expected behavior.

ArthurSens:
This is definitely a breaking change though. 

You can see the changes being made to golden files where `enforced.*limits` were declared but limits were not declared in the scrape objects. All of a sudden a new limit is added to the configuration.

If we proceed with this, updating prometheus-operator to v0.75 can become quite troublesome to some users 😕

pkg/prometheus/testdata/KeepDroppedTargets.golden
nicolastakashi: From what me and @simonpasquier discussed, this should be a consistent behaviour through all the limits.

nicolastakashi:
> This is definitely a breaking change though.
> 
> You can see the changes being made to golden files where `enforced.*limits` were declared but limits were not declared in the scrape objects. All of a sudden a new limit is added to the configuration.
> 
> If we proceed with this, updating prometheus-operator to v0.75 can become quite troublesome to some users 😕

I'm not sure if this is a breaking change, because the behavior is still the same in the end when you're using enforcedSampleLimits the only difference is, instead of setting the sample limit on the scrape config it will be placed on the global config.

wdyt @simonpasquier ?

pkg/prometheus/promcfg.go
```diff
@@ -1617,7 +1616,7 @@ func getLimit(user *uint64, enforced *uint64) *uint64 {
 	}
 
 	if user == nil {
-		return enforced
+		return user
```
simonpasquier, line 1619: I'm confused by this change.

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2256,39 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
 	return append(slice, yaml.MapItem{Key: "evaluation_interval", Value: evaluationInterval})
 }
 
+func (cg *ConfigGenerator) appendGlobalLimits(slice yaml.MapSlice, version string, limitKey string, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
+	if limit == nil && enforcedLimit != nil && *enforcedLimit > 0 {
```
simonpasquier, line 2260: (suggestion) and you can get rid of the next block IIUC
```suggestion
	if ptr.Deref(limit, 0) == 0 && ptr.Deref(enforcedLimit, 0) > 0 {
```

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2256,39 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
...
+		slice = cg.WithMinimumVersion(version).AppendMapItem(slice, limitKey, *enforcedLimit)
+		return slice
+	}
+
+	if enforcedLimit != nil && *enforcedLimit > 0 && limit != nil && *limit < 1 {
+		slice = cg.WithMinimumVersion(version).AppendMapItem(slice, limitKey, *enforcedLimit)
+	}
+
+	if limit != nil && *limit > 0 {
```
simonpasquier, line 2269:
```suggestion
	if ptr.Deref(limit, 0) > 0 {
```

simonpasquier:
Technically it's a change so it should be highlighted as such in the changelog. Today if you configure both `.spec.sampleLimit` and `.spec.enforcedSampleLimit` in the Prometheus object, only the latter is taken into account so I wouldn't expect users to define both fields in practice. If you've configure only one of them, there should be no visible change.
So my position is that the new behavior makes more sense and it should impact a small fraction of users (if any).

> I think it's also worth updating API docs for all enforced fields, clarifying the expected behavior.

Agreed.

pkg/prometheus/testdata/TargetLimits-1_Versionv2.21.0_Enforce1000.golden
```diff
@@ -66,4 +66,3 @@ scrape_configs:
     - __tmp_hash
     regex: $(SHARD)
     action: keep
-  target_limit: 1000
```
simonpasquier, line 69: hmm we had a bug?

pkg/prometheus/testdata/LabelLimits_NoLimit_v2.27.0_enforceLimit1000.golden
```diff
@@ -66,4 +66,3 @@ scrape_configs:
     - __tmp_hash
     regex: $(SHARD)
     action: keep
-  label_limit: 1000
```
simonpasquier, line 69: is this correct?

pkg/prometheus/promcfg.go
nicolastakashi: I just pushed another version, can you take a look?

pkg/prometheus/promcfg.go
```diff
@@ -1050,7 +1050,6 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 
 	relabelings = generateAddressShardingRelabelingRules(relabelings, shards)
 	cfg = append(cfg, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
-
```
simonpasquier, line 1053: (nit) can we keep the empty line?

pkg/prometheus/testdata/TargetLimits-1_Versionv2.21.0_Enforce1000.golden
simonpasquier: @nicolastakashi thoughts?

simonpasquier (commented):
I believe that we have an issue to fix depending on Prometheus versions... IIUC the global limits require 2.45.0 while scrape limits where are available since 2.21.0 & 2.27.0.

I'm afraid that we'll have to set the limits on the individual scrape configs rather than relying on global settings to accommodate older versions.

pkg/prometheus/testdata/TargetLimits-1_Versionv2.21.0_Enforce1000.golden
nicolastakashi: @simonpasquier checking yet, I'll update soon.

simonpasquier:
Hmm enforcedSampleLimits worked previously with versions <  2.45 so it's a bit extreme to say that it doesn't anymore.

pkg/prometheus/testdata/TargetLimits-1_Versionv2.21.0_Enforce1000.golden
simonpasquier: I suppose that it's the isssue with global sample_limit not being supported before 2.45.

nicolastakashi:
> Hmm enforcedSampleLimits worked previously with versions < 2.45 so it's a bit extreme to say that it doesn't anymore.
@simonpasquier I'm keeping the current behavior is version is < 2.45, so we don't introduce any breaking change on this case, wdyt?

pkg/prometheus/promcfg.go
```diff
@@ -1611,16 +1615,16 @@ func generateRunningFilter() yaml.MapSlice {
...
 		return user
 	}
 
-	if user == nil {
+	if isGlobalLimitNotSupported && user == nil {
 		return enforced
 	}
 
-	if *enforced > *user {
+	if ptr.Deref(user, 0) <= ptr.Deref(enforced, 0) {
```
simonpasquier, line 1627: (nit)
```suggestion
	if ptr.Deref(enforced, 0) > ptr.Deref(user, 0) {
```

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2261,35 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
 	return append(slice, yaml.MapItem{Key: "evaluation_interval", Value: evaluationInterval})
 }
 
+func (cg *ConfigGenerator) appendGlobalLimits(slice yaml.MapSlice, limitKey string, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
+	if ptr.Deref(limit, 0) == 0 && ptr.Deref(enforcedLimit, 0) > 0 {
+		slice = cg.AppendMapItem(slice, limitKey, *enforcedLimit)
+		return slice
```
simonpasquier, line 2267: (nit)
```suggestion
		return cg.AppendMapItem(slice, limitKey, *enforcedLimit)
```

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2261,35 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
 	return append(slice, yaml.MapItem{Key: "evaluation_interval", Value: evaluationInterval})
 }
 
+func (cg *ConfigGenerator) appendGlobalLimits(slice yaml.MapSlice, limitKey string, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
+	if ptr.Deref(limit, 0) == 0 && ptr.Deref(enforcedLimit, 0) > 0 {
```
simonpasquier, line 2265:
```suggestion
	// Use the enforced limit if no global limit is defined to ensure that scrape jobs without an explicit limit inherit the enforced limit value.
	if ptr.Deref(limit, 0) == 0 && ptr.Deref(enforcedLimit, 0) > 0 {
```

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2261,35 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
...
+		slice = cg.AppendMapItem(slice, limitKey, *enforcedLimit)
+		return slice
+	}
+
+	if ptr.Deref(limit, 0) > 0 {
+		slice = cg.AppendMapItem(slice, limitKey, *limit)
+		return slice
```
simonpasquier, line 2272:
```suggestion
		return cg.AppendMapItem(slice, limitKey, *limit)
```

pkg/prometheus/promcfg.go
```diff
@@ -1611,16 +1615,16 @@ func generateRunningFilter() yaml.MapSlice {
...
 }
 
-func getLimit(user *uint64, enforced *uint64) *uint64 {
+func getLimit(user *uint64, enforced *uint64, isGlobalLimitNotSupported bool) *uint64 {
 	if enforced == nil {
 		return user
 	}
 
-	if user == nil {
+	if isGlobalLimitNotSupported && user == nil {
```
simonpasquier, line 1623: I'd find it easier to have the version inline here
```suggestion
	// If the Prometheus version doesn't support global limit 
	if cg.version.LT(semver.MustParse("2.45.0")) && user == nil {
```

pkg/prometheus/promcfg.go
```diff
@@ -1611,16 +1615,16 @@ func generateRunningFilter() yaml.MapSlice {
...
 		return enforced
 	}
```
simonpasquier, line 1625: My understanding is that when Prometheus supports global limits then we don't want to set scrape limit == enforced.
I find it also easier to have the version inline here.

```suggestion
	// Prometheus >= 2.45.0 supports global limits in which case the global limit gets applied for all targets that don't specify an explicit limit.
	if cg.version.GTE(semver.MustParse("2.45.0")) && ptr.Deref(user, 0) == 0 {
		return nil
	}
```

pkg/prometheus/promcfg_test.go
```diff
@@ -3321,33 +3321,49 @@ func TestTrackTimestampsStaleness(t *testing.T) {
...
 			enforcedLimit: 1000,
 			limit:         500,
 			golden:        "SampleLimits_Limit500.golden",
 		},
+		{
+			globalLimit:   1000,
+			enforcedLimit: 2000,
+			limit:         -1,
+			golden:        "SampleLimits_GlobalLimit1000_Enforce2000.golden",
+		},
```
simonpasquier, line 3358: I wonder if we don't need more test cases? In particular with version < 2.4.50?

simonpasquier (commented):
Can you also update the API documentation to clarify the behavior change?

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2258,34 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
...
 
+func (cg *ConfigGenerator) appendGlobalLimits(slice yaml.MapSlice, limitKey string, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
+	// Use the enforced limit if no global limit is defined to ensure that scrape jobs without an explicit limit inherit the enforced limit value.
+	if ptr.Deref(limit, 0) == 0 && ptr.Deref(enforcedLimit, 0) > 0 {
+		return cg.AppendMapItem(slice, limitKey, *enforcedLimit)
+	}
+
+	if ptr.Deref(limit, 0) > 0 {
+		return cg.AppendMapItem(slice, limitKey, *limit)
+	}
```
simonpasquier, line 2269: It reads more easily for me. Though it brings the question of what to do when global limit > enforced limit? I suggest that we clamp down the global limit to the enforced value in this case (CEL could help later on to validate this at the API level).
```suggestion
	if ptr.Deref(limit, 0) > 0 {
		return cg.AppendMapItem(slice, limitKey, *limit)
	}

	// <explain why enforcedlimit is used>
	if ptr.Deref(enforcedLimit, 0) > 0 {
		return cg.AppendMapItem(slice, limitKey, *enforcedLimit)
	}
```

pkg/prometheus/promcfg.go
```diff
@@ -1611,16 +1611,17 @@ func generateRunningFilter() yaml.MapSlice {
 	}
 }
 
-func getLimit(user *uint64, enforced *uint64) *uint64 {
+func (cg *ConfigGenerator) getLimit(user *uint64, enforced *uint64) *uint64 {
 	if enforced == nil {
```
simonpasquier, line 1615: (suggestion) easier to use ptr.Deref() from the start?
```suggestion
	if ptr.Deref(enforced, 0) == 0 {
```

pkg/prometheus/promcfg.go
```diff
@@ -1611,16 +1611,17 @@ func generateRunningFilter() yaml.MapSlice {
...
 		return user
 	}
 
-	if user == nil {
+	// If the Prometheus version doesn't support global limit
```
simonpasquier, line 1619: Would it be easier to follow with this logic?
* if *user == 0
  * if version >= 2.45.0, return nil (global limit applies)
  * otherwise return enforced

simonpasquier (commented):
I don't remember if I said it already but I think that the new behavior needs to be documented in the API.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -514,6 +514,9 @@ type CommonPrometheusFields struct {
 	// It is meant to be used by admins to keep the overall number of
 	// samples/series under a desired limit.
 	//
+	// Starting from Prometheus v2.45.0, this field is used to define global limits if `spec.sampleLimit` is not set.
+	// Admins can combine this field with `spec.sampleLimit` to provide a global upper limit and a lower limit for individual objects.
```
simonpasquier, line 518: It's still a bit unclear IMHO. What if we explain in details the decision tree?

```
When both `enforcedSampleLimit` and `sampleLimit` are defined and greater than zero, the following rules apply:
* Scrape objets without a defined sampleLimit value will inherit the global sampleLimit value (Prometheus <= 2.45.0) or the enforcedSampleLimit value (Prometheus > v2.45.0).
* Scrape objets with a sampleLimit value less than or equal to enforcedSampleLimit keep their specific value.
* Scrape objets with a sampleLimit value greater than enforcedSampleLimit are set to enforcedSampleLimit.
```

From my POV, it's ok to have the detailed documentation for 1 field (e.g. enforced sample limit) and have a reference (e.g. "See `enforcedSampleLimit` to understand the interactions with `<field>`."). It might also good to have a reference to the enforced fields in each global field.

pkg/prometheus/promcfg.go
```diff
@@ -1611,16 +1611,20 @@ func generateRunningFilter() yaml.MapSlice {
...
-func getLimit(user *uint64, enforced *uint64) *uint64 {
-	if enforced == nil {
+func (cg *ConfigGenerator) getLimit(user *uint64, enforced *uint64) *uint64 {
+	if ptr.Deref(enforced, 0) == 0 {
 		return user
 	}
 
-	if user == nil {
+	if ptr.Deref(user, 0) == 0 {
+		// Prometheus 2.45.0 introduced the ability to use global sample_limit
```
simonpasquier, line 1620:
```suggestion
		// With Prometheus >= 2.45.0, the limit value in the global section will always apply, hence there's no need to set the value explicitly.
```

pkg/prometheus/promcfg.go
```diff
@@ -2257,29 +2261,34 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
 	return append(slice, yaml.MapItem{Key: "evaluation_interval", Value: evaluationInterval})
 }
 
+func (cg *ConfigGenerator) appendGlobalLimits(slice yaml.MapSlice, limitKey string, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
+	if ptr.Deref(limit, 0) > 0 {
```
simonpasquier, line 2265: should we check that limit <= enforced? Otheriwse is there a risk of the "soft" limit being higher than the enforced limit?

nicolastakashi: @simonpasquier wdyt?

pkg/prometheus/testdata/SampleLimits_GlobalLimit1000_Enforce2000.golden
```diff
@@ -0,0 +1,69 @@
+global:
+  evaluation_interval: 30s
+  scrape_interval: 30s
+  external_labels:
+    prometheus: default/test
+    prometheus_replica: $(POD_NAME)
+  sample_limit: 2000
```
simonpasquier, line 7: shouldn't it be 1,000?

pkg/prometheus/promcfg.go
```diff
@@ -2289,29 +2293,38 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
...
+		if ptr.Deref(enforcedLimit, 0) > 0 && *limit < *enforcedLimit {
```
simonpasquier, line 2298: shouldn't it be `*limit > *enforcedLimit`?

nicolastakashi: Don't think so!
Sample Limit == Soft Limit

simonpasquier: I'd rather expect the opposite: enforced (hard) limits should always be greater than global (soft) limits?
In any case it should be documented in the `sampleLimits` field.

nicolastakashi:
@ArthurSens @simonpasquier PTAL

pkg/prometheus/promcfg.go
```diff
@@ -2289,29 +2293,38 @@ func (cg *ConfigGenerator) appendEvaluationInterval(slice yaml.MapSlice, evaluat
 	return append(slice, yaml.MapItem{Key: "evaluation_interval", Value: evaluationInterval})
 }
 
+func (cg *ConfigGenerator) appendGlobalLimits(slice yaml.MapSlice, limitKey string, limit *uint64, enforcedLimit *uint64) yaml.MapSlice {
+	if ptr.Deref(limit, 0) > 0 {
+		if ptr.Deref(enforcedLimit, 0) > 0 && *limit > *enforcedLimit {
+			level.Warn(cg.logger).Log("msg", limitKey, "limit is greather than enforced limit, using enforced limit")
```
simonpasquier, line 2299:
```suggestion
			level.Warn(cg.logger).Log("msg", fmt.Sprintf("%q is greater than the enforced limit, using enforced limit", limitKey), "limit", *limit, "enforced_limit", *enforcedLimit)
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -514,6 +514,12 @@ type CommonPrometheusFields struct {
 	// It is meant to be used by admins to keep the overall number of
 	// samples/series under a desired limit.
 	//
+	// When both `enforcedSampleLimit` and `sampleLimit` are defined and greater than zero, the following rules apply:
```
simonpasquier, line 517: shall we say that when sampleLimit > enforcedSampleLimit, we consider only enforcedSampleLimit?

nicolastakashi: its already there I think.

```If Prometheus version is >= 2.45.0 and the `enforcedTargetLimit` is greater than the `targetLimit`, the `targetLimit` will be set to `enforcedTargetLimit`.```

simonpasquier: :+1:

simonpasquier (approved):
Thanks! :tada:

PR #6615, opened 2024-05-24
[BUGFIX] Fix PrometheusAgent reconciliation for the statefulset changes

junotx:
## Description

_Fix the prometheus agent statefulset key match format because the statefulset is named with a `prom-agent` prefix._

## Type of change

## Verification

## Changelog entry

```release-note
[BUGFIX] Fix PrometheusAgent reconciliation for the statefulset changes.
```

### DIFF
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -17,6 +17,7 @@ package prometheusagent
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nicolastakashi:
Hey @junotx 👋🏽  
Is this related to an issue? 
If so, could you link it to the PR description? so we can get more context about the change.

junotx:
@nicolastakashi no. 
before this change, it cannot be enqueued when the prometheus agent statefulset is updated, because its name does not match [the format](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/operator.go#L39). for example, if the prometheus agent statefulset is deleted, the operator will not rebuilt it at once, unless the PrometheusAgent cr is updated or the operator is restarted.

simonpasquier:
Thanks for the PR! We'd need unit tests (similar to TestStatefulSetKeyToPrometheusKey). Also `StatefulSetKeyToPrometheusKey` should be moved to pkg/prometheus/server since the agent and server implementations are different.

benjaminhuo:
That's great, thanks @junotx

PR #6620, opened 2024-05-27
Use functional options pattern for Prometheus Controller

ArthurSens:
## Description

After #6491 got merged, I went back to #6198. While I was trying to pass the feature-gate value to the controller, I noticed how we're passing way too many booleans to the function signature. Something that will only grow as we introduce more optionality as feature flags.

Here I'm proposing that we use a functional options pattern to create Prometheus Controller, allowing us to introduce more options without much effort. (The same can be done to other controllers as well)

In #6198, the plan would be to have a `WithGracefulShutdownSupport()` function, that would enable the feature in the controller's level.

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -95,8 +95,28 @@ type Operator struct {
 	eventRecorder record.EventRecorder
 }
 
+type ControllerOptions func(*Operator)
+
+func WithEndpointSlice() ControllerOptions {
+	return func(o *Operator) {
+		o.endpointSliceSupported = true
+	}
+}
+
+func WithScrapeConfig() ControllerOptions {
+	return func(o *Operator) {
+		o.scrapeConfigSupported = true
+	}
+}
+
+func WithStorageClassValidation() ControllerOptions {
+	return func(o *Operator) {
+		o.canReadStorageClass = true
+	}
+}
+
 // New creates a new controller.
-func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger log.Logger, r prometheus.Registerer, scrapeConfigSupported, canReadStorageClass bool, erf operator.EventRecorderFactory) (*Operator, error) {
+func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger log.Logger, r prometheus.Registerer, erf operator.EventRecorderFactory, opts ...ControllerOptions) (*Operator, error) {
 	logger = log.With(logger, "component", controllerName)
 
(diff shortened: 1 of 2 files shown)


pkg/prometheus/server/operator.go
```diff
@@ -88,15 +88,42 @@ type Operator struct {
...
+type controllerOptions struct {
 	endpointSliceSupported bool
 	scrapeConfigSupported  bool
 	canReadStorageClass    bool
+}
 
-	eventRecorder record.EventRecorder
+type ControllerOptions func(*controllerOptions)
+
+func WithEndpointSliceSupport() ControllerOptions {
```
simonpasquier, line 103: (nit)
```suggestion
func WithEndpointSlice() ControllerOptions {
```

pkg/prometheus/server/operator.go
```diff
@@ -88,15 +88,42 @@ type Operator struct {
...
+	return func(o *controllerOptions) {
+		o.endpointSliceSupported = true
+	}
+}
+
+func WithScrapeConfigSupport() ControllerOptions {
```
simonpasquier, line 109: (nit)
```suggestion
func WithScrapeConfig() ControllerOptions {
```

pkg/prometheus/server/operator.go
```diff
@@ -88,15 +88,42 @@ type Operator struct {
...
+	return func(o *controllerOptions) {
+		o.scrapeConfigSupported = true
+	}
+}
+
+func CanReadStorageClass() ControllerOptions {
```
simonpasquier, line 115: (nit)
```suggestion
func WithStorageClassValidation() ControllerOptions {
```

pkg/prometheus/server/operator.go
```diff
@@ -88,15 +88,42 @@ type Operator struct {
 	reconciliations *operator.ReconciliationTracker
 	statusReporter  prompkg.StatusReporter
 
+	eventRecorder record.EventRecorder
+	options       controllerOptions
+}
+
+type controllerOptions struct {
```
simonpasquier, line 95: Do we need an intermediate struct? `o.options.scrapeConfigSupported` is a bit heavy to read.

simonpasquier (commented):
Nice!

pkg/prometheus/server/operator.go
ArthurSens: No, no need for it :)

I was just grouping things together, but no strong opinions here

simonpasquier (approved):

PR #6624, opened 2024-05-28
Make a cluster of 2 nodes for e2e

haanhvu:
## Description

Currently we're running a cluster of only 1 node for e2e tests. For the future e2e tests for DaemonSet deployment of Prometheus Agent (proposal in #6600), we will need a cluster of at least 2 nodes.
The DaemonSet deployment hasn't been implemented yet. But we want to start with deploying 2 nodes for the current e2e tests, to see if it breaks anything.

## Type of change

## Verification
As long as the change doesn't break anything, it's good to go.

## Changelog entry

Make a cluster of 2 nodes for e2e


haanhvu:
> The PR title and commit message say "2 nodes" but the kind config is 1 control plane + 2 workers. I don't have strong preferences and the e2e tests pass with this new config. Can you update the commit & description?

Yeah, I'll change the PR and commit to be more precise.

simonpasquier:
Thanks @haanhvu! I didn't have time to check if there's any visible impact on the test duration. Could you check this?

haanhvu:
@simonpasquier @ArthurSens @kakkoyun Besides the kind config, do we need to modify any code for e2e tests to enable 2 (worker) nodes? For example: https://github.com/prometheus-operator/prometheus-operator/blob/605966a8328cc1c1b70bf946da42db2020323407/test/framework/framework.go#L760

simonpasquier:
Good catch! Can we check the number of available workers and adjust the deployment if workers == 1? It would avoid breaking people running the e2e tests locally with a single node cluster.

haanhvu:
> @simonpasquier @ArthurSens @kakkoyun Besides the kind config, do we need to modify any code for e2e tests to enable 2 (worker) nodes? For example:
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/605966a8328cc1c1b70bf946da42db2020323407/test/framework/framework.go#L760



> Good catch! Can we check the number of available workers and adjust the deployment if workers == 1? It would avoid breaking people running the e2e tests locally with a single node cluster.

@simonpasquier The code I pointed above is for admission webhook server deployment. You mean we need to have different admission webhook server deployments for `workers == 1` and `workers == 2`? Now rereading the code again I'm not sure we'll need to do this. Now we have more nodes but the features stay the same, so the validation/tests for the features should also stay the same? No matter how many nodes we have? 🤔

haanhvu:
> Thanks @haanhvu! I didn't have time to check if there's any visible impact on the test duration. Could you check this?

My PR:

- E2E tests (alertmanager): 35m 38s
- E2E tests (prometheus): 45m 35s
- E2E tests (prometheusAllNS): 26m 5s
- E2E tests (thanosruler): 10m 12s
- E2E tests (operatorUpgrade): 6m 34s


PR #6623 (I picked this PR because it has same size as mine, also passed, but only run on 1 node):

- E2E tests (alertmanager): 35m 44s 
- E2E tests (prometheus): 44m 39s
- E2E tests (prometheusAllNS): 25m 2s
- E2E tests (thanosruler): 10m 37s
- E2E tests (operatorUpgrade): 6m 34s


Seems like the test duration doesn't change much with 2 nodes.

nicolastakashi:
what about use k3d, the startup time is much faster

simonpasquier (approved):

PR #6636, opened 2024-05-30
Add NODE_NAME env in config reloader

haanhvu:
## Description

Add `NODE_NAME` env in config reloader for DaemonSet deployment of Prometheus Agent. In DaemonSet deployment, Prometheus Agent is deployed per node and only scrapes targets located on the same node. That's why it needs to know its `NODE_NAME` to filter targets. Detailed reasons and how `NODE_NAME` env is used are explained in the proposal #6600.

## Type of change

## Verification
Unit tests

## Changelog entry

Add NODE_NAME env in config reloader

### DIFF
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -62,6 +62,7 @@ type ConfigReloader struct {
 	volumeMounts       []v1.VolumeMount
 	watchedDirectories []string
 	useSignal          bool
+	needNodeNameEnv    bool
 }
 
 type ReloaderOption = func(*ConfigReloader)
@@ -178,6 +179,13 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 	}
 }
 
+// NeedNodeNameEnv sets the needNodeNameEnv option for the config-reloader container.
+func NeedNodeNameEnv(need bool) ReloaderOption {
+	return func(c *ConfigReloader) {
+		c.needNodeNameEnv = need
+	}
+}
+
 // CreateConfigReloader returns the definition of the config-reloader
 // container.
 func CreateConfigReloader(name string, options ...ReloaderOption) v1.Container {
@@ -196,15 +204,18 @@ func CreateConfigReloader(name string, options ...ReloaderOption) v1.Container {
 					FieldRef: &v1.ObjectFieldSelector{FieldPath: "metadata.name"},
 				},
 			},
-			{
-				Name: NodeNameEnvVar,
-				ValueFrom: &v1.EnvVarSource{
-					FieldRef: &v1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
-				},
-			},
 		}
 		ports []v1.ContainerPort
 	)
+
+	if configReloader.needNodeNameEnv {
+		envVars = append(envVars, v1.EnvVar{
+			Name: NodeNameEnvVar,
+			ValueFrom: &v1.EnvVarSource{
+				FieldRef: &v1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
+			},
+		})
+	}
 
 	if configReloader.initContainer {
 		args = append(args, fmt.Sprintf("--watch-interval=%d", 0))
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -180,9 +180,9 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 }
 
 // WithNodeNameEnv sets the withNodeNameEnv option for the config-reloader container.
-func WithNodeNameEnv(nodeNameNeeded bool) ReloaderOption {
+func WithNodeNameEnv() ReloaderOption {
 	return func(c *ConfigReloader) {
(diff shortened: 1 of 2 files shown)


kakkoyun (dismissed):

kakkoyun:
> (the reasons are explained in proposal #6600).

We could still drop one or two sentences to explain it for the sake of the reviewer, who doesn't have a lot of context.

simonpasquier (commented):
For safety, could we check the feature gate and only include the env var if it is set? I'm not too worried about breaking existing setups but it's also hard to be sure that nobody will be impacted. Having it behind the feature gate ensures that we're not changing anything now.

haanhvu:
> > (the reasons are explained in proposal #6600).
> 
> We could still drop one or two sentences to explain it for the sake of the reviewer, who doesn't have a lot of context.

Thanks. I added a small summary.


> For safety, could we check the feature gate and only include the env var if it is set? I'm not too worried about breaking existing setups but it's also hard to be sure that nobody will be impacted. Having it behind the feature gate ensures that we're not changing anything now.

Actually, I'm thinking of creating a separate `CreateConfigReloaderForDaemonSet` that has this NODE_NAME env. The existing `CreateConfigReloader` for StatefulSet will stay the same. Is that ok?


Also, do we need to add test for this? I don't see we have any similar tests for the existing `POD_NAME` env. And I'm not sure whether the implementation has been sufficient for testing. But if we want to test that `NODE_NAME` is correctly retrieved, maybe there's a way to set up a mock node for test? (not clearly sure how, just feel this seems viable)


@simonpasquier @kakkoyun @ArthurSens

simonpasquier:
My recommendation would be to create a new WithNodeNameEnvVar() option, similar to:

https://github.com/prometheus-operator/prometheus-operator/blob/e3412eec6b3843fd63d6c21f11c1ebabc9cc3aa9/pkg/operator/config_reloader.go#L84-L90

We have also some unit tests for this package (though not exhaustive) so it would be good to add some coverage for this new option.

Then when calling [BuildConfigReloader()](https://github.com/prometheus-operator/prometheus-operator/blob/e3412eec6b3843fd63d6c21f11c1ebabc9cc3aa9/pkg/prometheus/agent/statefulset.go#L291-L298) in the agent controller, the value is true if the feature gate is enabled.

simonpasquier:
> Then when calling [BuildConfigReloader()](https://github.com/prometheus-operator/prometheus-operator/blob/e3412eec6b3843fd63d6c21f11c1ebabc9cc3aa9/pkg/prometheus/agent/statefulset.go#L291-L298) in the agent controller, the value is true if the feature gate is enabled.

Discussed offline with @ArthurSens too and the other condition would be that the PrometheusAgent spec has `mode: DaemonSet` hence a preliminary PR needs to add the field to the CRD.

pkg/operator/config_reloader.go
```diff
@@ -174,6 +179,13 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 	}
 }
 
+// NeedNodeNameEnv sets the needNodeNameEnv option for the config-reloader container.
+func NeedNodeNameEnv(need bool) ReloaderOption {
```
ArthurSens, line 183:
```suggestion
func WithNodeNameEnv(need bool) ReloaderOption {
```

ArthurSens (commented):
just nitpicking :P

pkg/operator/config_reloader.go
```diff
@@ -58,6 +62,7 @@ type ConfigReloader struct {
 	volumeMounts       []v1.VolumeMount
 	watchedDirectories []string
 	useSignal          bool
+	needNodeNameEnv    bool
```
ArthurSens, line 65:
```suggestion
	addNodeNameEnv    bool
```

ArthurSens (dismissed):
LGTM 

now that https://github.com/prometheus-operator/prometheus-operator/pull/6640 got merged, we can find a way of using this variable if mode==daemoset and featuregate is enabled

It can be done in a separate PR if you prefer

haanhvu:
> LGTM
> 
> now that #6640 got merged, we can find a way of using this variable if mode==daemoset and featuregate is enabled
> 
> It can be done in a separate PR if you prefer

Now reading the code, I'm not sure we need to do this.

`withNodeNameEnv` is false by default. It's only true if `WithNodeNameEnv(true)` is called (do we just need `WithNodeNameEnv()` instead?). So maybe we would just need to activate this option in `BuildConfigReloader()` in `makeDaemonSetSpec()` (upcoming), the existing [BuildConfigReloader()](https://github.com/prometheus-operator/prometheus-operator/blob/e3412eec6b3843fd63d6c21f11c1ebabc9cc3aa9/pkg/prometheus/agent/statefulset.go#L291-L298) in `makeStatefulSetSpec` would just stay the same, i.e., without this option.

If it works this way, I don't think we need to check mode == daemoset and featuregate is enabled, right?

kakkoyun (approved):

simonpasquier:
> If it works this way, I don't think we need to check mode == DaemonSet and feature gate is enabled, right?

Correct :+1: 

I'm merging it as-is since it introduces no risk of breakage :)

PR #6640, opened 2024-06-03
Add `mode` field in PrometheusAgent CRD

haanhvu:
## Description

Add `mode` field in PrometheusAgent CRD that allows either StatefulSet (already supported) or DaemonSet (will be supported) in PrometheusAgent CRD.
At the moment, we just need the field to have different configuration for the two modes (for example, to add `NODE_NAME` env in config reloader for DaemonSet mode in [#6636](https://github.com/prometheus-operator/prometheus-operator/pull/6636#issuecomment-2139860025)). We don't change the operator's behaviors according to the field yet. This will be done in later PRs.

## Type of change

## Verification
At the moment we just need to add the field. No need for verification yet.

## Changelog entry

Add `mode` field in PrometheusAgent CRD


pkg/apis/monitoring/v1alpha1/prometheusagent_types.go
```diff
@@ -91,5 +91,9 @@ func (l *PrometheusAgentList) DeepCopyObject() runtime.Object {
 // https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
 // +k8s:openapi-gen=true
 type PrometheusAgentSpec struct {
+	// +kubebuilder:default="StatefulSet"
+	// +kubebuilder:validation:Enum=StatefulSet;DaemonSet
+	Mode string `json:"mode,omitempty"`
```
simonpasquier, line 96: I'd add a few sentences to set the expectations. I'll also drop the default marker to avoid confusing the users (otherwise they'll get `mode: StatefulSet` in their API responses).
(nit) I realize that the feature gate is `PrometheusAgentDaemonset` while the mode value says Daemon**S**et. Not a big deal but the latter is more [correct](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) so we might want to fix the feature gate name.

```suggestion
        // Mode defines how the Prometheus operator deploys the PrometheusAgent pods. 
        // For now this field has no effect.
        //
        // (Alpha) Using this field requires the `PrometheusAgentDaemonset` feature gate to be enabled.
        //
	// +kubebuilder:validation:Enum=StatefulSet;DaemonSet
	// +optional
	Mode *string `json:"mode,omitempty"`
```

haanhvu: Thanks. I fixed all these in the last commit (feature gate is not related to this PR, but it's just a typo fix though)

pkg/apis/monitoring/v1alpha1/prometheusagent_types.go
```diff
@@ -91,5 +91,14 @@ func (l *PrometheusAgentList) DeepCopyObject() runtime.Object {
...
+	// Mode defines how the Prometheus operator deploys the PrometheusAgent pod(s).
+	// For now this field has no effect.
+	//
+	// (Alpha) Using this field requires the `PrometheusAgentDaemonSet` feature gate to be enabled.
+	//
+	// +kubebuilder:validation:Enum=StatefulSet;DaemonSet
+	// +optional
+	Mode *string `json:"mode,omitempty"`
+
 	monitoringv1.CommonPrometheusFields `json:",inline"`
```
simonpasquier, line 103: stylistic suggestion: I'd put the inline struct first.

```suggestion
	monitoringv1.CommonPrometheusFields `json:",inline"`

	// Mode defines how the Prometheus operator deploys the PrometheusAgent pod(s).
	// For now this field has no effect.
	//
	// (Alpha) Using this field requires the `PrometheusAgentDaemonSet` feature gate to be enabled.
	//
	// +kubebuilder:validation:Enum=StatefulSet;DaemonSet
	// +optional
	Mode *string `json:"mode,omitempty"`
```

simonpasquier (commented):
LGTM. I'll let @ArthurSens and @kakkoyun chime in.

pkg/apis/monitoring/v1alpha1/prometheusagent_types.go
haanhvu: This makes sense. But I don't know if changing the order of fields in the codebase also changes the order of fields in the doc? I think it's best for users to see the `mode` field before the fields of the modes.

ArthurSens (approved):
LGTM, nothing to add :)

pkg/apis/monitoring/v1alpha1/prometheusagent_types.go
haanhvu: I just checked the doc. Turned out the order of fields in doc doesn't follow the order in code, but the alphabet... I'll change the order in the code as you suggested then. But eventually I think it would be best to somehow show `mode` at the top. @ArthurSens Is it possible to do that in the doc? (asking you since you're working on doc and website)

simonpasquier: cc @AshwinSriram11 :)

simonpasquier (approved):

pkg/apis/monitoring/v1alpha1/prometheusagent_types.go
ArthurSens: I don't know, we need to investigate 😬

AshwinSriram11: Currently `mode` is showing at the top(on the website). I will check how to change the order.

haanhvu: Oh, really? Then it's good. That's what we want. No need to do anything more.

> I just checked the doc. Turned out the order of fields in doc doesn't follow the order in code, but the alphabet...

Scratch this. I checked this in `bundle.yaml`, while I should have checked `api.md` instead. Sorry for that :)

PR #6537, opened 2024-04-19
chore: refactor the assets package

simonpasquier:
## Description

This commit simplifies the API of the assets package. To limit the impact, it tackles only Basic Auth secrets for now.

Previous API:

```
// storing the credentials from function A
err = store.AddBasicAuth(ctx, namespace, httpConfig.BasicAuth, "some key")

// retrieving the credentials from function B
basicAuth := store.BasicAuthAssets["some key"]
```

New API:

```
// storing the credentials from function A
err = store.AddBasicAuth(ctx, namespace, httpConfig.BasicAuth)

// retrieving the credentials from function B
s := store.ForNamespace(namespace)
username, err := s.GetSecretKey(basicAuth.Username)
password, err := s.GetSecretKey(basicAuth.Password)
```

The main simplification is that function B doesn't need to know how function A built the key value. It makes also testing more decoupled and reduces the risk of leaking data across namespaces.

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification

Please check the [Prometheus-Operator testing guidelines](../TESTING.md) for recommendations about automated tests.

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #6641, opened 2024-06-03
chore: refactor tokens management in the assets package

simonpasquier:
## Description

This is a follow-up of #6537 and #6557.

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/assets/store.go
+++ b/pkg/assets/store.go
@@ -44,7 +44,6 @@ type StoreBuilder struct {
 	objStore cache.Store
 
 	TLSAssets        map[TLSAssetKey]TLSAsset
-	TokenAssets      map[string]Token
 	SigV4Assets      map[string]SigV4Credentials
 	AzureOAuthAssets map[string]AzureOAuthCredentials
 }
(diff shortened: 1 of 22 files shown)


pkg/assets/store.go
```diff
@@ -251,14 +221,16 @@ func (s *StoreBuilder) AddSafeAuthorizationCredentials(ctx context.Context, name
 		return err
 	}
 
-	err := s.addToken(ctx, namespace, auth.Credentials, key)
-	if err != nil {
-		return fmt.Errorf("failed to get authorization token of type %q: %w", auth.Type, err)
+	if auth.Credentials.Name != "" {
```
simonpasquier, line 224: I've kept this check to preserve the behavior of the defunct `addToken()` which exited early if the name value was empty. Eventually we may want to reject secret/configmap selectors with empty name & key but it's another discussion to have.

ArthurSens (approved):

pkg/assets/store.go
simonpasquier: It doesn't make sense and it might be prevented for some core resources. Our CRDs don't validate though. It would be a nice improvement, at least for v1alpha1 resources like ScrapeConfig (poke @mviswanathsai).

PR #6652, opened 2024-06-06
Support makeDaemonSet() for Prometheus Agent's DaemonSet deployment

haanhvu:
## Description

Support `makeDaemonSet()` and inner function `makeDaemonSetSpec()` for Prometheus Agent's DaemonSet deployment

## Type of change

## Verification
Unit tests

## Changelog entry

Support makeDaemonSet() for Prometheus Agent's DaemonSet deployment

### DIFF
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -119,32 +119,23 @@ func makeDaemonSetSpec(
 
 	var configReloaderWebConfigFile string
 
-	// Mount web config and web TLS credentials as volumes.
-	// We always mount the web config file for versions greater than 2.24.0.
-	// With this we avoid redeploying prometheus when reconfiguring between
-	// HTTP and HTTPS and vice-versa.
-	webConfigGenerator := cg.WithMinimumVersion("2.24.0")
-	if webConfigGenerator.IsCompatible() {
-		confArg, configVol, configMount, err := prompkg.BuildWebconfig(cpf, p)
-		if err != nil {
-			return nil, err
-		}
-
-		promArgs = append(promArgs, confArg)
-		volumes = append(volumes, configVol...)
-		promVolumeMounts = append(promVolumeMounts, configMount...)
-
-		// To avoid breaking users deploying an old version of the config-reloader image.
-		// TODO: remove the if condition after v0.72.0.
-		if cpf.Web != nil {
-			configReloaderWebConfigFile = confArg.Value
-			configReloaderVolumeMounts = append(configReloaderVolumeMounts, configMount...)
-		}
-	} else if cpf.Web != nil {
-		webConfigGenerator.Warn("web.config.file")
+	confArg, configVol, configMount, err := prompkg.BuildWebconfig(cpf, p)
+	if err != nil {
+		return nil, err
 	}
 
-	startupProbe, readinessProbe, livenessProbe := prompkg.MakeProbes(cpf, webConfigGenerator)
+	promArgs = append(promArgs, confArg)
+	volumes = append(volumes, configVol...)
+	promVolumeMounts = append(promVolumeMounts, configMount...)
+
+	// To avoid breaking users deploying an old version of the config-reloader image.
+	// TODO: remove the if condition after v0.72.0.
+	if cpf.Web != nil {
+		configReloaderWebConfigFile = confArg.Value
+		configReloaderVolumeMounts = append(configReloaderVolumeMounts, configMount...)
+	}
+
+	startupProbe, readinessProbe, livenessProbe := prompkg.MakeProbes(cpf, cg)
 
 	podAnnotations, podLabels := prompkg.BuildPodMetadata(cpf, cg)
 	// In cases where an existing selector label is modified, or a new one is added, new sts cannot match existing pods.
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -14,3 +14,303 @@
 
 package prometheusagent
 
+import (
+	"fmt"
+	"strings"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
+	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
+	prompkg "github.com/prometheus-operator/prometheus-operator/pkg/prometheus"
+	"github.com/prometheus-operator/prometheus-operator/pkg/webconfig"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+func makeDaemonSet(
+	name string,
+	p monitoringv1.PrometheusInterface,
+	config *prompkg.Config,
+	cg *prompkg.ConfigGenerator,
+	tlsSecrets *operator.ShardedSecret,
+) (*appsv1.DaemonSet, error) {
+	cpf := p.GetCommonPrometheusFields()
+	objMeta := p.GetObjectMeta()
+
+	if cpf.PortName == "" {
+		cpf.PortName = prompkg.DefaultPortName
+	}
+
+	// We need to re-set the common fields because cpf is only a copy of the original object.
+	// We set some defaults if some fields are not present, and we want those fields set in the original Prometheus object before building the StatefulSetSpec.
+	p.SetCommonPrometheusFields(cpf)
+	spec, err := makeDaemonSetSpec(p, config, cg, tlsSecrets)
+	if err != nil {
+		return nil, fmt.Errorf("make DaemonSet spec: %w", err)
+	}
+
+	// do not transfer kubectl annotations to the statefulset so it is not
+	// pruned by kubectl
+	annotations := make(map[string]string, 0)
+	for key, value := range objMeta.GetAnnotations() {
+		if !strings.HasPrefix(key, "kubectl.kubernetes.io/") {
+			annotations[key] = value
+		}
+	}
+	daemonSet := &appsv1.DaemonSet{Spec: *spec}
+
+	operator.UpdateObject(
+		daemonSet,
+		operator.WithName(name),
+		operator.WithAnnotations(annotations),
+		operator.WithAnnotations(config.Annotations),
+		operator.WithLabels(objMeta.GetLabels()),
+		operator.WithLabels(map[string]string{
+			prompkg.PrometheusNameLabelName: objMeta.GetName(),
+			prompkg.PrometheusModeLabeLName: prometheusMode,
+		}),
+		operator.WithLabels(config.Labels),
+		operator.WithManagingOwner(p),
+	)
+
+	if cpf.ImagePullSecrets != nil && len(cpf.ImagePullSecrets) > 0 {
+		daemonSet.Spec.Template.Spec.ImagePullSecrets = cpf.ImagePullSecrets
+	}
+
+	if cpf.HostNetwork {
+		daemonSet.Spec.Template.Spec.DNSPolicy = v1.DNSClusterFirstWithHostNet
+	}
+
+	return daemonSet, nil
+}
+
+func makeDaemonSetSpec(
+	p monitoringv1.PrometheusInterface,
+	c *prompkg.Config,
+	cg *prompkg.ConfigGenerator,
+	tlsSecrets *operator.ShardedSecret,
+) (*appsv1.DaemonSetSpec, error) {
+	cpf := p.GetCommonPrometheusFields()
+
+	pImagePath, err := operator.BuildImagePath(
+		operator.StringPtrValOrDefault(cpf.Image, ""),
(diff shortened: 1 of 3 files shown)


pkg/prometheus/agent/daemonset_test.go
```diff
@@ -0,0 +1,384 @@
...
+
+func TestAutomountServiceAccountTokenForDaemonSet(t *testing.T) {
+	for _, tc := range []struct {
+		name                         string
+		automountServiceAccountToken *bool
+		expectedValue                bool
+	}{
+		{
+			name:                         "automountServiceAccountToken not set",
+			automountServiceAccountToken: nil,
```
ArthurSens, line 350:
```suggestion
```

The "default" value of a pointer is `nil`, so there's no need to add this line :)

ArthurSens (commented):
Sorry, couldn't find the time to review everything just yet. I've made some comments about test readability

I also have the impression that a lot of code is copy pasted from makeStatefulSet and makeStatefulSetSpec. Not a requirement for this PR, but I'm just wondering if we could reduce code duplication between daemonset and statefulsets

haanhvu:
> I also have the impression that a lot of code is copy pasted from makeStatefulSet and makeStatefulSetSpec. Not a requirement for this PR, but I'm just wondering if we could reduce code duplication between daemonset and statefulsets

@ArthurSens Yes we can. But as we discussed in previous meetings, I think we should do the refactoring later. Because the implementation for daemonset could change, e.g., if later we find a bug in the implementation through e2e tests. You agree?

Most of your change suggestions are for both daemonset and statefulset's implementation. I'll resolve for both modes then.

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,319 @@
...
+	cpf := p.GetCommonPrometheusFields()
+	objMeta := p.GetObjectMeta()
+
+	if cpf.PortName == "" {
+		cpf.PortName = prompkg.DefaultPortName
+	}
+
+	// We need to re-set the common fields because cpf is only a copy of the original object.
+	// We set some defaults if some fields are not present, and we want those fields set in the original Prometheus object before building the DaemonSetSpec.
+	p.SetCommonPrometheusFields(cpf)
```
simonpasquier, line 49: should we factorize since we have similar in Prometheus sts and PrometheusAgent sts?

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,319 @@
...
+	spec, err := makeDaemonSetSpec(p, config, cg, tlsSecrets)
+	if err != nil {
+		return nil, fmt.Errorf("make DaemonSet spec: %w", err)
+	}
+
+	// do not transfer kubectl annotations to the daemonset so it is not
+	// pruned by kubectl
```
simonpasquier, line 56: Same question here regarding this part. We could create an `operator.DropKubeCtlAnnotations()` function to reduce duplication?

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,319 @@
...
+		operator.WithLabels(objMeta.GetLabels()),
+		operator.WithLabels(map[string]string{
+			prompkg.PrometheusNameLabelName: objMeta.GetName(),
+			prompkg.PrometheusModeLabeLName: prometheusMode,
+		}),
+		operator.WithLabels(config.Labels),
+		operator.WithManagingOwner(p),
+	)
+
+	if cpf.ImagePullSecrets != nil && len(cpf.ImagePullSecrets) > 0 {
```
simonpasquier, line 79: same remark here regarding duplication

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,319 @@
...
+		daemonSet.Spec.Template.Spec.ImagePullSecrets = cpf.ImagePullSecrets
+	}
+
+	if cpf.HostNetwork {
```
simonpasquier, line 83: and here again

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,319 @@
...
+			MountPath: prompkg.ConfOutDir,
+		},
+	}
+
+	var configReloaderWebConfigFile string
+
+	// Mount web config and web TLS credentials as volumes.
+	// We always mount the web config file for versions greater than 2.24.0.
+	// With this we avoid redeploying prometheus when reconfiguring between
+	// HTTP and HTTPS and vice-versa.
```
simonpasquier, line 145: I'd simplify and just consider that users must use a recent Prometheus version for DaemonSet deployment.

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,319 @@
...
+		webConfigGenerator.Warn("web.config.file")
+	}
+
+	// The /-/ready handler returns OK only after the TSDB initialization has
+	// completed. The WAL replay can take a significant time for large setups
+	// hence we enable the startup probe with a generous failure threshold (15
+	// minutes) to ensure that the readiness probe only comes into effect once
+	// Prometheus is effectively ready.
+	// We don't want to use the /-/healthy handler here because it returns OK as
+	// soon as the web server is started (irrespective of the WAL replay).
```
simonpasquier, line 182: We can either use the same settings as with statefulset mode or consider that since daemonset doesn't support persistent storage, we can have tighter timeouts.

haanhvu:
We discussed offline and agreed to do some refactoring of the common between StatefulSet and DaemonSet first, before continuing this PR.

pkg/prometheus/agent/daemonset.go
haanhvu: Yeah, it's a new mode so we don't need this check

pkg/prometheus/agent/daemonset.go
haanhvu: What lines are you referring to? Line 40-49 do different things. I'm not sure we should refactor them into one function.

pkg/prometheus/agent/daemonset.go
haanhvu: Currently daemonet doesn't have `prometheus-operator-input-hash` annotation, as we agreed in Slack: https://cloud-native.slack.com/archives/C071STDT4FP/p1717580605023279 So this is not a duplication actually

pkg/prometheus/agent/daemonset.go
haanhvu: From here downward we set the fields in the DaemonSet object. It's a different object from StatefulSet. So I don't think it's a duplication here?

pkg/prometheus/agent/daemonset.go
haanhvu: same as https://github.com/prometheus-operator/prometheus-operator/pull/6652#discussion_r1643946300

haanhvu:
I just rebased, haven't done anything new yet. Will ping when this's ready for review.

pkg/prometheus/agent/daemonset.go
haanhvu: Now I'm not entirely sure yet we should remove this check in DaemonSet. If we remove this check, it means we will officially require users to use >=2.24 right? If that's the case, we can remove this check and should clearly write the requirement in the doc.

haanhvu:
@ArthurSens @simonpasquier @kakkoyun This is ready for review now. Please help review so that we can move on to the next thing. Next thing will be operator's logic and e2e tests, which are more difficult and interesting :smile:

ArthurSens (approved):
LGTM!

If we find something wrong along the way, we can just fix it in a follow-up PR :)

pkg/prometheus/agent/daemonset.go
simonpasquier: Let me try to clarify: I would declare that Daemonset mode requires a "recent" version of Prometheus (for instance at least 2.45.0 which the current long-term support version) so we can simplify the support matrix. The check should be done as early as possible so makeDaemonSet() can safely assume a minimal version.

haanhvu: Yeah I wanted to clarify this, because we had a discussion in Slack (with @ArthurSens too) about whether we should limit the Prometheus version in DaemonSet: https://cloud-native.slack.com/archives/C071STDT4FP/p1718703164200639

If we agree to require a specific recent version, I'll remove this check then. I'll also note so that we'll remember to clarify this is the doc of DaemonSet.

pkg/prometheus/agent/daemonset.go
haanhvu: Eh, now I realized the objects are different, but the fields are similar. I'll refactor then.

pkg/prometheus/agent/daemonset.go
haanhvu: same as https://github.com/prometheus-operator/prometheus-operator/pull/6652/files#r1653864712 again

pkg/prometheus/agent/daemonset.go
haanhvu: Now rethought I'm not sure I need to refactor, since this duplication is pretty short and simple code :thinking: Do you have any hard opinion?

pkg/prometheus/agent/daemonset.go
haanhvu: same as https://github.com/prometheus-operator/prometheus-operator/pull/6652#discussion_r1653880084 again

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,252 @@
...
+	promArgs = append(promArgs, confArg)
+	volumes = append(volumes, configVol...)
+	promVolumeMounts = append(promVolumeMounts, configMount...)
+
+	// To avoid breaking users deploying an old version of the config-reloader image.
+	// TODO: remove the if condition after v0.72.0.
+	if cpf.Web != nil {
+		configReloaderWebConfigFile = confArg.Value
+		configReloaderVolumeMounts = append(configReloaderVolumeMounts, configMount...)
+	}
```
haanhvu, line 136: @ArthurSens @simonpasquier @kakkoyun We already agreed on Prometheus version in https://github.com/prometheus-operator/prometheus-operator/pull/6652#discussion_r1653854976 What about config reloader version here? Do we need this version check?

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,243 @@
...
+		operator.WithName(name),
+		operator.WithAnnotations(objMeta.GetAnnotations()),
+		operator.WithAnnotations(config.Annotations),
+		operator.WithLabels(objMeta.GetLabels()),
+		operator.WithLabels(map[string]string{
+			prompkg.PrometheusNameLabelName: objMeta.GetName(),
+			prompkg.PrometheusModeLabeLName: prometheusMode,
+		}),
+		operator.WithLabels(config.Labels),
+		operator.WithManagingOwner(p),
```
simonpasquier, line 67: ;)
```suggestion
		operator.WithManagingOwner(p),
		operator.WithoutKubectlAnnotations(),
```

pkg/prometheus/agent/daemonset.go
simonpasquier: fair point!

pkg/prometheus/agent/daemonset.go
simonpasquier: agreed!

pkg/prometheus/agent/daemonset.go
simonpasquier: No, let's keep it simple.

pkg/prometheus/agent/daemonset.go
haanhvu: Not directly related to this change, but do you think we need an unit test for annotation and label? Something like https://github.com/prometheus-operator/prometheus-operator/blob/680654c9d110ab90c214f3185ca85926f1c72f6f/pkg/prometheus/server/statefulset_test.go#L79 but doesn't have `prometheus-operator-input-hash` and `operator.prometheus.io/shard`.

I think this is needed because this is the part where daemonset is different from statefulset (both agent and server).

simonpasquier: Yes a unit test would be good.

pkg/prometheus/agent/daemonset.go
```diff
@@ -0,0 +1,240 @@
...
+	configReloaderWebConfigFile = confArg.Value
+	configReloaderVolumeMounts = append(configReloaderVolumeMounts, configMount...)
+
+	startupProbe, readinessProbe, livenessProbe := prompkg.MakeProbes(cpf, cg)
+
+	podAnnotations, podLabels := prompkg.BuildPodMetadata(cpf, cg)
+	// In cases where an existing selector label is modified, or a new one is added, new sts cannot match existing pods.
+	// We should try to avoid removing such immutable fields whenever possible since doing
+	// so forces us to enter the 'recreate cycle' and can potentially lead to downtime.
+	// The requirement to make a change here should be carefully evaluated.
```
simonpasquier, line 132: (nit)
```suggestion
	// In cases where an existing selector label is modified, or a new one is added, new daemonset cannot match existing pods.
	// We should try to avoid removing such immutable fields whenever possible since doing
	// so forces us to enter the 'recreate cycle' and can potentially lead to downtime.
	// The requirement to make a change here should be carefully evaluated.
```

simonpasquier (approved):
One minor comment which you can address later.

PR #6633, opened 2024-05-29
ScrapeConfig CRD: Add DockerswarmSD configuration to the CRD

mviswanathsai:
## Description
This PR adds the DockerswarmSD config options to the ScrapeConfig CRD.

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [x] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification

Please check the [Prometheus-Operator testing guidelines](../TESTING.md) for recommendations about automated tests.

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```
Add Dockerswarm SD config to ScrapeConfig CRD
```

PR #6661, opened 2024-06-09
ScrapeConfig CRD: refactor `Filter` types

mviswanathsai:
## Description

Earlier, there was a little bit of code duplication concerning the `Filter` types for EC2, DockerSwarm and Docker Service discoveries. 
This PR: 
1. Removes that to add a singular `Filter` type and 
2. Adds a `addFilterToYaml()` function which will add Filters to the generated config while also checking for appropriate Prometheus version which was missing earlier.

## Type of change

## Verification

## Changelog entry

```release-note

```


pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -536,7 +524,7 @@ type EC2SDConfig struct {
 	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
 	// Filter API documentation: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Filter.html
 	// +optional
-	Filters []*EC2Filter `json:"filters"`
+	Filters []Filter `json:"filters"`
```
slashpai, line 527: This should be still a pointer as its a optional field

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -830,7 +818,7 @@ type DockerSDConfig struct {
 	HostNetworkingHost *string `json:"hostNetworkingHost,omitempty"`
 	// Optional filters to limit the discovery process to a subset of the available resources.
 	// +optional
-	Filters *[]DockerFilter `json:"filters,omitempty"`
+	Filters []Filter `json:"filters,omitempty"`
```
slashpai, line 821: same here should be a pointer

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -958,7 +946,7 @@ type DockerSwarmSDConfig struct {
 	// Tasks: https://docs.docker.com/engine/api/v1.40/#operation/TaskList
 	// Nodes: https://docs.docker.com/engine/api/v1.40/#operation/NodeList
 	// +optional
-	Filters []DockerSwarmFilter `json:"filters"`
+	Filters []Filter `json:"filters"`
```
slashpai, line 949: here as well field should be pointer

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
mviswanathsai: Regarding the `[]RelabelConfig` field, since it is an array, I believe we can determine if the field was entered simply by checking its length. This approach should be sufficient, as even with pointers, we wouldn't be checking each element for nil.

PS: The change was originally suggested [here](https://github.com/prometheus-operator/prometheus-operator/pull/6633#discussion_r1620243723).

mviswanathsai: I now see that `omitempty` is not present here. Should I add that?

simonpasquier: yes please add the omitempty tag.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required says:

````
Optional fields have the following properties:
[...]
* [...] have a built-in nil value (e.g. maps and slices).
````

File: pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go [hunk 4]
```diff
@@ -56,20 +56,8 @@ type AttachMetadata struct {
...
-	Values []string `json:"values"`
-}
-
-// DockerFilter is the configuration to limit the discovery process to a subset of available resources.
-type DockerFilter struct {
-	Name   string   `json:"name"`
-	Values []string `json:"values"`
-}
-
 // Filter is the configuration to limit the discovery process to a subset of available resources.
```
simonpasquier, line 59:
```suggestion
// Filter name and value pairs to limit the discovery process to a subset of available resources.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -56,20 +56,8 @@ type AttachMetadata struct {
...
-type DockerSwarmFilter struct {
+type Filter struct {
 	// Name is the key of the field to check against.
```
simonpasquier, line 61: Since the struct is used for different SDs, let's have a generic description. 
```suggestion
	// Name of the filter.
```

And for the Values field

````
// Values to filter on.
````

File: pkg/prometheus/promcfg.go [hunk 6]
```diff
@@ -4092,3 +4066,11 @@ func getLowerByteSize(v *monitoringv1.ByteSize, cpf *monitoringv1.CommonPromethe
 func isByteSizeEmpty(v *monitoringv1.ByteSize) bool {
 	return v == nil || *v == ""
 }
+
+func (cg *ConfigGenerator) addFiltersToYaml(cfg yaml.MapSlice, minimumVersion string, filters []monitoringv1alpha1.Filter) yaml.MapSlice {
+	if filters == nil {
```
simonpasquier, line 4071:
```suggestion
	if len(filters) == 0 {
```

pkg/prometheus/promcfg.go
```diff
@@ -4092,3 +4066,11 @@ func getLowerByteSize(v *monitoringv1.ByteSize, cpf *monitoringv1.CommonPromethe
...
+		return cfg
+	}
+
+	return cg.WithMinimumVersion(minimumVersion).AppendMapItem(cfg, "filters", filters)
```
simonpasquier, line 4075: (suggestion) for explicitness, I'd prefer to avoid relying on the CRD struct == the Prometheus struct. We also need to ensure that the generation is deterministic (see stringMapToMapSlice for example).

mviswanathsai: I originally had the logic be explicit, but then changed it. I will change it back to how it was.

pkg/prometheus/promcfg.go
mviswanathsai: Same here, I changed it after. I will change it back.

pkg/prometheus/promcfg.go
```diff
@@ -2998,12 +2998,10 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		configs := make([][]yaml.MapItem, len(sc.Spec.EC2SDConfigs))
 		for i, config := range sc.Spec.EC2SDConfigs {
 			if config.Region != nil {
-				configs[i] = []yaml.MapItem{
-					{
-						Key:   "region",
-						Value: config.Region,
-					},
-				}
+				configs[i] = append(configs[i], yaml.MapItem{
```
simonpasquier, line 3001: hmm nice catch! we had a latent bug here?

pkg/prometheus/promcfg.go
```diff
@@ -3056,12 +3054,7 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 				})
 			}
 
-			if config.Filters != nil {
-				configs[i] = append(configs[i], yaml.MapItem{
-					Key:   "filters",
-					Value: config.Filters,
-				})
-			}
+			configs[i] = cg.addFiltersToYaml(configs[i], "2.3.0", config.Filters)
```
simonpasquier, line 3057: (suggestion) a bit more verbose but more readable from my POV (you don't have to ask yourself what this "2.3.0" string is about). I bet that it's also simpler for Docker SDs since you don't have to pass the version again.
```suggestion
			configs[i] = cg.WithMinimumVersion("2.3.0").addFiltersToYaml(configs[i], config.Filters)
```

pkg/prometheus/promcfg.go
mviswanathsai: Yeah, I noticed that the Filters were not getting added, turns out this is the reason.

pkg/prometheus/promcfg.go
mviswanathsai: > I bet that it's also simpler for Docker SDs since you don't have to pass the version again.
Not sure I understand this.

pkg/prometheus/promcfg.go
```diff
@@ -3512,6 +3505,7 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 			configs[i] = cg.addOAuth2ToYaml(configs[i], s, config.OAuth2)
 			configs[i] = cg.addProxyConfigtoYaml(ctx, configs[i], sc.GetNamespace(), store, config.ProxyConfig)
 			configs[i] = cg.addBasicAuthToYaml(configs[i], s, config.BasicAuth)
+			configs[i] = cg.WithMinimumVersion("2.23.0").addFiltersToYaml(configs[i], config.Filters)
```
simonpasquier, line 3508: In fact, Docker SD already requires 2.23.0 (though it's not enforced in `validateDockerSDConfigs` and should be fixed in another PR).
```suggestion
			configs[i] = cg.addFiltersToYaml(configs[i], config.Filters)
```

pkg/prometheus/promcfg.go
```diff
@@ -3787,6 +3759,7 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 			configs[i] = cg.addSafeAuthorizationToYaml(configs[i], s, config.Authorization)
 			configs[i] = cg.addOAuth2ToYaml(configs[i], s, config.OAuth2)
 			configs[i] = cg.addProxyConfigtoYaml(ctx, configs[i], sc.GetNamespace(), store, config.ProxyConfig)
+			configs[i] = cg.WithMinimumVersion("2.23.0").addFiltersToYaml(configs[i], config.Filters)
```
simonpasquier, line 3762: similar remark here though in this case, `validateDockerSwarmSDConfigs()` already ensures that the SD is supported by the Prometheus version.

pkg/prometheus/promcfg.go
```diff
@@ -4092,3 +4044,24 @@ func getLowerByteSize(v *monitoringv1.ByteSize, cpf *monitoringv1.CommonPromethe
...
 	return v == nil || *v == ""
 }
+
+func (cg *ConfigGenerator) addFiltersToYaml(cfg yaml.MapSlice, filters []monitoringv1alpha1.Filter) yaml.MapSlice {
+	if len(filters) == 0 {
+		return cfg
+	}
+
+	filtersYamlMap := []yaml.MapSlice{}
+	for _, filter := range filters {
```
simonpasquier, line 4054: to have a generated config which is 100% deterministic and stable, I'd sort the filters slice by name and then iterate over the sorted names.

pkg/prometheus/promcfg.go
mviswanathsai: Got it. I will remove the version check here and then make a PR for the Docker refactor and handle it there.

pkg/prometheus/promcfg.go
mviswanathsai: I thought of this too. Should we sort all the fields for all the generated SDs as well?

pkg/prometheus/promcfg.go
```diff
@@ -4092,3 +4044,29 @@ func getLowerByteSize(v *monitoringv1.ByteSize, cpf *monitoringv1.CommonPromethe
...
+		filtersYamlMap = append(filtersYamlMap, yaml.MapSlice{
+			{
+				Key:   "name",
+				Value: filter.Name,
+			},
+			{
+				Key:   "values",
+				Value: filter.Values,
+			}})
+	}
```
simonpasquier, line 4069: let's avoid the extra step
```suggestion
	filtersYamlMap := []yaml.MapSlice{}
	for _, filter := range filters {
		filtersYamlMap = append(filtersYamlMap, yaml.MapSlice{
			{
				Key:   "name",
				Value: filter.Name,
			},
			{
				Key:   "values",
				Value: filter.Values,
			}})
	}
```

simonpasquier: AFAICT it's not required since it happens when serializing to YAML anyway

mviswanathsai: I removed the sorting logic and generated the config and `a_dummy_label` comes after `dummy_label`. 
Seems like the filters are not sorted, or do you mean something else?

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -536,7 +524,7 @@ type EC2SDConfig struct {
 	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
 	// Filter API documentation: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Filter.html
 	// +optional
-	Filters []*EC2Filter `json:"filters"`
+	Filters []Filter `json:"filters,omitempty"`
```
simonpasquier, line 527: It's not enforced by Prometheus (e.g. you can define several times the same fliter name in the list) but IMHO it makes sense to add some restriction at the API level here.
```suggestion
	// +listType:=map
	// +listMapKey:=Name
	Filters []Filter `json:"filters,omitempty"`
```

pkg/prometheus/promcfg.go
simonpasquier: My bad! Without the explicit sorting, the keys should always be listed as they are in the custom resource. While it's unlikely that the order changes repeatedly in the spec, it's always better to ensure predictability of the generated configuration.

suggestion: use https://pkg.go.dev/slices#SortStableFunc and https://pkg.go.dev/cmp#Compare

pkg/prometheus/promcfg_test.go
```diff
@@ -7231,11 +7231,13 @@ func TestScrapeConfigSpecConfigWithDockerSDConfig(t *testing.T) {
...
 						RefreshInterval:    ptr.To(monitoringv1.Duration("30s")),
 						HostNetworkingHost: ptr.To("localhost"),
-						Filters: &[]monitoringv1alpha1.DockerFilter{
+						Filters: []monitoringv1alpha1.Filter{
 							{Name: "dummy_label_1",
 								Values: []string{"dummy_value_1"}},
 							{Name: "dummy_label_2",
 								Values: []string{"dummy_value_2", "dummy_value_3"}},
+							{Name: "a_dummy_label_1",
+								Values: []string{"dummy_value_2", "dummy_value_3"}},
```
ArthurSens, line 7240: I see that you're adding this "alphabetic out-of-order" filter here with the goal to test for determinism, right?

Should we run this test like 50x times, just to guarantee the determinism? Or running just once is enough?

mviswanathsai: 🤔 I didn't think of it that way tbh. But since the sorting algo is from the standard lib, we can rest easy? I'm not sure tbh. cc: @simonpasquier

simonpasquier: running once is enough.

ArthurSens: Sounds good to me :)

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
mviswanathsai: Seems like `listMapKey` can only be applied for associative lists 🤔

mviswanathsai: Getting this error: `must apply listMapKey to an associative-list`

simonpasquier:
Another rebase needed :)

mviswanathsai:
Any idea why the unit tests are failing?

simonpasquier:
@mviswanathsai the change breaks `TestScrapeConfigSpecConfigWithDockerSDConfig` 
https://github.com/prometheus-operator/prometheus-operator/actions/runs/9692322799/job/26745433708?pr=6661#step:5:1982

PR #6665, opened 2024-06-11
chore: Add feature-gated tests to CI

ArthurSens:
## Description

The goal is that this test suite will be used for feature-gated features that are incoming, e.g. #6198 and #6600 

## Type of change

## Verification

## Changelog entry

```release-note
NONE
```


test/e2e/main_test.go
```diff
@@ -420,6 +416,20 @@ const (
 	prometheusOperatorServiceName = "prometheus-operator"
 )
 
+// TestFeatureGatedFeatures tests features that are behind feature gates.
+// Altough we'd like to have these tests green at all times, they are not
+// critical for the release process, therefore they can fail without blocking PRs.
+func TestFeatureGatedFeatures(t *testing.T) {
```
simonpasquier, line 422: (nit)
```suggestion
func TestGatedFeatures(t *testing.T) {
```

.github/workflows/e2e.yaml
```diff
@@ -99,7 +94,7 @@ jobs:
         EXCLUDE_PROMETHEUS_ALL_NS_TESTS=${{ matrix.prometheusAllNS }}
         EXCLUDE_THANOSRULER_TESTS=${{ matrix.thanosruler }}
         EXCLUDE_OPERATOR_UPGRADE_TESTS=${{ matrix.operatorUpgrade }}
-        FEATURE_GATED_TESTS=${{ matrix.featureGated }}
+        FEATURE_GATED_TESTS=exclude
```
simonpasquier, line 97: should we rename to EXCLUDE_FEATURE_GATED_TESTS for consistency?

test/e2e/main_test.go
```diff
@@ -420,6 +416,20 @@ const (
 	prometheusOperatorServiceName = "prometheus-operator"
 )
 
+// TestFeatureGatedFeatures tests features that are behind feature gates.
+// Altough we'd like to have these tests green at all times, they are not
+// critical for the release process, therefore they can fail without blocking PRs.
```
simonpasquier, line 421: I might not agree 100% with this :)

For now, all feature gates are alpha so yes they could break anytime but in the future, some of them (hopefully) will become beta. Past experience also showed me that non-gating jobs are very likely to degrade over time.

simonpasquier (commented):
Thanks for starting this!

.github/workflows/e2e.yaml
ArthurSens: yeah good idea

test/e2e/main_test.go
ArthurSens: Fair enough, I did not think of beta graduation of features.

Should we split tests again in the future? (Beta are required to pass, Alpha not required)

simonpasquier: We can still merge without all required tests being green so I'd rather be strict from the start and avoid considering these tests as "less important".

ArthurSens: Fair point, changes adjusted

simonpasquier (approved):
This is good enough to get started!

PR #6663, opened 2024-06-10
Add --use-endpoint-slices flag to toggle EndpointSlice vs Endpoints service discovery

thaniri:
## Description

This is a solution to https://github.com/prometheus-operator/prometheus-operator/issues/3862

This introduces a command line flag to toggle service discovery via [EndpointSlices](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/). This will necessarily require a patch to [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) that looks something like this:

<details>
  <summary>kube-prometheus-stack patch</summary>
  
  Shows what changes were required to the kube-prometheus-stack helm chart to configure this command line flag.
  
  ```
  diff --git a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
  index 342f3882..3ceaf0fe 100644
  --- a/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
  +++ b/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml
  @@ -132,6 +132,9 @@ spec:
               {{- if .Values.prometheusOperator.secretFieldSelector }}
               - --secret-field-selector={{ tpl (.Values.prometheusOperator.secretFieldSelector) $ }}
               {{- end }}
  +            {{- if .Values.prometheusOperator.useEndpointSlices }}
  +            - --use-endpoint-slices=true
  +            {{- end}}
               {{- if .Values.prometheusOperator.clusterDomain }}
               - --cluster-domain={{ .Values.prometheusOperator.clusterDomain }}
               {{- end }}
  diff --git a/charts/kube-prometheus-stack/values.yaml b/charts/kube-prometheus-stack/values.yaml
  index 5dd92fbd..3065598c 100644
  --- a/charts/kube-prometheus-stack/values.yaml
  +++ b/charts/kube-prometheus-stack/values.yaml
  @@ -2188,6 +2188,7 @@ prometheus-node-exporter:
   ##
   prometheusOperator:
     enabled: true
  +  useEndpointSlices: false
 
     ## Use '{{ template "kube-prometheus-stack.fullname" . }}-operator' by default
     fullnameOverride: ""
  ```
</details>

[...]

PR #6672, opened 2024-06-12
Add ServiceDiscoveryRole configuration to toggle service discovery via EndpointSlice vs Endpoints

thaniri:
## Description

This is a solution to https://github.com/prometheus-operator/prometheus-operator/issues/3862 and is a replacement for the proposed solution https://github.com/prometheus-operator/prometheus-operator/pull/6663. This builds from the work @haanhvu started here https://github.com/prometheus-operator/prometheus-operator/pull/6518.

This PR introduces a `serviceDiscoveryRole` field to the specs for `Prometheus` and `PrometheusAgent` via the `CommonPrometheusFields` configuration pattern.

This will necessarily require a patch to [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) that looks something like this:

<details>
  <summary>kube-prometheus-stack patch</summary>
  
  Shows what changes were required to the kube-prometheus-stack helm chart to make use of the `serviceDiscoverRole` configuration.
  
  ```
  diff --git a/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheusagents.yaml b/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheusagents.yaml
  index caa24793..e332c5d4 100644
  --- a/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheusagents.yaml
  +++ b/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheusagents.yaml
  @@ -5535,6 +5535,15 @@ spec:
                     Only valid in Prometheus versions 2.45.0 and newer.
                   format: int64
                   type: integer
  +              serviceDiscoveryRole:
  +                description: |-
  +                  Defines the service discovery mechanism used to discover scrape targets.
  +                  If set, the value should be either "Endpoints" or "EndpointSlice".
  +                default: Endpoints
  +                enum:
  +                - Endpoints
  +                - EndpointSlice
  +                type: string
                 scrapeClasses:
                   description: |-
                     List of scrape classes to expose to scraping objects such as
  diff --git a/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheuses.yaml b/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheuses.yaml
  index 2dc5ba9c..c511e988 100644
  --- a/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheuses.yaml
  +++ b/charts/kube-prometheus-stack/charts/crds/crds/crd-prometheuses.yaml
  @@ -6839,6 +6839,15 @@ spec:
                     Only valid in Prometheus versions 2.45.0 and newer.
                   format: int64
                   type: integer
  +              serviceDiscoveryRole:
  +                description: |-
  +                  Defines the service discovery mechanism used to discover scrape targets.
  +                  If set, the value should be either "Endpoints" or "EndpointSlice".
  +                default: Endpoints
  +                enum:
  +                - Endpoints
  +                - EndpointSlice
  +                type: string
                 scrapeClasses:
                   description: |-
                     List of scrape classes to expose to scraping objects such as
  diff --git a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
  index ae95fd02..400dce9f 100644
  --- a/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
  +++ b/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml
  @@ -185,6 +185,9 @@ spec:
   {{ else }}
     podMonitorNamespaceSelector: {}
   {{- end }}
  +{{- if .Values.prometheus.prometheusSpec.serviceDiscoveryRole }}
  +  serviceDiscoveryRole: {{ .Values.prometheus.prometheusSpec.serviceDiscoveryRole }}
  +{{- end }}
   {{- if .Values.prometheus.prometheusSpec.probeSelector }}
     probeSelector:
   {{ tpl (toYaml .Values.prometheus.prometheusSpec.probeSelector | indent 4) . }}
  diff --git a/charts/kube-prometheus-stack/values.yaml b/charts/kube-prometheus-stack/values.yaml
  index 5dd92fbd..75c03cdf 100644
  --- a/charts/kube-prometheus-stack/values.yaml
  +++ b/charts/kube-prometheus-stack/values.yaml
  @@ -3365,6 +3365,7 @@ prometheus:
     ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#prometheusspec
     ##
     prometheusSpec:
  +    serviceDiscoveryRole: Endpoints
       ## Statefulset's persistent volume claim retention policy
       ## pvcDeleteOnStsDelete and pvcDeleteOnStsScale determine whether
       ## statefulset's PVCs are deleted (true) or retained (false) on scaling down
  ```
</details>

## Type of change

## Verification

This can be tested by deploying the `kube-prometheus-stack` helm chart with the above patch. When `serviceDiscoveryRole` is set to `Endpoints` then the generated scrape configs will have a `kubernetes_sd_role` of `endpoints`. Otherwise if `serviceDiscoveryRole` is set to `EndpointSlice` then the generated scrape configs will have a `kubernetes_sd_role` of `endpointslices`. If `serviceDiscoveryRole` is unset, or is set to an incorrect value, then the generated scrape configs will have a `kubernetes_sd_role` of `endpoints`.

## Changelog entry

```release-note
[ENHANCEMENT] Add ServiceDiscoveryRole configuration for use of Endpoints vs EndpointSlice.
```


ArthurSens:
Just tagging @haanhvu here! Once you have the time, could you review this PR? :)

pkg/prometheus/promcfg_test.go
```diff
@@ -901,7 +911,7 @@ func TestK8SSDConfigGeneration(t *testing.T) {
 			sm.Namespace,
 			tc.apiServerConfig,
 			tc.store,
-			kubernetesSDRoleEndpoint,
+			tc.role,
```
haanhvu, line 914: This check is valid, so we can keep it. However, it doesn't actually test the change introduced in this PR.

Looking at the implementation in this PR, the change happens at the `sync()` function of Prometheus/PrometheusAgent operator. Basically, `sync()` checks the value of `ServiceDiscoveryRole`, then set `endpointSliceSupported` accordingly. Meanwhile, this check is actually calling `generateK8SSDConfig()`, which is not a change in this PR.

So to check the change of this PR, could you add a new test to verify that `sync()` sets `endpointSliceSupported` according to the value of `ServiceDiscoveryRole`? Thankfully we already had [TestServiceMonitorWithEndpointSliceEnable](https://github.com/prometheus-operator/prometheus-operator/blob/ad006f4d607c99062d66936a8c02dd8bbaa6b40e/pkg/prometheus/promcfg_test.go#L1535C6-L1535C47) that verifies EndpointSlice is correctly used if `endpointSliceSupported == true`, so we don't need to care about this later step.

haanhvu (commented):
Thanks for helping it move forward! I left a few comments. Feel free to ask for more information if you need.

pkg/prometheus/agent/operator.go
```diff
@@ -586,6 +583,22 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		level.Info(logger).Log("msg", "using endpointslice as service discovery role")
+		c.endpointSliceSupported = true
+	case "Endpoints":
+		level.Info(logger).Log("msg", "using endpoints as service discovery role")
+		c.endpointSliceSupported = false
+	default:
+		level.Info(logger).Log("msg",
+			"unknown service discovery role, defaulting to endpoints",
+			"serviceDiscoveryRole",
+			serviceDiscoveryRole)
```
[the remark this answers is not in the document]
thaniri: - I have updated the log level to Warn
- I have added additional information for the end user to know what to configure if that warning message appears
- Not technically required as Golang will default bools that are not explicitly given a value to false. But it is a good catch for the inconsistency and I think being explicit is a good practice so I have added this suggestion.

Please see 70d258855 for my changes for these comments.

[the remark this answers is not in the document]
haanhvu: > Not technically required as Golang will default bools that are not explicitly given a value to false. But it is a good catch for the inconsistency and I think being explicit is a good practice so I have added this suggestion.

Yes the default bool is false. But in this case `endpointSliceSupported` is set through 2 steps:

- Step 1: Check if the version supports EndpointSlice: https://github.com/prometheus-operator/prometheus-operator/blob/ad006f4d607c99062d66936a8c02dd8bbaa6b40e/pkg/prometheus/agent/operator.go#L320-L324
- Step 2: Check the value of `ServiceDiscoveryRole` (this PR's change)

So there may be cases when `endpointSliceSupported` is true in step 1, but false in step 2. E.g., if the version supports EndpointSlice but users still need to use Endpoints. So we need to update `endpointSliceSupported` in step 2 in any cases.

The reason why step 1 is not enough and we need step 2 is explained in https://github.com/prometheus-operator/prometheus-operator/issues/4634

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -701,6 +701,12 @@ type CommonPrometheusFields struct {
 	// +listType=map
 	// +listMapKey=name
 	ScrapeClasses []ScrapeClass `json:"scrapeClasses,omitempty"`
+
+	// Defines the service discovery mechanism used to discover scrape targets.
+	// If set, the value should be either "Endpoints" or "EndpointSlice".
+	// +optional
+	// +kubebuilder:validation:Enum=Endpoints;EndpointSlice
+	ServiceDiscoveryRole string `json:"serviceDiscoveryRole,omitempty"`
```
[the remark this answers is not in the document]
thaniri: I've seen it used as a pattern for optional elements for JSON objects to be represented as pointers to values in Golang. As this is an easy change and makes sense, I have implemented this suggestion in [2ddd628](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/2ddd628152e07af43f0d34dd241fbedbc84c22df)

pkg/prometheus/server/operator.go
```diff
@@ -811,7 +808,25 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		c.endpointSliceSupported = true
+	case "Endpoints":
+		level.Info(logger).Log("msg", "using endpoints as service discovery role")
+		c.endpointSliceSupported = false
+	default:
+		level.Info(logger).Log("msg",
+			"unknown service discovery role, defaulting to endpoints",
+			"serviceDiscoveryRole",
+			serviceDiscoveryRole)
+		c.endpointSliceSupported = false
```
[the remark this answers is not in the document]
thaniri: Please see comment in https://github.com/prometheus-operator/prometheus-operator/pull/6672#discussion_r1652041060

Suggestion has been accepted.

pkg/prometheus/testdata/K8SSDConfigGenerationThree.golden
```diff
@@ -0,0 +1,5 @@
+kubernetes_sd_configs:
+- role: endpointslice
+  namespaces:
+    names:
+    - test
```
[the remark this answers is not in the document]
thaniri: I manually created this file. After running `make test-unit-update-golden` nothing changes.

pkg/prometheus/promcfg_test.go
```diff
@@ -884,6 +893,7 @@ func TestK8SSDConfigGeneration(t *testing.T) {
 				Spec: monitoringv1.PrometheusSpec{
 					CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
 						IgnoreNamespaceSelectors: false,
+						ServiceDiscoveryRole:     "EndpointSlice",
```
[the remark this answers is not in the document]
thaniri: Thank you for pointing this out. I have removed this setting in [5cab2fc](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/5cab2fc5481a9816f3adfeb919c8353435322c55)

I have opted not to add `assert.False(t, cg.endpointSliceSupported)` as it will default to false.

pkg/prometheus/promcfg_test.go
thaniri: Hello @haanhvu , thank you for pointing out that an existing function `TestServiceMonitorWithEndpointSliceEnable` exists which tests that a scrape configuration is properly generated if EndpointSlices are supported.

I am having trouble finding where in the code we have an example of the `sync()` function being tested. As the function is quite large it would probably require an integration test or extensive mocking. You are correct that the core logic of this PR is in the reconciliation loop evaluating if both 1) the Kubernetes API supports EndpointSlices and 2) the user has configured their prometheus-operator to use EndpointSlices for service discovery. **Would you be able to push me in the correct direction to test the reconciliation loop of Prometheus?**

The key lines I can identify are:

1. Where `sync()` is setting the `endpointSliceSupported` field on the `Operator` object: https://github.com/thaniri/prometheus-operator/blob/1da76da45e78597f973ba1341780c2e406786fd2/pkg/prometheus/server/operator.go#L813-L825
2. The function that creates new `Operator` objects: https://github.com/thaniri/prometheus-operator/blob/1da76da45e78597f973ba1341780c2e406786fd2/pkg/prometheus/server/operator.go#L123
3. Where the `New()` function is called from: https://github.com/thaniri/prometheus-operator/blob/1da76da45e78597f973ba1341780c2e406786fd2/cmd/operator/main.go#L352
4. Where the `Operator` object actually starts https://github.com/thaniri/prometheus-operator/blob/1da76da45e78597f973ba1341780c2e406786fd2/cmd/operator/main.go#L531-L533

I am envisioning instantiating an `Operator` and calling the `sync` function on it to see if the `endpointSliceSupported` field on the Operator actually gets updated.

**If there is no current testing on the `sync()` function, could that perhaps be its own issue?**

pkg/prometheus/agent/operator.go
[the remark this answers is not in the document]
thaniri: Thank you for the further explanation. This is a good thing to catch and ensure predictable behaviour out of the operator.

Please let me know if there is further work you would like to see for this particular suggestion.

pkg/prometheus/promcfg_test.go
haanhvu: Let me think how we can test this change as simply as we can. @simonpasquier @ArthurSens any ideas?

haanhvu:
@thaniri I'll try to take the next review this week. In the meantime could you rebase?

simonpasquier (commented):
This is a great start!
IMHO the decision about which discovery role to use should be delegated to `ConfigGenerator`. It has all the information needed to make the decision: it has the Prometheus/PrometheusAgent spec (Prom version + discovery role) + the information about whether endpointslice is supported.

Also the `endpointSliceSupported` field in the struct is badly named, it should rather be `useEndpointSlice` (or something simliar) and initialized in NewConfigGenerator() based on the inputs:
https://github.com/prometheus-operator/prometheus-operator/blob/ad006f4d607c99062d66936a8c02dd8bbaa6b40e/pkg/prometheus/promcfg.go#L60-L68

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
thaniri: This suggestion has been added. Thank you for your suggestion.

pkg/prometheus/server/operator.go
```diff
@@ -811,7 +808,24 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return err
 	}
 
+	cpf := p.GetCommonPrometheusFields()
+
+	switch serviceDiscoveryRole := *cpf.ServiceDiscoveryRole; serviceDiscoveryRole {
```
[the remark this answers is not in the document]
thaniri: I have separately made `cpf.ServiceDiscoveryRole` a pointer to a string instead of a string so it might not be necessary to use the `ptr` package to check the service discovery role in this switch statement.

pkg/apis/monitoring/v1/prometheus_types.go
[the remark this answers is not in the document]
thaniri: Thanks for pointing out this pattern. I have implemented this in [ab692b0](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/ab692b0d9d94cd0d011a9b20a8497dc701aca870)

pkg/prometheus/server/operator.go
```diff
@@ -811,7 +808,24 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	case "EndpointSlice":
+		level.Info(logger).Log("msg", "using endpointslice as service discovery role")
+		c.endpointSliceSupported = true
```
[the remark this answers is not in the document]
thaniri: This improves the code clarity. I have modified this in [3c0a27a](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/3c0a27a93a4e2a7583e5e5d744da68c8d0fe9ec8)

pkg/prometheus/promcfg_test.go
thaniri: @simonpasquier recommended moving out the core logic of this PR into `NewConfigGenerator()` in [this comment](https://github.com/prometheus-operator/prometheus-operator/pull/6672#pullrequestreview-2130701946). This is a very good suggestion as we will be able to easily unit test this.

I have modified TestServiceMonitorWithEndpointSliceEnable to account for this logic in [8ca42f5](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/8ca42f506ccd1d3ae786aee76c004c9c65dee7d0)

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -771,6 +771,12 @@ type CommonPrometheusFields struct {
 	// +listType=map
 	// +listMapKey=name
 	ScrapeClasses []ScrapeClass `json:"scrapeClasses,omitempty"`
+
+	// Defines the service discovery role used to discover targets from `ServiceMonitor` objects.
+	// If set, the value should be either "Endpoints" or "EndpointSlice".
+	// If unset, the operator assumes the "Endpoints" role.
+	// +optional
+	ServiceDiscoveryRole *string `json:"serviceDiscoveryRole,omitempty"`
```
ArthurSens, line 779:
```suggestion
	ServiceDiscoveryRole *ServiceDiscoveryRole `json:"serviceDiscoveryRole,omitempty"`
```

Let's use the newly created Enum :)

pkg/prometheus/promcfg.go
```diff
@@ -92,11 +92,28 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
 		return nil, fmt.Errorf("failed to parse scrape classes: %w", err)
 	}
 
+	endpointSliceSupportedAndConfigured := false // Always assume false to preserve original prometheus-operator behaviour.
+
+	// Check if the user has explicitly set the service discovery role to use.
+	switch serviceDiscoveryRole := *cpf.ServiceDiscoveryRole; serviceDiscoveryRole {
+	case "EndpointSlice":
```
ArthurSens, line 99:
```suggestion
	case monitoringv1.EndpointSliceRole:
```

pkg/prometheus/promcfg.go
```diff
@@ -92,11 +92,28 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
+		level.Info(logger).Log("msg", "using endpointslice as service discovery role")
+		endpointSliceSupportedAndConfigured = true
+	case "Endpoints":
```
ArthurSens, line 102:
```suggestion
	case monitoringv1.EndpointsRole:
```

pkg/prometheus/promcfg.go
```diff
@@ -62,13 +62,13 @@ type ConfigGenerator struct {
...
 	prom                   monitoringv1.PrometheusInterface
-	endpointSliceSupported bool
+	useEndpointSlice       bool // Whether to use EndpointSlice for service discovery from `ServiceMonitor` objects.
 	scrapeClasses          map[string]monitoringv1.ScrapeClass
 	defaultScrapeClassName string
 }
 
 // NewConfigGenerator creates a ConfigGenerator for the provided Prometheus resource.
-func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, endpointSliceSupported bool) (*ConfigGenerator, error) {
+func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface) (*ConfigGenerator, error) {
```
ArthurSens, line 71: Do I understand correctly that you're removing the argument that tells if the k8s cluster where the operator is running supports EndpointSlices or not? Shouldn't this information be kept and used?

E.g. what happens if the user sets the role to EndpointSlice, but the cluster doesn't support it?

pkg/prometheus/promcfg_test.go
```diff
@@ -825,11 +825,13 @@ func TestK8SSDConfigGeneration(t *testing.T) {
 	testcases := []struct {
 		apiServerConfig *monitoringv1.APIServerConfig
 		store           *assets.StoreBuilder
+		role            string
```
ArthurSens, line 828:
```suggestion
		role            monitoringv1.ServiceDiscoveryRole
```

pkg/prometheus/promcfg.go
```diff
@@ -92,11 +92,28 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
+			serviceDiscoveryRole)
+		endpointSliceSupportedAndConfigured = false
+	}
+
 	return &ConfigGenerator{
 		logger:                 logger,
 		version:                version,
 		prom:                   p,
-		endpointSliceSupported: endpointSliceSupported,
+		useEndpointSlice:       endpointSliceSupportedAndConfigured,
```
ArthurSens, line 116: What I'm imagining is that the argument `endpointSliceSupported` still exists, that tells if the cluster itself supports endpointSlice, and `endpointSliceConfigured` is what you're checking with the switch case above.

```suggestion
		useEndpointSlice:       endpointSliceSupported && endpointSliceConfigured,
```

ArthurSens (commented):
Thanks a lot for the patience and determination in this PR :)

I've made a few comments, one because I've noticed we created an Enum for the ServiceDiscoveryRole, but we aren't using it anywhere.

The second comment I'm not really sure if I'm missing something, but I have the impression that we check if the cluster supports the endpointslice role, but it seems like we're not using it for anything. 🤔

ArthurSens:
The tests shows that we're not checking for `nil` somewhere in `NewConfigGenerator` 😬

pkg/prometheus/promcfg.go
```diff
@@ -92,11 +92,28 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
 		return nil, fmt.Errorf("failed to parse scrape classes: %w", err)
 	}
 
+	endpointSliceSupportedAndConfigured := false // Always assume false to preserve original prometheus-operator behaviour.
+
+	// Check if the user has explicitly set the service discovery role to use.
+	switch serviceDiscoveryRole := *cpf.ServiceDiscoveryRole; serviceDiscoveryRole {
```
simonpasquier, line 98:
```suggestion
	switch serviceDiscoveryRole := ptr.Deref(cpf.ServiceDiscoveryRole, monitoringv1.Endpoints); serviceDiscoveryRole {
```

simonpasquier (commented):
This is getting in the right direction! I concur with @ArthurSens comment: we still need to pass the information to NewConfigGenerator() about the API supporting the Endpointslice resource or not.

pkg/prometheus/promcfg.go
thaniri: Thank you for your suggestion. I have implemented this in [7c1adf2](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/7c1adf2433cbba2fb6c1bd84d736450767c03110)

pkg/prometheus/promcfg.go
thaniri: Thank you for the code style suggestion. I have implemented this in [7d5c6a5](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/7d5c6a5421819696a553ae02c0af9df0f6b2beb2)

pkg/prometheus/promcfg.go
thaniri: Thank you for the code style suggestion. I have implemented this in [7d5c6a5](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/7d5c6a5421819696a553ae02c0af9df0f6b2beb2)

pkg/prometheus/promcfg.go
thaniri: Thank you for pointing this out. It was an oversight on my part. Your comment [here](https://github.com/prometheus-operator/prometheus-operator/pull/6672#discussion_r1662943437) helped me realize the mistake I made.

I have re-added the argument to pass through the support of endpointslices here [16dc1f3](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/16dc1f3d8a62d9076512e7e47993d7153177c4a9)

pkg/prometheus/promcfg_test.go
thaniri: Thank you for pointing this out. I have fixed this in [16dc1f3](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/16dc1f3d8a62d9076512e7e47993d7153177c4a9)

pkg/prometheus/promcfg.go
thaniri: This is a wonderful comment that points out the missed logic in my implementation. I have fixed this in [16dc1f3](https://github.com/prometheus-operator/prometheus-operator/pull/6672/commits/16dc1f3d8a62d9076512e7e47993d7153177c4a9)

thaniri:
> The tests show that we're not checking for `nil` somewhere in `NewConfigGenerator` 😬

I have fixed this now. Thank you to the reviewers for the patience in implementing this PR. It's summer where I live and as such holiday breaks are being taken :D

Tagging @simonpasquier and @ArthurSens directly to hopefully get a new review in!

pkg/prometheus/agent/operator.go
```diff
@@ -317,15 +317,12 @@ func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger
...
 	if err != nil {
 		level.Warn(o.logger).Log("msg", "failed to check if the API supports the endpointslice resources", "err ", err)
 	}
-	level.Info(o.logger).Log("msg", "Kubernetes API capabilities", "endpointslices", endpointSliceSupported)
-	// The operator doesn't yet support the endpointslices API.
-	// See https://github.com/prometheus-operator/prometheus-operator/issues/3862
-	// for details.
-	o.endpointSliceSupported = false
+	level.Info(o.logger).Log("msg", "Kubernetes API capabilities", "endpointslices", k8sAPIEndpointSliceSupported)
+	o.endpointSliceSupported = k8sAPIEndpointSliceSupported
```
ArthurSens, line 325: I guess we could avoid a variable assignment here. Do we really need the extra variable `k8sAPIEndpointSliceSupported`?

```go
	o.endpointSliceSupported, err := k8sutil.IsAPIGroupVersionResourceSupported(o.kclient.Discovery(), schema.GroupVersion{Group: "discovery.k8s.io", Version: "v1"}, "endpointslices")
	if err != nil {
		level.Warn(o.logger).Log("msg", "failed to check if the API supports the endpointslice resources", "err ", err)
	}
	level.Info(o.logger).Log("msg", "Kubernetes API capabilities", "endpointslices", o.endpointSliceSupported)
```

pkg/prometheus/promcfg.go
```diff
@@ -94,11 +96,28 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
+	endpointSliceConfigured := false // Always assume false to preserve original prometheus-operator behaviour.
+
+	// Check if the user has explicitly set the service discovery role to use.
+	switch serviceDiscoveryRole := ptr.Deref(cpf.ServiceDiscoveryRole, monitoringv1.EndpointsRole); serviceDiscoveryRole {
+	case monitoringv1.EndpointSliceRole:
+		level.Info(logger).Log("msg", "using endpointslice as service discovery role")
+		endpointSliceConfigured = true
+	case monitoringv1.EndpointsRole:
+		level.Info(logger).Log("msg", "using endpoints as service discovery role")
+		endpointSliceConfigured = false
```
ArthurSens, line 108:
```suggestion
```
This is no-op because the value is already false :)

pkg/prometheus/promcfg.go
```diff
@@ -94,11 +96,28 @@ func NewConfigGenerator(logger log.Logger, p monitoringv1.PrometheusInterface, e
...
+	default:
+		level.Warn(logger).Log("msg",
+			"unknown service discovery role %q, defaulting to endpoints. Configure serviceDiscoveryRole to 'EndpointSlice' to use endpointslice as service discovery role.",
+			serviceDiscoveryRole)
+		endpointSliceConfigured = false
```
ArthurSens, line 113:
```suggestion
```

Same thing here

pkg/prometheus/server/operator.go
```diff
@@ -354,15 +354,12 @@ func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger
...
 	if err != nil {
 		level.Warn(o.logger).Log("msg", "failed to check if the API supports the endpointslice resources", "err ", err)
 	}
-	level.Info(o.logger).Log("msg", "Kubernetes API capabilities", "endpointslices", endpointSliceSupported)
-	// The operator doesn't yet support the endpointslices API.
-	// See https://github.com/prometheus-operator/prometheus-operator/issues/3862
-	// for details.
-	o.endpointSliceSupported = false
+	level.Info(o.logger).Log("msg", "Kubernetes API capabilities", "endpointslices", k8sAPIEndpointSliceSupported)
+	o.endpointSliceSupported = k8sAPIEndpointSliceSupported
```
ArthurSens, line 362: Same comments here as I've made above :)

ArthurSens (commented):
I didn't have much time to go too deep into this review, so please expect another one in a few days. Simon is also on PTO this week, I'd prefer to wait for him to comeback before merging anything anyway 😬

pkg/prometheus/agent/operator.go
thaniri: I have implemented this suggestion. It reduces the LoC by 1 and saves 1 byte of memory each time `sync()` gets called.

pkg/prometheus/promcfg.go
thaniri: Line 99 is also a no-op as we assign false to the bool `endpointSliceConfigured`.

Perhaps we keep this no-op because I believe being explicit about intention instead of implicit is more helpful to future readers of the code.

pkg/prometheus/promcfg.go
thaniri: See comment https://github.com/prometheus-operator/prometheus-operator/pull/6672#discussion_r1681836580

pkg/prometheus/server/operator.go
thaniri: I have implemented this as in https://github.com/prometheus-operator/prometheus-operator/pull/6672#discussion_r1681835729

pkg/prometheus/promcfg.go
ArthurSens: Fair point :)

ArthurSens (dismissed):
LGTM, thanks for your patience! I'm hoping Simon and Ha Ahn also express their opinion before merging here :)

thaniri:
> LGTM, thanks for your patience! I'm hoping Simon and Ha Ahn also express their opinion before merging here :)

Thank you for your attentiveness and review @ArthurSens. I will wait for at least one more pull request approval. Please let me know if it is appropriate to directly message Simon or Ha Ahn.

simonpasquier (approved):
Thanks for the hard work! Tested locally and it works but one needs to grant the permissions on the endpointslices resource to the prometheus SA. I'll submit a follow-up PR to adjust the default manifests. I've got a few stylistic remarks but they don't warrant to delay merging this PR. The last thing would be to resolve the conflict.

thaniri:
Thanks for helping with getting the code to a shippable state.

I've resolved the merge conflict, could one of you, @simonpasquier or @ArthurSens, please approve again? (And merge if you'd like because we are in different timezones)

simonpasquier (approved):
Thanks!

PR #6686, opened 2024-06-17
Refactor the common implementation code (not including tests) between Prometheus's modes

haanhvu:
## Description

Refactor the common between Prometheus Agent's StatefulSet and DaemonSet to reduce code duplication and make reviewing #6652 easier. If seeing any pieces can be refactored further into the Server mode then do it too.

This PR just targets the implementation code duplication. The test duplication will be handled in another PR.

## Type of change

## Verification
CI green


pkg/prometheus/agent/common.go
```diff
@@ -0,0 +1,75 @@
...
+
+	return pImagePath, nil
+}
+
+func buildAgentArgs(
+	cpf monitoringv1.CommonPrometheusFields,
+	cg *prompkg.ConfigGenerator,
+) []monitoringv1.Argument {
+	promArgs := prompkg.BuildCommonPrometheusArgs(cpf, cg)
+	promArgs = appendAgentArgs(promArgs, cg, cpf.WALCompression)
```
simonpasquier, line 49: Since WALCompression exists both for Server and Agent, I'm not sure there's a reason why it can't be handled by BuildCommonPrometheusArgs()?

pkg/prometheus/agent/common.go
```diff
@@ -0,0 +1,75 @@
...
+func buildVolumes(
+	p monitoringv1.PrometheusInterface,
+	tlsSecrets *operator.ShardedSecret,
+) ([]v1.Volume, []v1.VolumeMount, []v1.VolumeMount, error) {
+	volumes, promVolumeMounts, err := prompkg.BuildCommonVolumes(p, tlsSecrets)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	configReloaderVolumeMounts := []v1.VolumeMount{
```
simonpasquier, line 63: I see that we have the same snippet between Agent and Server. Should we rather create a `prompkg.ConfigReloaderVolumeMounts()`? Or even move it to `operator.CreateConfigReloader()`?

simonpasquier (commented):
Thanks for starting this!

pkg/prometheus/agent/common.go
haanhvu: Yeah I'll look further if any pieces can also be refactored into the server mode. Refactoring agent's modes is Step 1.

pkg/prometheus/agent/common.go
haanhvu: same above

pkg/prometheus/agent/common.go
haanhvu: WALCompression exists in both Server and Agent. But the args are different:

- Server arg: https://github.com/prometheus-operator/prometheus-operator/blob/cbf0f66c687656b6a96e2acb8b618ef0aaaf138c/pkg/prometheus/server/statefulset.go#L567
- Agent arg: https://github.com/prometheus-operator/prometheus-operator/blob/cbf0f66c687656b6a96e2acb8b618ef0aaaf138c/pkg/prometheus/agent/statefulset.go#L436

pkg/prometheus/agent/common.go
haanhvu: yeah this makes sense I'll do it

haanhvu: I'll move this to pkg/operator/config_reloader.go

haanhvu: Now reread the code I'll move it to pkg/prometheus/statefulset.go instead, becaue we use prompkg fields here.

haanhvu: Now reread the code I see we already had `pkg/prometheus/statefulset.go`, so I'll move it here.

pkg/prometheus/statefulset.go
```diff
@@ -518,3 +519,100 @@ func GetStatupProbePeriodSecondsAndFailureThreshold(cfp monitoringv1.CommonProme
...
+}
+
+// The /-/ready handler returns OK only after the TSDB initialization has
+// completed. The WAL replay can take a significant time for large setups
+// hence we enable the startup probe with a generous failure threshold (15
+// minutes) to ensure that the readiness probe only comes into effect once
+// Prometheus is effectively ready.
+// We don't want to use the /-/healthy handler here because it returns OK as
+// soon as the web server is started (irrespective of the WAL replay).
+func MakeProbes(
```
haanhvu, line 589: As @simonpasquier said in https://github.com/prometheus-operator/prometheus-operator/pull/6652#discussion_r1637887343 we could have either same settings with statefulset or tighter timeouts for daemonset. Let's have a decision here, because it affects the refactoring?

haanhvu (commented):
@simonpasquier @ArthurSens @kakkoyun I refactored as far as I see possible. Please help review!

As stated in the description, this PR just covers the code duplication. I'll handle the test duplication in the next PR.

pkg/prometheus/statefulset.go
haanhvu: This probably would affect how we'll refactor `TestStartupProbeTimeoutSeconds` too? https://github.com/prometheus-operator/prometheus-operator/blob/cbf0f66c687656b6a96e2acb8b618ef0aaaf138c/pkg/prometheus/agent/statefulset_test.go#L193

pkg/prometheus/agent/common.go
simonpasquier: Ah bummer! Sorry for the noise then :)

pkg/prometheus/agent/statefulset.go
```diff
@@ -155,48 +154,24 @@ func makeStatefulSetSpec(
...
-		operator.StringValOrDefault(cpf.Version, operator.DefaultPrometheusVersion),
-		"",
-		"",
-	)
+	pImagePath, err := buildImagePath(cpf, c)
 	if err != nil {
 		return nil, err
 	}
 
 	cpf.EnableFeatures = append(cpf.EnableFeatures, "agent")
```
simonpasquier, line 162: reading this, I realize that someone could define `.spec.enabledFeatures: [agent]`  in a PrometheusAgent object. It doesn't do any harm (Prometheus will start and log E`xperimental agent mode enabled.` twice) but we might want to "fix" it.

pkg/prometheus/agent/statefulset.go
```diff
@@ -155,48 +154,24 @@ func makeStatefulSetSpec(
...
-	promArgs := prompkg.BuildCommonPrometheusArgs(cpf, cg)
-	promArgs = appendAgentArgs(promArgs, cg, cpf.WALCompression)
+	promArgs := buildAgentArgs(cpf, cg)
 
 	var ports []v1.ContainerPort
 	if !cpf.ListenLocal {
```
simonpasquier, line 166: can we move the if condition to `MakeContainerPorts()`?

pkg/prometheus/agent/statefulset.go
```diff
@@ -155,48 +154,24 @@ func makeStatefulSetSpec(
...
+		ports = prompkg.MakeContainerPorts(cpf)
 	}
 
-	volumes, promVolumeMounts, err := prompkg.BuildCommonVolumes(p, tlsSecrets)
+	volumes, promVolumeMounts, configReloaderVolumeMounts, err := prompkg.BuildVolumes(p, tlsSecrets)
 	if err != nil {
 		return nil, err
 	}
 
-	configReloaderVolumeMounts := []v1.VolumeMount{
```
simonpasquier, line 189: sorry if you answered somewhere else but could this be moved directly to `pkg/operator/config_reloader.go`?

simonpasquier (commented):
Thanks for doing this, I know how tedious it can be :)

pkg/prometheus/statefulset.go
simonpasquier: Until we're sure which direction to go (different probe settings or not), let's delay the refactoring then.

pkg/prometheus/server/statefulset.go
```diff
@@ -247,13 +246,7 @@ func makeStatefulSetSpec(
 
 	var ports []v1.ContainerPort
 	if !cpf.ListenLocal {
```
ArthurSens, line 248: Similar to simon's comment above, let's move this if inside makecontainerports()

ArthurSens (commented):
Good progress! I've added a few comments but they are all non-blockers

pkg/prometheus/statefulset.go
haanhvu: This piece is already a duplication in server and agent's statefulsets, so I think just keep this refactoring here for now? Later if we decide to change the timeouts for daemonset we can modify this refactoring to work with all three modes.

pkg/prometheus/agent/statefulset.go
haanhvu: So we need to check if `cpf.EnableFeatures` already contains `agent` or not right? Should I fix this in a separate PR?

pkg/prometheus/agent/statefulset.go
haanhvu: Yeah I said in https://github.com/prometheus-operator/prometheus-operator/pull/6686#discussion_r1643751042 that I'd move it to `config_reloader.go`. But then I realized this piece uses `prompkg` fields, so I decided to move it to `prompkg`.

The way I see it, `pkg/operator/config_reloader.go` is just pure config reloader's logic. If `prompkg` fields are used, then it's Prometheus's config reloader container's logic, then it's should be in `prompkg` instead. You agree?

pkg/prometheus/agent/common.go
```diff
@@ -0,0 +1,50 @@
...
+		"",
+		"",
+	)
+
+	if err != nil {
+		return "", err
+	}
+
+	return pImagePath, nil
+}
```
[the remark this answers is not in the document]
haanhvu: Yeah one goal of the refactoring is to make reviewing #6652 easier. As @simonpasquier said in the last meeting, we need to refactor so that we won't need to go back to statefulset code to check if the duplicated daemonset code is correct or not.

If there are no hard opinions, I'll keep this refactoring. Without it you might need to go back to statefulset code to check if `operator.BuildImagePath()` is correctly called in daemonset or not, i.e., if the arguments are correctly set and the returned values are correctly handled.

pkg/prometheus/agent/common.go
```diff
@@ -0,0 +1,50 @@
...
+
+func buildAgentArgs(
+	cpf monitoringv1.CommonPrometheusFields,
+	cg *prompkg.ConfigGenerator,
+) []monitoringv1.Argument {
+	promArgs := prompkg.BuildCommonPrometheusArgs(cpf, cg)
+	promArgs = appendAgentArgs(promArgs, cg, cpf.WALCompression)
+
+	return promArgs
+}
```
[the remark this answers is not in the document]
haanhvu: I wasn't actually thinking about WALCompression when making this change. My intention was the same as https://github.com/prometheus-operator/prometheus-operator/pull/6686#discussion_r1645374035

pkg/prometheus/statefulset.go
haanhvu: I refactored `TestStartupProbeTimeoutSeconds` in #6694 too. We still need this test for DaemonSet now, so I think we should refactor to reduce duplication anyway. We can modify the refactored function later if we decide to have different timeouts for DaemonSet.

pkg/prometheus/agent/statefulset.go
```diff
@@ -155,48 +154,19 @@ func makeStatefulSetSpec(
...
-				Name:          cpf.PortName,
-				ContainerPort: 9090,
-				Protocol:      v1.ProtocolTCP,
-			},
-		}
-	}
+	promArgs := buildAgentArgs(cpf, cg)
 
-	volumes, promVolumeMounts, err := prompkg.BuildCommonVolumes(p, tlsSecrets)
+	volumes, promVolumeMounts, configReloaderVolumeMounts, err := prompkg.BuildVolumes(p, tlsSecrets)
```
simonpasquier, line 165: I'd be fine with keeping BuildCommonVolumes() here and having another call to `CreateConfigReloaderVolumeMounts() like in the server part.

pkg/prometheus/agent/statefulset.go
simonpasquier: :+1:

pkg/prometheus/statefulset.go
```diff
@@ -518,3 +519,104 @@ func GetStatupProbePeriodSecondsAndFailureThreshold(cfp monitoringv1.CommonProme
...
+	if err != nil {
+		return monitoringv1.Argument{}, nil, nil, err
+	}
+
+	confArg, configVol, configMount, err := webConfig.GetMountParameters()
+	if err != nil {
+		return monitoringv1.Argument{}, nil, nil, err
+	}
+
+	return confArg, configVol, configMount, nil
```
simonpasquier, line 583: This is equivalent, right?
```suggestion
	return webConfig.GetMountParameters()
```

pkg/operator/image.go
```diff
@@ -56,6 +56,32 @@ func BuildImagePath(specImage, baseImage, version, tag, sha string) (string, err
 	return baseImage + ":" + version, nil
 }
 
+// BuildImagePathForAgent builds a container image path based on
+// the given parameters for Prometheus Agent.
+// Return specImage if not empty.
+// If image contains a tag or digest then image will be returned.
+// Otherwise, return image suffixed by either SHA, tag or version(in that order).
+// Inspired by kubernetes code handling of image building.
+func BuildImagePathForAgent(specImage, baseImage, version string) (string, error) {
```
simonpasquier, line 65: it could also call `BuildImagePath()` directly with empty strings for arguments which are irrelevant?

pkg/prometheus/agent/common.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func buildAgentArgs(
+	cpf monitoringv1.CommonPrometheusFields,
+	cg *prompkg.ConfigGenerator,
+) []monitoringv1.Argument {
+	promArgs := prompkg.BuildCommonPrometheusArgs(cpf, cg)
+	promArgs = appendAgentArgs(promArgs, cg, cpf.WALCompression)
+
+	return promArgs
```
simonpasquier, line 29: (nit)
```suggestion
	return appendAgentArgs(promArgs, cg, cpf.WALCompression)
```

pkg/prometheus/agent/statefulset.go
```diff
@@ -155,47 +154,24 @@ func makeStatefulSetSpec(
 ) (*appsv1.StatefulSetSpec, error) {
 	cpf := p.GetCommonPrometheusFields()
 
-	pImagePath, err := operator.BuildImagePath(
+	pImagePath, err := operator.BuildImagePathForAgent(
 		operator.StringPtrValOrDefault(cpf.Image, ""),
 		operator.StringValOrDefault("", c.PrometheusDefaultBaseImage),
```
simonpasquier, line 159: (nit) is it the same, right?
```suggestion
		c.PrometheusDefaultBaseImage,
```

haanhvu: There're other places like this too. Like `operator.StringPtrValOrDefault(a.Spec.Image, "")` I'll fix this in a separate PR which we can merge fast, since #6703 depends on this.

haanhvu: Ah no. They're not similar. #6703 is for string pointer, here is the string value.

pkg/operator/image.go
haanhvu: ah ok, now I fully understand what you mean in https://github.com/prometheus-operator/prometheus-operator/pull/6686#discussion_r1645778377

pkg/operator/image.go
```diff
@@ -56,6 +56,16 @@ func BuildImagePath(specImage, baseImage, version, tag, sha string) (string, err
 	return baseImage + ":" + version, nil
 }
 
+// BuildImagePathForAgent builds a container image path based on
+// the given parameters for Prometheus Agent.
+// Return specImage if not empty.
+// If image contains a tag or digest then image will be returned.
+// Otherwise, return image suffixed by either SHA, tag or version(in that order).
+// Inspired by kubernetes code handling of image building.
```
simonpasquier, line 64: (nit)
```suggestion
// Otherwise a combination of baseImage and version.
```

simonpasquier (dismissed):

simonpasquier (approved):

haanhvu:
@simonpasquier auto-merge doesn't work here, no ideas why 🤔

ArthurSens:
Is it because of the merge conflicts?

ArthurSens (approved):

haanhvu:
> Is it because of the merge conflicts?

It didn't have any conflicts when Simon enabled auto-merge. I'll resolve the new conflicts.

simonpasquier (approved):

PR #6687, opened 2024-06-17
Feat: Add `servicename` field to the Prometheus server and agent CRDs

### LINKED ISSUE #6060: Prometheus Operators incorrectly removes gRPC port from shared prometheus-operated Service resource

mviswanathsai:
## Description

[#6060](https://github.com/prometheus-operator/prometheus-operator/issues/6060) relates to exposing services with custom service names since  "governing" services don’t play well when the Prometheus-es in the same namespace need to expose different sets of ports, which can only be done by using different services. 
This PR adds a field called `servicename` where the user can specify the name of the service that will expose that particular set of Prometheus instances which are governed by the Prometheus CR. 

The behavior is as follows: 
1. If .spec.serviceName is defined and the service doesn't exist, the operator fails the reconciliation.
2. If the service exists, it exposes the respective Promethei.
3. If the .spec.serviceName is not defined, a regular governing service is exposed.

It is up to the user to expose a service that suits their needs. However, the operator does help in preventing users from linking the Prometheus resource with a service that does not select it. 

Closes #6060 

## Type of change

## Verification

## Changelog entry

```
Add `servicename` field to common Prometheus fields
```


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -701,6 +701,13 @@ type CommonPrometheusFields struct {
 	// +listType=map
 	// +listMapKey=name
 	ScrapeClasses []ScrapeClass `json:"scrapeClasses,omitempty"`
+
+	// The service name by which we wish to expose the Prometheus pods
+	// governed by the given Prometheus CR.
```
simonpasquier, line 706:
```suggestion
	// The name of the service name used by the underlying StatefulSet as its governing service. If defined, it must be created before the Prometheus object.
	// If empty, the operator will create and manage a service named `prometheus-operated`.
	// See https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id for more details.
```

pkg/prometheus/agent/operator.go
```diff
@@ -605,10 +605,35 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("synchronizing web config secret failed: %w", err)
 	}
 
-	// Create governing service if it doesn't exist.
+	// If a ServiceName is specified, check that the service exists in the namespace. If it does not exist, fail
+	// the reconciliation.
+	// If the ServiceName is not specified, create a governing service if it doesn't exist.
+	// Also, ensure that the Prometheus instance is selected by the service. If not, fail the reconciliation.
+	// TODO[mviswanathsai]: is it possible to refactor this (and other) code repetitions between the Prometheus server and agent?
```
simonpasquier, line 612: yes please :)
side suggestion: please do in a preliminary PR.

pkg/prometheus/agent/operator.go
```diff
@@ -605,10 +605,35 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+			return fmt.Errorf("synchronizing service failed: %w", err)
+		}
+
+		// Check that the selectors in the service actually select the Prometheus instance.
+		svcSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: svc.Spec.Selector})
+		if err != nil {
+			return fmt.Errorf("failed to get label selector while synchronizing service: %w", err)
+		}
+
+		if svcSelector.Matches(labels.Set(p.Labels)) {
```
simonpasquier, line 630: you want to check against the pod's labels, not the Prometheus object label.

pkg/prometheus/agent/operator.go
mviswanathsai: Should we complete this PR and then move onto the refactor? Seems like the refactor might be a bit of work.

pkg/prometheus/agent/operator.go
mviswanathsai: Got it.

pkg/prometheus/agent/operator.go
simonpasquier: I'd rather do it before this PR otherwise there's a risk we never get to it.

mviswanathsai: Got it!

mviswanathsai: Can you give me pointers on where to start with the refactor? @simonpasquier

pkg/prometheus/agent/operator.go
simonpasquier: this is still not correct: the Kubernetes service's selector is used to select pods based on their labels. In practice you need to use `makeSelectorLabels()` (the function could use a more expressive name, e.g. `selectorForPrometheusPods`).

pkg/prometheus/agent/operator.go
simonpasquier: Sorry I was confused: it should be done in this PR. The logic when `p.Spec.ServiceName != nil` should be similar between agent and server.

pkg/prometheus/agent/operator.go
mviswanathsai: Not sure to follow here, I do use `makeSelectorLabels` and then check against the label selectors present in the service.

mviswanathsai:
The e2e test I'd written fails, I just wanted to push it to get some opinions on it as well.
CC: @slashpai @ArthurSens

mviswanathsai:
PTAL @simonpasquier. I also have a few things to discuss regarding the e2e tests, and I have added some concerns as `TODO` comments in the code.

pkg/prometheus/common.go
```diff
@@ -594,3 +598,31 @@ func MakeProbes(
...
+			return fmt.Errorf("service %s/%s does not exist: %w", prometheusNS, *serviceName, err)
+		}
+
+		return err
+	}
+
+	// Get the user defined label selectors from the service.
+	svcSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: svc.Spec.Selector})
+	if err != nil {
+		return fmt.Errorf("failed to get label selector while synchronizing service: %w", err)
```
simonpasquier, line 618: IIUC this should never happen in real life.
```suggestion
		return fmt.Errorf("failed to convert service selector: %w", err)
```

pkg/prometheus/common.go
```diff
@@ -594,3 +598,31 @@ func MakeProbes(
...
+	}
+
+	// Check if the svcSelector is the same as the Selector used to match the pods handled by this
+	// Prometheus resource.
+	if !svcSelector.Matches(labels.Set(selectorLabels)) {
+		return fmt.Errorf("service %s/%s does not select prometheus server %s", prometheusNS, *serviceName, prometheusName)
```
simonpasquier, line 624: (suggestion) let's print out the details.
```suggestion
		return fmt.Errorf("service %s/%s  with selector %q does not select Prometheus pods %s with labels %q", prometheusNS, *serviceName, prometheusName, svcSelector.String(), labels.Set(selectorLabels).String())
```

pkg/prometheus/common.go
```diff
@@ -594,3 +598,31 @@ func MakeProbes(
 
 	return startupProbe, readinessProbe, livenessProbe
 }
+
+// If a ServiceName is specified, check that the service exists in the namespace. If it does not exist, fail
+// the reconciliation.
+// Also, ensure that the Prometheus instance is selected by the service. If not, fail the reconciliation.
+func CheckCustomService(ctx context.Context, serviceName *string, prometheusNS string, prometheusName string, svcClient clientv1.ServiceInterface, selectorLabels map[string]string) error {
```
simonpasquier, line 605: What if we pass a PrometheusInterface parameter instead? It would simplify the function signature.

test/e2e/prometheus_test.go
```diff
@@ -5169,6 +5169,77 @@ func testPrometheusStatusScale(t *testing.T) {
 	}
 }
 
+// TODO: Also make sure no governing service was created?
```
simonpasquier, line 5172: I'd be ok not checking this.

test/e2e/prometheus_test.go
```diff
@@ -5169,6 +5169,77 @@ func testPrometheusStatusScale(t *testing.T) {
...
+
+	p := framework.MakeBasicPrometheus(ns, name, name, 1)
+	p.Spec.ServiceName = ptr.To(fmt.Sprintf("%s-service", name))
+
+	p, err := framework.CreatePrometheus(context.Background(), ns, p)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// TODO: Maybe we can just send a http request to the Prometheus pod to see if it is active?
```
simonpasquier, line 5228: IIUC the existing test is enough: WaitForHealthyTargets() will target the custom service so if it replies ok, you know that everything is wired properly.

test/e2e/prometheus_test.go
```diff
@@ -5169,6 +5169,77 @@ func testPrometheusStatusScale(t *testing.T) {
 	}
 }
 
+// TODO: Also make sure no governing service was created?
+// What do we want to do? we want to make sure that when you are creating your own service and
+// the service is actually selected. But this is pretty self explanatory.
+// Something that we also want to check for is to make sure that if the service does not select
+// our prometheus, and the prometheus selects our service then we reject such a config.
```
simonpasquier, line 5176: Yes this would useful to test the error cases.

pkg/prometheus/common.go
mviswanathsai: I have made the change, but could you elaborate a bit on that statement?

pkg/prometheus/common.go
mviswanathsai: Changed it a bit
```
		return fmt.Errorf("service %s/%s with selector %s does not select Prometheus sset pods %s/%s with labels %s",
			prometheusNS, serviceName, svcSelector.String(), prometheusNS, prometheusName, labels.Set(selectorLabels).String())
```

pkg/prometheus/common.go
mviswanathsai: I think at the time of writing, my POV was that do we need to pass the entire interface when we only need 3 string values from it. But no strong opinions on this one.

test/e2e/prometheus_test.go
mviswanathsai: Forgot to remove this, mb!

test/e2e/prometheus_test.go
mviswanathsai: I thought perhaps we could do away with the dummy target creation, but since it does mimic real life scenarios I think it should be good.

mviswanathsai:
Random issues kept popping up with the local e2e tests (not concerning the actual test itself) which lead to failures over and over again. So I decided to just push the changes and then deal with it.

mviswanathsai:
I think this is good for a review now @simonpasquier

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -821,6 +821,14 @@ type CommonPrometheusFields struct {
 	//
 	// +optional
 	TSDB *TSDBSpec `json:"tsdb,omitempty"`
+
+	// The name of the service name used by the underlying StatefulSet as its governing service.
```
simonpasquier, line 825: (nit) there are multiple statefulsets in case of sharding.
```suggestion
	// The name of the service name used by the underlying StatefulSet(s) as the governing service.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -821,6 +821,14 @@ type CommonPrometheusFields struct {
...
+	// If defined, it must be created before the Prometheus object in the same namespace and it must define a selector that will match the pod labels.
+	// If empty, the operator will create and manage a service named `prometheus-operated`.
+	// When deploying multiple Prometheus/PrometheusAgent resources in the same namespace, it is recommended to specify a different value for each.
+	// See https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id for more details.	// +optional
```
simonpasquier, line 829:
```suggestion
	// See https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id for more details.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -821,6 +821,14 @@ type CommonPrometheusFields struct {
...
+	// +optional
```
simonpasquier, line 830: We'll need a validation on the length too (>= 1).

pkg/prometheus/agent/operator.go
```diff
@@ -714,10 +714,18 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
...
-	if _, err := k8sutil.CreateOrUpdateService(ctx, svcClient, makeStatefulSetService(p, c.config)); err != nil {
-		return fmt.Errorf("synchronizing governing service failed: %w", err)
+	selectorLabels := makeSelectorLabels(p.Name)
+
+	if p.Spec.ServiceName != nil {
+		if err := prompkg.CheckCustomService(ctx, *p.Spec.ServiceName, p.Namespace, p.Name, svcClient, selectorLabels); err != nil {
+			return fmt.Errorf("synchronizing custom service failed: %w", err)
+		}
+	} else {
+		// If the ServiceName is not specified, create a governing service if one doesn't already exist.
```
simonpasquier, line 725:
```suggestion
		// If the ServiceName is not specified, the controller manages the governing service.
```

pkg/prometheus/server/operator.go
```diff
@@ -787,10 +787,22 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("failed to reconcile Thanos config secret: %w", err)
 	}
 
-	// Create governing service if it doesn't exist.
+	if err := c.createOrUpdateThanosConfigSecret(ctx, p); err != nil {
+		return fmt.Errorf("failed to reconcile Thanos config secret: %w", err)
+	}
```
simonpasquier, line 792: mistake?
```suggestion
```

pkg/prometheus/common.go
```diff
@@ -439,3 +443,32 @@ func BuildWebconfig(
 
 	return webConfig.GetMountParameters()
 }
+
+// If a ServiceName is specified, check that the service exists in the namespace. If it does not exist, fail
+// the reconciliation.
+// Also, ensure that the Prometheus instance is selected by the service. If not, fail the reconciliation.
+func CheckCustomService(ctx context.Context, serviceName string, prometheusNS string, prometheusName string, svcClient clientv1.ServiceInterface, selectorLabels map[string]string) error {
```
simonpasquier, line 450: If the function receives a `PrometheusInterface` argument + svcClient + selectorLabels, it could do:
* Verify that the service exists and matches the pods if p.Spec.ServiceName isn't nil
* Otherwise create/update the prometheus-operated service.

pkg/prometheus/common.go
simonpasquier: I mean to say that the service shouldn't have an invalid label selector value (how would it be allowed by the API in the first place?). But it's better to handle the error in any case.

pkg/prometheus/server/operator.go
mviswanathsai: Might've accidentally smashed the key binding

pkg/prometheus/common.go
mviswanathsai: I agree, that way we could do everything related to the service creation within this one function.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -862,6 +862,15 @@ type CommonPrometheusFields struct {
 	// +optional
 	TSDB *TSDBSpec `json:"tsdb,omitempty"`
 
+	// The name of the service name used by the underlying StatefulSet(s) as the governing service.
+	// If defined, it must be created before the Prometheus object in the same namespace and it must define a selector that will match the pod labels.
+	// If empty, the operator will create and manage a service named `prometheus-operated`.
```
simonpasquier, line 867:
```suggestion
	// If empty, the operator will create and manage a headless service named `prometheus-operated` for Prometheus resources or `prometheus-agent-operated` for PrometheusAgent resources.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -862,6 +862,15 @@ type CommonPrometheusFields struct {
 	// +optional
 	TSDB *TSDBSpec `json:"tsdb,omitempty"`
 
+	// The name of the service name used by the underlying StatefulSet(s) as the governing service.
+	// If defined, it must be created before the Prometheus object in the same namespace and it must define a selector that will match the pod labels.
```
simonpasquier, line 866:
```suggestion
	// If defined, it must be created before the Prometheus/PrometheusAgent resource in the same namespace and it must define a selector that will match the pod labels.
```

mviswanathsai:
Should we take the de-duplication commit up here? or maybe have it in a separate PR with tests (if necessary). I don't have a strong opinion but I would be leaning a bit towards the separate PR.

mviswanathsai:
```
Error: test/e2e/prometheus_test.go:5338:89: cannot use "web" (untyped string constant) as *string value in struct literal
```
Not sure this error is relevant? the type for Port is `string` AFAIK.

simonpasquier:
that's a compilation error in the e2e tests.

mviswanathsai:
Yes, the e2e tests compiled just fine earlier and nothing has changed since AFAIK...

mviswanathsai:
This is also how the type has been used in other parts of the code
```go
					PodMetricsEndpoints: []monitoringv1.PodMetricsEndpoint{
						{
							Port:     "web",
							Interval: "30s",
// Other specifications
                                                 }
```

simonpasquier:
Hmm it's strange. Maybe rebase on top of main?

simonpasquier:
you'll need to rebase since #7162 merged

mviswanathsai:
I rebased and changed the logic a bit. The `CheckGoverningService` function was initially doing two things: validation, or creation of a new service. Now, I decided to keep it in the same function because it ultimately has to do with the "governing" function for the statefulset, but come to think of it... I cannot really name the function properly.

It sounds silly but I decided if I cannot name the function then it is probably doing a couple of things that shouldn't be done by one function. Of course, it can, but it didn't really sit right for me at least in terms of readability. Let me know what you think about this @simonpasquier.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -862,6 +862,16 @@ type CommonPrometheusFields struct {
 	// +optional
 	TSDB *TSDBSpec `json:"tsdb,omitempty"`
 
+	// The name of the service name used by the underlying StatefulSet(s) as the governing service.
+	// If defined, it must be created before the Prometheus/PrometheusAgent resource in the same namespace and it must define a selector that will match the pod labels.
```
simonpasquier, line 866: (nit) just to be explicit
```suggestion
	// If defined, the Service must be created before the Prometheus/PrometheusAgent resource in the same namespace and define a selector that matches the pod labels.
```

pkg/prometheus/common.go
```diff
@@ -474,3 +478,39 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
 
 	return svc
 }
+
+// This function is responsible for the following:
+// If a ServiceName is specified, verify that the custom governing service exists in the namespace.
+// If it does not exist, fail the reconciliation.
+// Also, ensure that the Prometheus instance is selected by the custom governing service.
+// If it is not selected, fail the reconciliation.
+// If a ServiceName is not specified, create or update the default governing service.
```
simonpasquier, line 487: The comment needs an update.

pkg/prometheus/common.go
```diff
@@ -474,3 +478,39 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+func CheckCustomGoverningService(ctx context.Context, prometheus monitoringv1.PrometheusInterface, svcClient clientv1.ServiceInterface, selectorLabels map[string]string) error {
```
simonpasquier, line 488: (nit)
```suggestion
func EnsureGoverningService(ctx context.Context, prometheus monitoringv1.PrometheusInterface, svcClient clientv1.ServiceInterface, selectorLabels map[string]string) error {
```

pkg/prometheus/common.go
```diff
@@ -474,3 +478,39 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+	serviceName := prometheus.GetCommonPrometheusFields().ServiceName
+	prometheusNS := prometheus.GetObjectMeta().GetNamespace()
+	prometheusName := prometheus.GetObjectMeta().GetName()
+
+	if serviceName == nil {
+		return nil
+	}
```
simonpasquier, line 495: This would be a programmatic bug.

pkg/prometheus/common.go
```diff
@@ -474,3 +478,39 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+	// Check if the custom governing service is defined in the same namespace and selects the Prometheus pod.
+	svc, err := svcClient.Get(ctx, *serviceName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return fmt.Errorf("custom governing service %s/%s does not exist", prometheusNS, *serviceName)
+		}
+
```
simonpasquier, line 502: I wouldn't have a special handling for not-found errors.
```suggestion
```

pkg/prometheus/common.go
```diff
@@ -474,3 +478,39 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+		return fmt.Errorf("failed to check custom governing service %s/%s: %w", prometheusNS, *serviceName, err)
+	}
+
+	svcSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: svc.Spec.Selector})
+	if err != nil {
+		return fmt.Errorf("failed to get the selector labels for custom governing service %s/%s: %w", prometheusNS, *serviceName, err)
+	}
+
+	if !svcSelector.Matches(labels.Set(selectorLabels)) {
+		return fmt.Errorf("custom governing service %s/%s with selector %s does not select Prometheus sset pods %s/%s with labels %s",
```
simonpasquier, line 512:
```suggestion
		return fmt.Errorf("custom governing service %s/%s with selector %s does not select Prometheus pods %s/%s with labels %s",
```

pkg/prometheus/common.go
simonpasquier, line 488: it should be possible to write unit tests for this function.

test/e2e/prometheus_test.go
```diff
@@ -5300,6 +5300,70 @@ func testPrometheusStatusScale(t *testing.T) {
...
+				"app.kubernetes.io/managed-by": "prometheus-operator",
+			},
+		},
+	}
+
+	if _, err := framework.CreateOrUpdateService(context.Background(), ns, svc); err != nil {
+		t.Fatal(fmt.Errorf("failed to create app service: %w", err))
+	}
+
+	pm := framework.MakeBasicPodMonitor(name)
```
simonpasquier, line 5336: do we really need to create a pod monitor and check the targets?

test/e2e/prometheus_test.go
```diff
@@ -5300,6 +5300,70 @@ func testPrometheusStatusScale(t *testing.T) {
...
+	}
+
+	if err := framework.WaitForHealthyTargets(context.Background(), ns, svc.Name, 1); err != nil {
+		t.Fatal(err)
+	}
+
+	// Ensure that governing service was not created.
+	governingServiceName := "prometheus-operated"
+	if _, err := framework.KubeClient.CoreV1().Services(ns).Get(context.Background(), governingServiceName, metav1.GetOptions{}); err != nil {
+		if !apierrors.IsNotFound(err) {
```
simonpasquier, line 5361: we need to ensure that err != nil and apierrors.IsNotFound(err) is true.

test/e2e/prometheus_test.go
mviswanathsai: I think I rushed it at the time, maybe we could use `framework.PrometheusSVCGetRequest()`.

pkg/prometheus/common.go
mviswanathsai: Could you point me to an example that uses mock clients?

simonpasquier: sure!

https://github.com/prometheus-operator/prometheus-operator/blob/b2a6a6ed8c33ee9a9a8eedd2809fbbe5895e2aba/pkg/prometheus/resource_selector_test.go#L1079-L1102

pkg/prometheus/common.go
```diff
@@ -474,3 +477,36 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+// as the Prometheus/PrometheusAgent.
+// If it does not exist, fail the reconciliation.
+//
+// If the ServiceName is specified and a service with the same name exists in the same namespace as the
+// Prometheus/PrometheusAgent resource, ensure that the custom governing service's selector matches the
+// Prometheus/PrometheusAgent statefulsets.
+// If it is not selected, fail the reconciliation.
+func EnsureCustomGoverningService(ctx context.Context, prometheus monitoringv1.PrometheusInterface, svcClient clientv1.ServiceInterface, selectorLabels map[string]string) error {
+	serviceName := prometheus.GetCommonPrometheusFields().ServiceName
+	prometheusNS := prometheus.GetObjectMeta().GetNamespace()
```
simonpasquier, line 493: (nit)
```suggestion
	namespace := prometheus.GetObjectMeta().GetNamespace()
```

pkg/prometheus/common.go
```diff
@@ -474,3 +477,36 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+// If `ServiceName` is specified, verify that a service with the same name exists and is in the same namespace
+// as the Prometheus/PrometheusAgent.
+// If it does not exist, fail the reconciliation.
+//
+// If the ServiceName is specified and a service with the same name exists in the same namespace as the
+// Prometheus/PrometheusAgent resource, ensure that the custom governing service's selector matches the
+// Prometheus/PrometheusAgent statefulsets.
+// If it is not selected, fail the reconciliation.
+func EnsureCustomGoverningService(ctx context.Context, prometheus monitoringv1.PrometheusInterface, svcClient clientv1.ServiceInterface, selectorLabels map[string]string) error {
+	serviceName := prometheus.GetCommonPrometheusFields().ServiceName
```
simonpasquier, line 492: (nit)
```suggestion
	name := prometheus.GetCommonPrometheusFields().ServiceName
```

pkg/prometheus/common.go
```diff
@@ -474,3 +477,36 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+		return fmt.Errorf("failed to get custom governing service %s/%s: %w", prometheusNS, *serviceName, err)
+	}
+
+	svcSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: svc.Spec.Selector})
+	if err != nil {
+		return fmt.Errorf("failed to get the selector labels for custom governing service %s/%s: %w", prometheusNS, *serviceName, err)
+	}
+
+	if !svcSelector.Matches(labels.Set(selectorLabels)) {
+		return fmt.Errorf("custom governing service %s/%s with selector %s does not select Prometheus pods %s/%s with labels %s",
```
simonpasquier, line 508: (nit) I don't think that we need to repeat the prometheus resource's name. It should already be present in the logged information

pkg/prometheus/common.go
```diff
@@ -474,3 +477,36 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
 
 	return svc
 }
+
+// This function is responsible for the following:
+//
+// If `ServiceName` is specified, verify that a service with the same name exists and is in the same namespace
```
simonpasquier, line 483: an invariant of the function is that ServiceName isn't nil.
```suggestion
// Verify that the service exists in the Prometheus resource's namespace.
```

pkg/prometheus/common.go
```diff
@@ -474,3 +477,36 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+// as the Prometheus/PrometheusAgent.
+// If it does not exist, fail the reconciliation.
+//
+// If the ServiceName is specified and a service with the same name exists in the same namespace as the
+// Prometheus/PrometheusAgent resource, ensure that the custom governing service's selector matches the
+// Prometheus/PrometheusAgent statefulsets.
+// If it is not selected, fail the reconciliation.
```
simonpasquier, line 490:
```suggestion
// If it is not selected, fail the reconciliation.
// Warning: the function will panic if the resource's ServiceName is nil.
```

pkg/prometheus/server/operator.go
```diff
@@ -793,28 +793,33 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
-		map[string]string{"app.kubernetes.io/name": "prometheus"},
-		p,
-		c.config,
-	)
+	if p.Spec.ServiceName != nil {
+		if err := prompkg.EnsureCustomGoverningService(ctx, p, svcClient, selectorLabels); err != nil {
+			return fmt.Errorf("validating custom service failed: %w", err)
+		}
+	} else {
+		// If the ServiceName is not specified, create a governing service if one doesn't already exist.
```
simonpasquier, line 804:
```suggestion
		// Reconcile the default governing service.
```

pkg/prometheus/server/operator.go
```diff
@@ -793,28 +793,33 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
-	// Reconcile the governing service.
-	svc := prompkg.BuildStatefulSetService(
-		governingServiceName,
-		map[string]string{"app.kubernetes.io/name": "prometheus"},
-		p,
-		c.config,
-	)
+	if p.Spec.ServiceName != nil {
+		if err := prompkg.EnsureCustomGoverningService(ctx, p, svcClient, selectorLabels); err != nil {
+			return fmt.Errorf("validating custom service failed: %w", err)
```
simonpasquier, line 801: (nit) We can probably avoid wrapping the returned error which has enough information already IMHO.
```suggestion
			return err
```

pkg/prometheus/common.go
mviswanathsai: Got it.

pkg/prometheus/common.go
mviswanathsai: Thanks!

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -862,6 +862,16 @@ type CommonPrometheusFields struct {
 	// +optional
 	TSDB *TSDBSpec `json:"tsdb,omitempty"`
 
+	// The name of the service name used by the underlying StatefulSet(s) as the governing service.
+	// If defined, the Service  must be created before the Prometheus/PrometheusAgent resource in the same namespace and it must define a selector that will matches the pod labels.
```
simonpasquier, line 866:
```suggestion
	// If defined, the Service  must be created before the Prometheus/PrometheusAgent resource in the same namespace and it must define a selector that  matches the pod labels.
```

pkg/prometheus/common.go
```diff
@@ -474,3 +477,32 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+func EnsureCustomGoverningService(ctx context.Context, namespace string, serviceName string, svcClient clientv1.ServiceInterface, selectorLabels map[string]string) error {
+	// Check if the custom governing service is defined in the same namespace and selects the Prometheus pod.
+	svc, err := svcClient.Get(ctx, serviceName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get custom governing service %s/%s: %w", namespace, serviceName, err)
+	}
+
+	svcSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: svc.Spec.Selector})
+	if err != nil {
+		return fmt.Errorf("failed to get the selector labels for custom governing service %s/%s: %w", namespace, serviceName, err)
```
simonpasquier, line 500: (nit)
```suggestion
		return fmt.Errorf("failed to parse the selector labels for custom governing service %s/%s: %w", namespace, serviceName, err)
```

pkg/prometheus/common.go
```diff
@@ -474,3 +477,32 @@ func BuildStatefulSetService(name string, selector map[string]string, p monitori
...
+	}
+
+	if !svcSelector.Matches(labels.Set(selectorLabels)) {
+		return fmt.Errorf("custom governing service %s/%s with selector %s does not select Prometheus/PrometheusAgent pods with labels %s",
```
simonpasquier, line 504:
```suggestion
		return fmt.Errorf("custom governing service %s/%s with selector %q does not select Prometheus/PrometheusAgent pods with labels %q",
```

pkg/prometheus/agent/operator.go
```diff
@@ -714,15 +714,24 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
...
-	svc := prompkg.BuildStatefulSetService(
-		governingServiceName,
-		map[string]string{"app.kubernetes.io/name": "prometheus-agent"},
-		p,
-		c.config,
-	)
-	if _, err := k8sutil.CreateOrUpdateService(ctx, c.kclient.CoreV1().Services(p.Namespace), svc); err != nil {
-		return fmt.Errorf("synchronizing governing service failed: %w", err)
+	svcClient := c.kclient.CoreV1().Services(p.Namespace)
+	selectorLabels := makeSelectorLabels(p.Name)
```
simonpasquier, line 718: (nit) this could be moved in the first if block

pkg/prometheus/server/operator.go
```diff
@@ -793,28 +793,33 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("synchronizing web config secret failed: %w", err)
 	}
 
-	if err := c.createOrUpdateThanosConfigSecret(ctx, p); err != nil {
-		return fmt.Errorf("failed to reconcile Thanos config secret: %w", err)
-	}
+	svcClient := c.kclient.CoreV1().Services(p.Namespace)
+	selectorLabels := makeSelectorLabels(p.Name)
```
simonpasquier, line 797: same here

mviswanathsai:
Addressed comments, not sure why the other e2e test is failing though?

simonpasquier:
PrometheusAgentSSetServiceName fails with an error which might be transient:
````
=== CONT  TestAllNS/y/PrometheusAgentSSetServiceName
    prometheusagent_test.go:623: 
        	Error Trace:	/home/runner/work/prometheus-operator/prometheus-operator/test/e2e/prometheusagent_test.go:623
        	Error:      	Received unexpected error:
        	            	the server is currently unable to handle the request (get services http:test-agent-servicename-service:web)
        	Test:       	TestAllNS/y/PrometheusAgentSSetServiceName
````

TestAllNS/y/PromWebWithThanosSidecar looks more genuine but I can't tell what's happening except that the init config reloader fails to start:

````
=== NAME  TestAllNS/y/PromWebWithThanosSidecar
    prometheus_test.go:3801: Creating prometheus failed: waiting for 1 Prometheus instances timed out (basic-prometheus): prometheus allns-y-promwebwiththanossidecar-so4dvz-1-5a129d16/basic-prometheus failed to become available: client rate limiter Wait returned an error: context deadline exceeded: context deadline exceeded
allns-y-promwebwiththanossidecar-so4dvz-1-5a129d16: failed to get pod logs: failed to retrieve logs of container "prometheus" (pod allns-y-promwebwiththanossidecar-so4dvz-1-5a129d16/prometheus-basic-prometheus-0): the server rejected our request for an unknown reason (get pods prometheus-basic-prometheus-0)
````

**Comment by mviswanathsai** [comment]
> PrometheusAgentSSetServiceName fails with an error which might be transient:

Ah, this is probably because we are sending the request before the Prometheus(Agent) becomes available. I made the change with the Prometheus test, forgot the agent test.

**Comment by mviswanathsai** [comment]
> TestAllNS/y/PromWebWithThanosSidecar looks more genuine but I can't tell what's happening except that the init config reloader fails to start:

This was happening because I accidentally removed the code that creates the Thanossidecar's webconfig secret. Yikes!

**Comment by mviswanathsai** [comment]
Not entirely sure why the test is failing. But can you review the PR anyway @simonpasquier ?

File: test/e2e/prometheus_test.go [hunk 38]
```diff
@@ -5300,6 +5300,56 @@ func testPrometheusStatusScale(t *testing.T) {
...
+			Selector: map[string]string{
+				"prometheus":                   name,
+				"app.kubernetes.io/name":       "prometheus",
+				"app.kubernetes.io/instance":   name,
+				"app.kubernetes.io/managed-by": "prometheus-operator",
+			},
+		},
+	}
+
+	_, err := framework.CreateOrUpdateService(context.Background(), ns, svc)
```
simonpasquier, line 5332: we don't really need a create-or-update function?
```suggestion
	_, err := framework..KubeClient.CoreV1().Services(ns).Create(ctx, svc, metav1.CreateOptions{})
```

test/e2e/prometheus_test.go
```diff
@@ -5300,6 +5300,56 @@ func testPrometheusStatusScale(t *testing.T) {
...
+	require.NoError(t, err)
+
+	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, ns)
+
+	p := framework.MakeBasicPrometheus(ns, name, name, 1)
+	p.Spec.ServiceName = ptr.To(fmt.Sprintf("%s-service", name))
```
simonpasquier, line 5338:
```suggestion
	p.Spec.ServiceName = ptr.To(svc.Name)
```

test/e2e/prometheus_test.go
```diff
@@ -5300,6 +5300,56 @@ func testPrometheusStatusScale(t *testing.T) {
...
+
+	_, err = framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, p)
+	require.NoError(t, err)
+
+	success, err := framework.BasicQueryWorking(context.Background(), ns, svc.Name)
```
simonpasquier, line 5343: why not use `framework.GetActiveTargets` here too?

simonpasquier (commented):
just a few comments on the e2e tests otherwise lgtm, thanks!

test/framework/service.go
```diff
@@ -71,6 +71,35 @@ func (f *Framework) CreateOrUpdateServiceAndWaitUntilReady(ctx context.Context,
 	return finalizerFn, nil
 }
 
+func (f *Framework) CreateOrUpdateService(ctx context.Context, namespace string, service *v1.Service) (FinalizerFn, error) {
```
simonpasquier, line 74: I guess that we don't need this function anymore?

test/framework/prometheus.go
simonpasquier, line 1: We don't need this change anymore?

test/e2e/prometheusagent_test.go
```diff
@@ -579,6 +580,57 @@ func testPrometheusAgentDaemonSetSelectPodMonitor(t *testing.T) {
...
+		},
+	}
+
+	_, err := framework.KubeClient.CoreV1().Services(ns).Create(context.Background(), svc, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, ns)
+
+	p := framework.MakeBasicPrometheusAgent(ns, name, name, 1)
+	p.Spec.ServiceName = ptr.To(fmt.Sprintf("%s-service", name))
```
simonpasquier, line 617: (nit)
```suggestion
	p.Spec.ServiceName = &svc.Name
```

test/e2e/prometheusagent_test.go
```diff
@@ -579,6 +580,57 @@ func testPrometheusAgentDaemonSetSelectPodMonitor(t *testing.T) {
...
+
+	_, err = framework.CreatePrometheusAgentAndWaitUntilReady(context.Background(), ns, p)
+	require.NoError(t, err)
+
+	// The Agent doesn't support querying. So we just ensure that we are able to reach the
+	// Prometheus Agent through our service.
```
simonpasquier, line 623: (nit) since both tests request the same API endpoint.
```suggestion
```

test/e2e/prometheusagent_test.go
```diff
@@ -579,6 +580,57 @@ func testPrometheusAgentDaemonSetSelectPodMonitor(t *testing.T) {
...
+	targets, err := framework.GetActiveTargets(context.Background(), ns, svc.Name)
+	require.NoError(t, err)
+	require.Empty(t, targets)
+
+	// Ensure that governing service was not created.
+	governingServiceName := "prometheus-operated"
```
simonpasquier, line 629: it might be more reliable to list all the services in the namespace and check that we have only one item (same comment for the prometheus test).
```suggestion
	governingServiceName := "prometheus-agent-operated"
```

mviswanathsai: Will check the number of services instead.

mviswanathsai: And also added a check to ensure that the one service we have is the one we expect. This should be pretty reliable.

test/framework/service.go
mviswanathsai: Yep! missed it somehow. Also went ahead and reverted back the change to `basicQueryWorking`.

simonpasquier (approved):
thanks!

PR #6683, opened 2024-06-15
Replaced t.Fatal with require in Alertmanager Package

AshwinSriram11:
## Description
Continuation to #6659 and #6680. Alertmanager package was edited in this PR.

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

PR #6680, opened 2024-06-14
Replacing t.fatal with require Package 

AshwinSriram11:
## Description
Continuation to #6659 . In this PR, I have made changes to admission, assets, k8sutil and listwatch packages.

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

PR #6659, opened 2024-06-07
[Enhancement] Replacing t.Fatal with testify/require package 

AshwinSriram11:
## Description

Currently, there is an excessive usage of [t.Fatal ](https://github.com/search?q=repo%3Aprometheus-operator%2Fprometheus-operator+t.Fatal&type=code&p=3)in this repository. To make the code more readable, and to remove repetitive manual checks, I have replaced t.Fatal with functions of **testify/require package**. Currently, this PR only consists of changes applied in **pkg/operator**. 

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [x] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification
Since I made changes only in **pkg/operator**, I ran a test for this package using ```go test ./pkg/operator``` and received the following output -  ```ok      github.com/prometheus-operator/prometheus-operator/pkg/operator 0.065s```

PR #6688, opened 2024-06-18
WIP: Refactor common test code between Prometheus Agent's StatefulSet and DaemonSet modes

haanhvu:
## Description

Refactor common test code between Prometheus Agent's StatefulSet and DaemonSet modes to reduce duplication and make reviewing https://github.com/prometheus-operator/prometheus-operator/pull/6652 easier.
If seeing any pieces can be refactored further into the Server mode then do it too.
If seeing anywhere `testing` package is used when `require` and `assert` is the better option, make the change too.
If seeing any minor code-related improvement chance, do it too. 

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification
CI green

PR #6690, opened 2024-06-18
Replace t.Fatal with require in Prometheus package

AshwinSriram11:
## Description
Continuation to #6659, #6680 and #6683.

## Type of change

### DIFF
--- a/pkg/prometheus/agent/statefulset_test.go
+++ b/pkg/prometheus/agent/statefulset_test.go
@@ -85,13 +85,7 @@ func TestWALCompression(t *testing.T) {
 			}
 		}
 
-		if found != test.shouldContain {
-			if test.shouldContain {
-				t.Fatalf("expected Prometheus args to contain %v, but got %v", test.expectedArg, promArgs)
-			} else {
-				t.Fatalf("expected Prometheus args to NOT contain %v, but got %v", test.expectedArg, promArgs)
-			}
-		}
+		require.Equal(t, test.shouldContain, found)
 	}
 }
 
--- a/pkg/prometheus/agent/statefulset_test.go
+++ b/pkg/prometheus/agent/statefulset_test.go
@@ -84,7 +84,6 @@ func TestWALCompression(t *testing.T) {
 				break
 			}
 		}
-
 		require.Equal(t, test.shouldContain, found)
 	}
 }
--- a/pkg/prometheus/operator_test.go
+++ b/pkg/prometheus/operator_test.go
@@ -17,6 +17,7 @@ package prometheus
 import (
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/utils/ptr"
 
@@ -47,9 +48,7 @@ func TestKeyToStatefulSetKey(t *testing.T) {
 
 	for _, c := range cases {
 		got := KeyToStatefulSetKey(c.p, c.name, c.shard)
-		if c.expected != got {
-			t.Fatalf("Expected key %q got %q", c.expected, got)
-		}
+		require.Equal(t, c.expected, got, "Expected key %q got %q", c.expected, got)
 	}
 }
 
@@ -197,12 +196,11 @@ func TestValidateRemoteWriteConfig(t *testing.T) {
 		test := c
 		t.Run(test.name, func(t *testing.T) {
 			err := ValidateRemoteWriteSpec(test.spec)
-			if err != nil && !test.expectErr {
-				t.Fatalf("unexpected error occurred: %v", err)
+			if test.expectErr {
+				require.Error(t, err)
+				return
 			}
-			if err == nil && test.expectErr {
-				t.Fatalf("expected an error, got nil")
-			}
+			require.NoError(t, err)
 		})
 	}
 }
--- a/pkg/prometheus/resource_selector_test.go
+++ b/pkg/prometheus/resource_selector_test.go
@@ -420,12 +420,11 @@ func TestValidateRelabelConfig(t *testing.T) {
 	} {
 		t.Run(tc.scenario, func(t *testing.T) {
 			err := validateRelabelConfig(&tc.prometheus, tc.relabelConfig)
-			if err != nil && !tc.expectedErr {
-				t.Fatalf("expected no error, got: %v", err)
+			if tc.expectedErr {
+				require.Error(t, err)
+				return
 			}
-			if err == nil && tc.expectedErr {
-				t.Fatalf("expected an error, got nil")
-			}
+			require.NoError(t, err)
 		})
 	}
 }
@@ -765,12 +764,11 @@ func TestValidateScrapeIntervalAndTimeout(t *testing.T) {
 			for _, endpoint := range tc.smSpec.Endpoints {
 				err := validateScrapeIntervalAndTimeout(&tc.prometheus, endpoint.Interval, endpoint.ScrapeTimeout)
 				t.Logf("err %v", err)
-				if err != nil && !tc.expectedErr {
-					t.Fatalf("expected no error, got: %v", err)
+				if tc.expectedErr {
+					require.Error(t, err)
+					return
 				}
-				if err == nil && tc.expectedErr {
-					t.Fatalf("expected an error, got nil")
-				}
+				require.NoError(t, err)
 			}
 		})
 	}
--- a/pkg/prometheus/server/rules_test.go
+++ b/pkg/prometheus/server/rules_test.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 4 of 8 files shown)


pkg/prometheus/server/statefulset_test.go
```diff
@@ -1289,9 +1179,7 @@ func TestThanosObjectStorageFile(t *testing.T) {
 				}
 			}
 		}
-		if !found {
-			t.Fatalf("--tsdb.path argument should be given to the Thanos sidecar, got %q", strings.Join(sset.Spec.Template.Spec.Containers[3].Args, " "))
-		}
+		require.True(t, found, "--tsdb.path argument should be given to the Thanos sidecar, got %q", strings.Join(sset.Spec.Template.Spec.Containers[3].Args, " "))
```
AshwinSriram11, line 1182: I checked why the tests are failing and I got this error

```
--- FAIL: TestThanosObjectStorageFile (0.00s)
panic: runtime error: index out of range [3] with length 3 [recovered]
        panic: runtime error: index out of range [3] with length 3
```
When I changed it to ```sset.Spec.Template.Spec.Containers[2].Args```, it was working fine. Wdyt about this?

pkg/prometheus/agent/statefulset_test.go
```diff
@@ -181,11 +170,8 @@ func TestWALCompression(t *testing.T) {
 		}
 
 		if found != test.shouldContain {
-			if test.shouldContain {
-				t.Fatalf("expected Prometheus args to contain %v, but got %v", test.expectedArg, promArgs)
-			} else {
-				t.Fatalf("expected Prometheus args to NOT contain %v, but got %v", test.expectedArg, promArgs)
-			}
+			require.False(t, test.shouldContain, "expected Prometheus args to contain %v, but got %v", test.expectedArg, promArgs)
+			require.True(t, test.shouldContain, "expected Prometheus args to NOT contain %v, but got %v", test.expectedArg, promArgs)
```
AshwinSriram11, line 174: I feel this is not a good method. There are many other places with same code. Wdyt?

pkg/prometheus/operator_test.go
```diff
@@ -197,11 +196,11 @@ func TestValidateRemoteWriteConfig(t *testing.T) {
 		test := c
 		t.Run(test.name, func(t *testing.T) {
 			err := ValidateRemoteWriteSpec(test.spec)
-			if err != nil && !test.expectErr {
-				t.Fatalf("unexpected error occurred: %v", err)
+			if !test.expectErr {
+				require.NoError(t, err)
```
AshwinSriram11, line 200: This might be better. Wdyt?

```
require.False(t,(err!=nil && !test.expectErr),"unexpected error occurred: %v", err)
```

pkg/prometheus/agent/statefulset_test.go
AshwinSriram11, line 174: I feel this is not a good method. There are many other places with same code. Wdyt?

ArthurSens: What if we just use `require.Equal(t, test.shouldContain, found)`?

I think this could replace the whole if block 🤔

pkg/prometheus/operator_test.go
ArthurSens: I think it looks good the way it is now, we just don't need the second if because, if we get there, it's always true.

What do you think about the suggestion below?

```go

	for _, c := range cases {
		test := c
		t.Run(test.name, func(t *testing.T) {
			err := ValidateRemoteWriteSpec(test.spec)
			if test.expectErr {
				require.Error(t, err)
				return
			}
			require.NoError(t, err)
		})
	}
```

pkg/prometheus/resource_selector_test.go
```diff
@@ -420,11 +420,11 @@ func TestValidateRelabelConfig(t *testing.T) {
...
 			err := validateRelabelConfig(&tc.prometheus, tc.relabelConfig)
-			if err != nil && !tc.expectedErr {
-				t.Fatalf("expected no error, got: %v", err)
+			if !tc.expectedErr {
+				require.NoError(t, err)
 			}
-			if err == nil && tc.expectedErr {
-				t.Fatalf("expected an error, got nil")
+			if tc.expectedErr {
+				require.Error(t, err)
```
ArthurSens, line 427: It's a similar situation here, we can use early returns to not need so many ifs

```go

		t.Run(tc.scenario, func(t *testing.T) {
			err := validateRelabelConfig(&tc.prometheus, tc.relabelConfig)
			if tc.expectedErr {
				require.Error(t, err)
				return
			}
			require.NoError(t, err)
		})
```

pkg/prometheus/resource_selector_test.go
```diff
@@ -765,11 +765,11 @@ func TestValidateScrapeIntervalAndTimeout(t *testing.T) {
...
 				t.Logf("err %v", err)
-				if err != nil && !tc.expectedErr {
-					t.Fatalf("expected no error, got: %v", err)
+				if !tc.expectedErr {
+					require.NoError(t, err)
 				}
-				if err == nil && tc.expectedErr {
-					t.Fatalf("expected an error, got nil")
+				if tc.expectedErr {
+					require.Error(t, err)
```
ArthurSens, line 772: Same thing here :)

pkg/prometheus/server/rules_test.go
```diff
@@ -54,9 +51,8 @@ func shouldErrorOnTooLargeRuleFile(t *testing.T) {
 	ruleFiles["my-rule-file"] = strings.Repeat("a", v1.MaxSecretSize+1)
 
 	_, err := makeRulesConfigMaps(&monitoringv1.Prometheus{ObjectMeta: metav1.ObjectMeta{Name: "test"}}, ruleFiles)
-	if err == nil || err.Error() != expectedError {
-		t.Fatalf("expected makeRulesConfigMaps to return error '%v' but got '%v'", expectedError, err)
-	}
+	require.Error(t, err)
+	require.Equal(t, expectedError, err.Error(), "expected makeRulesConfigMaps to return error '%v' but got '%v'", expectedError, err)
```
ArthurSens, line 55:
```suggestion
	require.Equal(t, expectedError, err.Error(), "expected makeRulesConfigMaps to return error '%v' but got '%v'", expectedError, err)
```

If err is nil, the `require.Equal(...)` will catch it already, right?

ArthurSens (commented):
There was a big refactoring of the tests in #6688, which already replaced the t.Fatal appearances. Could you rebase your branch and forget about the file `pkg/prometheus/agent/statefulset_test.go`? :)

I've done an incomplete review, I'll try to go over it again at another time!

pkg/prometheus/server/statefulset_test.go
```diff
@@ -1288,9 +1174,7 @@ func TestThanosObjectStorageFile(t *testing.T) {
 				}
 			}
 		}
-		if !found {
-			t.Fatalf("--tsdb.path argument should be given to the Thanos sidecar, got %q", strings.Join(sset.Spec.Template.Spec.Containers[3].Args, " "))
-		}
+		require.True(t, found, "--tsdb.path argument should be given to the Thanos sidecar, got %q", strings.Join(sset.Spec.Template.Spec.Containers[2].Args, " "))
```
AshwinSriram11, line 1177: Same here too

> I changed the index from 3 to 2 because I received the following error locally on running the test
> 
> ```
> --- FAIL: TestThanosObjectStorageFile (0.00s)
> panic: runtime error: index out of range [3] with length 3 [recovered]
>         panic: runtime error: index out of range [3] with length 3
> ```
> After changing, it was running fine.

pkg/prometheus/server/statefulset_test.go
```diff
@@ -1538,21 +1405,8 @@ func TestRetentionAndRetentionSize(t *testing.T) {
...
-
-		if foundRetentionSize != test.shouldContainRetentionSize || foundRetentionSizeFlag != test.shouldContainRetentionSize {
-			if test.shouldContainRetentionSize {
-				t.Fatalf("test %d, expected Prometheus args to contain %v, but got %v", i, test.expectedRetentionSizeArg, promArgs)
-			} else {
-				t.Fatalf("test %d, expected Prometheus args to NOT contain %v, but got %v", i, test.expectedRetentionSizeArg, promArgs)
-			}
-		}
+		require.False(t, (foundRetention != test.shouldContainRetention || foundRetentionFlag != test.shouldContainRetention), "test %d, expected Prometheus args to %scontain %v, but got %v", i, map[bool]string{true: "", false: "NOT "}[test.shouldContainRetention], test.expectedRetentionArg, promArgs)
+		require.False(t, (foundRetentionSize != test.shouldContainRetentionSize || foundRetentionSizeFlag != test.shouldContainRetentionSize), "test %d, expected Prometheus args to %scontain %v, but got %v", i, map[bool]string{true: "", false: "NOT "}[test.shouldContainRetentionSize], test.expectedRetentionSizeArg, promArgs)
```
ArthurSens, line 1409: We don't necessarily need to keep the previous message with this refactoring. It would be even better to not add them at all I believe.

We added messages with t.Fatal because t.Fatal has no context whatsoever. The require package already adds a lot of context to the test, and emits nice human readable comparisons when the test fails.

We could simplify here with just:

```go
if test.shouldContainRetention {
      require.True(t, (foundRetention && foundRetentionFlag))
}

if test.shouldContainRetentionSize {
      require.True(t, (foundRetentionSize && foundRetentionSizeFlag))
}
```

ArthurSens (approved):
I've added just a few nitpicks, but generally LGTM :)

pkg/prometheus/server/statefulset_test.go
```diff
@@ -162,12 +151,11 @@ func TestPodLabelsAnnotations(t *testing.T) {
...
 	require.NoError(t, err)
 
-	if val, ok := sset.Spec.Template.ObjectMeta.Labels["testlabel"]; !ok || val != "testvalue" {
-		t.Fatal("Pod labels are not properly propagated")
-	}
-	if val, ok := sset.Spec.Template.ObjectMeta.Annotations["testannotation"]; !ok || val != "testvalue" {
-		t.Fatal("Pod annotations are not properly propagated")
-	}
+	valLabel, ok := sset.Spec.Template.ObjectMeta.Labels["testlabel"]
+	require.True(t, (ok && valLabel == "testvalue"), "Pod labels are not properly propagated")
```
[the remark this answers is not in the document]
AshwinSriram11: Yeah, didn't think of it. Nice catch.

ArthurSens (approved):
Good job, thanks!

PR #6694, opened 2024-06-19
Continue #6688: Refactor common test code between Prometheus modes

haanhvu:
## Description

Continues #6688. Do some changes that we agreed in Slack:

- Remove Thanos image in test config: https://cloud-native.slack.com/archives/C071STDT4FP/p1718722242845739
- Refactor `TestWALCompression` for both StatefulSet and DaemonSet: https://cloud-native.slack.com/archives/C071STDT4FP/p1718703164200639
- Refactor `TestStartupProbeTimeoutSeconds`: https://github.com/prometheus-operator/prometheus-operator/pull/6686#discussion_r1645711201
- Use some appropriate refactored functions in server mode

## Type of change

## Verification
CI green

### DIFF
--- a/pkg/prometheus/agent/statefulset_test.go
+++ b/pkg/prometheus/agent/statefulset_test.go
@@ -52,20 +52,39 @@ func TestListenTLS(t *testing.T) {
 }
 
 func TestWALCompression(t *testing.T) {
-	testcases := createTestCasesForTestWALCompression()
+	tests := []struct {
+		version       string
+		enabled       *bool
+		expectedArg   string
+		shouldContain bool
+	}{
+		// Nil should not have either flag.
+		{"v2.30.0", ptr.To(false), "--storage.agent.wal-compression", false},
+		{"v2.32.0", nil, "--storage.agent.wal-compression", false},
+		{"v2.32.0", ptr.To(false), "--no-storage.agent.wal-compression", true},
+		{"v2.32.0", ptr.To(true), "--storage.agent.wal-compression", true},
+	}
 
-	for _, test := range testcases {
-		sset, err := makeStatefulSetFromPrometheus(
-			makePrometheusAgentForTestWALCompression(
-				test.version,
-				test.enabled))
+	for _, test := range tests {
+		sset, err := makeStatefulSetFromPrometheus(monitoringv1alpha1.PrometheusAgent{
+			Spec: monitoringv1alpha1.PrometheusAgentSpec{
+				CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
+					Version:        test.version,
+					WALCompression: test.enabled,
+				},
+			},
+		})
 		require.NoError(t, err)
 
-		testPromArgsShouldContain(
-			t,
-			test.expectedArg,
-			sset.Spec.Template.Spec.Containers[0].Args,
-			test.shouldContain)
+		promArgs := sset.Spec.Template.Spec.Containers[0].Args
+		found := false
+		for _, flag := range promArgs {
+			if flag == test.expectedArg {
+				found = true
+				break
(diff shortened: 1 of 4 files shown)


pkg/prometheus/agent/statefulset_test.go
```diff
@@ -52,75 +52,29 @@ func TestListenTLS(t *testing.T) {
...
-		enabled       *bool
-		expectedArg   string
-		shouldContain bool
-	}{
-		// Nil should not have either flag.
-		{"v2.30.0", ptr.To(false), "--storage.agent.wal-compression", false},
-		{"v2.32.0", nil, "--storage.agent.wal-compression", false},
-		{"v2.32.0", ptr.To(false), "--no-storage.agent.wal-compression", true},
-		{"v2.32.0", ptr.To(true), "--storage.agent.wal-compression", true},
-	}
```
ArthurSens, line 66: Hmmmm, I'm assuming we made similar changes in other tests because the test cases array was taking waaaay more space than the test logic itself. For this test in particular it doesn't seem like it. Could you explain your thoughts when making the changes here?

pkg/prometheus/agent/statefulset_test.go
```diff
@@ -52,75 +52,29 @@ func TestListenTLS(t *testing.T) {
...
-					Version:        test.version,
-					WALCompression: test.enabled,
-				},
-			},
-		})
+	for _, test := range testcases {
+		sset, err := makeStatefulSetFromPrometheus(
+			makePrometheusAgentForTestWALCompression(
+				test.version,
+				test.enabled))
```
ArthurSens, line 61: same comment here... we're writing 10 lines for `makePrometheusAgentForTestWalCompression`, just to hide 2 lines here. Is it really worth it?

pkg/prometheus/agent/statefulset_test.go
```diff
@@ -52,75 +52,29 @@ func TestListenTLS(t *testing.T) {
...
-			}
-		}
-
-		if found != test.shouldContain {
-			if test.shouldContain {
-				t.Fatalf("expected Prometheus args to contain %v, but got %v", test.expectedArg, promArgs)
-			} else {
-				t.Fatalf("expected Prometheus args to NOT contain %v, but got %v", test.expectedArg, promArgs)
-			}
-		}
```
ArthurSens, line 94: The change you've made about this looks pretty good!
 
```go
require.Equal(t, shouldContain, found)
```
Looks a lot better than the nested ifs!

I'm just not sure about abstracting this into a separate function, for the same reasons above.

ArthurSens (commented):
I'm a bit confused with the goals here. The tests you're refactoring are super small, I'm not sure if creating tons of functions in test_utils is the way to go.

Ideally functions in `utils` are functions that are used several times across the code base, but `makePrometheusAgentForTestWalCompression` or `createTestCasesForTestWalCompression` looks very specific to a single test 🤔

haanhvu:
@ArthurSens As explained in Slack, right now these refactored functions are used only once, so the lines reduced maybe small. But later when daemonset code (or even later if we have other modes) use them, they will be used more than once, and the lines reduces will be bigger.

We can continue discussion here or in Slack, or even in today sync^^

haanhvu:
Some refactored functions created in #6688 can be used for server's tests too. I'll enable that in this PR.

haanhvu:
We discussed offline and agreed `TestWALCompression` doesn't change between StatefulSet and DaemonSet, so we don't need to refactor and test it again in DaemonSet.

We'll also look at other tests to see if which are also (not) needed to test again in DaemonSet.

ArthurSens (approved):
Thanks!

Issue #6682, opened 2024-06-14
liveness- & readiness probes for `config-reloader` container in prometheus stateful set fail when `listenLocal: true`


sdickhoven:
### Is there an existing issue for this?

- [X] I have searched the existing issues

## Description

hi 👋 

we use [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) v55.5.0 as well as v60.1.0

the former comes with prometheus-operator v0.70.0, the latter with v0.74.0

in both versions prometheus-operator creates a broken stateful set if the `Prometheus` resource is configured with `listenLocal: true`.

specifically, the liveness- & readiness probes for the `config-reloader` container are configured with

```yaml
httpGet:
  path: /healthz
  port: 8080
  scheme: HTTP
```

even though the container is also configured with

```yaml
args:
- --listen-address=127.0.0.1:8080
- ...
```

so the `config-reloader` container keeps getting murdered by kubernetes. 💥 

*without* `listenLocal: true`, the `config-reloader` container is configured with

```yaml
args:
- --listen-address=:8080
- ...
```

...in which case these liveness- & readiness probes work fine.

## Steps to Reproduce

create `Prometheus` resource with

```yaml
spec:
  listenLocal: true
```

## Expected Result

either when `listenLocal` is set to `true` or _always_, the liveness- & readiness probes for the `config-reloader` container should look something like this:

```yaml
exec:
  command:
  - sh
  - -c
  - if [ -x "$(command -v curl)" ]; then exec curl --fail http://localhost:8080/healthz;
    elif [ -x "$(command -v wget)" ]; then exec wget -q -O /dev/null http://localhost:8080/healthz;
    else exit 1; fi
```

...which is exactly what the `prometheus` container in the stateful set already uses (presumably to deal with precisely this issue).

## Actual Result

the liveness- & readiness probes for the `config-reloader` container are configured with

```yaml
httpGet:
  path: /healthz
  port: 8080
  scheme: HTTP
```

### Prometheus Operator Version

```shell
v0.70.0
v0.74.0
```

(presumably everything in between as well and likely also some previous versions :shrug:)

### Kubernetes Version

```yaml
clientVersion:
  buildDate: "2024-05-14T10:42:02Z"
  compiler: gc
  gitCommit: 6911225c3f747e1cd9d109c305436d08b668f086
  gitTreeState: clean
  gitVersion: v1.30.1
[...]

PR #6698, opened 2024-06-19
Create correct readiness/liveness probes for config-reloader when `listenLocal` is set to `true`

sdickhoven:
## Description

When the prometheus operator is started with `--enable-config-reloader-probes` it will now create `exec` probes that run curl/wget in the `config-reloader` container against localhost to check the /healthz endpoint if `listenLocal` is set to `true`.

Otherwise, it creates `httpGet` probes as before.

This change resolves https://github.com/prometheus-operator/prometheus-operator/issues/6682

## Type of change

## Verification

I ran unit tests and they all passed.

I also created a local [kind](https://kind.sigs.k8s.io/) cluster and modified the `scripts/run-external.sh` script to add the `--enable-config-reloader-probes` option for the operator here:

https://github.com/prometheus-operator/prometheus-operator/blob/8869970/scripts/run-external.sh#L128-L133

i then created a `Prometheus` resource with `listenLocal` set to both `true` and `false` (to ensure both branches of the new logic are exercised and result in the expected outcome).

the result in both cases was

```
$ kubectl get pod --watch
NAMESPACE   NAME                READY   STATUS     RESTARTS   AGE
default     prometheus-test-0   0/2     Init:0/1   0          0s       
default     prometheus-test-0   0/2     PodInitializing   0          18s      
default     prometheus-test-0   0/2     Running           0          26s      
default     prometheus-test-0   0/2     Running           0          30s      
default     prometheus-test-0   1/2     Running           0          31s      
default     prometheus-test-0   2/2     Running           0          40s      
```

(i.e. the probes returned a healthy status)

the `config-reloader` probes created by the operator were (as expected)...

`listenLocal: true`:
```yaml
exec:
  command:
  - sh
  - -c
  - if [ -x "$(command -v curl)" ]; then exec curl --fail http://localhost:8080/healthz;
    elif [ -x "$(command -v wget)" ]; then exec wget -q -O /dev/null http://localhost:8080/healthz;
    else exit 1; fi
```

`listenLocal: false`:
```yaml
httpGet:
  path: /healthz
  port: 8080
  scheme: HTTP
```

## Changelog entry

```release-note
Generate correct liveness/readiness probes for config-reloader in Prometheus stateful set when listenLocal is set to true.
```

### DIFF
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -309,22 +309,39 @@ func CreateConfigReloader(name string, options ...ReloaderOption) v1.Container {
 	}
 
 	if !configReloader.initContainer && configReloader.config.EnableProbes {
-		c = addProbes(c)
+		c = addProbes(c, configReloader.listenLocal)
 	}
 
 	return c
 }
 
-func addProbes(c v1.Container) v1.Container {
-	probe := &v1.Probe{
-		ProbeHandler: v1.ProbeHandler{
-			HTTPGet: &v1.HTTPGetAction{
-				Path: path.Clean("/healthz"),
-				Port: intstr.FromInt(configReloaderPort),
+func addProbes(c v1.Container, listenLocal bool) v1.Container {
+	probePath := path.Clean("/healthz")
+	handler := v1.ProbeHandler{}
+	if listenLocal {
+		probeURL := url.URL{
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
Thanks!

PR #6703, opened 2024-06-20
chore: Replace StringPtrValOrDefault with ptr.Deref

ArthurSens:
## Description

While reviewing a few PRs from @haanhvu, I've noticed we're using a function that does the same thing as `ptr.Deref`. The reason is because that code was written before generics were a thing.

I don't mean to create conflicts with @haanhvu's PRs, so we can merge this one after we get yours merged first :)

## Type of change

## Verification

## Changelog entry

```release-note

```


haanhvu (approved):
> I don't mean to create conflicts with @haanhvu's PRs, so we can merge this one after we get yours merged first :)

We can merge this now. It could cause conflicts, but those would just be minor conflicts. It's not a good idea to let a smaller PR wait for a bigger PR.

ArthurSens:
> > I don't mean to create conflicts with @haanhvu's PRs, so we can merge this one after we get yours merged first :)
> 
> We can merge this now. It could cause conflicts, but those would just be minor conflicts. I don't think we should let a simpler PR wait for some more complicated PRs.

Cool, I'm merging then!

PR #6708, opened 2024-06-26
Allow Prometheus Agent operator to create DaemonSet

haanhvu:
## Description

Allow Prometheus Agent operator to create DaemonSet if DaemonSet mode is chosen.

## Type of change

## Verification
WIP: Local tests and e2e tests for sure, unit tests maybe

## Changelog entry

Allow Prometheus Agent operator to create DaemonSet

### DIFF
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -179,13 +179,13 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 	}
 }
 
-// DaemonSet sets the options that only work for DaemonSet mode.
-// Currently we set SHARD env equal to 1, eventhough DaemonSet doesn't use this env.
+// DaemonSet sets the options that work for DaemonSet mode.
+// Currently we set SHARD env equal to 0, eventhough DaemonSet doesn't use this env.
 // TODO: Remove SHARD env for DaemonSet mode.
-func DaemonSet() ReloaderOption {
+func WithDaemonSetMode() ReloaderOption {
 	return func(c *ConfigReloader) {
 		c.withNodeNameEnv = true
-		c.shard = ptr.To(int32(1))
+		c.shard = ptr.To(int32(0))
 	}
 }
 
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -62,7 +62,7 @@ type ConfigReloader struct {
 	volumeMounts       []v1.VolumeMount
 	watchedDirectories []string
 	useSignal          bool
-	withNodeNameEnv    bool
+	daemonSet          bool
 }
 
 type ReloaderOption = func(*ConfigReloader)
@@ -180,9 +180,9 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 }
 
 // WithNodeNameEnv sets the withNodeNameEnv option for the config-reloader container.
-func WithNodeNameEnv() ReloaderOption {
+func DaemonSet() ReloaderOption {
 	return func(c *ConfigReloader) {
-		c.withNodeNameEnv = true
+		c.daemonSet = true
 	}
 }
 
@@ -208,13 +208,14 @@ func CreateConfigReloader(name string, options ...ReloaderOption) v1.Container {
 		ports []v1.ContainerPort
 	)
 
-	if configReloader.withNodeNameEnv {
+	if configReloader.daemonSet {
 		envVars = append(envVars, v1.EnvVar{
 			Name: NodeNameEnvVar,
 			ValueFrom: &v1.EnvVarSource{
 				FieldRef: &v1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
 			},
 		})
+		configReloader.shard = nil
 	}
 
 	if configReloader.initContainer {
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -62,7 +62,7 @@ type ConfigReloader struct {
 	volumeMounts       []v1.VolumeMount
 	watchedDirectories []string
 	useSignal          bool
-	daemonSet          bool
+	withNodeNameEnv    bool
 }
 
 type ReloaderOption = func(*ConfigReloader)
@@ -180,9 +180,9 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 }
 
 // WithNodeNameEnv sets the withNodeNameEnv option for the config-reloader container.
-func DaemonSet() ReloaderOption {
+func WithNodeNameEnv() ReloaderOption {
 	return func(c *ConfigReloader) {
-		c.daemonSet = true
+		c.withNodeNameEnv = true
 	}
 }
 
@@ -208,14 +208,13 @@ func CreateConfigReloader(name string, options ...ReloaderOption) v1.Container {
 		ports []v1.ContainerPort
 	)
 
-	if configReloader.daemonSet {
+	if configReloader.withNodeNameEnv {
 		envVars = append(envVars, v1.EnvVar{
 			Name: NodeNameEnvVar,
 			ValueFrom: &v1.EnvVarSource{
 				FieldRef: &v1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
 			},
 		})
-		configReloader.shard = nil
 	}
 
 	if configReloader.initContainer {
--- a/pkg/operator/config_reloader.go
+++ b/pkg/operator/config_reloader.go
@@ -179,10 +179,13 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 	}
 }
 
-// WithNodeNameEnv sets the withNodeNameEnv option for the config-reloader container.
-func WithNodeNameEnv() ReloaderOption {
+// DaemonSet sets the options that only work for DaemonSet mode.
+// Currently we set SHARD env equal to 1, eventhough DaemonSet doesn't use this env.
+// TODO: Remove SHARD env for DaemonSet mode.
+func DaemonSet() ReloaderOption {
 	return func(c *ConfigReloader) {
 		c.withNodeNameEnv = true
+		c.shard = ptr.To(int32(1))
 	}
 }
 
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -154,6 +154,7 @@ func makeDaemonSetSpec(
 			true,
 			configReloaderVolumeMounts,
 			watchedDirectories,
+			operator.WithDaemonSetMode(),
 		),
 	)
 
@@ -195,7 +196,7 @@ func makeDaemonSetSpec(
 			configReloaderVolumeMounts,
 			watchedDirectories,
 			operator.WebConfigFile(configReloaderWebConfigFile),
-			operator.DaemonSet(),
+			operator.WithDaemonSetMode(),
 		),
 	}, additionalContainers...)
 
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -195,8 +195,7 @@ func makeDaemonSetSpec(
 			configReloaderVolumeMounts,
 			watchedDirectories,
 			operator.WebConfigFile(configReloaderWebConfigFile),
-			// DaemonSet needs NODE_NAME env to filter targes on the same node.
-			operator.WithNodeNameEnv(),
+			operator.DaemonSet(),
 		),
 	}, additionalContainers...)
 
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -30,7 +30,6 @@ import (
 )
 
 func makeDaemonSet(
-	name string,
 	p monitoringv1.PrometheusInterface,
 	config *prompkg.Config,
 	cg *prompkg.ConfigGenerator,
@@ -55,7 +54,7 @@ func makeDaemonSet(
 
 	operator.UpdateObject(
 		daemonSet,
-		operator.WithName(name),
+		operator.WithName("agent-daemonset"),
 		operator.WithAnnotations(objMeta.GetAnnotations()),
 		operator.WithAnnotations(config.Annotations),
 		operator.WithLabels(objMeta.GetLabels()),
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -196,7 +196,7 @@ func makeDaemonSetSpec(
 			watchedDirectories,
 			operator.WebConfigFile(configReloaderWebConfigFile),
 			// DaemonSet needs NODE_NAME env to filter targes on the same node.
-			operator.WithNodeNameEnv(),
+			operator.DaemonSet(),
 		),
 	}, additionalContainers...)
 
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -102,7 +102,7 @@ func makeDaemonSetSpec(
 
 	promArgs := buildAgentArgs(cpf, cg)
 
-	volumes, promVolumeMounts, err := prompkg.BuildCommonVolumes(p, tlsSecrets)
+	volumes, promVolumeMounts, err := prompkg.BuildCommonVolumes(p, tlsSecrets, false)
 	if err != nil {
 		return nil, err
 	}
@@ -126,7 +126,7 @@ func makeDaemonSetSpec(
 	startupProbe, readinessProbe, livenessProbe := prompkg.MakeProbes(cpf, cg)
 
 	podAnnotations, podLabels := prompkg.BuildPodMetadata(cpf, cg)
-	// In cases where an existing selector label is modified, or a new one is added, new sts cannot match existing pods.
+	// In cases where an existing selector label is modified, or a new one is added, new daemonset cannot match existing pods.
 	// We should try to avoid removing such immutable fields whenever possible since doing
 	// so forces us to enter the 'recreate cycle' and can potentially lead to downtime.
 	// The requirement to make a change here should be carefully evaluated.
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -54,7 +54,7 @@ func makeDaemonSet(
 
 	operator.UpdateObject(
 		daemonSet,
-		operator.WithName("agent-daemonset"),
+		operator.WithName(prompkg.PrefixedName(p)),
 		operator.WithAnnotations(objMeta.GetAnnotations()),
 		operator.WithAnnotations(config.Annotations),
 		operator.WithLabels(objMeta.GetLabels()),
--- a/pkg/prometheus/agent/daemonset.go
+++ b/pkg/prometheus/agent/daemonset.go
@@ -196,7 +196,7 @@ func makeDaemonSetSpec(
 			watchedDirectories,
 			operator.WebConfigFile(configReloaderWebConfigFile),
 			// DaemonSet needs NODE_NAME env to filter targes on the same node.
-			operator.DaemonSet(),
+			operator.WithNodeNameEnv(),
 		),
 	}, additionalContainers...)
 
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -603,10 +603,9 @@ func (c *Operator) Sync(ctx context.Context, key string) error {
 	p = p.DeepCopy()
 	if p.Spec.Mode != nil && *p.Spec.Mode == "DaemonSet" {
 		err = c.syncDaemonSet(ctx, key, p)
-		c.reconciliations.SetStatus(key, err)
-		return err
+	} else {
+		err = c.syncStatefulSet(ctx, key, p)
 	}
-	err = c.syncStatefulSet(ctx, key, p)
 	c.reconciliations.SetStatus(key, err)
 	return err
 }
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -587,6 +587,7 @@ func statefulSetKeyToPrometheusAgentKey(key string) (bool, string) {
 }
 
 // Sync implements the operator.Syncer interface.
+// TODO: Consider refactoring the common code between syncDaemonSet() and syncStatefulSet().
 func (c *Operator) Sync(ctx context.Context, key string) error {
 	pobj, err := c.promInfs.Get(key)
 
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -659,7 +659,6 @@ func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv
 	}
 
 	dset, err := makeDaemonSet(
-		"daemonset",
 		p,
 		&c.config,
 		cg,
@@ -701,7 +700,6 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
 
 	level.Info(logger).Log("msg", "sync prometheus")
 
-	// 1: relate
 	if err := operator.CheckStorageClass(ctx, c.canReadStorageClass, c.kclient, p.Spec.Storage); err != nil {
 		return err
 	}
@@ -725,17 +723,14 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
 		return fmt.Errorf("synchronizing web config secret failed: %w", err)
 	}
 
-	// 2: relate
 	// Create governing service if it doesn't exist.
 	svcClient := c.kclient.CoreV1().Services(p.Namespace)
 	if err := k8sutil.CreateOrUpdateService(ctx, svcClient, makeStatefulSetService(p, c.config)); err != nil {
 		return fmt.Errorf("synchronizing governing service failed: %w", err)
 	}
 
-	// 3: relate
 	ssetClient := c.kclient.AppsV1().StatefulSets(p.Namespace)
 
-	// 4: relate
 	// Ensure we have a StatefulSet running Prometheus Agent deployed and that StatefulSet names are created correctly.
 	expected := prompkg.ExpectedStatefulSetShardNames(p)
 	for shard, ssetName := range expected {
@@ -827,13 +822,11 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
 		}
 	}
 
-	// 5: relate
 	ssets := map[string]struct{}{}
 	for _, ssetName := range expected {
 		ssets[ssetName] = struct{}{}
 	}
 
-	// 6: relate
 	err = c.ssetInfs.ListAllByNamespace(p.Namespace, labels.SelectorFromSet(labels.Set{prompkg.PrometheusNameLabelName: p.Name, prompkg.PrometheusModeLabeLName: prometheusMode}), func(obj interface{}) {
 		s := obj.(*appsv1.StatefulSet)
 
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -652,7 +652,7 @@ func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv
 
 	level.Debug(logger).Log("msg", "reconciling daemonset")
 
-	_, err = c.dsetInfs.Get(prompkg.KeyToDaemonSetKey(p, key))
+	_, err = c.dsetInfs.Get(keyToDaemonSetKey(p, key))
 	exists := !apierrors.IsNotFound(err)
 	if err != nil && !apierrors.IsNotFound(err) {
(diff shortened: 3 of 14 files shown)


haanhvu:
@ArthurSens @simonpasquier @kakkoyun This PR contains PR #6652 I'm too impatient to wait for #6652 merged, so copied that PR directly in this PR to allow prometheus agent operator to create daemonset.

I'll let you know when this's ready for local tests.

haanhvu:
@ArthurSens @simonpasquier @kakkoyun I tried locally with a simple CRD:
```
kind: PrometheusAgent
metadata:
  name: prometheus-agent
spec:
  mode: DaemonSet

```

and got this error:
`level=error ts=2024-06-26T11:00:05.311030089Z caller=klog.go:126 component=k8s_client_runtime func=ErrorDepth msg="sync \"default/prometheus-agent\" failed: creating daemonset failed: DaemonSet.apps \"daemonset\" is invalid: spec.template.spec.containers[0].volumeMounts[2].name: Not found: \"prom-agent-prometheus-agent-db\""`

pkg/prometheus/agent/operator.go
```diff
@@ -282,6 +283,20 @@ func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger
 		return nil, fmt.Errorf("error creating statefulset informers: %w", err)
 	}
 
+	o.dsetInfs, err = informers.NewInformersForResource(
```
simonpasquier, line 286: we need to check if the feature gate is enabled or not

```suggestion
	if c.Gates.Enabled(PrometheusAgentDaemonSetFeature) {
		o.dsetInfs, err = informers.NewInformersForResource(
```

pkg/prometheus/agent/operator.go
```diff
@@ -352,6 +368,7 @@ func (c *Operator) Run(ctx context.Context) error {
 	go c.cmapInfs.Start(ctx.Done())
 	go c.secrInfs.Start(ctx.Done())
 	go c.ssetInfs.Start(ctx.Done())
+	go c.dsetInfs.Start(ctx.Done())
```
simonpasquier, line 371: check for nil

simonpasquier (commented):
The main() function also needs to check that if the feature gate is enabled, the operator has the proper RBAC permissions on the Daemonset resource (like we do for PrometheusAgent for instance).

https://github.com/prometheus-operator/prometheus-operator/blob/680654c9d110ab90c214f3185ca85926f1c72f6f/cmd/operator/main.go#L360-L379

It can be done in a separate PR (before this one?).

The ResourceReconciler also needs to support Daemonset objects in addition to StatefulSet objects but I'm not sure what's the best approach.

https://github.com/prometheus-operator/prometheus-operator/blob/680654c9d110ab90c214f3185ca85926f1c72f6f/pkg/operator/resource_reconciler.go#L38-L46

pkg/prometheus/agent/operator.go
```diff
@@ -440,6 +458,8 @@ func (c *Operator) addHandlers() {
 
 	c.ssetInfs.AddEventHandler(c.rr)
 
+	c.dsetInfs.AddEventHandler(c.rr)
```
simonpasquier, line 461:
```suggestion
	// TODO: implement proper event handler for Daemonset.
	// c.dsetInfs.AddEventHandler(c.rr)
```

haanhvu:
Existing e2e run green now. No more failures. I'll local test tomorrow. @ArthurSens @simonpasquier @kakkoyun pls help local test also when you have time.

pkg/prometheus/agent/operator.go
```diff
@@ -564,6 +581,82 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	level.Debug(logger).Log("msg", "reconciling daemonset")
+
+	_, err = c.dsetInfs.Get(prompkg.KeyToDaemonSetKey(p, key))
+	exists := !apierrors.IsNotFound(err)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("retrieving daemonset failed: %w", err)
+	}
+
+	dset, err := makeDaemonSet(
+		"daemonset",
```
ArthurSens, line 638:
```suggestion
		p.Name,
```

Could you explain why we're using "daemonset" as the name here?

ArthurSens: For the statefulset we needed extra logix to handle shard names, but looks like we don't need that for daemonsets. Would it make sense to just remove this argument from `makeDaemonset`?

ArthurSens (commented):
I've done a quick test locally, after the change I've mentioned below, I still get some errors:

```
level=error ts=2024-06-30T15:07:36.17385Z caller=klog.go:126 component=k8s_client_runtime func=ErrorDepth msg="sync \"default/prometheus-agent\" failed: creating daemonset failed: DaemonSet.apps \"daemonset\" is invalid: spec.template.spec.containers[0].volumeMounts[2].name: Not found: \"prom-agent-prometheus-agent-db\""
```

haanhvu:
> I've done a quick test locally, after the change I've mentioned below, I still get some errors:
> 
> ```
> level=error ts=2024-06-30T15:07:36.17385Z caller=klog.go:126 component=k8s_client_runtime func=ErrorDepth msg="sync \"default/prometheus-agent\" failed: creating daemonset failed: DaemonSet.apps \"daemonset\" is invalid: spec.template.spec.containers[0].volumeMounts[2].name: Not found: \"prom-agent-prometheus-agent-db\""
> ```

Thanks I got this too. Let me trace :feet:

haanhvu:
> > I've done a quick test locally, after the change I've mentioned below, I still get some errors:
> > ```
> > level=error ts=2024-06-30T15:07:36.17385Z caller=klog.go:126 component=k8s_client_runtime func=ErrorDepth msg="sync \"default/prometheus-agent\" failed: creating daemonset failed: DaemonSet.apps \"daemonset\" is invalid: spec.template.spec.containers[0].volumeMounts[2].name: Not found: \"prom-agent-prometheus-agent-db\""
> > ```
> 
> Thanks I got this too. Let me trace 🐾

@ArthurSens @simonpasquier @kakkoyun The not found volumemount in the error is this one:
https://github.com/prometheus-operator/prometheus-operator/blob/22a0a9b49b5580c15607475cb52dbd69ebf2ba56/pkg/prometheus/common.go#L292-L294

haanhvu:
> > > I've done a quick test locally, after the change I've mentioned below, I still get some errors:
> > > ```
> > > level=error ts=2024-06-30T15:07:36.17385Z caller=klog.go:126 component=k8s_client_runtime func=ErrorDepth msg="sync \"default/prometheus-agent\" failed: creating daemonset failed: DaemonSet.apps \"daemonset\" is invalid: spec.template.spec.containers[0].volumeMounts[2].name: Not found: \"prom-agent-prometheus-agent-db\""
> > > ```
> > 
> > 
> > Thanks I got this too. Let me trace 🐾
> 
> @ArthurSens @simonpasquier @kakkoyun The not found volumemount in the error is this one:
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/22a0a9b49b5580c15607475cb52dbd69ebf2ba56/pkg/prometheus/common.go#L292-L294
> 
> DaemonSet doesn't support storage, so we can remove this volumemount, right?

@ArthurSens @simonpasquier @kakkoyun When I removed this volumemount, the error was gone and seemed like daemonset was created successfully?
Screenshot from 2024-07-01 18-09-36

haanhvu:
The tests are failing because I commented the logic unneeded for DaemonSet, but needed for StatefulSet. I'll uncommented in next commit.

pkg/prometheus/agent/operator.go
```diff
@@ -544,13 +568,6 @@ func statefulSetKeyToPrometheusAgentKey(key string) (bool, string) {
 
 // Sync implements the operator.Syncer interface.
 func (c *Operator) Sync(ctx context.Context, key string) error {
-	err := c.sync(ctx, key)
-	c.reconciliations.SetStatus(key, err)
-
-	return err
-}
-
-func (c *Operator) sync(ctx context.Context, key string) error {
```
simonpasquier, line 553: I would keep the `sync()` method and only branch when the operator updates the statefulsets. What do you think?

pkg/prometheus/agent/operator.go
```diff
@@ -564,6 +581,84 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		err = c.syncDaemonSet(ctx, key, p)
+		c.reconciliations.SetStatus(key, err)
+		return err
+	}
+	err = c.syncStatefulSet(ctx, key, p)
+	c.reconciliations.SetStatus(key, err)
+	return err
+}
+
+func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv1alpha1.PrometheusAgent) error {
```
simonpasquier, line 594: the first thing to verify should be that the feature gate is enabled and return an error if not.

pkg/prometheus/agent/operator.go
haanhvu: You mean keeping the creating logic in sync() for both DaemonSet and StatefulSet? But their creating logics are not entirely similar. For example, for StatefulSet we need to check storageclass and handle sharding before creating.

In the end, we can surely refactor the common operator logics between two modes, whether in creating, updating or deleting. But maybe we can do that later, when we already have a working implmentation of DaemonSet?

For #6652 I understand we needed to do some refactoring before it, because it would have been harder to review and verify without the refactoring. But for operator's logic we can easily verify with local tests and e2e tests, so I don't think refactoring is a priority at the moment?

simonpasquier: I assume that we can still optimize a bit (e.g. reduce duplication). In particular everything which isn't related to statefulset/daemonset should be common?

haanhvu: Yeah we can still do it. Tbh operator's logic is still somewhat cryptic for me, so I just want to focus on making it work for DaemonSet first. But if there're any parts that seems obvious and easy to refactor, I'll do it then^^

haanhvu:
@kakkoyun @ArthurSens @simonpasquier I just added an e2e test for creating DaemonSet. It passed locally for me.

haanhvu:
TODO: Only run StatefulSet code (for example, StatefulSet informers) when StatefulSet mode is used. For now, it runs even when DaemonSet mode is used.

test/framework/prometheusagent.go
```diff
@@ -66,21 +69,92 @@ func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas in
...
 	if err != nil {
-		return nil, fmt.Errorf("waiting for %v prometheus-agent instances timed out (%v): %v", p.Spec.Replicas, p.Name, err)
+		return err
+	}
+
+	var pollErr error
+	if err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 5*time.Minute, false, func(ctx context.Context) (bool, error) {
+		dmsList, err := f.KubeClient.AppsV1().DaemonSets(ns).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			pollErr = fmt.Errorf("failed to list DaemonSet: %w", err)
```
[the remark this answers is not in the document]
haanhvu: @ArthurSens Nah still hitting the wall :sweat_smile: I set the timeout in ListOptions here: https://github.com/prometheus-operator/prometheus-operator/pull/6708/files#diff-537af06bf2777ce63bf6df494158f0cbfb50ac7809137f6ac68c4d9c4d68bfddR113 but somehow this error persists.

Pls help me with this if you have time :smile: I'll continue investigating tmr.

ArthurSens:
> TODO: Only run StatefulSet code (for example, StatefulSet informers) when StatefulSet mode is used. For now, it runs even when DaemonSet mode is used.

I'm not sure we want to do that... We have this conditional for Daemonset Informers because daemonset mode is currently experimental and users need to opt-in for the future. It doesn't make sense to run informers when people don't opt-in. That's not the case for Statefulsets, it's a stable feature and not opt-in... this means we always want to run StatefulSet informers 🤔

test/framework/prometheusagent.go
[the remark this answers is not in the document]
ArthurSens: Ah damn, the client rate limit problem is often flakiness. We hit that from time to time. One option is to make requests less often, e.g. by increasing the interval used in the poll

haanhvu:
> > TODO: Only run StatefulSet code (for example, StatefulSet informers) when StatefulSet mode is used. For now, it runs even when DaemonSet mode is used.
> 
> I'm not sure we want to do that... We have this conditional for Daemonset Informers because daemonset mode is currently experimental and users need to opt-in for the future. It doesn't make sense to run informers when people don't opt-in. That's not the case for Statefulsets, it's a stable feature and not opt-in... this means we always want to run StatefulSet informers 🤔

But in case PA runs in DaemonSet mode, we don't need to activate StatefulSet informers, right? 🤔

test/framework/prometheusagent.go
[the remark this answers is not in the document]
haanhvu: Currently interval is 5s. I'll increase it to 10s and see how it goes.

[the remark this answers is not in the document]
haanhvu: Increasing timeout didn't help. So I changed from `PollUntilContextTimeout()` to `PollUntilContextCancel()`.

The test doesn't fail anymore. But now it runs forever...

ArthurSens:
> But in case PA runs in DaemonSet mode, we don't need to activate StatefulSet informers, right? 🤔

I think there's some misunderstanding about where/why informers run 🤔.

The informers are a kind of cache that runs on the operator's side, and one operator is responsible for managing one or several prometheus/prometheus-agent/thanos-ruler/alertmanager.

If a Prometheus-Operator container happens to manage one single Prometheus-Agent and this agent happens to run in daemonset mode, your statement is true... only the daemonset informer is needed and the statefulset informer is stale. But we can't predict that! What if someone deploys another PrometheusAgent resource and this time in statefulset mode? The informer needs to be running so we can react to that case

pkg/prometheus/agent/operator.go
haanhvu: Now took a deeper look into `syncDaemonSet()` and `syncStatefulSet()`, I think we have refactored as far as we can. Yes there's still some common codes left. But these common codes do different things, so I'm not sure if we shoudl refactor them further into one function.

@ArthurSens Could you check and see if you agree with this?

@simonpasquier Even if we decide to merge without refactoring further now, you can always take another review when you come back from vacation^^

test/framework/prometheusagent.go
```diff
@@ -66,17 +69,78 @@ func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas in
...
+		}
+		/*dmsList, err := f.KubeClient.AppsV1().DaemonSets(ns).List(ctx, metav1.ListOptions{LabelSelector: selector.String()})
+		if err != nil {
+			pollErr = fmt.Errorf("failed to list DaemonSet: %w", err)
+			return false, nil
+		}
+		if len(dmsList.Items) != 1 {
+			pollErr = fmt.Errorf("expected: 1 DaemonSet, actual: %d", len(dmsList.Items))
+			return false, nil
+		}*/
```
haanhvu, line 128: @ArthurSens In the last meeting with Simon, we agreed to `List()` daemonset for e2e, similar to how you `List()` statefulset in https://github.com/prometheus-operator/prometheus-operator/pull/6198/files#diff-b3d3450b51af206bb0009c2b9c2f863c3d03a3575d64cbdb18510a2bc834144aR5108

But then I rethought `Get()` daemonset we want would be more suitable here. So I decided to do that.

Do you agree with this? I'm still commenting the previous `List()` approach (not deleted yet), in case you want to go back to it.

pkg/prometheus/agent/daemonset.go
```diff
@@ -55,7 +54,7 @@ func makeDaemonSet(
 
 	operator.UpdateObject(
 		daemonSet,
-		operator.WithName(name),
+		operator.WithName("agent-daemonset"),
```
ArthurSens, line 57: Could you clarify what is the goal here?

If a daemonset has a hardcoded name, it will be impossible to deploy 2 daemonsets in the same namespace 🤔

test/framework/prometheusagent.go
```diff
@@ -66,17 +69,78 @@ func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas in
...
+						v1.ResourceMemory: resource.MustParse("400Mi"),
+					},
+				},
+			},
+		},
 	}
+}
 
-	result, err = f.WaitForPrometheusAgentReady(ctx, result, 5*time.Minute)
+func (f *Framework) CreatePrometheusAgentDS(ctx context.Context, ns string, p *monitoringv1alpha1.PrometheusAgent) error {
```
ArthurSens, line 94:
```suggestion
func (f *Framework) CreatePrometheusAgentDSAndWaitUntilReady(ctx context.Context, ns string, p *monitoringv1alpha1.PrometheusAgent) error {
```

ArthurSens (commented):
Nice work! I think we're almost there. I've added just some small comments and a question about the daemonset name 🤔

pkg/prometheus/agent/daemonset.go
haanhvu: Maybe `agent-daemonset-$NODE_NAME`? We already can retrieve `NODE_NAME` from config-reloader: https://github.com/prometheus-operator/prometheus-operator/pull/6636

pkg/prometheus/agent/operator.go
haanhvu: Yeah, this code block is a duplication. But it does different things here. Should we really refactor them into _one_ function?

Maybe one function `beforeCreatingAgentClient()`? :thinking:

test/framework/prometheusagent.go
haanhvu: Yeah I'll add that comment

pkg/prometheus/agent/daemonset.go
ArthurSens: Why not re-use the agent name from the PrometheusAgent resource, like we do with statefulset?

pkg/prometheus/agent/operator.go
haanhvu: @simonpasquier wdyt^^

pkg/prometheus/agent/daemonset.go
simonpasquier: agreed with Arthur, we need a non-unique name. It should be possible to deploy 2 daemonset agents on the same cluster (though it's a niche scenario). IIUC it's also inconsistent with the logic of keyToDaemonSetKey(), meaning that the operator can't link back to the DaemonSet object from a PrometheusAgent name.

pkg/prometheus/agent/operator.go
```diff
@@ -564,6 +585,86 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 
 	p := pobj.(*monitoringv1alpha1.PrometheusAgent)
 	p = p.DeepCopy()
+	if p.Spec.Mode != nil && *p.Spec.Mode == "DaemonSet" {
```
simonpasquier, line 588: (nit)
```suggestion
	if ptr.Deref(p.Spec.Mode, "StatefulSet") == "DaemonSet" {
```

pkg/prometheus/agent/operator.go
simonpasquier: Let's create an issue to revisit potential refactoring after this PR merges.

pkg/prometheus/agent/operator.go
```diff
@@ -564,6 +585,86 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		err = c.syncDaemonSet(ctx, key, p)
+	} else {
+		err = c.syncStatefulSet(ctx, key, p)
+	}
+	c.reconciliations.SetStatus(key, err)
+	return err
+}
+
+func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv1alpha1.PrometheusAgent) error {
+	if !c.daemonSetFeatureGateEnabled {
```
simonpasquier, line 598: After more thinking I wonder if we shouldn't fallback to "statefulset" sync when the feature gate isn't enabled instead of failing?

simonpasquier (commented):
Good work! I tried locally and the pod can't reach the ready state because the SHARD env variable is undefined. Either we hardcode the SHARD var to 1 or we improve the config generation code to deal with this situation (the first option is probably simpler for now).

pkg/prometheus/agent/operator.go
haanhvu: Hm, you mean if I set `mode: daemonSet`, but forgot to enable feature gate, I'll have statefulset? :thinking:

haanhvu:
> Good work! I tried locally and the pod can't reach the ready state because the SHARD env variable is undefined. Either we hardcode the SHARD var to 1 or we improve the config generation code to deal with this situation (the first option is probably simpler for now).

@simonpasquier Hm, I didn't get this when testing locally. @ArthurSens didn't either, right? Is it because the SHARD env may already be set in a previous local setup for statefulset, that we forgot to clean up before creating daemonset?

haanhvu:
> > Good work! I tried locally and the pod can't reach the ready state because the SHARD env variable is undefined.
> 
> @simonpasquier Hm, I didn't get this when testing locally. @ArthurSens didn't either, right? Is it because the SHARD env may already be set in a previous local setup for statefulset, that we forgot to clean up before deploying daemonset?
> 
> > Either we hardcode the SHARD var to 1 or we improve the config generation code to deal with this situation (the first option is probably simpler for now).
> 
> I think we can remove SHARD var in config-reloader for daemonset, similar to how we added the NODE_NAME env? IIRC the config-reloader code it's not complicated to do. Let me try.

@simonpasquier @ArthurSens  I fixed this in last commit. Could you local test again?

haanhvu:
@simonpasquier @ArthurSens I resolved all reviews. Pls take another review. Thanks!

pkg/prometheus/agent/operator.go
simonpasquier: Yes. WDYT?

simonpasquier:
hmm I'm still getting the error with the SHARD env var being unset. Output of the `kubectl describe pod ...` command:

````
    Last State:  Terminated                                                                    
      Reason:    Error                                                                         
      Message:   level=info ts=2024-07-25T12:41:43.520299622Z caller=main.go:142 msg="Starting prometheus-config-reloader" version="(version=0.75.2, branch=refs/tags/v0.75.2, revision=35b0f4
57d8f705f1ac8282b4e22caa68a19c6c43)"                                                           
level=info ts=2024-07-25T12:41:43.520370773Z caller=main.go:143 build_context="(go=go1.22.5, platform=linux/amd64, user=Action-Run-ID-10070178014, date=20240724-04:04:09, tags=unknown)"
level=info ts=2024-07-25T12:41:43.719565567Z caller=cpu.go:28 msg="Updating GOMAXPROCS=1: using minimum allowed GOMAXPROCS"
level=info ts=2024-07-25T12:41:44.019381172Z caller=reloader.go:270 msg="reloading via HTTP"   
level=error ts=2024-07-25T12:41:44.019710256Z caller=main.go:223 msg="Failed to run" err="expand environment variables: found reference to unset environment variable \"SHARD\""
````

File: pkg/prometheus/agent/operator.go (hunk 12 above)
Comment (ArthurSens) [reply pkg/prometheus/agent/operator.go:595]: I not a fan of the idea 😬

If someone explicitly tells the operator to run a daemonset, see no errors but see statefulset instead... that would be very confusing

Comment (simonpasquier) [reply pkg/prometheus/agent/operator.go:595]: We've discussed offline and agreed that failing would be better (more explicit).

File: pkg/prometheus/agent/operator.go (hunk 5 above)
Comment (haanhvu) [reply pkg/prometheus/agent/operator.go:569]: I didn't create an issue. Instead I created a TODO. All the things we wish to do later have been added a TODO in this PR. Will that work too?

**Comment by haanhvu** [comment]
@ArthurSens @simonpasquier I resolved all reviews. Pls help take another review and local test. Thanks!

File: pkg/operator/config_reloader.go [hunk 13]
```diff
@@ -179,10 +179,13 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
...
 
-// WithNodeNameEnv sets the withNodeNameEnv option for the config-reloader container.
-func WithNodeNameEnv() ReloaderOption {
+// DaemonSet sets the options that only work for DaemonSet mode.
+// Currently we set SHARD env equal to 1, eventhough DaemonSet doesn't use this env.
+// TODO: Remove SHARD env for DaemonSet mode.
+func DaemonSet() ReloaderOption {
 	return func(c *ConfigReloader) {
 		c.withNodeNameEnv = true
+		c.shard = ptr.To(int32(1))
```
simonpasquier, line 188: This needs to be zero because the shards are 0-indexed (otherwise all targets are dropped).
```suggestion
		c.shard = ptr.To(int32(0))
```

test/framework/prometheusagent.go
```diff
@@ -66,17 +69,62 @@ func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas in
...
+
+		return true, nil
+	}); err != nil {
+		return fmt.Errorf("%v: %w", pollErr, err)
 	}
 
+	return nil
+}
+
+func (f *Framework) CreatePrometheusAgentAndWaitUntilReady(ctx context.Context, t *testing.T, ns string, p *monitoringv1alpha1.PrometheusAgent) (*monitoringv1alpha1.PrometheusAgent, error) {
```
simonpasquier, line 122: Framework methods shouldn't take testing.T argument. If something bad happens, they need to return an error (possibly wrapped).

test/framework/prometheusagent.go
```diff
@@ -66,17 +69,62 @@ func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas in
...
+	_, err := f.MonClientV1alpha1.PrometheusAgents(ns).Create(ctx, p, metav1.CreateOptions{})
 	if err != nil {
-		return nil, fmt.Errorf("waiting for %v prometheus-agent instances timed out (%v): %v", p.Spec.Replicas, p.Name, err)
+		return fmt.Errorf("failed to create Prometheus Agent DaemonSet: %w", err)
+	}
+
+	var pollErr error
+	if err := wait.PollUntilContextTimeout(ctx, 30*time.Second, 30*time.Minute, true, func(ctx context.Context) (bool, error) {
+		name := fmt.Sprintf("prom-agent-%s", p.Name)
+		// TODO: Implement UpdateStatus() for DaemonSet and check status instead of using Get().
```
simonpasquier, line 103: Until we have a proper status subresource for the DaemonSet mode and in addition to the check on the existence of the DaemonSet object, the test should also verify the status of the daemonset (it could be numberUnavailable == 0 and numberReady > 0).

test/framework/prometheusagent.go
```diff
@@ -66,17 +69,62 @@ func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas in
...
+						v1.ResourceMemory: resource.MustParse("400Mi"),
+					},
+				},
+			},
+		},
 	}
+}
 
-	result, err = f.WaitForPrometheusAgentReady(ctx, result, 5*time.Minute)
+func (f *Framework) CreatePrometheusAgentDSAndWaitUntilReady(ctx context.Context, ns string, p *monitoringv1alpha1.PrometheusAgent) error {
```
simonpasquier, line 94: instead of creating a new specific function for daemonset, we could implement the logic in WaitForPrometheusAgentReady():
* If .spec.mode == ds, get the DaemonSet object.
* Otherwise check the PrometheusAgent status subresource

test/framework/prometheusagent.go
```diff
@@ -66,17 +69,62 @@ func (f *Framework) MakeBasicPrometheusAgent(ns, name, group string, replicas in
...
+	var pollErr error
+	if err := wait.PollUntilContextTimeout(ctx, 30*time.Second, 30*time.Minute, true, func(ctx context.Context) (bool, error) {
+		name := fmt.Sprintf("prom-agent-%s", p.Name)
+		// TODO: Implement UpdateStatus() for DaemonSet and check status instead of using Get().
+		dms, err := f.KubeClient.AppsV1().DaemonSets(ns).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			pollErr = fmt.Errorf("failed to get DaemonSet: %w", err)
+			return false, nil
+		}
+		if dms == nil {
```
simonpasquier, line 109: (nit) if there's no error then dms can't be nil.

haanhvu:
@simonpasquier I resolved all the reviews

pkg/operator/config_reloader.go
```diff
@@ -179,10 +179,13 @@ func ImagePullPolicy(imagePullPolicy v1.PullPolicy) ReloaderOption {
 	}
 }
 
-// WithNodeNameEnv sets the withNodeNameEnv option for the config-reloader container.
-func WithNodeNameEnv() ReloaderOption {
+// DaemonSet sets the options that work for DaemonSet mode.
+// Currently we set SHARD env equal to 0, eventhough DaemonSet doesn't use this env.
```
simonpasquier, line 183:
```suggestion
// Currently we set SHARD env equal to 0, eventhough DaemonSet doesn't (and will probably never) suport sharding.
```

simonpasquier (approved):
Thanks!

PR #6720, opened 2024-06-30
Check DaemonSet RBAC permissions for Prometheus Agent's DaemonSet mode

haanhvu:
## Description

Check DaemonSet RBAC permissions for Prometheus Agent's DaemonSet mode, if Prometheus Agent DaemonSet feature gate is enabled, as explained in https://github.com/prometheus-operator/prometheus-operator/pull/6708#pullrequestreview-2144720179

## Type of change

## Verification
For now we just need CI to be green. To test this behavior we need to implement other things in next PRs, after which we'll have e2e tests.

## Changelog entry

Check DaemonSet RBAC permissions for Prometheus Agent's DaemonSet mode

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/version"
 	"golang.org/x/sync/errgroup"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -396,18 +397,13 @@ func run(fs *flag.FlagSet) int {
 	// If Prometheus Agent runs in DaemonSet mode, check if
 	// the operator has proper RBAC permissions on the DaemonSet resource.
 	if cfg.Gates.Enabled(operator.PrometheusAgentDaemonSetFeature) {
-		daemonSetGroupVersion := schema.GroupVersion{
-			Group:   "apps",
-			Version: "v1",
-		}
-
 		allowed, errs, err := k8sutil.IsAllowed(ctx,
 			kclient.AuthorizationV1().SelfSubjectAccessReviews(),
 			cfg.Namespaces.PrometheusAllowList.Slice(),
 			k8sutil.ResourceAttribute{
-				Group:    daemonSetGroupVersion.Group,
-				Version:  daemonSetGroupVersion.Version,
-				Resource: "daemonset",
+				Group:    appsv1.SchemeGroupVersion.Group,
+				Version:  appsv1.SchemeGroupVersion.Version,
+				Resource: "daemonsets",
 				Verbs:    []string{"get", "list", "watch"},
 			})
 		if err != nil {
@@ -418,6 +414,8 @@ func run(fs *flag.FlagSet) int {
 		if !allowed {
 			for _, reason := range errs {
 				level.Error(logger).Log("msg", "missing permissions to manage Daemonset resource for Prometheus Agent", "reason", reason)
+				cancel()
+				return 1
 			}
 		}
 	}
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -394,42 +394,31 @@ func run(fs *flag.FlagSet) int {
 	}
 
 	// If Prometheus Agent runs in DaemonSet mode, check if
-	// the operator has proper RBAC permissions on the Daemonset resource.
+	// the operator has proper RBAC permissions on the DaemonSet resource.
 	if cfg.Gates.Enabled(operator.PrometheusAgentDaemonSetFeature) {
 		daemonSetGroupVersion := schema.GroupVersion{
 			Group:   "apps",
 			Version: "v1",
 		}
 
-		daemonSetSupported, err := checkPrerequisites(
-			ctx,
-			logger,
-			kclient,
+		allowed, errs, err := k8sutil.IsAllowed(ctx,
+			kclient.AuthorizationV1().SelfSubjectAccessReviews(),
 			cfg.Namespaces.PrometheusAllowList.Slice(),
-			daemonSetGroupVersion,
-			"daemonset",
 			k8sutil.ResourceAttribute{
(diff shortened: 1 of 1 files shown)


haanhvu:
@ArthurSens @simonpasquier @kakkoyun This implements Simon's suggestion in https://github.com/prometheus-operator/prometheus-operator/pull/6708#pullrequestreview-2144720179

> The main() function also needs to check that if the feature gate is enabled, the operator has the proper RBAC permissions on the Daemonset resource (like we do for PrometheusAgent for instance). It can be done in a separate PR (before this one?).

Please review to see if I understand the requirement correctly.

cmd/operator/main.go
```diff
@@ -393,6 +393,46 @@ func run(fs *flag.FlagSet) int {
...
 
+	// If Prometheus Agent runs in DaemonSet mode, check if
+	// the operator has proper RBAC permissions on the Daemonset resource.
+	if cfg.Gates.Enabled(operator.PrometheusAgentDaemonSetFeature) {
+		daemonSetGroupVersion := schema.GroupVersion{
+			Group:   "apps",
+			Version: "v1",
+		}
+
+		daemonSetSupported, err := checkPrerequisites(
```
simonpasquier, line 404: we can call `k8sutil.IsAllowed()` only: the DaemonSet resource should always be supported since it's a Kubernetes core API.

cmd/operator/main.go
```diff
@@ -393,6 +393,46 @@ func run(fs *flag.FlagSet) int {
...
+				Verbs:    []string{"update"},
+			},
+		)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to check DaemonSet support", "err", err)
+			cancel()
+			return 1
+		}
+		if !daemonSetSupported {
+			level.Error(logger).Log("msg", "failed to get RBAC permissions on Daemonset resource for Prometheus Agent", "err", err)
```
simonpasquier, line 430:
```suggestion
			for _, reason := range reasons {
				level.Error(logger).Log("msg", "missing permissions to manage Daemonset resources for Prometheus Agent", "reason", reason)
			}
```

haanhvu:
@simonpasquier @ArthurSens @kakkoyun I resolved Simon's reviews. Please also look at the current position of this check. Do we need this check to happen sooner/later?

cmd/operator/main.go
```diff
@@ -393,6 +393,35 @@ func run(fs *flag.FlagSet) int {
...
+		}
+
+		allowed, errs, err := k8sutil.IsAllowed(ctx,
+			kclient.AuthorizationV1().SelfSubjectAccessReviews(),
+			cfg.Namespaces.PrometheusAllowList.Slice(),
+			k8sutil.ResourceAttribute{
+				Group:    daemonSetGroupVersion.Group,
+				Version:  daemonSetGroupVersion.Version,
+				Resource: "daemonset",
+				Verbs:    []string{"get", "list", "watch"},
```
simonpasquier, line 411: I'd already account for write operations.
```suggestion
				Verbs:    []string{"get", "list", "watch", "create", "update", "delete"},
```

cmd/operator/main.go
```diff
@@ -393,6 +393,35 @@ func run(fs *flag.FlagSet) int {
...
+			})
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to check permissions on DaemonSet resource", "err", err)
+			cancel()
+			return 1
+		}
+		if !allowed {
+			for _, reason := range errs {
+				level.Error(logger).Log("msg", "missing permissions to manage Daemonset resource for Prometheus Agent", "reason", reason)
```
simonpasquier, line 420: Either we exit if the permissions aren't sufficient or we pass the information to the prometheus agent controller. I'd go with option 1 for now (if you enable the feature gate, you expect the controller to be functional).

cmd/operator/main.go
```diff
@@ -393,6 +393,35 @@ func run(fs *flag.FlagSet) int {
...
+		daemonSetGroupVersion := schema.GroupVersion{
+			Group:   "apps",
+			Version: "v1",
+		}
+
+		allowed, errs, err := k8sutil.IsAllowed(ctx,
+			kclient.AuthorizationV1().SelfSubjectAccessReviews(),
+			cfg.Namespaces.PrometheusAllowList.Slice(),
+			k8sutil.ResourceAttribute{
+				Group:    daemonSetGroupVersion.Group,
```
simonpasquier, line 408: we could use `SchemeGroupVersion.Group` from `k8s.io/api/apps/v1`

cmd/operator/main.go
```diff
@@ -393,6 +393,35 @@ func run(fs *flag.FlagSet) int {
...
+				Version:  daemonSetGroupVersion.Version,
```
simonpasquier, line 409: `SchemeGroupVersion.Version`

cmd/operator/main.go
```diff
@@ -393,6 +393,35 @@ func run(fs *flag.FlagSet) int {
...
+				Resource: "daemonset",
```
simonpasquier, line 410:
```suggestion
				Resource: "daemonsets",
```

cmd/operator/main.go
ArthurSens: +1 for exiting

cmd/operator/main.go
haanhvu: Why does DaemonSet recource have these, but Prometheus Agent resource doesn't have these?

cmd/operator/main.go
haanhvu: yeah let's exit for now

cmd/operator/main.go
simonpasquier: The operator doesn't need to create/update/delete PrometheusAgent objects.

haanhvu: Ah ok, those are users' roles right?

simonpasquier: yes only users need to modify PrometheusAgents, the operator only modifies the underlying objects such as statefulsets, secrets, configmaps and daemonsets (with your work).

haanhvu:
@ArthurSens @kakkoyun @simonpasquier I resolved all the reviews. I also moved this check to happen before the creation of Prometheus Agent's controller. Please review.

simonpasquier (approved):
thanks!

Issue #6610, opened 2024-05-21
Exclude irrelevant certs/tls secrets watch in monitoring namespace


qx121:
### What is missing? Please describe.

Prometheus operator seems to watch and manage all secrets in monitoring namespace. There are other types of secrets(e.g. IPS/certs/etc) that are irrelevant to prometheus stack. Is it possible to make an include/exclude list for prometheus-operator such that it makes fewer list secret API call to kube-api-server?

## Environment
Kubernetes Version:
Prometheus-Operator Version:

simonpasquier:
You can start the operator with `-secret-field-selector`

pbxqdown:
@simonpasquier thanks. I think having a selector would be the correct direction for this issue. The argument is a field-selector, wondering whether we have label/annotation selector?

Also would like to know more details about how the secret selector is being used. Does monitoring components such as prometheus/alertmanager watch the secret change and update it accordingly? Thank you!

simonpasquier:
context about why we have  a field selector: https://github.com/prometheus-operator/prometheus-operator/pull/3355

> Does monitoring components such as prometheus/alertmanager watch the secret change and update it accordingly? 

The operator needs to react on secret changes. Watching on all secrets in the namespace is the simplest solution. Having a label selector put us at risk of excluding noteworthy secrets.

What's exactly your concern here?

pbxqdown:
We would like to reduce calls to api-server, it is also from performance perspective similar to #3355 
In our stack, not all secrets in monitoring namespace are managed by prometheus-operator. IPS are injected by other controller, certs are managed by other components as well.

It would be good to have a label-selector since it provides larger flexibility. I have free cycle to help with implementation if needed.

But if it is not trivial work, maybe we can try field-selector as well. IPS has kubernetes.io/dockerconfigjson type, so we can use field selector to exclude IPS. Wondering whether prometheus-operator manages any IPS?

simonpasquier:
What is IPS?

I'd be ok adding a label selector argument for secrets like we have the field selector.

pbxqdown:
it is image-pull-secret, usually injected from external party.
It would be great if label selector can be added, I can create a PR if you would like to.

simonpasquier:
sure go ahead.

pbxqdown:
Great, this is the PR: #6731

PR #6731, opened 2024-07-03
cmd: Add secret label selector to filter secrets to watch

pbxqdown:
## Description
This commit adds a flag for secret label selector so user can use label to filter secrets that prometheus-operator watches. This reduces calls to kube-apiserver.

Fixes Exclude irrelevant certs/tls secrets watch in monitoring namespace #6610

## Type of change

## Verification

## Changelog entry

```release-note
cmd: Add secret label selector to filter secrets to watch
```

### DIFF
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -57,10 +57,11 @@ type Config struct {
 	LocalHost string
 
 	// Label and field selectors for resource watchers.
-	PromSelector            LabelSelector
-	AlertmanagerSelector    LabelSelector
-	ThanosRulerSelector     LabelSelector
-	SecretListWatchSelector FieldSelector
+	PromSelector                 LabelSelector
+	AlertmanagerSelector         LabelSelector
+	ThanosRulerSelector          LabelSelector
+	SecretListWatchSelector      FieldSelector
+	SecretListWatchLabelSelector LabelSelector
 
 	// Controller id for pod ownership.
 	ControllerID string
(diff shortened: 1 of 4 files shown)


simonpasquier:
Thanks! The same change needs to be applied to pkg/prometheus/agent/operator.go, and pkg/prometheus/server/operator.go.

pbxqdown:
Thank you for review! Updated the commit. Please let me know if there are other places that need update.

ArthurSens (approved):
LGTM! Ideally, we'd have an end-to-end test for this one, but I'm not really sure if we have a good example 🤔

pbxqdown:
Thank you!  It would be great if there is a framework testing cmd flags.

pkg/operator/config.go
```diff
@@ -57,10 +57,11 @@ type Config struct {
...
 
 	// Label and field selectors for resource watchers.
-	PromSelector            LabelSelector
-	AlertmanagerSelector    LabelSelector
-	ThanosRulerSelector     LabelSelector
-	SecretListWatchSelector FieldSelector
+	PromSelector                 LabelSelector
+	AlertmanagerSelector         LabelSelector
+	ThanosRulerSelector          LabelSelector
+	SecretListWatchSelector      FieldSelector
```
simonpasquier, line 63: could you rename for clarity?
```suggestion
	SecretListWatchFieldSelector      FieldSelector
```

pbxqdown: sure, renamed.

simonpasquier (approved):
Thanks!

PR #6743, opened 2024-07-09
Feat: new logger slog

### LINKED ISSUE #6183: Replace `github.com/go-kit/klog` by the new `slog` package
[slog](https://pkg.go.dev/log/slog) provides structured logging out of the box. It should be a good fit to replace  `github.com/go-kit/klog`  which is used currently.

Advantages:
* Reduce our dependency tree.
* Likely better integration with `k8s.io/klog/v2` . We currently use a [shim](https://github.com/simonpasquier/klog-gokit) but it doesn't support all klog's [features](https://github.com/simonpasquier/klog-gokit#limitations).

perebaj:
## Description

Fixes #6183

Following the @ArthurSens suggestion, first I'm adapting the new slog constructor to follow the same requirements of the `k8s.io/klog/v2` and afterwards I plan to replace all the imports.

## Type of change

## Verification

## Changelog entry

```release-note
feature: add new logger for slog library
```

### DIFF
--- a/internal/log/slog.go
+++ b/internal/log/slog.go
@@ -38,7 +38,7 @@ func NewLoggerSlog(c Config) (*slog.Logger, error) {
 	return logger, nil
 }
 
-func replaceSlogAttributes(groups []string, a slog.Attr) slog.Attr {
+func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
 	if a.Key == "time" {
 		return slog.Attr{
 			Key:   "ts",
--- a/internal/log/slog.go
+++ b/internal/log/slog.go
@@ -42,6 +42,9 @@ func NewLoggerSlog(c Config) (*slog.Logger, error) {
 	return slog.New(handler), nil
 }
 
+// replaceSlogAttributes replaces fields that were added by default by slog, but had different
+// formats or key names in github.com/go-kit/log. The operator was originally implemented with go-kit/log,
+// so we use these replacements to make the migration smoother.
 func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
 	if a.Key == "time" {
 		return slog.Attr{
@@ -67,6 +70,7 @@ func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
 	return a
 }
 
+// getHandlerFromFormat returns a slog.Handler based on the provided format and slog options.
 func getHandlerFromFormat(format string, opts slog.HandlerOptions) (slog.Handler, error) {
 	var handler slog.Handler
 	switch strings.ToLower(format) {
@@ -81,6 +85,7 @@ func getHandlerFromFormat(format string, opts slog.HandlerOptions) (slog.Handler
 	}
 }
 
+// parseLevel returns the slog.Level based on the provided string.
 func parseLevel(lvl string) (slog.Level, error) {
 	switch strings.ToLower(lvl) {
 	case LevelAll:
--- a/internal/log/slog.go
+++ b/internal/log/slog.go
@@ -1,0 +1,65 @@
+package log
+
+import (
+	"fmt"
+	"log/slog"
+	"os"
+	"strings"
+)
+
+func NewLoggerSlog(c Config) (*slog.Logger, error) {
+	var (
+		logger *slog.Logger
+	)
+
+	lvlOption, err := ParseLevel(c.Level)
+	if err != nil {
+		return nil, err
+	}
+
+	logger, err = ParseFmt(lvlOption, c.Format)
+	if err != nil {
+		return nil, err
+	}
+
+	return logger, nil
+}
+
+func ParseFmt(lvlOption slog.Level, format string) (*slog.Logger, error) {
+	var logger *slog.Logger
+	switch strings.ToLower(format) {
+	case FormatLogFmt:
+		h := slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+			Level: lvlOption,
+		})
+		logger = slog.New(h)
+		return logger, nil
+	case FormatJSON:
+		h := slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+			Level: lvlOption,
+		})
+		logger = slog.New(h)
+		return logger, nil
+	default:
+		return nil, fmt.Errorf("log format %s unknown, %v are possible values", format, AvailableLogFormats)
+	}
+}
+
+func ParseLevel(lvl string) (slog.Level, error) {
+	switch strings.ToLower(lvl) {
+	case LevelAll:
+		return slog.LevelDebug, nil
+	case LevelDebug:
+		return slog.LevelDebug, nil
+	case LevelInfo:
+		return slog.LevelInfo, nil
+	case LevelWarn:
+		return slog.LevelWarn, nil
+	case LevelError:
+		return slog.LevelError, nil
+	case LevelNone:
+		return -1, nil
+	default:
+		return -1, fmt.Errorf("log log_level %s unknown, %v are possible values", lvl, AvailableLogLevels)
+	}
+}
--- a/internal/log/slog.go
+++ b/internal/log/slog.go
@@ -30,12 +30,16 @@ func NewLoggerSlog(c Config) (*slog.Logger, error) {
 		return nil, err
 	}
 
-	logger, err := parseFmt(lvlOption, c.Format)
+	handler, err := getHandlerFromFormat(c.Format, slog.HandlerOptions{
+		Level:       lvlOption,
+		AddSource:   true,
+		ReplaceAttr: replaceSlogAttributes,
+	})
 	if err != nil {
 		return nil, err
 	}
 
-	return logger, nil
+	return slog.New(handler), nil
 }
 
 func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
@@ -63,25 +67,15 @@ func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
 	return a
 }
 
-func parseFmt(lvlOption slog.Level, format string) (*slog.Logger, error) {
-	var logger *slog.Logger
+func getHandlerFromFormat(format string, opts slog.HandlerOptions) (slog.Handler, error) {
+	var handler slog.Handler
 	switch strings.ToLower(format) {
 	case FormatLogFmt:
-		h := slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
-			Level:       lvlOption,
-			AddSource:   true,
-			ReplaceAttr: replaceSlogAttributes,
-		})
-		logger = slog.New(h)
-		return logger, nil
+		handler = slog.NewTextHandler(os.Stdout, &opts)
+		return handler, nil
 	case FormatJSON:
-		h := slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
-			Level:       lvlOption,
-			AddSource:   true,
-			ReplaceAttr: replaceSlogAttributes,
-		})
-		logger = slog.New(h)
-		return logger, nil
+		handler = slog.NewJSONHandler(os.Stdout, &opts)
+		return handler, nil
 	default:
 		return nil, fmt.Errorf("log format %s unknown, %v are possible values", format, AvailableLogFormats)
 	}
--- a/internal/log/slog.go
+++ b/internal/log/slog.go
@@ -19,19 +19,18 @@ import (
 	"log/slog"
 	"os"
 	"strings"
+	"time"
 )
 
+// NewLoggerSlog returns a *slog.Logger that prints in the provided format at the
+// provided level with a UTC timestamp and the caller of the log entry.
 func NewLoggerSlog(c Config) (*slog.Logger, error) {
-	var (
-		logger *slog.Logger
-	)
-
-	lvlOption, err := ParseLevel(c.Level)
+	lvlOption, err := parseLevel(c.Level)
 	if err != nil {
 		return nil, err
 	}
 
-	logger, err = ParseFmt(lvlOption, c.Format)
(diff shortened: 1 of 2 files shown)


internal/log/slog.go
```diff
@@ -0,0 +1,107 @@
...
+
+	logger, err := parseFmt(lvlOption, c.Format)
+	if err != nil {
+		return nil, err
+	}
+
+	return logger, nil
+}
+
+func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
```
nicolastakashi, line 41: I'm wondering why do we want to replace this attributes?

internal/log/slog.go
```diff
@@ -0,0 +1,107 @@
...
+			ReplaceAttr: replaceSlogAttributes,
+		})
+		logger = slog.New(h)
+		return logger, nil
+	default:
+		return nil, fmt.Errorf("log format %s unknown, %v are possible values", format, AvailableLogFormats)
+	}
+}
+
+func parseLevel(lvl string) (slog.Level, error) {
```
nicolastakashi, line 90:
```suggestion
func parseLevel(lvl string) (*slog.Level, error) {
	level := slog.Level(1)
	err := level.UnmarshalText([]byte(lvl))
	if err != nil {
		return nil, err
	}
	return &level, nil
}
```

Wdyt this approach?

internal/log/slog.go
```diff
@@ -0,0 +1,107 @@
...
+		return slog.Attr{
+			Key:   "caller",
+			Value: a.Value,
+		}
+	}
+
+	return a
+}
+
+func parseFmt(lvlOption slog.Level, format string) (*slog.Logger, error) {
```
nicolastakashi, line 66: IMO this functions is doing more than parsing the format, it's also initializing the logger with the proper handler.

Maybe you can have one approach like this.

```go
func getHandlerFromFormat(format string, opts slog.HandlerOptions) (slog.Handler, error) {
    var handler slog.Handler
    switch {
	case logFormat == "text":
		handler = slog.NewTextHandler(os.Stdout, handlerOptions)
	case logFormat == "json":
		handler = slog.NewJSONHandler(os.Stdout, handlerOptions)
	default:
		return nil, errors.New("unknown log format")
	}
}
```

So you can change your `NewLoggerSlog` to looks like this:

```go
func NewLoggerSlog(c Config) (*slog.Logger, error) {
	lvlOption, err := parseLevel(c.Level)
	if err != nil {
		return nil, err
	}

	handler, err := getHandlerFromFormat(lvlOption, slog.HandlerOptions{
			Level:       lvlOption,
			AddSource:   true,
			ReplaceAttr: replaceSlogAttributes,
		})
	if err != nil {
		return nil, err
	}

	return slog.New(h), nil
}
```

wdyt?

perebaj: yep, makes sense, will also simplify the tests. Adressing

internal/log/slog.go
perebaj: yep. it makes sense because simplify a lot of code. 

But taking a look at the implementation, the [UnmarshalText](https://github.com/golang/go/blob/master/src/log/slog/level.go#L107-L146) just works for DEBUG, INFO, WARN and ERROR. So the other cases (ALL and None) won't be covered

What do you suggest in that case?

internal/log/slog.go
perebaj: If you initialize the [other log structure](https://github.com/prometheus-operator/prometheus-operator/blob/main/internal/log/log.go#L55) it's possible to see that it had different attributes. The same ones that I'm putting in this new implementation

The steps that I took to discover it were:

```Go
func TestWhyWeNeed2Replace(t *testing.T) {
        // initialize the NewLogger from internal package
	t.Log("TestTmp")
	l, err := NewLogger(Config{
		Level:  LevelDebug,
		Format: FormatJSON,
	})
	if err != nil {
		t.Fatalf("unexpected error: %v", err)
	}
	level.Info(l).Log("msg", "test")

	// initialize a default slog.Logger
	l2 := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
		Level:     slog.LevelDebug,
		AddSource: true}))

	l2.Info("test")
}
```

Output NewLogger:

```JSON
{"caller":"slog_test.go:166","level":"info","msg":"test","ts":"2024-07-13T11:08:18.119068Z"}
```

Output default Slog(without changed the attributes)

```JSON
{"time":"2024-07-13T08:13:35.356893-03:00","level":"INFO","source":{"function":"github.com/prometheus-operator/prometheus-operator/internal/log.TestWhyWeNeed2Replace","file":"/Users/jonathan.silva/prometheus-operator/internal/log/slog_test.go","line":175},"msg":"test"}
```

In that case, it's possible to notice that the attributes that I changed are the ones that are different between them: 

- `time -> ts`
- `level string in lower case`
- `source -> caller`

internal/log/slog.go
perebaj: addressed it here -> https://github.com/prometheus-operator/prometheus-operator/pull/6743/commits/b9ea62456d78c47b27005fbeb945bd3f202d8377

internal/log/slog.go
nicolastakashi: IMH if we get none we should use default info and about all I don't know if I follow what this means?

If you set all, are you expecting getting, debug, info, warn, and error logs?

Log use to work like:
error -> only error logs
warn -> only warn and error logs
info -> only info, warn and error logs
debug -> all logs

wdyt?

internal/log/slog.go
nicolastakashi: I got, but why do we want to replace it?
Why not use the default info provided by slog?
Maybe I missing some part, sorry if it sounds weird hehe 😄

perebaj: > Why do we want to replace it?

I thought that was an agreement to keep both implementations using the same patterns... If isn't, who do I need to talk to decide it?

internal/log/slog.go
perebaj: > If you set all, are you expecting getting, debug, info, warn, and error logs?

Today the default behavior for `ALL`, is to use `debug` as I set.

> IMH if we get none we should use default info 

Today, the default behavior is to print nothing.

internal/log/slog.go
perebaj: but if you are ok to no replace it, that is it...

nicolastakashi: I mean change it might or might not be considered a breaking change.

WHat's your opinion @simonpasquier @ArthurSens @slashpai ?

internal/log/slog.go
nicolastakashi: I think debug is way to verbose to be used as default, we should go as info

perebaj: So, just to confirm if I understand the level thing

LevelAll = `debug level`
LevelNone = `info level` 

that is it?

---

About the code, using the `level.UnmarshalText`. Will look like this?

```Golang
func parseLevel(lvl string) (*slog.Level, error) {
	level := slog.Level(1)
	if lvl == LevelNone {
		level = slog.LevelInfo
		return &level, nil
	}
	if lvl == LevelAll {
		level = slog.LevelDebug
		return &level, nil
	}
	err := level.UnmarshalText([]byte(lvl))
	if err != nil {
		return nil, err
	}
	return &level, nil
}
```

internal/log/slog.go
ArthurSens: I think we can move forward with no breaking changes in the log attributes for now. If we decide this is too much work to main, we can change it later!

internal/log/slog.go
```diff
@@ -0,0 +1,101 @@
...
+		ReplaceAttr: replaceSlogAttributes,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return slog.New(handler), nil
+}
+
+func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
```
ArthurSens, line 45:
```suggestion
// replaceSlogAttributes replaces fields that were added by default by slog, but had different 
// formats or key names in github.com/go-kit/log. The operator was originally implemented with go-kit/log,
// so we use these replacements to make the migration smoother.
func replaceSlogAttributes(_ []string, a slog.Attr) slog.Attr {
```

internal/log/slog_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	l := slog.New(h)
+
+	l.Info("test")
+
+	var m map[string]interface{}
+	err := json.Unmarshal(buf.Bytes(), &m)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
```
ArthurSens, line 43: Could we use `require.NoError` instead? We're currently trying our best to decrease usage of t.Fatal 😬

internal/log/slog_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if _, ok := m["level"]; !ok {
+		t.Fatalf("key level not found in the JSON")
+	}
```
ArthurSens, line 47: Not sure which require method we could use here... maybe `require.Contains` 🤔 ?

internal/log/slog_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	wantJSONHandler := slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		Level:     slog.LevelDebug,
+		AddSource: true,
+	})
+
+	if !reflect.DeepEqual(handler, wantJSONHandler) {
+		t.Fatalf("handler not equal to wantJSONHandler")
+	}
```
ArthurSens, line 75: require.Equal

ArthurSens (dismissed):
LGTM!

I'd just document more the functions we created to smoothen out the migration. could you also remove the usage of t.Fatal from tests?

internal/log/slog_test.go
perebaj: Something like that, I think

```Golang
require.Contains(t, m, "level")
require.Contains(t, m, "msg")
require.Contains(t, m, "caller")
```

internal/log/slog_test.go
perebaj: addressed it here https://github.com/prometheus-operator/prometheus-operator/pull/6743/commits/d0a23439ca5b45595cd34b07496e874c596f5622

internal/log/slog_test.go
perebaj: addressed it here -> https://github.com/prometheus-operator/prometheus-operator/pull/6743/commits/e43dbdefbc13ceb03adb467e0b42e534325d31c8#diff-98b5be74814c0956b26ebdc84e9ba3b203b9cf5197e815912b6a6f6c2be000deR47-R50

internal/log/slog.go
perebaj: instead of commit your suggestion I took the opportunity to change everything https://github.com/prometheus-operator/prometheus-operator/pull/6743/commits/1978e2c56139ac0b6d2b64f3562a9f1aecff4668
in a single shot

ArthurSens (approved):
Perfect! I'll leave this PR open for a while in case other maintainers have different opinions.

If you want to start working on removing the go-kit import. Maybe you [could create chained PRs](https://blog.logrocket.com/using-stacked-pull-requests-in-github/)?

perebaj:
> Perfect! I'll leave this PR open for a while in case other maintainers have different opinions.
> 
> If you want to start working on removing the go-kit import. Maybe you [could create chained PRs](https://blog.logrocket.com/using-stacked-pull-requests-in-github/)?

For sure!! Just to confirm, we aren't going to address [this change](https://github.com/prometheus-operator/prometheus-operator/pull/6743#discussion_r1676456798) now, right? 

Reason: this will introduce a new behavior in the log library

ArthurSens:
> > Perfect! I'll leave this PR open for a while in case other maintainers have different opinions.
> > If you want to start working on removing the go-kit import. Maybe you [could create chained PRs](https://blog.logrocket.com/using-stacked-pull-requests-in-github/)?
> 
> For sure!! Just to confirm, we aren't going to address [this change](https://github.com/prometheus-operator/prometheus-operator/pull/6743#discussion_r1676456798) now, right?
> 
> Reason: this will introduce a new behavior in the log library

Good question, I'd move forward without this change, since this is a refactoring exercise. Removing `all` and `none` log levels is more of a breaking change than replacing certain log keys since it will break the start-up of the operator if any of those log levels are set.

But happy to be challenged @nicolastakashi!!

ArthurSens:
> this is a refactoring exercise. 

And about that, do you mind updating your PR description with `NONE` kind of change? I don't think we consider logging as a Prometheus-Operator feature. We usually only add PRs to the changelog if they are user facing changes :)

nicolastakashi:
> > > Perfect! I'll leave this PR open for a while in case other maintainers have different opinions.
> > > If you want to start working on removing the go-kit import. Maybe you [could create chained PRs](https://blog.logrocket.com/using-stacked-pull-requests-in-github/)?
> > 
> > 
> > For sure!! Just to confirm, we aren't going to address [this change](https://github.com/prometheus-operator/prometheus-operator/pull/6743#discussion_r1676456798) now, right?
> > Reason: this will introduce a new behavior in the log library
> 
> Good question, I'd move forward without this change, since this is a refactoring exercise. Removing `all` and `none` log levels is more of a breaking change than replacing certain log keys since it will break the start-up of the operator if any of those log levels are set.
> 
> But happy to be challenged @nicolastakashi!!

I was not thinking about removing it, but providing the same capapilities

nicolastakashi:
> > I was not thinking about removing it, but providing the same capapilities
> 
> I'm a bit confused, `all` and `debug` having the same behavior makes sense to me, but here you say:
> 
> > IMH if we get none we should use default info
> 
> The current behavior of `none` is to have no logs at all. If we change `none` to have the same behavior as `info`, it doesn't feel like "providing the same capabilities" 🤔

OK, that's true, I'm fine keeping the same interface for now.

perebaj:
😄!

PR #6753, opened 2024-07-15
Replace go kit log

perebaj:
## Description

Adapting the places where we are using the `go_kit_log library`, but doing that in baby steps to avoid changing many places simultaneously.

I'm Keeping the oldLogger (go_kit_log) in places where we have more dependencies and adding the new Logger where it's easy to change

This PR follows up on the changes made here https://github.com/prometheus-operator/prometheus-operator/pull/6743

## Type of change

## Verification

## Changelog entry

```release-note
replace go-kit-log
```


cmd/admission-webhook/main.go
```diff
@@ -61,7 +60,14 @@ func main() {
...
 
-	logger, err := logging.NewLogger(logConfig)
+	logger, err := logging.NewLoggerSlog(logConfig)
+	if err != nil {
+		stdlog.Fatal(err)
+	}
+
+	// The old logger will be removed in the future. For now, we are using it to avoid breaking changes and
+	// to have a smooth transition.
+	oldLogger, err := logging.NewLogger(logConfig)
```
perebaj, line 70: Does this approach make sense? I think that will be easier to review and change the dependencies code if we do that incrementally.

internal/goruntime/cpu.go
perebaj, line 1: the `internal/goruntime/cpu.go` and `internal/goruntime/memory.go` were the places where we had less dependencies between the code and the log library, for this reason, I'm pushing this code to use the new approach.

Does it make sense?

ArthurSens: Yep, makes sense!

cmd/prometheus-config-reloader/main.go
```diff
@@ -122,25 +121,31 @@ func main() {
...
 
-	logger, err := logging.NewLogger(logConfig)
+	logger, err := logging.NewLoggerSlog(logConfig)
+	if err != nil {
+		stdlog.Fatal(err)
+	}
+
+	// The old logger will be removed in the future. For now, we are using it to avoid breaking changes and
+	// to have a smooth transition.
+	oldLogger, err := logging.NewLogger(logConfig)
```
ArthurSens, line 131:
```suggestion
	// We're currently migrating our logging library from go-kit to slog.
	// The go-kit logger is being removed in small PRs. For now, we are creating 2 loggers to avoid breaking changes and
	// to have a smooth transition.
	goKitLogger, err := logging.NewLogger(logConfig)
```

ArthurSens: Could you make similar changes to the other two `main.go` files?

ArthurSens (approved):
I've made some comments just to make it clearer why we need 2 loggers for now, but LGTM!

Could we also fix the commits while we're at it? 1 commit for this PRs looks enough :)

cmd/prometheus-config-reloader/main.go
perebaj: For sure!

perebaj:
> I've made some comments just to make it clearer why we need 2 loggers for now, but LGTM!
> 
> Could we also fix the commits while we're at it? 1 commit for this PRs looks enough :)

I will take a look and try to solve it, but this amount of commits was because I did the branch chain...

cmd/operator/main.go
```diff
@@ -193,32 +193,38 @@ func run(fs *flag.FlagSet) int {
...
+	// The go-kit logger is being removed in small PRs. For now, we are creating 2 loggers to avoid breaking changes and
+	// to have a smooth transition.
+	goKitLogger, err := logging.NewLogger(logConfig)
 	if err != nil {
 		stdlog.Fatal(err)
 	}
 
 	if err := cfg.Gates.UpdateFeatureGates(*featureGates.Map); err != nil {
-		level.Error(logger).Log("error", err)
+		logger.Error("", "error", err)
```
ArthurSens, line 210: Interesting! I wonder what the output is like for this... will the `msg` key attribute not be logged at all? Or will the output be something like 
```json
{"msg": "", "error": "......."}
```
?

perebaj:
I tried to squash the commits and ruined the commit tree. Closing it to open a new PR

cmd/operator/main.go
perebaj: It will be like the output that you shared

PR #6755, opened 2024-07-17
replace go kit log

perebaj:
## Description

Adapting the places where we are using the `go_kit_log library`, but doing that in baby steps to avoid changing many places simultaneously.

I'm Keeping the oldLogger (go_kit_log) in places where we have more dependencies and adding the new Logger where it's easy to change

This PR follows up on the changes made here https://github.com/prometheus-operator/prometheus-operator/pull/6743

## Type of change

## Verification

## Changelog entry

```release-note
replace go-kit-log
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	stdlog "log"
+	"log/slog"
 	"net/http"
 	"net/http/pprof"
 	"os"
@@ -27,7 +28,6 @@ import (
 	"syscall"
 
 	"github.com/go-kit/log"
-	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	versioncollector "github.com/prometheus/client_golang/prometheus/collectors/version"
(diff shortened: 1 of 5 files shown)


perebaj:
@ArthurSens sorry for the mess, but I've already addressed the comments you claimed in this PR: https://github.com/prometheus-operator/prometheus-operator/pull/6753 and I squashed the commits...

cmd/operator/main.go
```diff
@@ -193,32 +193,38 @@ func run(fs *flag.FlagSet) int {
...
+	// The go-kit logger is being removed in small PRs. For now, we are creating 2 loggers to avoid breaking changes and
+	// to have a smooth transition.
+	goKitLogger, err := logging.NewLogger(logConfig)
 	if err != nil {
 		stdlog.Fatal(err)
 	}
 
 	if err := cfg.Gates.UpdateFeatureGates(*featureGates.Map); err != nil {
-		level.Error(logger).Log("error", err)
+		logger.Error("", "error", err)
```
ArthurSens, line 210: Interesting thing with slog. Looks like it will **always** have the `msg` field, where go-kit/log doesn't require that.

An empty message will still be logged, as seen in this [playground](https://go.dev/play/p/46QcymutNGw). Not that this is a problem, but we're introducing a new field to all places where `msg` isn't filled with something

ArthurSens: image

ArthurSens (approved):
LGTM, let's move on to the next!

PR #6768, opened 2024-07-23
replace go kit log admission pkg

perebaj:
## Description

Adapting the places where we are using the `go_kit_log library` into the admission package

## Type of change

## Verification

## Changelog entry

```release-note
replace kit log with slog in package admission
```

### DIFF
--- a/pkg/admission/admission_test.go
+++ b/pkg/admission/admission_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"math"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -314,7 +315,8 @@ func TestAlertmanagerConfigConversion(t *testing.T) {
 func api() *Admission {
 	a := New(
 		slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
-			Level: slog.LevelDebug,
+			// slog level math.MaxInt means no logging
+			Level: slog.Level(math.MaxInt),
 		})),
 	)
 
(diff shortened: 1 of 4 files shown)


pkg/admission/admission_test.go
```diff
@@ -313,7 +313,12 @@ func TestAlertmanagerConfigConversion(t *testing.T) {
 }
 
 func api() *Admission {
-	a := New(level.NewFilter(log.NewLogfmtLogger(log.NewSyncWriter(os.Stdout)), level.AllowNone()))
+	a := New(
+		slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+			// slog level math.MaxInt means no logging
+			Level: slog.Level(math.MaxInt),
```
perebaj, line 319: @ArthurSens testing this code and trying to log nothing I found that the approach that I pushed [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/internal/log/slog.go#L101-L104) is not working properly. 

[To disable the logs we need to use math.MaxInt instead of -1](https://stackoverflow.com/questions/16895651/how-to-implement-level-based-logging-in-golang) 

WDYT if we first merge these changes here and after create a different PR to fix LevelNone? [`slog.go`](https://github.com/prometheus-operator/prometheus-operator/blob/main/internal/log/slog.go)

ArthurSens (approved):
LGTM, let's fix the level none in a separate PR!

perebaj:
> LGTM, let's fix the level none in a separate PR!

Thanks. It's here -> https://github.com/prometheus-operator/prometheus-operator/pull/6769

Issue #6508, opened 2024-04-14
Support for RemoteWrite CRD


superbrothers:
### Component(s)

Prometheus

### What is missing? Please describe.

I would like to allow users to add arbitrary RemoteWrite/Read configs to Prometheus.

Scenario: cluster-admin provides a managed Prometheus server to users using Prometheus CR. This Prometheus CR can only be modified by cluster-admin. The user adds config to the Prometheus server using a RemoteWrite CR to forward metrics to an external managed Prometheus service.

```yaml
apiVersion: monitoring.coreos.com/v1alpha1
kind: RemoteWrite
metadata:
  name: example
  namespace: default
spec:
  url: xxxx/api/v1/remote_write
  authorization:
    name: cred
    key: token
  writeRelabelConfigs:
  - ...
```

```yaml
apiVersion: monitoring.coreos.com/v1alpha1
kind: RemoteRead
metadata:
  name: example
  namespace: default
spec:
  url: xxxx/api/v1/read
  sigv4:
    region: ap-northeast-1
```

```yaml
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
metadata:
  name: prometheus
  namespace: monitoring
spec:
  RemoteWriteNamespaceSelector: {}
  RemoteWriteSelector: {}
  RemoteReadNamespaceSelector: {}
  RemoteReadSelector: {}
```

Since RemoteWrite/Read CR are namespaced resources, it may be better to limit the target metrics to the namespace in which the object was created.

### Describe alternatives you've considered.

None.

## Environment
Kubernetes Version:
Prometheus-Operator Version:

mviswanathsai:
If the RemoteWrite is namespaced and is only supposed to export metrics from its own namespace, there would also be a relabel config in the final generated config I see. But is there a reason why it should be namespaced? 
Could (an alternative) dedicated relabeling rules that you could add in the CR help with the separation of concerns among teams a bit more flexibly? 
Just bouncing off some ideas.

superbrothers:
> I'm more curious about the remote-read use caes, can you provide more details about it?

Actually, I only want RemoteWrite CR; I suggested RemoteRead CR because it seems unnatural that RemoteWrite exists but RemoteRead CR does not.

> I suppose that you mean the operator should automatically add a relabeling config dropping all metrics without a namespace="<remote write's namesapce>" label?

[...]

PR #6781, opened 2024-07-29
Proposal for RemoteWrite CRD

superbrothers:
This PR adds a proposal document for RemoteWrite CRD.

ref. https://github.com/prometheus-operator/prometheus-operator/issues/6508

cc @simonpasquier

### DIFF
--- a/Documentation/proposals/202407-remote-write.md
+++ b/Documentation/proposals/202407-remote-write.md
@@ -113,6 +113,23 @@ The Prometheus operator will generate the Prometheus configuration including rem
 
 The operator will respect the --namespaces and --deny-namespaces flags when looking for RemoteWrite objects.
 
+### Administrative limits
+
+Similar to what exists for scrape resources (e.g. `enforcedSampleLimit` for  `ServiceMonitor`), we will allow the Prometheus resource's owners to setup upper-bound limits on the remote write options. For example, it will be possible to disable metadata sending or limit the queue capacity:
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: example
+  namespace: default
+spec:
+   remoteWriteLimits:
+     metadataConfig:
+       send: false
+     queueConfig:
+       capacity: 1024
+       maxShards: 16
 ## PrometheusAgent CRD
 
 The PrometheusAgent CRD also supports the RemoteWrite CRD. Everything explained about the Prometheus CRD works in the same way for the PrometheusAgent.
--- a/Documentation/proposals/202407-remote-write.md
+++ b/Documentation/proposals/202407-remote-write.md
@@ -1,0 +1,121 @@
+# RemoteWrite CRD
+
+- Owners:
+    - [@superbrothers](https://github.com/superbrothers)
+- Related Tickets:
+    - [#6508](https://github.com/prometheus-operator/prometheus-operator/issues/6508)
+- Other docs:
+    - n/a
+
+## TL;DR
+
+This design doc proposes RemoteWrite CRD, which enables cluster admins to delegate the ability to configure Prometheus remote_rewrite configuration to application developers/operators.
+
+## Why
+
+The Prometheus remote_write configuration is defined in the Prometheus CRD. Currently, the configuration data generation is the responsibility of cluster admins.
+
+## Goals
+
+The main goal is for application developers/operators to configure themselves how the remote endpoint sends metrics.
+
+This means exposing a new CRD to configure Prometheus remote_write configuration.
+
+## Non-goals
+
+We do not cover "RemoteRead" CRD here. We can still implement it later if needed.
+
+## How
+
+### RemoteWrite CRD
+
+The RemoteWrite CRD represents one of the Prometheus remote_write configuration scoped to the resource’s namespace.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: RemoteWrite
+metadata:
+  name: example
+  namespace: default
+spec:
+  url: "https://aps-workspaces.us-west-2.amazonaws.com/workspaces/<workspace id>/api/v1/remote_write"
+  sigv4:
+    region: us-west-2
+    accessKey:
+      name: aws_access
+      key: access_key
+    secretKey:
+      name: aws_access
+      key: secret_key
+```
+
+```go
+package v1alpha1
+
+import (
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type RemoteWrite struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec              monitoringv1.RemoteWriteSpec `json:"spec"`
+}
+```
+
+### Prometheus CRD
+
+The Prometheus CRD is extended with 2 new fields (remoteWriteSelector and remoteWriteNamespaceSelector) that define which RemoteWrite resources are associated with this Prometheus instance.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: example
+  namespace: default
+spec:
+  # RemoteWrite objects to be selected. If nil, nothing will be selected.
+  remoteWriteSelector: {}
+  # Namespace(s) to be selected for RemoteWrite discovery. If nil, only check in its own namespace.
+  remoteWriteNamespaceSelector: {}
+  ...
+```
+
+#### Respect enforceNamespaceLabel and excludedFromEnforcement
+
+The RemoteWrite CRD should respect the Prometheus `.spec.enforcedNamespaceLabel` field.
+
+The object’s namespace is added as the last item in the `write_relabel_config` of the remote_write configuration.
+
+```yaml
+write_relabel_configs:
+...
+- target_label: <enforced-namespace-label>
+  replacement: <namespace>
+```
+
+The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field.
+
+These were proposed at https://github.com/prometheus-operator/prometheus-operator/issues/6508#issuecomment-2058942336.
+
+### PrometheusAgent CRD
+
+TODO
+
+### Configuration generation
+
+The Prometheus operator will generate the Prometheus configuration including remote_write configuration from the Prometheus CRD and the RemoteWrite resources matching remoteWriteSelector from the namespace(s) selected by remoteWriteNamespaceSelector for additional remote_rewrite configuration. The generated configuration will be stored in a secret mounted in the Prometheus pod.
+
+The operator will respect the --namespaces and --deny-namespaces flags when looking for RemoteWrite objects.
+
+## Alternatives
+
+### Prometheus per namespace or team
+
+An application developer/operator can deploy Prometheus instances directly. However, Prometheus instances may be provided as managed by cluster admins.
+
+### The team responsible for the Prometheus object configuring individual remote write destinations on behalf of each "tenant"
+
+An application developer/operator will somehow share the remote_write configuration with the team responsible for Prometheus objects. This includes credential information for remote_write.
--- a/Documentation/proposals/202407-remote-write.md
+++ b/Documentation/proposals/202407-remote-write.md
@@ -1,11 +1,11 @@
 # RemoteWrite CRD
 
 - Owners:
-    - [@superbrothers](https://github.com/superbrothers)
+  - [@superbrothers](https://github.com/superbrothers)
 - Related Tickets:
-    - [#6508](https://github.com/prometheus-operator/prometheus-operator/issues/6508)
+  - [#6508](https://github.com/prometheus-operator/prometheus-operator/issues/6508)
 - Other docs:
-    - n/a
+  - n/a
 
 ## TL;DR
 
@@ -61,7 +61,7 @@ type RemoteWrite struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec              monitoringv1.RemoteWriteSpec `json:"spec"`
+	Spec monitoringv1.RemoteWriteSpec `json:"spec"`
 }
 ```
 
--- a/Documentation/proposals/202407-remote-write.md
+++ b/Documentation/proposals/202407-remote-write.md
@@ -105,7 +105,7 @@ write_relabel_configs:
 
 The item is inserted in the first position by the operator before any other user relabelings then there's no way that users can have access to "foreign" metrics.
 
-The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field.
+The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field. It is a list of references to objects to be excluded from `enforcedNamespaceLabel`.
 
 ### Configuration generation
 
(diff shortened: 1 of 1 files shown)


Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,121 @@
...
+The object’s namespace is added as the last item in the `write_relabel_config` of the remote_write configuration.
+
+```yaml
+write_relabel_configs:
+...
+- target_label: <enforced-namespace-label>
+  replacement: <namespace>
+```
+
+The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field.
```
simonpasquier, line 99: Side-note: the new CRD would need to be added to the list of supported resources.

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,121 @@
...
+
+These were proposed at https://github.com/prometheus-operator/prometheus-operator/issues/6508#issuecomment-2058942336.
+
+### PrometheusAgent CRD
+
+TODO
```
simonpasquier, line 105: From I can see, it should work the same as the Prometheus CRD. Is there anything that would differ?

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Configuration generation
+
+The Prometheus operator will generate the Prometheus configuration including remote_write configuration from the Prometheus CRD and the RemoteWrite resources matching remoteWriteSelector from the namespace(s) selected by remoteWriteNamespaceSelector for additional remote_rewrite configuration. The generated configuration will be stored in a secret mounted in the Prometheus pod.
```
simonpasquier, line 109: > The generated configuration will be stored in a secret mounted in the Prometheus pod.

No change to the current way, right?

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,121 @@
...
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: RemoteWrite
+metadata:
+  name: example
+  namespace: default
+spec:
+  url: "https://aps-workspaces.us-west-2.amazonaws.com/workspaces/<workspace id>/api/v1/remote_write"
+  sigv4:
+    region: us-west-2
+    accessKey:
```
simonpasquier, line 44: It might be good to mention that secret/configmap refs will be resolved in the RemoteWrite namespace.

simonpasquier (commented):
Thanks for starting this! My main question would be how admins can protect the Prometheus instance from "expensive" configurations (see also https://github.com/prometheus/prometheus/issues/13174). I don't have a good answer yet.

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,121 @@
...
+
+This design doc proposes RemoteWrite CRD, which enables cluster admins to delegate the ability to configure Prometheus remote_rewrite configuration to application developers/operators.
+
+## Why
+
+The Prometheus remote_write configuration is defined in the Prometheus CRD. Currently, the configuration data generation is the responsibility of cluster admins.
+
+## Goals
+
+The main goal is for application developers/operators to configure themselves how the remote endpoint sends metrics.
```
philipgough, line 20:
```suggestion
The main goal is to enable application developers/operators to self-service the remote write, and configure how the client sends metrics to the remote endpoint.
```

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,121 @@
...
+The Prometheus CRD is extended with 2 new fields (remoteWriteSelector and remoteWriteNamespaceSelector) that define which RemoteWrite resources are associated with this Prometheus instance.
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Prometheus
+metadata:
+  name: example
+  namespace: default
+spec:
+  # RemoteWrite objects to be selected. If nil, nothing will be selected.
```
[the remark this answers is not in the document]
superbrothers: Fixed at 4d8afb2c6.

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,121 @@
...
+  # Namespace(s) to be selected for RemoteWrite discovery. If nil, only check in its own namespace.
+  remoteWriteNamespaceSelector: {}
+  ...
+```
+
+#### Respect enforceNamespaceLabel and excludedFromEnforcement
+
+The RemoteWrite CRD should respect the Prometheus `.spec.enforcedNamespaceLabel` field.
+
+The object’s namespace is added as the last item in the `write_relabel_config` of the remote_write configuration.
```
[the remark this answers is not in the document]
superbrothers: It seems to me that simply replacing the namespace labels does not prevent other tenants from stealing metrics. Should we only keep metrics for the namespace where the RemoteWrite object is created, as follows?

```yaml
write_relabel_configs:
...
- sourceLabels:
  - namespace
  regex: <namespace>
  action: keep
```

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,124 @@
...
+```yaml
+write_relabel_configs:
+...
+- target_label: <enforced-namespace-label>
+  replacement: <namespace>
+```
+
+The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field.
+
+These were proposed at https://github.com/prometheus-operator/prometheus-operator/issues/6508#issuecomment-2058942336.
```
superbrothers, line 104:
```suggestion
```

Documentation/proposals/202407-remote-write.md
superbrothers: Yes, it is still the same. The configuration file is gzipped, but essentially unchanged.

Documentation/proposals/202407-remote-write.md
superbrothers:
```suggestion
    # Secret object references will be resolved in the namespace of this remoteWrite object.
    # In this manifest it is the "default" namespace.
    accessKey:
```

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,124 @@
...
+  namespace: default
+spec:
+  url: "https://aps-workspaces.us-west-2.amazonaws.com/workspaces/<workspace id>/api/v1/remote_write"
+  sigv4:
+    region: us-west-2
+    # Secret object references will be resolved in the namespace of this remoteWrite object.
+    # In this manifest it is the "default" namespace.
+    accessKey:
+      name: aws_access
+      key: access_key
```
superbrothers, line 48:
```suggestion
    accessKey:
      # Secret object references will be resolved in the namespace of this remoteWrite object.
      # In this manifest it is the "default" namespace.
      name: aws_access
      key: access_key
```

Documentation/proposals/202407-remote-write.md
simonpasquier: Thanks! Then it's probably simpler to say this at the beginning and that everything which is described for the Prometheus CRD will work the same for PrometheusAgent.

Documentation/proposals/202407-remote-write.md
simonpasquier: Right so let's omit the things which won't change.
```suggestion
The Prometheus operator will generate the Prometheus configuration including remote_write configuration from the Prometheus CRD and the RemoteWrite resources matching remoteWriteSelector from the namespace(s) selected by remoteWriteNamespaceSelector for additional remote_rewrite configuration.
```

Documentation/proposals/202407-remote-write.md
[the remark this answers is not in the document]
superbrothers: Updated in 29a6f2b21ea9c12efaf0ccf65b737b42995dd649.

Documentation/proposals/202407-remote-write.md
superbrothers: Updated in 29b96b41a6f0ce1a3acd7d4efeac0a58eab22d40.

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,128 @@
...
+  # only.
+  remoteWriteNamespaceSelector: {}
+  ...
+```
+
+#### Respect enforceNamespaceLabel and excludedFromEnforcement
+
+The RemoteWrite CRD should respect the Prometheus `.spec.enforcedNamespaceLabel` field.
+
+The object’s namespace is added as the first item in the `write_relabel_config` of the remote_write configuration.
```
ArthurSens, line 95: Nice approach to respect tenants!

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,128 @@
...
+- sourceLabels:
+  - namespace
+  regex: <namespace>
+  action: keep
+  ...
+```
+
+The item is inserted in the first position by the operator before any other user relabelings then there's no way that users can have access to "foreign" metrics.
+
+The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field.
```
ArthurSens, line 108: This might sound as a silly question coming from a maintainer, but what does `excludedFromEnforcement` do again? 😅

ArthurSens (commented):
One question:

Introducing a new CRD requires that the operator has RBAC permissions to read/list/watch this new resource. What happens if someone updates the operator to a version that has support for RemoteWrite CRD, but still hasn't configured the necessary permissions?

Could we do something similar to what we've done with the ScrapeConfig CRD? (allow starting the operator without permissions, do not reconcile if permissions aren't there)

Documentation/proposals/202407-remote-write.md
simonpasquier: right now it's used to enforce a label (typically "namespace") with the "namespace of origin" value and it's supported for:
* metrics scraped from ServiceMonitor, PodMonitor, Probe objects
* rules from PrometheusRule objects

simonpasquier:
+1 to the suggestion from @ArthurSens about checking the RBAC permissions of the operator. We could also add a feature gate but it's a bit overkill IMHO.

About protecting Prometheus against greedy remote write configs, we could allow the admins to define max values for the queue config settings (like queue capacity, sample age, max number of shards, ...).

Documentation/proposals/202407-remote-write.md
superbrothers:
```suggestion
The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field. It is a list of references to objects to be excluded from `enforcedNamespaceLabel`.
```

superbrothers:
> Could we do something similar to what we've done with the ScrapeConfig CRD? (allow starting the operator without permissions, do not reconcile if permissions aren't there)

@ArthurSens  If it is possible with ScrapeConfig CRD, it should be possible here as well. I will try to achieve this in a similar way to ScrapeConfig.

superbrothers:
> About protecting Prometheus against greedy remote write configs, we could allow the admins to define max values for the queue config settings (like queue capacity, sample age, max number of shards, ...).

@simonpasquier This can also be the case with other resources such as ServiceMonitor, where extreme configurations can affect the Prometheus server. It is difficult to support individual cases. I believe that admins should be responsible for protecting them, e.g. by using a ValidatingAdmissionPolicy.

mviswanathsai:
> Thanks for starting this! My main question would be how admins can protect the Prometheus instance from "expensive" configurations (see also https://github.com/prometheus/prometheus/issues/13174). I don't have a good answer yet.

Not sure if this will help, but I wrote to the OP of that issue and learnt that they are writing everything to one/ a few endpoints and then configuring the endpoints to do the fragmentation. Not sure if we will even be able to enforce something like this. But this goes to show we can limit the number of RemoteWrite CRDs in the cluster if we really wanted to and not directly impact user's abilities. However, it definitely makes their life harder :\

simonpasquier:
> > About protecting Prometheus against greedy remote write configs, we could allow the admins to define max values for the queue config settings (like queue capacity, sample age, max number of shards, ...).
> 
> @simonpasquier This can also be the case with other resources such as ServiceMonitor, where extreme configurations can affect the Prometheus server. It is difficult to support individual cases. I believe that admins should be responsible for protecting them, e.g. by using ValidatingAdmissionPolicy.


You can enforce scrape limits for ServiceMonitor. I don't want to dismiss your work here but we should document which options we have in the proposal :)

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,128 @@
...
+```yaml
+write_relabel_configs:
+- sourceLabels:
+  - namespace
+  regex: <namespace>
+  action: keep
+  ...
+```
+
+The item is inserted in the first position by the operator before any other user relabelings then there's no way that users can have access to "foreign" metrics.
```
mviswanathsai, line 106: Pretty neat. But I am still not sure about how reliable this would be.

mviswanathsai: Might be a nit, but for clarity, could we elaborate on what we mean by "foreign" metrics?

superbrothers:
```suggestion
The item is inserted in the first position by the operator before any other user relabelings, so there's no way that users can have access to metrics labeled with namespaces other than the one in which the RemoteWrite object is created.
```

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Goals
+
+The main goal is to enable application developers/operators to self-service the remote write, and configure how the client sends metrics to the remote endpoint.
+
+This means exposing a new CRD to configure Prometheus remote_write configuration.
+
+## Non-goals
+
+We do not cover "RemoteRead" CRD here. We can still implement it later if needed.
```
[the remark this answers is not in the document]
superbrothers:
```suggestion
Prometheus has two APIs for integrating with remote storage: 'remote write' and 'remote read'. Since there has been no confirmed demand to treat remote read as a CRD, we do not cover "RemoteRead" CRD here. We can still implement it later if needed.
```

superbrothers:
@simonpasquier Do I need to make any other changes?

simonpasquier (commented):
Thanks for the updated version! I'll get back to you this week hoperfully.
cc @philipgough who commented earlier :)

dublx:
Nice, keeping an eye on the progress here, we have a need for this as well.

simonpasquier (commented):
From my POV, it's ok to merge once you've the small suggested paragraph on admin limits.

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,128 @@
...
+The item is inserted in the first position by the operator before any other user relabelings, so there's no way that users can have access to metrics labeled with namespaces other than the one in which the RemoteWrite object is created.
+
+The RemoteWrite CRD should also respect the Prometheus `.spec.excludedFromEnforcement` field. It is a list of references to objects to be excluded from `enforcedNamespaceLabel`.
+
+### Configuration generation
+
+The Prometheus operator will generate the Prometheus configuration including remote_write configuration from the Prometheus CRD and the RemoteWrite resources matching remoteWriteSelector from the namespace(s) selected by remoteWriteNamespaceSelector for additional remote_rewrite configuration.
+
+The operator will respect the --namespaces and --deny-namespaces flags when looking for RemoteWrite objects.
+
```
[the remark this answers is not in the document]
superbrothers: Thanks!

Documentation/proposals/202407-remote-write.md
```diff
@@ -0,0 +1,147 @@
...
+### Administrative limits
+
+Similar to what exists for scrape resources (e.g. `enforcedSampleLimit` for  `ServiceMonitor`), we will allow the Prometheus resource's owners to setup upper-bound limits on the remote write options. For example, it will be possible to disable metadata sending or limit the queue capacity:
```
superbrothers, line 118:
```suggestion
Similar to what exists for scrape resources (e.g. `enforcedSampleLimit` for `ServiceMonitor`), we will allow the Prometheus resource's owners to setup upper-bound limits on the remote write options. For example, it will be possible to disable metadata sending or limit the queue capacity:
```

simonpasquier (approved):
Thanks!

PR #6787, opened 2024-08-01
Allow Prometheus Agent to update, delete, and handle changes to DaemonSet object

haanhvu:
## Description

Allow Prometheus Agent update, delete, and handle changes to DaemonSet object

## Type of change

## Verification
E2e that update/delete Prometheus Agent DaemonSet

## Changelog entry

Allow updating and deleting Prometheus Agent DaemonSet

### DIFF
--- a/pkg/operator/resource_reconciler.go
+++ b/pkg/operator/resource_reconciler.go
@@ -135,7 +135,8 @@ func NewResourceReconciler(
 
 	qname := strings.ToLower(kind)
 
-	for _, t := range []string{"StatefulSet", "DaemonSet", kind} {
+	// TODO: Support reconciling metrics for DaemonSet resource
+	for _, t := range []string{"StatefulSet", kind} {
 		for _, e := range []HandlerEvent{AddEvent, DeleteEvent, UpdateEvent} {
 			metrics.TriggerByCounter(t, e)
 		}
@@ -357,7 +358,6 @@ func (rr *ResourceReconciler) onDaemonSetAdd(ds *appsv1.DaemonSet) {
 	}
 
 	rr.logger.Debug("DaemonSet added")
-	rr.metrics.TriggerByCounter("DaemonSet", AddEvent).Inc()
 
 	rr.EnqueueForReconciliation(obj)
 }
@@ -409,7 +409,6 @@ func (rr *ResourceReconciler) onDaemonSetUpdate(old, cur *appsv1.DaemonSet) {
 	}
 
 	rr.logger.Debug("DaemonSet updated")
-	rr.metrics.TriggerByCounter("DaemonSet", UpdateEvent).Inc()
 	if !rr.hasStateChanged(old, cur) {
 		// If the daemonset state (spec, labels or annotations) hasn't
 		// changed, the operator can only update the status subresource instead
@@ -441,7 +440,6 @@ func (rr *ResourceReconciler) onDaemonSetDelete(ds *appsv1.DaemonSet) {
 	}
 
 	rr.logger.Debug("DaemonSet delete")
-	rr.metrics.TriggerByCounter("DaemonSet", DeleteEvent).Inc()
 
 	rr.EnqueueForReconciliation(obj)
 }
--- a/pkg/operator/resource_reconciler.go
+++ b/pkg/operator/resource_reconciler.go
@@ -410,7 +410,6 @@ func (rr *ResourceReconciler) onDaemonSetUpdate(old, cur *appsv1.DaemonSet) {
 
 	rr.logger.Debug("DaemonSet updated")
 	rr.metrics.TriggerByCounter("DaemonSet", UpdateEvent).Inc()
-	// TODO: Uncomment this when Prometheus Agent DaemonSet's status has been supported.
 	if !rr.hasStateChanged(old, cur) {
 		// If the daemonset state (spec, labels or annotations) hasn't
 		// changed, the operator can only update the status subresource instead
--- a/pkg/operator/resource_reconciler.go
+++ b/pkg/operator/resource_reconciler.go
@@ -265,12 +265,12 @@ func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
 
 // OnUpdate implements the cache.ResourceEventHandler interface.
 func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
-	switch cur.(type) {
+	switch v := cur.(type) {
 	case *appsv1.DaemonSet:
-		rr.onDaemonSetUpdate(old.(*appsv1.DaemonSet), cur.(*appsv1.DaemonSet))
+		rr.onDaemonSetUpdate(old.(*appsv1.DaemonSet), v)
 		return
 	case *appsv1.StatefulSet:
-		rr.onStatefulSetUpdate(old.(*appsv1.StatefulSet), cur.(*appsv1.StatefulSet))
+		rr.onStatefulSetUpdate(old.(*appsv1.StatefulSet), v)
 		return
 	}
 
@@ -411,13 +411,14 @@ func (rr *ResourceReconciler) onDaemonSetUpdate(old, cur *appsv1.DaemonSet) {
 	rr.logger.Debug("DaemonSet updated")
 	rr.metrics.TriggerByCounter("DaemonSet", UpdateEvent).Inc()
 	// TODO: Uncomment this when Prometheus Agent DaemonSet's status has been supported.
-	/*if !rr.hasStateChanged(old, cur) {
+	if !rr.hasStateChanged(old, cur) {
 		// If the daemonset state (spec, labels or annotations) hasn't
 		// changed, the operator can only update the status subresource instead
 		// of doing a full reconciliation.
-		rr.EnqueueForStatus(obj)
+		// TODO: Uncomment this when Prometheus Agent DaemonSet's status has been supported.
+		// rr.EnqueueForStatus(obj)
 		return
-	}*/
+	}
 
 	rr.EnqueueForReconciliation(obj)
 }
--- a/pkg/operator/resource_reconciler.go
+++ b/pkg/operator/resource_reconciler.go
@@ -381,14 +381,13 @@ func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet) 
 	rr.logger.Debug("StatefulSet updated")
 	rr.metrics.TriggerByCounter("StatefulSet", UpdateEvent).Inc()
 
-	// TODO: Uncomment this when Prometheus Agent DaemonSet's status has been supported.
-	/*if !rr.hasStateChanged(old, cur) {
+	if !rr.hasStateChanged(old, cur) {
 		// If the statefulset state (spec, labels or annotations) hasn't
 		// changed, the operator can only update the status subresource instead
 		// of doing a full reconciliation.
 		rr.EnqueueForStatus(obj)
 		return
-	}*/
+	}
 
 	rr.EnqueueForReconciliation(obj)
 }
@@ -411,14 +410,14 @@ func (rr *ResourceReconciler) onDaemonSetUpdate(old, cur *appsv1.DaemonSet) {
 
 	rr.logger.Debug("DaemonSet updated")
 	rr.metrics.TriggerByCounter("DaemonSet", UpdateEvent).Inc()
-
-	if !rr.hasStateChanged(old, cur) {
+	// TODO: Uncomment this when Prometheus Agent DaemonSet's status has been supported.
+	/*if !rr.hasStateChanged(old, cur) {
 		// If the daemonset state (spec, labels or annotations) hasn't
 		// changed, the operator can only update the status subresource instead
 		// of doing a full reconciliation.
 		rr.EnqueueForStatus(obj)
 		return
-	}
+	}*/
 
 	rr.EnqueueForReconciliation(obj)
 }
--- a/pkg/operator/resource_reconciler.go
+++ b/pkg/operator/resource_reconciler.go
@@ -34,14 +34,14 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-// Syncer knows how to synchronize statefulset-based resources.
+// Syncer knows how to synchronize statefulset-based or daemonset-based resources.
 type Syncer interface {
 	// Sync the state of the object identified by its key.
 	Sync(context.Context, string) error
 	// UpdateStatus updates the status of the object identified by its key.
 	UpdateStatus(context.Context, string) error
-	// Resolve returns the resource associated to the statefulset.
-	Resolve(*appsv1.StatefulSet) metav1.Object
+	// Resolve returns the resource owning the workload object (either StatefulSet or DaemonSet).
+	Resolve(obj interface{}) metav1.Object
 }
 
 // ReconcilerMetrics tracks reconciler metrics.
@@ -135,7 +135,7 @@ func NewResourceReconciler(
 
 	qname := strings.ToLower(kind)
 
-	for _, t := range []string{"StatefulSet", kind} {
+	for _, t := range []string{"StatefulSet", "DaemonSet", kind} {
 		for _, e := range []HandlerEvent{AddEvent, DeleteEvent, UpdateEvent} {
 			metrics.TriggerByCounter(t, e)
 		}
@@ -234,8 +234,12 @@ func (rr *ResourceReconciler) objectKey(obj interface{}) (string, bool) {
 // OnAdd implements the cache.ResourceEventHandler interface.
 func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
 
-	if _, ok := obj.(*appsv1.StatefulSet); ok {
-		rr.onStatefulSetAdd(obj.(*appsv1.StatefulSet))
+	switch v := obj.(type) {
+	case *appsv1.DaemonSet:
+		rr.onDaemonSetAdd(v)
+		return
+	case *appsv1.StatefulSet:
+		rr.onStatefulSetAdd(v)
 		return
 	}
 
@@ -261,7 +265,11 @@ func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
 
 // OnUpdate implements the cache.ResourceEventHandler interface.
 func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
-	if _, ok := cur.(*appsv1.StatefulSet); ok {
+	switch cur.(type) {
+	case *appsv1.DaemonSet:
+		rr.onDaemonSetUpdate(old.(*appsv1.DaemonSet), cur.(*appsv1.DaemonSet))
+		return
+	case *appsv1.StatefulSet:
 		rr.onStatefulSetUpdate(old.(*appsv1.StatefulSet), cur.(*appsv1.StatefulSet))
 		return
 	}
@@ -301,8 +309,12 @@ func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
 
 // OnDelete implements the cache.ResourceEventHandler interface.
 func (rr *ResourceReconciler) OnDelete(obj interface{}) {
-	if _, ok := obj.(*appsv1.StatefulSet); ok {
-		rr.onStatefulSetDelete(obj.(*appsv1.StatefulSet))
+	switch v := obj.(type) {
+	case *appsv1.DaemonSet:
+		rr.onDaemonSetDelete(v)
+		return
+	case *appsv1.StatefulSet:
+		rr.onStatefulSetDelete(v)
 		return
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/prometheus/agent/operator.go
```diff
@@ -481,8 +481,7 @@ func (c *Operator) addHandlers() {
 
 	c.ssetInfs.AddEventHandler(c.rr)
 
-	// TODO: implement proper event handler for Daemonset.
-	// c.dsetInfs.AddEventHandler(c.rr)
+	c.dsetInfs.AddEventHandler(c.rr)
```
simonpasquier, line 484: you need to test c.destInfs != nil :)

haanhvu: aha!

pkg/operator/resource_reconciler.go
```diff
@@ -374,6 +401,36 @@ func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet)
...
+	}
+
+	level.Debug(rr.logger).Log("msg", "DaemonSet updated")
+	rr.metrics.TriggerByCounter("DaemonSet", UpdateEvent).Inc()
+
+	if !rr.hasStateChanged(old, cur) {
+		// If the statefulset state (spec, labels or annotations) hasn't
+		// changed, the operator can only update the status subresource instead
+		// of doing a full reconciliation.
+		rr.EnqueueForStatus(obj)
```
haanhvu, line 427: @simonpasquier Now we haven't supported status subresource for DaemonSet yet, so we cannot use it for now right?

haanhvu:
@simonpasquier I added a prometheus agent delete e2e in the last commit to see if it's work. It passed now. Do you want to move this to the next PR like we discussed in the meeting?

pkg/operator/resource_reconciler.go
haanhvu: Ok I think I got it now. We haven't supported status subresource for Prometheus Agent in DaemonSet mode. But DaemonSet's status has already been supported by K8s. So we can `EnqueueForStatus()` for DaemonSet, right?

pkg/operator/resource_reconciler.go
```diff
@@ -245,6 +245,11 @@ func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
 		return
 	}
 
+	if _, ok := obj.(*appsv1.DaemonSet); ok {
```
simonpasquier, line 248: (nit) we could use a type switch

````
switch v := obj.(type) {
    case *appsv1.DaemonSet:
        rr.onDaemonSetAdd(v)
        return
    case *appsv1.StatefulSet:
        rr.onStatefulSetAdd(v)
        return
}

File: pkg/operator/resource_reconciler.go (hunk 2 above)
Comment (simonpasquier) [reply pkg/operator/resource_reconciler.go:427]: AFAICT we don't support the status subresource when mode=daemonset so it's not worth calling EnqueueForStatus() but keep the if loop and add a TODO comment that we can call EnqueueForStatus() when implementing the status.

File: pkg/prometheus/agent/operator.go [hunk 4]
```diff
@@ -550,8 +551,18 @@ func (c *Operator) addHandlers() {
...
+	var key string
+	var ok bool
+
+	if c.daemonSetFeatureGateEnabled {
+		ds := obj.(*appsv1.DaemonSet)
+		key, ok = c.accessor.MetaNamespaceKey(ds)
+	} else {
+		ss := obj.(*appsv1.StatefulSet)
+		key, ok = c.accessor.MetaNamespaceKey(ss)
+	}
```
simonpasquier, line 564: Resolve() should return the PrometheusAgent object backing the statefulset or daemonset but MetaNamespaceKey() works on generic Kubernetes objects. Hence the type check should be done later to decide whether we need to call statefulSetKeyToPrometheusAgentKey() or daemonSetKeyToPrometheusAgentKey() (to be implemented).

simonpasquier (commented):
Nice! A few comments but it looks good overall.

pkg/prometheus/agent/operator.go
haanhvu: Yeah actually now I reread the operator's code there're still some logics missing to support updating PA DaemonSet, beside this one that you pointed out. No surprise why update e2e is failing^^ Let me try to complete these logics

test/e2e/prometheusagent_test.go
```diff
@@ -204,3 +209,81 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+	}
+
+	p, err = framework.CreatePrometheusAgentAndWaitUntilReady(context.Background(), ns, p)
+	require.NoError(t, err)
+
+	pods, err := framework.KubeClient.CoreV1().Pods(ns).List(ctx, pa.ListOptions(name))
+	require.NoError(t, err)
+
+	res := pods.Items[0].Spec.Containers[0].Resources
+	require.Equal(t, res, p.Spec.Resources)
```
ArthurSens, line 246: Not sure what's the goal when listing pods here, could you explain?

test/e2e/prometheusagent_test.go
```diff
@@ -204,3 +209,81 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
+				Resources: v1.ResourceRequirements{
+					Requests: v1.ResourceList{
+						v1.ResourceMemory: resource.MustParse("200Mi"),
+					},
+				},
+			},
+		},
+	)
+	require.NoError(t, err)
```
ArthurSens, line 263: We're adding `OnDaemonsetAdd`, which means we want to react to Daemonset changes. We want to patch daemonset, not a PrometheusAgent, right?

ArthurSens: Feel free to tell me if I'm missing something, but the flow we want to check is:

* PrometheusAgent is created in daemonsetMode.
* Operator creates a daemonset
* We manually change the daemonset by calling k8s API
* Operator reverts the change made by us, because it is different from the desired spec.
* We verify that the Operator reacted by our changed, by making sure things were reverted

ArthurSens (commented):
Very good start! I've added a comment regarding the new test, but please tell me if I'm missing the point completely!

test/e2e/prometheusagent_test.go
```diff
@@ -183,3 +188,81 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+	var pollErr error
+	err = wait.PollUntilContextTimeout(context.Background(), 30*time.Second, 30*time.Minute, false, func(ctx context.Context) (bool, error) {
+		pods, err := framework.KubeClient.CoreV1().Pods(ns).List(ctx, pa.ListOptions(name))
+		if err != nil {
+			pollErr = fmt.Errorf("failed to list Prometheus Agent DaemonSet pods: %w", err)
+			return false, nil
+		}
+
+		if len(pods.Items) != 1 {
+			pollErr = fmt.Errorf("expected 1 pod, actual number of pods: %d", len(pods.Items))
```
haanhvu, line 253: @simonpasquier @ArthurSens The test is failing in this line: `expected 1 pod, actual number of pods: 2`

I'll investigate later today or tomorrow. If you have time pls investigate too!

As I said here, the test follow a similar Prometheus test, to check if Prometheus Agent can be updated: https://github.com/prometheus-operator/prometheus-operator/pull/6787#discussion_r1706605187

haanhvu: Actually now I realize. We're running the test on 2 nodes, so we should actually expect 2 pods, right? :thinking:

test/e2e/prometheusagent_test.go
simonpasquier: To verify that the resource settings are propagated to the pods? But I suggest that you check the daemonset directly.

test/e2e/prometheusagent_test.go
simonpasquier: What @ArthurSens describes would be the correct thing to test but I think that we're not doing it anyway for the other CRDs :shrug: 
This new test is still useful but it's not directly related to this change.

test/e2e/prometheusagent_test.go
simonpasquier: See my first comment: it's fine to check the daemonset definition rather than the pods.

test/e2e/prometheusagent_test.go
ArthurSens: > @ArthurSens Actually this test follows the similar Prometheus test here:
> 
> https://github.com/prometheus-operator/prometheus-operator/blob/4e254f018dd0ec9624409ba429c04de192ef8db0/test/e2e/prometheus_test.go#L856
> 
> Basically I would like to test if we can update Prometheus Agent DaemonSet.

I see! Like @simonpasquier said, your change is not about updates to the Agent resource. You're adding handlers to react on daemonset changes and therefore the test will be different :)

haanhvu:
@simonpasquier @ArthurSens I resolved all reviews. As I said in https://github.com/prometheus-operator/prometheus-operator/pull/6787#discussion_r1711414291 I checked DaemonSet resource in both update and delete e2e.

Tests all passed now. Pls take another review. Thanks!

pkg/operator/resource_reconciler.go
```diff
@@ -261,7 +265,11 @@ func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
 
 // OnUpdate implements the cache.ResourceEventHandler interface.
 func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
-	if _, ok := cur.(*appsv1.StatefulSet); ok {
+	switch cur.(type) {
```
simonpasquier, line 268: (nit) you could use the same `switch v := obj.(type)` trick and pass `v` to the function.

pkg/operator/resource_reconciler.go
```diff
@@ -357,12 +381,43 @@ func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet)
 	rr.logger.Debug("StatefulSet updated")
 	rr.metrics.TriggerByCounter("StatefulSet", UpdateEvent).Inc()
 
-	if !rr.hasStateChanged(old, cur) {
+	// TODO: Uncomment this when Prometheus Agent DaemonSet's status has been supported.
+	/*if !rr.hasStateChanged(old, cur) {
```
simonpasquier, line 385: why is this part commented? `onStatefulSetUpdate()` shouldn't be changed?

pkg/operator/resource_reconciler.go
```diff
@@ -357,12 +381,43 @@ func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet)
...
+	}
+
+	rr.logger.Debug("DaemonSet updated")
+	rr.metrics.TriggerByCounter("DaemonSet", UpdateEvent).Inc()
+
+	if !rr.hasStateChanged(old, cur) {
+		// If the daemonset state (spec, labels or annotations) hasn't
+		// changed, the operator can only update the status subresource instead
+		// of doing a full reconciliation.
+		rr.EnqueueForStatus(obj)
```
simonpasquier, line 419: this is where we'd need a TODO?

pkg/prometheus/agent/operator.go
```diff
@@ -550,13 +551,21 @@ func (c *Operator) addHandlers() {
...
+	key, ok := c.accessor.MetaNamespaceKey(obj)
 	if !ok {
 		return nil
 	}
 
-	match, promKey := statefulSetKeyToPrometheusAgentKey(key)
+	var match bool
+	var promKey string
+
+	if c.daemonSetFeatureGateEnabled {
```
simonpasquier, line 563: we also need to cast the given object to know if we should call the statefulset or daemonset function.

pkg/operator/resource_reconciler.go
haanhvu: How? Here we don't have `obj` interface, we have two `old` and `cur` interface :thinking:

haanhvu:
@simonpasquier @ArthurSens As discussed in Slack, I added all the required tests. They all passed now. Pls review!

There are many duplications in the tests right now. Once we agree the tests are correct I'll refactor those duplications^^

pkg/operator/resource_reconciler.go
```diff
@@ -261,7 +265,11 @@ func (rr *ResourceReconciler) OnAdd(obj interface{}, _ bool) {
...
 func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
-	if _, ok := cur.(*appsv1.StatefulSet); ok {
+	switch cur.(type) {
+	case *appsv1.DaemonSet:
+		rr.onDaemonSetUpdate(old.(*appsv1.DaemonSet), cur.(*appsv1.DaemonSet))
+		return
+	case *appsv1.StatefulSet:
 		rr.onStatefulSetUpdate(old.(*appsv1.StatefulSet), cur.(*appsv1.StatefulSet))
 		return
 	}
```
ArthurSens, line 275:
```suggestion
	switch v := cur.(type) {
	case *appsv1.DaemonSet:
		rr.onDaemonSetUpdate(old.(*appsv1.DaemonSet), v)
		return
	case *appsv1.StatefulSet:
		rr.onStatefulSetUpdate(old.(*appsv1.StatefulSet), v)
		return
	}
```

similar to what you did with `OnDelete`

pkg/operator/resource_reconciler.go
ArthurSens: I'm unresolving this comment because I'm also a bit confused here. 

I think we still want to keep the `if !rr.hasStateChanged(old,cur) ` here, and only comment `rr.EnqueueForStatus(obj)` because there's an early return inside the if.

In other words, if the old and new daemonset are fundamentally equal, we'd like to avoid calling `rr.EnqueueForReconciliation` below.

pkg/prometheus/agent/operator.go
```diff
@@ -550,13 +551,24 @@ func (c *Operator) addHandlers() {
...
 
-	match, promKey := statefulSetKeyToPrometheusAgentKey(key)
+	var match bool
+	var promKey string
+
+	switch obj.(type) {
+	case *appsv1.DaemonSet:
+		if c.daemonSetFeatureGateEnabled {
+			match, promKey = daemonSetKeyToPrometheusAgentKey(key)
+		}
```
ArthurSens, line 567: I still need to investigate this, but isn't the daemonset informer only created if the featuregate is enabled? I.e. it's impossible to `Resolve` to be called by a daemonset change if the featuregate is disabled.

If what I'm saying is correct, we probably can remove the conditional here 🤔

pkg/operator/resource_reconciler.go
```diff
@@ -135,7 +135,7 @@ func NewResourceReconciler(
 
 	qname := strings.ToLower(kind)
 
-	for _, t := range []string{"StatefulSet", kind} {
+	for _, t := range []string{"StatefulSet", "DaemonSet", kind} {
```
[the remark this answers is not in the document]
ArthurSens: Testing locally I can see some new metrics being exposed that don't make much sense 😬 
image

Maybe we could revert this change?

pkg/prometheus/agent/operator.go
ArthurSens: I've just tested locally without the conditional and it seems to have worked without problems 🤔

pkg/prometheus/agent/operator.go
```diff
@@ -591,6 +603,19 @@ func statefulSetKeyToPrometheusAgentKey(key string) (bool, string) {
 	return true, matches[0][1] + "/" + matches[0][2]
 }
 
+func daemonSetKeyToPrometheusAgentKey(key string) (bool, string) {
```
ArthurSens, line 606: nit:

A comment explaining what this function does would help a ton, an example in the comment would be even better. I know `statefulSetKeyToPrometheusAgentKey` doesn't have a comment as well, but a comment there would also be useful

pkg/prometheus/agent/operator.go
```diff
@@ -683,6 +713,29 @@ func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv
 		return nil
 	}
 
+	err = k8sutil.UpdateDaemonSet(ctx, dsetClient, dset)
+	sErr, ok := err.(*apierrors.StatusError)
+
```
ArthurSens, line 718: I've noticed that we don't compare input hashes for daemonsets while we do for statefulsets. The tests pass, so I'm not too worried, but I'm curious how this impacts the update frequency.

If this proves to be a real problem, it's ok to fix in a follow up PR

ArthurSens (commented):
I've tested locally and everything seems to work just fine 🥳 

I've added a few comments, but I'm confident we'll get it ready for merge tomorrow or early next week

pkg/operator/resource_reconciler.go
ArthurSens: I think simon had something like this https://github.com/prometheus-operator/prometheus-operator/pull/6787#discussion_r1718719044 in mind

pkg/prometheus/agent/operator.go
haanhvu: Yeah we discussed some time ago in Slack, I and Simon both agreed to do this later (after MVP): https://cloud-native.slack.com/archives/C071STDT4FP/p1717580605023279

pkg/prometheus/agent/operator.go
haanhvu: Yeah makes sense. I'll remove this condition

pkg/operator/resource_reconciler.go
haanhvu: Yeah I got it now. Will uncomment the return part

pkg/operator/resource_reconciler.go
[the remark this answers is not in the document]
haanhvu: This is still necessary for Prometheus Agent. So how about only doing this for Prometheus Agent controller?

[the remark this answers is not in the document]
haanhvu: Also, in case we want to keep this for Prometheus Agent, is there any way to check these metrics in e2e?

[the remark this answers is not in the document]
simonpasquier: I suggest that we don't record the "Daemonset" label for now and leave a TODO. The only downside is that folks using the daemonset mode won't see the counters with 0 value.

simonpasquier (commented):
one small nit otherwise lgtm!

pkg/operator/resource_reconciler.go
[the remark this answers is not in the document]
ArthurSens: If they have the feature-flag created-timestamp-zero-ingestion enabled in their server, they will see 0 value even if we don't initialize metrics like we do today :)

haanhvu:
@simonpasquier @ArthurSens I resolved all reviews. Some remaining questions:

-  Seems like `statefulSetKeyToPrometheusAgentKey()` can be used for DaemonSet mode too right? If it's true we can remove `daemonSetKeyToPrometheusAgentKey()`? :thinking:
- Can you check the timeouts in e2e? I set them very generously, but not sure if necessary :smile: 
- Have you local tested for all cases: update ds, delete ds, reconcile ds update, reconcile ds delete?

test/e2e/prometheusagent_test.go
```diff
@@ -183,3 +187,181 @@ func testPrometheusAgentStatusScale(t *testing.T) {
...
+						v1.ResourceMemory: resource.MustParse("200Mi"),
+					},
+				},
+			},
+		},
+	)
+	require.NoError(t, err)
+
+	var pollErr error
+	err = wait.PollUntilContextTimeout(context.Background(), 30*time.Second, 30*time.Minute, false, func(ctx context.Context) (bool, error) {
```
simonpasquier, line 244: usually 5 minutes for the timeout is more than enough. Polling more often (every second) is also not an issue.

simonpasquier:
Tested locally and LGTM!

ArthurSens (approved):
:shipit:

pkg/operator/resource_reconciler.go
```diff
@@ -135,7 +135,8 @@ func NewResourceReconciler(
 
 	qname := strings.ToLower(kind)
 
-	for _, t := range []string{"StatefulSet", "DaemonSet", kind} {
+	// TODO: Support reconciling metrics for DaemonSet resource
```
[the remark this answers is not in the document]
simonpasquier: @ArthurSens IIUC your point you're alluding to the created timestamp for counters. IMHO we can have the discussion in another issue/PR.
```suggestion
	// TODO: initialize metrics for the DaemonSet label.
```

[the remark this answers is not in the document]
ArthurSens: Yes, you're right! I'll commit your suggestion and merge the PR :)

ArthurSens (approved):

haanhvu:
Auto-merge doesn't work. Probably because I didn't rebase on main. I'll do it now and https://github.com/prometheus-operator/prometheus-operator/pull/6787#discussion_r1734907384 too

haanhvu:
@simonpasquier I decreased timeouts to 5min and the tests failed :sweat_smile: let's see if 10min work

test/e2e/prometheusagent_test.go
haanhvu: > usually 5 minutes for the timeout is more than enough. Polling more often (every second) is also not an issue.

@simonpasquier I tried different numbers from 5 to 20. 20s poll with 20min timeout is what works for me. Lower numbers failed the tests.

haanhvu:
> usually 5 minutes for the timeout is more than enough. Polling more often (every second) is also not an issue.

@simonpasquier I tried different numbers from 5 to 20. 20s poll with 20min timeout is what works for me. Lower numbers failed the tests.

ArthurSens (approved):

Issue #6774, opened 2024-07-24
Native histogram configuration support in {Pod,Service}Monitor


faustuzas:
### Component(s)

ServiceMonitor, PodMonitor

### What is missing? Please describe.

Prometheus has recently introduced a new feature called "native histograms" (detailed documentation available [here](https://docs.google.com/document/d/1ch6ru8GKg03N02jRjYriurt-CZqUVY09evPg6yKTA1s/edit#heading=h.fxf1i4tcd03i)). This feature includes several configuration options in both the static file-based target scrape config and global configuration.

Given that the native histogram public API appears to have reached a stable state with no anticipated significant changes, it seems appropriate to incorporate support for these configurations into {Pod,Service}Monitor. This addition would enable dynamic Kubernetes-based applications to configure and utilize native histograms effectively.

—

I propose adding the following configuration options to {Pod,Service}Monitor, corresponding to those described in the [Prometheus docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/) and other relevant sources:

1. enableNativeHistograms (bool, Optional, default: false)
   - Corresponds to the `native-histogram` feature flag of the Prometheus server. It changes Prometheus to prefer Protobuf exposition format when scraping targets and starts ingesting native histograms if any are found.
   - Rationale: Per-target control is beneficial due to potential "le" label formatting discrepancies between TEXT and Protobuf exposition formats ("3" vs "3.0"), which could impact queries. (For more details, refer to the [Prometheus feature flags docs](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms) under "Note about the format of le and quantile label values".)
   - Enabling this feature globally in a large, diverse system may have unintended consequences and reduce adoption rates.

2. scrapeClassicHistograms (bool, Optional, default: false)
   - Controls whether to scrape a classic histogram that is also exposed as a native histogram.

3. nativeHistogramBucketLimit (int, Optional, default: 0)
   - Limits the total number of positive and negative buckets allowed in a single native histogram.
   - If exceeded, the entire scrape will be treated as failed.
   - A value of 0 indicates no limit.

4. convertClassicHistograms (bool, Optional, default: false)
   - Controls the conversion of classic histograms to Native Histogram Custom Buckets (NHCB).
   - Note: While this option is not yet present in mainline Prometheus, it is described in this [Prometheus issue](https://github.com/prometheus/prometheus/issues/13532) and is expected to be added soon.

—

[...]

PR #6796, opened 2024-08-02
feat: support add native histogram config

dongjiang1989:
## Description

Close: #6774 
support add native histogram config

## Type of change

## Verification

## Changelog entry

```release-note
support add native histogram config
```


simonpasquier (commented):
should we create a custom type and embed it in each struct (probe, podmonitor and servicemonitor)?

dongjiang1989:
> should we create a custom type and embed it in each struct (probe, podmonitor and servicemonitor)?

Thanks @simonpasquier . Got it. I'll fix it.

dongjiang1989:
@simonpasquier @slashpai  PTAL, thanks

pkg/apis/monitoring/v1/types.go
```diff
@@ -880,3 +880,27 @@ const (
...
+	//
+	// +optional
+	NativeHistogramBucketLimit *uint64 `json:"nativeHistogramBucketLimit,omitempty"`
+
+	// If the growth factor of one bucket to the next is smaller than this,
+	// buckets will be merged to increase the factor sufficiently.
+	// It requires Prometheus >= v2.50.0.
+	//
+	// +optional
+	NativeHistogramMinBucketFactor *float64 `json:"nativeHistogramMinBucketFactor,omitempty"`
```
simonpasquier, line 905: float fields are strongly discouraged because they can't be round-tripped (e.g. the returned value might be different from the value set by the client which could break reconcilers). The recommendation is to use `resource.Quantity` instead.

pkg/apis/monitoring/v1/podmonitor_types.go
```diff
@@ -123,6 +123,11 @@ type PodMonitorSpec struct {
 	//
 	// +optional
 	LabelValueLengthLimit *uint64 `json:"labelValueLengthLimit,omitempty"`
+
+	// Optional NativeHistogramConfig.
+	// +optional
+	NativeHistogramConfig `json:",inline"`
```
simonpasquier, line 129: the field being inlined, I think that the comment isn't used by the doc/API generation. Same for all the other types.
```suggestion
	NativeHistogramConfig `json:",inline"`
```

pkg/apis/monitoring/v1/types.go
dongjiang1989: Thanks @simonpasquier .
Change `*float64 -> *resource.Quantity` done.

dongjiang1989:
@simonpasquier Please re-check it, thanks.

simonpasquier (approved):
Thanks!

PR #6819, opened 2024-08-07
replace gokitlog in the resource reconciler

perebaj:
## Description

The `ResourceReconciler` structure is used across all operators. I duplicated the logger and replace their names to a meaningful one (`goKitLogger`), like i'm doing the in the past PRs.

But in the end, this PR just aim to change the log library in the following file: `pgk/operator/resource_reconciler.go`

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/operator/resource_reconciler.go
+++ b/pkg/operator/resource_reconciler.go
@@ -17,12 +17,11 @@ package operator
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"reflect"
 	"strings"
 	"time"
 
-	"github.com/go-kit/log"
-	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/sync/errgroup"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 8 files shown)


pkg/operator/resource_reconciler.go
```diff
@@ -65,7 +64,7 @@ type ReconcilerMetrics interface {
 // ResourceReconciler will trigger object and status reconciliations based on
 // the events received from the informer.
 type ResourceReconciler struct {
-	logger log.Logger
+	logger *slog.Logger
```
ArthurSens, line 67: Any special reason to change this to a pointer?

ArthurSens (commented):
Just one question

ArthurSens:
We need to be careful with this one because it definitely conflicts with #6787

pkg/operator/resource_reconciler.go
perebaj: or maybe you confused *slog.Logger with *log.Logger?

perebaj:
> We need to be careful with this one because it definitely conflicts with #6787

Yep, let's merge the other one first, then I fix the conflicts afterward

ArthurSens (approved):
Actually I think that one is a bit further from being merged, let's merge this one and @haanhvu can easily solve the conflicts on her PR :)

perebaj:
> Actually I think that one is a bit further from being merged, let's merge this one and @haanhvu can easily solve the conflicts on her PR :)

Nice!

PR #6825, opened 2024-08-08
replace go kit log in resource selector

perebaj:
## Description

- replace go kit log in resource_selector.go
- replace go kit log in accessor.go
- remove unused err validation

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -123,6 +123,7 @@ func WithStorageClassValidation() ControllerOption {
 // New creates a new controller.
 func New(ctx context.Context, restConfig *rest.Config, c operator.Config, goKitLogger log.Logger, logger *slog.Logger, r prometheus.Registerer, options ...ControllerOption) (*Operator, error) {
 	goKitLogger = log.With(goKitLogger, "component", controllerName)
+	logger = logger.With("component", controllerName)
 
 	client, err := kubernetes.NewForConfig(restConfig)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


test/framework/framework.go
```diff
@@ -97,9 +97,6 @@ func New(kubeconfig, opImage, exampleDir, resourcesDir string, operatorVersion s
 	}
 
 	httpc := cli.CoreV1().RESTClient().(*rest.RESTClient).Client
-	if err != nil {
-		return nil, fmt.Errorf("creating http-client failed: %w", err)
-	}
```
perebaj, line 102: this error validation here doesn't make sense. For this reason I'm removing it

pkg/alertmanager/operator.go
```diff
@@ -123,6 +123,7 @@ func WithStorageClassValidation() ControllerOption {
 // New creates a new controller.
 func New(ctx context.Context, restConfig *rest.Config, c operator.Config, goKitLogger log.Logger, logger *slog.Logger, r prometheus.Registerer, options ...ControllerOption) (*Operator, error) {
 	goKitLogger = log.With(goKitLogger, "component", controllerName)
+	logger = logger.With("component", controllerName)
```
ArthurSens, line 126: oh thanks for fixing that, somehow I missed when reviewing the other PR

test/framework/framework.go
ArthurSens: Ideally, we keep the scope of PRs consistent. I.e. this change would make more sense in another PR. This way if we need to revert one thing, we don't revert unrelated changes.

Since this is suuuuper small, let's just call it nitpicking and approve it :P

ArthurSens (approved):
Thanks!

test/framework/framework.go
perebaj: Ok, If I found something like this one, I will address in a different PR. Thanks for the hint.

PR #6832, opened 2024-08-12
chore: replace goKitLogger in k8sutil and event handler

perebaj:
## Description

Replace goKitLogger in k8sutil package and EventHandler struct 

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/k8sutil/secrets_test.go
+++ b/pkg/k8sutil/secrets_test.go
@@ -16,9 +16,11 @@ package k8sutil
 
 import (
 	"context"
+	"log/slog"
+	"math"
+	"os"
 	"testing"
 
-	"github.com/go-kit/log"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


pkg/k8sutil/secrets_test.go
```diff
@@ -39,7 +41,10 @@ func TestLoadSecretRef(t *testing.T) {
 	}
 
 	sClient := fake.NewSimpleClientset(secret).CoreV1().Secrets("ns")
-	logger := log.NewNopLogger()
+	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+		// slog level math.MaxInt means no logging
+		Level: slog.Level(math.MaxInt),
+	}))
```
perebaj, line 47: The previous logger isn't logging nothing, for this reason, I'm using `Level: sloglLevel(math.MaxInt)`

ArthurSens: Just did a little research and looks like slog doesn't have a builtin No-op logger. It's in the works tho (See https://github.com/golang/go/issues/62005) :)

Could you add a comment here saying that we want to use slog's builtin No-op logger once it's available? You can add the Go's issue link to make it clearer

perebaj: good catch! tnks.

[079a260](https://github.com/prometheus-operator/prometheus-operator/pull/6832/commits/079a2602d650a5c8e2ef7f8c6815c5d1d7fb7085)

We are using this level across other parts of the code, in the next PR I will address them putting a similar comment

ArthurSens (approved):
Looks good, thanks!

PR #6867, opened 2024-08-22
ScrapeConfig: Refactor StaticConfig

mviswanathsai:
## Description

PR adds additional validation and e2e testcases for StaticConfig.

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
+++ b/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
@@ -337,7 +337,7 @@ type StaticConfig struct {
 	// List of targets for this static configuration.
 	// +kubebuilder:validation:MinItems:=1
 	// +listType=set
-	// +optional
+	// +required
 	Targets []Target `json:"targets,omitempty"`
 	// Labels assigned to all metrics scraped from the targets.
 	// +mapType:=atomic
--- a/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
+++ b/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
@@ -338,7 +338,7 @@ type StaticConfig struct {
 	// +kubebuilder:validation:MinItems:=1
 	// +listType=set
 	// +required
-	Targets []Target `json:"targets,omitempty"`
+	Targets []Target `json:"targets"`
 	// Labels assigned to all metrics scraped from the targets.
 	// +mapType:=atomic
 	// +optional
--- a/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
+++ b/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
@@ -28,6 +28,7 @@ const (
 )
 
 // Target represents a target for Prometheus to scrape
+// kubebuilder:validation:MinLength:=1
 type Target string
 
 // SDFile represents a file used for service discovery
@@ -334,6 +335,8 @@ type ScrapeConfigSpec struct {
 // +k8s:openapi-gen=true
 type StaticConfig struct {
 	// List of targets for this static configuration.
+	// +kubebuilder:validation:MinItems:=1
+	// +listType=set
 	// +optional
 	Targets []Target `json:"targets,omitempty"`
 	// Labels assigned to all metrics scraped from the targets.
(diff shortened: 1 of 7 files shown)


mviswanathsai:
hmm... isn't "11owned-by" an invalid label name, judging by this regex: 
`// +kubebuilder:validation:Pattern:="^[a-zA-Z_][a-zA-Z0-9_]*$"` 
Why then is the test failing? 
CC: @simonpasquier 

I also checked locally,
```
kind: ScrapeConfig
metadata:
  name: static-config
  namespace: my-namespace
  labels:
    prometheus: system-monitoring-prometheus
spec:
  staticConfigs:
    - targets:
       - "1.1.1.1"
       - "1.1.1.2"
      labels: {"*owned-by": "prom"}
```
is not failed.

Am I missing something?

mviswanathsai:
```
kind: ScrapeConfig
metadata:
  name: static-config
  namespace: my-namespace
  labels:
    prometheus: system-monitoring-prometheus
spec:
  staticConfigs:
    - labels:
        /job: prometheus
      targets:
        - prometheus.demo.do.prometheus.io:9090
```
This works as well.

test/e2e/scrapeconfig_test.go
```diff
@@ -1065,3 +1068,78 @@ var EC2SDTestCases = []scrapeCRDTestCase{
...
+		expectedError: true,
+	},
+	{
+		name: "Valid labels",
+		scrapeConfigSpec: monitoringv1alpha1.ScrapeConfigSpec{
+
+			StaticConfigs: []monitoringv1alpha1.StaticConfig{
+				{
+					Targets: []monitoringv1alpha1.Target{"1.1.1.1:9090", "0.0.0.0:9090"},
+					Labels:  map[monitoringv1.LabelName]string{"owned-by": "prometheus"},
```
simonpasquier, line 1126: "owned-by" isn't valid.

simonpasquier (commented):
> Am I missing something?

hmm the reason might be that map keys aren't validated properly? It seems to work for scalar or array fields like `sourceLabels`.

test/e2e/scrapeconfig_test.go
mviswanathsai: I see what you mean. However, the test does not fail nor does the cluster fail to accept this object when deployed locally. What should we do?
One thing I can think of is try making the Labels a struct with a `listType=Map` and `listKey=LabelName` and see if that works.

mviswanathsai: But then, how would the JSON look like. I don't quite understand this.

simonpasquier: > I see what you mean. However, the test does not fail nor does the cluster fail to accept this object when deployed locally. What should we do?

https://groups.google.com/g/kubebuilder/c/CMvdWa4et_k?pli=1

Indeed the only option *would* be to use `+listType=map` but it would be against the idea that ScrapeConfig is a 1:1 mapping of the Prometheus config format...

As such we'd need a new PR to change the `map[v1.LabelName]string` type to `map[string]string` and validate the keys in the resource selector.

test/e2e/scrapeconfig_test.go
```diff
@@ -1065,3 +1068,78 @@ var EC2SDTestCases = []scrapeCRDTestCase{
...
+				},
+			},
+		},
+		expectedError: false,
+	},
+	{
```
simonpasquier, line 1132: this test can be removed then.

mviswanathsai: Right, so lets remove this test from here and create a new PR for the change (or could we do it here, AFAIK it only seems to be in `StaticConfig`).

mviswanathsai:
I have removed the test, I also went ahead and logged an issue. Will work on it shortly.

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -325,7 +326,9 @@ type ScrapeConfigSpec struct {
 // +k8s:openapi-gen=true
 type StaticConfig struct {
 	// List of targets for this static configuration.
-	// +optional
+	// +kubebuilder:validation:MinItems:=1
+	// +listType=set
+	// +required
 	Targets []Target `json:"targets,omitempty"`
```
simonpasquier, line 332:
```suggestion
	Targets []Target `json:"targets"`
```

mviswanathsai:
PTAL @simonpasquier

simonpasquier (approved):

PR #6879, opened 2024-08-27
Support PodMonitor for Prometheus Agent in DaemonSet mode

haanhvu:
## Description

Support PodMonitor for Prometheus Agent in DaemonSet mode. The reasons and design of this feature are described in https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202405-agent-daemonset.md#63-targets-filtering-for-pods-podmonitor-support

## Type of change

## Verification
WIP: e2e for sure, unit tests maybe

## Changelog entry

Support PodMonitor for Prometheus Agent in DaemonSet mode

### DIFF
--- a/pkg/apis/monitoring/v1/prometheus_types.go
+++ b/pkg/apis/monitoring/v1/prometheus_types.go
@@ -53,6 +53,8 @@ type PrometheusInterface interface {
 	GetStatus() PrometheusStatus
 }
 
+var _ = PrometheusInterface(&Prometheus{})
+
 func (l *Prometheus) GetCommonPrometheusFields() CommonPrometheusFields {
 	return l.Spec.CommonPrometheusFields
 }
--- a/pkg/apis/monitoring/v1/prometheus_types.go
+++ b/pkg/apis/monitoring/v1/prometheus_types.go
@@ -53,8 +53,6 @@ type PrometheusInterface interface {
 	GetStatus() PrometheusStatus
 }
 
-var _ = PrometheusInterface(&Prometheus{})
-
 func (l *Prometheus) GetCommonPrometheusFields() CommonPrometheusFields {
 	return l.Spec.CommonPrometheusFields
 }
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -661,8 +661,7 @@ func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv
 
 	logger.Info("sync prometheus")
 
-	opts := []prompkg.ConfigGeneratorOption{}
-	opts = append(opts, prompkg.DaemonSet())
+	opts := []prompkg.ConfigGeneratorOption{prompkg.WithDaemonSet()}
 	if c.endpointSliceSupported {
 		opts = append(opts, prompkg.WithEndpointSliceSupport())
 	}
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -662,6 +662,7 @@ func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv
 	logger.Info("sync prometheus")
 
 	opts := []prompkg.ConfigGeneratorOption{}
+	opts = append(opts, prompkg.DaemonSet())
 	if c.endpointSliceSupported {
 		opts = append(opts, prompkg.WithEndpointSliceSupport())
 	}
@@ -915,10 +916,6 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
 }
 
 func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *monitoringv1alpha1.PrometheusAgent, cg *prompkg.ConfigGenerator, store *assets.StoreBuilder) error {
-	if c.daemonSetFeatureGateEnabled {
-		cg.PromAgentDaemonSet = true
-	}
-
 	resourceSelector, err := prompkg.NewResourceSelector(c.logger, p, store, c.nsMonInf, c.metrics, c.eventRecorder)
 	if err != nil {
 		return err
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -915,6 +915,10 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
 }
 
 func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *monitoringv1alpha1.PrometheusAgent, cg *prompkg.ConfigGenerator, store *assets.StoreBuilder) error {
+	if c.daemonSetFeatureGateEnabled {
+		cg.PromAgentDaemonSet = true
+	}
+
 	resourceSelector, err := prompkg.NewResourceSelector(c.logger, p, store, c.nsMonInf, c.metrics, c.eventRecorder)
 	if err != nil {
 		return err
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -68,7 +68,7 @@ type ConfigGenerator struct {
 	useEndpointSlice       bool // Whether to use EndpointSlice for service discovery from `ServiceMonitor` objects.
 	scrapeClasses          map[string]monitoringv1.ScrapeClass
 	defaultScrapeClassName string
-	PromAgentDaemonSet     bool
+	daemonSet              bool
 }
 
 type ConfigGeneratorOption func(*ConfigGenerator)
@@ -77,6 +77,12 @@ func WithEndpointSliceSupport() ConfigGeneratorOption {
 	return func(cg *ConfigGenerator) {
 		cpf := cg.prom.GetCommonPrometheusFields()
 		cg.useEndpointSlice = ptr.Deref(cpf.ServiceDiscoveryRole, monitoringv1.EndpointsRole) == monitoringv1.EndpointSliceRole
+	}
+}
+
+func DaemonSet() ConfigGeneratorOption {
+	return func(cg *ConfigGenerator) {
+		cg.daemonSet = true
 	}
 }
 
@@ -185,7 +191,7 @@ func (cg *ConfigGenerator) WithKeyVals(keyvals ...interface{}) *ConfigGenerator 
 		useEndpointSlice:       cg.useEndpointSlice,
 		scrapeClasses:          cg.scrapeClasses,
 		defaultScrapeClassName: cg.defaultScrapeClassName,
-		PromAgentDaemonSet:     cg.PromAgentDaemonSet,
+		daemonSet:              cg.daemonSet,
 	}
 }
 
@@ -205,7 +211,7 @@ func (cg *ConfigGenerator) WithMinimumVersion(version string) *ConfigGenerator {
 			useEndpointSlice:       cg.useEndpointSlice,
 			scrapeClasses:          cg.scrapeClasses,
 			defaultScrapeClassName: cg.defaultScrapeClassName,
-			PromAgentDaemonSet:     cg.PromAgentDaemonSet,
+			daemonSet:              cg.daemonSet,
 		}
 	}
 
@@ -228,7 +234,7 @@ func (cg *ConfigGenerator) WithMaximumVersion(version string) *ConfigGenerator {
 			useEndpointSlice:       cg.useEndpointSlice,
 			scrapeClasses:          cg.scrapeClasses,
 			defaultScrapeClassName: cg.defaultScrapeClassName,
-			PromAgentDaemonSet:     cg.PromAgentDaemonSet,
+			daemonSet:              cg.daemonSet,
 		}
 	}
 
@@ -1140,8 +1146,8 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 	labeler := namespacelabeler.New(cpf.EnforcedNamespaceLabel, cpf.ExcludedFromEnforcement, false)
 	relabelings = append(relabelings, generateRelabelConfig(labeler.GetRelabelingConfigs(m.TypeMeta, m.ObjectMeta, ep.RelabelConfigs))...)
 
-	// Prometheus Agent in DaemonSet mode doesn't support sharding.
-	if !cg.PromAgentDaemonSet {
+	// DaemonSet mode doesn't support sharding.
+	if !cg.daemonSet {
 		relabelings = generateAddressShardingRelabelingRules(relabelings, shards)
 	}
 
@@ -1883,8 +1889,8 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 			})
 	}
 
-	// Specific configuration generated for Prometheus Agent in DaemonSet mode.
-	if cg.PromAgentDaemonSet {
+	// Specific configuration generated for DaemonSet mode.
+	if cg.daemonSet {
 		k8sSDConfig = cg.AppendMapItem(k8sSDConfig, "selectors", []yaml.MapSlice{
 			{
 				{
@@ -1895,12 +1901,6 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 					Key:   "field",
 					Value: "spec.nodeName=$(NODE_NAME)",
 				},
-			},
-		})
-		k8sSDConfig = cg.AppendMapItem(k8sSDConfig, "attach_metadata", yaml.MapSlice{
-			{
-				Key:   "node",
-				Value: true,
 			},
 		})
 	}
@@ -2014,8 +2014,12 @@ func (cg *ConfigGenerator) generateAdditionalScrapeConfigs(
 	if err != nil {
 		return nil, fmt.Errorf("unmarshalling additional scrape configs failed: %w", err)
 	}
-	if shards == 1 {
-		return additionalScrapeConfigsYaml, nil
+
+	// DaemonSet mode doesn't support sharding.
+	if !cg.daemonSet {
+		if shards == 1 {
+			return additionalScrapeConfigsYaml, nil
+		}
 	}
 
 	var addlScrapeConfigs []yaml.MapSlice
@@ -2041,7 +2045,13 @@ func (cg *ConfigGenerator) generateAdditionalScrapeConfigs(
 				relabelings = append(relabelings, relabeling)
 			}
 		}
-		relabelings = cg.generateAddressShardingRelabelingRulesIfMissing(relabelings, shards)
+		// DaemonSet mode doesn't support sharding.
+		if !cg.daemonSet {
+			if shards == 1 {
+				relabelings = cg.generateAddressShardingRelabelingRulesIfMissing(relabelings, shards)
+			}
+		}
+
 		addlScrapeConfig = append(addlScrapeConfig, otherConfigItems...)
 		addlScrapeConfig = append(addlScrapeConfig, yaml.MapItem{Key: "relabel_configs", Value: relabelings})
 		addlScrapeConfigs = append(addlScrapeConfigs, addlScrapeConfig)
@@ -2606,19 +2616,18 @@ func (cg *ConfigGenerator) GenerateAgentConfiguration(
 	)
 
 	scrapeConfigs = cg.appendPodMonitorConfigs(scrapeConfigs, pMons, apiserverConfig, store, shards)
+	scrapeConfigs, err := cg.appendAdditionalScrapeConfigs(scrapeConfigs, additionalScrapeConfigs, shards)
+	if err != nil {
+		return nil, fmt.Errorf("generate additional scrape configs: %w", err)
+	}
 
-	// Currently, Prometheus Agent in DaemonSet mode doesn't support these.
-	if !cg.PromAgentDaemonSet {
+	// Currently, DaemonSet mode doesn't support these.
+	if !cg.daemonSet {
 		scrapeConfigs = cg.appendServiceMonitorConfigs(scrapeConfigs, sMons, apiserverConfig, store, shards)
 		scrapeConfigs = cg.appendProbeConfigs(scrapeConfigs, probes, apiserverConfig, store, shards)
-		scrapeConfigs, err := cg.appendScrapeConfigs(scrapeConfigs, sCons, store, shards)
+		scrapeConfigs, err = cg.appendScrapeConfigs(scrapeConfigs, sCons, store, shards)
 		if err != nil {
 			return nil, fmt.Errorf("generate scrape configs: %w", err)
-		}
-
-		scrapeConfigs, err = cg.appendAdditionalScrapeConfigs(scrapeConfigs, additionalScrapeConfigs, shards)
-		if err != nil {
-			return nil, fmt.Errorf("generate additional scrape configs: %w", err)
 		}
 	}
 
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -80,7 +80,7 @@ func WithEndpointSliceSupport() ConfigGeneratorOption {
 	}
 }
 
-func DaemonSet() ConfigGeneratorOption {
+func WithDaemonSet() ConfigGeneratorOption {
 	return func(cg *ConfigGenerator) {
 		cg.daemonSet = true
 	}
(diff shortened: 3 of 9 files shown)


pkg/prometheus/agent/operator.go
```diff
@@ -856,7 +856,7 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
 	return nil
 }
 
-func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *monitoringv1alpha1.PrometheusAgent, cg *prompkg.ConfigGenerator, store *assets.StoreBuilder) error {
+func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *monitoringv1alpha1.PrometheusAgent, cg *prompkg.ConfigGenerator, store *assets.StoreBuilder, daemonSet bool) error {
```
nicolastakashi, line 859: (nit) I think an enum would be more idiomatic

like 

```
const (
	DaemonSet   WorkloadType = "DaemonSet"
	StatefulSet WorkloadType = "StatefulSet"
)

switch workload {
	case DaemonSet:
		fmt.Println("The workload is a DaemonSet")
	case StatefulSet:
		fmt.Println("The workload is a StatefulSet")
	default:
		fmt.Println("Unknown workload type")
	}
```

Just an example

pkg/prometheus/promcfg.go
```diff
@@ -1142,6 +1143,263 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
...
+	cfg = cg.AddHonorLabels(cfg, ep.HonorLabels)
+	cfg = cg.AddHonorTimestamps(cfg, ep.HonorTimestamps)
+	cfg = cg.AddTrackTimestampsStaleness(cfg, ep.TrackTimestampsStaleness)
+
+	attachMetaConfig := mergeAttachMetadataWithScrapeClass(m.Spec.AttachMetadata, scrapeClass, "2.35.0")
+
+	s := store.ForNamespace(m.Namespace)
+
+	k8SSDConfig := cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, s, kubernetesSDRolePod, attachMetaConfig)
+	k8SSDConfigValue := k8SSDConfig.Value.([]yaml.MapSlice)
```
nicolastakashi, line 1169:
```suggestion
k8SSDConfig.Value = append(k8SSDConfig.Value.([]yaml.MapSlice), yaml.MapSlice{
	{
		Key:   "role",
		Value: "pod",
	},
	{
		Key: "selectors",
		Value: yaml.MapSlice{
			{
				Key:   "role",
				Value: "pod",
			},
			{
				Key:   "field",
				Value: "spec.nodeName=" + os.Getenv("NODE_NAME"),
			},
		},
	},
})
```

I'm not sure if it works, but wdyt?

pkg/prometheus/agent/operator.go
haanhvu: Yeah we repeat these everywhere, not only in this PR. I'll take a look later how far we can generalize^^

pkg/prometheus/promcfg.go
haanhvu: Basically it's combining my two steps into one, so it'll probably work. Let me see

haanhvu:
@ArthurSens @simonpasquier AFAICS the implementation is complete. Now the next things to do are:

- Local test
- Unit test in promcfg_test.go
- e2e that checks if PA DaemonSet picks the correct pods in the same node

Please help with local test and tell me wyt^^

pkg/prometheus/agent/operator.go
simonpasquier: I'm not sure that we need an extra argument since the information is already present in the PrometheusAgent object.

simonpasquier (commented):
Can we modify the existing `generatePodMonitorConfig()` function to check for the daemonset mode and adjust the generated config accordingly? I think that it would be more "maintan-able" over time.

pkg/prometheus/agent/operator.go
haanhvu: Aha! Funny silly mistake...

haanhvu:
> Can we modify the existing generatePodMonitorConfig() function to check for the daemonset mode and adjust the generated config accordingly? I think that it would be more "maintan-able" over time.

Yeah we actually can refactor further more, in other functions too. I'll do it later (before this PR's merged).

pkg/prometheus/promcfg.go
```diff
@@ -2513,6 +2840,15 @@ func (cg *ConfigGenerator) appendAdditionalScrapeConfigs(scrapeConfigs []yaml.Ma
 	return append(scrapeConfigs, addlScrapeConfigs...), nil
 }
 
+func (cg *ConfigGenerator) appendAdditionalScrapeConfigsForAgentDaemonSet(scrapeConfigs []yaml.MapSlice, additionalScrapeConfigs []byte) ([]yaml.MapSlice, error) {
```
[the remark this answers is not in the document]
haanhvu: So we should remove this?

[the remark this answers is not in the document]
simonpasquier: It would be my preference.

test/e2e/prometheusagent_test.go
```diff
@@ -365,3 +374,240 @@ func testPromAgentReconcileDaemonSetResourceDelete(t *testing.T) {
...
+	_, err = framework.CreatePrometheusAgentAndWaitUntilReady(ctx, ns, prometheusAgentDS)
+	require.NoError(t, err)
+
+	var pollErr error
+	err = wait.PollUntilContextTimeout(context.Background(), 20*time.Second, 20*time.Minute, false, func(ctx context.Context) (bool, error) {
+		appPods, err := framework.KubeClient.CoreV1().Pods(ns).List(ctx, metav1.ListOptions{
+			LabelSelector: "group=test",
+		})
+		if err != nil {
+			pollErr = fmt.Errorf("can't list app pods: %w", err)
```
haanhvu, line 434: @simonpasquier @ArthurSens The test is failing here. Can you help me figure out?

pkg/prometheus/agent/operator.go
```diff
@@ -954,15 +954,25 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *mon
...
-		pmons,
-		bmons,
-		scrapeConfigs,
-		p.Spec.TSDB,
-		store,
-		additionalScrapeConfigs,
-	)
+	var conf []byte
+	if ptr.Deref(p.Spec.Mode, "StatefulSet") == "DaemonSet" {
+		conf, err = cg.GenerateAgentDaemonSetConfiguration(
```
ArthurSens, line 959: I'm not sure if we need separate functions `GenerateAgentDaemonSetConfiguration` and `GenerateAgentConfiguration` here. The object that calls them, `cg`, has `PrometheusInterface` as one of its attributes, which means that `cg` is able to tell if it is generating the configuration to a daemonset or not.


https://github.com/prometheus-operator/prometheus-operator/blob/bf3a738f8ce631035005884f99ed1f1a7ef453a2/pkg/prometheus/promcfg.go#L2558

Here we could do something like:

```go
	if cg.prom.(*monitoringv1alpha1.PrometheusAgent).Spec.Mode == "Daemonset" {
		scrapeConfigs = cg.appendPodMonitorDaemonsetConfigs(scrapeConfigs, pMons, apiserverConfig, store, shards)
	} else {
		scrapeConfigs = cg.appendPodMonitorConfigs(scrapeConfigs, pMons, apiserverConfig, store, shards)
	}
```

and avoid a lot of duplicated code we have in this PR

haanhvu: Yeah a know the PR is a bit messy at this moment^^" as I said in https://github.com/prometheus-operator/prometheus-operator/pull/6879#issuecomment-2322821270 I'll clean it up before we merge

Can you help me with the error in https://github.com/prometheus-operator/prometheus-operator/pull/6879#discussion_r1741078817 too? This error is about not being able to retrieve our sample app's pods, which doesn't really relate to podmonitor logic anyway

cc @simonpasquier

test/e2e/prometheusagent_test.go
```diff
@@ -365,3 +374,240 @@ func testPromAgentReconcileDaemonSetResourceDelete(t *testing.T) {
...
+	require.NoError(t, err)
+
+	prometheusAgentDS := framework.MakeBasicPrometheusAgentDaemonSet(ns, name)
+	prometheusAgentDS.Spec.ScrapeInterval = "1s"
+
+	_, err = framework.CreatePrometheusAgentAndWaitUntilReady(ctx, ns, prometheusAgentDS)
+	require.NoError(t, err)
+
+	var pollErr error
+	err = wait.PollUntilContextTimeout(context.Background(), 20*time.Second, 20*time.Minute, false, func(ctx context.Context) (bool, error) {
```
simonpasquier, line 429:
```suggestion
	err = wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 10*time.Minute, false, func(_ context.Context) (bool, error) {
		ctx := context.Background()
```

test/e2e/prometheusagent_test.go
```diff
@@ -365,3 +374,240 @@ func testPromAgentReconcileDaemonSetResourceDelete(t *testing.T) {
...
+		}
+
+		// The prometheus certificate is issued to <pod>.<namespace>.svc,
+		// but port-forwarding is done through localhost.
+		// This is why we use an http client which skips the TLS verification.
+		// In the test we will verify the TLS certificate manually to make sure
+		// the prometheus instance is configured properly.
+		transport := &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
haanhvu: So we just need to initialize http client without configuring transport like this `httpClient := http.Client{}`?

test/e2e/prometheusagent_test.go
```diff
@@ -365,3 +374,249 @@ func testPromAgentReconcileDaemonSetResourceDelete(t *testing.T) {
...
+			return false, nil
+		}
+
+		httpClient := http.Client{
+			Transport: transport,
+		}
+
+		resp, err := httpClient.Do(req)
+		if err != nil {
+			pollErr = fmt.Errorf("can't send http request to first prometheus server: %w", err)
```
haanhvu, line 508: @ArthurSens @simonpasquier The test is now failing here: `can't send http request to first prometheus server: Get "https://localhost:9090/api/v1/targets": http: server gave HTTP response to HTTPS client`

So AFAIU to fix this there are two options:

- Make http client receive http response
- Make `https://localhost:9090/api/v1/targets` serve https response

Seems like option 1 is the way to go right. If we follow this option I need to remove custom transport as stated in https://github.com/prometheus-operator/prometheus-operator/pull/6879#discussion_r1742064354 right?

test/e2e/prometheusagent_test.go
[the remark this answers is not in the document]
simonpasquier: correct

test/e2e/prometheusagent_test.go
```diff
@@ -365,3 +374,249 @@ func testPromAgentReconcileDaemonSetResourceDelete(t *testing.T) {
...
+		defer cancel()
+
+		closer, err := testFramework.StartPortForward(ctx, cfg, "https", paPods.Items[0].Name, ns, "9090")
+		if err != nil {
+			pollErr = fmt.Errorf("can't start port forward to first prometheus agent pod: %w", err)
+			return false, nil
+		}
+		defer closer()
+
+		req, err := http.NewRequestWithContext(ctx, "GET", "https://localhost:9090/api/v1/targets", nil)
```
simonpasquier, line 480: we should use http instead of https.
```suggestion
		req, err := http.NewRequestWithContext(ctx, "GET", "http://localhost:9090/api/v1/targets", nil)
```

test/e2e/prometheusagent_test.go
simonpasquier: see my comment above (we need to fix the url scheme).

pkg/prometheus/promcfg.go
```diff
@@ -1148,6 +1148,263 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
...
+		{
+			Key: "selectors",
+			Value: yaml.MapSlice{
+				{
+					Key:   "role",
+					Value: "pod",
+				},
+				{
+					Key:   "field",
+					Value: "spec.nodeName=" + os.Getenv("NODE_NAME"),
```
simonpasquier, line 1189: we don't want the NODE_NAME value from the prometheus operator container. Instead the config reloader should replace the value at runtime.

```suggestion
					Value: "spec.nodeName=$(NODE_NAME)",
```

simonpasquier:
Hmm I've tested locally and while the pod monitor got selected, the Prometheus config is empty :thinking: 

````
[...] msg="selected PodMonitors" component=prometheusagent-controller podmonitors=default/example-app namespace=default prometheus=agent
````

````
global:
  scrape_interval: 30s
  scrape_timeout: 10s
  scrape_protocols:
  - OpenMetricsText1.0.0
  - OpenMetricsText0.0.1
  - PrometheusText0.0.4
  evaluation_interval: 1m
  external_labels:
    prometheus: default/agent
    prometheus_replica: prom-agent-agent-qxb7q
runtime:
  gogc: 75
````

**Comment by haanhvu** [comment]
> Hmm I've tested locally and while the pod monitor got selected, the Prometheus config is empty 🤔
> 
> ```
> [...] msg="selected PodMonitors" component=prometheusagent-controller podmonitors=default/example-app namespace=default prometheus=agent
> ```
> 
> ```
> global:
>   scrape_interval: 30s
>   scrape_timeout: 10s
>   scrape_protocols:
>   - OpenMetricsText1.0.0
>   - OpenMetricsText0.0.1
>   - PrometheusText0.0.4
>   evaluation_interval: 1m
>   external_labels:
>     prometheus: default/agent
>     prometheus_replica: prom-agent-agent-qxb7q
> runtime:
>   gogc: 75
> ```

@simonpasquier Yeah I think we should finish unit test first. To see how the config is generated. But I tried updating golden files and nothing changed:
Screenshot from 2024-08-31 16-05-03

**Comment by haanhvu** [comment]
> > Hmm I've tested locally and while the pod monitor got selected, the Prometheus config is empty 🤔
> > ```
> > [...] msg="selected PodMonitors" component=prometheusagent-controller podmonitors=default/example-app namespace=default prometheus=agent
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > global:
> >   scrape_interval: 30s
> >   scrape_timeout: 10s
> >   scrape_protocols:
> >   - OpenMetricsText1.0.0
> >   - OpenMetricsText0.0.1
> >   - PrometheusText0.0.4
> >   evaluation_interval: 1m
> >   external_labels:
> >     prometheus: default/agent
> >     prometheus_replica: prom-agent-agent-qxb7q
> > runtime:
> >   gogc: 75
> > ```
> 
> @simonpasquier Yeah I think we should finish unit test first, to see how the config is generated. But I tried updating golden files and nothing changed: Screenshot from 2024-08-31 16-05-03
> 
> But seems like this isn't a big problem? I could just compare the functions' results to empty golden files, to check the functions' results right?
> 
> But first, let me clean up this PR. It's harder to find out what's wrong in a messy PR...

@simonpasquier Actually now I realize why some configs were not generated. It was a mistake in my new code. I'll fix it

File: pkg/prometheus/promcfg.go [hunk 11]
```diff
@@ -1854,6 +1859,36 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+						Key:   "role",
+						Value: "pod",
+					},
+					{
+						Key:   "field",
+						Value: "spec.nodeName=$(NODE_NAME)",
+					},
+				},
+			},
+		}
```
haanhvu, line 1882: @simonpasquier @ArthurSens The new unit test `TestPromAgentDaemonSetPodMonitorConfig` shows:
```
kubernetes_sd_configs:
          - role: pod
            namespaces:
              names:
              - default
```

It means the new necessary config for daemonset implemented here was not added.

Could you help me figure out why?

pkg/prometheus/promcfg.go
```diff
@@ -2554,18 +2576,23 @@ func (cg *ConfigGenerator) GenerateAgentConfiguration(
...
+	// Currently, Prometheus Agent in DaemonSet mode doesn't support these.
+	if !cg.PromAgentDaemonSet {
+		scrapeConfigs = cg.appendServiceMonitorConfigs(scrapeConfigs, sMons, apiserverConfig, store, shards)
+		scrapeConfigs = cg.appendProbeConfigs(scrapeConfigs, probes, apiserverConfig, store, shards)
+		scrapeConfigs, err := cg.appendScrapeConfigs(scrapeConfigs, sCons, store, shards)
+		if err != nil {
+			return nil, fmt.Errorf("generate scrape configs: %w", err)
+		}
+
+		scrapeConfigs, err = cg.appendAdditionalScrapeConfigs(scrapeConfigs, additionalScrapeConfigs, shards)
```
simonpasquier, line 2590: Sorry if my earlier comment wasn't clear: I would still support additional scrape configs with daemonset mode but I would drop the sharding config.

pkg/prometheus/agent/operator.go
```diff
@@ -907,6 +907,10 @@ func (c *Operator) syncStatefulSet(ctx context.Context, key string, p *monitorin
 }
 
 func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, p *monitoringv1alpha1.PrometheusAgent, cg *prompkg.ConfigGenerator, store *assets.StoreBuilder) error {
+	if c.daemonSetFeatureGateEnabled {
+		cg.PromAgentDaemonSet = true
```
simonpasquier, line 911: We should set true only when the feature gate is enabled + mode == daemonset.

pkg/prometheus/promcfg.go
```diff
@@ -68,6 +68,7 @@ type ConfigGenerator struct {
 	useEndpointSlice       bool // Whether to use EndpointSlice for service discovery from `ServiceMonitor` objects.
 	scrapeClasses          map[string]monitoringv1.ScrapeClass
 	defaultScrapeClassName string
+	PromAgentDaemonSet     bool
```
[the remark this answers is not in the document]
ArthurSens: Is this field really needed though? ConfigGenerator already has the Prometheus object, which holds the `Mode` information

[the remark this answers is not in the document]
simonpasquier: It's a PrometheusAgent field only.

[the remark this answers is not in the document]
ArthurSens: Yes, but we can do a cast switch and grab it. But I understand the extra effort, no strong opinions here

pkg/prometheus/promcfg.go
```diff
@@ -1868,6 +1876,28 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+					Value: "spec.nodeName=$(NODE_NAME)",
+				},
+			},
+		})
+		k8sSDConfig = cg.AppendMapItem(k8sSDConfig, "attach_metadata", yaml.MapSlice{
+			{
+				Key:   "node",
+				Value: true,
+			},
+		})
```
simonpasquier, line 1898: I don't remember why we need to attach node metadata?

simonpasquier (commented):
Tested locally and it worked fine!

pkg/prometheus/promcfg.go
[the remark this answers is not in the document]
haanhvu: I actually did a simple cast switch with `if *cg.prom.(*monitoringv1alpha1.PrometheusAgent).Spec.Mode == "Daemonset"`. It returned an error somewhat like "PrometheusInterface is Prometheus, not Prometheus Agent". That's why I created this new variable instead.

pkg/prometheus/promcfg.go
haanhvu: Yeah we never discussed about this. I added this from reading Prometheus config doc. At first I thought we needed to set this to get node information. But now reread the doc, seems like this is only needed if we want to "attach" node info to targets, which we don't need right?

simonpasquier: yes if users want to retrieve node metadata, it's ok and can be done setting `attachMetadata.node: true` in the endpoint definition but it shouldn't be mandatory.

haanhvu:
@ArthurSens @simonpasquier e2e passed now :partying_face: (I'll do some refactoring for the common steps of two port forwarding)

The only problem now is this unit test case is failing, and I don't know why:
https://github.com/prometheus-operator/prometheus-operator/blob/f4af7da70647e050874edb4b3da87237119de156/pkg/prometheus/promcfg_test.go#L1469-L1471

I checked my changes in `promcfg.go` and they don't seem to have any problems. And all other unit tests and e2e passed. So I'm not sure why this specific test cases is failing.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -53,8 +53,6 @@ type PrometheusInterface interface {
 	GetStatus() PrometheusStatus
 }
 
-var _ = PrometheusInterface(&Prometheus{})
```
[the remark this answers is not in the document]
simonpasquier: Yes this construct ensures that the Prometheus struct implements the PrometheusInterface interface. Though not required, it makes it easier to spot when we change the interface definition.

pkg/prometheus/promcfg.go
```diff
@@ -79,6 +80,12 @@ func WithEndpointSliceSupport() ConfigGeneratorOption {
 	}
 }
 
+func DaemonSet() ConfigGeneratorOption {
```
simonpasquier, line 83: (nit)
```suggestion
func WithDaemonSetMode() ConfigGeneratorOption {
```

pkg/prometheus/promcfg.go
```diff
@@ -1984,8 +2014,12 @@ func (cg *ConfigGenerator) generateAdditionalScrapeConfigs(
...
 		return nil, fmt.Errorf("unmarshalling additional scrape configs failed: %w", err)
 	}
-	if shards == 1 {
-		return additionalScrapeConfigsYaml, nil
+
+	// DaemonSet mode doesn't support sharding.
+	if !cg.daemonSet {
+		if shards == 1 {
+			return additionalScrapeConfigsYaml, nil
+		}
```
simonpasquier, line 2022:
```suggestion
	// DaemonSet mode doesn't support sharding.
	if cg.daemonSet || shards == 1 {
			return additionalScrapeConfigsYaml, nil
```

pkg/prometheus/promcfg.go
```diff
@@ -2011,7 +2045,13 @@ func (cg *ConfigGenerator) generateAdditionalScrapeConfigs(
 				relabelings = append(relabelings, relabeling)
 			}
 		}
-		relabelings = cg.generateAddressShardingRelabelingRulesIfMissing(relabelings, shards)
+		// DaemonSet mode doesn't support sharding.
+		if !cg.daemonSet {
+			if shards == 1 {
+				relabelings = cg.generateAddressShardingRelabelingRulesIfMissing(relabelings, shards)
+			}
```
simonpasquier, line 2052: this is why the unit tests fail IIUC
```suggestion
			relabelings = cg.generateAddressShardingRelabelingRulesIfMissing(relabelings, shards)
```

pkg/prometheus/promcfg.go
```diff
@@ -2575,18 +2615,22 @@ func (cg *ConfigGenerator) GenerateAgentConfiguration(
...
 	)
 
-	scrapeConfigs = cg.appendServiceMonitorConfigs(scrapeConfigs, sMons, apiserverConfig, store, shards)
 	scrapeConfigs = cg.appendPodMonitorConfigs(scrapeConfigs, pMons, apiserverConfig, store, shards)
-	scrapeConfigs = cg.appendProbeConfigs(scrapeConfigs, probes, apiserverConfig, store, shards)
-	scrapeConfigs, err := cg.appendScrapeConfigs(scrapeConfigs, sCons, store, shards)
+	scrapeConfigs, err := cg.appendAdditionalScrapeConfigs(scrapeConfigs, additionalScrapeConfigs, shards)
 	if err != nil {
-		return nil, fmt.Errorf("generate scrape configs: %w", err)
+		return nil, fmt.Errorf("generate additional scrape configs: %w", err)
```
simonpasquier, line 2621: :+1:

pkg/prometheus/agent/operator.go
```diff
@@ -662,6 +662,7 @@ func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv
 	logger.Info("sync prometheus")
 
 	opts := []prompkg.ConfigGeneratorOption{}
+	opts = append(opts, prompkg.DaemonSet())
```
simonpasquier, line 665: (nit) a bit shorter
```suggestion
	opts := []prompkg.ConfigGeneratorOption{prompkg.DaemonSet())
```

simonpasquier (approved):
:partying_face:

PR #6882, opened 2024-08-28
feat: support EndpointSlice for the kubelet controller

### LINKED ISSUE #3862: Support the endpointslice role for Prometheus

simonpasquier:
## Description

This change adds support for managing `EndpointSlice` objects for the kubelet service. The controller can manage either one of `Endpoints` and EndpointSlice` or both.
    
The migration path can be:
    1. Configure the operator to manage both objects in the kubelet controller.
    2. Verify that the generated `EndpointSlice` objects are correct.
    3. Configure the Prometheus object to use the `EndpointSlice` role instead of `Endpoints`.
    1. Configure the operator to manage only `Endpoints` objects in the kubelet controller.
    
The removal of the legacy `Endpoints` object is left to the user.

Closes #3862

## Type of change

## Verification

## Changelog entry

```release-note

```


Documentation/operator.md
```diff
@@ -57,6 +57,10 @@ Usage of ./operator:
     	  PrometheusAgentDaemonSet: Enables the DaemonSet mode for PrometheusAgent (enabled: false)
   -key-file string
     	- NOT RECOMMENDED FOR PRODUCTION - Path to private TLS certificate file.
+  -kubelet-endpoints
```
nicolastakashi, line 60: Is it one or another right?
What I mean is we can't use both together?

example/rbac/prometheus-operator/prometheus-operator-deployment.yaml
```diff
@@ -26,7 +26,7 @@ spec:
       containers:
       - args:
         - --kubelet-service=kube-system/kubelet
-        - --prometheus-config-reloader=quay.io/prometheus-operator/prometheus-config-reloader:v0.76.0
+        - --prometheus-config-reloader=quay.io/prometheus-operator/prometheus-config-reloader:v0.76.1
```
nicolastakashi: Interesting!
Should it be 0.76.2?

simonpasquier: It should be fixed now that conflicts are resolved.

simonpasquier:
Since we have an upcoming release of the operator, I'd like to land this PR before. Given that nothing changes by default, it should be safe.
On Red Hat side, we will be turning on the endpoinslice flag in OpenShift and report back.

nicolastakashi (approved):

PR #6890, opened 2024-08-29
replace goKitLogger in controller and namespace denylist

perebaj:
## Description

- Replace goKitLogger in controller and namespace denylist

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification

Please check the [Prometheus-Operator testing guidelines](../TESTING.md) for recommendations about automated tests.

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #6902, opened 2024-09-03
chore: redirect klog to slog

simonpasquier:
## Description

PR #6890 removed the final bits from go-kit/log in the operator code base but forgot to redirect the klog logs to the new slog logger. `ResourceReconciler` has a dependency on klog because it uses `utilruntime.HandleError()` when it fails to sync a resource. As a consequence, the operator didn't log reconciliation failures anymore.

_I didn't qualify it as a bug because the latest operator version doesn't have the issue._

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	k8sflag "k8s.io/component-base/cli/flag"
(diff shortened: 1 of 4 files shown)


ArthurSens (approved):
Nice catch and thanks for fixing it!

cc @perebaj for awareness!

simonpasquier:
I found it while testing a failure case and I was like: "hmm it sure fails to reconcile but I see no log???"

PR #6906, opened 2024-09-04
feat: support  `allow_overlapping_compaction` for prometheus

### LINKED ISSUE #6829: Disable overlapping compaction in Prometheus when Thanos sidecar and OOO enabled
### Component(s)

Prometheus

When Prometheus is deployed together with Thanos sidecar and out of order is enabled, Prometheus should disable overlapping compaction otherwise Thanos sidecar couldn't upload vertical compacted blocks to object storage.

If Prometheus version is v2.55.0 or newer, enable flag https://github.com/prometheus/prometheus/pull/14581 in Prometheus if Thanos sidecar and OOO enabled. 

_No response_

## Environment
Kubernetes Version:
Prometheus-Operator Version:

dongjiang1989:
## Description

Closes #6829 

## Type of change

## Verification

## Changelog entry

```release-note
support  `allow_overlapping_compaction` for prometheus
```


simonpasquier:
I'm not sure if @yeya24 asked for an explicit field in the CRD or for the operator to automatically set `--no-storage.tsdb.allow-overlapping-compaction` when all the conditions are met:
1. Prometheus >= v2.55.0
2. Thanos sidecar configured for uploading blocks to object storage
3. out-of-order window is > 0

I assume the latter but I might be wrong.

yeya24:
Hi @simonpasquier, for me the latter should be good enough. I am also fine with an explicit field in the CRD but sounds unnecessary as the only usecase is what you mentioned.

simonpasquier:
For the record, I prefer the second option (automatic activation of `--no-storage.tsdb.allow-overlapping-compaction`) too.  

@yeya24 do you foresee any situation where it would not be desired?

dongjiang1989:
> * Prometheus >= v2.55.0
> * Thanos sidecar configured for uploading blocks to object storage
> * out-of-order window is > 0

@simonpasquier  @yeya24  Just keep `--no-storage.tsdb.allow-overlapping-compaction`(no field in the CRD). 
Whether to retain the above verification in the code?

dongjiang1989:
> @dongjiang1989 not sure about your question but I'd be ok with the operator adding or not the flag based on the 3 conditions that I've listed.

Thanks @simonpasquier . I Got it.
PTAL
cc @yeya24

pkg/prometheus/server/statefulset.go
```diff
@@ -303,6 +305,10 @@ func makeStatefulSetSpec(
 		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.min-block-duration", Value: thanosBlockDuration})
 	}
 
+	if allowOverlappingCompaction && cpf.TSDB != nil && cpf.TSDB.OutOfOrderTimeWindow != "" && disableCompaction {
```
simonpasquier, line 308: I suggest that we don't make it a visible option and automatically turn on overlapping compaction when all conditions are met.
```suggestion
	if cpf.TSDB != nil && cpf.TSDB.OutOfOrderTimeWindow != "" && disableCompaction && cg.WithMinimumVersion("2.55.0").IsCompatible() {
```

dongjiang1989: > I suggest that we don't make it a visible option and automatically turn on overlapping compaction when all conditions are met.
> 
> ```suggestion
> 
> 	if cpf.TSDB != nil && cpf.TSDB.OutOfOrderTimeWindow != "" && disableCompaction && cg.WithMinimumVersion("2.55.0").IsCompatible() {
> 
> ```

Thanks. I got it. 
But i s it possible for users to actively set it to false? 🤔

mviswanathsai: I understand that the suggestion was to not make it a configurable option from the user: check for the 3 conditions and enable behind the scenes. (which is not what the current implementation does)
Joining @dongjiang1989 though, should we still allow the user to set it to false @simonpasquier ? It sounds reasonable, I won't be able to give a use case though.

dongjiang1989: > The option is hidden in Prometheus (https://github.com/prometheus/prometheus/blob/e0260930d6ffbd945f7b6d030f6258b27078a442/cmd/prometheus/main.go#L408-L409) and the only (?) use case is when used with the Thanos sidecar to upload blocks to object storage. To avoid API bloat, I'd prefer not to expose it for now.

Thanks @simonpasquier . I got it.  
Remove `allowOverlappingCompaction` it done.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1884,6 +1884,7 @@ type TSDBSpec struct {
 	// in a breaking way.
 	//
 	// It requires Prometheus >= v2.39.0 or PrometheusAgent >= v2.54.0.
+	// +optional
 	OutOfOrderTimeWindow Duration `json:"outOfOrderTimeWindow,omitempty"`
```
simonpasquier, line 1888: We need a pointer.
```suggestion
	OutOfOrderTimeWindow *Duration `json:"outOfOrderTimeWindow,omitempty"`
```

pkg/prometheus/server/statefulset.go
```diff
@@ -302,6 +302,12 @@ func makeStatefulSetSpec(
 		promArgs = append(promArgs, monitoringv1.Argument{Name: "storage.tsdb.min-block-duration", Value: thanosBlockDuration})
 	}
 
+	if cpf.TSDB != nil && cpf.TSDB.OutOfOrderTimeWindow != "" &&
```
simonpasquier, line 305: Can we add a comment explaining the rationale here? A link to the original issue would also work.

pkg/apis/monitoring/v1/prometheus_types.go
dongjiang1989: Thanks @simonpasquier . 
Change it done.

pkg/prometheus/server/statefulset.go
dongjiang1989: Thanks @simonpasquier .
Add ref issue link done.

dongjiang1989:
@simonpasquier Pleaser re-check it, thanks

simonpasquier (approved):
Thanks!

trip-pguedes:
^^ Would like to know the answer to this as well

slashpai:
https://github.com/prometheus-operator/prometheus-operator/pull/7731
A fix PR in progress

PR #6943, opened 2024-09-17
WIP: Disable direct switch between Prometheus Agent modes

haanhvu:
## Description

Disable direct switch between DaemonSet and StatefulSet modes in Prometheus Agent, which is explained in https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202405-agent-daemonset.md#61-crd

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification
WIP: e2e

## Changelog entry

Disable direct switch between Prometheus Agent modes

PR #6944, opened 2024-09-17
Disable Prometheus Agent fields that are unrelated in DaemonSet mode

haanhvu:
## Description

Fail Prometheus Agent operator in DaemonSet mode if unrelated fields are set

## Type of change

## Verification

## Changelog entry

Disable Prometheus Agent fields that are unrelated in DaemonSet mode

### DIFF
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -571,6 +571,30 @@ func (c *Operator) syncDaemonSet(ctx context.Context, key string, p *monitoringv
 		return fmt.Errorf("feature gate for Prometheus Agent's DaemonSet mode is not enabled")
 	}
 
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
Same as for https://github.com/prometheus-operator/prometheus-operator/pull/6943: it would be nice to list the fields which are not supported in the API documentation.

ArthurSens (commented):
Should we be more explicit in the error message about what forbidden field must be removed?

If a user reads `DaemonSet mode doesn't support storage setting but storage fields are set in Prometheus Agent CRD in DaemonSet mode`, it might not be obvious to him/her which field needs to be removed

simonpasquier:
My suggestion would be something like

`field '.spec.xxx' is not supported with DaemonSet mode and should be unset`

PR #6782, opened 2024-07-30
Replace t.Fatal with require package in tests-e2e

AshwinSriram11:
## Description
Continuation to #6749, #6690, #6683, #6680 

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification

Please check the [Prometheus-Operator testing guidelines](../TESTING.md) for recommendations about automated tests.

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #6952, opened 2024-09-19
Replace t.Fatal with require package

AshwinSriram11:
## Description
 Continuation to #6680, #6683, #6690, #6749 and #6782 .

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/test/framework/namespace.go
+++ b/test/framework/namespace.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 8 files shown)


test/framework/namespace.go
```diff
@@ -31,19 +31,15 @@ func (f *Framework) CreateNamespace(ctx context.Context, t *testing.T, testCtx *
 	name := testCtx.ID()
 	rn := k8sutil.ResourceNamer{}
 	name, err := rn.UniqueDNS1123Label(name)
-	if err != nil {
-		t.Fatal(fmt.Errorf("failed to generate a namespace name %v: %w", name, err))
-	}
+	require.NoError(t, err)
```
nicolastakashi, line 34: I almost sure we can also provide the same message while using require.NoError

AshwinSriram11: I will add it.

ArthurSens: We can, but why though?

We added messages before because t.Fatal provides no context about why we were failing a test.
NoError provides that context: "We expected no error but we got one"

nicolastakashi: I don’t have a strong preference on this, but I aim to maintain consistency with previous behavior. Additionally, when dealing with multiple assertions, I find it helpful to include messages that clarify where the failure occurred.

AshwinSriram11: It depends. You may also want to see additional details like namespace name in this case. But I don't know if it is that important.

ArthurSens: NoError already includes the error message `Received unexpected error: %v`[[1](https://github.com/stretchr/testify/blob/b074924938f86d417f1c9a845c7e8b0784d7f937/assert/assertions.go#L1584)].

The stack trace, that includes the line where the test fails, and the text message provided by require.NoError is enough context in my opinion 😬

PR #6961, opened 2024-09-23
build: move build step to Dockerfiles

jan--f:
## Description

This introduces multi-stage Dockerfiles. The go build step for images now happens in the build step in the container file definition. The prometheus golang builder image is used.

The main benefit is that external build systems can simply build the Dockerfile, no need to know how to build binaries or what dependencies to install.

In the future there could be benefits for building on non-x86 architectures and perhaps testing.

## Type of change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,18 @@
 ARG ARCH="amd64"
 ARG OS="linux"
+ARG GOLANG_BUILDER="1.23"
+FROM quay.io/prometheus/golang-builder:${GOLANG_BUILDER}-base as builder
+WORKDIR /workspace
+
+# Copy source files
+COPY . .
+
+# Build
+RUN make operator
+
 FROM quay.io/prometheus/busybox-${OS}-${ARCH}:latest
 
-COPY operator /bin/operator
+COPY --from=builder workspace/operator /bin/operator
 
 # On busybox 'nobody' has uid `65534'
 USER 65534
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ IMAGE_RELOADER?=quay.io/prometheus-operator/prometheus-config-reloader
 IMAGE_WEBHOOK?=quay.io/prometheus-operator/admission-webhook
 TAG?=$(shell git rev-parse --short HEAD)
 VERSION?=$(shell cat VERSION | tr -d " \t\n\r")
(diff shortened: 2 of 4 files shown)


nicolastakashi:
I like the idea, especially because we could leverage buidx cache from the registry.

Dockerfile
```diff
@@ -1,8 +1,27 @@
 ARG ARCH="amd64"
 ARG OS="linux"
+FROM quay.io/prometheus/golang-builder:main as builder
+WORKDIR /workspace
+
+# Copy the required source files
+COPY .git/ .git/
```
nicolastakashi, line 7: I was wondering if we could COPY . since it's a multi stage we will only get what we need in the end, using COPY might make our lives a bit easier, did you see any problem on that @jan--f ?

Dockerfile
```diff
@@ -1,8 +1,27 @@
 ARG ARCH="amd64"
 ARG OS="linux"
+FROM quay.io/prometheus/golang-builder:main as builder
```
simonpasquier, line 3: i wonder if we should pin to a specific major version in line with the go.mod?
The GitHub runner already has access to the ${golang-version} variable defined in `.github/env`.

Dockerfile
jan--f: :+1: 
I was mostly not sure if `COPY . .` was a thing...but TIL :)

Dockerfile
jan--f: I didn't because I figured the lattest go version will (or should) always be able to build projects that specify an older version in `go.mod`.
Afaiu Prometheus only keeps the builders up to date for n-1 versions because they test with the previous version. Unless p-o does something similar, I don't see an advantage to pining a specific builder version.

simonpasquier: Yeah I don't have a concrete example which would require pinning the Go version but I'd prefer if we ensure that the same Go version is used everywhere. Can we have a way to inject the Go image tag and default to main if unspecified?

jan--f: How about this?
Added a new make target that updates the version as specified in `.github/env` (maybe use go.mod instead?) and integrate it into the generate target.

jan--f: Oh even better...those are much smaller.

nicolastakashi (approved):
LGTM!
@simonpasquier waiting your final review.

simonpasquier (approved):
:tada: 
Thanks!

PR #7000, opened 2024-10-09
feat: Support add the ability to set custom `http_headers` in config.

dongjiang1989:
## Description

ref: https://github.com/prometheus/prometheus/pull/14817

Add the ability to set custom `http_headers` in config.

## Type of change

## Verification

## Changelog entry

```release-note
Add the ability to set custom `http_headers` in config.
```


pkg/apis/monitoring/v1/types.go
```diff
@@ -102,6 +102,11 @@ type ProxyConfig struct {
 	// +optional
 	// +mapType:=atomic
 	ProxyConnectHeader map[string][]v1.SecretKeySelector `json:"proxyConnectHeader,omitempty"`
+
+	// It requires Prometheus >= v2.55.0.
+	//
+	// +optional
+	HTTPHeadersConfig `json:",inline"`
```
simonpasquier, line 109: IIUC the field is added to HTTP configuration, not Proxy configuration?

simonpasquier (commented):
To make the review easier, can we start adding only `Secrets []v1.SecretKeySelector` ?

pkg/apis/monitoring/v1/types.go
dongjiang1989: There is currently no global HTTPConfig, add a new HTTPConfig struct as name `HTTPConfig`
```
type HTTPConfig struct {
        // It requires Prometheus >= v2.55.0.
	//
	// +optional
	HTTPHeadersConfig `json:",inline"`
}
```
@simonpasquier WDYT?

dongjiang1989:
> To make the review easier, can we start adding only `Secrets []v1.SecretKeySelector` ?

Thanks @simonpasquier .

I'll split into several PRs:

- [ ]  add the ability to set custom http_headers only `Secrets []v1.SecretKeySelector` in config
- [ ] add the ability to set custom http_headers with `File []string` and `values []string` in config

pkg/apis/monitoring/v1/types.go
simonpasquier: Hmm I think that the `inline` stanza isn't at the right place?

````

type Endpoint struct {
    ...

    HTTPConfig `json:",inline"`
}

type HTTPConfig struct {
        // It requires Prometheus >= v2.55.0.
	//
	// +optional
	HTTPHeaders map[string]HTTPHeader `json:"httpHeaders,omitempty"`
}
````

Over time we could move all the common HTTP fields to the new HTTPConfig (it will be tedious but it doesn't have to happen all at once).

dongjiang1989: Thanks. I got it.
Change `HTTPConfig` to `CustomHTTPConfig` because it conflicts with the `HTTPConfig` struct with the same name in `alertmanager_types.go`

PR #7004, opened 2024-10-11
feat: operator support set basic auth for prometheus server

xu-wentao:
## Description

Support set basic auth info from secret and set statefulset probe's httpGet headers.
Aim to solve https://github.com/prometheus-operator/prometheus-operator/issues/5836

## Type of change

## Verification

## Changelog entry

```release-note

```


pkg/apis/monitoring/v1/types.go
```diff
@@ -288,6 +288,16 @@ type WebConfigFileFields struct {
...
+	// Defines the basic auth users for server.
+	// +optional
+	BasicAuthUsers []*WebBasicAuthUsers `json:"basicAuthUsers,omitempty"`
+}
+
+// WebBasicAuthUsers defines the basic auth users for --web.config.file flag.
+// +k8s:openapi-gen=true
+type WebBasicAuthUsers struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
```
simonpasquier, line 300: you don't want passwords in clear text inside the Prometheus CRD.

simonpasquier (commented):
We have to solve the issue of Kubernetes probes not working if you configure basic-auth.

pkg/apis/monitoring/v1/types.go
Xu-Wentao: yes, i'm on the way to slove this. i'm going to store `username` and `password` in secret, and when reconcile `Prometheus` resource, we can get info from secret and then set basic auth. Also, set probe to solve the kubernetes readiness check.

Xu-Wentao:
@simonpasquier  hi, i have changed the type of `basicAuthUsers` to `BasicAuth` and basis auth info store in secret (should be created maunally first) currently. and when reconcile prometheus, we will set statefulset's probes with basic auth headers automaticly. But there's still have a promblem that, user can get basic auth info by decoding authorization info which in prometheus liveness probe or readiness probe.

pkg/apis/monitoring/v1/types.go
simonpasquier: It's not acceptable from a security standpoint to store credentials in clear-text in the statefulset spec. We need https://github.com/prometheus/exporter-toolkit/pull/151 to allow unauthenticated access to the probe endpoints.

simonpasquier:
There's also the question of the Thanos sidecar and config reloader which require authentication credentials too.
Another option instead of picking up a random user/password from the list provided by the user would be to create a special username with a random password and inject the credentials in the generated passwd file.

Xu-Wentao:
> er
I will add authentication credentials  to thanos and config reloader.  
Regarding the second question, the automatically generated username and random password will also be encoded by the user from the pod information, and then decoded to obtain the original data. Is this situation also unsafe?

Xu-Wentao:
> There's also the question of the Thanos sidecar and config reloader which require authentication credentials too. Another option instead of picking up a random user/password from the list provided by the user would be to create a special username with a random password and inject the credentials in the generated passwd file.

Regarding the second question, the automatically generated username and random password will also be encoded by the user from the pod information, and then decoded to obtain the original data. Is this situation also unsafe?

Xu-Wentao:
> There's also the question of the Thanos sidecar and config reloader which require authentication credentials too. Another option instead of picking up a random user/password from the list provided by the user would be to create a special username with a random password and inject the credentials in the generated passwd file.

ok.I will finish it. bytheway, maybe prometheus and other components can exclued health check api for authicated like the issue https://github.com/prometheus/prometheus/issues/9166 mentioned.

simonpasquier:
I'm on the fence regarding the API: I wonder if the operator should be in charge of producing the brypt hashes or if it should pass the secret data as-is to Prometheus (and let the user provide the hashes). In particular involving the operator with security concerns is always a risk for the long-term maintenance.

simonpasquier:
After more thinking, I realize that the simplest API might be for the user to provide a secret containing all the credentials where the username would be the key and the password the value:

````yaml
apiVersion: v1
kind: Secret
metadata:
  name: web-users
stringData:
  foo: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
  bar: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
  prometheus: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
---
apiVersion: v1
kind: Secret
metadata:
  name: example-webconfig
stringData:
  password: test
---
kind: Prometheus
metadata:
  name: example
spec:
  web:
    basicAuthUsers:
      secretName: web-users
      podCredentials:
        username: prometheus
        password:
           key: password
           name: example-webconfig
````

The operator can fail the reconciliation if it detects invalid hashes in the secret (e.g. using `bcrypt.Cost()`).

Again sorry for the back and forth but I didn't put a lot of effort in this part before your PR so thanks for doing it!

**Comment by Xu-Wentao** [comment]
> After more thinking, I realize that the simplest API might be for the user to provide a secret containing all the credentials where the username would be the key and the password the value:
> 
> ```yaml
> apiVersion: v1
> kind: Secret
> metadata:
>   name: web-users
> stringData:
>   foo: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
>   bar: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
>   prometheus: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
> ---
> apiVersion: v1
> kind: Secret
> metadata:
>   name: example-webconfig
> stringData:
>   password: test
> ---
> apiVersion: monitoring.coreos.com/v1
> kind: Prometheus
> metadata:
>   name: example
> spec:
>   web:
>     basicAuthUsers:
>       secretName: web-users
>       podCredentials:
>         username: prometheus
>         password:
>            key: password
>            name: example-webconfig
> ```
> 
> The operator can fail the reconciliation if it detects invalid hashes in the secret (e.g. using `bcrypt.Cost()`).
> 
> Again sorry for the back and forth but I didn't put a lot of effort in this part before your PR so thanks for doing it!

Ok, I'll try to change the API as you recommended, But the way, when i try to set a correct config of `thanos sidecar` to make sidecar can access to prometheus server normally, i find out that the sidecar currently set `--prometheus.http-client` config as a plain text, is that ok? coz if we want to use a config file for `--promethes.http-client-config` https://thanos.io/tip/components/sidecar.md/#configuration, this a lot of work need to do.

**Comment by simonpasquier** [comment]
> when i try to set a correct config of `thanos sidecar` to make sidecar can access to prometheus server normally, i find out that the sidecar currently set `--prometheus.http-client` config as a plain text, is that ok? coz if we want to use a config file for `--promethes.http-client-config` https://thanos.io/tip/components/sidecar.md/#configuration, this a lot of work need to do.

For sure, we don't want the HTTP client config to be in clear-text if it contains sensitive information like password so yes, we'll need to change this and probably use a dedicated secret mounted in the sidecar. And we need the same for the config reloader sidecar. I'd say that it should be a preliminary PR to this one to facilitate the review process.

For context, both thanos and config-reloader sidecars had to disable TLS verification because they only connect through localhost and we didn't want to bother with configuring the right CA.

**Comment by Xu-Wentao** [comment]
> process

ok, i'll create another new PR to support config secrets for sidecar.

**Comment by Xu-Wentao** [comment]
> > when i try to set a correct config of `thanos sidecar` to make sidecar can access to prometheus server normally, i find out that the sidecar currently set `--prometheus.http-client` config as a plain text, is that ok? coz if we want to use a config file for `--promethes.http-client-config` https://thanos.io/tip/components/sidecar.md/#configuration, this a lot of work need to do.
> 
> For sure, we don't want the HTTP client config to be in clear-text if it contains sensitive information like password so yes, we'll need to change this and probably use a dedicated secret mounted in the sidecar. And we need the same for the config reloader sidecar. I'd say that it should be a preliminary PR to this one to facilitate the review process.
> 
> For context, both thanos and config-reloader sidecars had to disable TLS verification because they only connect through localhost and we didn't want to bother with configuring the right CA.

**Comment by Xu-Wentao** [comment]
> > when i try to set a correct config of `thanos sidecar` to make sidecar can access to prometheus server normally, i find out that the sidecar currently set `--prometheus.http-client` config as a plain text, is that ok? coz if we want to use a config file for `--promethes.http-client-config` https://thanos.io/tip/components/sidecar.md/#configuration, this a lot of work need to do.
> 
> For sure, we don't want the HTTP client config to be in clear-text if it contains sensitive information like password so yes, we'll need to change this and probably use a dedicated secret mounted in the sidecar. And we need the same for the config reloader sidecar. I'd say that it should be a preliminary PR to this one to facilitate the review process.
> 
> For context, both thanos and config-reloader sidecars had to disable TLS verification because they only connect through localhost and we didn't want to bother with configuring the right CA.

I find that the config-reloader's `ReloaderURL` is defined here https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/common.go#L456 and did not support config basic auth headers, should i change the thanos config reloader to support this param?

**Comment by simonpasquier** [comment]
For the config reloader, the idea should be to add CLI arguments to pass the basic-auth user/password in https://github.com/prometheus-operator/prometheus-operator/blob/main/cmd/prometheus-config-reloader/main.go.

**Comment by Xu-Wentao** [comment]
> After more thinking, I realize that the simplest API might be for the user to provide a secret containing all the credentials where the username would be the key and the password the value:经过更多思考，我意识到最简单的 API 可能是让用户提供一个包含所有凭据的秘密，其中用户名是密钥，密码是值：
> 
> ```yaml
> apiVersion: v1
> kind: Secret
> metadata:
>   name: web-users
> stringData:
>   foo: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
>   bar: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
>   prometheus: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
> ---
> apiVersion: v1
> kind: Secret
> metadata:
>   name: example-webconfig
> stringData:
>   password: test
> ---
> apiVersion: monitoring.coreos.com/v1
> kind: Prometheus
> metadata:
>   name: example
> spec:
>   web:
>     basicAuthUsers:
>       secretName: web-users
>       podCredentials:
>         username: prometheus
>         password:
>            key: password
>            name: example-webconfig
> ```
> 
> The operator can fail the reconciliation if it detects invalid hashes in the secret (e.g. using `bcrypt.Cost()`).如果操作员检测到秘密中的无效哈希值（例如使用`bcrypt.Cost()` ），则协调可能会失败。
> 
> Again sorry for the back and forth but I didn't put a lot of effort in this part before your PR so thanks for doing it!再次对来来回回表示歉意，但在你的 PR 之前我没有在这部分投入很多精力，所以感谢你这样做！

how about like this, this `podCredentialUser` must be in the secret. User can specify which user info is used for pod probe check. the operator will check if the  `podCredentialUser` in secret when reconcile.
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: web-users
stringData:
  foo: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
  bar: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
  prometheus: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
---
apiVersion: v1
kind: Secret
metadata:
  name: example-webconfig
stringData:
  password: test
---
kind: Prometheus
metadata:
  name: example
spec:
  web:
    basicAuthUsers:
      secretName: web-users
      podCredentialUser: prometheus
```

Xu-Wentao:
> There's already an identity for the "pod user" which is the service account (SA) name. To make things simple, we could require that the secret contains a key equal to the SA name (or `default` if the resource doesn't define a SA). If the key isn't present, the reconciliation fails.

So, if we pick up username from serviceaccount, then maybe we should create hashed value automatically? operator should merge the auth user infos from `basicAuthUsers.secretName` and the user for probe check (name is SA and hashed value is created by operator) in web-config.

Xu-Wentao:
> There's already an identity for the "pod user" which is the service account (SA) name. To make things simple, we could require that the secret contains a key equal to the SA name (or `default` if the resource doesn't define a SA). If the key isn't present, the reconciliation fails.

If we use the SA name as the username of basic auth for pod probe check, how to get the password of it, from another secret specified by user? or create a secrete to save password  automatically?

simonpasquier:
Right, we need both the hashed password and the password in clear text. Something like this?
```yaml
spec:
  web:
    basicAuthUsers:
      secretName: web-users # secret
      serviceAccountPassword:
        name: credentials
        key: prometheus-web-password
````

**Comment by Xu-Wentao** [comment]
> After more thinking, I realize that the simplest API might be for the user to provide a secret containing all the credentials where the username would be the key and the password the value:
> 
> ```yaml
> apiVersion: v1
> kind: Secret
> metadata:
>   name: web-users
> stringData:
>   foo: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
>   bar: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
>   prometheus: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
> ---
> apiVersion: v1
> kind: Secret
> metadata:
>   name: example-webconfig
> stringData:
>   password: test
> ---
> apiVersion: monitoring.coreos.com/v1
> kind: Prometheus
> metadata:
>   name: example
> spec:
>   web:
>     basicAuthUsers:
>       secretName: web-users
>       podCredentials:
>         username: prometheus
>         password:
>            key: password
>            name: example-webconfig
> ```
> 
> The operator can fail the reconciliation if it detects invalid hashes in the secret (e.g. using `bcrypt.Cost()`).
> 
> Again sorry for the back and forth but I didn't put a lot of effort in this part before your PR so thanks for doing it!

I finally used this desgin, coz it's more clearly for user to setup the `podCredentials` which is used  for pod health check. and if we use the ServiceAccount as username, when we build the `web-config`, it did not have the `prometheusSpec` param, it may make it code more complex.

File: cmd/prometheus-config-reloader/main.go [hunk 2]
```diff
@@ -109,6 +110,8 @@ func main() {
 	runtimeInfoURL := app.Flag("runtimeinfo-url", "URL to check the status of the runtime configuration").
 		Default("http://127.0.0.1:9090/api/v1/status/runtimeinfo").URL()
 
+	serverBasicAuth := app.Flag("server-basic-auth", "Basic Auth credentials for the prometheus server. The format should be 'username:password'").Default("").String()
```
simonpasquier, line 113: We can't have the password in clear-text.

pkg/apis/monitoring/v1/types.go
```diff
@@ -314,6 +314,24 @@ type WebConfigFileFields struct {
...
 	HTTPConfig *WebHTTPConfig `json:"httpConfig,omitempty"`
+	// Defines the basic auth users for server.
+	// +optional
+	BasicAuthUsers *BasicAuthUsers `json:"basicAuthUsers,omitempty"`
+}
+
+// BasicAuthUsers defines the basic auth users for prometheus server.
+type BasicAuthUsers struct {
+	// Defines the name of secret stored basic auth users.
+	SecretName *string `json:"secretName,omitempty"`
```
simonpasquier, line 325: all fields of BasicAuthUsers are mandatory/required.

simonpasquier:
> I finally used this desgin, coz it's more clearly for user to setup the podCredentials which is used for pod health check. and if we use the ServiceAccount as username, when we build the web-config, it did not have the prometheusSpec param, it may make it code more complex.

I would like to avoid adding more fields than necessary to the API. Let's see if we can get rid of the `username` somehow.

cmd/prometheus-config-reloader/main.go
Xu-Wentao: > We can't have the password in clear-text.

So we should use a config file to save the basic auth info in `reloader` ? like we mount an additional file into contaienr, and the flag may like `server-pod-credentials-config-file`.

Xu-Wentao:
> > I finally used this desgin, coz it's more clearly for user to setup the podCredentials which is used for pod health check. and if we use the ServiceAccount as username, when we build the web-config, it did not have the prometheusSpec param, it may make it code more complex.
> 
> I would like to avoid adding more fields than necessary to the API. Let's see if we can get rid of the `username` somehow.

How about that? (maybe we can generate the passowrd automatically, user no need to create it)
---
apiVersion: v1
kind: Secret
metadata:
  name: web-users
stringData:
  foo: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
  bar: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay
---
apiVersion: v1
kind: Secret
metadata:
  name: example-webconfig
stringData:
  password: test
---
kind: Prometheus
metadata:
  name: example
spec:
  web:
    basicAuthUsers:
      secretName: web-users
      podCredentials:
        password:
           key: password
           name: example-webconfig
---
# generate by operator
apiVersion: v1
kind: Secret
metadata:
  name: pod-credentials
data:
  # username and password is used for health check probe/thanos-sidecar/config-reloader
  username: serviceAccountName
  password: test
  # hashedValue is used for prometheus server merged with `web-users` and save into `web-config`
  hashedValue: "serviceAccountName: $2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay"

simonpasquier:
I would simplify the struct a bit:

```yaml
kind: Prometheus
metadata:
  name: example
spec:
  web:
    basicAuthUsers:
      secretName: web-users
      podCredentials:
         key: password
         name: example-webconfig
````

Indeed the operator could merge the user-provided user list (`web-users` secret) with the bcrypt hash from the SA password though I'm a bit nervous of dealing with security/crypto topics. As a first approach, I would make it mandatory to have an entry in `web-users` for the service account (and fail reconciliation if it's not the case).

In all cases, we still need https://github.com/prometheus/exporter-toolkit/pull/151 before proceeding.

**Comment by Xu-Wentao** [comment]
> 情况

Agree with that, and i was tried that way. But there is a unwanted thing happend:  the username:password will be bcrypted by operator automatically, and when the secret created (merged with `web-user` secret), it will cause trigger another reconcile, and then a different bcryped hashed value will gerneated, and then cause one another reconcile and so on.

There's some thing i think we can work on it: when we try to create a secret `web-config`, we can compare the current user list if we need to update.

**Comment by simonpasquier** [comment]
> But there is a unwanted thing happend: the username:password will be bcrypted by operator automatically, and when the secret created (merged with web-user secret), it will cause trigger another reconcile, and then a different bcryped hashed value will gerneated, and then cause one another reconcile and so on.

I'm not sure to understand what's happening here. Also my suggestion would be not having the operator to deal with password encryption.

**Comment by Xu-Wentao** [comment]
@simonpasquier sorry for update late. Currently the basic auth is designed like this.
```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: basic-auth-users
stringData:
  foo: "$2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay"
  bar: "$2b$12$hNf2lSsxfm0.i4a.1kVpSOVyBCfIB51VRjgBUyv6kdnyTlgWj81Ay"
  prometheus: "$2a$12$6H/IeGCSIqdl2Bg1Mk3D5ebqbc6qESkmIwBXiJLLg/nDN3OVQBF76"
---
apiVersion: v1
kind: Secret
metadata:
  name: basic-auth-pod-credentials
stringData:
  password: "prometheus-password"
---
kind: Prometheus
metadata:
  name: prometheus
  labels:
    prometheus: basic-auth
spec:
  replicas: 1
  thanos:
    logLevel: info
  serviceMonitorSelector: {}
  web:
    basicAuthUsers:
      secretName: basic-auth-users
      podCredentials:
        key: password
        name: basic-auth-pod-credentials
```
User need to createe a `Secret` to store all basic-auth-users info, included the name from `ServiceAccount` (in this example is `prometheus`), and create a `Secret` to store the password for user name from `ServiceAccount`.
Is this design like waht you mentoned?

simonpasquier:
> User need to createe a `Secret` to store all basic-auth-users info, included the name from `ServiceAccount` (in this example is `prometheus`), and create a `Secret` to store the password for user name from `ServiceAccount`.
> Is this design like waht you mentoned?

yes this is what i had in mind.

Xu-Wentao:
> > User need to createe a `Secret` to store all basic-auth-users info, included the name from `ServiceAccount` (in this example is `prometheus`), and create a `Secret` to store the password for user name from `ServiceAccount`.
> > Is this design like waht you mentoned?
> 
> yes this is what i had in mind.
I have completed the writing of relevant code logic. Please take a look at which logic we can continue to adjust and optimize.

cmd/prometheus-config-reloader/main.go
```diff
@@ -242,10 +245,47 @@ func createHTTPClient(timeout *time.Duration) http.Client {
...
 	}
 
+	var rt http.RoundTripper = transport
+
+	if podCredentialsUsername != "" || podCredentialsPasswordFile != "" {
+		password, err := os.ReadFile(fmt.Sprintf("%s/password", podCredentialsPasswordFile))
+		if err != nil {
+			fmt.Printf("Failed to read password file: %v\n", err)
+		} else {
+			rt = newBasicAuthRoundTripper(podCredentialsUsername, strings.TrimSpace(string(password)), rt)
```
RouxAntoine, line 255: There is an existing basic roundtripper in prometheus common `config.NewBasicAuthRoundTripper()` https://github.com/prometheus/common/blob/main/config/http_config.go#L839 maybe it would be best to reuse it

RouxAntoine (approved):

pkg/apis/monitoring/v1/types.go
```diff
@@ -316,6 +316,17 @@ type WebConfigFileFields struct {
...
 	HTTPConfig *WebHTTPConfig `json:"httpConfig,omitempty"`
+	// Defines the basic auth users for server.
+	// +optional
+	BasicAuthUsers *BasicAuthUsers `json:"basicAuthUsers,omitempty"`
+}
+
+// BasicAuthUsers defines the basic auth users for prometheus server.
+type BasicAuthUsers struct {
+	// Defines the name of secret stored basic auth users.
+	SecretName *string `json:"secretName"`
```
simonpasquier, line 327: To follow https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references
```suggestion
	  // Defines the secret containing the list of users (keys) with their hashed passwords (values).
	// The secret must contain a key for the service account.
	// +required
	SecretRef SecretReference `json:"secretRef"`
```
with
````golang
type SecretReference struct {
	// Name of the secret in the resource's namespace.
	// +required
	Name string `json:"name"`
}
````

File: pkg/apis/monitoring/v1/types.go [hunk 6]
```diff
@@ -316,6 +316,17 @@ type WebConfigFileFields struct {
...
+	// Defines the pod credentials for health check probe of prometheus server.
```
simonpasquier, line 328:
```suggestion
	// Defines the secret's key holding the password used by the service account to query the web server.
	// The service account must exist in the users secret.
```

pkg/apis/monitoring/v1/types.go
```diff
@@ -316,6 +316,17 @@ type WebConfigFileFields struct {
...
+	PodCredentials *v1.SecretKeySelector `json:"podCredentials"`
```
simonpasquier, line 329:
```suggestion
	// +required
	ServiceAccountPasswordRef SecretKeySelector `json:"serviceAccountPasswordRef"`
```

````golang
type SecretKeySelector struct {
	SecretReference
	// The key of the secret to select from.
	// +required
	Key string `json:"key"`
}
````

**Review by simonpasquier — COMMENTED** [review]
I've got remarks on the API. We need to add more validations.

File: cmd/prometheus-config-reloader/Dockerfile [hunk 8]
```diff
@@ -15,7 +15,7 @@ ARG GOARCH
 ENV GOARCH=${GOARCH}
 RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache/go-build make prometheus-config-reloader
 
-FROM quay.io/prometheus/busybox-${OS}-${ARCH}:latest
+FROM quay.io/prometheus/busybox-linux-${ARCH}:latest
```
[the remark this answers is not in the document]
Xu-Wentao: Sorry, i built on my mac and run on linux machine. i will revert this.

Xu-Wentao:
@simonpasquier hi, i have use object reference to secrets as you mentioned, is there any other things i can to to move on？

PR #7006, opened 2024-10-11
Refactor Digital Ocean Service Discovery

### LINKED ISSUE #6991: Refactor `DigitalOceanSDConfig`
### Component(s)

_No response_

We want to refactor the ScrapeConfig API, at least for all the tier-1 and tier-2 Service Discoveries mentioned in #6634. Currently all of tier-1 stands refactored and some from the tier-2 are left incomplete and we want to work on them.

Some common rules of thumb for refactoring the API are:
1. The API needs to be restrictive wherever possible/sensible.
2. Arrays should not be empty if an empty array does not make sense in configuration. (empty arrays have some meaning in some cases, so be mindful of that)
3. String values (standalone, or part of a []string) should not be empty string. i.e, "".
4. Numerical values need to have a Min and Max value wherever possible. 
5. Test cases for all the API behavior we expect needs to be written.
6. Keep an eye on the minimum Prometheus version required to configure certain fields, you will have to comment and also add test cases to ensure that the unsupported field is dealt with properly.

You will find that most (hopefully all) of the above rules are followed in prior refactoring PRs. You may take inspiration from the PRs that refactored different SDs before:
#6869 
#6800
#6744 

Especially test functions may be daunting to write at first, it's really useful to take reference from prior PRs.
I tried to cover as much as I can. This might not be a good first issue in the truest sense, but it is definitely something that is exciting to do as a newcomer. If you have any doubts, you can reach out to me!

_No response_

## Environment
Kubernetes Version:
Prometheus-Operator Version:

AshwinSriram11:
## Description

Closes #6991.

## Type of change

### DIFF
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -7340,9 +7340,10 @@ func TestScrapeConfigSpecConfigWithOpenStackSD(t *testing.T) {
 
 func TestScrapeConfigSpecConfigWithDigitalOceanSD(t *testing.T) {
 	for _, tc := range []struct {
-		name   string
-		scSpec monitoringv1alpha1.ScrapeConfigSpec
-		golden string
+		name    string
+		scSpec  monitoringv1alpha1.ScrapeConfigSpec
+		golden  string
+		version string
 	}{
 		{
 			name: "digitalocean_sd_config",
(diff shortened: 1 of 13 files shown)


pkg/prometheus/promcfg_test.go
```diff
@@ -7374,14 +7375,51 @@ func TestScrapeConfigSpecConfigWithDigitalOceanSD(t *testing.T) {
...
 					},
 				},
 			},
-			golden: "ScrapeConfigSpecConfig_DigitalOceanSD.golden",
-		},
-		{
+			version: "2.40.0",
+			golden:  "ScrapeConfigSpecConfig_DigitalOceanSD.golden",
+		}, {
+			name: "digitalocean_sd_config_with_unsupported_version",
```
mviswanathsai, line 7386: Not sure we want to test for cases which won't go through the resource selector in the first place. I would suggest removing this test case, the one that you added at `resource_selector_test.go` should be sufficient.

pkg/prometheus/promcfg_test.go
```diff
@@ -7452,7 +7491,50 @@ func TestScrapeConfigSpecConfigWithDigitalOceanSD(t *testing.T) {
 					},
 				},
 			},
-			golden: "ScrapeConfigSpecConfig_DigitalOceanSD_with_TLSConfig.golden",
+			version: "2.40.0",
+			golden:  "ScrapeConfigSpecConfig_DigitalOceanSD_with_TLSConfig.golden",
+		}, {
+			name: "digitalocean_sd_config_tls_with_unsupported_version",
```
mviswanathsai, line 7497: Why are we testing for TLS config when even DigitalOcean is not valid at that version?
This test case can be removed, it seems that the TLS configuration for DigitalOcean is supported across all versions that support DigitalOcean.

mviswanathsai (commented):
Had a look. Looks good overall, just a few comments.
Can you look into the failing tests? also, did you try running `make test-unit-update-golden` to generate golden files? If not, I would advise to do it after you have addressed my comments.

mviswanathsai:
Looks like you have build errors, try running `make fix-golang`.

mviswanathsai:
LGTM, I would wait for @simonpasquier to review and merge.

mviswanathsai:
This is pretty good, I think you can mark it ready for review!

simonpasquier (approved):
thanks!

PR #7007, opened 2024-10-11
DONOTMERGE: Testing comment action

slashpai:
## Description

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -47,6 +47,7 @@ jobs:
         uses: thollander/actions-comment-pull-request@v2
         with:
          message: 'Auto-generated files are likely not up-to-date. Please make sure to run `make --always-make format generate`, commit the modifications and push the updated branch.'
+         pr-number: 7007
 
   check-docs:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


simonpasquier:
https://github.com/thollander/actions-comment-pull-request?tab=readme-ov-file#permissions

simonpasquier:
Hmm it looks like there's no easy solution :disappointed: 
https://stackoverflow.com/questions/61419552/workaround-to-post-comments-from-github-actions-from-forked-repos

simonpasquier:
maybe https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run could work but it would still be convoluted to pass the PR id to the second workflow.

slashpai:
I think the problem here is forked repo won't have access to write 
https://github.com/thollander/actions-comment-pull-request?tab=readme-ov-file#permissions
we may need to have another workflow with pull_request_target

> Note that, if the PR comes from a fork, it will have only read permission despite the permissions given in the action for the pull_request event. In this case, you may use the pull_request_target event. With this event, permissions can be given without issue (the difference is that it will execute the action from the target branch and not from the origin PR).

simonpasquier:
I don't think that `pull_request_target` will work because it's only reacting on events like "PR opened", "Review requested", ... but nothing like "CI has ended with status=X".

slashpai:
https://github.com/peter-evans/create-or-update-comment suggests same for forked repos. We probably should be able to handle that using artifact

- check-docs and generate to create artifact to denote failure
- Wait for artifact to be there to execute the job in pull_request_target workflow which has comment job

Issue #7002, opened 2024-10-10
Documentation: advertised OTLPConfig in Prometheus CRD not available yet


r-asiebert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Attempted to use `….otlp.promoteResourceAttributes` feature exposed in Prometheus CRD, which doesn't exist yet in Prometheus.

### Prometheus Operator Version

```shell
0.77.1
```

### Kubernetes Version

```yaml
N/A
```

### Kubernetes Cluster Type

Other (please comment)

### How did you deploy Prometheus-Operator?

helm chart:prometheus-community/kube-prometheus-stack

### Manifests

```yaml
# Helm values
prometheus:
  prometheusSpec:
    additionalConfig:
      otlp:
        # https://github.com/prometheus-operator/prometheus-operator/blob/3993785/Documentation/api.md#monitoring.coreos.com/v1.OTLPConfig
        promoteResourceAttributes:
          # https://prometheus.io/docs/guides/opentelemetry/
          - service.instance.id
          - service.name
          - service.namespace
          - cloud.region
          - container.name
          - deployment.environment
          - k8s.cluster.name
          - k8s.namespace.name
          - k8s.pod.name
```

### prometheus-operator log output

```shell
N/A - Operator correctly writes /etc/prometheus/config_out/prometheus.env.yaml and triggers Prom conf reload
```

## Doc location

https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.CommonPrometheusFields advertises this field in the `Prometheus` CRD spec:
> `otlp`   [OTLPConfig](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.OTLPConfig) 
> (Optional) Settings related to the OTLP receiver feature. It requires Prometheus >= v2.54.0.

([current source](https://github.com/prometheus-operator/prometheus-operator/blob/880f847f6a553019d968ba65e5eb19ade7170b59/Documentation/api.md#L11528-L11542))

This `otlp` block contains the `promoteResourceAttributes` configuration.

## Issue

After upgrading to the latest Helm Chart (kube-prometheus-stack 65.1.1) finally including the 0.77.1 Operator with this CRD field, I found out Prometheus doesn't support it yet. Prometheus 2.54.1 rejects the resulting config file with `field otlp not found in type config.plain`.

[...]

PR #7017, opened 2024-10-15
Fix issue #7002

heliapb:
## Description

Fix issue https://github.com/prometheus-operator/prometheus-operator/issues/7002
## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/operator/defaults.go
+++ b/pkg/operator/defaults.go
@@ -103,5 +103,6 @@ var (
 	// This is only added for users to try the unstable versions.
 	PrometheusExperimentalVersions = []string{
 		"v3.0.0-beta.0",
+		"v2.55.0-rc.0",
 	}
 )
--- a/pkg/operator/defaults.go
+++ b/pkg/operator/defaults.go
@@ -103,6 +103,5 @@ var (
 	// This is only added for users to try the unstable versions.
 	PrometheusExperimentalVersions = []string{
 		"v3.0.0-beta.0",
-		"v2.55.0-rc.0",
 	}
 )
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -4265,8 +4265,8 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 }
 
 func (cg *ConfigGenerator) generateOTLPConfig() (yaml.MapItem, error) {
-	if cg.version.LT(semver.MustParse("2.54.0")) {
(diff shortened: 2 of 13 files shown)


pkg/operator/defaults.go
```diff
@@ -103,5 +103,6 @@ var (
 	// This is only added for users to try the unstable versions.
 	PrometheusExperimentalVersions = []string{
 		"v3.0.0-beta.0",
+		"v2.55.0-rc.0",
```
slashpai, line 106: As we discussed in https://github.com/prometheus-operator/prometheus-operator/issues/7002#issuecomment-2406740176, option 2 is what we can do for the fix

heliapb: Ok, then should the change be done here https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L112

slashpai: here https://github.com/prometheus-operator/prometheus-operator/blob/d1fb4196594218b9aea80bfb588881c293fb6e56/pkg/prometheus/promcfg.go#L4268-L4269

pkg/prometheus/promcfg.go
```diff
@@ -4265,8 +4265,8 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 }
 
 func (cg *ConfigGenerator) generateOTLPConfig() (yaml.MapItem, error) {
-	if cg.version.LT(semver.MustParse("2.54.0")) {
-		return yaml.MapItem{}, fmt.Errorf("OTLP configuration is only supported from Prometheus version 2.54.0")
+	if cg.version.LT(semver.MustParse("2.55.0.rc.0")) {
```
slashpai, line 4268:
```suggestion
	if cg.version.LT(semver.MustParse("2.55.0")) {
```

pkg/prometheus/promcfg.go
```diff
@@ -4265,8 +4265,8 @@ func (cg *ConfigGenerator) generateScrapeConfig(
...
+		return yaml.MapItem{}, fmt.Errorf("OTLP configuration is only supported from Prometheus version 2.55.0.rc.0")
```
slashpai, line 4269:
```suggestion
		return yaml.MapItem{}, fmt.Errorf("OTLP configuration is only supported from Prometheus version 2.55.0")
```

slashpai:
You should update unit test as well

heliapb:
Hi @slashpai Done, thanks for the help

simonpasquier (approved):
Thanks!

Issue #6970, opened 2024-09-25
Add `enableHTTP2` field to the ScrapeConfig CRD


mviswanathsai:
### Component(s)

ScrapeConfig

### What is missing? Please describe.

The ScrapeConfig CRD is meant to be a 1:1 mirror of the Prometheus scrape configuration. To that extent, it is currently missing a few fields like `enable_http2` and we want to add them. It is important to note that it may not be possible to add some fields.

You may take examples of `EnableHTTP` from existing service discoveries. (#6421 might be a good place to start)

## Environment
Kubernetes Version:
Prometheus-Operator Version:

chetak123:
I would like to work on this issue please

slashpai:
go ahead 🙂

mviswanathsai:
@chetak123 do you need any help with this?

PR #7024, opened 2024-10-19
Feat: Add enableHTTP2 to ScrapeConfig CRD

nutmos:
## Description

Add the enableHTTP2 to the ScrapeConfig CRD to map with enable_http2 in Prometheus Scrape Config.

Refer to #6970.

## Type of change

## Verification

I have written tests under pkg/prometheus/promfg_test.go for verifying the working of enableHTTP2 config.

## Changelog entry

```release-note
Add support for enableHTTP2 in ScrapeConfig
```

### DIFF
--- a/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
+++ b/pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
@@ -280,6 +280,11 @@ type ScrapeConfigSpec struct {
 	// If unset, Prometheus uses true by default.
 	// +optional
 	EnableCompression *bool `json:"enableCompression,omitempty"`
+	// When false, Prometheus will request uncompressed response from the scraped target.
+	//
+	// If unset, Prometheus uses true by default.
+	// +optional
+	EnableHTTP2 *bool `json:"enableHTTP2,omitempty"`
 	// BasicAuth information to use on every scrape request.
 	// +optional
 	BasicAuth *v1.BasicAuth `json:"basicAuth,omitempty"`
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -2714,6 +2714,10 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		cfg = cg.WithMinimumVersion("2.49.0").AppendMapItem(cfg, "enable_compression", *sc.Spec.EnableCompression)
 	}
 
+	if sc.Spec.EnableHTTP2 != nil {
+		cfg = cg.AppendMapItem(cfg, "enable_http2", *sc.Spec.EnableHTTP2)
+	}
+
 	if sc.Spec.ScrapeInterval != nil {
 		cfg = append(cfg, yaml.MapItem{Key: "scrape_interval", Value: *sc.Spec.ScrapeInterval})
 	}
--- a/pkg/prometheus/promcfg_test.go
+++ b/pkg/prometheus/promcfg_test.go
@@ -5962,6 +5962,20 @@ func TestScrapeConfigSpecConfig(t *testing.T) {
 			},
 			golden: "ScrapeConfigSpecConfig_EnableCompression_False.golden",
 		},
+		{
+			name: "enable_http2_is_set_to_true",
+			scSpec: monitoringv1alpha1.ScrapeConfigSpec{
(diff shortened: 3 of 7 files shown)


pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -280,6 +280,11 @@ type ScrapeConfigSpec struct {
 	// If unset, Prometheus uses true by default.
 	// +optional
 	EnableCompression *bool `json:"enableCompression,omitempty"`
+	// When false, Prometheus will request uncompressed response from the scraped target.
+	//
+	// If unset, Prometheus uses true by default.
```
slashpai, line 285: nit: remove this line

pkg/prometheus/promcfg_test.go
```diff
@@ -5962,6 +5962,20 @@ func TestScrapeConfigSpecConfig(t *testing.T) {
...
+			},
+			golden: "ScrapeConfigSpecConfig_EnableHTTP2_True.golden",
+		},
+		{
+			name: "enable_http2_is_set_to_false",
+			scSpec: monitoringv1alpha1.ScrapeConfigSpec{
+				EnableHTTP2: ptr.To(false),
+			},
+			golden: "ScrapeConfigSpecConfig_EnableHTTP2_False.golden",
+		},
```
slashpai, line 5978: may be you need a case when enable http2 not set as well?

slashpai (commented):
lgtm just a couple of nits

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
nutmos: I updated the description to match another EnableHTTP2 instead.

slashpai: aah sorry, we do have a few places which explicitly states Prometheus uses certain default value.
May be we should remove that, since this is rare scenario like bool value is true by default its ok to keep.
But I don't have a strong opinion.

@simonpasquier wdyt?

simonpasquier (approved):

nutmos:
@simonpasquier I updated with rebase already.

pkg/prometheus/promcfg.go
```diff
@@ -2737,6 +2737,10 @@ func (cg *ConfigGenerator) generateScrapeConfig(
 		cfg = cg.WithMinimumVersion("2.49.0").AppendMapItem(cfg, "enable_compression", *sc.Spec.EnableCompression)
 	}
 
+	if sc.Spec.EnableHTTP2 != nil {
+		cfg = cg.AppendMapItem(cfg, "enable_http2", *sc.Spec.EnableHTTP2)
```
simonpasquier, line 2741: sorry I missed this in my previous review but we need to check the Prometheus version.
```suggestion
		cfg = cg.WithMinimumVersion("2.35.0").AppendMapItem(cfg, "enable_http2", *sc.Spec.EnableHTTP2)
```

pkg/prometheus/promcfg_test.go
```diff
@@ -6041,6 +6041,20 @@ func TestScrapeConfigSpecConfig(t *testing.T) {
 			},
 			golden: "ScrapeConfigSpecConfig_EnableCompression_False.golden",
 		},
+		{
```
simonpasquier, line 6044: we would need to add a test with Prometheus version < 2.35.0.

nutmos:
@simonpasquier I updated the check and added the new tests already.

mviswanathsai:
LGTM, only concern is how we deal with unsupported versions. I remember that we wanted to reject resources containing unsupported fields. Would we want to do that here or in a follow up PR maybe? @simonpasquier 
Besides that, I think we should log an issue, will go ahead and do that.

simonpasquier (approved):
Thanks!

mviswanathsai:
> > LGTM, only concern is how we deal with unsupported versions. I remember that we wanted to reject resources containing unsupported fields. Would we want to do that here or in a follow up PR maybe? @simonpasquier Besides that, I think we should log an issue, will go ahead and do that.
> 
> This is https://github.com/prometheus-operator/prometheus-operator/discussions/6792 right?

Right! Forgot about it.

PR #7054, opened 2024-10-29
feat: support `additionalArgs` in the Alertmanager CRD

verejoel:
## Description

Currently it's not supported to add additional arguments to the Alertmanager. Previously this functionality was added for Prometheus and ThanosRuler. This PR also makes it possible to add extra arguments for the Alertmanager.

## Type of change

## Verification

## Changelog entry

```release-note
- feat: support `additionalArgs` in the Alertmanager CRD (#7054)
```


verejoel:
@simonpasquier I think this is ready. The checks are failing with 
```
code-of-conduct.md:50: "http://contributor-covenant.org/version/1/2/0/" not accessible even after retry; status code 0: Get "http://contributor-covenant.org/version/1/2/0/": dial tcp 13.57.148.141:80: i/o timeout
		code-of-conduct.md:49: "http://contributor-covenant.org/" not accessible even after retry; status code 0: Get "http://contributor-covenant.org/": dial tcp 13.57.148.141:80: i/o timeout
```
which also happens on my local machine.

I just noticed there is a release planned for 30.10. I guess it's a little late to try and squeak this one by? 🙈

verejoel:
Actuall I found a potentiall thorny issue. Currently to disable HA we must call `--cluster.listen-address=` with an empty string. That is not possible with the current way `Argument` is implemented...I think we would need to have `Value` as a pointer to string, and then we can distinguish between `nil`, and an empty string. But then probably we change a lot of code for this one use case.

WDYT?

simonpasquier:
https://github.com/prometheus-operator/prometheus-operator/pull/7057 is the PR to drop support for Alertmanager < 0.15.

simonpasquier:
Also you'll need to rebase on top of the main branch which has the fix for the documentation CI check.

simonpasquier:
> Actually I found a potentially thorny issue.
> 
> Currently to disable HA we must call `--cluster.listen-address=` with an empty string. That is not possible with the current way `Argument` is implemented...I think we would need to have `Value` as a pointer to string, and then we can distinguish between `nil`, and an empty string. But then probably we change a lot of code for this one use case.
> 
> WDYT?

Would we get the same effect if we set the value to `""`? e.g.  

```
clusterListenAddressFlag.Value = `""`
````

**Comment by verejoel** [comment]
> Would we get the same effect if we set the value to `""`? e.g.

Have tried this, it renders out as 
```
|| 	            	 (string) (len=27) "--cluster.listen-address=\"\"",
```
I can call Alertmanager with this `--cluster.listen-address=""` and it works the same as `--cluster.listen-address=` but I'm still a little nervous about it. Have updated the tests to show the new state.

File: pkg/alertmanager/statefulset.go [hunk 1]
```diff
@@ -223,71 +223,73 @@ func makeStatefulSetSpec(logger *slog.Logger, a *monitoringv1.Alertmanager, conf
...
 	if a.Spec.RoutePrefix != "" {
 		webRoutePrefix = a.Spec.RoutePrefix
 	}
-	amArgs = append(amArgs, fmt.Sprintf("--web.route-prefix=%v", webRoutePrefix))
+	amArgs = append(amArgs, monitoringv1.Argument{Name: "web.route-prefix", Value: webRoutePrefix})
 
 	web := a.Spec.Web
 	if version.GTE(semver.MustParse("0.17.0")) && web != nil && web.GetConcurrency != nil {
-		amArgs = append(amArgs, fmt.Sprintf("--web.get-concurrency=%d", *web.GetConcurrency))
+		amArgs = append(amArgs, monitoringv1.Argument{Name: "web.get-concurrency", Value: fmt.Sprintf("%d", *web.GetConcurrency)})
```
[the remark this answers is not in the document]
verejoel: Had to additionally cast to int() inside Itoa.

pkg/alertmanager/statefulset.go
```diff
@@ -223,71 +224,73 @@ func makeStatefulSetSpec(logger *slog.Logger, a *monitoringv1.Alertmanager, conf
...
+		{Name: "config.file", Value: path.Join(alertmanagerConfigDir, alertmanagerConfigFileCompressed)},
+		{Name: "storage.path", Value: alertmanagerStorageDir},
+		{Name: "data.retention", Value: string(a.Spec.Retention)},
 	}
 
-	if *a.Spec.Replicas == 1 && !a.Spec.ForceEnableClusterMode {
-		amArgs = append(amArgs, "--cluster.listen-address=")
-	} else {
-		amArgs = append(amArgs, "--cluster.listen-address=[$(POD_IP)]:9094")
+	clusterListenAddressArg := monitoringv1.Argument{Name: "cluster.listen-address", Value: `""`}
```
verejoel, line 233: @simonpasquier I think this works. I guess if the e2e tests pass we are i the clear.

pkg/alertmanager/statefulset_test.go
```diff
@@ -870,7 +870,7 @@ func TestClusterListenAddressForSingleReplica(t *testing.T) {
 	containsEmptyClusterListenAddress := false
 
 	for _, arg := range amArgs {
-		if arg == "--cluster.listen-address=" {
+		if arg == `--cluster.listen-address=""` {
```
verejoel, line 873: Updated the test here to reflect the change in how this will be passed.

verejoel:
@simonpasquier I extended the Argument struct in an attempt to do this in a slightly cleaner way. The upgrade e2e test is still failing though. Do you see a reason for that?

PR #7081, opened 2024-11-07
chore: Optimize go runtime metrics

dongjiang1989:
## Description

Ref: https://github.com/golang/go/issues/49881
Optimize go runtime metrics for operator, admission-webhook and prometheus-config-reloader

## Type of change

## Verification

## Changelog entry

```release-note
Optimize go runtime metrics for operator, admission-webhook and prometheus-config-reloader
```

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package util
+package metrics
 
 import (
 	"regexp"
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -22,7 +22,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/collectors/version"
 )
 
-func NewCollector(name string, cs ...prometheus.Collector) *prometheus.Registry {
+func NewRegistry(program string) *prometheus.Registry {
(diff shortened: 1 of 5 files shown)


simonpasquier:
thanks! since we have 3 times the same code, can we add a small function (in internal/metrics for instance)?

dongjiang1989:
> thanks! since we have 3 times the same code, can we add a small function (in internal/metrics for instance)?

Thanks @simonpasquier  PTAL

internal/metrics/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"regexp"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
+	"github.com/prometheus/client_golang/prometheus/collectors/version"
+)
+
+func NewCollector(name string, cs ...prometheus.Collector) *prometheus.Registry {
```
simonpasquier, line 25: if we don't pass custom collectors for now, let's drop the optional args.
```suggestion
func NewRegistry(program string) *prometheus.Registry {
```

dongjiang1989: Thanks. Fixed.

simonpasquier (approved):
thanks!

PR #6580, opened 2024-05-12
WIP[#5345]: RoleSelector field in ServiceMonitor (#5345)

yp969803:
## Description

If it fixes a bug or resolves a feature request, be sure to link to that issue. #5345 

Be able to define selectors in kubernetes_sd_config

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [x] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

Issue #5345, opened 2023-02-16
Be able to define role selectors in ServiceMonitor and PodMonitor


torrescd:
**What is missing?**

Parameter to define selectors config in generated kubernetes_sd_configs.

Documentation here. https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config
```

# Optional label and field selectors to limit the discovery process to a subset of available resources.
# See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
# and https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ to learn more about the possible
# filters that can be used. The endpoints role supports pod, service and endpoints selectors.
# The pod role supports node selectors when configured with `attach_metadata: {node: true}`.
# Other roles only support selectors matching the role itself (e.g. node role can only contain node selectors).

# Note: When making decision about using field/label selector make sure that this
# is the best approach - it will prevent Prometheus from reusing single list/watch
# for all scrape configs. This might result in a bigger load on the Kubernetes API,
# because per each selector combination there will be additional LIST/WATCH. On the other hand,
# if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
# Decision, if selectors should be used or not depends on the particular situation.
[ selectors:
  [ - role: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string)
    [ label: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string) ]
    [ field: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string) ] ]]
```
Im willing to define them for ServiceMonitor CRDs, but probably applies in more use cases.

**Why do we need it?**

Prometheus allows it

JoaoBraveCoding:
Hello, @torrescd I'm not sure I fully understand your vision for adding `kubernetes_sd_configs` to ServiceMonitor. Could you please provide a bit more detail on your use case?

Also I'm not sure if you are aware but we are now implementing the ScrapeConfig CR that it might fit your use case https://github.com/prometheus-operator/prometheus-operator/pull/5335 based on the design doc https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202212-scrape-config.md

torrescd:
Hi Joao, thank you for your answer.

I think that ScrapeConfig it is not needed in my use case.

See here this function.

[...]

PR #7086, opened 2024-11-07
[FEATURE] adding RoleSelector field in ServiceMonitor

nicolastakashi:
## Description

If it fixes a bug or resolves a feature request, be sure to link to that issue. https://github.com/prometheus-operator/prometheus-operator/issues/5345

Be able to define selectors in kubernetes_sd_config

Overtaking https://github.com/prometheus-operator/prometheus-operator/pull/6580 to solve https://github.com/prometheus-operator/prometheus-operator/issues/5345

## Type of change

## Verification

## Changelog entry

```release-note
Adding RoleSelector field in ServiceMonitor
```

### DIFF
--- a/pkg/apis/monitoring/v1/servicemonitor_types.go
+++ b/pkg/apis/monitoring/v1/servicemonitor_types.go
@@ -83,6 +83,18 @@ type ServiceMonitorSpec struct {
 
 	// Label selector to select the Kubernetes `Endpoints` objects to scrape metrics from.
 	Selector metav1.LabelSelector `json:"selector"`
+
+	// Mechanism used to select the endpoints to scrape.
+	// By default, the selection process relies on relabel configurations to filter the discovered targets.
+	// Alternatively, you can opt in for role selectors, which may offer better efficiency in large clusters.
+	// Using role selectors might reduce Prometheus resource requirements Which strategy is best for your use case needs to be carefully evaluated.
+	//
+	// It requires Prometheus >= v2.17.0.
+	//
+	// +kubebuilder:validation:Enum=RelabelConfig;RoleSelector
+	// +kubebuilder:default=RelabelConfig
+	SelectorMechanism SelectorMechanism `json:"selectorMechanism,omitempty"`
+
 	// `namespaceSelector` defines in which namespace(s) Prometheus should discover the services.
 	// By default, the services are discovered in the same namespace as the `ServiceMonitor` object but it is possible to select pods across different/all namespaces.
 	NamespaceSelector NamespaceSelector `json:"namespaceSelector,omitempty"`
--- a/pkg/apis/monitoring/v1/servicemonitor_types.go
+++ b/pkg/apis/monitoring/v1/servicemonitor_types.go
@@ -87,7 +87,7 @@ type ServiceMonitorSpec struct {
 	// Mechanism used to select the endpoints to scrape.
 	// By default, the selection process relies on relabel configurations to filter the discovered targets.
 	// Alternatively, you can opt in for role selectors, which may offer better efficiency in large clusters.
-	// Using role selectors might reduce Prometheus resource requirements Which strategy is best for your use case needs to be carefully evaluated.
+	// Which strategy is best for your use case needs to be carefully evaluated.
 	//
 	// It requires Prometheus >= v2.17.0.
 	//
--- a/pkg/apis/monitoring/v1/servicemonitor_types.go
+++ b/pkg/apis/monitoring/v1/servicemonitor_types.go
@@ -92,8 +92,8 @@ type ServiceMonitorSpec struct {
 	// It requires Prometheus >= v2.17.0.
 	//
 	// +kubebuilder:validation:Enum=RelabelConfig;RoleSelector
-	// +kubebuilder:default=RelabelConfig
-	SelectorMechanism SelectorMechanism `json:"selectorMechanism,omitempty"`
+	// +optional
+	SelectorMechanism *SelectorMechanism `json:"selectorMechanism,omitempty"`
 
 	// `namespaceSelector` defines in which namespace(s) Prometheus should discover the services.
 	// By default, the services are discovered in the same namespace as the `ServiceMonitor` object but it is possible to select pods across different/all namespaces.
--- a/pkg/apis/monitoring/v1/servicemonitor_types.go
+++ b/pkg/apis/monitoring/v1/servicemonitor_types.go
@@ -91,7 +91,6 @@ type ServiceMonitorSpec struct {
 	//
 	// It requires Prometheus >= v2.17.0.
 	//
-	// +kubebuilder:validation:Enum=RelabelConfig;RoleSelector
 	// +optional
 	SelectorMechanism *SelectorMechanism `json:"selectorMechanism,omitempty"`
 
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -2163,7 +2163,7 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 	}
 }
 
-func generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice, roles []string, selector *metav1.LabelSelector, cg *ConfigGenerator) yaml.MapSlice {
+func (cg *ConfigGenerator) generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice, roles []string, selector *metav1.LabelSelector) yaml.MapSlice {
 	selectors := []yaml.MapSlice{}
 	for _, role := range roles {
 		yml := yaml.MapSlice{
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -2178,27 +2178,19 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 }
 
 func (cg *ConfigGenerator) generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice, roles []string, selector metav1.LabelSelector) yaml.MapSlice {
-	selectors := []yaml.MapSlice{}
+	selectors := make([]yaml.MapSlice, 0, len(roles))
+	labelSelector := labels.SelectorFromValidatedSet(labels.Set(selector.MatchLabels))
+
+	for _, exp := range selector.MatchExpressions {
+		requirement, _ := labels.NewRequirement(exp.Key, selection.Operator(strings.ToLower(string(exp.Operator))), exp.Values)
+		labelSelector = labelSelector.Add(*requirement)
+	}
+
 	for _, role := range roles {
-		yml := yaml.MapSlice{
-			{
-				Key:   "role",
-				Value: role,
-			},
-		}
-
-		labelSelector := labels.SelectorFromValidatedSet(labels.Set(selector.MatchLabels))
-
-		if len(selector.MatchExpressions) > 0 {
-			for _, exp := range selector.MatchExpressions {
-				// Needs to lower because the selection expects lowercase, and the exp.Operator is CamelCase.
-				// Ignoring error because we are validating on the resource selection.
-				requirement, _ := labels.NewRequirement(exp.Key, selection.Operator(strings.ToLower(string(exp.Operator))), exp.Values)
-				labelSelector = labelSelector.Add(*requirement)
-			}
-		}
-		yml = append(yml, yaml.MapItem{Key: "label", Value: labelSelector.String()})
-		selectors = append(selectors, yml)
+		selectors = append(selectors, yaml.MapSlice{
+			{Key: "role", Value: role},
+			{Key: "label", Value: labelSelector.String()},
+		})
 	}
 
 	for i, item := range k8sSDConfig {
@@ -2207,6 +2199,7 @@ func (cg *ConfigGenerator) generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice,
 			return k8sSDConfig
 		}
 	}
+
 	return cg.AppendMapItem(k8sSDConfig, "selectors", selectors)
 }
 
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -2165,10 +2165,6 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 		})
 	}
 
-	// if ptr.Deref(selectorMechanism, monitoringv1.SelectorMechanismRelabel) == monitoringv1.SelectorMechanismRole {
-	// 	k8sSDConfig = cg.generateRoleSelectorConfig(k8sSDConfig, selectorRoles, selector)
-	// }
-
 	for _, opt := range opts {
 		k8sSDConfig = opt(k8sSDConfig)
 	}
@@ -2195,13 +2191,9 @@ func (cg *ConfigGenerator) generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice,
 
 		if len(selector.MatchExpressions) > 0 {
 			for _, exp := range selector.MatchExpressions {
-				// Needs to lower because the selection expects lowercase.
-				// exp.Operator is CamelCase.
-				requirement, err := labels.NewRequirement(exp.Key, selection.Operator(strings.ToLower(string(exp.Operator))), exp.Values)
-				if err != nil {
-					cg.logger.Error("failed to create label requirement", "err", err)
-					continue
-				}
+				// Needs to lower because the selection expects lowercase, and the exp.Operator is CamelCase.
+				// Ignoring error because we are validating on the resource selection.
+				requirement, _ := labels.NewRequirement(exp.Key, selection.Operator(strings.ToLower(string(exp.Operator))), exp.Values)
 				labelSelector = labelSelector.Add(*requirement)
 			}
 		}
@@ -2209,6 +2201,12 @@ func (cg *ConfigGenerator) generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice,
 		selectors = append(selectors, yml)
 	}
 
+	for i, item := range k8sSDConfig {
+		if item.Key == "selectors" {
+			k8sSDConfig[i].Value = append(k8sSDConfig[i].Value.([]yaml.MapSlice), selectors...)
+			return k8sSDConfig
+		}
+	}
 	return cg.AppendMapItem(k8sSDConfig, "selectors", selectors)
 }
 
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -33,6 +33,8 @@ import (
 	"gopkg.in/yaml.v2"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
 
@@ -1187,7 +1189,7 @@ func (cg *ConfigGenerator) generatePodMonitorConfig(
 
 	s := store.ForNamespace(m.Namespace)
 
-	cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, s, kubernetesSDRolePod, attachMetaConfig))
+	cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, s, kubernetesSDRolePod, attachMetaConfig, nil, nil, nil))
 
 	if ep.Interval != "" {
 		cfg = append(cfg, yaml.MapItem{Key: "scrape_interval", Value: ep.Interval})
@@ -1571,7 +1573,7 @@ func (cg *ConfigGenerator) generateProbeConfig(
 			}
 		}
 
-		cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.Targets.Ingress.NamespaceSelector, m.Namespace, apiserverConfig, s, kubernetesSDRoleIngress, nil))
+		cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.Targets.Ingress.NamespaceSelector, m.Namespace, apiserverConfig, s, kubernetesSDRoleIngress, nil, nil, nil, nil))
 
 		// Relabelings for ingress SD.
 		relabelings = append(relabelings, []yaml.MapSlice{
@@ -1675,7 +1677,7 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 	s := store.ForNamespace(m.Namespace)
 
 	role := cg.endpointRoleFlavor()
-	cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, s, role, attachMetaConfig))
+	cfg = append(cfg, cg.generateK8SSDConfig(m.Spec.NamespaceSelector, m.Namespace, apiserverConfig, s, role, attachMetaConfig, &m.Spec.Selector, &m.Spec.SelectorMechanism, ptr.To(monitoringv1alpha1.KubernetesRoleService)))
 
 	if ep.Interval != "" {
 		cfg = append(cfg, yaml.MapItem{Key: "scrape_interval", Value: ep.Interval})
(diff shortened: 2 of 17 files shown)


pkg/apis/monitoring/v1/servicemonitor_types.go
```diff
@@ -82,7 +82,8 @@ type ServiceMonitorSpec struct {
 	Endpoints []Endpoint `json:"endpoints"`
 
 	// Label selector to select the Kubernetes `Endpoints` objects to scrape metrics from.
-	Selector metav1.LabelSelector `json:"selector"`
+	Selector EndpointSelector `json:"selector"`
```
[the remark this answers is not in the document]
simonpasquier: Reading @bboreham's [comment](https://github.com/prometheus-operator/prometheus-operator/issues/5345#issuecomment-2451684926), I assume that the request is to offer a way to generate a config which is

````
  kubernetes_sd_configs:
  - role: endpoints
    selectors:
    - role: service
      label: foo=bar
    - role: endpoints
      label: foo=bar
  relabel_configs:
  - ...
````

instead of the current

```
  kubernetes_sd_configs:
  - role: endpoints
  relabel_configs:
  - ...
  - action: keep
    source_labels:
    - __meta_kubernetes_service_label_foo
    - __meta_kubernetes_service_labelpresent_foo
    regex: (bar);true
  - ...
```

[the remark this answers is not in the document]
Comment (nicolastakashi) [reply pkg/apis/monitoring/v1/servicemonitor_types.go:85]: friendly reminder @simonpasquier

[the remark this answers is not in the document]
Comment (nicolastakashi) [reply pkg/apis/monitoring/v1/servicemonitor_types.go]: @simonpasquier PTAL 🙏🏽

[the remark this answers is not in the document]
Comment (nicolastakashi) [reply pkg/apis/monitoring/v1/servicemonitor_types.go]: @simonpasquier @slashpai friendly ping

File: pkg/apis/monitoring/v1/servicemonitor_types.go [hunk 2]
```diff
@@ -83,6 +85,13 @@ type ServiceMonitorSpec struct {
 
 	// Label selector to select the Kubernetes `Endpoints` objects to scrape metrics from.
 	Selector metav1.LabelSelector `json:"selector"`
+
+	// Role selector to select the Kubernetes `Endpoints` objects to scrape metrics from.
+	// If `roleSelector` is defined, the `selector` field is ignored.
```
simonpasquier, line 90: I feel that the API doesn't provide a super-clean user experience.

Say that you have

````
spec:
  selector:
    matchLabels: {foo: bar}
````

If you want to use role selectors instead, you could change it to the following but it requires to do the translation "by hand".

````
spec:
  roleSelector:
  - role: Service
    labels:
    - key: foo
      operator: In
      values: [bar]
  - role: Endpoints
    labels:
    - key: foo
      operator: In
      values: [bar]
````

But you can also define bogus configs like:

````
spec:
  selector:
    matchLabels: {foo: bar}
  # because role selector list isn't empty, the selector field is completely ignored
  roleSelector:
  - role: Pod
    labels:
    - key: other_label
      operator: In
      values: [a,b,c]
````

Which is why I advocated for an enum field which would tell the operator to use role selectors instead of relabeling config to filter the targets: it would probably be enough for most users (and less error prone). And we can still expose role selectors for other resources.

File: pkg/apis/monitoring/v1/servicemonitor_types.go [hunk 3]
```diff
@@ -83,6 +83,16 @@ type ServiceMonitorSpec struct {
 
 	// Label selector to select the Kubernetes `Endpoints` objects to scrape metrics from.
 	Selector metav1.LabelSelector `json:"selector"`
+
+	// Mechanism used to select the endpoints to scrape.
```
simonpasquier, line 87: It requires Prometheus >= v2.17.0. It's very unlikely that people run such old versions but to be consistent, I'd advise to reject the resource if the role selector mechanism is defined and the version doesn't meet the requirement.

pkg/prometheus/testdata/ServiceMonitorObjectWithMatchExpressionSelector.golden
```diff
@@ -0,0 +1,78 @@
...
+scrape_configs:
+- job_name: serviceMonitor/default/defaultServiceMonitor/0
+  honor_labels: false
+  kubernetes_sd_configs:
+  - role: endpoints
+    namespaces:
+      names:
+      - default
+    selectors:
+    - role: endpoints
```
simonpasquier, line 16: should we also add a service role selector?
```suggestion
    - role: service
      label: group in (group1)
    - role: endpoints
```

pkg/prometheus/promcfg_test.go
```diff
@@ -12051,3 +12061,89 @@ func TestAlertmanagerTLSConfig(t *testing.T) {
 
 	}
 }
+
+func TestServiceMonitorSelectors(t *testing.T) {
```
simonpasquier, line 12065: can we add a test with a Prometheus resource which has `.spec.serviceDiscoveryRole: EndpointSlice`? It should generate `role: endpointslice` rather than endpoints.

pkg/prometheus/testdata/ServiceMonitorObjectWithSelectorAndRoleSelector.golden
simonpasquier, line 1: is it used?

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2151,44 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 	}
 }
 
+func generateRoleSelectorConfig(role string, selector *metav1.LabelSelector, k8sSDConfig yaml.MapSlice, cg *ConfigGenerator) yaml.MapSlice {
```
simonpasquier, line 2154: can we use https://pkg.go.dev/k8s.io/apimachinery@v0.31.3/pkg/labels to generate the selector string instead of building it ourselves?

pkg/prometheus/testdata/ServiceMonitorObjectWithMatchExpressionSelector.golden
nicolastakashi: Actually I'm not sure, I'd say no.
But wdyt? @simonpasquier Do you have any nice use case?

pkg/prometheus/promcfg_test.go
nicolastakashi: I'll move away from endpoints and endpoints slices and only use service role, since we're already filtering by service labels, this will simplify tests and ensure we're operate over the already used labels

pkg/apis/monitoring/v1/servicemonitor_types.go
```diff
@@ -83,6 +83,18 @@ type ServiceMonitorSpec struct {
...
+	// By default, the selection process relies on relabel configurations to filter the discovered targets.
+	// Alternatively, you can opt in for role selectors, which may offer better efficiency in large clusters.
+	// Using role selectors might reduce Prometheus resource requirements Which strategy is best for your use case needs to be carefully evaluated.
```
simonpasquier, line 90: (nit) reading again we say the same thing twice.
```suggestion
	// Which strategy is best for your use case needs to be carefully evaluated.
```

pkg/apis/monitoring/v1/servicemonitor_types.go
```diff
@@ -83,6 +83,18 @@ type ServiceMonitorSpec struct {
...
+	//
+	// It requires Prometheus >= v2.17.0.
```
simonpasquier, line 92: BTW we need to enforce this in the resource selector.

pkg/apis/monitoring/v1/servicemonitor_types.go
```diff
@@ -83,6 +83,16 @@ type ServiceMonitorSpec struct {
...
 	Selector metav1.LabelSelector `json:"selector"`
+
+	// Mechanism used to select the endpoints to scrape.
+	// By default, the selection process relies on relabel configurations to filter targets.
+	// Alternatively, you can opt for role selectors, which may offer better efficiency in large clusters.
+	// Using role selectors bypasses the need for complex relabeling, making it more suitable for high-scale environments.
+	//
+	// +kubebuilder:validation:Enum=RelabelConfig;RoleSelector
+	// +kubebuilder:default=RelabelConfig
+	SelectorMechanism SelectorMechanism `json:"selectorMechanism,omitempty"`
```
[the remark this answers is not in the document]
simonpasquier: still valid?

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2154,39 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 	}
 }
 
+func generateRoleSelectorConfig(role *monitoringv1alpha1.KubernetesRole, selector *metav1.LabelSelector, k8sSDConfig yaml.MapSlice, cg *ConfigGenerator) yaml.MapSlice {
```
simonpasquier, line 2157: (nit) can we have `k8sSDConfig yaml.MapSlice` as the first arg (to mimic the append() function)?

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2154,39 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+			Value: strings.ToLower(string(*role)),
+		},
+	}
+
+	if len(selector.MatchLabels) > 0 {
+		yml = append(yml, yaml.MapItem{Key: "label", Value: labels.FormatLabels(selector.MatchLabels)})
+		selectors = append(selectors, yml)
+	}
+
+	if len(selector.MatchExpressions) > 0 {
```
simonpasquier, line 2171: what if both `MatchLabels` and `MatchExpressions` are defined? should (can?) we merge them?

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2154,39 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+		expressions := []string{}
```
simonpasquier, line 2172: can be labels.Requirements?

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2154,39 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+			if err != nil {
+				cg.logger.Error("failed to create label requirement", "err", err)
+				continue
+			}
+			expressions = append(expressions, requirement.String())
+		}
+		yml = append(yml, yaml.MapItem{Key: "label", Value: strings.Join(expressions, ",")})
+		selectors = append(selectors, yml)
+	}
+	return cg.WithMinimumVersion("2.17.0").AppendMapItem(k8sSDConfig, "selectors", selectors)
```
simonpasquier, line 2187: I would rather drop the resource earlier to avoid false expectations.

pkg/prometheus/promcfg_test.go
simonpasquier: I think that it makes sense to filter both services and endpoints(lice). Otherwise we're not meeting the user expectations (reducing the number of resources being watched).

pkg/apis/monitoring/v1/servicemonitor_types.go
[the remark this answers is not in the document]
nicolastakashi: IIRC yes, we need expose a selection option for the users, from what I understood from our call we're going to keep this field and if selected Role it will switch behind the scenes the same role

pkg/prometheus/promcfg.go
nicolastakashi: Maybe I'm missing some package, but I didn't find anything to convert `[]*labels.Requirement{}` to the string representation

pkg/apis/monitoring/v1/servicemonitor_types.go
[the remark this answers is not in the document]
simonpasquier: sorry I was dense: I'd advise to have a pointer field and handle the default value in the controller directly (instead of having the `+kubebuilder:default=RelabelConfig` marker).

pkg/apis/monitoring/v1/servicemonitor_types.go
simonpasquier: we can add a version check in pkg/prometheus/resource_selector.go and reject the ServiceMonitor resource if it defines `selectorMechanism: `RoleSelector

pkg/prometheus/promcfg.go
```diff
@@ -1657,7 +1659,15 @@ func (cg *ConfigGenerator) generateServiceMonitorConfig(
 	apiserverConfig *monitoringv1.APIServerConfig,
 	store *assets.StoreBuilder,
 	shards int32,
-) yaml.MapSlice {
+) (yaml.MapSlice, error) {
+
+	if m.Spec.SelectorMechanism == monitoringv1.SelectorMechanismRole && !cg.version.GTE(semver.MustParse("2.17.0")) {
+		return nil, fmt.Errorf(
+			"RoleSelector is only supported in Prometheus 2.17.0 and newer: serviceMonitor %s/%s",
```
simonpasquier, line 1666: We should reject the resource in `*ResourceSelector.SelectServiceMonitors()`. Otherwise a single invalid ServiceMonitor would block the whole reconciliation.

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2163,46 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 	}
 }
 
+func generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice, roles []string, selector *metav1.LabelSelector, cg *ConfigGenerator) yaml.MapSlice {
```
simonpasquier, line 2166:
```suggestion
func (cg *ConfigGenerator) generateRoleSelectorConfig(k8sSDConfig yaml.MapSlice, roles []string, selector *metav1.LabelSelector) yaml.MapSlice {
```

pkg/prometheus/promcfg.go
```diff
@@ -2134,6 +2151,10 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 		})
 	}
 
+	if ptr.Deref(selectorMechanism, monitoringv1.SelectorMechanismRelabel) == monitoringv1.SelectorMechanismRole {
+		k8sSDConfig = generateRoleSelectorConfig(k8sSDConfig, selectorRoles, selector, cg)
+	}
```
simonpasquier, line 2156: (nit) we could always call generateRoleSelectorConfig() and have the test on the selector mechanism there.
```suggestion
	k8sSDConfig = generateRoleSelectorConfig(k8sSDConfig, selectorRoles, selector, cg)
```

pkg/prometheus/promcfg.go
```diff
@@ -2064,6 +2078,9 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 	store assets.StoreGetter,
 	role string,
 	attachMetadataConfig *attachMetadataConfig,
+	selector *metav1.LabelSelector,
+	selectorMechanism *monitoringv1.SelectorMechanism,
+	selectorRoles []string,
```
simonpasquier, line 2083: I'm not a fan of adding more parameters to this method. Maybe we need to refactor generateK8SSDConfig() so it does less? (the comment could also apply to the `attachMetadataConfig` parameter).
One option would be to introduce wrappers

```
appendRoleSelectors(
    attachMetaConfig.append(
        generateK8SSDConfig(...),
        ...,
    )
    ...,
)
````

File: pkg/prometheus/promcfg.go [hunk 18]
```diff
@@ -2134,6 +2151,10 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 		})
 	}
 
+	if ptr.Deref(selectorMechanism, monitoringv1.SelectorMechanismRelabel) == monitoringv1.SelectorMechanismRole {
+		k8sSDConfig = generateRoleSelectorConfig(k8sSDConfig, selectorRoles, selector, cg)
```
simonpasquier, line 2155: hmm I notice that we also inject the "selectors" key when running in daemonset mode. While DaemonSet mode doesn't support service monitors, it might in the future and it would get messy.

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2163,46 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+	selectors := []yaml.MapSlice{}
+	for _, role := range roles {
+		yml := yaml.MapSlice{
+			{
+				Key:   "role",
+				Value: role,
+			},
+		}
+
+		labelValues := []string{}
```
simonpasquier, line 2176: (nit) 
```suggestion
		labelSelectors := []string{}
```

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2163,46 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+		if len(selector.MatchLabels) > 0 {
+			labelValues = append(labelValues, labels.FormatLabels(selector.MatchLabels))
+		}
+
```
simonpasquier, line 2180: (suggestion) We can create a labels.Selector object so we call labelSelector.Add(requirement) after.
```suggestion
		labelSelector := labels.SelectorFromValidatedSet(labels.Set(selector.MatchLabels))

```

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2163,46 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+		if len(selector.MatchExpressions) > 0 {
+			expressions := []string{}
+
+			for _, exp := range selector.MatchExpressions {
+				// Needs to lower because the selection expects lowercase.
+				// exp.Operator is CamalCase.
```
simonpasquier, line 2186:
```suggestion
				// exp.Operator is CamelCase.
```

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2163,46 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+				requirement, err := labels.NewRequirement(exp.Key, selection.Operator(strings.ToLower(string(exp.Operator))), exp.Values)
+				if err != nil {
+					cg.logger.Error("failed to create label requirement", "err", err)
```
simonpasquier, line 2189: Should we check earlier (in resource selector) that the input is correct?

pkg/prometheus/promcfg.go
nicolastakashi: @simonpasquier you're a genius

pkg/prometheus/promcfg.go
nicolastakashi: I don't think so since it's a LabelSelector and K8S API Will ensure it for us, right?

simonpasquier: Yeah, I don't have a concrete example where validation markers would fail us but it might be a good idea to prevent the potential issue though?

pkg/apis/monitoring/v1/servicemonitor_types.go
```diff
@@ -83,6 +83,18 @@ type ServiceMonitorSpec struct {
...
 	Selector metav1.LabelSelector `json:"selector"`
+
+	// Mechanism used to select the endpoints to scrape.
+	// By default, the selection process relies on relabel configurations to filter the discovered targets.
+	// Alternatively, you can opt in for role selectors, which may offer better efficiency in large clusters.
+	// Which strategy is best for your use case needs to be carefully evaluated.
+	//
+	// It requires Prometheus >= v2.17.0.
+	//
+	// +kubebuilder:validation:Enum=RelabelConfig;RoleSelector
```
simonpasquier, line 94: (nit) we already have validation on the SelectorMechanism type.
```suggestion
```

pkg/prometheus/promcfg.go
```diff
@@ -2134,6 +2165,14 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 		})
 	}
 
+	// if ptr.Deref(selectorMechanism, monitoringv1.SelectorMechanismRelabel) == monitoringv1.SelectorMechanismRole {
```
simonpasquier, line 2168: to be removed?

pkg/prometheus/__debug_bin263883692
simonpasquier, line 1: are you sure that we need  a 60MB file? :stuck_out_tongue_closed_eyes:

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2181,37 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+					continue
+				}
+				labelSelector = labelSelector.Add(*requirement)
+			}
+		}
+		yml = append(yml, yaml.MapItem{Key: "label", Value: labelSelector.String()})
+		selectors = append(selectors, yml)
+	}
+
+	return cg.AppendMapItem(k8sSDConfig, "selectors", selectors)
```
simonpasquier, line 2212: should we check if the "selectors" is already present in k8sSDConfig, just to be safe?

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2177,39 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+		if len(selector.MatchExpressions) > 0 {
+			for _, exp := range selector.MatchExpressions {
+				// Needs to lower because the selection expects lowercase, and the exp.Operator is CamelCase.
+				// Ignoring error because we are validating on the resource selection.
+				requirement, _ := labels.NewRequirement(exp.Key, selection.Operator(strings.ToLower(string(exp.Operator))), exp.Values)
+				labelSelector = labelSelector.Add(*requirement)
+			}
+		}
+		yml = append(yml, yaml.MapItem{Key: "label", Value: labelSelector.String()})
+		selectors = append(selectors, yml)
```
simonpasquier, line 2201: (nit) can we reduce the number of intermediate variables?
```suggestion
		selectors = append(selectors, 
yaml.MapSlice{
			{
				Key:   "role",
				Value: role,
			},
			{
				Key:   "label",
				Value: selector,
			},
		})
```

pkg/prometheus/promcfg_test.go
```diff
@@ -1584,6 +1584,7 @@ func TestNoEnforcedNamespaceLabelServiceMonitor(t *testing.T) {
 					Namespace: "default",
 				},
 				Spec: monitoringv1.ServiceMonitorSpec{
+					SelectorMechanism: ptr.To(monitoringv1.SelectorMechanismRelabel),
```
simonpasquier, line 1587: do we need to define the SelectorMechanism value?

pkg/prometheus/promcfg_test.go
```diff
@@ -12051,3 +12059,156 @@ func TestAlertmanagerTLSConfig(t *testing.T) {
...
 }
+
+func TestServiceMonitorSelectors(t *testing.T) {
+	for _, tc := range []struct {
+		name                 string
+		golden               string
+		serviceMonitor       *monitoringv1.ServiceMonitor
+		serviceDiscoveryRole monitoringv1.ServiceDiscoveryRole
+	}{
+		// {
```
simonpasquier, line 12070: to be removed?

pkg/prometheus/promcfg.go
```diff
@@ -2142,6 +2177,39 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
...
+			},
+		}
+
+		labelSelector := labels.SelectorFromValidatedSet(labels.Set(selector.MatchLabels))
+
+		if len(selector.MatchExpressions) > 0 {
+			for _, exp := range selector.MatchExpressions {
+				// Needs to lower because the selection expects lowercase, and the exp.Operator is CamelCase.
+				// Ignoring error because we are validating on the resource selection.
+				requirement, _ := labels.NewRequirement(exp.Key, selection.Operator(strings.ToLower(string(exp.Operator))), exp.Values)
```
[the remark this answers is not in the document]
nicolastakashi: Do you mean if we have an error on this line panic it?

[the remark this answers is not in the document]
simonpasquier: yes I'm always nervous when we silently swallow errors.

pkg/prometheus/promcfg_test.go
nicolastakashi: Not really, removed

pkg/prometheus/promcfg_test.go
nicolastakashi: it was commented by mistake

simonpasquier (approved):
One small (non-blocking) comment. It would be good to add an e2e test but it can be done as a follow-up.

simonpasquier (approved):

simonpasquier:
:partying_face:

bboreham:
Thanks for this.  Should #5345 be closed now?

bboreham:
Also, I believe this feature was part of release 0.79, but I don't see it on the [release notes](https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.79.0).

slashpai:
> Also, I believe this feature was part of release 0.79, but I don't see it on the [release notes](https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.79.0).
> 
> 

It's there in enhancement section 🙂

[ENHANCEMENT] Add support for role selector field selectorMechanism on ServiceMonitor. #7086

Issue #6437, opened 2024-03-26
Allow a custom label for sharding.


arnecls:
### Component(s)

Prometheus

### What is missing? Please describe.

Currently the recording rule assigning endpoint to shards is hardcoded to `__address__`.

https://github.com/prometheus-operator/prometheus-operator/blob/d87710bd7201425d078374eef1312cb985bd9c08/pkg/prometheus/promcfg.go#L1613

We would like to be able to change the label and/or the method to, for example, allow zone aware sharding.
If all endpoints expose a certain label, let's say `zone`, either directly or through a rewrite rules, one could use this label instead of `__address__`. This would allow more control over how endpoints are distributed amongst the available shards.

In addition to this it would be nice to switch from hash based sharding to "value based sharding" where each shard is reacting to a specific value. This could be used to, for example, assign a specific shard to a specific zone, removing the need to find out which shard is scraping which zone after hashing.

### Describe alternatives you've considered.

We're not aware of any alternative method to inject a "per-shard" rewrite rule to all metrics handled by a specific shard.

## Environment
Kubernetes Version: -
Prometheus-Operator Version: 0.72.0

simonpasquier:
I thought we already had an issue about this but I can't find it but definitely :+1: for the idea of zone sharding.
Can you elaborate a bit more how this could work? My understanding is that to shard on zones, Prometheus needs to attach node's metadata to the targets.
Also do you foresee other target labels which would be good candidates for sharding?

arnecls:
> Can you elaborate a bit more how this could work? My understanding is that to shard on zones, Prometheus needs to attach node's metadata to the targets.

The main issue here is that zones are cloud provider specific and not all scraping configs support the same set of labels.

My idea (for our case) is that we have re-labeling rules for different cases which generate a unified label. That would only work though if dropping happens _after_ these rules have been evaluated.

For pod based metrics the idea was to use Kyverno (or some other suitable controller) to add a label on each pod containing the zone. A rule for this case could look like this:

```yaml
  - source_labels: [__meta_kubernetes_pod_label_zone] # generated by us, same for all shards
    separator: ;
    regex: (.+)
    target_label: zone
    replacement: ${1}
    action: replace
  - source_labels: [zone] # generated by operator per shard
    separator: ;
    regex: (.*)
    modulus: 1
    target_label: __tmp_hash
    replacement: $1
    action: hashmod
  - source_labels: [__tmp_hash]
    separator: ;
[...]

PR #7110, opened 2024-11-13
[DOC] add a proposal for zone aware scraping

arnecls:
## Description

Document Proposal for zone aware scraping
Related to: #6437

## Type of change

## Verification

The proposed configurations were not validated yet.

## Changelog entry

N/A


Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,215 @@
...
+assigned target), we need to generate [node affinities](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity).
+
+The corresponding section in a kubernetes podspec will look like as follow.  
+We use the same Helm templating notation as above.
+
+```yaml
+{{ $shardsPerZone := max 1 (.spec.shards | div (len .spec.shardingStrategy.Topology)) }}
+spec:
+  affinity:
+    nodeAffinity:
```
arnecls, line 199: Just realized we could also use `NodeSelector` instead.
https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes/#create-a-pod-that-gets-scheduled-to-your-chosen-node

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
...
+
+### topology mode
+
+```yaml
+  {{ $labelSuffix := .spec.shardingStrategy.label | replace "/" "_" | replace "." "_" }}
+  {{ $shardsPerZone := max 1 (.spec.shards | div (len .spec.shardingStrategy.Topology)) }}
+  {{ $promZoneIdx := .operator.shardIndex | mod $shardsPerZone }}
+  
+  {{ .spec.shardingStrategy.additionalRelabelConfig | toYaml }}
+  - source_labels: [ '__meta_kubernetes_pod_label_{{ default "topology_kubernetes_io_zone" $labelSuffix }}' ]
```
arnecls, line 170: We might want to make this configurable, too.
Something like this:

```suggestion
{{ $labelPrefix := .spec.shardingStrategy.labelPrefix | replace "/" "_" | replace "." "_" }}
  - source_labels: [ '{{ default "__meta_kubernetes_pod_label_" $labelPrefix }}{{ default "topology_kubernetes_io_zone" $labelSuffix }}' ]
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
...
+When `num_shards` is 2 and `len(zones)` is 3 as in `[A..C]`.
+`shards_per_zone` is 1. This yields the following distribution
+
+```
+0 1   | shard index
+A B C | zone
+0 0 0 | assignment index
+```
+
+In this case the 1st assert will warn about zone C not being scraped.
```
simonpasquier, line 114: should it be prevented by the operator? E.g. it could fail the reconciliation if shards and zones don't match.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Required configuration options
+
+Following the algorithm presented above, we suggest the following configuration
+options to be added to the [Prometheus custom resource definition](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.Prometheus):
+
+```yaml
+spec:
+  shardingStrategy:
+    mode: 'classic'             # can be 'classic' or 'topology'
```
simonpasquier, line 124: (nit) it should be `Classic` and `Topology` to follow the API conventions.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
...
+- Implement mechanisms to avoid configuration errors by the user
+- Support mixed environments (kubernetes and non-kubernetes targets are scraped)
+
+## Algorithm
+
+In order to do calculate a stable assignment, following parameters are required:
+
+1. `num_shards`: The number of prometheus shards
+2. `shard_index`: A number of the range `[0..num_shards[` identifying a single prometheus 
+   instance
```
simonpasquier, line 48: (nit)
```suggestion
   instance inside a given shard
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
...
+
+The `additionalRelabelConfig` section is meant to allow the `label` to be
+generated if needed. This should allow enough flexibility to cover edgecases
+not anticiapted by this proposal.
+
+## Generated configuration
+
+The following configuration uses Helm templating syntax and the proposed
+configuration structure as proposed above. Functions from the [sprig](https://masterminds.github.io/sprig/)
+library are used, as they are also present in Helm.
```
simonpasquier, line 139: I'm not sure to understand: are you saying that the configuration generated by the operator will processed by a template engine before being loaded into Prometeus? or is it only provided as an example?

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
...
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: '{{ .spec.shardingStrategy.label }}'
+            operator: 'In'
+            values:
+            - '{{ index .spec.shardingStrategy.Topology (.operator.shardIndex | div $shardsPerZone) }}'
+```
+
+The following section shows how a node selector would be implemented.
```
simonpasquier, line 212: node selector would have my preference because it's simpler to use.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
...
+
+```yaml
+spec:
+  shardingStrategy:
+    mode: 'classic'             # can be 'classic' or 'topology'
+    label: '__address__'        # defaults to '__address__` for 'classic' and 
+                                # 'topology.kubernetes.io/zone' for 'topology'
+    topology: []                # expects list of zones for 'toplogy'
+    additionalRelabelConfig: [] # Optional: array of relabel configurations
+```
```
simonpasquier, line 129: I would have different subsections to define each mode's settings, similar to what exists for Deployment:

````yaml
spec:
  shardingStrategy:
    type: Topology
    topology:
      zones: [...]
````

File: Documentation/proposals/202411-zone-aware-sharding.md [hunk 9]
```diff
@@ -0,0 +1,227 @@
...
+
```
simonpasquier, line 130: suggestion: 1 example snippet per mode would be useful to image how the API would look like (like you did for the generated config).

simonpasquier (commented):
This is a great start, thanks for doing this! Out of curiosity is it something that you've already implemented outside of the operator?

cc @slashpai @nicolastakashi @ArthurSens

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: No, I just wanted to use some commonly used „language“ to express what should happen here.

I also wanted to use this to test the algorithm but I have not yet set up a test environment to actually do that.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: I was under the impression that a „shard“ is an instance. Or is it a group of instances in the prometheus context?

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: I don’t know what mechanisms exist in the operator.
Both could be an error stopping the reconciliation loop. At minimum the first one, because data is lost in that case. In the second case there is duplicate data, which is not ideal, but at least nothing is lost.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,227 @@
+# Status subresource for Prometheus operator CRDs
```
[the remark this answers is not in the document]
arnecls: Oops 😅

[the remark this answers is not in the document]
arnecls: fixed

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: fixed.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: added an explanation to the section

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: Added more details: input and expected output.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: removed node affinity and explained why.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: fixed.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: Ah - because a shard can also have multiple replica?

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+spec:
+  shards: 4
+  shardingStrategy:
+    mode: 'Classic'    
+    additionalRelabelConfig: [] 
+    classic:
+      sourceLabel: '__address__'
+```
+
+And the relabel configuration being genereated through this template
```
simonpasquier, line 189: (nit)
```suggestion
And the relabel configuration being generated through this template
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+## Generated configuration
+
+In order to illustrate configuration generator mechanics, the following
+examples use Helm templating syntax.
+Variables are based on the proposed configuration structure as defined above.
+Functions from the [sprig](https://masterminds.github.io/sprig/) library are
+used, as they are also present in Helm.
+
+Values generated by the operator, like `shard_index` are retrieved from an
+`.operator` object.
```
simonpasquier, line 173: (suggestion) I would drop the intermediate template representation and keep only the final generated config with comments to help the understanding.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+  replacement: '$1'
+  action: 'keep'
+```
+
+we would get the following output for `shard_index == 2`
+
+```yaml
+- source_labels: [ '__address__' ] 
+  separator: ;
+  modulus: 4
```
simonpasquier, line 211:
```suggestion
  modulus: 4 # number of shards
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+  target_label: '__tmp_hash'
+  replacement: '$1'
+  action: 'hashmod'
+- source_labels: ['__tmp_hash']
+  separator: ;
+  regex: '2'
```
simonpasquier, line 217:
```suggestion
  regex: '2' # index of the current shard
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+  separator: ;
+  regex: '{{ .operator.shardIndex | div $shardsPerZone }}'
+  replacement: '$1'
+  action: 'keep'
+```
+
+we would get the following output for `shard_index == 2`:
+
+```yaml
+- source_labels: [ '__meta_kubernetes_pod_label_topology_kubernetes_io_zone' ]
```
simonpasquier, line 268: if I'm not wrong the `topology.kubernetes.io/zone` label is only set on [nodes](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone), not pods. I suspect that we'll need to attach node metadata to the service discovery and filter on `__meta_kubernetes_node_label_topology_kubernetes_io_zone`.

As a side note, endpointslices can carry topology information so in this case, node metadata might not be required (https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#topology).

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+
+```yaml
+spec:
+  nodeSelector:
+    'topology.kubernetes.io/zone': 'europe-west4-a'
+```
+
+## Alternatives
+
+N/A
```
simonpasquier, line 331: An alternative could be to expose a way for users to customize the shard relabelings and let them manage all the settings manually (nodeSelector + relabelings) but it would be way more less user-friendly.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+Values generated by the operator, like `shard_index` are retrieved from an
+`.operator` object.
+
+### classic mode
+
+Given the following configuration
+
+```yaml
+spec:
+  shards: 4
```
simonpasquier, line 181: (nit) let's add the replica count to be explicit and assume the standard HA-pair mode.
```suggestion
  replicas: 2
  shards: 4
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+  action: 'keep'
+```
+
+### topology mode
+
+Given the following configuration
+
+```yaml
+spec:
+  shards: 4
```
simonpasquier, line 228: (nit)
```suggestion
  replicas: 2
  shards: 4
```

simonpasquier:
Ah, another point to take into consideration would be how/if the design could work with the (experimental) daemonset agent mode (for context, https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202405-agent-daemonset.md).

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+
+* Implement mechanisms to automatically fix configuration errors by the user
+* Support mixed environments (kubernetes and non-kubernetes targets are scraped)
+
+## Algorithm
+
+In order to do calculate a stable assignment, following parameters are required:
+
+1. `num_shards`: The number of prometheus shards
+2. `shard_index`: A number of the range `[0..num_shards[` identifying a single
```
nicolastakashi, line 47:
```suggestion
2. `shard_index`: A number of the range `[0..num_shards]` identifying a single
```

simonpasquier: I think that the current wording is correct, we want `[0..num_shards-1]`.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,335 @@
...
+        sourceLabel: '__address__'
+      
+    # The following section is only valid if "mode" is set to "Topology"
+    # 'topology.kubernetes.io/zone' for 'topology'
+    topology: 
+        # A kubernetes node label defining the topology to be sharded on
+        nodeLabel: 'topology.kubernetes.io/zone'
+        
+        # A prometheus metric containing the topology value of a given target
+        sourceLabel: '__meta_kubernetes_pod_label_topology_kubernetes_io_zone'
```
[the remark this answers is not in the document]
arnecls: This allows for more flexibility and is also more future proof.
Maybe you want to scrape based on region, or some other concept. Maybe the source label is generated.
We cannot anticipate what this will look like, but we can allow overrides - and this is what is done here.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: That‘s probably a good idea.
Helm templates are a mess to read and the algorithm is explained above, so the comments are likely shorter and easier to read.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: I guess not everybody knows the `[a..b[` notation. I'll use the `b-1]` instead. That should be clear for everybody.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: > I suspect that we'll need to attach node metadata to the service discovery

You're right! I totally forgot about this. In the original issue I had this addressed by having the label injected into a pod ... somehow.

How can we do this through prometheus?
Is this even possible?

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: Essentially this is coming from the comment below, that pods do not have a zone label by default. If this label is injected by the user, we don't know the name.

However, it depends a bit on if and how we can solve the issue of the missing label on pods in general.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: added

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: done

simonpasquier:
> IMO these two proposals are mutually exclusive as daemonSets only scrape their node and hence have a fixed sharding configuration. So - all "shardingStratgies" would become ineffective if DaemonSet mode is active.

Yes, I agree but we need to clarify this in the doc :)

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,311 @@
...
+
+```yaml
+spec:
+  shardingStrategy:
+     # Select a sharding mode. Can be 'Classic' or 'Topology'
+    mode: 'Classic'    
+
+    # Optional: array of relabel configurations to insert before the sharding
+    # relabel configuration.
+    additionalRelabelConfig: []
```
simonpasquier, line 141: Scrape classes can also be leveraged to inject relabel configs. While it's more explicit to have a dedicated field here, it duplicates the same functionality. I'd rather avoid adding a new field for now and document the possibility to use scrape classes.

Also we could already provide more flexibility to users by letting them set the __tmp_hash label before the hashmod rule.E.g.  instead of having this snippet at the end of the generated relabel configs:

````yaml
relabel_configs:
  - ...
    separator: ;
    modulus: 2
    target_label: __tmp_hash
    replacement: $1
    action: hashmod
  - source_labels: [__tmp_hash]
    separator: ;
    regex: "0"
    replacement: $1
    action: keep
````

It could be

```yaml
relabel_configs:
  - ...
  # If the user has set __tmp_hash before then use this value otherwise default to __address__.
  - source_labels: [__address__, __tmp_hash]
    target_label: __tmp_hash
    regex: "(.+);(.+)"
    replacement: $1
    action: replace
  - source_labels: [__tmp_hash]
    separator: ;
    modulus: 2
    target_label: __tmp_hash
    replacement: $1
    action: hashmod
  - source_labels: [__tmp_hash]
    separator: ;
    regex: "0"
    replacement: $1
    action: keep
````

File: Documentation/proposals/202411-zone-aware-sharding.md (hunk 15 above)
Comment (arnecls) [reply Documentation/proposals/202411-zone-aware-sharding.md]: Ok, seems like this is going to get complicated.
According to the docs, there is no label available we could use here:

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config

Or am I mistaken?

Comment (arnecls) [reply Documentation/proposals/202411-zone-aware-sharding.md]: > I suspect that we'll need to attach node metadata to the service discovery

This means a code change to the `kubernetes_sd` ?

File: Documentation/proposals/202411-zone-aware-sharding.md [hunk 22]
```diff
@@ -0,0 +1,311 @@
...
+    mode: 'Classic'    
+
+    # Optional: array of relabel configurations to insert before the sharding
+    # relabel configuration.
+    additionalRelabelConfig: [] 
+
+    # The following section is only valid if "mode" is set to "Classic"
+    classic:
+        # Metric label used for sharding.
+        sourceLabel: '__address__'
```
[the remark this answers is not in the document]
arnecls: I can't think of any ...?

But aside from this, the field _is_ an array, and if we later realize that we _do_ have a case, making this field an array will be a breaking change. What do you think? Allow an array to avoid a potential future breaking change, or keep it simple?

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: I removed the field and added a note on scrape classes in the "alternatives discussed" section.

simonpasquier: > If scrape classes are always injected before this block and essentially provide the same functionality as additionalRelabelConfig we can remove that field.

the API [documentation](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.ScrapeClass) says:

> The Operator automatically adds relabelings for a few standard Kubernetes fields like __meta_kubernetes_namespace and __meta_kubernetes_service_name. Then the Operator adds the scrape class relabelings defined here. Then the Operator adds the target-specific relabelings defined in the scrape object.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
simonpasquier: I think that the discussion in https://github.com/prometheus-operator/prometheus-operator/pull/7110#discussion_r1843548809 voids the need for a `sourceLabel` field.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: As the field is already gone -> resolved

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,295 @@
...
+Given the following configuration
+
+```yaml
+spec:
+  shards: 4
+  replicas: 2
+  shardingStrategy:
+    mode: 'Classic'    
+    classic:
+      sourceLabel: '__address__'
```
nicolastakashi, line 186: could this be optional?

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,295 @@
+# Zone aware sharding for prometheus
```
nicolastakashi, line 1: @simonpasquier do you think makes sense we highlght that additionalScrapeConfig will not automatically work?

nicolastakashi, line 1: @simonpasquier should we highlight about additionalScrapeConfigs not working out of the box with this solution?

simonpasquier: yes I'd probably put in in the non-goals section.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
simonpasquier: > However, it depends a bit on if and how we can solve the issue of the missing label on pods in general.

Now I remember that I had an idea which avoids the need to attach node metadata to the service discovery. As explained in https://gist.github.com/simonpasquier/0a7df5b0899721167b5c56f9cdc9668c (which was mentioned in https://github.com/prometheus-operator/prometheus-operator/issues/5495), the operator could discover the list of nodes + AZs and inject a relabel config rule which would filter the pods based on their `__meta_kubernetes_pod_node_name` label.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,295 @@
...
+0 0 0 | assignment index
+```
+
+In this case the 1st assert will warn about zone C not being scraped.
+
+This second case - a zone not being scraped - should lead to an error in the
+operator, causing the change to not be rolled out, as data would be lost.
+The first case - double scraping - should at minimum cause a warning.
+
+## API changes
```
simonpasquier, line 125: I wonder if we need a feature flag for this feature? I'm not sure that we can cram everything in a single PR so having it behind a feature flag would allow for a more gradual development.

@nicolastakashi thoughts?

Documentation/proposals/202411-zone-aware-sharding.md
simonpasquier: https://github.com/prometheus-operator/prometheus-operator/pull/7126, let's see if it breaks the e2e tests (it shouldn't).

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: Yes. I think I put somewhere that `__address__` is just the default value here.
I'll make this more explicit in the API changes section,

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: @simonpasquier If this requires a change to the `kubernetes_sd` this needs to be mentioned in the doc, right?

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: @simonpasquier @nicolastakashi I added a new section to discuss this topic to the document.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: I added a new section to the document to discuss this.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: I disagree (a bit).
The fact that `__address__` is currently hardcoded is not ideal. Allowing users to change this allows for more flexibility and is a (comparable) small change.
As we're currently discussing a change to this section I think it is a good point to add this here, especially as it does not break any existing functionality.

If we had this field now, we could already test a few things in regard to sharding, which we currently cannot.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
simonpasquier: > I'm afraid this method will cause issues, as the operator would have to constantly update the relabel config. Especially in clusters where nodes scale up and down a lot, this would put quite a bit of load on prometheus and the operator.

I'm not too worried about the additional load unless the node churn is very high. Attaching node metadata to all targets has also a cost for Prometheus and the Kubernetes API (more resources being watched). I still think that it would make sense as an option for clusters which don't expect a high churn on nodes.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
simonpasquier: Have you seen that https://github.com/prometheus-operator/prometheus-operator/pull/7126 has merged? The next release of the operator will allow to customize the label(s) on which the hashmod operation is performed.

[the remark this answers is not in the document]
simonpasquier: To clarify why I'm relunctant to add new fields: any field that we add we need to live with it until a v2 API happens (if ever) which creates cognitive load for users and maintainers. Not adding it now doesn't mean that it's a permanent no.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: But then you still have the race condition, don't you?

BTW - I just saw that prometheus 3 has `__meta_kubernetes_endpointslice_endpoint_zone` which means we have that information for serviceMonitors, right?

Documentation/proposals/202411-zone-aware-sharding.md
simonpasquier: @arnecls can you add that the feature will be behind a specific feature gate initially (https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/operator/feature_gates.go for context)?

simonpasquier (commented):
Thanks again for working hard on this! It's really promising.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: I personally prefer explicit over implicit, but I get your point.
I'll remove the sourceLabel from both sections and will reference the `__tmp_hash` override.

In case of the `Topology` mode, this is probably a good idea anyways, as the label might be different for pods or services.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,321 @@
...
+0 1   | shard index
+A B C | zone
+0 0 0 | assignment index
+```
+
+In this case the 1st assert will warn about zone C not being scraped.
+
+This second case - a zone not being scraped - should lead to an error in the
+operator, causing the change to not be rolled out, as data would be lost.
+The first case - double scraping - should at minimum cause a warning.
```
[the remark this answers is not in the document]
arnecls: Done

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,321 @@
...
+2. The operator could do this discovery and add a relabel rule based on the
+   node name.
+
+The second solution would require the operator to constantly update the relabel
+configuration. This could lead to increased load on clusters with agressive
+autoscaling as well as race conditions for pods on newly created nodes, as the
+config change is not atomic/instant.
+
+As of that, a change to the kubernetes service discovery is considered the more
+stable, and thus preferrable solution.
```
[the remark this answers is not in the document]
arnecls: Added.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,321 @@
...
+        # Defaults to `__meta_kubernetes_node_label_topology_kubernetes_io_zone`
+        sourceLabel: '__meta_kubernetes_node_label_topology_kubernetes_io_zone'
+
+        # All topology values to be used by nodeLabel and sourceLabel
+        values: []
+```
+
+The `additionalRelabelConfig` section is meant to allow the `sourceLabel` to be
+generated if needed. This should allow enough flexibility to cover edgecases
+not anticiapted by this proposal.
```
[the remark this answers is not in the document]
arnecls: yes. Removed

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: Added.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,321 @@
...
+* Define the relabel configuration to be generated for zone aware sharding
+* Define changes to the prometheus pod spec to support zone stickyness
+* Stay backwards compatible to the current mechanism by default
+
+## Non-goals
+
+* Implement mechanisms to automatically fix configuration errors by the user
+* Support mixed environments (kubernetes and non-kubernetes targets are scraped)
+
+## Algorithm
```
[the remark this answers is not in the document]
arnecls: This would make it a sub-section of non-goals?

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,321 @@
...
+* **Other docs:**
+  * [Well known kubernetes labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone)
+  * [AWS zone names](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones)
+  * [GCP zone names](https://cloud.google.com/compute/docs/regions-zones#available)
+
+This proposal describes how we can implement zone-aware sharding by adding
+support for custom labels and zone configuration options to the existing
+prometheus configuration resources.
+
+> [!NOTE]
```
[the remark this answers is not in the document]
arnecls: done

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Prometheus instance zone assignment
+
+To make sure that prometheus instances are deployed to the correct zone (of their
+assigned target), we need to generate a [node affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity)
+or a [node selector](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes/#create-a-pod-that-gets-scheduled-to-your-chosen-node).
+
+As node selectors are simpler to manage, and node affinities might run into
+ordering issues when a user defines their own affinities, node selectors should
+be used.
```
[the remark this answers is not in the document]
arnecls: done

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,321 @@
...
+      nodeLabel: 'topology.kubernetes.io/zone'
+      sourceLabel: '__meta_kubernetes_node_label_topology_kubernetes_io_zone'
+      values:
+        - europe-west4-a
+        - europe-west4-b
+```
+
+we would get the following output for `shard_index == 2`:
+
+```yaml
```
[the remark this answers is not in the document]
arnecls: Sure, but I currently cannot see/remember where this comes from 😅

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: Mentioned.

Documentation/proposals/202411-zone-aware-sharding.md
arnecls: Can you add a suggestion on what to put there?

arnecls: added

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: done

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,342 @@
...
+The `topology` section does not use the term `zone`. This makes the feature
+more flexible in case a user needs to shard on e.g. regions instead.
+
+Both modes do not contain an explicit overwrite of the label used for sharding.
+This feature is already possible by generating a `__tmp_hash` label through
+[scrape classes](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.ScrapeClass).
+
+In the `Classic` mode this allows overriding the main label used for sharding.
+In case of the `Topology` mode, a different label might need to be used for
+e.g. services or pods. This label is automatically generated.
```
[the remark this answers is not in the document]
arnecls: The zone aware sharding relabel config has two parts

1. Filter out targets that are not in the same zone
2. Hash based sharding over the targets inside the desired zone

The second part is essentially the same thing as the classic mode and needs to implement the same mechanics.

The first part is added for topology aware scraping.
From the ongoing discussion, it is not yet clear which label we can use.
There already is a label for endpointslices, but endpointslices are only available for services, not for pods. So for pods we might have to use a different label.
And this means - we cannot have _one_ override, but maybe have two or more.

While writing this I just realized that we could simply have a list of source fields. D'oh.

Okay - but this still leaves the question on how to handle the sourceLabel for topology here. For classic mode it's implicit, i.e. the user needs to provide a label that we use. How do we do this here? We cannot use the same label, as it's already in use. We could add a sourceLabel field (array) again, but then we have two different methods. Also - what's the default value for this field?

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,342 @@
...
+    mode: 'Classic'    
+
+    # The following section is only valid if "mode" is set to "Topology"
+    # 'topology.kubernetes.io/zone' for 'topology'
+    topology: 
+        # A kubernetes node label used in the nodeSelector field of prometheus
+        # instances. This assures that instances are placed in the zone they
+        # are meant to be scraping.
+        # Defaults to `topology.kubernetes.io/zone`
+        nodeLabel: 'topology.kubernetes.io/zone'
```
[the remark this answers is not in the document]
arnecls: Older clusters might use the legacy label `failure-domain.beta.kubernetes.io/zone` but I don't know if this is still in use anywhere or if prometheus-operator even runs on versions that still have it (It was removed after 1.26).

Seems like the label is injected by the kubelet so we can rely on it after 1.26.
https://kubernetes.io/docs/reference/node/node-labels/

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,342 @@
...
+        - 'europe-west4-a'
+        - 'europe-west4-b'
+```
+
+we would get the following output for `shard_index == 2`:
+
+```yaml
+# zones := shardingStrategy.topology.values
+# shards_per_zone := max(1, floor(shards / len(zones)))
+- source_labels: ['__meta_kubernetes_endpointslice_endpoint_zone']
```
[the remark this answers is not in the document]
arnecls: I already mentioned this above, but we could actually add both, the `__meta_kubernetes_endpointslice_endpoint_zone` and the node metadata field, couldn't we?

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,342 @@
...
+```
+
+In this case the 1st assert will warn about zone C not being scraped.
+
+This second case - a zone not being scraped - should lead to an error during
+reconciliation, causing the change to not be rolled out, as data would be lost.
+While the first case - double scraping - is not as severe, it should cause an
+error during reconciliation, too, as this would be otherwise hard to spot.
+It's also to be mentioned that replicas are to be used to achieve redundant
+scraping.
```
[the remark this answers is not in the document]
arnecls: That's simpler. Changed.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,342 @@
...
+  * [GCP zone names](https://cloud.google.com/compute/docs/regions-zones#available)
+
+This proposal describes how we can implement zone-aware sharding by adding
+support for custom labels and zone configuration options to the existing
+prometheus configuration resources.
+
+> [!NOTE]
+> Due to the size of this feature, it will be placed behind a
+> [feature gate](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/operator/feature_gates.go)
+> to allow incremental testing.
```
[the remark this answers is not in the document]
arnecls: done

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
simonpasquier: yes the operator can pick up the right label based on the following decision tree:
* Use `__meta_kubernetes_endpointslice_endpoint_zone` if the resource is a service monitor + `.spec.serviceDiscoveryRole: EndpointSlice`
* Otherwise use `__meta_kubernetes_node_label_topology_kubernetes_io_zone` + attach node metadata.

And if we want to offer flexibility for users, we can take the same approach as with the `__tmp_hash` label and have the `keep` rule checking against a specific temporary label:

```
- source_labels: ['__meta_kubernetes_endpointslice_endpoint_zone', __tmp_topology]
  target_label: __tmp_topology
  regex: "(.+);"
- source_labels: ['__meta_kubernetes_node_label_topology_kubernetes_io_zone', '__meta_kubernetes_node_labelpresent_topology_kubernetes_io_zone','__tmp_topology']
  regex: "(.+);true;"
  target_label: __tmp_topology
- source_labels: ['__tmp_topology']
  regex: 'europe-west4-a'          # zones[shard_index % shards_per_zone]
  action: 'keep'
```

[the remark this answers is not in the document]
arnecls: I added this to the example and improved the overwrite mechanics explanation.
@simonpasquier is this now better/clearer?

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: I removed the node label field and added the non-goal.
I also added the outcome of this discussion (and the one of the label overwrites) to the alternatives section so the reasoning doesn't get lost.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: Is this what you meant?

```yaml
spec:
  scrapeClasses:
    - name: 'topology'
      default: true
      attachMetadata:
        node: true
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,345 @@
...
+```yaml
+spec:
+  shardingStrategy:
+     # Select a sharding mode. Can be 'Classic' or 'Topology'.
+     # Defaults to `Classic`.
+    mode: 'Classic'    
+
+    # The following section is only valid if "mode" is set to "Topology"
+    # 'topology.kubernetes.io/zone' for 'topology'
+    topology:
```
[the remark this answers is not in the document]
arnecls: Isn't this field redundant to the labels used for topology detection?
I mean `__meta_kubernetes_endpointslice_endpoint_zone` and `__meta_kubernetes_node_label_topology_kubernetes_io_zone`?

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: Continued in a [discussion below](https://github.com/prometheus-operator/prometheus-operator/pull/7110#discussion_r1856872505)

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
simonpasquier: None of the __meta labels will be copied to the metric labels by default. The general recommendation is to use external labels for labels that apply to all metrics of a Prometheus instance.

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,357 @@
...
+
+When running large, multi-zone clusters, prometheus scraping can lead to an
+increase in inter-zone traffic costs. The current sharding mechanics will
+allow multiple instances of prometheus to run, but as the `__address__` label
+is hard coded, all instances will always fetch all zones.
+
+It is not sufficient to simply switch this label with another label, as
+multiple prometheus instances per zone might be required. Furthermore we
+must be able to calculate the "assignment" of a specific instance to a zone,
+and that assignment must be stable.
```
simonpasquier, line 28: maybe rephrase it since the next version of the operator enables power users to use something other than the default `__adddress__` label?

```suggestion
When running large, multi-zone clusters, Prometheus scraping can lead to an
increase in inter-zone traffic costs. A solution would be to deploy 1 Prometheus shard
per zone and configure each shard to scrape only the targets local to its zone. The 
current sharding implementation can't solve the issue though. While
it's possible to customize the label (`__address__` by default) used for distributing the
targets to the Prometheus instances, there's no way to configure a single Prometheus
resource so that each shard is bound to a specific zone.
```

Documentation/proposals/202411-zone-aware-sharding.md
```diff
@@ -0,0 +1,357 @@
...
+  shardingStrategy:
+     # Select a sharding mode. Can be 'Classic' or 'Topology'.
+     # Defaults to `Classic`.
+    mode: 'Classic'    
+
+    # The following section is only valid if "mode" is set to "Topology"
+    topology:
+        # All topology values to be used by the cluster, i.e. a list of all
+        # zones in use.
+        values: []
```
simonpasquier, line 182: saying it here to remember later but it doesn't have to be in the proposal: if the mode is Topology then the values list should have at least 1 item.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
simonpasquier: No I meant to add it to the generated Prometheus configuration but don't add it now (it will be obvious when implementing).

````
- job_name: serviceMonitor/foo/bar/0
  honor_labels: false
  kubernetes_sd_configs:                                                                       
  - role: endpoints                                                                            
    attach_metadata: <--- here
      node: true
    namespaces:    
      names:       
      - foo                                                                                
    relabel_configs:
    - ...

````

**Review by simonpasquier — COMMENTED** [review]
This is excellent! These are my last remarks otherwise it's a LGTM for me.

File: Documentation/proposals/202411-zone-aware-sharding.md (hunk 38 above)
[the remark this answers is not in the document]
Comment (simonpasquier) [reply Documentation/proposals/202411-zone-aware-sharding.md]: this is still relevant IMHO.

File: Documentation/proposals/202411-zone-aware-sharding.md [hunk 41]
```diff
@@ -0,0 +1,357 @@
...
+* **Owners:**
+  * [arnecls](https://github.com/arnecls)
+
+* **Related Tickets:**
+  * [#6437](https://github.com/prometheus-operator/prometheus-operator/issues/6437)
+
+* **Other docs:**
+  * [Well known kubernetes labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone)
+  * [AWS zone names](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones)
+  * [GCP zone names](https://cloud.google.com/compute/docs/regions-zones#available)
```
simonpasquier, line 12: It would be good to mention the shard autoscaling proposal. I don't think that it has a significant impact on this proposal but it should mentioned as a generic item in the goals (e.g. `Compatibility with the Shard autoscaling functionality`).
```suggestion
  * [GCP zone names](https://cloud.google.com/compute/docs/regions-zones#available)
  * [Shard Autoscaling](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md) design proposal.
```

simonpasquier: cc @nicolastakashi

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: If the label is not there, then it is certainly required, especially for debugging.
I'll add it.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: I added this as a note so it is documented.

Documentation/proposals/202411-zone-aware-sharding.md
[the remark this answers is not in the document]
arnecls: done

simonpasquier (approved):
Excellent! That's an approval from my side. I'll let @nicolastakashi review it one more time.
If you're willing to contribute the feature, feel free to start the work :)

simonpasquier:
(nit) the doc linter asks for a few cosmetic changes.

nicolastakashi (approved):

nicolastakashi:
Amazing work @arnecls 
Looking forward to see this in actions 💪🏽

simonpasquier:
:partying_face: 

@arnecls thanks again for the tremendous effort!

PR #7127, opened 2024-11-19
Document disableCompaction defaulting when using thanos sidecar

VannTen:
## Description

For the Prometheus CRD, the operator default to true for `spec.disableCompaction` when `spec.thanos.objectStorageConfig(|File)` is defined, but this is not documented anywhere but in code (and in kube-prometheus-stack helm chart values, as it turns out).

https://github.com/prometheus-operator/prometheus-operator/blob/35790cdbeec2f302516781365313e7d6fc007ac8/pkg/prometheus/server/statefulset.go#L753-L761

Document the behavior in the API reference directly.

## Type of change

## Verification
N/A

## Changelog entry

```release-note
Document default behavior of `spec.disableCompaction` in Prometheus CRD in relation to Thanos sidecar object store configuration.
```

### DIFF
--- a/pkg/apis/monitoring/v1/prometheus_types.go
+++ b/pkg/apis/monitoring/v1/prometheus_types.go
@@ -992,6 +992,7 @@ type PrometheusSpec struct {
 	RetentionSize ByteSize `json:"retentionSize,omitempty"`
 
 	// When true, the Prometheus compaction is disabled.
+	// Defaults to true if `spec.thanos.objectStorageConfig` or `spec.objectStorageConfigFile` are defined.
 	DisableCompaction bool `json:"disableCompaction,omitempty"`
 
 	// Defines the configuration of the Prometheus rules' engine.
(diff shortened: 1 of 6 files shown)


pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -957,6 +957,7 @@ type PrometheusSpec struct {
 	RetentionSize ByteSize `json:"retentionSize,omitempty"`
 
 	// When true, the Prometheus compaction is disabled.
+	// Defaults to true if `spec.thanos.objectStorageConfig` or `spec.objectStorageConfigFile` are defined.
```
simonpasquier, line 960: (suggestion) let's explain the rationale.
```suggestion
	// When `spec.thanos.objectStorageConfig` or `spec.objectStorageConfigFile` are defined, the operator automatically disables block compaction to avoid race conditions during block uploads (as the Thanos documentation recommends).
```

simonpasquier (commented):
Thanks for clarifying the doc! I've got only one small suggestion.

VannTen:
Not sure what the test results are about...
Unless I should rebase on main ?

simonpasquier:
@VannTen the failures are unrelated to your PR but we had 2 recent merges that collided.

simonpasquier:
you'd need to rebase on top of main once https://github.com/prometheus-operator/prometheus-operator/pull/7138 merges.

VannTen:
will do :+1:

Issue #7077, opened 2024-11-06
Allow to configure `fallback_scrape_protocol` for each scrape CRD and globally


simonpasquier:
### Component(s)

Prometheus, PrometheusAgent

### What is missing? Please describe.

https://github.com/prometheus/prometheus/pull/15136 introduces a (breaking) change in Prometheus 3.0: Prometheus will reject targets not specifying an explicit and valid `Content-Type` (typically applications not using official/supported client libraries). As a workaround, users can define a fallback protocol. It needs to be supported for ServiceMonitor, PodMonitor and ScrapeConfig. It would also be useful to support as a global Prometheus/PrometheusAgent setting.

## Environment
Kubernetes Version:
Prometheus-Operator Version:

dongjiang1989:
@simonpasquier I would like to work on this issue. 😊

slashpai:
@dongjiang1989 go ahead I didnt get enough time to work on it :)

patrickshan:
For anyone who is looking for solution for this with prometheus-operator, a default `scrapeClass` can be defined with `fallbackScrapeProtocol` parameter which will be applied to all the scrape config.

hack4e:
> For anyone who is looking for solution for this with prometheus-operator, a default `scrapeClass` can be defined with `fallbackScrapeProtocol` parameter which will be applied to all the scrape config.

@patrickshan Thanks, I've managed to resolve the issue by updating the Operator configuration using the configuration shown below.


```yaml
    scrapeClasses:
    - default: true
      fallbackScrapeProtocol: PrometheusText0.0.4
      name: default-scrape
```

PR #7197, opened 2024-12-16
Fix: Remove scrapeFallbackProtocol in Prometheus CommonSpec

### LINKED ISSUE #7196: scrapeFallbackProtocol in PrometheusSpec is broken
### Is there an existing issue for this?

- [X] I have searched the existing issues

## Description

The `scrapeFallbackProtocol` field in `PrometheusSpec` is invalid, this is not a global configuration parameter as per https://prometheus.io/docs/prometheus/latest/configuration/configuration/

It's only valid on `scrape_config`.

This might be better as part of the `ScrapeClass` spec, then you can create a default scrape class that sets this value everywhere!

Alternatively the current field needs to affect every generated scrape job rather than go into the global config

## Steps to Reproduce

Set `spec.scrapeFallbackProtocol` to anything in a Prometheus resource.

## Expected Result

Valid Prometheus config yaml is created

## Actual Result

Prometheus (v3.0.1) reports an error reloading configuration

```
time=2024-12-16T01:29:12.923Z level=ERROR source=main.go:1162 msg="Error reloading config" err="couldn't load configuration (--config.file=\"/etc/prometheus/config_out/prometheus.env.yaml\"): parsing YAML file /etc/prometheus/config_out/prometheus.env.yaml: yaml: unmarshal errors:\n  line 3: field fallback_scrape_protocol not found in type config.plain"
```

### Prometheus Operator Version

```shell
0.79.0
```

### Kubernetes Version

```yaml
1.31.2
```

### Kubernetes Cluster Type

EKS

### How did you deploy Prometheus-Operator?

helm chart:prometheus-community/kube-prometheus-stack

### Manifests

_No response_

### prometheus-operator log output

```shell
N/A
```

### Anything else?

_No response_

dongjiang1989:
## Description

Close #7196 
Remove  scrapeFallbackProtocol in Prometheus CommonSpec broken

Cherry pick to release-0.79

## Type of change

## Verification

## Changelog entry

```release-note
Remove scrapeFallbackProtocol in Prometheus CommonSpec
```


slashpai (approved):

slashpai:
Not a blocker, I was wondering if we should fix the API too with the name `FallbackScrapeProtocol` instead of `ScrapeFallbackProtocol` so that the json field matches similar to Prometheus config

```
# Fallback protocol to use if a scrape returns blank, unparseable, or otherwise
# invalid Content-Type.
# Supported values (case sensitive): PrometheusProto, OpenMetricsText0.0.1,
# OpenMetricsText1.0.0, PrometheusText0.0.4, PrometheusText1.0.0.
[ fallback_scrape_protocol: [<string>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#string) ]
```
Ref: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config

Lets wait what @simonpasquier thinks about that

simonpasquier:
> I was wondering if we should fix the API too with the name FallbackScrapeProtocol instead of ScrapeFallbackProtocol so that the json field matches similar to Prometheus config

Given that the feature is broken anyway, we could make an exception here and rename the field. Though I like that the field is prefixed by `scrape` like `scrapeInterval`, `scrapeTimeout`, ...

slashpai:
> > I was wondering if we should fix the API too with the name FallbackScrapeProtocol instead of ScrapeFallbackProtocol so that the json field matches similar to Prometheus config
> 
> 
> 
> Given that the feature is broken anyway, we could make an exception here and rename the field. Though I like that the field is prefixed by `scrape` like `scrapeInterval`, `scrapeTimeout`, ...

I don't have strong opinion but do we need to try to keep field names same as in Prometheus?

simonpasquier:
@dongjiang1989 sorry for the back and forth but after more thinking, it seems more appropriate to remove the field for now (as you did initially) and wait for a concrete use case requiring a global option.

pkg/prometheus/promcfg_test.go
```diff
@@ -265,22 +264,6 @@ func TestGlobalSettings(t *testing.T) {
 			},
 			Golden: "valid_global_config_with_unsupported_scrape_protocols.golden",
 		},
-		{
-			Scenario:               "valid global config with scrape fallback protocol",
-			Version:                "v3.0.0-rc.0",
-			ScrapeInterval:         "30s",
-			EvaluationInterval:     "30s",
-			ScrapeFallbackProtocol: ptr.To(monitoringv1.PrometheusText1_0_0),
-			Golden:                 "valid_global_config_with_scrape_fallback_protocol.golden",
```
simonpasquier, line 274: can we remove the golden file?

pkg/prometheus/promcfg_test.go
```diff
@@ -265,22 +264,6 @@ func TestGlobalSettings(t *testing.T) {
...
-		},
-		{
-			Scenario:               "valid global config scrape protocols with unsupported version",
-			Version:                "v2.55.0",
-			ScrapeInterval:         "30s",
-			EvaluationInterval:     "30s",
-			ScrapeFallbackProtocol: ptr.To(monitoringv1.PrometheusProto),
-			Golden:                 "valid_global_config_with_unsupported_scrape_fallback_protocols.golden",
```
simonpasquier, line 282: same here

dongjiang1989: Done.

dongjiang1989:
> @dongjiang1989 sorry for the back and forth but after more thinking, it seems more appropriate to remove the field for now (as you did initially) and wait for a concrete use case requiring a global option.

Thanks @slashpai @simonpasquier 
I Got it.
Please re-check it

slashpai (approved):

simonpasquier (approved):
thanks!

c3-davidtran:
> Reading again #7077
> 
> > It would also be useful to support as a global Prometheus/PrometheusAgent setting.
> 
> I missed that we have no global setting in the Prometheus config but the intention here was to provide a global setting at the Prometheus/PrometheusAgent level to avoid users changing all their scrape resources. It should work the following way:
> 
> * if the Prometheus resource defines no fallback protocol then use whatever is defined by the scrape resource.
> * if the Prometheus resource defines a fallback protocol and the scrape resource defines none then use the former.
> * if the Prometheus resource defines a fallback protocol and the scrape resource defines its own then use the latter.

@simonpasquier  I see that even in latest operaor version 0.81.0, this behavior to use a global fallback config you've suggested, is not yet implemented.
Would you know if there is an effort to implement it or is it already planned for later releases ?

Issue #7221, opened 2024-12-19
Add the PrometheusTopologySharding feature gate.


simonpasquier:
### Component(s)

Prometheus, PrometheusAgent

### What is missing? Please describe.

See https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202411-zone-aware-sharding.md#action-plan

## Environment
Kubernetes Version:
Prometheus-Operator Version:

dongjiang1989:
@simonpasquier  I want to try this job.

Issue #7222, opened 2024-12-19
[Zone aware sharding] Implement the API changes with pre-flight validations


simonpasquier:
### Component(s)

Prometheus, PrometheusAgent

### What is missing? Please describe.

See https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202411-zone-aware-sharding.md#action-plan

## Environment
Kubernetes Version:
Prometheus-Operator Version:

PR #7233, opened 2024-12-23
feat: add PrometheusTopologySharding feature gate

dongjiang1989:
## Description

ref #7221 

## Type of change

## Verification

## Changelog entry

```release-note
add the PrometheusTopologySharding feature gate
```


simonpasquier (commented):
Thanks for starting the PR! The linked issue is only about adding the feature gate to the operator, nothing more. The API changes are tracked in https://github.com/prometheus-operator/prometheus-operator/issues/7222

dongjiang1989:
> Thanks for starting the PR! The linked issue is only about adding the feature gate to the operator, nothing more. The API changes are tracked in #7222

Thanks @simonpasquier 
Got it
Move apis struct to other PR

pkg/prometheus/server/operator.go
```diff
@@ -92,6 +92,8 @@ type Operator struct {
 	canReadStorageClass           bool
 	disableUnmanagedConfiguration bool
 
+	prometheusTopologyShardingFeatureGateEnabled bool
```
simonpasquier, line 95: Instead of adding a new field for each feature gate, we could extend `prompkg.Config` to hold a copy of Gates. When the controller needs to check if a gate is enabled, it would call `c.Config.Gates.Enabled(...)`.
It would also have the nice benefit that the operator always reconciles the workload resources when gates are enabled/disabled.

simonpasquier (commented):
Thanks! The agent controller also needs to have access to the feature gate but my other comment should resolve this.

pkg/prometheus/server/operator.go
dongjiang1989: Thanks @simonpasquier 
Got it
Remove flag done.

simonpasquier (approved):
thanks!

PR #7274, opened 2025-01-15
Add ShardRetentionPolicy to Prometheus API

mviswanathsai:
## Description

Picking up from #6198 (all credits to @ArthurSens). I just stepped in since he's a bit busy!

Partially addresses https://github.com/prometheus-operator/prometheus-operator/issues/4967, by implementing half of what was proposed [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#graceful-scale-down-of-prometheus-servers).

When Scaling down Prometheus servers, there are good chances of data loss because the Operator is not aware if the data stored in a Prometheus server is being backed up by a Long-Term-Storage solution. Users who want to avoid such situations might want to tell the operator to keep the instance alive for a while before safe deletion.

This PR introduces the new field ShardRetentionPolicy to the Prometheus CRD to address the problem mentioned above and in the linked proposal.

For the sake of keeping PRs small, in this PR I'm only worrying about pod retention when a retention period is set. Follow-up PRs will be opened with the goals:

Scale up will re-use already existing pod that was marked for future deletion
Scale up/down will correctly redistribute targets.
Retention duration is not taken into consideration right now. If ShardRetentionPolicy is set to Retain, we'll keep the shard alive forever.

## Type of change

## Verification

## Changelog entry

```release-note

```


mviswanathsai:
CC: @simonpasquier @nicolastakashi @slashpai

pkg/prometheus/server/operator.go
```diff
@@ -951,6 +962,21 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
 
+// As the ShardRetentionPolicy feature evolves, should delete will evolve accordingly.
+// For now, shouldDelete just returns the appropriate boolean based on the retention type.
+func (c *Operator) shouldDelete(p *monitoringv1.Prometheus) (bool, error) {
+	if !c.retentionPoliciesEnabled {
+		// Feature-gate is disabled, default behavior is always to delete.
+		return true, nil
+	}
+
+	if p.Spec.ShardRetentionPolicy.WhenScaled == monitoringv1.WhenScaledRetentionTypeDelete {
```
simonpasquier, line 973: we need to test for nil pointers.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1115,6 +1127,23 @@ type PrometheusSpec struct {
...
+type WhenScaledRetentionType string
+
+var (
+	WhenScaledRetentionTypeRetain WhenScaledRetentionType = "Retain"
+	WhenScaledRetentionTypeDelete WhenScaledRetentionType = "Delete"
+)
+
+type ShardRetentionPolicy struct {
+	// WhenScaled defines the retention policy when the Prometheus shards are scaled down.
+	// When set to `Retain`, the operator will keep the pods from the scaled-down shard(s), so the data can still be queried.
```
simonpasquier, line 1139:
```suggestion
	// * `Delete`, the operator will delete the pods from the scaled-down shard(s).
	// * `Retain`, the operator will keep the pods from the scaled-down shard(s).
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1115,6 +1127,23 @@ type PrometheusSpec struct {
...
+	//
+	// Default: "Delete"
```
simonpasquier, line 1141:
```suggestion
	// If not defined, the operator assumes the `` Delete` value.
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1115,6 +1127,23 @@ type PrometheusSpec struct {
...
+	// +kubebuilder:validation:Enum=Retain;Delete
+	// +required
+	WhenScaled WhenScaledRetentionType `json:"whenScaledDown"`
```
simonpasquier, line 1144: we had a discussion on #6198 that we wanted to align with the statefulset spec which uses `whenScaled` already.
```suggestion
	WhenScaled *WhenScaledRetentionType `json:"whenScaled,omitempty"`
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1115,6 +1127,23 @@ type PrometheusSpec struct {
...
+	WhenScaledRetentionTypeDelete WhenScaledRetentionType = "Delete"
+)
+
+type ShardRetentionPolicy struct {
+	// WhenScaled defines the retention policy when the Prometheus shards are scaled down.
+	// When set to `Retain`, the operator will keep the pods from the scaled-down shard(s), so the data can still be queried.
+	//
+	// Default: "Delete"
+	// +kubebuilder:validation:Enum=Retain;Delete
+	// +required
```
simonpasquier, line 1143: I would define the field as optional for now.
```suggestion
	// +optional
```

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1115,6 +1127,23 @@ type PrometheusSpec struct {
 	EnableAdminAPI bool `json:"enableAdminAPI,omitempty"`
 }
 
+type WhenScaledRetentionType string
+
+var (
+	WhenScaledRetentionTypeRetain WhenScaledRetentionType = "Retain"
+	WhenScaledRetentionTypeDelete WhenScaledRetentionType = "Delete"
```
simonpasquier, line 1134: (nit)
```suggestion
	RetainWhenScaledRetentionType WhenScaledRetentionType = "Retain"
	DeleteWhenScaledRetentionType WhenScaledRetentionType = "Delete"
```

mviswanathsai:
Not sure why the generate and format is failing.

slashpai:
> Not sure why the generate and format is failing.

Because import updated

```
 import (
-	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 )
 
 // ShardRetentionPolicyApplyConfiguration represents a declarative configuration of the ShardRetentionPolicy type for use
 // with apply.
 type ShardRetentionPolicyApplyConfiguration struct {
-	WhenScaled *v1.WhenScaledRetentionType `json:"whenScaled,omitempty"`
+	WhenScaled *monitoringv1.WhenScaledRetentionType `json:"whenScaled,omitempty"`
 }
 ```

File: pkg/apis/monitoring/v1/prometheus_types.go [hunk 7]
```diff
@@ -996,6 +996,18 @@ type PrometheusSpec struct {
 	// Maximum number of bytes used by the Prometheus data.
 	RetentionSize ByteSize `json:"retentionSize,omitempty"`
 
+	// ShardRetentionPolicy defines the retention policy for the Prometheus shards.
+	// (Alpha) Using this field requires the 'PrometheusShardRetentionPolicy' feature gate to be enabled.
+	//
+	// The final goals for this feature can be seen at https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#graceful-scale-down-of-prometheus-servers,
+	// however, the feature is not yet fully implemented and has known limitations:
+	// * Scale up after scale down will not re-use the previously marked-for-deletion shards.
+	// * Target redistribution doesn't happen on scale up/down.
```
nicolastakashi, line 1005: Hmm I'm not sure, I think whenever we increase or decrease number of replicas the shards will be distributed, but the data won't.

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -996,6 +996,18 @@ type PrometheusSpec struct {
...
+	// * Retention duration is not settable, for now shards are retained forever.
```
nicolastakashi, line 1006: I don't know if I follow this.
Are we retaining data forever? What does it means?

pkg/prometheus/server/operator.go
```diff
@@ -951,6 +962,22 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
 
+// As the ShardRetentionPolicy feature evolves, should delete will evolve accordingly.
+// For now, shouldDelete just returns the appropriate boolean based on the retention type.
+func (c *Operator) shouldDelete(p *monitoringv1.Prometheus) (bool, error) {
+	if !c.retentionPoliciesEnabled {
+		// Feature-gate is disabled, default behavior is always to delete.
+		return true, nil
+	}
+
+	if p.Spec.ShardRetentionPolicy.WhenScaled != nil &&
```
nicolastakashi, line 973: I think this work as well.

```suggestion
	if ptr.Dref(p.Spec.ShardRetentionPolicy.WhenScaled, monitoringv1.DeleteWhenScaledRetentionType) == monitoringv1.DeleteWhenScaledRetentionType {
	return true, nil
}
```

pkg/apis/monitoring/v1/prometheus_types.go
simonpasquier: I read it as: "if you choose policy=retain, the shards are never removed"

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -996,6 +996,18 @@ type PrometheusSpec struct {
 	// Maximum number of bytes used by the Prometheus data.
 	RetentionSize ByteSize `json:"retentionSize,omitempty"`
 
+	// ShardRetentionPolicy defines the retention policy for the Prometheus shards.
+	// (Alpha) Using this field requires the 'PrometheusShardRetentionPolicy' feature gate to be enabled.
+	//
+	// The final goals for this feature can be seen at https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202310-shard-autoscaling.md#graceful-scale-down-of-prometheus-servers,
+	// however, the feature is not yet fully implemented and has known limitations:
+	// * Scale up after scale down will not re-use the previously marked-for-deletion shards.
```
simonpasquier, line 1004: I don't think that it's true either. IIUC the only thing not supported after this PR is merged is the "delete shards based on their retention period" capability.

pkg/prometheus/server/operator.go
```diff
@@ -951,6 +962,22 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
+// As the ShardRetentionPolicy feature evolves, should delete will evolve accordingly.
+// For now, shouldDelete just returns the appropriate boolean based on the retention type.
+func (c *Operator) shouldDelete(p *monitoringv1.Prometheus) (bool, error) {
```
simonpasquier, line 967: If I read correctly, the function never returns an error.
```suggestion
func (c *Operator) shouldDelete(p *monitoringv1.Prometheus) bool {
```

pkg/prometheus/server/operator.go
```diff
@@ -940,8 +942,17 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 			return
 		}
 
+		shouldDelete, err := c.shouldDelete(p)
```
simonpasquier, line 945: (nit) it might be easier to read if the function is `shouldRetain` (e.g. no need to negate the returned value to exit early).

pkg/apis/monitoring/v1/prometheus_types.go
mviswanathsai: I think Arthur just wanted to list out what is/isn't possible after this PR alone. I may be wrong though.

>I think whenever we increase or decrease number of replicas the shards will be distributed, but the data won't.
Yep, Agreed.

mviswanathsai: I will update it a bit to reflect the current changes.

pkg/prometheus/server/operator.go
mviswanathsai: I think it was added to "future proof" this function. Later on, `shouldDelete` would also involve asking the API server for the object and getting the deletion annotation.
We can leave this too, or remove it if it seems to cause confusion.

pkg/apis/monitoring/v1/prometheus_types.go
mviswanathsai: >IIUC the only thing not supported after this PR is merged is the "delete shards based on their retention period" capability.

Correct

pkg/prometheus/server/operator.go
simonpasquier: we can always change the function signature later. I'm more on the side of keeping it consistent (e.g. if it can't return an error, it shouldn't have an error return value).

pkg/apis/monitoring/v1/prometheus_types.go
```diff
@@ -1128,6 +1138,24 @@ type PrometheusSpec struct {
...
 
+type WhenScaledRetentionType string
+
+var (
+	RetainWhenScaledRetentionType WhenScaledRetentionType = "Retain"
+	DeleteWhenScaledRetentionType WhenScaledRetentionType = "Delete"
+)
+
+type ShardRetentionPolicy struct {
+	// WhenScaled defines the retention policy when the Prometheus shards are scaled down.
```
simonpasquier, line 1149: (nit)
```suggestion
	// Defines the retention policy when the Prometheus shards are scaled down.
```

pkg/prometheus/server/operator.go
```diff
@@ -940,8 +942,17 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+			c.logger.Error("failed to determine if StatefulSet should be retained", "err", err, "name", s.GetName(), "namespace", s.GetNamespace())
+			return
+		}
+		if shouldRetain {
+			return
+		}
+
 		if err := ssetClient.Delete(ctx, s.GetName(), metav1.DeleteOptions{PropagationPolicy: ptr.To(metav1.DeletePropagationForeground)}); err != nil {
-			c.logger.Error("failed to delete StatefulSet object", "err", err, "name", s.GetName(), "namespace", s.GetNamespace())
+			c.logger.Error("deleting StatefulSet failed", "err", err, "name", s.GetName(), "namespace", s.GetNamespace())
```
simonpasquier, line 955: (nit) I would retain the current form.

simonpasquier (approved):
thanks!

PR #7288, opened 2025-01-20
feat: add fallbackScrapeProtocol to ScrapeClass

slashpai:
## Description

## Type of change

## Verification

## Changelog entry

```release-note
Add fallbackScrapeProtocol to ScrapeClass
```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -402,13 +402,13 @@ func (cg *ConfigGenerator) addScrapeProtocols(cfg yaml.MapSlice, scrapeProtocols
 	return cg.WithMinimumVersion("2.49.0").AppendMapItem(cfg, "scrape_protocols", sps)
 }
 
-// addScrapeFallbackProtocol adds the fallback_scrape_protocol field into the configuration.
-func (cg *ConfigGenerator) addScrapeFallbackProtocol(cfg yaml.MapSlice, fallbackScrapeProtocol *monitoringv1.ScrapeProtocol) yaml.MapSlice {
+// addFallbackScrapeProtocol adds the fallback_scrape_protocol field into the configuration.
+func (cg *ConfigGenerator) addFallbackScrapeProtocol(cfg yaml.MapSlice, fallbackScrapeProtocol *monitoringv1.ScrapeProtocol) yaml.MapSlice {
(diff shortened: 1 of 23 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -535,6 +535,18 @@ func mergeAttachMetadataWithScrapeClass(attachMetadata *monitoringv1.AttachMetad
...
 }
 
+func mergeFallbackScrapeProtocolWithScrapeClass(fallbackScrapeProtocol *monitoringv1.ScrapeProtocol, scrapeClass monitoringv1.ScrapeClass) *monitoringv1.ScrapeProtocol {
+	if fallbackScrapeProtocol == nil {
+		fallbackScrapeProtocol = scrapeClass.FallbackScrapeProtocol
+	}
+
+	if fallbackScrapeProtocol == nil {
+		return nil
+	}
```
dongjiang1989, line 545:
```suggestion
```
Redundant code, can it be removed?

slashpai: not redundant code, first statement checks fallbackScrapeProtocol is set or not. Since fallbackScrapeProtocol is set from ScrapeClass in that case.

So in second statetment fallbackScrapeProtocol == nil actually would be checking if it was nil even in scrapeClass

simonpasquier: I think that we can remove the test though since the next line will return  fallbackScrapeProtocol.

slashpai:
@simonpasquier can you review again?

pkg/prometheus/promcfg.go
```diff
@@ -535,6 +535,14 @@ func mergeAttachMetadataWithScrapeClass(attachMetadata *monitoringv1.AttachMetad
 	}
 }
 
+func mergeFallbackScrapeProtocolWithScrapeClass(fallbackScrapeProtocol *monitoringv1.ScrapeProtocol, scrapeClass monitoringv1.ScrapeClass) *monitoringv1.ScrapeProtocol {
+	if fallbackScrapeProtocol == nil {
```
nicolastakashi, line 539: I think we can do something like that.
```suggestion
	return ptr.Deref(fallbackScrapeProtocol, scrapeClass.FallbackScrapeProtocol)
```

PR #7294, opened 2025-01-22
[DOC] add a proposal for BlackboxExporter CRD

Pger-Y:
## Description
image

As previously discussed, I am submitting a proposal for the implementation. 

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/Documentation/proposals/20250122-blackbox-crd.md
+++ b/Documentation/proposals/20250122-blackbox-crd.md
@@ -1,0 +1,153 @@
+# Add BlackboxExporter and BlackboxExporterModule CRDs
+
+* **Owners:**
+  * [Pger-Y](https://github.com/Pger-Y)
+
+* **Related Tickets:**
+  * [Discussion](https://github.com/prometheus-operator/prometheus-operator/discussions/7134#discussioncomment-11569949)
+
+* **Other docs:**
+  * n/a
+
+This document proposes the creation of a Blackbox Operator and associated CRDs, similar to Prometheus-Operator, to manage blackbox-exporter instances and introduce a Module CRD for handling dynamic configuration of blackbox modules in a Kubernetes-native way.
+
+
+
+## Why
+The Probe CRD can define all the parameters that the ICMP Probe requires and is quite convenient for most use cases. But configuring the HTTP Probe using the Probe CRD is a lot more complex, requiring the user to define plenty of parameters. This makes the Probe CRD an insufficient solution and potentially error-prone for HTTP Probe use cases.
+
+The situation is very similar to Alertmanager/AlertmanagerConfig that prometheus-operator offers. We need a Kubernetes-native way to generate blackbox config dynamically by reading the Probe Module Config.
+
+
+## Pitfalls of the current solution
+
+The Probe CRD provides a partial solution for handling complex modules like HTTP and gRPC. However, it introduces a separate lifecycle for Blackbox management. Users must pre-define modules containing specific configurations (such as headers and request bodies) by manually updating and reloading the Blackbox configuration, before referencing these modules in their probes.
+
+## Goals
+
+- Provide a way for users to self-service adding module config for their Probe CRD
+- Consolidate the module configuration generation logic in a central point for other resources to use
+
+## Audience
+
+* Administrators providing Prometheus monitoring who want to enable self-service probe configuration for their users
+* Teams wanting to manage blackbox module configurations declaratively through Kubernetes CRDs
+* Users requiring a standardized, Kubernetes-native approach to monitor external targets
+
+## Non-Goals
+
+- This propose don't aim to replace the probe crd
+
+## How
+
+We propose creating two new Custom Resource Definitions (CRDs):
+- The BlackboxExporter CRD, which functions similarly to the Prometheus CRD, will manage the deployment configuration and lifecycle of blackbox-exporter instances
+- The BlackboxExporterModule CRD will define module specifications and automatically generate blackbox configurations, analogous to how Alertmanager works in Prometheus Operator
+
+## Alternatives
+Users may need to update or create a Probe CRD to monitor an HTTP endpoint with specific headers or body content. This often requires updating the Blackbox Exporter configuration, such as creating a new module or modifying an existing one. While users can manage this manually, it is a more cumbersome and error-prone process.
+
+## Action Plan
+
+1. Create the BlackboxExporter CRD which will manage blackbox-exporter instances as either a DaemonSet or Deployment
+2. Implement a configuration reloader sidecar container to automatically reload blackbox-exporter when configuration changes, similar to how Prometheus Operator handles configuration updates
+
+## CRDs Draft
+### 1. BlackBoxExporter CRD
+```yaml
+apiVersion: monitoring.zhipu.ai/v1alpha1
+kind: BlackboxExporter
+metadata:
+  name: blackbox-exporter-draft
+spec:
+  # Deployment mode - how the BlackboxExporter should be deployed 
+    # Or Daemonset mode
+  deployMode: Deployment
+  replicas: 2
+  
+  # Key configuration options
+  image: prom/blackbox-exporter
+  timeoutOffset: "0.5"
+  logLevelProber: "info"
+  
+  # Module selection - allows selecting which probe modules to include
+  moduleNamespaceSelector:
+    matchLabels:
+        kubernetes.io/metadata.name: monitoring
+  moduleSelector:
+    matchLabels:
+      type: http-probe
+  
+  # Resource requirements
+  resources:
+    limits:
+      memory: 200Mi
+      cpu: 100m
+    requests:
+      memory: 100Mi
+      cpu: 50m
+      
+  # Advanced configuration options
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 1000
+  
+  # Network configuration
+  webListenAddresses:
+    - ":9115"
+  externalURL: "https://blackbox.example.com"
+  
+status:
+  # Status fields will be managed by the controller
+  replicas: 2
+  availableReplicas: 2
+  updatedReplicas: 2
+  unavailableReplicas: 0
+  conditions:
+    - type: Available
+      status: "True"
+      lastTransitionTime: "2025-02-28T12:00:00Z"
+      reason: MinimumReplicasAvailable
+      message: "All replicas are available"
+
+```
+### 2. BlackBoxExporterModule CRD
+```
+---
+apiVersion: monitoring.zhipu.ai/v1alpha1
+kind: BlackboxExporterModule
+metadata:
+  name: example-http-probe
+  labels:
+    type: http-probe
+spec:
+  # HTTP probe configuration
+  # Common probe settings
+  timeout: 5s
+  prober: http
+  http:
+    # Request definition
+    method: POST
+    headers:
+      Accept: "application/json"
+    body: |
+      {
+        "name": "example",
+        "value": 42,
+        "active": true,
+        "metadata": {
+          "source": "kubernetes-crd",
+          "version": "1.0"
+        }
+      }
+    # Validation options
+    validStatusCodes: [200, 301, 302]
+    validHTTPVersions: ["HTTP/1.1", "HTTP/2.0"]
+    failIfSSL: false
+    failIfNotSSL: true
+    # Response validation
+    failIfBodyMatchesRegexp:
+      - "Error|Failure"
+    failIfBodyNotMatchesRegexp:
+      - "Success"
+```


Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,52 @@
...
+* **Owners:**
+  * [Pger-Y](https://github.com/Pger-Y)
+
+* **Related Tickets:**
+  * [Discussion](https://github.com/prometheus-operator/prometheus-operator/discussions/7134#discussioncomment-11569949)
+
+* **Other docs:**
+  * n/a
+
+This document proposes the creation of a Blackbox Operator and associated CRDs, similar to Prometheus-Operator, to manage blackbox-exporter instances and introduce a Module CRD for handling dynamic configuration of blackbox modules in a Kubernetes-native way.
```
rouke-broersma, line 12: I don't see a reason for this to be a new operator. prometheus-operator already manages AlertManager and ThanosRuler resources for example.

nicolastakashi: I think he means operator in sense of adding this capability to the actual prometheus operator, this might be confuse especially looking for our structure, where you have operator.go inside many different resources like Prometheus, Thanos and many others.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Why
+As the above discussion described the Probe CRD is extremely convenient for icmp prober, the Probe CRD can define anything icmp needed ,but when we want to use http Probe , we must define header/body and plenty of something else in blackbox config first,because http is much more complex than icmp.
+
+The situation is very similar to ScrapeConfig/ServiceMonitor that prometheus-operator does,We need a k8s way for generate blackbox config dynamic by reading the Probe Module Config 
+
+
+### Pitfalls of the current solution
+
+Using HTTP Probe with custom header and body,we must modify the confgiMap of config file and reload blackbox by hand which is not that much cloud native :(,this situation is really a copy of prometheus-operator does
```
nicolastakashi, line 24: I'd extend to something like


The ProbeCRD partially solves the current solution when users require the utilisation of complex modules, such as HTTP, GRPC and many others, requiring users have a different life cycle of the Blackbox management`

Just a general idea.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Audience
+
+* Administrators providing Prometheus monitoring who want to enable self-service probe configuration for their users
+* Teams wanting to manage blackbox module configurations declaratively through Kubernetes CRDs
+* Users requiring a standardized, Kubernetes-native approach to monitor external targets
+
+## Non-Goals
+
+* This proposal aims to be an expansion of the Probe CRD rather than replacing it.
```
nicolastakashi, line 39: I feel this is not a right definition of a Non-Goal.

Maybe we can rewrite to be: This propose don't aim to replace the probe crd.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,52 @@
...
+
+## How
+
+As discussed in [Discussion](https://github.com/prometheus-operator/prometheus-operator/discussions/7134#discussioncomment-11569949), we propose to create a new Blackbox CRD that will function similarly to the Prometheus CRD. This controller will be responsible for managing blackbox-exporter instances and handling dynamic configuration updates and reloading.
```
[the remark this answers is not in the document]
nicolastakashi: I tend to agree with this separation of concerns especially because you might hand-over the configuration of modules to other teams, like we do with Monitor Objects.

The only down (or not) side is the fact of BlackBoxCRD will be minimal

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Alternatives
+
+- Update blackbox configuration whenever a probe module is modified or added
+
+## Action Plan
```
nicolastakashi, line 49: Thanks for starting this @Pger-Y 
I'd love to see some drafts of the CRDs to discuss an initial API Definition

Documentation/proposals/20250122-blackbox-crd.md
heliapb, line 12: Hi @Pger-Y question, when you have "Kubernetes-native way", do you mean to be able to pick targets with selectors, or is it another thing I'm missing? Just because I've worked with the probe in the past and the downside at the time was the [static](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api-reference/api.md#monitoring.coreos.com/v1.ProbeTargetStaticConfig) config for targets.

simonpasquier (commented):
Thanks @nicolastakashi for pointing me to the PR and thanks @Pger-Y for starting the discussion! I've quickly read over the proposal and I've got a few initial remarks/questions:
* First of all, I think that it's an effort pertaining to this project in general. 
* Having said that, I wonder if we should extend the existing operator or if it needs to be a new project under the same org. My main arguments in favor of the second option would be that we could use modern controller frameworks (e.g. controller-runtime) and it should also decrease the cognitive load for contributors and maintainers. In general, the recommendation is to keep operators narrow-focused and the Prometheus operator surface is already (too) large.
* What would be the relationship between the new CRDs and the Probe CRD? E.g. is there a model where a user would tell "use this blackbox exporter to probe targets A, B and C using module Z"?
* Regarding the BlackboxExporter CRD, I'd want to avoid reimplementing all of the Deployment spec in a new CRD.

Documentation/proposals/20250122-blackbox-crd.md
heliapb: Ok thanks I was just considering the targets since for my specific use case there were targets inside and outside K8s, and maybe expanding on the current Probe CRD would be a good idea. But I'm guessing that should be better suited for an issue instead.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,52 @@
...
+The situation is very similar to ScrapeConfig/ServiceMonitor that prometheus-operator does,We need a k8s way for generate blackbox config dynamic by reading the Probe Module Config 
+
+
+### Pitfalls of the current solution
+
+Using HTTP Probe with custom header and body,we must modify the confgiMap of config file and reload blackbox by hand which is not that much cloud native :(,this situation is really a copy of prometheus-operator does
+
+## Goals
+
+- Provide a way for users to self-service adding module config for their Probe CRD
```
jkroepke, line 28: I feel it make more sense to have a BlackboxExporter CRD and BlackboxExporterModule CRD similar to Prometheus and PrometheusRule or Alertmanager and AlertmanagerConfig

nicolastakashi:
> > Thanks @nicolastakashi for pointing me to the PR and thanks @Pger-Y for starting the discussion! I've quickly read over the proposal and I've got a few initial remarks/questions:
> > 
> > * First of all, I think that it's an effort pertaining to this project in general.
> > * Having said that, I wonder if we should extend the existing operator or if it needs to be a new project under the same org. My main arguments in favor of the second option would be that we could use modern controller frameworks (e.g. controller-runtime) and it should also decrease the cognitive load for contributors and maintainers. In general, the recommendation is to keep operators narrow-focused and the Prometheus operator surface is already (too) large.
> > * What would be the relationship between the new CRDs and the Probe CRD? E.g. is there a model where a user would tell "use this blackbox exporter to probe targets A, B and C using module Z"?
> > * Regarding the BlackboxExporter CRD, I'd want to avoid reimplementing all of the Deployment spec in a new CRD.
> 
> Hi @simonpasquier sorry for the delayed reply—I’ve been on holiday the past week.
> 
> * As discussed in Slack, I’m quite unsure whether we should extend the existing operator or start a new one at beginning. If we decide to extend the current project, it could lead to a new operator focused on the blackbox, which would make the existing project increasingly complex. After thinking it over these past few days, I’m leaning toward starting a new project focused specifically on the blackbox. However, in the Probe CRD, we would need to import modules defined in the new CRDs. I believe this is a flaw in the design, as these two CRDs (Probe and Module), which are related, are split across two projects. :(
> * I think the relation between CRDs and the Probe CRD is very close,for example we create a CRDs like module-a which defind the head/body,then we can use this module in Probe
> 
> ```
> kind: Probe
> apiVersion: monitoring.coreos.com/v1
> metadata:
>   name: example-com-website
>   namespace: monitoring
> spec:
>   interval: 60s
>   module: module-a
>   prober:
>     url: blackbox-exporter.monitoring.svc.cluster.local:19115
>   targets:
>     staticConfig:
>       static:
>       - http://example.com
>       - https://example.com
> ```

What @simonpasquier means about a new project, is a new project under the prometheus-operator organisation, which means we could leverage the same api name `monitoring.coreos.com` (IIRC) but with the benefity of using a brand new set of tools available on the kubebuilder project.

I tend to agree that might a bit weird ProbCR ref one CR managed by other controller, even if it belongs to the same org.

rouke-broersma:
Probe already allows setting module name, I think it's fine to keep that relationship indirect as it is now, and later add a ExporterRef and ModuleRef-like structure if it's wanted in the future. I don't see an issue with that being different controllers in different projects working together. However that sounds like a design decision that needs to be made holistically for all new product additions and should be documented in a process so this doesn't need to be a question every time (if that doesn't exist yet of course).

nicolastakashi:
> Probe already allows setting module name, I think it's fine to keep that relationship indirect as it is now, and later add a ExporterRef and ModuleRef-like structure if it's wanted in the future. I don't see an issue with that being different controllers in different projects working together. However that sounds like a design decision that needs to be made holistically for all new product additions and should be documented in a process so this doesn't need to be a question every time (if that doesn't exist yet of course).

I agree, we need to try as much as we can document the reason behind the decision to make a new controller, I think one of our main concerns if the fact the Prometheus is a large ecosystem and in the future we can have requests about deploying other softwares which are part of the ecosystem but we're not sure if they fits properly the Prom Operator use case.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,2722 @@
...
+
+* **Other docs:**
+  * n/a
+
+This document proposes the creation of a Blackbox Operator and associated CRDs, similar to Prometheus-Operator, to manage blackbox-exporter instances and introduce a Module CRD for handling dynamic configuration of blackbox modules in a Kubernetes-native way.
+
+
+
+## Why
+As the above discussion described the Probe CRD is extremely convenient for icmp prober, the Probe CRD can define anything icmp needed ,but when we want to use http Probe , we must define header/body and plenty of something else in blackbox config first,because http is much more complex than icmp.
```
heliapb, line 17: Hi @Pger-Y just a nit pic about punctuation:

```MD
As the above discussion described the Probe CRD is extremely convenient for icmp prober, the Probe CRD can define anything icmp needed, but when we want to use http Probe we must define header/body and plenty of something else in blackbox config first, because http is much more complex than icmp.

The situation is very similar to ScrapeConfig/ServiceMonitor that prometheus-operator does. We need a k8s way for generate blackbox config dynamic by reading the Probe Module Config 
```

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,2722 @@
...
+
+The situation is very similar to ScrapeConfig/ServiceMonitor that prometheus-operator does,We need a k8s way for generate blackbox config dynamic by reading the Probe Module Config
```
heliapb, line 19: Also I do believe the ScrapeCRD|ServiceMonitor is perhaps not the best comparison, since by design was done to add [additional service discover capabilities to targets outside k8s](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202212-scrape-config.md#why). IMO Alertmanager|AlertmanagerConfig is a better comparison

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,2722 @@
...
+
+## Non-Goals
+
+- This propose don't aim to replace the probe crd
+
+## How
+
+As discussed in [Discussion](https://github.com/prometheus-operator/prometheus-operator/discussions/7134#discussioncomment-11569949), we propose creating two new Custom Resource Definitions (CRDs):
+- The BlackboxExporter CRD, which functions similarly to the Prometheus CRD, will manage the deployment configuration and lifecycle of blackbox-exporter instances
+- The BlackboxExporterModule CRD will define module specifications and automatically generate blackbox configurations, analogous to how ScrapeConfig works in Prometheus Operator
```
heliapb, line 45: I guess it needs to be changed here as well

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,2722 @@
...
+* Teams wanting to manage blackbox module configurations declaratively through Kubernetes CRDs
+* Users requiring a standardized, Kubernetes-native approach to monitor external targets
+
+## Non-Goals
+
+- This propose don't aim to replace the probe crd
+
+## How
+
+As discussed in [Discussion](https://github.com/prometheus-operator/prometheus-operator/discussions/7134#discussioncomment-11569949), we propose creating two new Custom Resource Definitions (CRDs):
```
simonpasquier, line 43: (nit) I'd remove the link here.
```suggestion
We propose creating two new Custom Resource Definitions (CRDs):
```

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,2722 @@
...
+## Alternatives
+
+- Update blackbox configuration whenever a probe module is modified or added
+
+## Action Plan
+
+1. Create the BlackboxExporter CRD which will manage blackbox-exporter instances as either a DaemonSet or Deployment
+2. Implement a configuration reloader sidecar container to automatically reload blackbox-exporter when configuration changes, similar to how Prometheus Operator handles configuration updates
+
+## CRDs draft
```
simonpasquier, line 56: Thanks for adding the CRD structure. Having said that it's too detailed for a design proposal. In general, it's sufficient to add an example of the custom resource with the most important fields.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -14,12 +14,12 @@ This document proposes the creation of a Blackbox Operator and associated CRDs,
 
 
 ## Why
-As the above discussion described, the Probe CRD is extremely convenient for ICMP prober. The Probe CRD can define anything ICMP needed, but when we want to use HTTP Probe, we must define header/body and plenty of other parameters in blackbox config first, because HTTP is much more complex than ICMP.
+In our daily use, we have discovered that the Probe CRD is extremely convenient for ICMP prober. The Probe CRD can define anything ICMP needed, but when we want to use HTTP Probe, we must define header/body and plenty of other parameters in blackbox config first, because HTTP is much more complex than ICMP.
```
mviswanathsai, line 17: (nit)
```suggestion
The Probe CRD can define all the parameters that the ICMP Probe requires and is quite convenient for most use cases. But configuring the HTTP Probe using the Probe CRD is a lot more complex, requiring the user to define plenty of parameters. This makes the Probe CRD an insufficient solution and potentially error-prone for HTTP Probe use cases.
```

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -14,12 +14,12 @@ This document proposes the creation of a Blackbox Operator and associated CRDs,
...
 
 The situation is very similar to Alertmanager/AlertmanagerConfig that prometheus-operator does. We need a Kubernetes-native way to generate blackbox config dynamically by reading the Probe Module Config.
```
mviswanathsai, line 19: (nit)
```suggestion
The situation is very similar to Alertmanager/AlertmanagerConfig that prometheus-operator offers. We need a Kubernetes-native way to generate blackbox config dynamically by reading the Probe Module Config.
```

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -40,9 +40,9 @@ The Probe CRD provides a partial solution for handling complex modules like HTTP
 
 ## How
 
-As discussed in [Discussion](https://github.com/prometheus-operator/prometheus-operator/discussions/7134#discussioncomment-11569949), we propose creating two new Custom Resource Definitions (CRDs):
+We propose creating two new Custom Resource Definitions (CRDs):
 - The BlackboxExporter CRD, which functions similarly to the Prometheus CRD, will manage the deployment configuration and lifecycle of blackbox-exporter instances
-- The BlackboxExporterModule CRD will define module specifications and automatically generate blackbox configurations, analogous to how ScrapeConfig works in Prometheus Operator
+- The BlackboxExporterModule CRD will define module specifications and automatically generate blackbox configurations, analogous to how Alertmanager works in Prometheus Operator
```
mviswanathsai, line 45: Might be a stupid question. But isn't generating a configuration based on the CRD more relevant to the `ScrapeConfig` - `Prometheus` relationship?

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -53,2670 +53,96 @@ As discussed in [Discussion](https://github.com/prometheus-operator/prometheus-o
 1. Create the BlackboxExporter CRD which will manage blackbox-exporter instances as either a DaemonSet or Deployment
 2. Implement a configuration reloader sidecar container to automatically reload blackbox-exporter when configuration changes, similar to how Prometheus Operator handles configuration updates
 
-## CRDs draft
+## CRDs Drafg
```
mviswanathsai, line 56: Typo :)
```suggestion
## CRDs Draft
```

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -53,2670 +53,96 @@ As discussed in [Discussion](https://github.com/prometheus-operator/prometheus-o
...
```
mviswanathsai, line 57: (suggestion)
### 1. `BlackBoxExporter` CRD

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -53,2670 +53,96 @@ As discussed in [Discussion](https://github.com/prometheus-operator/prometheus-o
...
-            - updatedReplicas
-            type: object
-        type: object
-    served: true
-    storage: true
-    subresources:
-      status: {}
-```
-
-## 2. `Module` CRD
```
mviswanathsai: keeping this seems reasonable.
### 2. `BlackBoxExporterModule` CRD

mviswanathsai (commented):
A few nits. Thanks for the effort!

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,148 @@
...
+
+```yaml
+apiVersion: monitoring.zhipu.ai/v1alpha1
+kind: BlackboxExporter
+metadata:
+  name: blackbox-exporter-draft
+spec:
+  # Deployment mode - how the BlackboxExporter should be deployed 
+    # Or Daemonset mode
+  deployMode: Deployment
```
simonpasquier, line 66: is there a real need for a daemonset  mode? what would be the use case?

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,148 @@
...
+  replicas: 2
+  
+  # Key configuration options
+  baseImage: prom/blackbox-exporter
```
simonpasquier, line 70: baseImage is  deprecated, we should just have an `image` field.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,148 @@
...
+  timeoutOffset: "0.5"
+  logLevelProber: "info"
+  
+  # Module selection - allows selecting which probe modules to include
+  moduleSelector:
+    matchLabels:
+      type: http-probe
+  
+  # Resource requirements
```
simonpasquier, line 79: I'd be wary to include more deployment-specific fields to the CRD because from experience, we'll end up re-implementing the deployment spec. One option could be to let users customize the generated deployment as they want relying on [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/#custom-resources-and-server-side-apply). I've used it in other projects and it works fine (provided that you run a recent Kubernetes version).

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,148 @@
...
+  deployMode: Deployment
+  replicas: 2
+  
+  # Key configuration options
+  baseImage: prom/blackbox-exporter
+  timeoutOffset: "0.5"
+  logLevelProber: "info"
+  
+  # Module selection - allows selecting which probe modules to include
+  moduleSelector:
```
simonpasquier, line 75: I assume that we want a moduleNamespaceSelector for consistency with other resources.

Documentation/proposals/20250122-blackbox-crd.md
mviswanathsai: +1

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,148 @@
...
+metadata:
+  name: example-http-probe
+  labels:
+    type: http-probe
+spec:
+  # HTTP probe configuration
+  # Common probe settings
+  timeout: 5s
+  prober: http
+  http:
```
heliapb, line 123: This resembles the BB configuration as is, but I was thinking maybe we can add something like, WDYT?

```yaml
prober: http // (http, tcp, dns, icmp, grpc)
proberConfg:
    http:
      method: GET
      headers:
        Host: host.example.com
    icmp:
      preferred_ip_protocol: "ip4"
```

Documentation/proposals/20250122-blackbox-crd.md
heliapb: Well IMO considering how BB works and it needs the prober with the prober type and relevant configuration for each probe I think is more a relation if configuration, so I think is more closer related to Alertmanager|AlermanagerConfig

Documentation/proposals/20250122-blackbox-crd.md
heliapb: Hi @Pger-Y you're right, I was thinking a bit about the current AlertmanagerConfig structure when I_ suggested this - https://doc.crds.dev/github.com/prometheus-operator/prometheus-operator/monitoring.coreos.com/AlertmanagerConfig/v1alpha1@v0.80.0. But we can just keep this simpler for a first version. But at the same time probably keeping as close as the current configuration could be a good idea, just in terms of structure, for context https://github.com/prometheus/blackbox_exporter/blob/master/example.yml

simonpasquier:
> I am currently implementing the design we discussed earlier. During this process, I am considering the optimal way to configure the `configReloader` for images and resources. I have observed that in the Prometheus Operator, image and resource limits are typically set through operator arguments. Therefore, I am exploring the possibility of defining `configReloader` parameters within the Custom Resource Definition (CRD),for example blackbox.spec.configReloader.image/resource. What are your thoughts on this approach?

This should be part of the discussion on this design doc. I'd encourage that you add these questions in the doc (add an "open questions" section at the end if you prefer).

Documentation/proposals/20250122-blackbox-crd.md
simonpasquier: @Pger-Y the API structure should follow the general Kube API conventions. While it may seem counter-intuitive to not always follow the Prometheus config format, the Kube conventions exist for good reasons.

https://prometheus-operator.dev/docs/community/contributing/#changes-to-the-apis

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Action Plan
+
+1. Create the BlackboxExporter CRD which will manage blackbox-exporter instances as either a DaemonSet or Deployment
+2. Implement a configuration reloader sidecar container to automatically reload blackbox-exporter when configuration changes, similar to how Prometheus Operator handles configuration updates
+
+## CRDs Draft
+### 1. BlackBoxExporter CRD
+```yaml
+apiVersion: monitoring.zhipu.ai/v1alpha1
```
simonpasquier, line 58: (nit) wrong group

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Alternatives
+Users may need to update or create a Probe CRD to monitor an HTTP endpoint with specific headers or body content. This often requires updating the Blackbox Exporter configuration, such as creating a new module or modifying an existing one. While users can manage this manually, it is a more cumbersome and error-prone process.
+
+## Action Plan
+
+1. Create the BlackboxExporter CRD which will manage blackbox-exporter instances as either a DaemonSet or Deployment
+2. Implement a configuration reloader sidecar container to automatically reload blackbox-exporter when configuration changes, similar to how Prometheus Operator handles configuration updates
+
+## CRDs Draft
```
simonpasquier, line 55: (nit) this should go in the `How` section

Documentation/proposals/20250122-blackbox-crd.md
simonpasquier: I was talking about all the fields which are specific to container definitions like `resources`

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,153 @@
...
+spec:
+  # Deployment mode - how the BlackboxExporter should be deployed 
+    # Or Daemonset mode
+  deployMode: Deployment
+  replicas: 2
+  
+  # Key configuration options
+  image: prom/blackbox-exporter
+  timeoutOffset: "0.5"
+  logLevelProber: "info"
```
simonpasquier, line 71: we need fields for `--log.level`, `--log.format` and `--history.limit` too.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,153 @@
...
+      memory: 100Mi
+      cpu: 50m
+      
+  # Advanced configuration options
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 1000
+  
+  # Network configuration
+  webListenAddresses:
```
simonpasquier, line 96: I'd recommend defining all web fields in the same group

````
spec:
  web:
    listenAddresses: [...]
    externalURL: ...
    ...
```

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,153 @@
...
+spec:
+  # HTTP probe configuration
+  # Common probe settings
+  timeout: 5s
+  prober: http
+  http:
+    # Request definition
+    method: POST
+    headers:
+      Accept: "application/json"
```
simonpasquier, line 132: see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps
maps aren't recommended in Kubernetes APIs.

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,153 @@
...
+      status: "True"
+      lastTransitionTime: "2025-02-28T12:00:00Z"
+      reason: MinimumReplicasAvailable
+      message: "All replicas are available"
+
+```
+### 2. BlackBoxExporterModule CRD
+```
+---
+apiVersion: monitoring.zhipu.ai/v1alpha1
```
simonpasquier, line 117: same here

Documentation/proposals/20250122-blackbox-crd.md
```diff
@@ -0,0 +1,153 @@
...
+- The BlackboxExporter CRD, which functions similarly to the Prometheus CRD, will manage the deployment configuration and lifecycle of blackbox-exporter instances
+- The BlackboxExporterModule CRD will define module specifications and automatically generate blackbox configurations, analogous to how Alertmanager works in Prometheus Operator
+
+## Alternatives
+Users may need to update or create a Probe CRD to monitor an HTTP endpoint with specific headers or body content. This often requires updating the Blackbox Exporter configuration, such as creating a new module or modifying an existing one. While users can manage this manually, it is a more cumbersome and error-prone process.
+
+## Action Plan
+
+1. Create the BlackboxExporter CRD which will manage blackbox-exporter instances as either a DaemonSet or Deployment
+2. Implement a configuration reloader sidecar container to automatically reload blackbox-exporter when configuration changes, similar to how Prometheus Operator handles configuration updates
```
simonpasquier, line 53: is there a need for another config reloader? I imagine that we can reuse the existing one.

simonpasquier (commented):
I don't see any decision in the document about the creation (or not) of a new operator.

PR #7295, opened 2025-01-24
chore: update go dependencies for release v0.80

slashpai:
## Description

## Type of change

## Verification

## Changelog entry

```release-note

```


o.txt
slashpai, line 1: This file came as by mistake in my last PR https://github.com/prometheus-operator/prometheus-operator/pull/7288/files#diff-31dbfad25b1660facfe7fc6205e4d9067a87a0cb150c41fa56eb23f2fda80779 hence removing it

slashpai:
Unit test failures after dependency update:
    --- FAIL: TestMakeRulesConfigMaps/shouldRejectRuleWithInvalidLabels (0.00s)
    --- FAIL: TestSelectScrapeConfigs/staticConfig_with_invalid_Labels (0.00s)
Looking into it

slashpai:
> Unit test failures after dependency update: --- FAIL: TestMakeRulesConfigMaps/shouldRejectRuleWithInvalidLabels (0.00s) --- FAIL: TestSelectScrapeConfigs/staticConfig_with_invalid_Labels (0.00s) Looking into it

I tried without bumping prometheus common so tests are passing. Issue seems to be because default validation scheme is changed to UTF in 0.62 https://github.com/prometheus/common/commit/cf5f48f181f5270eeac48e077e974ba0e00f5c58#diff-1cb3678072281aa172be75f8f835b1b70374aa1fc069afdaadc0df269b6868d3

simonpasquier:
Thanks for the investigation! I think that for now we want to minimize changes and thus retain the legacy mode (e.g. not UTF-8 compatible). I'd advise to set the global `NameValidationScheme` to `LegacyValidation` in all binaries that make use of it (e.g. admission webhook + Prometheus operator).

Eventually we want to support UTF-8 labels but it's more work required and I'm not sure what's the right approach without digging further in the code. Maybe something like:
1. A new CLI flag for the admission webhook to switch between the 2 validations.
2. Teaching the operator to switch the NameValidationScheme variable on the fly depending on the Prometheus version?
WDYT?

slashpai:
> Thanks for the investigation! I think that for now we want to minimize changes and thus retain the legacy mode (e.g. not UTF-8 compatible). I'd advise to set the global `NameValidationScheme` to `LegacyValidation` in all binaries that make use of it (e.g. admission webhook + Prometheus operator).
> 
> Eventually we want to support UTF-8 labels but it's more work required and I'm not sure what's the right approach without digging further in the code. Maybe something like:
> 
> 1. A new CLI flag for the admission webhook to switch between the 2 validations.
> 2. Teaching the operator to switch the NameValidationScheme variable on the fly depending on the Prometheus version?
>    WDYT?

My first thought was to make legacy validation by default as there could be users using Prometheus 2 still even though operator default is Prometheus 3. Users who want to use UTF 8 can set NameValidationScheme to UTF 8 which should be ok for now I think as it been supported already.

slashpai:
This is a bit hard than I thought for a quick fix before release. I tried setting Legacy as default for https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/prometheus_types.go#L691 but for unit test we can't do legacy validation till the logic is been updated for rules to handle v2 vs v3. Also an e2e test would be good to make sure it works correctly.

I think its better to update prometheus/common in next release so that we get enough time for reviewing as we are a bit behind the schedule for release. I can log a github issue so that we don't forget about this

slashpai:
I updated the PR with skipping prometheus/common update
@simonpasquier can you review?

simonpasquier (approved):

Issue #5408, opened 2023-03-13
The prometheusRule took a long time to take effect


lycclsltt:
Hi, we are a Chinese Internet company based on a mass surveillance system developed by prometheus.

Now, there's a really nasty problem with a prometheusRule node loaded with 2000 PrometheusRules, each prometheusRule with 10 alerts. Each time a new prometheusRule is created, prometheusRule takes about 3-5 minutes to load into prometheus, which is quite a long time. I looked at the PrometheusOperator code and it seems that every time a PrometheusRule is updated, In the event, all resources will be updated again. I wonder if it takes a long time to take effect because of this. This problem has been bothering us for a long time. Is there any way to shorten the effective time? Looking forward to your reply

slashpai:
👋🏻  Thanks for the report

Could you please provide more details as in issue template?
https://github.com/prometheus-operator/prometheus-operator/blob/main/.github/ISSUE_TEMPLATE/bug.md

lycclsltt:
> 👋🏻 Thanks for the report
> 
> Could you please provide more details as in issue template? https://github.com/prometheus-operator/prometheus-operator/blob/main/.github/ISSUE_TEMPLATE/bug.md

What happened?

After creating a prometheusRule, prometheus pod takes 3-5 minutes to load and take effect

Did you expect to see something different?

It shouldn't take this long

How to reproduce it (as minimally and precisely as possible):

I don't know yet

Environment

k8s environment

Prometheus Operator version:

0.47.0

Insert image tag or Git SHA here

Tag: 0.47.0

Kubernetes version information:

Client Version:  Version. The Info {Major: "1", Minor: "18", GitVersion: "v1.18.6 GitCommit:" dff82dc0de47299ab66c83c626e08b245ab19037." GitTreeState:"clean", BuildDate:"2020-07-15T16:58:53Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version:  Version. The Info {Major: "1", Minor: "18", GitVersion: "v1.18.6 GitCommit:" dff82dc0de47299ab66c83c626e08b245ab19037." GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

kubectl version

Refer to above

Kubernetes cluster kind:

lycclsltt:
[...]

PR #7366, opened 2025-02-24
fix(reloader): fix broken config file watcher

machine424:
explicitly watch the config file parent directory as the underlying fsnotify library used by the reloader does not recommend watching individual files since it is easy to misuse. See https://github.com/fsnotify/fsnotify/blob/a9bc2e01792f868516acf80817f7d7d7b3315409/README.md?plain=1#L128

For example, atomic file updates will render the initial file watcher obsolete. That watcher will only receive the first DELETE and then stop sending any events, causing the reloader to fall back to defaultWatchInterval (3 minutes).

kubelet updates the ConfigMap/Secret mounted files atomically using https://pkg.go.dev/k8s.io/kubernetes/pkg/volume/util#AtomicWriter

Because the operator is responsible for defining config file mounting, I suppose no other files exist in the parent directory that might trigger unneeded events.

Also, watching the config file's parent directory will catch events corresponding to kubelet's AtomicWriter  manipulations on symlinks and modifications to the timestamped directories which contain the various actual config file versions.

Given that, watching the entire parent directory may result in multiple events at once, but the reloader is able to smooth them out.

Because this "bug" has been around for a long time, I'm not sure about the consequences; perhaps some setups might not be able to cope with this acceleration of the reconciliation loop, but I'm sure this will be a significant time saver for a lot of users and "wait for config to get applied" tests/checks.

Maybe we should also think about whether to want/can teach the reloader to handle this on its side.

fixes https://github.com/prometheus-operator/prometheus-operator/issues/5408

- [ ] add regression tests
- [ ] this is not even manually tests

## Description

## Type of change

## Verification

## Changelog entry

```release-note

```


pkg/operator/config_reloader.go
```diff
@@ -264,6 +265,13 @@ func CreateConfigReloader(name string, options ...ReloaderOption) v1.Container {
 
 	if len(configReloader.configFile) > 0 {
 		args = append(args, fmt.Sprintf("--config-file=%s", configReloader.configFile))
+		// The underlying fsnotify library used by the reloader does not recommend watching individual files.
+		// For example, atomic file updates will make the file watcher obsolete. Watch the dir instead.
+		// See https://github.com/fsnotify/fsnotify/blob/a9bc2e01792f868516acf80817f7d7d7b3315409/README.md?plain=1#L128
+		confDir := configReloader.configFile
+		if !slices.Contains(configReloader.watchedDirectories, confDir) {
```
machine424, line 272: Not sure how many elements the slice can contain (if this can be slow)

simonpasquier:
Thanks!

> but I'm sure this will be a significant time saver for a lot of users and "wait for config to get applied" tests/checks.

I haven't looked into the details yet but from my personal experience, the main reason why users experience delays in the configuration update propagation is because the kubelet doesn't update the mounted secrets/configmaps immediately (https://github.com/kubernetes/kubernetes/issues/30189 and https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically). From my POV, it's not a "bad" thing: in highly dynamic environments, it's probably preferred to have some slack in config reloads instead of hammering the /-/reload endpoint.

machine424:
Yes, with this change, reloads are no longer dictated by the defaultWatchInterval (3 minutes), even when the kubelet has already updated the files. Instead, the updates will be mainly subject to the kubelet’s update delay (which can be customized if needed) and the reloader’s throttling.

On a `kind` with the default configuration/latest image, the frequency moves from "a reload every 3 minutes" to a "a reload every ~1m20s" (mostly due to the kubelet delay).

Yes, I was also planning to open an issue on the reloader side. Thought we may need a quick fix in here.

machine424:
opened https://github.com/thanos-io/thanos/issues/8127

pkg/operator/config_reloader.go
machine424: The `append` is probably duplicated for some cases

sshambar:
I've been working around this issue by adding "--watched-dir=/etc/prometheus/config" to the config-reloader container args in my Prometheus object -- this works, and reloads prometheus as soon as kubelet updates the config ConfigMap

... but it's a "hack" in the sense that I can't "append the arg" but need to provide all the args in my Prometheus object - which will likely break in some future operator change to the config-reloader container. 

@simonpasquier IMHO the operator is the correct place to add the watchedDirectory.  Patching the thanos reloader would prevent it from actually watching just a specific "file" (and not also it's directory), and so may be rejected.  The operator mounts /etc/prometheus/config from the Secret, and so "owns" the directory and knows all files there are of interest.  Realistically, it _only_ needs to watch the /etc/prometheus/config directory, and not the file (you never know, there may be more in the secret that are of interest!)

Ideally, the Prometheus object would benefit from having some element for "watchedDirectories" which could be added to the reloader by the operator, but that's beyond what this PR is doing :)

machine424:
As discussed with @simonpasquier, the change also speeds up some of the e2e tests.

Compared with https://github.com/prometheus-operator/prometheus-operator/pull/8471, `e2e / E2E tests (prometheus)` duration drops from ~42min to ~25min.

The e2e tests that see noticeable improvements are those that trigger more than one config change. This is because the Thanos reloader already watches the config file https://github.com/thanos-io/thanos/blob/aec3c15b6803a4c24286ad3d3a9dca6f357657aa/pkg/reloader/reloader.go#L288 which causes inotify to fire on the first kubelet atomic swap. After that event, the watcher is dead, so subsequent config changes can only be detected by the `defaultWatchInterval` polling timer (~3 minutes).

With this PR, the config file's parent directory is watched instead. Directory watches survive atomic symlink swaps, so every config update is detected immediately via fsnotify.

The time saved is therefore proportional to `(number_of_config_changes - 1) * defaultWatchInterval`.

For example, `TestPromInstanceNs/ScrapeConfigLifecycle` changes the config 3 times. Its execution time dropped from 405s to 74s, consistent with saving roughly `2 * 180s = 360s`.

The `e2e / E2E tests (prometheusAllNS)` suite did not see any notable improvement because most of its tests should be changing the config at most once and because the tests run in parallel, the suite's exec duration is determined by the slowest test, which may not be reload-dependent at all...

simonpasquier:
> another approach would be to make each component set this for its reloader

I'd lean towards "fixing" the issue into config-reloader because other projects (such as https://github.com/thanos-community/thanos-operator) depend on the reloader and they would benefit for the fix without further modification.

PR #7415, opened 2025-03-16
feat: Use protobuf encoding for core K8s APIs in prometheus-operator

tosi3k:
## Description
This PR makes Prometheus Operator's static K8s client use Protobuf encoding when talking to K8s API.

If not specified explicitly, JSON encoding is used by default which is less efficient than protobuf.

For core K8s API objects like Pods, Nodes, etc., we can use protobuf encoding which reduces CPU consumption related to (de)serialization, reduces overall latency of the API call, reduces memory footprint, reduces the amount of work performed by the GC and results in quicker propagation of objects to event handlers of shared informers.

Standard system components of K8s default their serialization method to protobuf for some time already:
- `kube-scheduler`: [link](https://github.com/kubernetes/kubernetes/blob/d62b797c16fdffa55a8e2fc95f04bf72c019be70/pkg/scheduler/apis/config/v1/defaults.go#L143-L145)
- `kubelet`: [link](https://github.com/kubernetes/kubernetes/blob/d62b797c16fdffa55a8e2fc95f04bf72c019be70/pkg/kubelet/apis/config/v1beta1/defaults.go#L199-L201)
- `kube-proxy`: [link](https://github.com/kubernetes/kubernetes/blob/d62b797c16fdffa55a8e2fc95f04bf72c019be70/pkg/proxy/apis/config/v1alpha1/defaults.go#L126-L128)
- `kube-controller-manager`: [link](https://github.com/kubernetes/kubernetes/blob/d62b797c16fdffa55a8e2fc95f04bf72c019be70/staging/src/k8s.io/controller-manager/config/v1alpha1/defaults.go#L49) and [method's contents](https://github.com/kubernetes/kubernetes/blob/d62b797c16fdffa55a8e2fc95f04bf72c019be70/staging/src/k8s.io/component-base/config/v1alpha1/defaults.go#L66-L68)

Some benchmarks comparing JSON vs. protobuf showcasing how the latter data format (de)serializes faster and uses less memory:
- https://medium.com/@akresling/go-benchmark-json-v-protobuf-4ec3c62ec8d4
- https://www.codingexplorations.com/blog/performance-comparison-protobuf-marshaling-vs-json-marshaling-in-go
- https://medium.com/@david_turner/protocol-buffers-just-how-fast-are-they-9ec19ea580db

## Type of change

## Verification
Ran unit and e2e tests.

## Changelog entry
```release-note
Use protobuf encoding for core K8s APIs in Prometheus Operator.
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -35,6 +35,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -244,7 +245,9 @@ func run(fs *flag.FlagSet) int {
 		return 1
 	}
 
-	kclient, err := kubernetes.NewForConfig(restConfig)
+	restConfigProtobuf := rest.CopyConfig(restConfig)
+	restConfigProtobuf.ContentType = runtime.ContentTypeProtobuf
+	kclient, err := kubernetes.NewForConfig(restConfigProtobuf)
 	if err != nil {
 		logger.Error("failed to create Kubernetes client", "err", err)
 		cancel()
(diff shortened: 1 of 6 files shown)


simonpasquier:
Thanks for the PR! I believe that k8s.io/client-go has switched to protobuf by default in v1.32.0:
https://github.com/kubernetes/kubernetes/commit/7b7a7968d4c90c4430b1009546e13ce78bcde134
https://github.com/kubernetes/client-go/commit/6f44458e5efa4a1d73063e438616bac9f980b67b

for instance => https://github.com/kubernetes/client-go/blob/release-1.32/kubernetes/typed/core/v1/service.go#L62-L74

tosi3k:
Thanks for pointing this out - I wasn't aware of that :P

That said, using client-go 1.32+ makes this PR redundant - let me close it then.

PR #7421, opened 2025-03-18
fix(scrapeConfigs): do not insert sharding relabeling if already present

bodji:
Use of the function introduced in #6135 for ScrapeConfig, in ServiceMonitor objects

## Description

The PR #6135 introduced a check to not add sharding labels (with modulus) if there was already a hashmod action.

This PR uses this function in ServiceMonitor function.

## Type of change

## Verification

I compiled the Prometheus Operator locally and tested on Minikube cluster by adding a custom  label : 

```
kind: ServiceMonitor
metadata:
  name: nginx-exporter-servicemonitor
  labels:
    app: nginx-exporter
spec:
  selector:
    matchLabels:
      app: nginx-exporter
  endpoints:
    - targetPort: 9113
      interval: 30s
      path: /metrics
      relabelings:
        - action: hashmod
          sourceLabels: [__address__]
          modulus: 1
          targetLabel: __tmp_bodji
```

## Changelog entry

`fix(scrapeConfigs): do not insert sharding relabeling if already present
`


simonpasquier:
I'm not sure what you're trying to achieve but if you want to customize the hash value used for sharding, you should be copying it to `__tmp_hash`. From the documentation of the `.spec.shards` field (https://prometheus-operator.dev/docs/api-reference/api/):

```
By default, the sharding of targets is performed on:
* The __address__ target’s metadata label for PodMonitor, ServiceMonitor and ScrapeConfig resources.
* The __param_target__ label for Probe resources.

Users can define their own sharding implementation by setting the __tmp_hash label during the
target discovery with relabeling configuration (either in the monitoring resources or via scrape class).
````

**Comment by bodji** [comment]
No it's not about customizing the hash value.

I want to disable sharding stuff on a specific ServiceMonitor.

It has been done in #6135 by adding the `generateAddressShardingRelabelingRulesIfMissing` function, but this function is called only in `generateScrapeConfig` and not in `generateServiceMonitorConfig`

**Comment by simonpasquier** [comment]
Thanks for the details. Out of curiosity what's your use case for scraping a target from all shards?
 
Technically, I would go with defining another `__tmp_` attribute which could be used to prevent target sharding. Something like:

```
  - source_labels:
    - __address__
    target_label: __tmp_hash
    modulus: 1
    action: hashmod
  - source_labels:
    - __tmp_hash
    - __tmp_disable_sharding
    # The regex should keep the target if the __tmp_hash value matches the shard ID *or* if the user decided to disable sharding explicitly by setting the __tmp_disable_sharding label to a non-empty value.
    regex: "($(SHARD);|;.+)"
    action: keep
```

AFAICT the operator has always injected the hashmod relabel config for service & pod monitors so changing the behavior would potentially be a breaking change. The "don't insert if present" case only applies to `.spec.additionalScrapeConfigs` and ScrapeConfig CRD because they are lower-level APIs and it makes sense there to leave more user freedom.

**Comment by bodji** [comment]
We need sharding to divide the load for scraping pods metrics (kubelet / cAdvisor)

But we also need kube_pod_labels from kube-state-metrics, but each shard must have all kube-state-metrics to have all kube_pod_labels series.

File: test/e2e/prometheus_test.go
Comment (simonpasquier) [inline test/e2e/prometheus_test.go:1]: The testShardingProvisioning test would need a rewrite because it doesn't really make sense to check the generated config at the e2e testing level. But it's a work for another PR.

**Review by simonpasquier — COMMENTED** [review]
Can we also document the new tmp label here?
https://github.com/prometheus-operator/prometheus-operator/blob/b8e2df1c092b464deee69b178451a7f068a9ffdb/pkg/apis/monitoring/v1/prometheus_types.go#L259-L267

**Comment by simonpasquier** [comment]
Sorry for the delay, I missed your update.

File: pkg/apis/monitoring/v1/prometheus_types.go [hunk 1]
```diff
@@ -265,6 +265,9 @@ type CommonPrometheusFields struct {
 	// `__tmp_hash` label during the target discovery with relabeling
 	// configuration (either in the monitoring resources or via scrape class).
 	//
+	// You can also disable sharding on a specific scrapeConfig by setting the
+	// `__tmp_disable_sharding` label, the operator won't add the modulus action
+	// when this label is present on the relabelings.
```
simonpasquier, line 270: (nit)
```suggestion
	// You can also disable sharding on a specific target by setting the
	// `__tmp_disable_sharding` label` with relabeling configuration. When
	// the label value isn't empty, all Prometheus shards will scrape the target.
```

simonpasquier (commented):
you'll need to rebase on main to get the CI green.

bodji:
CI is failing with a lot of "429: Too Many Requests" :(

simonpasquier:
@bodji yes it's unrelated to your PR.

simonpasquier:
you'll need to rebase on top of the main branch once https://github.com/prometheus-operator/prometheus-operator/pull/7507 merges.

simonpasquier:
@bodji once you rebase the PR, I'll merge :)

simonpasquier:
some files need to be regenerated.

bodji:
> some files need to be regenerated.

Yes, I'm trying to do that,  but I encounter some errors : 
```
Error: load packages in root "/Users/mathieu/go/src/github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1": err: exit status 1: stderr: go: downloading go1.24 (darwin/arm64)
go: download go1.24 for darwin/arm64: toolchain not available

```

I try to fix

bodji:
```
error: 2 errors:
		Documentation/proposals/202305-scrapeclasses.md:194: "https://istio.io/latest/docs/reference/config/security/authorization-policy/" not accessible even after retry; status code 0: Get "https://istio.io/latest/docs/reference/config/security/authorization-policy/": http2: timeout awaiting response headers
		Documentation/proposals/202305-scrapeclasses.md:189: "https://istio.io/latest/docs/concepts/security/#permissive-mode" not accessible even after retry; status code 0: Get "https://istio.io/latest/docs/concepts/security/#permissive-mode": http2: timeout awaiting response headers
make: *** [Makefile:[341](https://github.com/prometheus-operator/prometheus-operator/actions/runs/14879170815/job/41782969731?pr=7421#step:5:342): check-docs] Error 1
```

bodji:
> thanks!

thank you for your help and review :)

PR #7439, opened 2025-04-04
feat: add `terminationGracePeriodSeconds` field

### LINKED ISSUE #3433: hope add terminationGracePeriodSeconds support

simonpasquier:
## Description

This change adds the `terminationGracePeriodSeconds` field to the Alertmanager, Prometheus, PrometheusAgent and ThanosRuler CRDs.

Closes #3433

## Type of change

## Verification

## Changelog entry

```release-note

```


simonpasquier:
cc @saswatamcode

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -309,6 +309,16 @@ type AlertmanagerSpec struct {
 	// by the given Alertmanager version.
 	// +optional
 	AdditionalArgs []Argument `json:"additionalArgs,omitempty"`
+
+	// Optional duration in seconds the pod needs to terminate gracefully.
+	// Value must be non-negative integer. The value zero indicates stop immediately via
+	// the kill signal (no opportunity to shut down) which may lead to data corruption.
+	//
+	// Defaults to 120 seconds.
```
slashpai, line 317: @simonpasquier  curious why we didn't add default in API validation but as a constant?

slashpai (commented):
lgtm just had a question for my understanding :)

pkg/apis/monitoring/v1/alertmanager_types.go
simonpasquier: In general, we don't use default in the API definition so I wanted to stay consistent. Default values are returned in the response of the Kubernetes API which may cause issues with some clients (e.g. they might interpret the value being present as a need for another update).

slashpai: We do have `// +kubebuilder:default:` set for many fields do you think we should remove it then for them?

simonpasquier: I don't think that we can remove any of them for compatibility reasons but I'd rather not add more :)

slashpai (approved):

PR #7448, opened 2025-04-09
feat: add `convertClassicHistogramsToNHCB` to scrape CRDs

smuth4:
## Description

This adds the ability to generate the `convert_classic_histograms_to_nhcb` option for the various scrape configs/monitors. Requires prometheus 3.0.0.

Most of the hard work has already been done in https://github.com/prometheus-operator/prometheus-operator/pull/6796, this just adds the final new option. While testing, I  found that `scripts/run-external.sh` didn't seem to be set up for contexts where the username/cluster might be named differently, so I updated that script as well.

Note that a PR was recently merged to add the same option to the top-level config: https://github.com/prometheus/prometheus/pull/16226, but since that hasn't made it into any releases yet I explicitly did not try to add support for it.

## Type of change

## Verification

I updated the unit tests to use v3.0.0 for the golden, and also tested on a live v3.2.1 prometheus instance with `scripts/run-external.sh` against an existing `nginx_ingress_controller_request_duration_seconds` classic histogram.

## Changelog entry

```release-note
- Add `convertClassicHistogramsToNHCB` to the scrape CRDs for enabling the conversion of classic histograms to native histograms with custom buckets.
```


simonpasquier (approved):
Thanks!

Issue #6528, opened 2024-04-18
Create end-to-end test to verify Prometheus operator metrics


simonpasquier:
This is a follow-up of #6513 and #6525.

To avoid future regression on key operator metrics disappearing when we update dependencies, we need to add an end-to-end test.

Key metrics involve at least the ones used in the alerts mixin the but others can be included.

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/issues/6525#issuecomment-2063462256_

afzal442:
Hey @simonpasquier this should go into metrics_test.go new file to add an end-to-end test . Can we take an example of it https://github.com/kubernetes-sigs/controller-runtime/blob/67b27f27e514bd9ac4cf9a2d84dec089ece95bf7/pkg/metrics/client_go_adapter.go#L30-L55 to create test cases for prom-op metrics?

Mberga14:
@simonpasquier attempted to add coverage for prometheus operator metrics in #7488

PR #7488, opened 2025-04-29
e2e: test for all prometheus-operator metrics in e2e

Mberga14:
fixes: #6528

## Description

Adding all metrics that prometheus-operator registers to the e2e coverage

## Type of change

## Verification

## Changelog entry

```release-note
test for presence of all prometheus-operator metrics in e2e test
```


dongjiang1989:
Needs https://github.com/prometheus-operator/prometheus-operator/pull/7490

simonpasquier (commented):
Thanks for the PR! I'd be fine with updating the list of metrics in `testPrometheusOperatorMetrics` instead of creating new test cases.

Mberga14:
@simonpasquier getting back to this - I am not exactly sure what you mean by your comment. 

To test all metrics being present the test needs to be invoked at different times (while CRs are present and at end of e2e tests). Or do you mean just having the same `testPrometheusOperatorMetrics` case (just renaming the additional calls I make at different points of the tests)?

simonpasquier:
> Or do you mean just having the same testPrometheusOperatorMetrics case (just renaming the additional calls I make at different points of the tests)?

This. I meant to have only one e2e test covering all metrics. If the test needs to create a custom resource for the metrics to be present, it is fine.

Mberga14:
@simonpasquier i amended the test so its now one test that checks for all metrics 👍

pkg/prometheus/collector.go
```diff
@@ -84,7 +85,6 @@ func (c *Collector) collectPrometheus(ch chan<- prometheus.Metric, p v1.Promethe
 	if cpf.EnforcedSampleLimit != nil {
 		ch <- prometheus.MustNewConstMetric(descPrometheusEnforcedSampleLimit, prometheus.GaugeValue, float64(*cpf.EnforcedSampleLimit), namespace, name)
 	}
-	if cpf.Shards != nil {
-		ch <- prometheus.MustNewConstMetric(descPrometheusSpecShards, prometheus.GaugeValue, float64(*cpf.Shards), namespace, name)
-	}
+
+	ch <- prometheus.MustNewConstMetric(descPrometheusSpecShards, prometheus.GaugeValue, float64(ptr.Deref(cpf.Shards, 1)), namespace, name)
```
Mberga14, line 89: @simonpasquier interesting side note - the `prometheus_operator_spec_shards` metric was only present when the number of shards is set on the CRD - the metric should exist for all values I think.

simonpasquier: good catch!

test/e2e/prometheus_operator_test.go
```diff
@@ -40,25 +40,81 @@ func testServerTLS(t *testing.T, namespace string) {
...
+		"https",
+		namespace,
+		prometheusOperatorServiceName,
+		"https",
+		append(operatorMetrics, operatorOperationalMetrics...)...,
 	)
+
 	require.NoError(t, err)
+
+	if err := framework.DeletePrometheusAndWaitUntilGone(context.Background(), ns, name); err != nil {
```
simonpasquier, line 117: The namespace should be deleted at the end of the test (via `defer`) so it's fine to rely on the namespace finalizer to clean up the Prometheus resource.

test/e2e/prometheus_operator_test.go
```diff
@@ -40,25 +40,81 @@ func testServerTLS(t *testing.T, namespace string) {
...
+		t.Fatal(err)
+	}
+
+	err = framework.EnsureMetricsFromService(
+		ctx,
+		"https",
+		namespace,
+		prometheusOperatorServiceName,
+		"https",
+		append(operatorMetrics, operatorOperationalMetrics...)...,
```
simonpasquier, line 112: (nit) since we already check `operatorMetrics` before we can just check resource-dependent metrics here.
```suggestion
		operatorOperationalMetrics...,
```

simonpasquier (approved):
Thanks!

PR #7512, opened 2025-05-08
fix(alertmanager): cluster.peer with custom service

denysvitali:
## Description

When a custom service is used (`spec.serviceName`) and AlertManager is used in HA mode - the cluster peering fails because the cluster assumes the service name is `alertmanager-operated`.

This results in:

```
Failed to resolve my-alertmanager-0.alertmanager-operated:9094: lookup my-alertmanager-0.alertmanager-operated on 10.33.0.10:53: no such host
```

With a service name of `foo`, the hostname should be:

```
my-alertmanager-0.foo:9094
```

## Type of change

## Verification

N/A

 

## Changelog entry

```release-note
- Fix AlertManager's `cluster.peer` argument when a custom service is used
```


simonpasquier:
Thanks for the PR! It might warrant a v0.82.2 patch release. Can you please make the change against the release-0.82 branch?
It would also be good to extend the `testAlertManagerServiceName` test to ensure that when setting a custom service name, the gossip cluster is working as expected (but it's more involved and it can be tracked in a follow-up issue/PR).

denysvitali:
Sure. Will do now!

I'll also try to add more tests for this

denysvitali:
Added a test and rebased on top of `release-0.82` :rocket:

simonpasquier (approved):
Thanks!

PR #7548, opened 2025-05-20
docs: update ingress monitoring documentation to use ingress-nginx instead of Ambassador

thaniri:
## Description

The previous docs referenced the Ambassador project which is not available any longer and does not use a license that the prometheus-operator developers want to promote. This updates the doc to use the popular ingress-nginx Ingress Controller which is one of the officially supported Ingress Controllers by the Kubernetes project.

## Type of change

## Verification

This is a documentation change, not a code change.

However I tested it by:

1. Starting a local Kind cluster
2. Following the steps in this guide (default kube-prometheus-stack and ingress-nginx helm chart installations)
3. Verification that monitoring works.

## Changelog entry

```release-note
Updated ingress monitoring documentation to use ingress-nginx instead of Ambassador.
```


Documentation/user-guides/monitoring-kubernetes-ingress.md
```diff
@@ -1,363 +1,115 @@
...
+> By default, the prometheus-operator will be installed in the default namespace.
+> You may need to update the spec.serviceMonitorSelector.matchLabels in order to correctly monitor ingress-nginx.
 
-```sh
-kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/master/bundle.yaml
-```
+## Configure ingress-nginx Monitoring
 
-Once the Prometheus operator is running, we need to create a Prometheus instance. The Prometheus Operator manages Prometheus deployments with the `Prometheus` CRD. To create a Prometheus instance and Kubernetes service, copy the following YAML to a file called `prometheus.yaml` and deploy it with `kubectl`:
+The ingress-nginx Service does not expose the Prometheus monitoring port by default. We need to first ensure the Service exposes the port for Prometheus metrics and then create a ServiceMonitor to tell Prometheus to scrape those metrics.
```
heliapb, line 41: Hi @thaniri, a nit:

```suggestion
After deploying Prometheus, we need to expose ingress-nginx Service to scrape metrics, since it doesn't create that Service by default. Afterwards, we need to create a ServiceMonitor to tell Prometheus to scrape those metrics.
```

thaniri: Thank you for the review, @heliapb 🎉 

I do not agree with your suggested change, as in this guide I patched the default Service provided by the `ingress-nginx` chart as opposed to creating a new one. Unless there is a prometheus-operator best practice that I am not aware of which recommends creating a new Service for monitoring purposes.

The key difference between my sentence and yours is found here:

```
The ingress-nginx **Service does not expose the Prometheus monitoring port by default**. We need to first **ensure the Service exposes the port** ...
```

Implying a Service _does_ exist, but doesn't expose the port we want.

Whereas yours:

```
After deploying Prometheus, we need to expose ingress-nginx Service to scrape metrics, **since it doesn't create that Service** ...
```

Implies we must create a new Service.

As we don't need to create a new Service, mine reads more correctly. Though, yours would still work! (With a minor change to the guide as well to not include the Service patch.)

simonpasquier (commented):
Thanks for starting this! I realize that https://kubernetes.github.io/ingress-nginx/user-guide/monitoring/#prometheus-and-grafana-installation-using-service-monitors already exists and provide similar instructions. Is there a need to describe again all the deployment steps here?

The previous documentation on Ambassador also had an interesting part where it described how to expose the Prometheus service via ingress (especially the externalUrl and routePrefix fields).

thaniri:
> Thanks for starting this! I realize that https://kubernetes.github.io/ingress-nginx/user-guide/monitoring/#prometheus-and-grafana-installation-using-service-monitors already exists and provide similar instructions. Is there a need to describe again all the deployment steps here?
> 
> The previous documentation on Ambassador also had an interesting part where it described how to expose the Prometheus service via ingress (especially the externalUrl and routePrefix fields).

To be honest, deleting this document entirely and allowing users to achieve this task by following the ingress-nginx docs would be better than having similar docs here. The docs here will get out of date as prometheus-operator developers are not in the business of keeping up with ingress-nginx developments.

I made this PR as an attempt to see how well an LLM could do a task like this. I just used the out-of-the-box agent mode in Cursor and it was... bad. So I fixed it up :)

@simonpasquier if you agree, I can simply delete this document and that will be one less thing to need to maintain :tada: . Even the steps to expose Prometheus via Ingress can be explained in ingess-nginx documentation.

simonpasquier:
@thaniri thanks for update! I agree that we can simply delete that page: less code makes maintainers happy (it doesn't even appear on https://prometheus-operator.dev/docs/).
Also we have documentation to expose Prometheus via ingress (https://prometheus-operator.dev/docs/platform/exposing-prometheus-and-alertmanager/).

thaniri:
Thanks @simonpasquier . Deleting things is one of the greatest joys of software development work.

Thank you for your careful consideration as a maintainer of prometheus-operator.

I've updated this PR to simply delete this document.

simonpasquier (approved):
LGTM!

PR #7516, opened 2025-05-09
fix: Fix parsing `role`  field in openstackSDConfigs in ScrapeConfig CRD

slashpai:
Fixes #7514

## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
If it fixes a bug or resolves a feature request, be sure to link to that issue._

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [x] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification

Please check the [Prometheus-Operator testing guidelines](../TESTING.md) for recommendations about automated tests.

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

Issue #7514, opened 2025-05-09
Scraped_config CSD not compatible with prometheus openstack_sd_config


rpo-fr:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Description
scraped config CRD need to define openstack role to   "Hypervisor" /  "Instance" / "Loadbalancer"

but prometheus need to be  "hypervisor" /  "instance" / "loadbalancer"
[https://github.com/prometheus/prometheus/blob/main/discovery/openstack/openstack.go](https://github.com/prometheus/prometheus/blob/main/discovery/openstack/openstack.go)

## Steps to Reproduce
Define a scraped_config

## Expected Result
Prometheus config reloaded, openstack target present

## Actual Result
Error message in prometheus log: "unknown OpenStack SD role \"Instance\"" 

### Prometheus Operator Version

```shell
kube-prometheus-stack:72.2.0
```

### Kubernetes Version

```yaml
1.31
```

### Kubernetes Cluster Type

Rancher

### How did you deploy Prometheus-Operator?

helm chart:prometheus-community/kube-prometheus-stack

### Manifests

```yaml

```

### prometheus-operator log output

```shell
ts=2025-05-09T14:07:09.033049447Z level=info caller=/workspace/pkg/prometheus/server/operator.go:768 msg="sync prometheus" component=prometheus-controller key=monitoring/kube-prometheus-stack-prometheus
```

rpo-fr:
change made by => https://github.com/prometheus-operator/prometheus-operator/pull/7370

slashpai:
Thanks for reporting the issue should be fixed in #7516 and available in patch release v0.82.2

PR #7559, opened 2025-05-23
fix: allow lower case role field in openstackSDConfigs in ScrapeConfig CRD

david-vavra:
## Description
This is a follow-up for https://github.com/prometheus-operator/prometheus-operator/issues/7514, and its fix - https://github.com/prometheus-operator/prometheus-operator/pull/7516/files. The fix is valid, but current CRD requires (`Instance`) the exact opposite of what Prometheus requires (`instance`).

I think that CRD should support both cases - lower case (`instance`) as the one with Capital letter (`Instance`). If only one case should be supported, imho it should not be the opposite of what Prometheus requires (as it is now). :)

## Type of change

## Verification

## Changelog entry

* [BUGFIX] Allow lowercase `role` field in openstackSDConfigs in ScrapeConfig CRD.

```release-note

```

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -57591,8 +57591,11 @@ spec:
                         Note: The `LoadBalancer` role requires Prometheus >= v3.2.0.
                       enum:
                       - Instance
+                      - instance
                       - Hypervisor
+                      - hypervisor
                       - LoadBalancer
+                      - loadbalancer
                       type: string
                     tlsConfig:
                       description: TLS configuration applying to the target HTTP endpoint.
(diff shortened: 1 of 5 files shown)


slashpai:
This was intentionally removed in https://github.com/prometheus-operator/prometheus-operator/pull/7370 to follow Kubernetes conventions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants which missed the lowercase conversion in code later fixed in bugfix you linked 🙂

david-vavra:
@slashpai, thx for adding the context. Closing this.

Issue #7583, opened 2025-06-02
finalizer in workload resources when configForStatusResource is enabled


yp969803:
### What is missing? Please describe.

finalizer in workload resources when configForStatusResource is enabled

## Environment
Kubernetes Version:
Prometheus-Operator Version:

PR #7584, opened 2025-06-02
finalizer added to workload resources if statusFOrConfigRes flag is enabled

yp969803:
## Description

fixes: #7583 

## Type of change

## Verification

## Changelog entry

```finalizer added to workload resources if statusFOrConfigRes flag is enabled

```

### DIFF
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -17,10 +17,12 @@ package alertmanager
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
 	"path"
+	"slices"
 	"strings"
 	"time"
 
@@ -34,6 +36,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	authv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/client-go/metadata"
@@ -59,6 +62,7 @@ import (
 const (
 	resyncPeriod   = 5 * time.Minute
 	controllerName = "alertmanager-controller"
+	finalizerName  = "monitoring.coreos.com/status-cleanup"
 )
 
 // Config defines the operator's parameters for the Alertmanager controller.
@@ -535,8 +539,50 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("failed to set Alertmanager type information: %w", err)
 	}
 
+	if c.configResourcesStatusEnabled {
+		// Add finalizer to the Prometheus resource if it doesn't have one.
+		finalizers := am.GetFinalizers()
+		if !slices.Contains(finalizers, finalizerName) {
+			finalizers = append(finalizers, finalizerName)
+			patchData := map[string]interface{}{
+				"metadata": map[string]interface{}{
+					"finalizers": finalizers,
+				},
+			}
+			patchBytes, err := json.Marshal(patchData)
+			if err != nil {
+				return fmt.Errorf("failed to marshal patch: %w", err)
+			}
+			if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+				return fmt.Errorf("failed to add %q finalizer: %w", finalizerName, err)
+			}
+		}
+	}
+
 	// Check if the Alertmanager instance is marked for deletion.
 	if c.rr.DeletionInProgress(am) {
+		if c.configResourcesStatusEnabled {
+			// If the Alertmanger instance is marked for deletion, we remove the finalizer.
+			finalizers := am.GetFinalizers()
+			finalizers = slices.DeleteFunc(finalizers, func(f string) bool {
+				return f == finalizerName
+			})
+			patchData := map[string]interface{}{
+				"metadata": map[string]interface{}{
+					"finalizers": finalizers,
+				},
+			}
+			patchBytes, err := json.Marshal(patchData)
+			if err != nil {
+				return fmt.Errorf("failed to marshal patch: %w", err)
+			}
+			if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+				return fmt.Errorf("failed to add %q finalizer: %w", finalizerName, err)
+			}
+
+			c.reconciliations.ForgetObject(key)
+			return nil
+		}
 		return nil
 	}
 
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -34,7 +34,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	authv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/client-go/metadata"
@@ -544,16 +543,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 
 	logger.Info("sync alertmanager")
 
-	if !c.configResourcesStatusEnabled && c.rr.DeletionInProgress(am) {
-		return nil
-	}
-
-	finalizersChanged, err := c.syncFinalizers(ctx, am, key)
-	if err != nil {
-		return err
-	}
-	if finalizersChanged {
-		c.rr.EnqueueForReconciliation(am)
+	if c.rr.DeletionInProgress(am) {
 		return nil
 	}
 
@@ -662,51 +652,6 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	}
 
 	return nil
-}
-
-// add or remove finalizers for the Alertmanager resource.
-// Returns true if the finalizers were modified, otherwise false. The second return value is an error, if any.
-func (c *Operator) syncFinalizers(ctx context.Context, am *monitoringv1.Alertmanager, key string) (bool, error) {
-	if !c.configResourcesStatusEnabled {
-		return false, nil
-	}
-
-	// The resource isn't being deleted, add the finalizer if missing.
-	if !c.rr.DeletionInProgress(am) {
-		// Add finalizer to the Alertmanger resource if it doesn't have one.
-		finalizers := am.GetFinalizers()
-		patchBytes, err := k8sutil.FinalizerAddPatch(finalizers, k8sutil.StatusCleanupFinalizerName)
-		if err != nil {
-			return false, fmt.Errorf("failed to marshal patch: %w", err)
-		}
-		if len(patchBytes) == 0 {
-			return false, nil
-		}
-		if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
-			return false, fmt.Errorf("failed to add %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
-		}
-
-		c.logger.Debug("added finalizer to Alertmanager", "name", am.Name, "namespace", am.Namespace, "finalizer", k8sutil.StatusCleanupFinalizerName)
-		return true, nil
-	}
-	// If the Alertmanager instance is marked for deletion, we remove the finalizer.
-	finalizers := am.GetFinalizers()
-	patchBytes, err := k8sutil.FinalizerDeletePatch(finalizers, k8sutil.StatusCleanupFinalizerName)
-	if err != nil {
-		return false, fmt.Errorf("failed to marshal patch: %w", err)
-	}
-	if len(patchBytes) == 0 {
-		c.reconciliations.ForgetObject(key)
-		return false, nil
-	}
-
-	if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
-		return false, fmt.Errorf("failed to remove %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
-	}
-
-	c.logger.Debug("removed finalizer from Alertmanager", "name", am.Name, "namespace", am.Namespace, "finalizer", k8sutil.StatusCleanupFinalizerName)
-	c.reconciliations.ForgetObject(key)
-	return true, nil
 }
 
 // getAlertmanagerFromKey returns a copy of the Alertmanager object identified by key.
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -539,7 +539,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("failed to set Alertmanager type information: %w", err)
 	}
 
-	if c.configResourcesStatusEnabled {
+	if c.configResourcesStatusEnabled && !c.rr.DeletionInProgress(am) {
 		// Add finalizer to the Prometheus resource if it doesn't have one.
 		finalizers := am.GetFinalizers()
 		if !slices.Contains(finalizers, finalizerName) {
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -553,6 +553,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return err
 	}
 	if finalizersChanged {
+		c.rr.EnqueueForReconciliation(am)
 		return nil
 	}
 
@@ -684,6 +685,8 @@ func (c *Operator) syncFinalizers(ctx context.Context, am *monitoringv1.Alertman
 		if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
 			return false, fmt.Errorf("failed to add %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
 		}
+
+		c.logger.Debug("added finalizer to Alertmanager", "name", am.Name, "namespace", am.Namespace, "finalizer", k8sutil.StatusCleanupFinalizerName)
 		return true, nil
 	}
 	// If the Alertmanager instance is marked for deletion, we remove the finalizer.
@@ -701,6 +704,7 @@ func (c *Operator) syncFinalizers(ctx context.Context, am *monitoringv1.Alertman
 		return false, fmt.Errorf("failed to remove %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
 	}
 
+	c.logger.Debug("removed finalizer from Alertmanager", "name", am.Name, "namespace", am.Namespace, "finalizer", k8sutil.StatusCleanupFinalizerName)
 	c.reconciliations.ForgetObject(key)
 	return true, nil
 }
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -104,6 +104,8 @@ type Operator struct {
 	canReadStorageClass bool
 
 	config Config
+
+	configResourcesStatusEnabled bool
 }
 
 type ControllerOption func(*Operator)
@@ -161,6 +163,7 @@ func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger
 			Annotations:                  c.Annotations,
 			Labels:                       c.Labels,
 		},
+		configResourcesStatusEnabled: c.Gates.Enabled(operator.StatusForConfigurationResourcesFeature),
 	}
 	for _, opt := range options {
 		opt(o)
@@ -178,6 +181,7 @@ func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger
 		monitoringv1.AlertmanagersKind,
 		r,
 		o.controllerID,
+		o.configResourcesStatusEnabled,
 	)
 
 	return o, nil
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -182,7 +182,6 @@ func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger
 		monitoringv1.AlertmanagersKind,
 		r,
 		o.controllerID,
-		o.configResourcesStatusEnabled,
 	)
 
 	return o, nil
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -669,38 +669,40 @@ func (c *Operator) syncFinalizers(ctx context.Context, am *monitoringv1.Alertman
 	if !c.configResourcesStatusEnabled {
 		return false, nil
 	}
-	finalizersChanged := false
+
+	// The resource isn't being deleted, add the finalizer if missing.
 	if !c.rr.DeletionInProgress(am) {
 		// Add finalizer to the Alertmanger resource if it doesn't have one.
 		finalizers := am.GetFinalizers()
 		patchBytes, err := k8sutil.FinalizerAddPatch(finalizers, k8sutil.StatusCleanupFinalizerName)
 		if err != nil {
-			return finalizersChanged, fmt.Errorf("failed to marshal patch: %w", err)
+			return false, fmt.Errorf("failed to marshal patch: %w", err)
 		}
-		if len(patchBytes) > 0 {
-			if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
-				return finalizersChanged, fmt.Errorf("failed to add %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
-			}
-			finalizersChanged = true
+		if len(patchBytes) == 0 {
+			return false, nil
 		}
-		return finalizersChanged, nil
+		if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+			return false, fmt.Errorf("failed to add %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
+		}
+		return true, nil
 	}
 	// If the Alertmanager instance is marked for deletion, we remove the finalizer.
 	finalizers := am.GetFinalizers()
 	patchBytes, err := k8sutil.FinalizerDeletePatch(finalizers, k8sutil.StatusCleanupFinalizerName)
 	if err != nil {
-		return finalizersChanged, fmt.Errorf("failed to marshal patch: %w", err)
+		return false, fmt.Errorf("failed to marshal patch: %w", err)
 	}
-	if len(patchBytes) > 0 {
-		if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
-			return finalizersChanged, fmt.Errorf("failed to remove %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
-		}
-		finalizersChanged = true
+	if len(patchBytes) == 0 {
+		c.reconciliations.ForgetObject(key)
+		return false, nil
+	}
+
+	if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+		return false, fmt.Errorf("failed to remove %s finalizer: %w", k8sutil.StatusCleanupFinalizerName, err)
 	}
 
 	c.reconciliations.ForgetObject(key)
-
-	return finalizersChanged, nil
+	return true, nil
 }
 
 // getAlertmanagerFromKey returns a copy of the Alertmanager object identified by key.
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -697,7 +697,7 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 		return err
 	}
 
-	if c.rr.DeletionInProgress(a) {
+	if a == nil || c.rr.DeletionInProgress(a) {
 		return nil
 	}
 
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -17,12 +17,10 @@ package alertmanager
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
 	"path"
-	"slices"
 	"strings"
 	"time"
 
@@ -62,7 +60,6 @@ import (
 const (
 	resyncPeriod   = 5 * time.Minute
 	controllerName = "alertmanager-controller"
-	finalizerName  = "monitoring.coreos.com/status-cleanup"
 )
 
 // Config defines the operator's parameters for the Alertmanager controller.
@@ -540,21 +537,15 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	}
 
 	if c.configResourcesStatusEnabled && !c.rr.DeletionInProgress(am) {
-		// Add finalizer to the Prometheus resource if it doesn't have one.
+		// Add finalizer to the Alertmanager resource if it doesn't have one.
 		finalizers := am.GetFinalizers()
-		if !slices.Contains(finalizers, finalizerName) {
-			finalizers = append(finalizers, finalizerName)
-			patchData := map[string]interface{}{
-				"metadata": map[string]interface{}{
-					"finalizers": finalizers,
-				},
-			}
-			patchBytes, err := json.Marshal(patchData)
-			if err != nil {
-				return fmt.Errorf("failed to marshal patch: %w", err)
-			}
+		patchBytes, err := k8sutil.AddStatusCleanupFinalizer(finalizers)
+		if err != nil {
+			return fmt.Errorf("failed to marshal patch: %w", err)
+		}
+		if patchBytes != nil {
 			if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
-				return fmt.Errorf("failed to add %q finalizer: %w", finalizerName, err)
+				return fmt.Errorf("failed to add statusCleanup finalizer: %w", err)
 			}
 		}
 	}
@@ -564,20 +555,14 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		if c.configResourcesStatusEnabled {
 			// If the Alertmanger instance is marked for deletion, we remove the finalizer.
 			finalizers := am.GetFinalizers()
-			finalizers = slices.DeleteFunc(finalizers, func(f string) bool {
-				return f == finalizerName
-			})
-			patchData := map[string]interface{}{
-				"metadata": map[string]interface{}{
-					"finalizers": finalizers,
-				},
-			}
-			patchBytes, err := json.Marshal(patchData)
+			patchBytes, err := k8sutil.DeleteStatusCleanupFinalizer(finalizers)
 			if err != nil {
 				return fmt.Errorf("failed to marshal patch: %w", err)
 			}
-			if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
-				return fmt.Errorf("failed to add %q finalizer: %w", finalizerName, err)
+			if patchBytes != nil {
+				if _, err = c.mclient.MonitoringV1().Alertmanagers(am.Namespace).Patch(ctx, am.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+					return fmt.Errorf("failed to remove statusCleanup finalizer: %w", err)
+				}
 			}
 
 			c.reconciliations.ForgetObject(key)
--- a/pkg/alertmanager/operator.go
+++ b/pkg/alertmanager/operator.go
@@ -663,42 +663,44 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
-// add or remove finalizers for the Alertmanger resource.
+// add or remove finalizers for the Alertmanager resource.
+// Returns true if the finalizers were modified, otherwise false. The second return value is an error, if any.
 func (c *Operator) syncFinalizers(ctx context.Context, am *monitoringv1.Alertmanager, key string) (bool, error) {
-	if c.configResourcesStatusEnabled {
-		var finalizersChanged bool
-		if !c.rr.DeletionInProgress(am) {
-			// Add finalizer to the Alertmanger resource if it doesn't have one.
-			finalizers := am.GetFinalizers()
-			patchBytes, err := k8sutil.AddStatusCleanupFinalizer(finalizers)
-			if err != nil {
-				return false, fmt.Errorf("failed to marshal patch: %w", err)
+	if !c.configResourcesStatusEnabled {
+		return false, nil
+	}
+	finalizersChanged := false
+	if !c.rr.DeletionInProgress(am) {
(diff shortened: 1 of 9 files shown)


pkg/prometheus/server/operator.go
```diff
@@ -753,8 +755,34 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
 	c.logDeprecatedFields(logger, p)
 
+	if c.configResourcesStatusEnabled {
+		// Add finalizer to the Prometheus resource if it doesn't have one.
+		finalizers := p.GetFinalizers()
+		if !slices.Contains(finalizers, finalizerName) {
+			finalizers = append(finalizers, finalizerName)
+			p.SetFinalizers(finalizers)
+			if _, err := c.mclient.MonitoringV1().Prometheuses(p.Namespace).Update(ctx, p, metav1.UpdateOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+				return fmt.Errorf("adding finalizer %q to Prometheus %q failed: %w", finalizerName, p.Name, err)
```
simonpasquier: (nit) the resource name is already logged in the `key` attribute by the logger.

simonpasquier:
```suggestion
				return fmt.Errorf("failed to add %q finalizer: %w", finalizerName, err)
```

pkg/prometheus/server/operator.go
```diff
@@ -753,8 +755,34 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	logger := c.logger.With("key", key)
 	c.logDeprecatedFields(logger, p)
 
+	if c.configResourcesStatusEnabled {
+		// Add finalizer to the Prometheus resource if it doesn't have one.
+		finalizers := p.GetFinalizers()
+		if !slices.Contains(finalizers, finalizerName) {
+			finalizers = append(finalizers, finalizerName)
+			p.SetFinalizers(finalizers)
+			if _, err := c.mclient.MonitoringV1().Prometheuses(p.Namespace).Update(ctx, p, metav1.UpdateOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
```
simonpasquier: (suggestion) we can use Patch() instead of a full Update().

pkg/operator/resource_reconciler.go
```diff
@@ -336,10 +336,6 @@ func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
 		return
 	}
 
-	if rr.DeletionInProgress(mCur) {
```
simonpasquier, line 339: to avoid unwanted side-effect for users running without the feature gate, it would be good to change the current behavior iif the feature gate is enabled, meaning that we need to pass the information to the resource reconciler (I would advise for an option to the constructor).

pkg/prometheus/server/operator.go
```diff
@@ -753,8 +755,34 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
 	// Check if the Prometheus instance is marked for deletion.
 	if c.rr.DeletionInProgress(p) {
+		if c.configResourcesStatusEnabled {
+			// If the Prometheus instance is marked for deletion, we remove the finalizer.
+			finalizers := p.GetFinalizers()
+			finalizers = slices.DeleteFunc(finalizers, func(f string) bool {
+				return f == finalizerName
+			})
+			p.SetFinalizers(finalizers)
+			if _, err := c.mclient.MonitoringV1().Prometheuses(p.Namespace).Update(ctx, p, metav1.UpdateOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
```
simonpasquier: same here about Patch()

pkg/prometheus/server/operator.go
```diff
@@ -53,6 +54,7 @@ import (
 const (
 	resyncPeriod   = 5 * time.Minute
 	controllerName = "prometheus-controller"
+	finalizerName  = "monitoring.coreos.com/config-res-status-cleanup"
```
simonpasquier: (nit)
``suggestion
	finalizerName  = "monitoring.coreos.com/status-cleanup"
```

File: pkg/prometheus/server/operator.go (hunk 1 above)
Comment (yp969803) [reply pkg/prometheus/server/operator.go]: ok

**Comment by yp969803** [comment]
prometheus is deleted successfully but getting this log,  
```
ts=2025-06-03T17:16:25.43770307Z level=info caller=/home/yash/go/src/github.com/prometheus-operator/prometheus-operator/pkg/prometheus/server/operator.go:1059 msg="failed to apply prometheus status subresource, trying again without scale fields" component=prometheus-controller err="prometheuses.monitoring.coreos.com \"prometheus\" not found"
ts=2025-06-03T17:16:25.440609933Z level=error caller=/home/yash/go/src/github.com/prometheus-operator/prometheus-operator/pkg/operator/resource_reconciler.go:567 msg="Unhandled Error" logger=UnhandledError err="status \"default/prometheus\" failed: failed to apply prometheus status subresource: prometheuses.monitoring.coreos.com \"prometheus\" not found"
```

**Comment by yp969803** [comment]

File: pkg/alertmanager/operator.go [hunk 6]
```diff
@@ -531,8 +539,50 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("failed to set Alertmanager type information: %w", err)
 	}
 
+	if c.configResourcesStatusEnabled {
```
simonpasquier: I'd advise for sharding the code across the different controllers (easier to maintain).

pkg/prometheus/agent/operator.go
```diff
@@ -570,8 +577,50 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 
 	logger := c.logger.With("key", key)
 
-	// Check if the Agent instance is marked for deletion.
+	if c.configResourcesStatusEnabled {
+		// Add finalizer to the Prometheus resource if it doesn't have one.
```
simonpasquier: From https://pkg.go.dev/k8s.io/apimachinery@v0.33.1/pkg/apis/meta/v1#ObjectMeta

```
If the deletionTimestamp of the object is non-nil, entries
in this list can only be removed.
````

To avoid unwanted errors, we might want to check that deletion timestamp isn't set.

File: pkg/prometheus/agent/operator.go [hunk 8]
```diff
@@ -570,8 +577,50 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+			patchData := map[string]interface{}{
+				"metadata": map[string]interface{}{
+					"finalizers": finalizers,
+				},
+			}
+			patchBytes, err := json.Marshal(patchData)
+			if err != nil {
+				return fmt.Errorf("failed to marshal patch: %w", err)
+			}
+			if _, err = c.mclient.MonitoringV1alpha1().PrometheusAgents(p.Namespace).Patch(ctx, p.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
```
simonpasquier: I wonder if Apply() wouldn't be safer to use? To be more precise, can we tell Apply() to only append our finalizer to the list instead of replacing the whole list?

pkg/prometheus/agent/operator.go
yp969803: done

pkg/prometheus/agent/operator.go
yp969803: No we cannot tell Apply to only append at last we have to replace whole list. Using patch is more safer than apply

pkg/alertmanager/operator.go
yp969803: done

yp969803:
@simonpasquier do u want integration tests also ?? if yes, what should i do , 1 possiblity create and delete workload res enabling this feature gate

pkg/operator/resource_reconciler.go
simonpasquier: I can see another alternative where instead of passing the feature gate information to ResourceReconciler, it could check the presence of the finalizer in the object metadata. If it's present then it means that a reconciliation is required.

pkg/operator/resource_reconciler.go
```diff
@@ -407,7 +411,7 @@ func (rr *ResourceReconciler) onDaemonSetAdd(ds *appsv1.DaemonSet) {
 func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet) {
 	rr.logger.Debug("update handler", "resource", "statefulset", "old", old.ResourceVersion, "cur", cur.ResourceVersion)
 
-	if rr.DeletionInProgress(cur) {
+	if !rr.configResourcesStatusEnabled && rr.DeletionInProgress(cur) {
```
simonpasquier: I think that we can keep the code unchanged here since we're looking at the statefulset not the top-level resource?

pkg/operator/resource_reconciler.go
```diff
@@ -437,7 +441,7 @@ func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet)
 func (rr *ResourceReconciler) onDaemonSetUpdate(old, cur *appsv1.DaemonSet) {
 	rr.logger.Debug("update handler", "resource", "daemonset", "old", old.ResourceVersion, "cur", cur.ResourceVersion)
 
-	if rr.DeletionInProgress(cur) {
+	if !rr.configResourcesStatusEnabled && rr.DeletionInProgress(cur) {
```
simonpasquier: same here

pkg/prometheus/server/operator.go
```diff
@@ -753,8 +755,38 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		finalizers := p.GetFinalizers()
+		patchBytes, err := k8sutil.AddStatusCleanupFinalizer(finalizers)
+		if err != nil {
+			return fmt.Errorf("failed to marshal patch: %w", err)
+		}
+		if patchBytes != nil {
+			if _, err = c.mclient.MonitoringV1().Prometheuses(p.Namespace).Patch(ctx, p.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+				return fmt.Errorf("failed to add statusCleanup finalizer: %w", err)
+			}
+		}
```
simonpasquier: A controller doesn't need to converge immediately, it might simpler to return here since it should be a rare operation (the finalizer has to be added only once in the object lifecycle).

Another suggestion would be to extract the logic into its own method for better readability

```golang
stop, err := reconcileFinalizers(p)
if err != nil {
    return err
}
if stop {
    // stop is true when the feature gate is on and the finalizer has been added/removed.
    return nil
}

...
```

pkg/prometheus/server/operator.go
```diff
@@ -753,8 +755,38 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
 
+	if c.configResourcesStatusEnabled && !c.rr.DeletionInProgress(p) {
+		// Add finalizer to the Prometheus resource if it doesn't have one.
+		finalizers := p.GetFinalizers()
+		patchBytes, err := k8sutil.AddStatusCleanupFinalizer(finalizers)
+		if err != nil {
+			return fmt.Errorf("failed to marshal patch: %w", err)
+		}
+		if patchBytes != nil {
+			if _, err = c.mclient.MonitoringV1().Prometheuses(p.Namespace).Patch(ctx, p.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
```
simonpasquier: maybe we can use a JSON patch to be more chirurgical and remove/add our specific element. e.g.

```json
[{"op":"add","path":"/metadata/finalizers/-","value":"example.com/baz"}]
```

https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment

pkg/prometheus/server/operator.go
yp969803: why do you want to return nil, we have to process other things such creating statefull or daemon sets ?? If we return nil, when will the reconciler will again trigger ??

yp969803: got it, you are right we should not coverge immediately

yp969803:

pkg/prometheus/server/operator.go
yp969803: docs say that
```
A strategic merge patch is different from a [JSON merge patch](https://tools.ietf.org/html/rfc7386). With a JSON merge patch, if you want to update a list, you have to specify the entire new list. And the new list completely replaces the existing list
```

yp969803:
```
ts=2025-06-06T10:10:50.097743559Z level=error caller=/home/yash/go/src/github.com/prometheus-operator/prometheus-operator/pkg/operator/resource_reconciler.go:560 msg="Unhandled Error" logger=UnhandledError err="sync \"default/prometheus\" failed: failed to add statusCleanup finalizer: the server rejected our request due to an error in our request"
```
getting this error while adding finalizer using json patch, i think this only works if our finalizer list in not null, means the finalizer list should contain some element

yp969803:
i am shifing back to json merge patch, as it do not give any error, and a safer method

simonpasquier:
> getting this error while adding finalizer using json patch, i think this only works if our finalizer list in not null, means the finalizer list should contain some element

can we then set the array to `[<finalizer>]` instead of adding to it?

simonpasquier:
> @simonpasquier do u want integration tests also ?? if yes, what should i do , 1 possiblity create and delete workload res enabling this feature gate

Yes please! Indeed creating and deleting a Prometheus resource should be good enough to start. You can look at existing tests because we have lots of utility functions that abstract all the complexity.

Tests for feature gates go to the `TestGatedFeatures()` function. I'd advise to put your tests into a new file (e.g. `test/e2e/status_subresource_test.go`).

pkg/operator/resource_reconciler.go
```diff
@@ -336,7 +339,7 @@ func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
 		return
 	}
 
-	if rr.DeletionInProgress(mCur) {
+	if !k8sutil.IsStatusCleanupFinalizerPresent(mCur.GetFinalizers()) && rr.DeletionInProgress(mCur) {
```
simonpasquier: what about updating DeletionInProgress() to check for the presence of the finalizer? It would simplify the change.

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,44 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
+// add or remove finalizers for the Prometheus resource.
+func (c *Operator) syncFinalizers(ctx context.Context, p *monitoringv1.Prometheus, key string) (bool, error) {
+	if c.configResourcesStatusEnabled {
```
simonpasquier: (suggestion) a good practice is to check first for the exit condition(s).
```suggestion
	if !c.configResourcesStatusEnabled {
		return false, nil
	}
```

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,44 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
+// add or remove finalizers for the Prometheus resource.
```
simonpasquier: Can we update the comment to explain what the first return value means?

pkg/prometheus/server/operator.go
```diff
@@ -753,8 +754,15 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	logger := c.logger.With("key", key)
 	c.logDeprecatedFields(logger, p)
 
-	// Check if the Prometheus instance is marked for deletion.
-	if c.rr.DeletionInProgress(p) {
+	if !c.configResourcesStatusEnabled && c.rr.DeletionInProgress(p) {
+		return nil
+	}
+
+	finalizersChanged, err := c.syncFinalizers(ctx, p, key)
```
simonpasquier, line 757: (nit)
```suggestion
	stop, err := c.syncFinalizers(ctx, p, key)
```

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,44 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+			patchBytes, err := k8sutil.AddStatusCleanupFinalizer(finalizers)
+			if err != nil {
+				return false, fmt.Errorf("failed to marshal patch: %w", err)
+			}
+			if patchBytes != nil {
+				if _, err = c.mclient.MonitoringV1().Prometheuses(p.Namespace).Patch(ctx, p.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
+					return false, fmt.Errorf("failed to add statusCleanup finalizer: %w", err)
+				}
+				finalizersChanged = true
+			}
```
simonpasquier: (suggestion) same here, by returning at the end of the if branch, we can avoid the else and make the code easier to read.
```suggestion
			}
			
			return true, nil
```

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,34 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
 	}
 	return nil
 }
+
+func AddStatusCleanupFinalizer(finalizers []string) ([]byte, error) {
```
simonpasquier: (nit) the purpose of k8sutils is to be k8s oriented so passing a metav1.Object feels more natural.
```suggestion
// AddFinalizerPatch generates the JSON patch payload which adds the finalizer to the object's metadata.
// If the finalizer is already present, it returns an empty []byte slice. 
func FinalizerAddPatch(o metav1.Object) ([]byte, error) {
```

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,34 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
...
+	if !slices.Contains(finalizers, statusCleanupFinalizerName) {
```
simonpasquier:
```suggestion
	if slices.Contains(finalizers, statusCleanupFinalizerName) {
		return nil, nil
	}
	
```

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,34 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
...
+				"finalizers": finalizers,
+			},
+		}
+		patchBytes, err := json.Marshal(patchData)
+		return patchBytes, err
+	}
+	return nil, nil
+}
+
+func DeleteStatusCleanupFinalizer(finalizers []string) ([]byte, error) {
```
simonpasquier: same remarks here

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,34 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
...
+func AddStatusCleanupFinalizer(finalizers []string) ([]byte, error) {
+	if !slices.Contains(finalizers, statusCleanupFinalizerName) {
+		finalizers = append(finalizers, statusCleanupFinalizerName)
+		patchData := map[string]interface{}{
+			"metadata": map[string]interface{}{
+				"finalizers": finalizers,
+			},
+		}
+		patchBytes, err := json.Marshal(patchData)
+		return patchBytes, err
```
simonpasquier: (nit)
```suggestion
		return json.Marshal(patchData)
```

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,44 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	if c.configResourcesStatusEnabled {
+		var finalizersChanged bool
+		if !c.rr.DeletionInProgress(p) {
+			// Add finalizer to the Prometheus resource if it doesn't have one.
+			finalizers := p.GetFinalizers()
+			patchBytes, err := k8sutil.AddStatusCleanupFinalizer(finalizers)
+			if err != nil {
+				return false, fmt.Errorf("failed to marshal patch: %w", err)
+			}
+			if patchBytes != nil {
```
simonpasquier: (nit) we usually check on the length of the byte slice.
```suggestion
			if len(patchBytes) > 0 {
```

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,44 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+				if _, err = c.mclient.MonitoringV1().Prometheuses(p.Namespace).Patch(ctx, p.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
```
simonpasquier: (suggestion for later) we might even be able to abstract further and write code which could work for virtually any CRD.

pkg/prometheus/server/operator.go
```diff
@@ -995,6 +1041,9 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	p := pobj.(*monitoringv1.Prometheus)
 	p = p.DeepCopy()
 
+	if p == nil || c.rr.DeletionInProgress(p) {
```
simonpasquier: i suppose that you've hit a race condition? Should we be more defensive in DeletionInProgress (e.g. if p is nil then it would return true)? If yes, the fix could deserve its own PR :)

yp969803:

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,49 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
 	}
 	return nil
 }
+
+// AddFinalizerPatch generates the JSON patch payload which adds the finalizer to the object's metadata.
+// If the finalizer is already present, it returns empty []byte slice.
+func FinalizerAddPatch(finalizers []string, finalizerName string) ([]byte, error) {
+	if slices.Contains(finalizers, finalizerName) {
+		return []byte{}, nil
+	} else if len(finalizers) == 0 {
```
simonpasquier:
```suggestion
	}
	
	if len(finalizers) == 0 {
```

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,46 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
 }
 
+// add or remove finalizers for the Prometheus resource.
+// Returns true if the finalizers were modified, otherwise false. The second return value is an error, if any.
+func (c *Operator) syncFinalizers(ctx context.Context, p *monitoringv1.Prometheus, key string) (bool, error) {
+	if !c.configResourcesStatusEnabled {
+		return false, nil
+	}
+	finalizersChanged := false
+	if !c.rr.DeletionInProgress(p) {
```
simonpasquier: (nit)
```suggestion
	var finalizersChanged bool
	
	// The resource isn't being deleted, add the finalizer if missing.
	if !c.rr.DeletionInProgress(p) {
```

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,46 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		// Add finalizer to the Prometheus resource if it doesn't have one.
+		finalizers := p.GetFinalizers()
+		patchBytes, err := k8sutil.FinalizerAddPatch(finalizers, k8sutil.StatusCleanupFinalizerName)
+		if err != nil {
+			return finalizersChanged, fmt.Errorf("failed to marshal patch: %w", err)
+		}
+		if len(patchBytes) > 0 {
```
simonpasquier: (nit) It reduces the number of nested levels.
```suggestion
		if len(patchBytes) == 0 {
			return false, nil
		}
```

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +972,46 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		}
+		return finalizersChanged, nil
+	}
+	// If the Prometheus instance is marked for deletion, we remove the finalizer.
+	finalizers := p.GetFinalizers()
+	patchBytes, err := k8sutil.FinalizerDeletePatch(finalizers, k8sutil.StatusCleanupFinalizerName)
+	if err != nil {
+		return finalizersChanged, fmt.Errorf("failed to marshal patch: %w", err)
+	}
+	if len(patchBytes) > 0 {
```
simonpasquier: same remark here

yp969803:
@simonpasquier done the changes, can u look at the written e2e test, dont know why it is failing, everything is working fine in local

yp969803:
getting this error in the test
```
Error Trace:	/home/runner/work/prometheus-operator/prometheus-operator/test/e2e/status_subresource_test.go:51
        	Error:      	Received unexpected error:
        	            	waiting for 1 Prometheus instances timed out (status-cleanup-finalizer-test): prometheus gatedfeatures-statusforconfigresources-sxno9v-1-37052[93](https://github.com/prometheus-operator/prometheus-operator/actions/runs/15568672743/job/43838952550?pr=7584#step:4:94)d/status-cleanup-finalizer-test failed to become available: client rate limiter Wait returned an error: context deadline exceeded: context deadline exceeded
```

test/e2e/status_subresource_test.go
```diff
@@ -0,0 +1,57 @@
...
+			AllowedNamespaces:   []string{ns},
+			EnabledFeatureGates: []operator.FeatureGateName{operator.StatusForConfigurationResourcesFeature},
+		},
+	)
+	require.NoError(t, err)
+
+	name := "status-cleanup-finalizer-test"
+	t.Run(name, func(t *testing.T) {
+		p := framework.MakeBasicPrometheus(ns, name, name, 1)
+		framework.SetupPrometheusRBAC(ctx, t, testCtx, ns)
```
simonpasquier: do we need to call SetupPrometheusRBAC() again?

test/e2e/status_subresource_test.go
```diff
@@ -0,0 +1,57 @@
...
+		pm, err := framework.CreatePrometheusAndWaitUntilReady(ctx, ns, p)
+		require.NoError(t, err, "failed to create Prometheus")
+		finalizers := pm.GetFinalizers()
+		require.Contains(t, finalizers, k8sutil.StatusCleanupFinalizerName, "expected Prometheus to have status-cleanup finalizer")
```
simonpasquier: (nit) we can only check that the finalizer list is empty after callling MakeBasicPrometheus() and it isn't empty here.

pkg/prometheus/agent/operator.go
```diff
@@ -630,6 +640,52 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		finalizers := p.GetFinalizers()
+		patchBytes, err := k8sutil.FinalizerAddPatch(finalizers, k8sutil.StatusCleanupFinalizerName)
+		if err != nil {
+			return false, fmt.Errorf("failed to marshal patch: %w", err)
+		}
+
+		if len(patchBytes) == 0 {
+			return false, nil
+		}
+
```
simonpasquier: we can have a (debug) log message saying that we're adding the finalizer.

pkg/prometheus/agent/operator.go
```diff
@@ -630,6 +640,52 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to marshal patch: %w", err)
+	}
+
+	if len(patchBytes) == 0 {
+		c.reconciliations.ForgetObject(key)
+		return false, nil
+	}
+
+	if _, err = c.mclient.MonitoringV1alpha1().PrometheusAgents(p.Namespace).Patch(ctx, p.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{FieldManager: operator.PrometheusOperatorFieldManager}); err != nil {
```
simonpasquier: same here

pkg/prometheus/agent/operator.go
```diff
@@ -630,6 +640,52 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+
+	// If the PrometheusAgent instance is marked for deletion, we remove the finalizer.
+	finalizers := p.GetFinalizers()
+	patchBytes, err := k8sutil.FinalizerDeletePatch(finalizers, k8sutil.StatusCleanupFinalizerName)
+	if err != nil {
+		return false, fmt.Errorf("failed to marshal patch: %w", err)
+	}
+
+	if len(patchBytes) == 0 {
+		c.reconciliations.ForgetObject(key)
```
simonpasquier: do we need this here?

pkg/operator/resource_reconciler.go
```diff
@@ -336,7 +338,7 @@ func (rr *ResourceReconciler) OnUpdate(old, cur interface{}) {
 		return
 	}
 
-	if rr.DeletionInProgress(mCur) {
+	if !k8sutil.IsFinalizerPresent(mCur.GetFinalizers(), k8sutil.StatusCleanupFinalizerName) && rr.DeletionInProgress(mCur) {
```
simonpasquier, line 341: maybe we should have a new method which returns true if the finalizer is not set and the deletion timestamp is not nil?

yp969803:
@simonpasquier i can see added finalizer log in operator state of e2e test, the finalizers are added successfully, 
```
ts=2025-06-11T20:53:00.472352169Z level=debug caller=/workspace/pkg/operator/event_handler.go:53 msg="Secret added" component=prometheus-controller
ts=2025-06-11T20:53:02.071548281Z level=debug caller=/workspace/pkg/operator/resource_reconciler.go:305 msg="Prometheus added" component=prometheus-controller key=gatedfeatures-statusforconfigresources-sxplzs-1-b302e653/status-cleanup-finalizer-test
ts=2025-06-11T20:53:02.076883465Z level=info caller=/workspace/pkg/prometheus/server/operator.go:997 msg="added finalizer to Prometheus resource" component=prometheus-controller name=status-cleanup-finalizer-test namespace=gatedfeatures-statusforconfigresources-sxplzs-1-b302e653 finalizer=monitoring.coreos.com/status-cleanup
ts=2025-06-11T20:56:00.26981196Z level=debug caller=/workspace/pkg/kubelet/controller.go:364 msg="Synchronizing nodes" component=kubelet_endpoints kubelet_object=kube-system/kubelet
ts=2025-06-11T20:56:00.270725716Z level=error caller=/workspace/pkg/kubelet/controller.go:369 msg="Failed to list nodes" component=kubelet_endpoints kubelet_object=kube-system/kubelet err="nodes is forbidden: User \"system:serviceaccount:gatedfeatures-statusforconfigresources-sxplzs-1-b302e653:prometheus-operator\" cannot list resource \"nodes\" in API group \"\" at the cluster scope"
```

yp969803:
logs of promethes state
```
Prometheus=gatedfeatures-statusforconfigresources-sxplzs-1-b302e653/status-cleanup-finalizer-test: condition type="Available" status="False" reason="StatefulSetNotFound" message="shard 0: statefulset gatedfeatures-statusforconfigresources-sxplzs-1-b302e653/prometheus-status-cleanup-finalizer-test not found"
Prometheus=gatedfeatures-statusforconfigresources-sxplzs-1-b302e653/status-cleanup-finalizer-test: condition type="Reconciled" status="True" reason="" message=""
```

yp969803:
should i increase the context deadline time, as i cannot see any error messages

yp969803:
i dont think the issue is with context deadline, as i increased it to 10min from 5min, still the test fails

yp969803:
if i remove syncFinalizer function test is running, hence the issue is somewhere there

yp969803:
```
//	if finalizersChanged {
//	return nil
//	}
```
if i comment these lines tests passed successfully @simonpasquier

yp969803:
btw things are workiing fine without doing any change locally

yp969803:
@simonpasquier can u merge this now

yp969803:
the deletionInProgress stuff i will do it in another pr

simonpasquier:
> i dont think the issue is with context deadline, as i increased it to 10min from 5min, still the test fails

In general, the retry timeouts are generous enough that it's rarely them causing trouble.

pkg/prometheus/server/operator.go
```diff
@@ -964,6 +973,52 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
+// add or remove finalizers for the Prometheus resource.
+// Returns true if the finalizers were modified, otherwise false. The second return value is an error, if any.
```
simonpasquier:
```suggestion
// syncFinalizers adds or removes the finalizer for the Prometheus resource.
// It returns true if the finalizers were modified, otherwise false. The second return value is an error, if any.
```

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,50 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
 	}
 	return nil
 }
+
+// AddFinalizerPatch generates the JSON patch payload which adds the finalizer to the object's metadata.
+// If the finalizer is already present, it returns empty []byte slice.
```
simonpasquier:
```suggestion
// If the finalizer is already present, it returns an empty []byte slice.
```

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,50 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
...
+			"op":    "add",
+			"path":  "/metadata/finalizers/-",
+			"value": finalizerName,
+		},
+	}
+	return json.Marshal(patch)
+}
+
+// FinalizerDeletePatch generates the JSON patch payload which delete the finalizer to the object's metadata.
+// If the finalizer is not present, it returns empty []byte slice.
```
simonpasquier:
```suggestion
// If the finalizer is not present, it returns an empty []byte slice.
```

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,50 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
...
+		{
+			"op":    "add",
+			"path":  "/metadata/finalizers/-",
+			"value": finalizerName,
+		},
+	}
+	return json.Marshal(patch)
+}
+
+// FinalizerDeletePatch generates the JSON patch payload which delete the finalizer to the object's metadata.
```
simonpasquier:
```suggestion
// FinalizerDeletePatch generates the JSON patch payload which deletes the finalizer from the object's metadata.
```

pkg/prometheus/server/operator.go
```diff
@@ -753,8 +754,16 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	logger := c.logger.With("key", key)
 	c.logDeprecatedFields(logger, p)
 
-	// Check if the Prometheus instance is marked for deletion.
-	if c.rr.DeletionInProgress(p) {
+	if !c.configResourcesStatusEnabled && c.rr.DeletionInProgress(p) {
```
simonpasquier: wouldn't it be more logical to move this after the call to syncFinalizers? 

E.g.

```
finalizersChanged, err := c.syncFinalizers(ctx, p, key)
if err != nil {
...

if c.rr.DeletionInProgress(p) {
  return nil
}
````

File: pkg/prometheus/server/operator.go [hunk 40]
```diff
@@ -753,8 +754,16 @@ func (c *Operator) sync(ctx context.Context, key string) error {
...
+		return nil
+	}
+
+	finalizersChanged, err := c.syncFinalizers(ctx, p, key)
+	if err != nil {
+		return err
+	}
+	if finalizersChanged {
+		c.rr.EnqueueForReconciliation(p)
```
simonpasquier, line 763: (nit)
```suggestion
		// Since the object has been updated, let's trigger another sync.
		c.rr.EnqueueForReconciliation(p)
```

Related remark: I think that the resource reconciler didn't trigger another reconciliation because hasStateChanged() doesn't take into account the finalizers list (e.g. it considers that the 2 objects are "identical").

test/e2e/status_subresource_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
+	testFramework "github.com/prometheus-operator/prometheus-operator/test/framework"
+)
+
+// testStatusForConfigResources tests the adding/removing of status-cleanup finalizer for Prometheus when StatusForConfigurationResourcesFeature is enable.
```
simonpasquier:
```suggestion
// testStatusForConfigResources tests the adding/removing of status-cleanup finalizer for Prometheus when StatusForConfigurationResourcesFeature is enabled.
```

test/e2e/status_subresource_test.go
```diff
@@ -0,0 +1,54 @@
...
+func testStatusForConfigResources(t *testing.T) {
```
simonpasquier: note for later (when we extend the tests): the test name could be more specific since it is about testing the finalizer mechanics.

simonpasquier (commented):
I'm worried that we're adding duplicated code to all controllers, especially since we're only testing with the Prometheus CR. Can you make changes only to the Prometheus server controller for now? We can refactor the reset later once we have a clearer picture.

I'll commit the other stylistic suggestions I've made.

yp969803:
@simonpasquier done the changes, it can be merged now

pkg/k8sutil/k8sutil_test.go
```diff
@@ -556,6 +558,102 @@ func TestConvertToK8sDNSConfig(t *testing.T) {
...
+		name          string
+		finalizers    []string
+		finalizerName string
+		expectedPatch []map[string]interface{}
+		expectEmpty   bool
+	}{
+		{
+			name:          "empty finalizers",
+			finalizers:    []string{},
+			finalizerName: "cleanup.kubernetes.io/finalizer",
```
mviswanathsai: (nit) might be cleaner to declare the finalizer name inside the test and then using it's value.

pkg/operator/resource_reconciler.go
mviswanathsai: Agreed. There is a bit re-use that would justify it

yp969803:
this part
```
 !k8sutil.IsFinalizerPresent(mCur.GetFinalizers(), k8sutil.StatusCleanupFinalizerName) 
```
i will do in a separate pr, as i have to figure it out how to converge finalizer checking and deleting timestamp checking in one function @mviswanathsai

pkg/alertmanager/operator.go
```diff
@@ -545,6 +543,10 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 
 	logger.Info("sync alertmanager")
 
+	if c.rr.DeletionInProgress(am) {
+		return nil
+	}
+
```
simonpasquier: I would revert this change.

pkg/k8sutil/k8sutil.go
```diff
@@ -599,3 +603,50 @@ func EnsureCustomGoverningService(ctx context.Context, namespace string, service
...
+			patch := []map[string]interface{}{
+				{
+					"op":   "remove",
+					"path": fmt.Sprintf("/metadata/finalizers/%d", i),
+				},
+			}
+			return json.Marshal(patch)
+		}
+	}
+	return []byte{}, nil
```
simonpasquier: (nit) we can return nil
```suggestion
	return nil, nil
```

pkg/operator/resource_reconciler.go
```diff
@@ -407,7 +409,7 @@ func (rr *ResourceReconciler) onDaemonSetAdd(ds *appsv1.DaemonSet) {
 func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet) {
 	rr.logger.Debug("update handler", "resource", "statefulset", "old", old.ResourceVersion, "cur", cur.ResourceVersion)
 
-	if rr.DeletionInProgress(cur) {
+	if !k8sutil.IsFinalizerPresent(cur.GetFinalizers(), k8sutil.StatusCleanupFinalizerName) && rr.DeletionInProgress(cur) {
```
simonpasquier: AFAICT we don't need to check for finalizers on the statefulsets, we're only interested on the parent object (e.g. Prometheus, Alertmanager, ...).
```suggestion
	if rr.DeletionInProgress(cur) {
```

pkg/operator/resource_reconciler.go
```diff
@@ -437,7 +439,7 @@ func (rr *ResourceReconciler) onStatefulSetUpdate(old, cur *appsv1.StatefulSet)
 func (rr *ResourceReconciler) onDaemonSetUpdate(old, cur *appsv1.DaemonSet) {
 	rr.logger.Debug("update handler", "resource", "daemonset", "old", old.ResourceVersion, "cur", cur.ResourceVersion)
 
-	if rr.DeletionInProgress(cur) {
+	if !k8sutil.IsFinalizerPresent(cur.GetFinalizers(), k8sutil.StatusCleanupFinalizerName) && rr.DeletionInProgress(cur) {
```
simonpasquier: same here
```suggestion
	if rr.DeletionInProgress(cur) {
```

pkg/prometheus/agent/operator.go
```diff
@@ -955,7 +956,9 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	}
 	p := pobj.(*monitoringv1alpha1.PrometheusAgent)
 	p = p.DeepCopy()
-
+	if p == nil || c.rr.DeletionInProgress(p) {
```
simonpasquier: we discussed offline but having a `nil` pointer here would be a programmatic bug so there's no real need to check for nil.
Also the check on the object's deletion being in progress should go in a separate PR for clarity.

pkg/thanos/operator.go
```diff
@@ -453,7 +456,9 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 		return fmt.Errorf("failed to set ThanosRuler type information: %w", err)
 	}
 
-	// Check if the Thanos instance is marked for deletion.
+	logger := o.logger.With("key", key)
```
simonpasquier: not related to this PR so let's revert :)

pkg/prometheus/agent/operator.go
```diff
@@ -955,7 +957,9 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 	}
 	p := pobj.(*monitoringv1alpha1.PrometheusAgent)
 	p = p.DeepCopy()
-
+	if c.rr.DeletionInProgress(p) {
+		return nil
+	}
```
simonpasquier, line 962: note for a following PR: we can move the check in c.statusReporter.Process().

pkg/thanos/operator.go
```diff
@@ -613,7 +616,7 @@ func (o *Operator) UpdateStatus(ctx context.Context, key string) error {
 		return err
 	}
 
-	if tr == nil || o.rr.DeletionInProgress(tr) {
+	if o.rr.DeletionInProgress(tr) {
```
simonpasquier: we don't need to modify this.
```suggestion
	if tr == nil || o.rr.DeletionInProgress(tr) {
```

pkg/alertmanager/operator.go
```diff
@@ -694,7 +697,7 @@ func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
 		return err
 	}
 
-	if a == nil || c.rr.DeletionInProgress(a) {
+	if c.rr.DeletionInProgress(a) {
```
simonpasquier:
```suggestion
	if a == nil || c.rr.DeletionInProgress(a) {
```

yp969803:
@simonpasquier merge request

PR #7598, opened 2025-06-12
Ability to add, get and delete objects from asset store from package consumers

### LINKED ISSUE #7591: Ability to update asset store when watch is received
### Component(s)

ServiceMonitor

[Opentelemetry operator](https://github.com/open-telemetry/opentelemetry-operator/tree/main) has a dependeny on prometheus-operator's packages to support detection of pod and service monitors. The targetallocator's [promOperator](https://github.com/open-telemetry/opentelemetry-operator/blob/16e6ccb3aa51c149278432e399ce31c8902af0e2/cmd/otel-allocator/internal/watcher/promOperator.go#L279) component responsible for watching pod and service monitors lacks the ability to pick up the latest changes to the secrets when they are updated. In order to do so a new secretinformer can be added which watches for the events and when a change is detected it needs to be able to update/delete objects from the object store so that the CRs can pick up the latest changes.

Currently the [storebuilder](https://github.com/prometheus-operator/prometheus-operator/blob/c4ebc762d0d2263541c67ebfe1ba7f2b419ed547/pkg/assets/store.go#L63) doesn't provide the ability to update the objStore when objects are updated. Can we extend this to add a couple methods that will provide for updating the store from consumers of this package - in this case promOperator.

Happy to make this change, if this proposal sounds good. 

_No response_

## Environment
Kubernetes Version:
Prometheus-Operator Version:

rashmichandrashekar:
## Description

[Opentelemetry operator](https://github.com/open-telemetry/opentelemetry-operator/tree/main) has a dependeny on prometheus-operator's packages to support detection of pod and service monitors. The targetallocator's [promOperator](https://github.com/open-telemetry/opentelemetry-operator/blob/16e6ccb3aa51c149278432e399ce31c8902af0e2/cmd/otel-allocator/internal/watcher/promOperator.go#L279) component responsible for watching pod and service monitors lacks the ability to pick up the latest changes to the secrets when they are updated/deleted. In order to do so a new secretinformer can be added which watches for the events and when a change is detected it needs to be able to update/delete objects( need add and get too to check the store before performing the operation) so that the CRs can pick up the latest changes.

Currently the [storebuilder](https://github.com/prometheus-operator/prometheus-operator/blob/c4ebc762d0d2263541c67ebfe1ba7f2b419ed547/pkg/assets/store.go#L63) doesn't provide the ability to update the objStore when objects are updated. Can we extend this to add methods to get, add, update and delete objects that will provide for updating the store from consumers of this package - in this case promOperator.

Resolves #7591 
## Type of change

## Verification
Tested with unit tests

## Changelog entry

Ability to get, add and delete objects from assets obj store

```release-note
Ability to add, get and delete objects from assets obj store for consumers of the assets package
```

### DIFF
--- a/pkg/assets/store.go
+++ b/pkg/assets/store.go
@@ -389,6 +389,24 @@ func (cos *cacheOnlyStore) TLSAsset(sel interface{}) string {
 	return k.toString()
 }
 
+func (s *StoreBuilder) HasObject(obj interface{}) (bool, error) {
+	if obj == nil {
+		return false, errors.New("object cannot be nil")
+	}
+
+	key, err := assetKeyFunc(obj)
+	if err != nil {
+		return false, fmt.Errorf("failed to get key for object: %w", err)
+	}
+
+	_, exists, err := s.objStore.GetByKey(key)
+	if err != nil {
+		return false, fmt.Errorf("failed to check object in store: %w", err)
+	}
+
+	return exists, nil
+}
+
 // AddObject adds an object to the underlying store.
 // This method is only used by external clients of the assets package such as the OpenTelemetry collector operator.
 
--- a/pkg/assets/store.go
+++ b/pkg/assets/store.go
@@ -402,3 +402,17 @@ func (s *StoreBuilder) UpdateObject(obj interface{}) error {
 
 	return nil
 }
+
+// DeleteObject updates the object in the underlying store.
+// This method is only used by external clients of the assets package such as the OpenTelemetry collector operator.
+func (s *StoreBuilder) DeleteObject(obj interface{}) error {
+	if obj == nil {
+		return errors.New("object cannot be nil")
+	}
+
+	if err := s.objStore.Delete(obj); err != nil {
+		return fmt.Errorf("failed to delete object in store: %w", err)
+	}
+
+	return nil
+}
--- a/pkg/assets/store.go
+++ b/pkg/assets/store.go
@@ -389,26 +389,6 @@ func (cos *cacheOnlyStore) TLSAsset(sel interface{}) string {
 	return k.toString()
 }
 
-// HasObject checks if the object exists in the underlying store.
-// This method is only used by external clients of the assets package such as the OpenTelemetry collector operator.
-func (s *StoreBuilder) HasObject(obj interface{}) (bool, error) {
-	if obj == nil {
-		return false, errors.New("object cannot be nil")
-	}
-
-	key, err := assetKeyFunc(obj)
-	if err != nil {
-		return false, fmt.Errorf("failed to get key for object: %w", err)
-	}
-
-	_, exists, err := s.objStore.GetByKey(key)
-	if err != nil {
-		return false, fmt.Errorf("failed to check object in store: %w", err)
-	}
-
-	return exists, nil
-}
-
 // AddObject adds an object to the underlying store.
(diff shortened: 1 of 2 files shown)


pkg/assets/store_test.go
```diff
@@ -1218,3 +1218,86 @@ func TestUpdateObject(t *testing.T) {
...
+	}
+	err = store.DeleteObject(secretObj)
+	require.NoError(t, err)
+
+	// Also delete the secret from the fake clientset to simulate full removal
+	err = c.CoreV1().Secrets("ns1").Delete(context.Background(), "secret", metav1.DeleteOptions{})
+	require.NoError(t, err)
+
+	// Now, getting the key should fail since the secret is deleted from the store
+	val, err = store.GetSecretKey(context.Background(), "ns1", secretSel)
```
simonpasquier: this needs fixing.

slashpai (dismissed):

pkg/assets/store.go
```diff
@@ -389,6 +389,64 @@ func (cos *cacheOnlyStore) TLSAsset(sel interface{}) string {
...
+	if err := s.objStore.Add(obj); err != nil {
+		return fmt.Errorf("failed to add object to store: %w", err)
+	}
+
+	return nil
+}
+
+// GetObject retrieves an object from the underlying store.
+// This method is only used by external clients of the assets package such as the OpenTelemetry collector operator.
+func (s *StoreBuilder) GetObject(obj interface{}) (interface{}, error) {
```
simonpasquier: I'd go with a bool return value that returns whether the object is found or not to avoid the HasObject() method.
```suggestion
func (s *StoreBuilder) GetObject(obj interface{}) (interface{}, bool, error) {
```

pkg/assets/store.go
```diff
@@ -402,3 +460,17 @@ func (s *StoreBuilder) UpdateObject(obj interface{}) error {
 
 	return nil
 }
+
+// DeleteObject updates the object in the underlying store.
```
simonpasquier:
```suggestion
// DeleteObject deletes the object in the underlying store.
```

pkg/assets/store.go
```diff
@@ -389,6 +389,64 @@ func (cos *cacheOnlyStore) TLSAsset(sel interface{}) string {
...
+
+	key, err := assetKeyFunc(obj)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get key for object: %w", err)
+	}
+
+	item, exists, err := s.objStore.GetByKey(key)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get object from store: %w", err)
+	}
```
simonpasquier: why not using s.objStore.Get() directly?

pkg/assets/store.go
rashmichandrashekar: Thanks @simonpasquier. I was doing some integration testing with the same changes, hence had converted my PR to draft, yes GetObject serves the purpose and we don't need an additional method. Have removed it.

pkg/assets/store.go
rashmichandrashekar: Thanks! Yes, was testing out the changes while this was in draft, updated them to do Get directly :)

rashmichandrashekar:
@simonpasquier @slashpai - Will there be a release on the 9th of July? I am waiting on the release to pick up the changes in this PR. Thanks!

PR #7604, opened 2025-06-17
feat:  Add alertmanagerConfigMatcherStrategy "OnNamespaceExceptForAlertmanagerNamespace"

### LINKED ISSUE #7603: Feature Request: Disable namespace matcher for just alertmanager namespace
### Component(s)

AlertManager

Presently, the only options for the alertmanagerConfigMatcherStrategy are inappropriate 
for a multi-tenant cluster. When "None" is selected, tenant namespace isolation is removed
for alerts. When "OnNamespace" is selected, cluster administrators cannot use the operator
to manage cluster-scoped alerts, or alerts on metrics from tenant namespaces.

A simple solution to this would be to add a third option, "OnOtherNamespace", which applies
the namespace matcher, but not to AlertManagerConfig's in the same namespace as the
AlertManager itself. This preserves the per-namespace isolation while allowing the 
administrators of the central AlertManager to create alerts for any namespace, and alerts
which have no namespace.

Using a second alertmanager instance.

## Environment
Kubernetes Version: N/A
Prometheus-Operator Version: N/A

meln5674:
Adds option to disable namespace matcher only for namespace of the alertmanager, to allow admins of multi-tenant clusters to create cluster-scoped alerts without disabling tenant isolation.

## Description

Fixes #7603

## Type of change

## Verification

Added unit test cases demonstrating selective adding of namespace matcher.

## Changelog entry

```release-note
- alertmanagerConfigMatcherStrategy now allows the value "OnOtherNamespace" to only exclude the namespace matcher from AlertManagerConfigs in the same namespace as the AlertManager.
```

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -200,6 +200,27 @@ func (ne *namespaceEnforcer) processRoute(crKey types.NamespacedName, r *route) 
 	return r
 }
 
+type otherNamespaceEnforcer struct {
+	alertmanagerNamespace string
+	namespaceEnforcer
+}
+
+var _ enforcer = &otherNamespaceEnforcer{}
+
+func (one *otherNamespaceEnforcer) processInhibitRule(crKey types.NamespacedName, ir *inhibitRule) *inhibitRule {
+	if crKey.Namespace == one.alertmanagerNamespace {
+		return ir
+	}
+	return one.namespaceEnforcer.processInhibitRule(crKey, ir)
+}
+
+func (one *otherNamespaceEnforcer) processRoute(crKey types.NamespacedName, r *route) *route {
+	if crKey.Namespace == one.alertmanagerNamespace {
+		return r
+	}
+	return one.namespaceEnforcer.processRoute(crKey, r)
+}
+
 // ConfigBuilder knows how to build an Alertmanager configuration from a raw
 // configuration and/or AlertmanagerConfig objects.
 // The API is public because it's used by Grafana Alloy (https://github.com/grafana/alloy).
@@ -212,21 +233,28 @@ type ConfigBuilder struct {
 	enforcer  enforcer
 }
 
-func NewConfigBuilder(logger *slog.Logger, amVersion semver.Version, store *assets.StoreBuilder, matcherStrategy monitoringv1.AlertmanagerConfigMatcherStrategy) *ConfigBuilder {
+func NewConfigBuilder(logger *slog.Logger, amVersion semver.Version, store *assets.StoreBuilder, am *monitoringv1.Alertmanager) *ConfigBuilder {
 	cg := &ConfigBuilder{
 		logger:    logger,
 		amVersion: amVersion,
 		store:     store,
-		enforcer:  getEnforcer(matcherStrategy, amVersion),
+		enforcer:  getEnforcer(am.Spec.AlertmanagerConfigMatcherStrategy, amVersion, am.Namespace),
 	}
 	return cg
 }
 
-func getEnforcer(matcherStrategy monitoringv1.AlertmanagerConfigMatcherStrategy, amVersion semver.Version) enforcer {
+func getEnforcer(matcherStrategy monitoringv1.AlertmanagerConfigMatcherStrategy, amVersion semver.Version, amNamespace string) enforcer {
 	var e enforcer
 	switch matcherStrategy.Type {
 	case monitoringv1.NoneConfigMatcherStrategyType:
 		e = &noopEnforcer{}
+	case monitoringv1.OnNamespaceExceptForAlertmanagerNamespaceConfigMatcherStrategyType:
+		e = &otherNamespaceEnforcer{
+			alertmanagerNamespace: amNamespace,
+			namespaceEnforcer: namespaceEnforcer{
+				matchersV2Allowed: amVersion.GTE(semver.MustParse("0.22.0")),
+			},
+		}
 	default:
 		e = &namespaceEnforcer{
 			matchersV2Allowed: amVersion.GTE(semver.MustParse("0.22.0")),
--- a/pkg/alertmanager/amcfg_test.go
+++ b/pkg/alertmanager/amcfg_test.go
@@ -872,7 +872,10 @@ func TestInitializeFromAlertmanagerConfig(t *testing.T) {
 			newNopLogger(t),
 			*tt.amVersion,
 			assets.NewStoreBuilder(kclient.CoreV1(), kclient.CoreV1()),
-			tt.matcherStrategy,
+			&monitoringv1.Alertmanager{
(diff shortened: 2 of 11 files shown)


pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +343,12 @@ const (
 	// label equal to the namespace of the object.
 	OnNamespaceConfigMatcherStrategyType AlertmanagerConfigMatcherStrategyType = "OnNamespace"
 
+	// With `OnOtherNamespace`, the route and inhibition rules of an
+	// AlertmanagerConfig object only process alerts that have a `namespace`
+	// label equal to the namespace of the object, but only if the namespace
+	// is not the namespace of the AlertManager itself.
+	OnOtherNamespaceConfigMatcherStrategyType AlertmanagerConfigMatcherStrategyType = "OnOtherNamespace"
```
simonpasquier: Not sure about the name but I fail to find a better proposition :)

Maybe `ExceptOnAlertmanagerNamespace`?

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +343,12 @@ const (
 	// label equal to the namespace of the object.
 	OnNamespaceConfigMatcherStrategyType AlertmanagerConfigMatcherStrategyType = "OnNamespace"
 
+	// With `OnOtherNamespace`, the route and inhibition rules of an
```
simonpasquier: The value needs to be added at L333 too (`+kubebuilder:validation:Enum`).

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -343,6 +343,12 @@ const (
...
+	// AlertmanagerConfig object only process alerts that have a `namespace`
+	// label equal to the namespace of the object, but only if the namespace
+	// is not the namespace of the AlertManager itself.
```
simonpasquier:
```suggestion
	// label equal to the namespace of the object unless the AlertmanagerConfig object is in the same namespace
	// as the Alertmanager object itself.
```

pkg/alertmanager/amcfg.go
```diff
@@ -212,21 +233,33 @@ type ConfigBuilder struct {
 	enforcer  enforcer
 }
 
-func NewConfigBuilder(logger *slog.Logger, amVersion semver.Version, store *assets.StoreBuilder, matcherStrategy monitoringv1.AlertmanagerConfigMatcherStrategy) *ConfigBuilder {
+func NewConfigBuilderInNamespace(logger *slog.Logger, amVersion semver.Version, store *assets.StoreBuilder, amNamespace string, matcherStrategy monitoringv1.AlertmanagerConfigMatcherStrategy) *ConfigBuilder {
```
simonpasquier: do we need a new function name? it might also be simpler to just pass the Alertmanager object?

meln5674: Because the function is exported, adding a new parameter would technically be a breaking change. If that's not considered part of the public API of this package, I can squash them into one function. In either case, I agree that passing the entire object is good future-proofing.

pkg/apis/monitoring/v1/alertmanager_types.go
meln5674: I interpreted the original "OnNamespace" to mean "enforce matching on namespace", so this new enum would likewise translate as "enforce on other namespace". I think the "namespace" here is getting overloaded as both matching the alert on namespace and enforcing based on object namespace. I feel that "ExceptOnAlertmanagerNamespace" doesn't capture both of those meanings as well as just "OnOtherNamespace", but I can go ahead and use whatever the project decides on.

simonpasquier: I meant something prefixed by OnNamespace so `OnNamespaceExceptForAlertmanagerNamespace` but it's (too) verbose :thinking:

pkg/alertmanager/amcfg.go
simonpasquier: This is why the project itself is v0.x, we don't promise any stability guarantee for the Go code, only for the CRDs.

pkg/apis/monitoring/v1/alertmanager_types.go
meln5674: I split the difference and went with `OnNamepaceInOtherNamespaces`, does that work?

pkg/apis/monitoring/v1/alertmanager_types.go
meln5674: Done

pkg/apis/monitoring/v1/alertmanager_types.go
meln5674: Done

pkg/alertmanager/amcfg.go
meln5674: Understood. Done.

meln5674:
@simonpasquier Sorry to be a nuisance, but I believe I've addressed all of your feedback, is there anything else you're waiting on from me? Do I need to rebase or squash or anything else?

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -330,7 +330,7 @@ type AlertmanagerConfigMatcherStrategy struct {
 	//
 	// The default value is `OnNamespace`.
 	//
-	// +kubebuilder:validation:Enum="OnNamespace";"None"
+	// +kubebuilder:validation:Enum="OnNamespace";"OnNamespaceInOtherNamespaces";"None"
```
simonpasquier:
```suggestion
	// +kubebuilder:validation:Enum="OnNamespace";"OnNamespaceExceptForAlertmanagerNamespace";"None"
```

pkg/alertmanager/amcfg_test.go
```diff
@@ -1264,6 +1267,60 @@ func TestGenerateConfig(t *testing.T) {
...
+		},
+		{
+			name:    "skeleton base in different namespace to alertmanager, simple CR with namespaceMatcher disabled for alertmanager namespace",
+			kclient: fake.NewSimpleClientset(),
+			baseConfig: alertmanagerConfig{
+				Route:     &route{Receiver: "null"},
+				Receivers: []*receiver{{Name: "null"}},
+			},
+			matcherStrategy: monitoringv1.AlertmanagerConfigMatcherStrategy{
+				Type: "OnOtherNamespace",
```
simonpasquier: :thinking:

simonpasquier (commented):
Sorry for the lag, we have a bug flow of PRs coming in these days and I was still pondering about the new strategy name. I made a suggestion which is long-ish but explicit IMHO. Let me know what you think.

meln5674:
That name works for me. I've updated it again, along with the commit message and PR title to match, as well as fixed that errant test.

simonpasquier (approved):
thanks!

Issue #7600, opened 2025-06-16
move the check in c.statusReporter.Process()


simonpasquier:
note for a following PR: we can move the check in c.statusReporter.Process().

_Originally posted by @simonpasquier in https://github.com/prometheus-operator/prometheus-operator/pull/7584#discussion_r2150082099_

yp969803:
this check 
```
if c.rr.DeletionInProgress(p) {
		return nil
	}
```

PR #7606, opened 2025-06-17
feat: getPrometheusFromKey to sync with codestyle of other controller

yp969803:
## Description

fixes: #7600 

feat: getPrometheusFromKey to sync with codestyle of other controller

## Type of change

## Verification

## Changelog entry

```
NONE

```

### DIFF
--- a/pkg/prometheus/agent/operator.go
+++ b/pkg/prometheus/agent/operator.go
@@ -943,21 +943,31 @@ func createSSetInputHash(p monitoringv1alpha1.PrometheusAgent, c prompkg.Config,
 	return fmt.Sprintf("%d", hash), nil
 }
 
+// getPrometheusFromKey returns a copy of the Prometheus object identified by key.
+// If the object is not found, it returns a nil pointer.
+func (c *Operator) getPrometheusFromKey(key string) (*monitoringv1alpha1.PrometheusAgent, error) {
+	obj, err := c.promInfs.Get(key)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			c.logger.Info("Prometheus not found", "key", key)
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to retrieve Prometheus from informer: %w", err)
+	}
+
+	return obj.(*monitoringv1alpha1.PrometheusAgent).DeepCopy(), nil
+}
+
 // UpdateStatus updates the status subresource of the object identified by the given
 // key.
 // UpdateStatus implements the operator.Syncer interface.
 func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
-	pobj, err := c.promInfs.Get(key)
-
-	if apierrors.IsNotFound(err) {
-		return nil
-	}
+	p, err := c.getPrometheusFromKey(key)
 	if err != nil {
 		return err
 	}
-	p := pobj.(*monitoringv1alpha1.PrometheusAgent)
-	p = p.DeepCopy()
-	if c.rr.DeletionInProgress(p) {
+
+	if p == nil || c.rr.DeletionInProgress(p) {
 		return nil
 	}
 	pStatus, err := c.statusReporter.Process(ctx, p, key)
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -1035,23 +1035,31 @@ func (c *Operator) shouldRetain(p *monitoringv1.Prometheus) (bool, error) {
 	return false, nil
 }
 
+// getPrometheusFromKey returns a copy of the Prometheus object identified by key.
+// If the object is not found, it returns a nil pointer.
+func (c *Operator) getPrometheusFromKey(key string) (*monitoringv1.Prometheus, error) {
+	obj, err := c.promInfs.Get(key)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			c.logger.Info("Prometheus not found", "key", key)
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to retrieve Prometheus from informer: %w", err)
+	}
+
+	return obj.(*monitoringv1.Prometheus).DeepCopy(), nil
+}
+
 // UpdateStatus updates the status subresource of the object identified by the given
 // key.
 // UpdateStatus implements the operator.Syncer interface.
 func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
-	pobj, err := c.promInfs.Get(key)
-
-	if apierrors.IsNotFound(err) {
-		return nil
-	}
+	p, err := c.getPrometheusFromKey(key)
 	if err != nil {
 		return err
 	}
 
-	p := pobj.(*monitoringv1.Prometheus)
-	p = p.DeepCopy()
-
-	if c.rr.DeletionInProgress(p) {
+	if p == nil || c.rr.DeletionInProgress(p) {
 		return nil
 	}
 	pStatus, err := c.statusReporter.Process(ctx, p, key)


pkg/prometheus/server/operator.go
```diff
@@ -1035,23 +1035,31 @@ func (c *Operator) shouldRetain(p *monitoringv1.Prometheus) (bool, error) {
...
+	}
+
+	return obj.(*monitoringv1.Prometheus).DeepCopy(), nil
+}
+
 // UpdateStatus updates the status subresource of the object identified by the given
 // key.
 // UpdateStatus implements the operator.Syncer interface.
 func (c *Operator) UpdateStatus(ctx context.Context, key string) error {
-	pobj, err := c.promInfs.Get(key)
```
simonpasquier, line 1042: The sync() method uses the same construct. Can we also refactor it?

yp969803:
I agree with u

yp969803:
@simonpasquier check now

pkg/prometheus/server/operator.go
yp969803: yea

pkg/common/common.go
simonpasquier, line 1: `common` is too generic, let's put it under pkg/operator

yp969803:
@simonpasquier PTAL

pkg/operator/common.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"log/slog"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
+)
+
+// GetObjectFromKey retrieves an object from the informer cache using the provided key.
+func GetObjectFromKey[T interface {
```
simonpasquier: I'd recommend using a constraint enumerating supported types.

```golang
type CustomResource interface {
    *monv1.Prometheus | *monv1.Alertmanager | *monv1alpha1.PrometheusAgent | *monv1.ThanosRuler
}

// GetObjectFromKey retrieves an object from the informer cache using the provided key.
func GetObjectFromKey[T CustomResource](infs *informers.ForResource, key string, logger *slog.Logger) (T, error) {
...

```

pkg/prometheus/server/operator.go
```diff
@@ -1039,19 +1039,12 @@ func (c *Operator) shouldRetain(p *monitoringv1.Prometheus) (bool, error) {
...
+	p, err := operator.GetObjectFromKey[*monitoringv1.Prometheus](c.promInfs, key, c.logger)
 	if err != nil {
 		return err
 	}
 
-	p := pobj.(*monitoringv1.Prometheus)
-	p = p.DeepCopy()
-
-	if c.rr.DeletionInProgress(p) {
+	if p == nil || c.rr.DeletionInProgress(p) {
```
simonpasquier: I would keep the 2 tests separated for clarity

````golang
if p == nil {
    return nil
}

if c.rr.DeletionInProgress(p) {
    return nil
}

File: pkg/operator/common.go [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+	DeepCopy() T
+}](infs *informers.ForResource, key string, logger *slog.Logger) (T, error) {
+	obj, err := infs.Get(key)
+	var zero T
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("Object not found", "key", key)
```
simonpasquier: do we really need to log? In general for "simple" functions, I prefer to avoid logging and defer to the caller.

pkg/operator/common.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+	"log/slog"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
+)
+
+// GetObjectFromKey retrieves an object from the informer cache using the provided key.
```
simonpasquier: We should note that the returned value can be nil if the object isn't found.

pkg/operator/common.go
simonpasquier, line 1: I'd suggeset pkg/operator/informers.go since it's related to informers.

pkg/operator/common.go
yp969803: The function can be used for any resource, including both workload and config rsource, the customResouce interface will become very big to include all the resources type, thats why i use any,  it will be the duty of the caller to give correct types

pkg/operator/common.go
yp969803: yes

pkg/operator/common.go
yp969803: ok

pkg/operator/common.go
```diff
@@ -0,0 +1,46 @@
...
+	obj, err := infs.Get(key)
+	var zero T
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("Object not found", "key", key)
+			return zero, nil
+		}
+		return zero, fmt.Errorf("failed to retrieve object from informer: %w", err)
+	}
+
```
simonpasquier: we need to call `k8sutil.AddTypeInformationToObject` too.

pkg/operator/common.go
simonpasquier: > the customResouce interface will become very big to include all the resources type

It's not really an issue but yeah no need to enumerate all types. I'd suggest  to reuse an existing interface though.

```
func GetObjectFromKey[T runtime.Object](...
```

pkg/operator/common.go
```diff
@@ -0,0 +1,46 @@
...
+	typed, ok := any(obj).(T)
+	if !ok {
+		return zero, fmt.Errorf("unexpected type %T", obj)
+	}
+
+	return typed.DeepCopy(), nil
```
simonpasquier: I'd recommend copying the object first.

pkg/operator/common.go
yp969803: ok

pkg/operator/common.go
yp969803: ok

pkg/operator/common.go
yp969803: ok

yp969803:

pkg/operator/informers.go
```diff
@@ -0,0 +1,49 @@
...
+	var zero T
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return zero, nil
+		}
+		return zero, fmt.Errorf("failed to retrieve object from informer: %w", err)
+	}
+
+	copy, ok := obj.DeepCopyObject().(T)
```
simonpasquier:
```suggestion
	obj, ok := obj.DeepCopyObject()

File: pkg/operator/informers.go [hunk 9]
```diff
@@ -0,0 +1,49 @@
...
+	if !ok {
+		return zero, fmt.Errorf("object %T is not of type %T", copy, zero)
+	}
+
+	if err = k8sutil.AddTypeInformationToObject(obj); err != nil {
+		return zero, err
+	}
+	return copy, nil
```
simonpasquier:
```suggestion
	return obj.(T), nil
```

pkg/operator/informers.go
yp969803: ok

pkg/operator/informers.go
yp969803: ok

yp969803:
@simonpasquier review and merge request

pkg/operator/informers.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
+	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
+)
+
+// GetObjectFromKey retrieves an object from the informer cache using the provided key.
+// Returns nil,nil if the object is not found, or an error if there is a problem retrieving it.
+func GetObjectFromKey[T runtime.Object](infs *informers.ForResource, key string) (T, error) {
+	obj, err := infs.Get(key)
+	var zero T
```
simonpasquier, line 31: TIL (Today I learned) :)

https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md#the-zero-value

pkg/operator/informers.go
```diff
@@ -0,0 +1,46 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/prometheus-operator/prometheus-operator/pkg/informers"
+	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
+)
+
+// GetObjectFromKey retrieves an object from the informer cache using the provided key.
+// Returns nil,nil if the object is not found, or an error if there is a problem retrieving it.
```
simonpasquier: (nit) I wouldn't describe the last return value but I would rephrase to make it clear that no error is returned when not found. 
```suggestion
// GetObjectFromKey retrieves an object from the informer cache using the provided key. It returns a nil value and no error if the object is not found.
// The function will panic if the caller provides an informer which doesn't reference objects of type T.
```

pkg/operator/informers.go
```diff
@@ -0,0 +1,46 @@
...
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return zero, nil
+		}
+		return zero, fmt.Errorf("failed to retrieve object from informer: %w", err)
+	}
+
+	obj = obj.DeepCopyObject()
+
```
simonpasquier: (nit)
```suggestion
```

pkg/alertmanager/operator.go
```diff
@@ -517,21 +517,18 @@ func (c *Operator) Sync(ctx context.Context, key string) error {
...
-
-	if apierrors.IsNotFound(err) {
-		c.reconciliations.ForgetObject(key)
-		// Dependent resources are cleaned up by K8s via OwnerReferences
-		return nil
-	}
+	am, err := operator.GetObjectFromKey[*monitoringv1.Alertmanager](c.alrtInfs, key)
 	if err != nil {
 		return err
 	}
```
simonpasquier, line 524: I would check for am == nil here. Also why do we remove the call to ForgetObject?

pkg/alertmanager/operator.go
```diff
@@ -543,11 +540,8 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 		return nil
 	}
 
-	logger := c.logger.With("key", key)
 	logDeprecatedFields(logger, am)
 
-	logger.Info("sync alertmanager")
```
simonpasquier: It's an info log which is why it's printed only when the controller does actual work (e.g. no deletion in progress or pause).

simonpasquier (commented):
I made a few comments (which apply to all other files, not only pkg/alertmanager) but LGTM on the approach :bow:

pkg/operator/informers.go
yp969803: so instead of returning nil we should return zero value
```
var zero T
	return zero
```

yp969803: already returning zero instead of nil

pkg/operator/informers.go
yp969803: ok

pkg/operator/informers.go
yp969803: ok

pkg/prometheus/agent/operator.go
```diff
@@ -947,16 +941,17 @@ func createSSetInputHash(p monitoringv1alpha1.PrometheusAgent, c prompkg.Config,
...
-		return nil
-	}
 	if err != nil {
 		return err
 	}
-	p := pobj.(*monitoringv1alpha1.PrometheusAgent)
-	p = p.DeepCopy()
+
+	if p == nil {
+		c.logger.Info("PrometheusAgent object not found", "key", key)
```
simonpasquier: we didn't log before?
```suggestion
```

pkg/prometheus/server/operator.go
```diff
@@ -734,26 +734,23 @@ func (c *Operator) Sync(ctx context.Context, key string) error {
...
+		c.reconciliations.ForgetObject(key)
+		// Dependent resources are cleaned up by K8s via OwnerReferences
+		return nil
 	}
 
 	logger := c.logger.With("key", key)
 	c.logDeprecatedFields(logger, p)
 
+	logger.Info("sync prometheus")
+
```
simonpasquier:
```suggestion
```

pkg/prometheus/server/operator.go
```diff
@@ -1039,17 +1036,15 @@ func (c *Operator) shouldRetain(p *monitoringv1.Prometheus) (bool, error) {
...
-	}
+	p, err := operator.GetObjectFromKey[*monitoringv1.Prometheus](c.promInfs, key)
 	if err != nil {
 		return err
 	}
 
-	p := pobj.(*monitoringv1.Prometheus)
-	p = p.DeepCopy()
+	if p == nil {
+		c.logger.Info("Prometheus object not found", "key", key)
```
simonpasquier:
```suggestion
```

pkg/thanos/operator.go
```diff
@@ -611,12 +592,17 @@ func (o *Operator) getStatefulSetFromThanosRulerKey(key string) (*appsv1.Statefu
...
 func (o *Operator) UpdateStatus(ctx context.Context, key string) error {
-	tr, err := o.getThanosRulerFromKey(key)
+	tr, err := operator.GetObjectFromKey[*monitoringv1.ThanosRuler](o.thanosRulerInfs, key)
 	if err != nil {
 		return err
 	}
 
-	if tr == nil || o.rr.DeletionInProgress(tr) {
+	if tr == nil {
+		o.logger.Info("ThanosRuler object not found, skipping status update", "key", key)
```
simonpasquier:
```suggestion
```

yp969803:
@simonpasquier @slashpai done the chages

simonpasquier (approved):
Nice!

PR #7486, opened 2025-04-25
Use the AlertmanagerConfig builder outside of the operator repo

ptodev:
## Description

Fixes #7418

I have been working on a new Grafana Alloy [feature](https://github.com/grafana/alloy/pull/3448) which gathers AlertmanagerConfig CRD resources, merges them, and sends them to an Alertmanager embedded into Grafana Mimir. It would be very helpful if there is a way to use the config builder. I suppose this change could be useful for other projects too, at some point. 

In this PR I only exported the functions which are useful for the Alloy feature. I still haven't finished working on it, but I'm pretty sure that at the very least I would need the functionality from this PR. I'm happy to change the PR if you think it makes sense to export the functionality differently 😊 

I'm not sure if there should be a changelog entry? There is no change from the point of view of Operator users.

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [x] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Verification

Please check the [Prometheus-Operator testing guidelines](../TESTING.md) for recommendations about automated tests.

## Changelog entry

_Please put a one-line changelog entry below. This will be copied to the changelog file during the release process._

```release-note

```

PR #7614, opened 2025-06-19
[release-0.82] Use the AlertmanagerConfig builder outside of the operator repo

ptodev:
This change cherry picks  #7486 onto the [release-0.82](https://github.com/prometheus-operator/prometheus-operator/commits/release-0.82/) branch. I hope that this is ok? I [asked](https://github.com/prometheus-operator/prometheus-operator/pull/7486#issuecomment-2980632149) about it in the PR which was merged to `main` but didn't get a reply so I decided to open a PR for us to discuss it here.

It is difficult to use v0.83 in projects which also import OTel Collector and Prometheus, because the Prometheus Operator uses a more recent version of `k8s.io/api`.
* [Updating OTel](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39725) is not straightforward because it also requires an update of the Go version, but the project [can't do that](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38970#issuecomment-2980476964) until Go 1.25 is released.
* Updating Prometheus would be easier, but it requires transitioning away from some deprecated fields which I didn't have time to do and test just for the Alertmanager CRD project.

If it's a standard practice to patch the release branch then I'd be very grateful if we do 🙏  But if it's not desirable I could wait a while for Prometheus and OTel to be updated, it shouldn't be a huge problem.

### DIFF
--- a/pkg/alertmanager/amcfg.go
+++ b/pkg/alertmanager/amcfg.go
@@ -200,9 +200,11 @@ func (ne *namespaceEnforcer) processRoute(crKey types.NamespacedName, r *route) 
 	return r
 }
 
-// configBuilder knows how to build an Alertmanager configuration from a raw
+// ConfigBuilder knows how to build an Alertmanager configuration from a raw
 // configuration and/or AlertmanagerConfig objects.
-type configBuilder struct {
+// The API is public because it's used by Grafana Alloy (https://github.com/grafana/alloy).
+// Note that the project makes no API stability guarantees.
+type ConfigBuilder struct {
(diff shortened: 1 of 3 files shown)


slashpai:
Thanks for the PR

Since v0.83 is already released we don't patch a release before that usually

simonpasquier:
What @slashpai said. An option for you would be to maintain a fork with your cherry-pick and use a replace directive in go.mod.

ptodev:
No worries, thank you for replying! I will try to setup a fork.

PR #7615, opened 2025-06-19
fix: reconcile workload configurations when a referenced secret/configmap key is updated

### LINKED ISSUE #6018: ServiceMonitor using secrets that are created later
**What did you do?**
I am deploying Elastic using the elastic operator.
When deploying the Elastic custom resource the operator creates elastic pods, as well as secrets that contain certificates and the password to the instance.
I want to scrape metrics from the Elastic instance, so I deploy a `serviceMonitor` to instruct the prometheus operator to do this.
The `serviceMonitor` needs the username, password and CA certificate in order to connect to the elastic instance.
This information is contained within secrets that is created by the elastic operator.

I am deploying both the elastic custom resource and the `ServiceMonitor` using helm.
After applying the elastic custom resource it takes a little while for the operator to deploy the elastic pods and create the secrets containing the certificated and password for the instance.
This means that the `ServiceMonitor` is created before these secrets exists.
When this happens prometheus disables the `ServiceMonitor` and does not re-try it.

```
level=warn ts=2023-10-17T11:32:39.147133443Z caller=operator.go:2255 component=prometheusoperator msg="skipping servicemonitor" error="failed to get basic auth username: unable to get secret \"prometheus-elastic-basic-auth-username\": secrets \"prometheus-elastic-basic-auth-username\" not found" servicemonitor=NAMESPACE/elk-prometheus-metrics namespace=CLUSTER-NAMESPACE prometheus=user-workload
```

A possible solution to this is to deploy the secrets using helm, but empty. The operator will then update the secrets with the correct information later.
I did this, but got another error

```
ts=2023-10-18T09:53:44.698Z caller=manager.go:216 level=error component="scrape manager" msg="error creating new scrape pool" 
err="error creating HTTP client: unable to load specified CA cert /etc/prometheus/certs/secret_NAMESPACE_-s-http-ca-internal_tls.crt: 
open /etc/prometheus/certs/secret_NAMESPACE-es-http-ca-internal_tls.crt: no such file or directory" scrape_pool=serviceMonitor/NAMESPACE/elk-prometheus-metrics-data/0
```

First I expected prometheus to re-try adding the `ServiceMonitor` even after the secret did not exist at the first attempt.
This would be the best solution and would require no work-around on my part.

I then have a question. If I add empty secrets instead, will prometheus re-load the info from the secrets when the content changes? It looks like prometheus mounts the secret into itself as a file and reads it. When the secret updates it will not automatically reflect on the prometheus server as it needs to reload the secret. Is there a mechanism for this?
And does this mechanism work, even if it fails to load the certificate the first attempt because it is empty?

**Environment**

operator: 0.60.1
prometheus:  2.39.1

v1.25.4+a34b9e9

* Kubernetes cluster kind:

    OpenShift

* Manifests:

```
---
kind: ServiceMonitor
metadata:
  name: elk-prometheus-metrics-master
  labels:
    app: elk
spec:
  selector:
    matchLabels:
      common.k8s.elastic.co/type: elasticsearch
      elasticsearch.k8s.elastic.co/statefulset-name: {{ .Release.Name }}-es-master
  endpoints:
    - path: "/_prometheus/metrics"
      port: https
      scheme: https
      basicAuth:
        password:
          name: {{ .Release.Name }}-es-elastic-user
          key: elastic
        username:
          name: prometheus-elastic-basic-auth-username
          key: username
      tlsConfig:
        ca: 
          secret: 
            name: {{ .Release.Name }}-es-http-ca-internal
            key: tls.crt
        serverName: {{ .Release.Name }}-es-http.{{ .Release.Namespace }}.es.local
```

simonpasquier:
## Description

Closes #6018 

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -168,6 +168,7 @@ func parseFlags(fs *flag.FlagSet) {
 	fs.Var(cfg.Namespaces.AlertmanagerAllowList, "alertmanager-instance-namespaces", "Namespaces where Alertmanager custom resources and corresponding StatefulSets are watched/created. If set this takes precedence over --namespaces or --deny-namespaces for Alertmanager custom resources.")
 	fs.Var(cfg.Namespaces.AlertmanagerConfigAllowList, "alertmanager-config-namespaces", "Namespaces where AlertmanagerConfig custom resources and corresponding Secrets are watched/created. If set this takes precedence over --namespaces or --deny-namespaces for AlertmanagerConfig custom resources.")
 	fs.Var(cfg.Namespaces.ThanosRulerAllowList, "thanos-ruler-instance-namespaces", "Namespaces where ThanosRuler custom resources and corresponding StatefulSets are watched/created. If set this takes precedence over --namespaces or --deny-namespaces for ThanosRuler custom resources.")
+	fs.BoolVar(&cfg.WatchObjectRefsInAllNamespaces, "watch-referenced-objects-in-all-namespaces", false, "When true the operator watches for configmaps and secrets in both workload and configuration resource namespaces.\nWhen false (default), the operator will only watch for secrets and configmaps in:\n* Workload namespaces for Prometheus and PrometheusAgent resources.\n* Configuration namespaces for Alertmanager resources.")
 
 	fs.Var(&cfg.Annotations, "annotations", "Annotations to be add to all resources created by the operator")
 	fs.Var(&cfg.Labels, "labels", "Labels to be add to all resources created by the operator")
@@ -212,19 +213,21 @@ func run(fs *flag.FlagSet) int {
 	}
 
 	logger.Info("Starting Prometheus Operator", "version", version.Info(), "build_context", version.BuildContext(), "feature_gates", cfg.Gates.String())
+	logger.Info("Operator's configuration",
+		"watch_referenced_objects_in_all_namespaces", cfg.WatchObjectRefsInAllNamespaces,
+		"controller_id", cfg.ControllerID,
+		"enable_config_reloader_probes", cfg.ReloaderConfig.EnableProbes)
 	goruntime.SetMaxProcs(logger)
 	goruntime.SetMemLimit(logger, memlimitRatio)
 
 	if len(cfg.Namespaces.AllowList) > 0 && len(cfg.Namespaces.DenyList) > 0 {
-		logger.Error(
-			"--namespaces and --deny-namespaces are mutually exclusive, only one should be provided",
-			"namespaces", cfg.Namespaces.AllowList,
-			"deny_namespaces", cfg.Namespaces.DenyList,
-		)
 		return 1
 	}
-	cfg.Namespaces.Finalize()
-	logger.Info("namespaces filtering configuration ", "config", cfg.Namespaces.String())
+	if err := cfg.Namespaces.Finalize(); err != nil {
+		logger.Error("failed to parse namespaces configuration", "configuration", cfg.Namespaces.String(), "error", err)
+		return 1
+	}
+	logger.Info("Namespaces filtering configuration ", "config", cfg.Namespaces.String())
 
 	ctx, cancel := context.WithCancel(context.Background())
 	wg, ctx := errgroup.WithContext(ctx)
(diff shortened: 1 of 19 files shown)


jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -30,6 +30,7 @@ local defaults = {
   kubeletService: 'kube-system/kubelet',
   kubeletEndpointsEnabled: true,
   kubeletEndpointSliceEnabled: false,
+  watchReferencedObjectsInAllNamespaces: true,
```
slashpai: IIUC from description of the field this should be false by default?

cmd/operator/main.go
```diff
@@ -212,19 +213,21 @@ func run(fs *flag.FlagSet) int {
 	}
 
 	logger.Info("Starting Prometheus Operator", "version", version.Info(), "build_context", version.BuildContext(), "feature_gates", cfg.Gates.String())
+	logger.Info("Operator's configuration",
+		"watch_referenced_objects_in_all_namespaces", cfg.WatchObjectRefsInAllNamespaces,
+		"controller_id", cfg.ControllerID,
+		"enable_config_reloader_probes", cfg.ReloaderConfig.EnableProbes)
```
slashpai, line 219: May be good to add a metric to track reconciliation triggered by referenced objects too?

pkg/operator/informers.go
```diff
@@ -18,12 +18,69 @@ import (
...
+	refResolver RefResolver,
+) FilterFunc {
+	return func(obj interface{}) bool {
+		partialObjMeta, ok := obj.(*metav1.PartialObjectMetadata)
+		if !ok {
+			return false
+		}
+
+		for _, informer := range informerGetter.GetInformers() {
+			workloads, err := informer.Lister().List(labels.Everything())
```
slashpai, line 54: do we need debug log for error here or would it be noisy?

slashpai (commented):
Significant fix 🎉 

I just had a few questions

pkg/prometheus/server/operator.go
```diff
@@ -798,6 +806,7 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	if err := c.createOrUpdateConfigurationSecret(ctx, logger, p, cg, ruleConfigMapNames, assetStore); err != nil {
 		return fmt.Errorf("creating config failed: %w", err)
 	}
+	c.reconciliations.UpdateReferenceTracker(key, assetStore.RefTracker())
```
slashpai, line 809: IIUC we need to call this in Alertmanager and PrometheusAgent too?

jsonnet/prometheus-operator/prometheus-operator.libsonnet
simonpasquier: Correct. If enabled by default, it might create excessive load in existing setups where the operator watches for configuration and workload resources in difference namespaces (e.g. OCP). Still I wanted the E2E tests to run with the flag turned on hence setting the flag to true in the jsonnet and by consequence in `examples/rbac/prometheus-operator/prometheus-operator-deployment.yaml`. But after more thinking, I might as well tweak the operator's deployment in the e2e framework...

My longer-term plan would be to turn it on by default after a few releases.

cmd/operator/main.go
simonpasquier: You mean a different metric than `prometheus_operator_triggered_total`?

pkg/operator/informers.go
simonpasquier: ~~I fear that it would be noisy indeed.~~
Just for error would be ok but we'd need to pass a logger.

pkg/prometheus/server/operator.go
simonpasquier: Good catch!

cmd/operator/main.go
slashpai: I haven't thought much but I was thinking if we could identify a specific secret triggering reconciliation that could help to understand who is culprit incase of excessive API calls? So the metric can have label `triggered by`:`<secret>`. If I am not wrong currently we have say only triggered by which controller through `prometheus_operator_triggered_total`

simonpasquier: having the secret's name & namespace as labels would be too much in terms of cardinality.

slashpai: ya you are right :)

slashpai:
lgtm

PR #7622, opened 2025-06-21
feat: add api url global config

nutmos:
## Description
Add support for API URLs in the global config of the alert manager.

## Type of change

## Verification
In Progress

## Changelog entry

```release-note
- Add support for API URLs in the global configuration for the alert manager for Telegram, Webex
```


pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -505,6 +523,23 @@ type GlobalSMTPConfig struct {
 	TLSConfig *SafeTLSConfig `json:"tlsConfig,omitempty"`
 }
 
+// GlobalWeChatConfig configures global WeChat parameters.
+// See https://prometheus.io/docs/alerting/latest/configuration/#configuration-file
+type GlobalWeChatConfig struct {
+	// WeChat API URL
+	// The default value is "https://qyapi.weixin.qq.com/cgi-bin/"
+	// +optional
+	APIURL *string `json:"apiUrl,omitempty"`
```
simonpasquier, line 532: (nit)
```suggestion
	APIURL *string `json:"apiURL,omitempty"`
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -390,6 +390,24 @@ type AlertmanagerGlobalConfig struct {
...
+	TelegramAPIURL *string `json:"telegramApiUrl,omitempty"`
+
+	// The default Webex API URL.
+	// Default value will be "https://webexapis.com/v1/messages"
+	// +optional
+	WebexAPIURL *string `json:"webexApiUrl,omitempty"`
+
+	// The default Webex API URL.
+	// +optional
+	JiraAPIURL *string `json:"jiraApiUrl,omitempty"`
```
simonpasquier, line 406:
```suggestion
	JiraAPIURL *string `json:"jiraAPIURL,omitempty"`
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -390,6 +390,24 @@ type AlertmanagerGlobalConfig struct {
...
+
+	// The default Telegram API URL.
+	// Default value will be "https://api.telegram.org"
+	// +optional
+	TelegramAPIURL *string `json:"telegramApiUrl,omitempty"`
+
+	// The default Webex API URL.
+	// Default value will be "https://webexapis.com/v1/messages"
+	// +optional
+	WebexAPIURL *string `json:"webexApiUrl,omitempty"`
```
simonpasquier, line 402:
```suggestion
	WebexAPIURL *string `json:"webexAPIURL,omitempty"`
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -390,6 +390,24 @@ type AlertmanagerGlobalConfig struct {
 
 	// The default Pagerduty URL.
 	PagerdutyURL *string `json:"pagerdutyUrl,omitempty"`
+
+	// The default Telegram API URL.
+	// Default value will be "https://api.telegram.org"
+	// +optional
+	TelegramAPIURL *string `json:"telegramApiUrl,omitempty"`
```
simonpasquier, line 397: Let's respect [Kubernetes conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions) on field naming. We're not being consistent in the past but I'd rather follow them from now on.
```suggestion
	TelegramAPIURL *string `json:"telegramAPIURL,omitempty"`
```

simonpasquier (commented):
Can we introduce an URL type which would validate that URL values start with http:// or https://?
Similar to (but without sock5)
https://github.com/prometheus-operator/prometheus-operator/blob/ae005bd59f510e5737a644596ee3d9c3693049a2/pkg/apis/monitoring/v1/types.go#L96

General advice: I would have preferred 1 PR for each field being added (or at least each receiver) as it is easier to review.

simonpasquier:
Also for SMTP we've introduced an intermediate struct so that SMTP global settings are located together. I'd recommend following the same approach even for receivers which only expose 1 global field for now.

pkg/apis/monitoring/v1/alertmanager_types.go
nutmos: This case I reflect slackApiUrl above. That's why it becomes ApiUrl instead of APIURL.

simonpasquier: Yes I totally understand. Unfortunately we didn't always follow the conventions but I'd prefer to stick with them for new stuff.

nutmos: @simonpasquier Okay, got it.

nutmos:
@simonpasquier I'll open new pull requests separately by each group of fields. Will close this when all of them are completed.

PR #7624, opened 2025-06-23
Add OTEL support

denysvitali:
## Description

> [!WARNING]  
> This solution was "vibe coded" - but of course I was always in the loop.

With this change we increase the visibility into the Prometheus Operator(s). As we recently found a nasty bug involving also parts of the Prometheus Operator, we found the lack of tracing / observability into such an important element of our stack to be concerning.

With this change, users will be able to use an OTEL Collector to get telemetry data from the Prometheus Operator.  
  
As the change would become breaking if metrics are introduced (we'll have to switch from the Prometheus metrics to OTEL) - this change wasn't performed as part of this PR. In the future I envision this project to switch the metrics to use the OTEL SDK and guide users on how to set up the metric reader collector so that we maintain the compatibility with the previous Prometheus-based metrics.

## Type of change

## Verification

Running this in our development clusters. No major issues noticed so far.

## Changelog entry

```release-note
Add OTEL SDK, increase tracing
```

### DIFF
--- a/pkg/thanos/rules.go
+++ b/pkg/thanos/rules.go
@@ -43,6 +43,7 @@ func (o *Operator) createOrUpdateRuleConfigMaps(ctx context.Context, t *monitori
 
 	namespaces, err := o.selectRuleNamespaces(t)
 	if err != nil {
+		span.RecordError(err)
 		return nil, err
 	}
 


simonpasquier:
First of all thanks for the PR. I had something similar in the back of mind but didn't have time for implementation.

To make sure that we're all the same page and given that it has an impact on all components, I'd ask for a (small) [design doc](https://prometheus-operator.dev/docs/community/contributing/#proposal-process) first (for instance, what's our testing strategy?). Also I'd recommend splitting the PR into different pieces to make reviews easier.

> With this change we increase the visibility into the Prometheus Operator(s). As we recently found a nasty bug involving also parts of the Prometheus Operator, we found the lack of tracing / observability into such an important element of our stack to be concerning.

Can you expand a bit? Is it related to https://github.com/prometheus-operator/prometheus-operator/pull/7512? How would have tracing helped you?

>  In the future I envision this project to switch the metrics to use the OTEL SDK and guide users on how to set up the metric reader collector so that we maintain the compatibility with the previous Prometheus-based metrics.

This would require a broader discussion. Personally I'm not sure that the migration is worth the effort.

denysvitali:
> I'd ask for a (small) [design doc](https://prometheus-operator.dev/docs/community/contributing/#proposal-process) first

I'll make a separate PR for that :)

> Can you expand a bit? Is it related to https://github.com/prometheus-operator/prometheus-operator/pull/7512? How would have tracing helped you?

Oh no, not at all.

This is something that happaned recently in our prod environment - it's a bit of a concatenation of issues (in multiple components) that caused an outage in the `thanos-ruler`. It was a very hard bug to debug, and the lack of observability in the prometheus-operator caused some delays in the investigation.

The situation was similar to the following:

Team A adds a Sloth `PrometheusServiceLevel` to our GitOps repo on $DATE.
Unfortunately, the name of the `PrometheusServiceLevel` collides with a `PrometheusRule` in the same namespace - which causes ArgoCD and Sloth to compete (every 15m) for the resource. ArgoCD tries to sync the `PrometheusRule` resource every 15m and Sloth replaces it immediately after it realizes that the `PrometheusRule` is replaced.

All went well until $DATE+15d (which made it even harder to find what caused the issue) where Team B removes some PrometheusRules (cleanup). this casued the Prometheus Operator to only use 2 configmaps for the ruler (`thanos-ruler-default-rulefiles-0` and `thanos-ruler-default-rulefiles-1`) instead of the original 3 configmaps (`thanos-ruler-default-rulefiles-0`, `thanos-ruler-default-rulefiles-1`, `thanos-ruler-default-rulefiles-2`).

Since the Thanos ruler was at the threshold of its rules (before creating a new configmap to hold the new ones), every 15 minutes the StatefulSet was updated to have a new configmap (`thanos-ruler-default-rulefiles-2`) and the rulefile configmap was _immediately_ removed from the StatefulSet (since the `PrometheusRule` was "corrected" by Sloth).  
  
This caused periodic pod restarts (`thanos-ruler-1`) which we tried to debug for a while. `prometheus-operator` was just mentioning a Thanos rule sync (w/o mentioning the addition / removal of the CM).

We had to dig into the Kubernetes logs to find out what was happening :upside_down_face: 

TL;DR: A reconciliation issue / concurrency between ArgoCD and Sloth caused one of our `thanos-ruler` pods to be constantly killed (SIGTERM) every 15 minutes - debugging the source of the issue was... very tricky. In the end it was a ConfigMap being added and immediately removed because the new rules didn't fit in 2 configmaps.

> This would require a broader discussion. Personally I'm not sure that the migration is worth the effort.

Not sure either, but it could be very helpful as many are switching from a pull-based approach to a push based approach ([including, somehow, Prometheus)](https://prometheus.io/docs/guides/opentelemetry/)).

internal/telemetry/otel.go
```diff
@@ -0,0 +1,262 @@
...
+		}, nil
+	}
+
+	// Create resource describing this application
+	res, err := resource.New(ctx,
+		resource.WithAttributes(
+			semconv.ServiceName(serviceName),
+			semconv.ServiceVersion(serviceVersion),
+		),
+		resource.WithFromEnv(),   // Merge with environment variables
```
nicolastakashi, line 94: I'm wondering if we shouldn't let this to be enhanced by the collector, this give more flexibility to users on what they want to enhance their telemetry

simonpasquier:
> but it could be very helpful as many are switching from a pull-based approach to a push based approach ([including, somehow, Prometheus)](https://prometheus.io/docs/guides/opentelemetry/)).

The goal is to make Prometheus the best (opensource) backend option for OpenTelemetry but it doesn't mean that pull-based monitoring is going away any time soon.

simonpasquier:
> TL;DR: A reconciliation issue / concurrency between ArgoCD and Sloth caused one of our `thanos-ruler` pods to be constantly killed (SIGTERM) every 15 minutes - debugging the source of the issue was... very tricky. In the end it was a ConfigMap being added and immediately removed because the new rules didn't fit in 2 configmaps.

Thanks for the details. It looks pretty much similar to the issue described in https://github.com/prometheus-operator/prometheus-operator/issues/5085

internal/telemetry/otel.go
```diff
@@ -0,0 +1,262 @@
...
+	// clusterResourcePathRegex matches Kubernetes API paths for cluster-scoped resources with specific names
+	// Example: /api/v1/nodes/my-node -> groups: ["/api/v1", "nodes", "my-node"]
+	// This should NOT match collection URLs like /api/v1/nodes or namespace-related paths
+	clusterResourcePathRegex = regexp.MustCompile(`^(/api/v[^/]+|/apis/[^/]+/v[^/]+)/([^/]+)/([^/]+)(?:/.*)?$`)
+)
+
+// Telemetry holds the telemetry providers and shutdown functions.
+type Telemetry struct {
+	TracerProvider trace.TracerProvider
+	MeterProvider  metric.MeterProvider
```
simonpasquier, line 61: As I commented elsewhere, I'd prefer to defer support for OTEL metrics.

internal/telemetry/otel.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	return nil
+}
+
+// Setup initializes OpenTelemetry with autoexport exporters.
+// It configures both tracing and metrics based on environment variables.
+// See https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/
+func Setup(ctx context.Context, serviceName, serviceVersion string, logger *slog.Logger) (*Telemetry, error) {
+	// Check if OTEL is disabled
+	if os.Getenv("OTEL_SDK_DISABLED") == "true" {
```
simonpasquier, line 80: side-note: I'm surprised that the autoexport package doesn't support the env variable.

internal/telemetry/otel.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// clusterResourcePathRegex matches Kubernetes API paths for cluster-scoped resources with specific names
+	// Example: /api/v1/nodes/my-node -> groups: ["/api/v1", "nodes", "my-node"]
+	// This should NOT match collection URLs like /api/v1/nodes or namespace-related paths
+	clusterResourcePathRegex = regexp.MustCompile(`^(/api/v[^/]+|/apis/[^/]+/v[^/]+)/([^/]+)/([^/]+)(?:/.*)?$`)
+)
+
+// Telemetry holds the telemetry providers and shutdown functions.
+type Telemetry struct {
+	TracerProvider trace.TracerProvider
```
simonpasquier, line 60: Do we even need to store the TracerProvider? I assumed that the PR relies on the global tracer provider.

pkg/alertmanager/operator.go
```diff
@@ -144,7 +149,9 @@ func New(ctx context.Context, restConfig *rest.Config, c operator.Config, logger
 		mclient:    mclient,
 		ssarClient: client.AuthorizationV1().SelfSubjectAccessReviews(),
 
-		logger:   logger,
+		logger: logger,
+		tracer: telemetry.GetComponentTracer("alertmanager-operator", "alertmanager-operator"),
```
simonpasquier, line 153: can we use `controllerName`?

simonpasquier: Also Tracer's name parameter is documented as

> 	// The passed name needs to uniquely identify instrumentation code.
	// Therefore, it is recommended that name is the Go package name of the
	// library providing instrumentation (note: not the code being
	// instrumented). Instrumentation libraries can have multiple versions,
	// therefore, the WithInstrumentationVersion option should be used to
	// distinguish these different codebases. Additionally, instrumentation
	// libraries may sometimes use traces to communicate different domains of
	// workflow data (i.e. using spans to communicate workflow events only). If
	// this is the case, the WithScopeAttributes option should be used to
	// uniquely identify Tracers that handle the different domains of workflow
	// data.


https://pkg.go.dev/go.opentelemetry.io/otel/trace#TracerProvider

Should we use `github.com/prometheus-operator/prometheus-operator`?

pkg/alertmanager/operator.go
```diff
@@ -506,14 +513,22 @@ func (c *Operator) handleNamespaceUpdate(oldo, curo interface{}) {
...
+		span.RecordError(err)
+	}
 	c.reconciliations.SetStatus(key, err)
 
 	return err
-
 }
 
 func (c *Operator) sync(ctx context.Context, key string) error {
+	ctx, span := c.tracer.Start(ctx, "sync")
```
simonpasquier, line 529: for now I'd prefer to minimize the changes. I think that instrumenting the main Sync() method is good enough (especially the error is recorded here and in Sync()).

pkg/alertmanager/operator.go
```diff
@@ -506,14 +513,22 @@ func (c *Operator) handleNamespaceUpdate(oldo, curo interface{}) {
 
 // Sync implements the operator.Syncer interface.
 func (c *Operator) Sync(ctx context.Context, key string) error {
+	ctx, span := c.tracer.Start(ctx, "Sync")
+	defer span.End()
+
 	err := c.sync(ctx, key)
+	if err != nil {
+		span.RecordError(err)
```
simonpasquier, line 521: we should call `span.SetStatus(codes.Error)` too.

internal/telemetry/otel.go
```diff
@@ -0,0 +1,262 @@
...
+		resource.WithHost(),      // Add host attributes
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create OpenTelemetry resource: %w", err)
+	}
+
+	tel := &Telemetry{}
+
+	// Setup trace provider with autoexport
+	traceExporter, err := autoexport.NewSpanExporter(ctx)
```
simonpasquier, line 107: While autoexport has lots of advantages but I've got one complaint against it: if you don't set anything, it will export spans to `http://localhost:4318` which isn't going to work. We need a way to make the export opt-in. One solution would be to default to `none` unless OTEL_TRACES_EXPORTER is defined.

https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_endpoint
https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport#NewSpanExporter

PR #7627, opened 2025-06-23
Feat: Add WeChat in Global Config

nutmos:
## Description
Add global config support in the alert manager for WeChat.

## Type of change

## Verification
Using the unit test for validation.

## Changelog entry

```release-note
- Add support for API URLs in the global configuration for the alert manager for WeChat
```

Relate #7622


pkg/alertmanager/amcfg_test.go
```diff
@@ -901,6 +905,360 @@ func TestInitializeFromAlertmanagerConfig(t *testing.T) {
...
+					},
+				},
+			},
+			matcherStrategy: monitoringv1.AlertmanagerConfigMatcherStrategy{
+				Type: "OnNamespace",
+			},
+			wantErr: true,
+		},
+		{
+			name:      "valid global config telegram api url",
```
AshwinSriram11: It might be better to add Telegram test cases in a separate PR.

pkg/alertmanager/amcfg_test.go
```diff
@@ -901,6 +905,360 @@ func TestInitializeFromAlertmanagerConfig(t *testing.T) {
...
+					},
+				},
+			},
+			matcherStrategy: monitoringv1.AlertmanagerConfigMatcherStrategy{
+				Type: "OnNamespace",
+			},
+			wantErr: true,
+		},
+		{
+			name:      "valid global config jira api url",
```
AshwinSriram11: Same for the jira ones

pkg/alertmanager/amcfg_test.go
```diff
@@ -901,6 +905,360 @@ func TestInitializeFromAlertmanagerConfig(t *testing.T) {
...
+					},
+				},
+			},
+			matcherStrategy: monitoringv1.AlertmanagerConfigMatcherStrategy{
+				Type: "OnNamespace",
+			},
+			wantErr: true,
+		},
+		{
+			name:      "invalid global config wechat config missing secret",
```
AshwinSriram11: We can also have a test case for an empty `APICorpID`

nutmos: @AshwinSriram11 I tried writing some test cases but it seems not to produce error as expected. Is it expected to be tested in E2e?

pkg/alertmanager/amcfg_test.go
nutmos: @AshwinSriram11 Both Telegram and Jira are duplicated from the existing config that was already merged. I removed all of them in the latest comment already.

pkg/alertmanager/amcfg_test.go
AshwinSriram11: I don't think so, maybe we need to validate this manually in the `convertGlobalWeChatConfig` 

Let's wait for @simonpasquier or @slashpai to confirm this

pkg/alertmanager/amcfg.go
```diff
@@ -1766,6 +1770,34 @@ func (cb *ConfigBuilder) convertGlobalJiraConfig(out *globalConfig, in *monitori
...
+
+	if in.APISecret != nil {
+		apiSecret, err := cb.store.GetSecretKey(ctx, crKey.Namespace, *in.APISecret)
+		if err != nil {
+			return fmt.Errorf("failed to get WeChat Secret: %w", err)
+		}
+		out.WeChatAPISecret = apiSecret
+	}
+
+	if in.APICorpID != nil {
```
AshwinSriram11, line 1858: Like I mentioned, maybe we need to do the check here but let's wait

AshwinSriram11 (commented):
Other than the one test case we discussed, everything looks great :+1:

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -557,6 +561,22 @@ type GlobalRocketChatConfig struct {
 	TokenID *v1.SecretKeySelector `json:"tokenID,omitempty"`
 }
 
+type GlobalWeChatConfig struct {
+	// WeChat API URL
```
simonpasquier:
```suggestion
	// The default WeChat API URL.
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -557,6 +561,22 @@ type GlobalRocketChatConfig struct {
...
+	// The default value is "https://qyapi.weixin.qq.com/cgi-bin/"
+	// +optional
+	APIURL *URL `json:"apiURL,omitempty"`
+
+	// WeChat API Secret
```
simonpasquier:
```suggestion
	// The default WeChat API Secret.
```

pkg/apis/monitoring/v1/alertmanager_types.go
```diff
@@ -557,6 +561,22 @@ type GlobalRocketChatConfig struct {
...
+	// +optional
+	APISecret *v1.SecretKeySelector `json:"apiSecret,omitempty"`
+
+	// WeChat API Corporate ID
```
simonpasquier:
```suggestion
	// The default WeChat API Corporate ID.
```

pkg/alertmanager/amcfg.go
```diff
@@ -1806,6 +1810,37 @@ func (cb *ConfigBuilder) convertGlobalRocketChatConfig(ctx context.Context, out
...
+		if *in.APICorpID == "" {
```
simonpasquier: it should be prevented by the kubebuilder validation.

pkg/alertmanager/amcfg.go
```diff
@@ -1806,6 +1810,37 @@ func (cb *ConfigBuilder) convertGlobalRocketChatConfig(ctx context.Context, out
...
 
+func (cb *ConfigBuilder) convertGlobalWeChatConfig(ctx context.Context, out *globalConfig, in *monitoringv1.GlobalWeChatConfig, crKey types.NamespacedName) error {
+	if in == nil {
+		return nil
+	}
+
+	if in.APIURL != nil {
+		u, err := url.Parse(string(*in.APIURL))
+		if err != nil {
+			return fmt.Errorf("parse WeChat API URL: %w", err)
```
simonpasquier:
```suggestion
			return fmt.Errorf("wechat API URL: %w", err)
```

pkg/alertmanager/amcfg.go
nutmos: I'd like to do so but @AshwinSriram11 mentioned before to have a check logic implemented here as well. I'm not sure what is the practice I should follow.

AshwinSriram11: Sorry if I misguided you. I was not sure what would be the correct method and asked for @simonpasquier's advice. I would suggest proceeding with the original method

nutmos: @AshwinSriram11 No problem. I will proceed with the original way.

nutmos:
@slashpai @simonpasquier This is ready to review now.

simonpasquier (approved):
Thanks!

pkg/alertmanager/amcfg_test.go
simonpasquier: Ideally we could test in the e2e tests but the validation rules are simple enough that I wouldn't mandate it. It's required for more complex validations like regex.

Issue #7362, opened 2025-02-21
Support UTF-8 for label names and rules


simonpasquier:
### Component(s)

Prometheus, ThanosRuler

### What is missing? Please describe.

One of the biggest changes in Prometheus v3 is that label and metric names support any UTF-8 character. Right now the operator enforces the legacy validation (e.g. no UTF-8) to avoid breaking Prometheus v2 users.

We want to support UTF-8 label names in the operator and admission webhook but it's unclear how it can be implemented while supporting both pre-v3 and v3 setups.

Impacted areas:
* Relabeling configurations
* Prometheus rule validations (including PromQL)

Note: the operator code base relies on github.com/prometheus/prometheus for rule validation (e.g. `github.com/prometheus-operator/prometheus-operator/model/rulefmt`) but the validation logic is governed by a global variable (`NameValidationScheme`). It works fine for Prometheus which sets the value on startup but it's harder to use in the operator if we want to support both schemes concurrently.

## Environment
Kubernetes Version:
Prometheus-Operator Version:

slashpai:
We had brief discussion on this in last dependency update adding it here for reference https://github.com/prometheus-operator/prometheus-operator/pull/7295#issuecomment-2615855010

simonpasquier:
Current status and plan (after discussing offline with @slashpai).

## UTF-8 label names in operands

* Prometheus >= v3.x
* Thanos >= v0.38.0
*  Alertmanager when operating with `--enable-feature="utf8-strict-mode`.

## API

* Relabel configuration (Prometheus + scrape resources).
* Static target labels (ScrapeConfig).
* External labels (Prometheus).
* Rules (Prometheus + Thanos Ruler + Alertmanager).

We don't have any other choice than relaxing the existing validations on the CRD fields. Right now, the only change would be for relabel configs (to be confirmed) because other fields are validated during reconciliations (if ever).

By definition, all string values exposed via the Kubernetes API are UTF-8 strings. There's no need for additional validation.

## Components

### Operator

Right now the validation relies on the prometheus/common library which has a global variable to decide the scheme. It's expected that the library clients set this variable during their init phase. It doesn't work for the operator (*) because the validation scheme is based on the version of the workload being defined (e.g. a service monitor with utf-8 relabel config is valid for Prometheus v3 but not for a v2 instance). Instead we will duplicate the legacy validation logic in the operator and switch the global validation scheme to utf-8 eventually. 

[...]

PR #7637, opened 2025-06-25
feat: add support for UTF8 labels and rules

slashpai:
Related-to #7362

## Description

## Type of change

## Verification

## Changelog entry

```release-note
feat: add support for UTF8 labels and rules
```


slashpai:
Working on to support in rules

pkg/prometheus/resource_selector.go
```diff
@@ -363,8 +365,14 @@ func (lcv *LabelConfigValidator) validate(rc monitoringv1.RelabelConfig) error {
 		}
 	}
 
-	if action == string(relabel.HashMod) && !model.LabelName(rc.TargetLabel).IsValid() {
-		return fmt.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
+	if action == string(relabel.HashMod) {
+		if !utf8Supported && !model.LabelName(rc.TargetLabel).IsValidLegacy() {
```
slashpai: model.NameValidationScheme will be deprecated in near future hence I am using the method suggested there directly https://github.com/prometheus/common/commit/6b9636ca14ecb6f8bb49d70aea6019c4cef93aab

slashpai:
I am adding a e2e test as well

slashpai:
more changes needed realised after testing e2e locally

pkg/operator/rules.go
```diff
@@ -36,15 +36,6 @@ import (
 	namespacelabeler "github.com/prometheus-operator/prometheus-operator/pkg/namespacelabeler"
 )
 
-func init() {
-	// For now, the operator only supports legacy label names.
-	// Eventually the operator should support UTF-8 label names too and the
-	// issue is tracked by
-	// https://github.com/prometheus-operator/prometheus-operator/issues/7362
-	// nolint:staticcheck
-	model.NameValidationScheme = model.LegacyValidation
```
simonpasquier, line 44: hmm it still has an impact on relabeling configurations, no?

slashpai: We are updating relabel config here https://github.com/prometheus-operator/prometheus-operator/pull/7637/files#diff-aca3d056c83da1ddc448c8c40960960d7ed00ef61f8d5f08a40e165564d74dcdR322

pkg/operator/rules.go
```diff
@@ -81,6 +75,11 @@ func NewPrometheusRuleSelector(ruleFormat RuleConfigurationFormat, version strin
 		return nil, fmt.Errorf("convert rule label selector to selector: %w", err)
 	}
 
+	if componentVersion.LT(semver.MustParse("3.0.0")) {
+		// nolint:staticcheck
+		model.NameValidationScheme = model.LegacyValidation
```
simonpasquier: we can't modify a global variable like this because it may cause race conditions.

pkg/operator/rules.go
simonpasquier: I meant that the LabelName type still has kubebuilder validation restricting to the legacy validation.

simonpasquier (commented):
Options
1. Have a global lock on `model.NameValidationScheme` but it's ugly and brittle. 
2. Expose a CLI flag in the operator to choose between legacy and utf8 validation.
3. Declare that we only support Prometheus > 3.x from now on.


1 and 2 don't solve the issue of validations in the CRD.

pkg/operator/rules.go
slashpai: I will review again 🙂

slashpai:
With https://github.com/prometheus/prometheus/commit/0a40df33fb4e23695a5eff5f387aa7651c5ec8cf#diff-b8c18b1e4cc9c9beec1598127742d0a82c9ffb2805996e1887aaa1021ed90587 it looks to be easier to have the option available in operator so current change is based on that

go.mod
```diff
@@ -25,9 +25,9 @@ require (
 	github.com/prometheus-operator/prometheus-operator/pkg/client v0.85.0
 	github.com/prometheus/alertmanager v0.28.1
 	github.com/prometheus/client_golang v1.23.0
-	github.com/prometheus/common v0.65.0
+	github.com/prometheus/common v0.65.1-0.20250801071412-c79a891c6c28
 	github.com/prometheus/exporter-toolkit v0.14.0
-	github.com/prometheus/prometheus v0.305.0
+	github.com/prometheus/prometheus v0.305.1-0.20250818080900-0a40df33fb4e
```
slashpai, line 31: For utilising the change https://github.com/prometheus/prometheus/commit/0a40df33fb4e23695a5eff5f387aa7651c5ec8cf#diff-b8c18b1e4cc9c9beec1598127742d0a82c9ffb2805996e1887aaa1021ed90587

slashpai:
TODO: relabel validations needed to be updated with logic as in prometheus

simonpasquier:
> it looks to be easier to have the option available in operator so current change is based on that

IIUC the operator would need to gate on the Prometheus version (for any version not having the ability to configure the validation scheme, the option wouldn't apply and the operator needs to compute the validation scheme applying to the version).
Anyway this is way nicer than before.

go.mod
slashpai: unfortunately this change is not included in prometheus 3.6 so had to get commit hash

pkg/admission/admission.go
```diff
@@ -233,7 +234,7 @@ func (a *Admission) validatePrometheusRules(ar v1.AdmissionReview) *v1.Admission
 		return toAdmissionResponseFailure(errUnmarshalRules, prometheusRuleResource, []error{err})
 	}
 
-	errors := promoperator.ValidateRule(promRule.Spec)
+	errors := promoperator.ValidateRule(promRule.Spec, model.UTF8Validation)
```
slashpai: TODO: updating this to be a flag instead of hard coded

pkg/operator/rules.go
```diff
@@ -195,10 +195,24 @@ func ValidateRule(promRuleSpec monitoringv1.PrometheusRuleSpec) []error {
...
 	}
 
-	_, errs := rulefmt.Parse(content, false)
+	_, errs := rulefmt.Parse(content, false, validationScheme)
 	return errs
 }
 
+func (prs *PrometheusRuleSelector) getValidationScheme() model.ValidationScheme {
+	// For Prometheus/PrometheusAgent 3.0.0+, use UTF-8 validation
+	// For ThanosRuler, always use legacy validation (ThanosRuler doesn't support UTF-8 yet)
```
slashpai: TODO: need to confirm this

pkg/admission/admission.go
slashpai: I am not fully sure whether we need update in admission webhook and should allow utf8 by default

simonpasquier: It shouldn't cost us a lot to implement a flag.

pkg/operator/rules.go
```diff
@@ -195,10 +195,24 @@ func ValidateRule(promRuleSpec monitoringv1.PrometheusRuleSpec) []error {
...
+	if prs.ruleFormat == ThanosFormat {
+		return model.LegacyValidation
+	}
+
+	if prs.version.GTE(semver.MustParse("3.0.0")) {
```
simonpasquier: should we factorize with `getValidationSchemeForVersion()` from pkg/prometheus/resource_selector.go?

pkg/operator/rules.go
simonpasquier: https://github.com/thanos-io/thanos/commit/0414eef64d6a3897ae4c09ac2fcb0aa1c1fb9cd1 landed in v0.38.0

pkg/operator/rules_test.go
```diff
@@ -133,29 +133,6 @@ func shouldAcceptRulesWithEmptyDurations(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func shouldRejectRuleWithInvalidLabels(t *testing.T) {
```
simonpasquier, line 136: should we keep this test and set prometheus version to v2.something?

pkg/prometheus/resource_selector.go
```diff
@@ -418,7 +438,7 @@ func (lcv *LabelConfigValidator) validate(rc monitoringv1.RelabelConfig) error {
 		return fmt.Errorf("relabel configuration for %s action needs targetLabel value", rc.Action)
 	}
 
-	if (action == string(relabel.Replace) || action == string(relabel.Lowercase) || action == string(relabel.Uppercase) || action == string(relabel.KeepEqual) || action == string(relabel.DropEqual)) && !relabelTarget.MatchString(rc.TargetLabel) {
+	if (action == string(relabel.Replace) || action == string(relabel.Lowercase) || action == string(relabel.Uppercase) || action == string(relabel.KeepEqual) || action == string(relabel.DropEqual)) && !lcv.isValidLabelName(rc.TargetLabel) {
```
simonpasquier: Wouldn't this break for `"${1}"` which is a valid target label in Prometheus v2?

pkg/prometheus/resource_selector.go
```diff
@@ -363,7 +373,8 @@ func validateScrapeIntervalAndTimeout(p monitoringv1.PrometheusInterface, scrape
 }
 
 type LabelConfigValidator struct {
-	v semver.Version
+	v           semver.Version
+	utf8Allowed bool
```
simonpasquier: (nit) do we need a bool field?

pkg/prometheus/resource_selector_test.go
simonpasquier, line 1: I expect that we extend the unit tests to cover v2 and v3.

pkg/prometheus/resource_selector.go
slashpai: addressed

slashpai: updated validation similar to https://github.com/prometheus/prometheus/blob/747c5ee2b19a9e6a51acfafae9fa2c77e224803d/model/relabel/relabel.go#L120-L186

cmd/admission-webhook/main.go
```diff
@@ -35,20 +36,23 @@ import (
...
+		memlimitRatio        float64
+		nameValidationScheme string
 	)
 
 	server.RegisterFlags(flagset, &serverConfig)
 	versionutil.RegisterFlags(flagset)
 	logging.RegisterFlags(flagset, &logConfig)
 
 	flagset.Float64Var(&memlimitRatio, "auto-gomemlimit-ratio", defaultGOMemlimitRatio, "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory. The value should be greater than 0.0 and less than 1.0. Default: 0.0 (disabled).")
+	flagset.StringVar(&nameValidationScheme, "name-validation-scheme", defaultValidationScheme, "The name validation scheme to use. Default: legacy.")
```
simonpasquier:
```suggestion
	flagset.StringVar(&nameValidationScheme, "name-validation-scheme", defaultValidationScheme, "The name validation scheme to use ('legacy' or 'utf-8'). Default: legacy.")
```

pkg/thanos/rules.go
```diff
@@ -62,6 +62,7 @@ func (o *Operator) createOrUpdateRuleConfigMaps(ctx context.Context, t *monitori
 	logger := o.logger.With("thanos", t.Name, "namespace", t.Namespace)
 	thanosVersion := operator.StringValOrDefault(ptr.Deref(t.Spec.Version, ""), operator.DefaultThanosVersion)
 
+	// Since ThanosRuler doesn't support UTF-8 yet
```
simonpasquier: are we sure?

pkg/operator/rules.go
```diff
@@ -92,6 +89,14 @@ func NewPrometheusRuleSelector(ruleFormat RuleConfigurationFormat, version strin
 	}, nil
 }
 
+// GetValidationSchemeForVersion returns the appropriate validation scheme based on Prometheus version.
+func GetValidationSchemeForVersion(version semver.Version) model.ValidationScheme {
```
simonpasquier: should we make it obvious that it's based on Prometheus version by renaming it to `ValidationSchemeForPrometheus()`? 
I would also recommend moving it outside rules.go because it's not limited to rules.

pkg/operator/rules.go
```diff
@@ -103,7 +108,9 @@ func (prs *PrometheusRuleSelector) generateRulesConfiguration(promRule *monitori
 		return "", fmt.Errorf("failed to marshal content: %w", err)
 	}
 
-	errs := ValidateRule(promRuleSpec)
+	validationScheme := GetValidationSchemeForVersion(prs.version)
```
simonpasquier: It doesn't work for Thanos?

pkg/prometheus/resource_selector.go
```diff
@@ -388,9 +388,16 @@ func (lcv *LabelConfigValidator) Validate(rcs []monitoringv1.RelabelConfig) erro
 	return nil
 }
 
-func (lcv *LabelConfigValidator) validate(rc monitoringv1.RelabelConfig) error {
-	relabelTarget := regexp.MustCompile(`^(?:(?:[a-zA-Z_]|\$(?:\{\w+\}|\w+))+\w*)+$`)
+func varInRegexTemplate(template string) bool {
```
simonpasquier, line 392: can we add a comment that it's copied from upstream Prometheus code base?

pkg/prometheus/resource_selector.go
```diff
@@ -388,9 +388,16 @@ func (lcv *LabelConfigValidator) Validate(rcs []monitoringv1.RelabelConfig) erro
...
+	return strings.Contains(template, "$")
+}
 
+func (lcv *LabelConfigValidator) isValidLabelName(labelName string) bool {
```
simonpasquier, line 396: unrelated to this PR but I think that we move LabelConfigValidator to its own package (right now we have parts of the code that access private fields/methods, it's never a good idea).

pkg/prometheus/resource_selector.go
```diff
@@ -418,21 +425,32 @@ func (lcv *LabelConfigValidator) validate(rc monitoringv1.RelabelConfig) error {
...
 	}
 
-	if (action == string(relabel.Replace) || action == string(relabel.Lowercase) || action == string(relabel.Uppercase) || action == string(relabel.KeepEqual) || action == string(relabel.DropEqual)) && !relabelTarget.MatchString(rc.TargetLabel) {
+	if (action == string(relabel.Replace)) && !varInRegexTemplate(rc.TargetLabel) && !lcv.isValidLabelName(rc.TargetLabel) {
+		return fmt.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
+	}
+
+	isValidLabelNameWithRegexVarFn := func(value string) bool {
+		// UTF-8 allows ${} characters, so standard validation allow $variables by default.
+		// Relabelling users cannot put $ and ${<...>} characters in metric names or values.
```
simonpasquier: I don't understand the comment (I've looked at Prometheus code and don't understand either).

pkg/prometheus/resource_selector.go
```diff
@@ -418,21 +425,32 @@ func (lcv *LabelConfigValidator) validate(rc monitoringv1.RelabelConfig) error {
...
+		return lcv.isValidLabelName(value)
+	}
+	if (action == string(relabel.Replace)) && varInRegexTemplate(rc.TargetLabel) && !isValidLabelNameWithRegexVarFn(rc.TargetLabel) {
+		return fmt.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
```
simonpasquier, line 434: note for later: we mix snake_case and camelCase fields. We should standardize on the latter.

pkg/prometheus/resource_selector.go
slashpai: I have logged issue for this will address in follow-up PR https://github.com/prometheus-operator/prometheus-operator/issues/7951

pkg/prometheus/resource_selector.go
slashpai: updated to simplify it

pkg/prometheus/resource_selector.go
slashpai: https://github.com/prometheus-operator/prometheus-operator/issues/7952

slashpai:
@simonpasquier addressed comments please take a look

cmd/admission-webhook/main.go
```diff
@@ -35,20 +36,23 @@ import (
...
+		memlimitRatio        float64
+		nameValidationScheme string
 	)
 
 	server.RegisterFlags(flagset, &serverConfig)
 	versionutil.RegisterFlags(flagset)
 	logging.RegisterFlags(flagset, &logConfig)
 
 	flagset.Float64Var(&memlimitRatio, "auto-gomemlimit-ratio", defaultGOMemlimitRatio, "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory. The value should be greater than 0.0 and less than 1.0. Default: 0.0 (disabled).")
+	flagset.StringVar(&nameValidationScheme, "name-validation-scheme", defaultValidationScheme, "The name validation scheme to use ('legacy' or 'utf8'). Default: legacy.")
```
simonpasquier: my bad, the default value is already printed.
```suggestion
	flagset.StringVar(&nameValidationScheme, "name-validation-scheme", defaultValidationScheme, "The name validation scheme to use ('legacy' or 'utf8').")
```

pkg/admission/testdata/goodRulesWithUTF8.golden
```diff
@@ -0,0 +1,70 @@
...
+                "for": "5m",
+                "labels": {
+                  "severity": "critical",
+                  "team": "platform",
+                  "测试标签": "测试_label_with_中文"
+                }
+              },
+              {
+                "record": "http_requests_rate_sum",
+                "expr": "sum(rate(http_requests_total[5m])) by (service, job)",
```
simonpasquier: can we also change the expression to use utf8 metric name/labels?
```suggestion
                "expr": 'sum by(service,job,"http.request") (rate({"http.requests.total","error.type"!=""}[5m]))',
```

pkg/admission/testdata/goodRulesWithUTF8.golden
```diff
@@ -0,0 +1,70 @@
...
+                "expr": "up{app=\"myapp\",service=\"测试服务\"} > 0",
+                "for": "5m",
+                "labels": {
+                  "severity": "critical",
+                  "team": "platform",
+                  "测试标签": "测试_label_with_中文"
+                }
+              },
+              {
+                "record": "http_requests_rate_sum",
```
simonpasquier:
```suggestion
                "record": "http.requests.rate_5m",
```

pkg/operator/validations.go
```diff
@@ -0,0 +1,36 @@
...
+func ValidationSchemeForPrometheus(version semver.Version) model.ValidationScheme {
+	if version.GTE(semver.MustParse("3.0.0")) {
+		return model.UTF8Validation
+	}
+	return model.LegacyValidation
+}
+
+// ValidationSchemeForThanos returns the appropriate validation scheme based on Thanos version.
+func ValidationSchemeForThanos(version semver.Version) model.ValidationScheme {
+	if version.GTE(semver.MustParse("0.39.0")) {
```
simonpasquier: https://github.com/thanos-io/thanos/commit/0414eef64d6a3897ae4c09ac2fcb0aa1c1fb9cd1 shows that it was shipped starting with v0.38

go.mod
simonpasquier: It's a bit unfortunate but can we add a comment at least?

slashpai:
I have created change for externalLabels utf-8 support in separate branch will have PR once this is accepted :)

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,280 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+	})
+}
+
+func testUTF8RelabelConfigWithPrometheus3(t *testing.T, ns string) {
+	name := "prometheus-utf8-relabel"
+
+	prom := framework.MakeBasicPrometheus(ns, name, "test-app", 1)
+	prom.Spec.Version = operator.DefaultPrometheusVersion
+
+	_, err := framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, prom)
```
simonpasquier: It's safer/faster to create the Prometheus resource last because it ensures that the generated config is up-to-date.

I would also recommend testing with a PrometheusRule resource.

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,280 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	_, err = framework.MonClientV1.ServiceMonitors(ns).Create(context.Background(), sm, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	// Verify that UTF-8 labels appear in the config with Prometheus 3.0+
+	err = wait.PollUntilContextTimeout(context.Background(), time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
```
simonpasquier: rather than checking the config can we configure a service monitor that scrapes a real target (e.g. Prometheus), query the metric in Prometheus and check that we get the expected labels?

simonpasquier (commented):
For the e2e tests, I would suggest that we don't check the config but rather the expected output (e.g. query metrics).

pkg/operator/rules.go
simonpasquier, line 42: the warning will go away if you rebase

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,83 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+	}
+	_, err = framework.MonClientV1.ServiceMonitors(ns).Create(context.Background(), sm, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	prom := framework.MakeBasicPrometheus(ns, name, "test-app", 1)
+	prom.Spec.Version = operator.DefaultPrometheusVersion
+	_, err = framework.CreatePrometheusAndWaitUntilReady(context.Background(), ns, prom)
+	require.NoError(t, err)
+
+	err = framework.WaitForDiscoveryWorking(context.Background(), ns, service.Name, name)
```
simonpasquier: do we need this?

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,83 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+					Action:       "replace",
+				}},
+			}},
+		},
+	}
+	_, err = framework.MonClientV1.ServiceMonitors(ns).Create(context.Background(), sm, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	prom := framework.MakeBasicPrometheus(ns, name, "test-app", 1)
+	prom.Spec.Version = operator.DefaultPrometheusVersion
```
simonpasquier: shouldn't be necessary since we default to this version

test/e2e/prometheus_test.go
slashpai: ok looks like overkill for this test as we are doing self scraping in this test, will remove this

pkg/admission/admission.go
```diff
@@ -166,7 +169,10 @@ func (a *Admission) serveAdmission(w http.ResponseWriter, r *http.Request, admit
 		responseAdmissionReview.Response = admit(requestedAdmissionReview)
 	}
 
-	responseAdmissionReview.Response.UID = requestedAdmissionReview.Request.UID
+	// Adding nil check here since request can be nil if deserialization fails.
```
slashpai, line 172: I added the check when my one of the testcase failed due to invalid json , so even if deserialisation fails request was accessed and it was causing nil pointer exception

after adding exception till json was fixed it showed exact error below

```
admission_test.go:521: UTF-8 rules should pass with UTF-8 validation: Expected admission to be allowed, but it was rejected. Message: Unable to deserialize request
```

slashpai:
@simonpasquier should be good now. Please review

pkg/admission/admission.go
simonpasquier: ok so it's a pre-existing logical flaw which we can fix outside this PR?

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,79 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+	t.Parallel()
+
+	testCtx := framework.NewTestCtx(t)
+	defer testCtx.Cleanup(t)
+	ns := framework.CreateNamespace(context.Background(), t, testCtx)
+	framework.SetupPrometheusRBAC(context.Background(), t, testCtx, ns)
+
+	name := "prometheus-utf8-test"
+
+	// Create service with UTF8 labels
```
simonpasquier: see https://github.com/prometheus-operator/prometheus-operator/pull/7955. After it merges, I can push an updated version of instrumented-app which exposes a metric with dots so we can test with a "real" example.

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,79 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+		Spec: v1.ServiceSpec{
+			Ports:    []v1.ServicePort{{Name: "web", Port: 9090}},
+			Selector: map[string]string{"app": "prometheus", "prometheus": name},
+		},
+	}
+	_, err := framework.KubeClient.CoreV1().Services(ns).Create(context.Background(), service, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	// Create ServiceMonitor that uses UTF8 labels in selectors and relabeling
+	sm := &monitoringv1.ServiceMonitor{
```
simonpasquier, line 5567: can we also test with a PrometheusRule and utf-8 (dots) label/annotation names?

pkg/admission/admission.go
slashpai: yes correct

slashpai:
@simonpasquier PTAL

pkg/prometheus/resource_selector_test.go
```diff
@@ -176,11 +176,11 @@ func TestValidateRelabelConfig(t *testing.T) {
 		},
 		// Test valid labelmap relabel config with replacement specified
 		{
-			scenario: "valid labelmap config",
+			scenario: "valid labelmap config with replacement",
 			relabelConfig: monitoringv1.RelabelConfig{
 				Action:      "labelmap",
 				Regex:       "__meta_kubernetes_service_label_(.+)",
-				Replacement: ptr.To("${2}"),
```
simonpasquier, line 183: do we test this combination elsewhere?

pkg/admission/admission.go
simonpasquier: ok we may want to return early if the request isn't valid JSON but let's address in a follow-up.

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,220 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Wait for target discovery
+	err = framework.WaitForDiscoveryWorking(context.Background(), ns, promSVC.Name, name)
+	require.NoError(t, err)
+
+	// Verify UTF8 labels work in queries.
+	err = wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
+		// Query for metric with UTF8 relabeled label - using sample app metrics
+		results, err := framework.PrometheusQuery(ns, promSVC.Name, "http", `app_info{service_clustér_label="dev"}`)
```
simonpasquier: hmm I would expect that we query `{"app.info","service_clustér_label"="dev"}`.

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,220 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+				"π.value":              "3.14159",
+			},
+		},
+		Spec: monitoringv1.PrometheusRuleSpec{
+			Groups: []monitoringv1.RuleGroup{{
+				Name: "utf8.test.rules",
+				Rules: []monitoringv1.Rule{
+					{
+						Alert: "UTF8TestAlert",
+						Expr:  intstr.FromString(`app_info{service_clustér_label="dev"} == 0`),
```
simonpasquier: same here

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,220 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+					},
+				},
+			}},
+		},
+	}
+	_, err = framework.MonClientV1.PrometheusRules(ns).Create(context.Background(), prometheusRule, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	// Verify UTF8 labels work in queries and rules.
+	err = wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
```
simonpasquier, line 5689: can we split the 4 checks into individual polls?

pkg/prometheus/resource_selector_test.go
slashpai: we were validating with regex before but I see its been updated and it doesn't allow `$` for label map https://github.com/prometheus/prometheus/blob/main/model/relabel/relabel.go#L159-L161

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,231 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Wait for target discovery
+	err = framework.WaitForDiscoveryWorking(context.Background(), ns, promSVC.Name, name)
+	require.NoError(t, err)
+
+	// Verify UTF8 labels work in queries.
+	err = wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
+		// Query for metric with UTF8 relabeled label - using sample app metrics
+		results, err := framework.PrometheusQuery(ns, promSVC.Name, "http", `app.info{service_clustér_label="dev"}`)
```
simonpasquier: there's something broken in the test logic because the expression is invalid (it should be `{"app.info","service_clustér_label="dev"}`). ~~Before fixing it we need to understand why the check passes while it should fail.~~ edit: I think that we should remove `skipPrometheusTests(t)` at the beginning of the tests.

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,231 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func testPrometheusUTF8LabelSupport(t *testing.T) {
+	skipPrometheusTests(t)
```
simonpasquier:
```suggestion
```

simonpasquier:
So we have a chicken-and-egg problem because the current CRDs don't allow for utf-8 labels. I'm fine merging this PR as it is and have the end-to-end tests in the follow-up PR which relaxes the validations.

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+	promSVC := framework.MakePrometheusService(name, "test-app", v1.ServiceTypeClusterIP)
+	_, err = framework.CreateOrUpdateServiceAndWaitUntilReady(context.Background(), ns, promSVC)
+	require.NoError(t, err)
+
+	// Verify UTF8 metrics work in queries.
+	err = wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
+		// Query for UTF8 metric
+		results, err := framework.PrometheusQuery(ns, promSVC.Name, "http", `{"app.info"}`)
+		if err != nil {
+			t.Logf("UTF8 PromQL Results: %v", results)
```
simonpasquier: If err isn't nil then we don't need to print the results?

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+			t.Logf("UTF8 query failed: %v", err)
+			return false, nil
+		}
+
+		return len(results) > 0, nil
```
simonpasquier: (suggestion)
```suggestion
		if len(results) == 0 {
				t.Log("got no result")
				return false, nil
		}

		return true, nil
```

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+			},
+		},
+		Spec: monitoringv1.PrometheusRuleSpec{
+			Groups: []monitoringv1.RuleGroup{{
+				Name: "utf8.test.rules",
+				Rules: []monitoringv1.Rule{
+					{
+						Alert: "UTF8TestAlert",
+						Expr:  intstr.FromString(`{"app.info"} == 0`),
+						For:   ptr.To(monitoringv1.Duration("1m")),
```
simonpasquier:
```suggestion
```

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+				"description": "Test rule",
+			},
+		},
+		Spec: monitoringv1.PrometheusRuleSpec{
+			Groups: []monitoringv1.RuleGroup{{
+				Name: "utf8.test.rules",
+				Rules: []monitoringv1.Rule{
+					{
+						Alert: "UTF8TestAlert",
+						Expr:  intstr.FromString(`{"app.info"} == 0`),
```
simonpasquier: (suggestion) to have the alert always active + return a utf8 label name.
```suggestion
						Expr:  intstr.FromString(`count by("app.version") ({"app.info"}`),
```

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+						For:   ptr.To(monitoringv1.Duration("1m")),
+						Labels: map[string]string{
+							"severity": "warning",
```
simonpasquier: I suppose that we can also set utf8 label names here?
```suggestion
							"severity": "warning",
							"service.name": "web",
```

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+			t.Logf("UTF8 PromQL Results: %v", results)
+			t.Logf("UTF8 query failed: %v", err)
+			return false, nil
+		}
+
+		return len(results) > 0, nil
+	})
+	require.NoError(t, err, "UTF-8 metrics should work in Prometheus 3.0+ queries")
+
+	// Create PrometheusRule with UTF-8 metrics.
```
simonpasquier, line 5615: (nit) we can create the rule before Prometheus so it's evaluated immediately.

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+							"severity": "warning",
+						},
+						Annotations: map[string]string{
+							"summary":             "Service is down",
+							"description.cluster": "The cluster service is not responding",
+							"runbook":             "https://runbook.example.com/cluster",
+						},
+					},
+					{
+						Record: "cluster:app_info:5m",
```
simonpasquier: let's go all-in on utf8 :)
```suggestion
						Record: "cluster.app_info.5m",
```

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,210 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+		},
+	}
+	_, err = framework.MonClientV1.PrometheusRules(ns).Create(context.Background(), prometheusRule, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	// Check UTF8 recording rule from PrometheusRule
+	err = wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
+		results, err := framework.PrometheusQuery(ns, promSVC.Name, "http", `cluster:app_info:5m`)
+		if err != nil {
+			t.Logf("PromQL Results: %v", results)
```
simonpasquier: same comments here than with the metric query

test/e2e/prometheus_test.go
slashpai: I was printing that while test was failing to see, will remove

test/e2e/prometheus_test.go
```diff
@@ -5495,6 +5495,217 @@ func testPrometheusReconciliationOnSecretChanges(t *testing.T) {
...
+	require.NoError(t, err, "UTF-8 PrometheusRule recording rule should work")
+
+	// Verify the alert rule exists in Prometheus
+	err = wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
+		_, err := framework.PrometheusQuery(ns, promSVC.Name, "http", `ALERTS{alertname="UTF8TestAlert"}`)
+		if err != nil {
+			t.Logf("UTF8 alert rule query failed: %v", err)
+			return false, nil
+		}
+		// Alert might not be firing, so we just check that the query doesn't error
```
simonpasquier: hmm the alert should be active?

simonpasquier (approved):

test/e2e/prometheus_test.go
slashpai: fixed

simonpasquier (dismissed):

slashpai:
I was finally able to fix my podman issue to test the change locally as depending CI was taking time, I am debugging the root cause for failure locally

slashpai:
It should pass now it was issue of promql query it should have been `{"cluster.app_info:5m"}` 🤦🏻‍♀️

simonpasquier (approved):
thanks!

Issue #7638, opened 2025-06-25
Support all proxy configuration fields for ServiceMonitor


simonpasquier:
### Component(s)

ServiceMonitor

### What is missing? Please describe.

Unlike the ScrapeConfig CRD, ServiceMonitor supports only proxyUrl. It should be _fairly_ straightforward to implement all proxy options.

see also https://github.com/prometheus-operator/prometheus-operator/issues/7639 and https://github.com/prometheus-operator/prometheus-operator/issues/7640

## Environment
Kubernetes Version:
Prometheus-Operator Version:

AshwinSriram11:
I would like to work on this

PR #7647, opened 2025-06-27
Support ProxyConfig fields for ServiceMonitor

AshwinSriram11:
## Description

Related issue: #7638 

Previously, ServiceMonitor only supported the `proxyUrl` field. This PR enhances the ServiceMonitor CRD by adding support for all proxy configuration options, similar to what is already available in the ScrapeConfig CRD. 

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/Documentation/api-reference/api.md
+++ b/Documentation/api-reference/api.md
@@ -9855,8 +9855,51 @@ string
 </td>
 <td>
 <em>(Optional)</em>
-<p><code>proxyURL</code> configures the HTTP Proxy URL (e.g.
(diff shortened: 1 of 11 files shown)


simonpasquier (approved):
Thanks!

simonpasquier:
Note that we're technically introducing a breaking change because the proxyURL field now has validation on the value (should start with "http", "https" or "sock5") but I think that it's fine (as long as advertised in the changelog) because invalid values would be rejected today.

PR #7658, opened 2025-06-30
feat: Add StackitSDConfig

heliapb:
## Description

Add support for new `stackit_sd_config` - https://github.com/prometheus/prometheus/pull/16401, from [prometheus v3.5.0-rc.0](https://github.com/prometheus/prometheus/releases/tag/v3.5.0-rc.0)

## Type of change

## Verification

## Changelog entry

```release-note

```


pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -223,6 +223,9 @@ type ScrapeConfigSpec struct {
 	// IonosSDConfigs defines a list of IONOS service discovery configurations.
 	// +optional
 	IonosSDConfigs []IonosSDConfig `json:"ionosSDConfigs,omitempty"`
+	//StackitSDConfigs defines a list of STACKIT service discovery configurations.
```
AshwinSriram11:
```suggestion
	// StackitSDConfigs defines a list of STACKIT service discovery configurations.
```

pkg/prometheus/resource_selector_test.go
```diff
@@ -4580,6 +4580,148 @@ func TestSelectScrapeConfigs(t *testing.T) {
...
+							},
+						},
+					},
+				}
+			},
+			selected:    true,
+			promVersion: "3.5.0",
+		},
+		{
+			scenario: "StackitSDconfig with invalid proxy settings",
```
AshwinSriram11:
```suggestion
			scenario: "StackitSDconfig with invalid proxy config with proxyurl defined and proxy from environment set to true",
```

AshwinSriram11: We can add more cases: invalid proxyurl, invalid secret in proxy connect header

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1464,3 +1467,58 @@ type IonosSDConfig struct {
...
+	// The STACKIT endpoint.
+	// +kubebuilder:validation:Pattern:="^http(s)?://.+$"
+	// +optional
+	Endpoint *string `json:"endpoint"`
+	// Default port to scrape metrics from.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	// +optional
+	Port *int32 `json:"port,omitempty"`
+	// The STACKIT Raw private key string used for authenticating a service account.
```
AshwinSriram11:
```suggestion
	// Raw private key string used for authenticating a service account.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1464,3 +1467,58 @@ type IonosSDConfig struct {
...
+	// +optional
+	PrivateKey *string `json:"privateKey"`
+	// The STACKIT Path to a file containing the raw private key string.
```
AshwinSriram11:
```suggestion
	// Path to a file containing the raw private key string.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1464,3 +1467,58 @@ type IonosSDConfig struct {
...
+	// Full JSON-formatted service account key used for authentication.
+	// +optional
+	ServiceAccountKey *string `json:"serviceAccountKey"`
+	// Path to a file containing the JSON-formatted service account key.
+	// +optional
+	ServiceAccountKeyPath *string `json:"serviceAccountKeyPath"`
+	// Path to a file containing STACKIT credentials.
+	// +optional
+	CredentialsFilePath *string `json:"credentialsFilePath"`
+	// RefreshInterval configures the refresh interval at which Prometheus will re-read the instance list.
```
AshwinSriram11:
```suggestion
	// RefreshInterval configures the time after which the servers are refreshed.
```

AshwinSriram11: https://prometheus.io/docs/prometheus/3.5/configuration/configuration/#stackit_sd_config:~:text=%3Cstring%3E%20%5D-,%23%20The%20time%20after%20which%20the%20servers%20are%20refreshed.,-%5B%20refresh_interval%3A Just to be consistent with the docs in Prometheus :slightly_smiling_face:

AshwinSriram11 (commented):
I think we can just have more test cases in `resource_selector_test.go` for other fields like `oauth2` covering valid and invalid cases.

heliapb:
> I think we can just have more test cases in `resource_selector_test.go` for other fields like `oauth2` covering valid and invalid cases.

Hi @AshwinSriram11 thanks for the feedback, added more tests and fixed documentation as suggested

pkg/prometheus/resource_selector_test.go
```diff
@@ -4588,6 +4588,231 @@ func TestSelectScrapeConfigs(t *testing.T) {
...
+										Name: "secret",
+									},
+								},
+							},
+						},
+					},
+				}
+			},
+			valid:       false,
+			promVersion: "3.2.0",
```
AshwinSriram11:
```suggestion
			promVersion: "3.5.0",
```

I would rather keep the supported version just to make sure the test fails due to invalid CA data

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	// Path to a file containing the JSON-formatted service account key.
+	// +optional
+	ServiceAccountKeyPath *string `json:"serviceAccountKeyPath"`
+	// Path to a file containing STACKIT credentials.
+	// +optional
+	CredentialsFilePath *string `json:"credentialsFilePath"`
+	// RefreshInterval configures the time after which the servers are refreshed.
+	// +optional
+	RefreshInterval *v1.Duration `json:"refreshInterval,omitempty"`
+	// Authorization` header configuration, required when using STACKIT.
```
AshwinSriram11:
```suggestion
	// Authorization header configuration, required when using STACKIT.
```

AshwinSriram11 (changes requested):

pkg/prometheus/resource_selector_test.go
heliapb: thanks make sense, fixed

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+
+// StackitSDConfig configurations enables service discovery for STACKIT.
+// See https://prometheus.io/docs/prometheus/3.5/configuration/configuration/#stackit_sd_config
+type StackitSDConfig struct {
+	// The STACKIT Project ID.
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Project string `json:"project"`
+	// The STACKIT region.
+	// +optional
```
mviswanathsai:
```suggestion
	// +required
```
Aside from having the `omitempty` tag, `Region` seems like a required field. https://github.com/prometheus/prometheus/blob/main/discovery/stackit/stackit.go#L106

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	Region *string `json:"region,omitempty"`
+	// The STACKIT endpoint.
+	// +kubebuilder:validation:Pattern:="^http(s)?://.+$"
+	// +optional
```
mviswanathsai:
```suggestion
	// +required
```
Seems like Endpoint is required, in spite of having a `omitempty` tag in the upstream repo.  https://github.com/prometheus/prometheus/blob/main/discovery/stackit/stackit.go#L106

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
 	// +optional
 	OAuth2 *v1.OAuth2 `json:"oauth2,omitempty"`
 }
+
+// StackitSDConfig configurations enables service discovery for STACKIT.
+// See https://prometheus.io/docs/prometheus/3.5/configuration/configuration/#stackit_sd_config
+type StackitSDConfig struct {
```
mviswanathsai, line 1532: Can we add the `omitempty` tag to all the optional fields?

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	// The STACKIT Project ID.
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Project string `json:"project"`
+	// The STACKIT region.
+	// +optional
+	Region *string `json:"region,omitempty"`
```
mviswanathsai:
```suggestion
	Region *string `json:"region"`
```

AshwinSriram11 (approved):
LGTM

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	// The STACKIT endpoint.
+	// +kubebuilder:validation:Pattern:="^http(s)?://.+$"
+	// +optional
+	Endpoint *string `json:"endpoint"`
+	// Default port to scrape metrics from.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	// +optional
+	Port *int32 `json:"port,omitempty"`
+	// Raw private key string used for authenticating a service account.
```
mviswanathsai:
```suggestion
	// Raw private key string used for authenticating a service account.
	// +kubebuilder:validation:MinLength=1
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	// +optional
+	PrivateKey *string `json:"privateKey"`
+	// Path to a file containing the raw private key string.
+	// +optional
+	PrivateKeyPath *string `json:"privateKeyPath"`
```
mviswanathsai: We would prefer to not have fields that point to filepaths in the config resources. 
```suggestion
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	// Full JSON-formatted service account key used for authentication.
+	// +optional
+	ServiceAccountKey *string `json:"serviceAccountKey"`
+	// Path to a file containing the JSON-formatted service account key.
+	// +optional
+	ServiceAccountKeyPath *string `json:"serviceAccountKeyPath"`
```
mviswanathsai:
```suggestion
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	// Path to a file containing STACKIT credentials.
+	// +optional
+	CredentialsFilePath *string `json:"credentialsFilePath"`
```
mviswanathsai:
```suggestion
```
What does the credentials file contain? If it's just the private key and service account key then we don't need to support it IMO.

Judging by this: https://registry.terraform.io/providers/stackitcloud/stackit/0.9.0/docs and https://github.com/stackitcloud/stackit-sdk-python
I think that's the case. We should be able to remove this field without any issues.

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,58 @@ type IonosSDConfig struct {
...
+	// RefreshInterval configures the time after which the servers are refreshed.
+	// +optional
+	RefreshInterval *v1.Duration `json:"refreshInterval,omitempty"`
+	// Authorization header configuration, required when using STACKIT.
```
mviswanathsai: https://github.com/prometheus/prometheus/blob/main/discovery/stackit/stackit.go#L114
The validation on the client config seems to be the same as we have with the other SDs. So I don't see why the Authorization needs to be a required field. Could you point me to where this is specified?

mviswanathsai (changes requested):

heliapb:
@mviswanathsai thanks fixed the issues.

slashpai:
@mviswanathsai we cannot add this in this release as it needs prometheus 3.5

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,52 @@ type IonosSDConfig struct {
...
+	// +required
+	Project string `json:"project"`
+	// The STACKIT region.
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Region string `json:"region"`
+	// The STACKIT endpoint.
+	// +kubebuilder:validation:Pattern:="^http(s)?://.+$"
+	// +required
+	Endpoint *string `json:"endpoint"`
```
slashpai: If this is required field it shouldn't be pointer

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,52 @@ type IonosSDConfig struct {
 	// +optional
 	OAuth2 *v1.OAuth2 `json:"oauth2,omitempty"`
 }
+
+// StackitSDConfig configurations enables service discovery for STACKIT.
+// See https://prometheus.io/docs/prometheus/3.5/configuration/configuration/#stackit_sd_config
+type StackitSDConfig struct {
+	// The STACKIT Project ID.
```
slashpai:
```suggestion
	// STACKIT Project ID.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,52 @@ type IonosSDConfig struct {
...
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Project string `json:"project"`
+	// The STACKIT region.
```
slashpai:
```suggestion
	// STACKIT region to use.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,52 @@ type IonosSDConfig struct {
...
+	// +kubebuilder:validation:MinLength:=1
+	// +required
+	Region string `json:"region"`
+	// The STACKIT endpoint.
```
slashpai:
```suggestion
	// Custom API endpoint to be used.
```

pkg/apis/monitoring/v1alpha1/scrapeconfig_types.go
```diff
@@ -1469,3 +1472,52 @@ type IonosSDConfig struct {
...
+	// +kubebuilder:validation:Pattern:="^http(s)?://.+$"
+	// +required
+	Endpoint *string `json:"endpoint"`
+	// Default port to scrape metrics from.
```
slashpai:
```suggestion
	// Port to scrape metrics from.
```

pkg/prometheus/resource_selector.go
```diff
@@ -1441,3 +1445,33 @@ func (rs *ResourceSelector) validateIonosSDConfigs(ctx context.Context, sc *moni
 	}
 	return nil
 }
+
+func (rs *ResourceSelector) validateStackitSDConfigs(ctx context.Context, sc *monitoringv1alpha1.ScrapeConfig) error {
+	if len(sc.Spec.StackitSDConfigs) == 0 {
```
slashpai, line 1430: I guess we don't need this check?

slashpai (commented):
a few nits, lgtm otherwise

pkg/prometheus/resource_selector.go
heliapb: Hi thanks fixed the docs and pointers. About this I added this check since since other scrape configs types seemed to have a similar validation done in https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/resource_selector.go#L666.

slashpai: I am not sure if I fully understand. Did you mean other validate() methods?

heliapb: sorry @slashpai my bad you meant the` len check`, removed, pls review

heliapb: I guess removing this broke the tests, at this point to exactly sure why, will look into it

pkg/prometheus/resource_selector.go
```diff
@@ -1457,3 +1461,33 @@ func (rs *ResourceSelector) validateIonosSDConfigs(ctx context.Context, sc *moni
...
```
heliapb, line 1430: hi @slashpai I guess removing this check as you recommended here https://github.com/prometheus-operator/prometheus-operator/pull/7658#discussion_r2227922508 is causing other tests to fail for example.

```
=== RUN   TestSelectScrapeConfigs/DigitalOcean_SD_config_with_valid_TLS_Config
time=2025-08-25T10:26:22.035+01:00 level=WARN msg="skipping object" kind=ScrapeConfig error="StackitSDConfigs: StackitSDConfig is only supported for Prometheus version >= 3.5.0" object=test/test
    resource_selector_test.go:4897:
        	Error Trace:	/Users/helbar1/workspace/opensource/prometheus-operator/pkg/prometheus/resource_selector_test.go:4897
        	Error:      	"map[]" should have 1 item(s), but has 0
        	Test:       	TestSelectScrapeConfigs/DigitalOcean_SD_config_with_valid_TLS_Config
```
And it seems from what I tested thus far the only way to ensure the not needing to do changes in the way `checkScrapeConfig`, since it seems to validate all sdconfigs even if they are not defined. Regardless if you have a another suggestion about how to fix this because  the other way I see to fix this would be to add a check in `checkScrapeConfig` as

```go

func (rs *ResourceSelector) checkScrapeConfig(ctx context.Context, sc *monitoringv1alpha1.ScrapeConfig) error {
(...)
if len(sc.Spec.StackitSDConfigs) > 0 {
    if err := rs.validateStackitSDConfigs(ctx, sc); err != nil {
        return fmt.Errorf("StackitSDConfigs: %w", err)
    }
}
```
 Or as currently I have 

```go
	if len(sc.Spec.StackitSDConfigs) == 0 {
		return nil
	}
```

heliapb:
Hi @slashpai @AshwinSriram11 and @mviswanathsai if you could check again pls. Thanks

Issue #7639, opened 2025-06-25
Support all proxy configuration fields for Probe


simonpasquier:
### Component(s)

Probe

### What is missing? Please describe.

Unlike the ScrapeConfig CRD, Probe supports only proxyUrl. It should be fairly straightforward to implement all proxy options.

see also https://github.com/prometheus-operator/prometheus-operator/issues/7640 and https://github.com/prometheus-operator/prometheus-operator/issues/7638

## Environment
Kubernetes Version:
Prometheus-Operator Version:

PR #7660, opened 2025-07-01
Support ProxyConfig fields for Probe

AshwinSriram11:
## Description

Related Issue: #7639 

Previously, Probe only supported the `proxyUrl` field. This PR enhances the Probe CRD by adding support for all proxy configuration options. Similar to #7647 and #7652

## Type of change

## Verification

## Changelog entry

```release-note

```


AshwinSriram11:
Now that ProxyURL is replaced in all the resources, I think we need to remove `validateProxyURL` function https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/resource_selector.go#L578-L585 . Should I open a separate PR for that or remove it in this one? Wdyt @simonpasquier @slashpai

nicolastakashi:
> Now that ProxyURL is replaced in all the resources, I think we need to remove `validateProxyURL` function https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/resource_selector.go#L578-L585 . Should I open a separate PR for that or remove it in this one? Wdyt @simonpasquier @slashpai

I would be fine removing on this one

simonpasquier:
yeah, it's fine to remove in this PR if it's dead code.

simonpasquier:
in fact golangci-lint already asks for removal :sweat_smile:

pkg/prometheus/resource_selector.go
```diff
@@ -564,7 +564,7 @@ func (rs *ResourceSelector) checkProbe(ctx context.Context, probe *monitoringv1.
 		}
 	}
 
-	if err := validateProxyURL(&probe.Spec.ProberSpec.ProxyURL); err != nil {
+	if err := addProxyConfigToStore(ctx, probe.Spec.ProberSpec.ProxyConfig, rs.store, probe.GetNamespace()); err != nil {
 		return fmt.Errorf("proxyURL: %w", err)
```
simonpasquier:
```suggestion
		return fmt.Errorf("proxy configuration: %w", err)
```

simonpasquier (approved):

Issue #7629, opened 2025-06-23
feat: add option to enable `always_scrape_classic_histograms` globally


skoenig:
### What is missing? Please describe.

When https://github.com/prometheus/prometheus/pull/16452 is released, it would be beneficial to add a global configuration option for `always_scrape_classic_histograms` in Prometheus Operator. This feature complements `convert_classic_histograms_to_nhcb` (implemented in https://github.com/prometheus-operator/prometheus-operator/pull/7448) and would be useful for migrating from classic to native histograms.

### Describe alternatives you've considered.

AFAIK it is possible to use `always_scrape_classic_histograms` on ServiceMonitor and PodMonitor level, but it would be beneficial for the migration to native histograms if this could be configured in a central (global) place, so that not every developer team that has monitoring configured for their services would have to take care of it separately.

## Environment
Kubernetes Version: v1.32.5
Prometheus-Operator Version: v0.83.0

heliapb:
Hi @skoenig as always will try to keep support for new changes when they get released from , will try to add support for this as well.

heliapb:
Hi @skoenig the global config was merged, that feature is being released with the new prometheus version. Will closed the issue since the work was done. Thanks

PR #7667, opened 2025-07-02
feat: global always_scrape_classic_histograms

heliapb:
## Description
From issue https://github.com/prometheus-operator/prometheus-operator/issues/7629, add `always_scrape_classic_histograms` that is going to be release in [prometheus 3.5](https://github.com/prometheus/prometheus/releases/tag/v3.5.0-rc.0)
## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/promcfg.go
+++ b/pkg/prometheus/promcfg.go
@@ -4907,6 +4907,16 @@ func (cg *ConfigGenerator) appendConvertClassicHistogramsToNHCB(cfg yaml.MapSlic
 	return cg.WithMinimumVersion("3.4.0").AppendMapItem(cfg, "convert_classic_histograms_to_nhcb", *cpf.ConvertClassicHistogramsToNHCB)
 }
 
+func (cg *ConfigGenerator) appendConvertScrapeClassicHistograms(cfg yaml.MapSlice) yaml.MapSlice {
+	cpf := cg.prom.GetCommonPrometheusFields()
+
+	if cpf.ScrapeClassicHistograms == nil {
+		return cfg
+	}
+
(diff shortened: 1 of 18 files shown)


pkg/prometheus/promcfg.go
```diff
@@ -479,6 +479,9 @@ func (cg *ConfigGenerator) addNativeHistogramConfig(cfg yaml.MapSlice, nhc monit
 		cfg = cg.WithMinimumVersion("3.0.0").AppendMapItem(cfg, "convert_classic_histograms_to_nhcb", nhc.ConvertClassicHistogramsToNHCB)
 	}
 
+	if nhc.AlwaysScrapeClassicHistograms != nil {
+		cfg = cg.WithMinimumVersion("3.5.0").AppendMapItem(cfg, "always_scrape_classic_histograms", nhc.AlwaysScrapeClassicHistograms)
```
simonpasquier: hmm `always_scrape_classic_histograms`was available since 3.x for scrape configs?
The change with 3.5 is that it's also a global config option now.
https://github.com/prometheus/prometheus/blob/main/CHANGELOG.md#300--2024-11-14

simonpasquier: Ah I see that we have `scrapeClassicHistograms` which is the equivalent of `always_scrape_classic_histogramswas` because the field was renamed in Prometheus 3.x.

heliapb: yes good point - I guess this can be closed then as adding the scrapeClassicHistograms already provides the option for scrapeClassicHistograms

heliapb: I guess I just need to move to buildGlobalConfig() instead, and probably just keep the naming for scrapeClassicHistograms - will work on that

simonpasquier (approved):

Issue #7666, opened 2025-07-02
Split API reference into Multiple Pages


AshwinSriram11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Description
Currently, the API reference doc is rendered as a single page, which includes types from all API packages (v1, v1alpha1, and v1beta1). To improve readability and usability, we want to split the API reference documentation by package, with one page per package.

### Prometheus Operator Version

```shell
NA
```

### Kubernetes Version

```yaml
NA
```

### Kubernetes Cluster Type

kind

### How did you deploy Prometheus-Operator?

prometheus-operator/kube-prometheus

### Manifests

```yaml

```

### prometheus-operator log output

```shell
NA
```

AshwinSriram11:
I will start working on this.

PR #7668, opened 2025-07-02
Split API reference into multiple pages

AshwinSriram11:
## Description

Related Issue: #7666
This PR aims to split the API reference page into multiple pages(one page per package) for better readability and usability. 

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/scripts/docs/config.json
+++ b/scripts/docs/config.json
@@ -46,7 +46,7 @@
     "k8s.io/apimachinery/pkg/apis/": "Kubernetes ",
     "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1": "Monitoring v1",
     "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1": "Monitoring v1alpha1",
-    "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1": "Monitoring v1beta1"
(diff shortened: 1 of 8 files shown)


scripts/docs/config.json
```diff
@@ -46,7 +46,7 @@
     "k8s.io/apimachinery/pkg/apis/": "Kubernetes ",
     "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1": "Monitoring v1",
     "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1": "Monitoring v1alpha1",
-    "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1": "Monitoring v1beta1"
+    "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1beta1": "Monitoring v1beta1"
```
AshwinSriram11, line 49: I am not sure what this does, but thought to add this with this PR

AshwinSriram11:
One thing I am seeing is that the header(toc) I specified for each page is not being read. All the pages are reading the v1beta header. Maybe having a separate template folder for each page might be better, but this also means we need to duplicate the other templates as well.

AshwinSriram11:
Another problem would be handling the linking between multiple packages which might be not possible right now on splitting the docs with package.

Issue #7716, opened 2025-07-11
getSelectedConfigResources to get selected config resources by prometheus


yp969803:
### What is missing? Please describe.

getSelectedConfigResources to get selected config resources by prometheus

## Environment
Kubernetes Version:
Prometheus-Operator Version:

PR #5335, opened 2023-02-11
feat: Add ScrapeConfig CRD

xiu:
## Description

This PR adds the `ScrapeConfig` CRD.

Proposal: [ScrapeConfig](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/proposals/202212-scrape-config.md)
Issue: #2787 

## Type of change

_What type of changes does your code introduce to the Prometheus operator? Put an `x` in the box that apply._

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [x] `FEATURE` (non-breaking change which adds functionality)
- [ ] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [ ] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

```release-note
Add ScrapeConfig CRD, allowing users to scrape targets outside the kubernetes cluster
```

PR #7717, opened 2025-07-11
feat: added getConfigResources function

yp969803:
## Description

added getConfigResources function to get the config resources selected by the prometheus

Closes: #7716 

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -1154,36 +1154,35 @@ func ListOptions(name string) metav1.ListOptions {
 }
 
 // getSeletedConfigResources returns all the configuration resources (PodMonitor, ServiceMonitor, Probes and ScrapeConfigs) selected by the Prometheus.
-func (c *Operator) getSelectedConfigResources(ctx context.Context, logger *slog.Logger, p *monitoringv1.Prometheus, store *assets.StoreBuilder) (selectedConfigResources, error) {
-	var resources selectedConfigResources
-
+func (c *Operator) getSelectedConfigResources(ctx context.Context, logger *slog.Logger, p *monitoringv1.Prometheus, store *assets.StoreBuilder) (*selectedConfigResources, error) {
 	resourceSelector, err := prompkg.NewResourceSelector(logger, p, store, c.nsMonInf, c.metrics, c.eventRecorder)
+
 	if err != nil {
-		return resources, err
+		return nil, err
 	}
 	smons, err := resourceSelector.SelectServiceMonitors(ctx, c.smonInfs.ListAllByNamespace)
 	if err != nil {
-		return resources, fmt.Errorf("selecting ServiceMonitors failed: %w", err)
+		return nil, fmt.Errorf("selecting ServiceMonitors failed: %w", err)
 	}
 
 	pmons, err := resourceSelector.SelectPodMonitors(ctx, c.pmonInfs.ListAllByNamespace)
 	if err != nil {
-		return resources, fmt.Errorf("selecting PodMonitors failed: %w", err)
+		return nil, fmt.Errorf("selecting PodMonitors failed: %w", err)
 	}
 
 	bmons, err := resourceSelector.SelectProbes(ctx, c.probeInfs.ListAllByNamespace)
 	if err != nil {
-		return resources, fmt.Errorf("selecting Probes failed: %w", err)
+		return nil, fmt.Errorf("selecting Probes failed: %w", err)
 	}
 
 	var scrapeConfigs prompkg.ResourcesSelection[*monitoringv1alpha1.ScrapeConfig]
 	if c.sconInfs != nil {
 		scrapeConfigs, err = resourceSelector.SelectScrapeConfigs(ctx, c.sconInfs.ListAllByNamespace)
 		if err != nil {
-			return resources, fmt.Errorf("selecting ScrapeConfigs failed: %w", err)
+			return nil, fmt.Errorf("selecting ScrapeConfigs failed: %w", err)
 		}
 	}
-	return selectedConfigResources{
+	return &selectedConfigResources{
 		sMons:         smons,
 		bMons:         bmons,
 		pMons:         pmons,
@@ -1191,7 +1190,7 @@ func (c *Operator) getSelectedConfigResources(ctx context.Context, logger *slog.
 	}, nil
 }
 
-func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, logger *slog.Logger, p *monitoringv1.Prometheus, cg *prompkg.ConfigGenerator, ruleConfigMapNames []string, store *assets.StoreBuilder, resources selectedConfigResources) error {
+func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, logger *slog.Logger, p *monitoringv1.Prometheus, cg *prompkg.ConfigGenerator, ruleConfigMapNames []string, store *assets.StoreBuilder, resources *selectedConfigResources) error {
 	// If no service/pod monitor and probe selectors are configured, the user
 	// wants to manage configuration themselves. Let's create an empty Secret
 	// if it doesn't exist.
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -805,6 +805,9 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	}
 
 	resources, err := c.getSelectedConfigResources(ctx, logger, p, assetStore)
+	if err != nil {
+		return err
+	}
 
 	if err := c.createOrUpdateConfigurationSecret(ctx, logger, p, cg, ruleConfigMapNames, assetStore, resources); err != nil {
 		return fmt.Errorf("creating config failed: %w", err)
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -1153,7 +1153,7 @@ func ListOptions(name string) metav1.ListOptions {
 	}
 }
 
-// getSeletedConfigResources returns the selected configuration resources (PodMonitor, ServiceMonitor, Probes and ScrapeConfigs) by the Prometheus.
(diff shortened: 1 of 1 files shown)


yp969803:
@slashpai @mviswanathsai @simonpasquier review and merge request

pkg/prometheus/server/operator.go
```diff
@@ -1139,56 +1153,66 @@ func ListOptions(name string) metav1.ListOptions {
...
-			if _, err := c.kclient.CoreV1().Secrets(p.Namespace).Create(ctx, s, metav1.CreateOptions{}); err != nil && !apierrors.IsAlreadyExists(err) {
-				return fmt.Errorf("failed to create an empty configuration secret: %w", err)
-			}
-
-			return nil
-		}
-
-		return err
-	}
+// getSeletedConfigResources returns the selected configuration resources (PodMonitor, ServiceMonitor, Probes and ScrapeConfigs) by the Prometheus.
```
mviswanathsai: (nit)
```suggestion
// getSeletedConfigResources returns all the configuration resources (PodMonitor, ServiceMonitor, Probes and ScrapeConfigs) selected by the Prometheus.
```

mviswanathsai (commented):
Just a small nit. LGTM otherwise

pkg/prometheus/server/operator.go
yp969803: done

yp969803:
the lint issue is not with my change, you can merge the branch if you want

pkg/prometheus/server/operator.go
```diff
@@ -1139,56 +1153,66 @@ func ListOptions(name string) metav1.ListOptions {
...
-		return err
-	}
+// getSeletedConfigResources returns all the configuration resources (PodMonitor, ServiceMonitor, Probes and ScrapeConfigs) selected by the Prometheus.
+func (c *Operator) getSelectedConfigResources(ctx context.Context, logger *slog.Logger, p *monitoringv1.Prometheus, store *assets.StoreBuilder) (selectedConfigResources, error) {
+	var resources selectedConfigResources
 
 	resourceSelector, err := prompkg.NewResourceSelector(logger, p, store, c.nsMonInf, c.metrics, c.eventRecorder)
 	if err != nil {
-		return err
+		return resources, err
```
slashpai: what would be value of resources here when its errored? Should return as `nil,err`?
IMO we should return it as nil to avoid confusion for caller to avoid using invalid data

Verify by running in actual cluster

pkg/prometheus/server/operator.go
```diff
@@ -1139,56 +1153,66 @@ func ListOptions(name string) metav1.ListOptions {
...
 	}
-
 	smons, err := resourceSelector.SelectServiceMonitors(ctx, c.smonInfs.ListAllByNamespace)
 	if err != nil {
-		return fmt.Errorf("selecting ServiceMonitors failed: %w", err)
+		return resources, fmt.Errorf("selecting ServiceMonitors failed: %w", err)
```
slashpai: same question as above

pkg/prometheus/server/operator.go
```diff
@@ -1139,56 +1153,66 @@ func ListOptions(name string) metav1.ListOptions {
...
 	}
 
 	pmons, err := resourceSelector.SelectPodMonitors(ctx, c.pmonInfs.ListAllByNamespace)
 	if err != nil {
-		return fmt.Errorf("selecting PodMonitors failed: %w", err)
+		return resources, fmt.Errorf("selecting PodMonitors failed: %w", err)
```
slashpai: same question as above

pkg/prometheus/server/operator.go
```diff
@@ -1139,56 +1153,66 @@ func ListOptions(name string) metav1.ListOptions {
...
 	}
 
 	bmons, err := resourceSelector.SelectProbes(ctx, c.probeInfs.ListAllByNamespace)
 	if err != nil {
-		return fmt.Errorf("selecting Probes failed: %w", err)
+		return resources, fmt.Errorf("selecting Probes failed: %w", err)
```
slashpai: same question as above

pkg/prometheus/server/operator.go
```diff
@@ -1139,56 +1153,66 @@ func ListOptions(name string) metav1.ListOptions {
...
 	}
 
 	var scrapeConfigs prompkg.ResourcesSelection[*monitoringv1alpha1.ScrapeConfig]
 	if c.sconInfs != nil {
 		scrapeConfigs, err = resourceSelector.SelectScrapeConfigs(ctx, c.sconInfs.ListAllByNamespace)
 		if err != nil {
-			return fmt.Errorf("selecting ScrapeConfigs failed: %w", err)
+			return resources, fmt.Errorf("selecting ScrapeConfigs failed: %w", err)
```
slashpai: same question as above

pkg/prometheus/server/operator.go
yp969803: var resources selectedConfigResources.
resources contains empty slices, initially, 
if there is error we return empty slices

yp969803: also i am not returning nil, because then later in the code, i have to check null condition, as it is not needed because if there is a error, code do not proceed further and reconcillation happens again

yp969803: also i am not returning nil, because then later in the code, i have to check null condition, as it is not needed because if there is a error, code do not proceed further and reconcillation happens again

slashpai: We return as nil for the object if there is error in operator code. We should follow same pattern for consistency. 

For eg:

https://github.com/prometheus-operator/prometheus-operator/blob/5e7e76e437b94fa49edea72f34d937b37048b8f8/pkg/prometheus/promcfg.go#L189-L208

https://github.com/prometheus-operator/prometheus-operator/blob/5e7e76e437b94fa49edea72f34d937b37048b8f8/pkg/prometheus/promcfg.go#L153-L156

yp969803: dome

pkg/prometheus/server/operator.go
yp969803: done

pkg/prometheus/server/operator.go
yp969803: done

pkg/prometheus/server/operator.go
yp969803: done

pkg/prometheus/server/operator.go
yp969803: done

yp969803:
@slashpai done the changes, also i am not checking nil condition for resources, as if it is nil means there is error, and sync func do not run futher and return errr

pkg/prometheus/server/operator.go
```diff
@@ -1139,7 +1153,44 @@ func ListOptions(name string) metav1.ListOptions {
...
+		return nil, fmt.Errorf("selecting PodMonitors failed: %w", err)
+	}
+
+	bmons, err := resourceSelector.SelectProbes(ctx, c.probeInfs.ListAllByNamespace)
+	if err != nil {
+		return nil, fmt.Errorf("selecting Probes failed: %w", err)
+	}
+
+	var scrapeConfigs prompkg.ResourcesSelection[*monitoringv1alpha1.ScrapeConfig]
+	if c.sconInfs != nil {
```
slashpai, line 1179: Is there a reason only for scrapeConfigs we are checking this first?

yp969803: i am not aware of this, i have copied it from previous code, do not why we are checking scrapeconfig informers for this case only

mviswanathsai: In `cmd/operator/main.go` [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/cmd/operator/main.go#L327-L340), out of all the config resources, ScrapeConfig is the only resource for which we perform a check to see if the CRD is present and accessible in the cluster. Based on this, the operator may or may not reconcile the ScrapeConfig (and thus may/may not have an informer). 

The reason for the check has been given by Simon here: https://github.com/prometheus-operator/prometheus-operator/pull/5335#pullrequestreview-1328886441

These checks make sense for the other resources as well, but a lack of permissions for the older resources is probably a very small edge case since they have been around for a long time.

mviswanathsai: TLDR; it is fine to check for a nil pointer for the informer. We may want to remove it (or not) in a future PR. But that is beyond the scope of the changes in this PR.

slashpai: Yes you are right.We don't install scrapeconfig crds by default as it's still in v1alpha1.

pkg/prometheus/server/operator.go
```diff
@@ -1238,10 +1261,10 @@ func (c *Operator) createOrUpdateConfigurationSecret(ctx context.Context, logger
...
 		p,
-		smons.ValidResources(),
-		pmons.ValidResources(),
-		bmons.ValidResources(),
-		scrapeConfigs.ValidResources(),
+		resources.sMons.ValidResources(),
+		resources.pMons.ValidResources(),
+		resources.bMons.ValidResources(),
+		resources.scrapeConfigs.ValidResources(),
 		store,
```
mviswanathsai, line 1268: I know it is not possible for a nil value of resources to make it to this point in the execution flow. But would it make sense to add a nil pointer check anywhere just in case?

yp969803: But remember what we discussed with Simon earlier — in this case, if resources is nil, it indicates something malicious in the program and it should crash. If we simply check for nil and allow the code to continue, it means we're knowingly running a malicious program

simonpasquier:
I miss the context behind this change. Is it a refactoring required for a following PR?

yp969803:
> I miss the context behind this change. Is it a refactoring required for a following PR?

i need resources in the sync function so that i can pass it to another function which updates the status

simonpasquier (approved):

PR #7725, opened 2025-07-16
feat: Add global config option for always_scrape_classic_histograms 

dongjiang1989:
## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request._

_If it fixes an existing issue (bug or feature), use the following keyword:_

_Closes: #ISSUE-NUMBER_

ref: https://github.com/prometheus/prometheus/pull/16452

## Type of change

## Verification

## Changelog entry

```release-note
Add global config option for always_scrape_classic_histograms 
```


slashpai (dismissed):

slashpai:
@heliapb IIUC this field is global one `alwaysScrapeClassicHistograms`?

slashpai:
ok I guess a confusion was caused here as json tag is not matching as in prometheus https://github.com/prometheus-operator/prometheus-operator/pull/7667/files#diff-b4a3c944005e67c959c04552108ff3b629b572d7244d586b0558651bd96eacdaR723

heliapb:
> ok I guess a confusion was caused here as json tag is not matching as in prometheus https://github.com/prometheus-operator/prometheus-operator/pull/7667/files#diff-b4a3c944005e67c959c04552108ff3b629b572d7244d586b0558651bd96eacdaR723

I think it was renamed at some point, at was kept here as is since it was implemented first as https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/types.go#L992 and https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L474

slashpai:
So for global config since it was introduced in prom3 cant we keep name as in Prometheus ie `AlwaysScrapeClassicHistograms`

heliapb:
> So for global config since it was introduced in prom3 cant we keep name as in Prometheus ie `AlwaysScrapeClassicHistograms`

yeah could be renamed  - but since global option as `scrapeClassicHistograms` was already part of the [v0.84.0](https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.84.0) that would mean a breaking change for a next release.

slashpai:
Yes but IIUC it was not useful in last release as we didn't ship with Prometheus 3.5. What we can do is patch release for this probably to avoid waiting till next release

heliapb:
> Yes but IIUC it was not useful in last release as we didn't ship with Prometheus 3.5. What we can do is patch release for this probably to avoid waiting till next release

Makes sense then, I guess @dongjiang1989 could you use this pr to remane `scrapeClassicHistograms` to `alwaysScrapeClassicHistograms` for the change done previously in the global config, as in https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/prometheus_types.go#L723

dongjiang1989:
Thanks @slashpai  @heliapb 
I'll remane `scrapeClassicHistograms` to `alwaysScrapeClassicHistograms` just in the global config in this PR

dongjiang1989:
Thanks @slashpai @heliapb 
Fixed

pkg/apis/monitoring/v1/types.go
```diff
@@ -985,8 +985,9 @@ const (
 // NativeHistogramConfig extends the native histogram configuration settings.
 // +k8s:openapi-gen=true
 type NativeHistogramConfig struct {
-	// Whether to scrape a classic histogram that is also exposed as a native histogram.
-	// It requires Prometheus >= v2.45.0.
+	// Whether to scrape a classic histogram, even if it is also exposed as a native histogram.
+	//
+	// It requires Prometheus >= v3.5.0.
```
heliapb: Hi this should not be changed as the `scrape_classic_histograms` is supported since 2.45.0, only the global.always_scrape_classic_histograms requires the version 3.5 as in https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go#L469

dongjiang1989: Thanks @heliapb 
I miss it
Fixed

slashpai:
@dongjiang1989  can you change the PR base to release-0.84?

cc @mviswanathsai (our release shepherd for v0.84 release)

dongjiang1989:
> @dongjiang1989 can you change the PR base to release-0.84?
> 
> cc @mviswanathsai (our release shepherd for v0.84 release)

@slashpai  PTAL

Done

heliapb (approved):

simonpasquier:
I'm on the fence about this one: technically it's still a breaking change even if it's only been shipped in 0.84.0. The fact that we didn't officially support 3.5.0 isn't really relevant because the operator supports versions which aren't officially listed in pkg/operator/defaults.go.

heliapb:
> I'm on the fence about this one: technically it's still a breaking change even if it's only been shipped in 0.84.0. The fact that we didn't officially support 3.5.0 isn't really relevant because the operator supports versions which aren't officially listed in pkg/operator/defaults.go.

Well personally I feel having the global option different than the config everywhere else is a concern. https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/types.go#L992 - as I feel this can cause confusion to the users. If we want to keep closer to prometheus maybe would be better to change everywhere else even if it becomes a breaking change .

simonpasquier:
> If we want to keep closer to prometheus maybe would be better to change everywhere else even if it becomes a breaking change.

Unfortunately we guarantee [non-breaking changes](https://github.com/prometheus-operator/prometheus-operator/blob/main/CONTRIBUTING.md#changes-to-the-apis) for v1 types and https://github.com/prometheus-operator/prometheus-operator/commit/a26ca7273e778f4775770b56b63c317fe5dcd33f was shipped as part of v0.78.0 so it's too late to change the field name in ServiceMonitor, PodMonitor et al. We may want to introduce a new field matching the new Prometheus format and deprecate the existing one but it's additional burden. IIRC the only exception to the rule was when we added a new field to the CRD and we had to ship a patch release because the feature didn't work as expected (for instance #7197 + #7199).

Our options for this patch release are:
1. Use `alwaysScrapeClassicHistograms` for the global config.
2. Don't change anything but highlight in the API documentation that the fields correspond to `always_scrape_classic_histograms` in Prometheus land.

slashpai:
I think highlighting in API doc would be enough if we cannot change field name

dongjiang1989:
> 2\. `always_scrape_classic_histograms`不要更改任何内容，但在 API 文档中突出显示字段与 Prometheus land 中对应的内容。

+1

heliapb:
> I think highlighting in API doc would be enough if we cannot change field name

Well maybe it can be documented somewhere besides the api doc, explaining why the fields have different names from the global cfg vs everywhere else

simonpasquier:
@heliapb @slashpai @mviswanathsai do we agree that we keep the current API as it is? If yes, we should be able to cut a new patch release for 0.84.

slashpai:
Ya I am ok with current change. API doc update can be done later probably to avoid reworking?

heliapb:
> @heliapb @slashpai @mviswanathsai do we agree that we keep the current API as it is? If yes, we should be able to cut a new patch release for 0.84.

sounds good.

slashpai:
We can close this and create separate PR to main to update just api docs?
I have created patch release PR for v0.84.1

mviswanathsai:
> @heliapb @slashpai @mviswanathsai do we agree that we keep the current API as it is? If yes, we should be able to cut a new patch release for 0.84.

I am very late to this one, unfortunately. But for the record, I agree with keeping the current API as it is and updating the docs to make the mapping clear to the users.

slashpai:
@dongjiang1989 can you open PR to main with API doc update if updating this PR is difficult?
Feel free to close this one.

dongjiang1989:
> @dongjiang1989 can you open PR to main with API doc update if updating this PR is difficult? Feel free to close this one.

Thanks @slashpai 
I will close this pr and open new PR to update API doc.

dongjiang1989:
The Next PR： https://github.com/prometheus-operator/prometheus-operator/pull/7784

So. Close this PR

Issue #7745, opened 2025-07-25
Support client_secret_file for oauth2 to load credentials from file


Fresa:
### Component(s)

Prometheus

### What is missing? Please describe.

Prometheus supports loading oauth2 secrets from file, would be great to support that via the operator. I'm using a CSI driver to mount sensitive information like oauth credentials directly to the Prometheus pods through volumes, instead of having to detour via a secret, which has several security downsides.

OAuth2 reference in the operator: https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.OAuth2
OAuth2 reference for Prometheus (note `client_secret_file`): https://prometheus.io/docs/prometheus/latest/configuration/configuration/#oauth2

### Describe alternatives you've considered.

Setting this configuration via CLI or environment variables, but it doesn't seem like Prometheus support that?

## Environment
Kubernetes Version: v1.31.2
Prometheus-Operator Version: v0.83.0

heliapb:
Hi @Fresa currently it supports both secret and configmap if you go with  `clientId`  and `clientSecret`, but I see no issue with adding `client_secret_file` support as well. Unless it was not done for some specific reasoning before, /cc @slashpai @simonpasquier if you could give your feedback.

simonpasquier:
For which parts do you need this capability?
For security reasons, we don't allow PodMonitor, Probe and ScrapeConfig to reference files which are local to the Prometheus pods (e.g. somebody able to create a PodMonitor could exfiltrate sensitive content like SA token from the Prometheus' filesystem).

Fresa:
When configuring remote write, I'd like to read an oauth2 client secret from a volume mounted by a [(secret store) CSI driver](https://github.com/kubernetes-sigs/secrets-store-csi-driver)

simonpasquier:
This would be possible then: any oauth2 config embedded in the Prometheus CRD is safe to have a reference to a local file.

PR #7760, opened 2025-07-31
feat: add clientSecretFile

heliapb:
## Description

https://github.com/prometheus-operator/prometheus-operator/issues/7745

## Verification

## Changelog entry

```release-note

```


pkg/apis/monitoring/v1/types.go
```diff
@@ -700,6 +700,10 @@ type OAuth2 struct {
 	// client's secret.
 	ClientSecret v1.SecretKeySelector `json:"clientSecret"`
 
+	// `clientSecret` specifies a key of a Secret containing the OAuth2
+	// client's secret.
+	ClientSecretFile *string `json:"clientSecretFile"`
```
simonpasquier: Warning: we don't want this to be added to all resources. It's only safe within workload resources (Promehteus, Alertmanager and ThanosRuler).

heliapb: created  a v2 for the oauth2, don't know if is the best approach but do let me know

simonpasquier: It's a good approach but we should only switch from oauth2 to oauth2v2 when the field is in a workload spec (e.g. not for config resources like ServiceMonitor, PodMonitor, etc.).

PR #7775, opened 2025-08-05
feat: Add reconcile-delay flag to controllers

slashpai:
## Description

_Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request._

_If it fixes an existing issue (bug or feature), use the following keyword:_

_Closes: #ISSUE-NUMBER_

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/jsonnet/prometheus-operator/prometheus-operator.libsonnet
+++ b/jsonnet/prometheus-operator/prometheus-operator.libsonnet
@@ -30,6 +30,7 @@ local defaults = {
   kubeletService: 'kube-system/kubelet',
   kubeletEndpointsEnabled: true,
   kubeletEndpointSliceEnabled: false,
+  reconcileDelay: '',
 };
 
 function(params) {
@@ -186,6 +187,8 @@ function(params) {
       if value != '' then [arg + '=' + value] else [];
     local enableReloaderProbesArg(value) =
       if value == true then ['--enable-config-reloader-probes=true'] else [];
+    local reconcileDelayArg(flag, value) =
+      if value != '' then [flag + '=' + value] else [];
 
     local container = {
       name: po.config.name,
@@ -200,7 +203,8 @@ function(params) {
             reloaderResourceArg('--config-reloader-memory-limit', po.config.configReloaderResources.limits.memory) +
             reloaderResourceArg('--config-reloader-cpu-request', po.config.configReloaderResources.requests.cpu) +
             reloaderResourceArg('--config-reloader-memory-request', po.config.configReloaderResources.requests.memory) +
-            enableReloaderProbesArg(po.config.enableReloaderProbes),
+            enableReloaderProbesArg(po.config.enableReloaderProbes) +
+            reconcileDelayArg('--reconcile-delay', po.config.reconcileDelay),
       ports: [{
         containerPort: po.config.port,
         name: 'http',
--- a/pkg/operator/resource_reconciler.go
+++ b/pkg/operator/resource_reconciler.go
@@ -96,6 +96,9 @@ type ResourceReconciler struct {
 	g errgroup.Group
 
 	controllerID string
+
+	reconcileDelay time.Duration
+	lastReconcile  map[string]time.Time
 }
 
 var (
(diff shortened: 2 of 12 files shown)


slashpai:
This is not right approach I am working on adding the delay through ResourceReconciler

simonpasquier (commented):
Thanks! I wonder if we shouldn't have only one global argument applied to all controllers for now?

jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -30,6 +30,9 @@ local defaults = {
   kubeletService: 'kube-system/kubelet',
   kubeletEndpointsEnabled: true,
   kubeletEndpointSliceEnabled: false,
+  alertmanagerReconcileDelay: '0s',
+  prometheusReconcileDelay: '0s',
+  thanosRulerReconcileDelay: '0s',
```
simonpasquier: maybe default to ''?

slashpai:
I initially thought to have single argument for all controllers but had second thoughts on it. Because if we need delay only for one controller we would have option. But I am ok to make it into one argument if that's what we are looking for.

simonpasquier:
We can always add more options later if needed. If you need the functionality, you're probably going to apply it for all controllers.

slashpai:
@simonpasquier updated, can you review again?

jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -30,6 +30,7 @@ local defaults = {
   kubeletService: 'kube-system/kubelet',
   kubeletEndpointsEnabled: true,
   kubeletEndpointSliceEnabled: false,
+  reconcileDelay: '0s',
```
simonpasquier:
```suggestion
  reconcileDelay: '',
```

jsonnet/prometheus-operator/prometheus-operator.libsonnet
```diff
@@ -186,6 +187,8 @@ function(params) {
       if value != '' then [arg + '=' + value] else [];
     local enableReloaderProbesArg(value) =
       if value == true then ['--enable-config-reloader-probes=true'] else [];
+    local reconcileDelayArg(flag, value) =
+      if value != '0s' && value != '' then [flag + '=' + value] else [];
```
simonpasquier:
````suggestion
      if value != '' then [flag + '=' + value] else [];
````

File: pkg/operator/resource_reconciler.go [hunk 4]
```diff
@@ -666,13 +671,22 @@ func (rr *ResourceReconciler) processNextReconcileItem(ctx context.Context) bool
 	defer rr.reconcileQ.Done(key)
 	defer rr.statusQ.Add(key) // enqueues the object's key to update the status subresource
 
+	// Check if reconciliation should be delayed
```
simonpasquier, line 674: I wonder if we shouldn't move this earlier (e.g. right after `rr.reconcileQ.Get()`)?

pkg/operator/resource_reconciler.go
```diff
@@ -666,13 +671,22 @@ func (rr *ResourceReconciler) processNextReconcileItem(ctx context.Context) bool
...
+
 	rr.reconcileTotal.Inc()
 	startTime := time.Now()
 	err := rr.syncer.Sync(ctx, key)
 	rr.reconcileDuration.Observe(time.Since(startTime).Seconds())
 
 	if err == nil {
 		rr.reconcileQ.Forget(key)
+		// Update the last reconcile time on successful reconciliation
+		rr.updateLastReconcileTime(key)
```
simonpasquier: (nit) should we set the timestamp to `startTime`?

pkg/operator/resource_reconciler.go
slashpai: since Done() and Add() is called through `defer` I believe current approach is correct, correct me if I am wrong :)

pkg/operator/resource_reconciler.go
slashpai: ya you are right good catch, with time.Now() delay varies based on how long reconciliation takes

pkg/operator/resource_reconciler.go
```diff
@@ -96,6 +96,9 @@ type ResourceReconciler struct {
 	g errgroup.Group
 
 	controllerID string
+
+	reconcileDelay time.Duration
+	lastReconcile  map[string]time.Time
```
slashpai, line 101: do we need to use sync.map instead of map here?

pkg/operator/resource_reconciler.go
simonpasquier: I expect that we don't enqueue to statusQ if the object isn't effectively reconciled.

slashpai:
This is not needed

Issue #7794, opened 2025-08-08
Remove smon status bindings if the refrenced workload is deleted


yp969803:
### What is missing? Please describe.

Remove smon status bindings if the refrenced workload is deleted

## Environment
Kubernetes Version:
Prometheus-Operator Version:

PR #7795, opened 2025-08-08
Remove binding from the smon status, if the ref workload is deleted

yp969803:
## Description
Remove binding from the smon status, if the ref workload is deleted

Closes: #7794 

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/server/operator.go
+++ b/pkg/prometheus/server/operator.go
@@ -1091,7 +1091,7 @@ func (c *Operator) configResStatusCleanup(ctx context.Context, p *monitoringv1.P
 			getErr = prompkg.RemoveServiceMonitorBinding(ctx, configResourceSyncer, s)
 		}
 	}); err != nil {
-		return err
+		return fmt.Errorf("listing all ServiceMonitors from cache failed: %w", err)
 	}
 	return getErr
 }


yp969803:
@simonpasquier review request

simonpasquier:
For this one, I'd expect that we leverage the FinalizerSyncer. It should be its responsibility to remove the bindings (maybe with help from the controller?) and then remove the finalizer only after everything's cleaned up.

yp969803:
RemoveBinding requires many helper functions and structs from controller, also finalizerSyncer responsiblity is to add/remove finalizer, adding/removing status is done by controller sync function

yp969803:
@simonpasquier PTAL!

pkg/prometheus/server/operator.go
```diff
@@ -804,12 +804,22 @@ func (c *Operator) sync(ctx context.Context, key string) error {
 	logger := c.logger.With("key", key)
 	c.logDeprecatedFields(logger, p)
 
-	finalizersChanged, err := c.finalizerSyncer.Sync(ctx, p, logger, c.rr.DeletionInProgress(p))
+	assetStore := assets.NewStoreBuilder(c.kclient.CoreV1(), c.kclient.CoreV1())
+	resources, err := c.getSelectedConfigResources(ctx, logger, p, assetStore)
```
simonpasquier: wouldn't it be better to iterate over all config resources and remove bindings which refer to the deleted prometheus? Otherwise we're at risk of missing some config resources.

pkg/prometheus/server/operator.go
```diff
@@ -1070,6 +1073,28 @@ func (c *Operator) updateConfigResourcesStatus(ctx context.Context, p *monitorin
 	}
 }
 
+// configResStatusCleanup removes prometheus bindings from the selected configuration resources (ServiceMonitor, PodMonitor, ScrapeConfig and PodMonitor).
+// It returns true if another reconciliation is needed to complete the cleanup and and an error if the cleanup fails.
```
simonpasquier: like discussed directly I don't think that we need to throttle the updates for now.

yp969803: ok

pkg/prometheus/server/operator.go
yp969803: ok

yp969803:
@simonpasquier PTAL done the changes

simonpasquier (approved):
LGTM

PR #7798, opened 2025-08-11
Add follow-up silence CRD implementation proposal

danielmellado:
## description

this pr adds a new, follow-up proposal for the silence crd implementation original one proposed by mcbenjemaa.

## type of change

- [ ] `change` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `feature` (non-breaking change which adds functionality)
- [ ] `bugfix` (non-breaking change which fixes an issue)
- [ ] `enhancement` (non-breaking change which improves existing functionality)
- [x] `none` (if none of the other choices apply. example, tooling, build system, ci, docs, etc.)

### DIFF
--- a/Documentation/proposals/202508-silence-crd-implementation.md
+++ b/Documentation/proposals/202508-silence-crd-implementation.md
@@ -1,0 +1,379 @@
+# Silence CRD Implementation
+
+* **Owners:**
+  * [mcbenjemaa](https://github.com/mcbenjemaa)
+  * [@danielmellado](https://github.com/danielmellado)
+  * [@mariofer](https://github.com/mariofer)
+* **Status:**
+  * `Accepted`
+* **Related Tickets:**
+  * [#5452](https://github.com/prometheus-operator/prometheus-operator/issues/5452)
+  * [#2398](https://github.com/prometheus-operator/prometheus-operator/issues/2398)
+* **Other docs:**
+  * [Original Silence CRD Proposal](https://github.com/prometheus-operator/prometheus-operator/pull/5485)
+
+> TL;DR: This proposal outlines the implementation approach for the Silence
+> CRD, enabling GitOps-friendly management of Alertmanager silences through
+> Kubernetes resources by extending the existing Alertmanager controller.
+
+## Why
+
+Currently, managing Alertmanager silences requires manual UI interaction or
+direct API calls to the Alertmanager API. This creates operational friction
+and prevents GitOps integration, making it difficult to manage silences as
+code alongside other monitoring configuration.
+
+### Pitfalls of the current solution
+
+* **Manual Operations**: Users must access Alertmanager UI or use custom
+  scripts for silence management
+* **No GitOps Integration**: Silences cannot be version-controlled or managed
+  through standard Kubernetes workflows
+* **Limited Access Control**: No fine-grained RBAC for silence operations
+* **Multi-cluster Complexity**: Managing silences across multiple clusters
+  requires custom tooling and processes
+* **No Audit Trail**: Difficult to track who created/modified silences and
+  when changes occurred
+
+## Audience
+
+* Platform teams managing monitoring infrastructure
+* DevOps engineers implementing GitOps workflows
+* SRE teams needing programmatic silence management
+* Organizations requiring compliance and audit trails
+
+## Goals
+
+* Enable GitOps workflows for silence management
+* Provide RBAC integration for access control
+* Support multi-Alertmanager deployments
+* Maintain backward compatibility with existing Alertmanager functionality
+
+## Non-Goals
+
+* Automatic cleanup of expired Silence CRs (users manage lifecycle, though the controller
+  will detect when Alertmanager has deleted expired silences after the retention period
+  (default: 120h, configurable via .spec.retention) and update the CRD status accordingly)
+* Cross-cluster silence management (creating silences that span multiple Kubernetes
+  clusters or federated Alertmanager deployments is out of scope)
+* Real-time sync guarantees (eventual consistency is acceptable, minor delays between
+  CRD updates and Alertmanager API changes are expected due to reconciliation loops)
+* Authentication support for Alertmanager API access (the controller will only support
+  unauthenticated Alertmanager instances, see [#5836](https://github.com/prometheus-operator/prometheus-operator/issues/5836)
+  and [#7004](https://github.com/prometheus-operator/prometheus-operator/pull/7004) for ongoing authentication work), other than that this may also be handled by a reverse proxy such as `kube-rbac-proxy`
+
+## How
+
+*Note: This feature will be protected by a feature gate `EnableSilenceCRD`, with graceful degradation when CRD or RBAC prerequisites are not met.*
+
+### Architecture Choice
+
+The implementation uses a dedicated Silence controller within prometheus-operator
+rather than extending the existing Alertmanager controller. This provides:
+
+* Clean separation of concerns between StatefulSet/config management and silences
+* Independent scaling and reconciliation patterns
+* Simplified testing and maintenance
+* Decoupled lifecycle management
+
+### API Design
+
+The Silence CRD will be defined in `monitoring.coreos.com/v1alpha1`:
+
+**Silence Resource Example:**
+
+```yaml
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: Silence
+metadata:
+  name: maintenance-silence
+  namespace: monitoring
+  labels:
+    app: my-app
+    team: platform
+spec:
+  comment: "Database upgrade maintenance window - disabling alerts for API services"
+  expiresAt: "2024-01-15T12:00:00Z"
+  matchers:
+  - name: "alertname"
+    value: "ServiceUnavailable"
+    matchType: "="
+  - name: "service"
+    value: "api-.*"
+    matchType: "=~"
+  - name: "severity"
+    value: "warning"
+    matchType: "!="
+status:
+  observedGeneration: 1
+  bindings: # to re-use the same naming than for ServiceMonitor
+    - name: "main-alertmanager"
+      namespace: "monitoring"
+      silenceID: "550e8400-e29b-41d4-a716-446655440000"
+      lastSyncTime: "2024-01-15T08:00:15Z"
+      syncedInstances: 3
+      totalInstances: 3
+      conditions:
+        - type: Ready
+          status: "True"
+          lastTransitionTime: "2024-01-15T08:00:15Z"
+          reason: SilenceApplied
+          message: "Silence applied successfully to Alertmanager instances"
+```
+
+**Alertmanager Resource with Silence Selection:**
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  name: main-alertmanager
+  namespace: monitoring
+spec:
+  replicas: 3
+  silenceSelector:
+    matchLabels:
+      team: platform
+  silenceNamespaceSelector:
+    matchLabels:
+      name: monitoring
+  # ... other Alertmanager configuration
+```
+
+### Implementation Strategy
+
+**Controller Integration**: A dedicated Silence controller will be implemented
+within prometheus-operator with its own informers, event handlers, and
+reconciliation logic. The controller will:
+
+1. **Watch Silence custom resources**: Detect creation, updates, and deletions
+2. **Watch Alertmanager custom resources**: React to changes in silence selectors
+3. **Reconcile on Events**: Process changes and sync with Alertmanager APIs
+4. **Periodic Reconciliation**: Detect drift and ensure consistency
+
+**Discovery Mechanism**: Alertmanager resources will define `silenceSelector`
+and `silenceNamespaceSelector` fields to select relevant Silence resources,
+following the same pattern as AlertmanagerConfig.
+
+**Reconciliation Logic**: The controller determines CRUD operations based on:
+
+* **Silence CRD Events**: Create/update/delete operations trigger API calls
+* **Generation Changes**: Spec modifications are detected via `observedGeneration`
+* **Orphaned Silences**: Periodic cleanup of silences no longer matching any CRD
+* **Alertmanager Changes**: Selector modifications trigger re-evaluation of Silence bindings
+
+**API Interaction**: The controller interacts with Alertmanager via REST API
+v2 for CRUD operations. Authentication support will follow the patterns
+established for Prometheus/Alertmanager authentication in issues
+[#5836](https://github.com/prometheus-operator/prometheus-operator/issues/5836)
+and [#7004](https://github.com/prometheus-operator/prometheus-operator/pull/7004).
+
+Supported authentication mechanisms:
+
+* **No Authentication**: Direct HTTP access to Alertmanager API
+* **Reverse Proxy Authentication**: Support for proxy layers that handle
+  authentication transparently (users may deploy various reverse proxies
+  depending on their authentication requirements)
+
+The operator will read authentication configuration from the Alertmanager CRD
+specification, reusing existing authentication fields where possible to maintain
+consistency with Prometheus authentication patterns.
+
+### Authentication and Reverse Proxy Support
+
+**Authentication Configuration**: The Silence controller will support authenticated
+access to Alertmanager APIs. The specific authentication configuration mechanism
+will align with the ongoing work in [#5836](https://github.com/prometheus-operator/prometheus-operator/issues/5836)
+and [#7004](https://github.com/prometheus-operator/prometheus-operator/pull/7004)
+to avoid duplication and ensure consistency across the operator.
+
+The controller will support the following authentication scenarios:
+
+* **No Authentication**: Direct access to Alertmanager API
+* **Reverse Proxy Authentication**: Access through authentication proxies
+* **Native Authentication**: Once #5836 is merged, native authentication support
+
+Authentication configuration will be handled through a dedicated field structure
+(not at the main spec level) to be defined as part of the #5836 implementation.
+
+**Reverse Proxy Support**: The controller supports Alertmanager instances exposed
+through reverse proxies that handle authentication. In this setup:
+
+1. The reverse proxy handles authentication (e.g., validating service account tokens)
+2. The controller accesses the proxy endpoint using its service account credentials
+3. The proxy forwards authenticated requests to the underlying Alertmanager instance
+
+Example reverse proxy configuration:
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  name: rbac-protected-alertmanager
+  namespace: monitoring
+spec:
+  replicas: 3
+  # Controller will access the internal service directly
+  # Authentication configuration will be added once #5836 is implemented
+```
+
+**Security Considerations**: All authentication credentials are stored in Kubernetes
+secrets and mounted securely. The controller follows the principle of least privilege,
+only requiring permissions to read authentication secrets and access Alertmanager APIs.
+
+**Status Management**: The Silence status tracks the observed generation,
+a single Ready condition, and per-Alertmanager status including silence IDs
+and sync state to handle cases where multiple Alertmanager resources select
+the same Silence.
+
+**Cluster Synchronization**: When Alertmanager runs in cluster mode, the controller
+creates silences on one healthy Alertmanager instance and relies on the gossip protocol
+for automatic replication to other cluster members. The controller monitors sync status
+by querying all instances and updates the binding status with `syncedInstances` vs
+`totalInstances`. If replication is broken/degraded, the controller will detect
+inconsistencies and attempt to create the silence directly on affected instances
+to ensure eventual consistency.
+
+**Configuration Drift Detection**: The controller detects drift by periodically
+comparing the Silence CRD spec with the actual silence configuration in Alertmanager
+(using the `createdBy` correlation). When drift is detected (e.g., user modifies
+silence matchers, comment, or expiration via Alertmanager UI), the controller
+reconciles by updating the Alertmanager silence to match the CRD spec, ensuring
+the CRD remains the source of truth.
+
+**Lifecycle Management**: Use finalizers (`monitoring.coreos.com/silence-cleanup`)
+to ensure proper cleanup from Alertmanager API when Silence resources are deleted.
+
+**Coexistence and Correlation**: Only manage silences created by the operator and leave manually-created
+silences untouched. Correlation between Alertmanager API silences and Silence custom
+resources is achieved using the `createdBy` field in the Alertmanager API, which contains `<namespace>/<name>`
+(e.g., `monitoring/maintenance-silence`) to uniquely identify the source custom resource.
+
+**CRUD Operations**:
+
+* **Create**: When a new Silence CRD is created, the controller calls the Alertmanager API to create the silence and stores the returned silence ID in the status
+* **Update**: When a Silence CRD is modified, the controller uses the stored silence ID to update the existing silence via the Alertmanager API
+* **Delete**: When a Silence CRD is deleted, finalizers ensure the controller calls the Alertmanager API to delete the silence before removing the CRD
+* **Drift Detection**: The controller periodically compares Silence CRDs with actual silences in Alertmanager (matching by `createdBy` field) and reconciles any differences
+
+### Namespace-based Tenancy and Matcher Strategy
+
+**AlertmanagerConfig Matcher Strategy Integration**: The Silence CRD will leverage
+the existing `alertmanagerConfigMatcherStrategy` field from the Alertmanager CRD
+to ensure consistent behavior with AlertmanagerConfig resources. This field controls
+how namespace-based matchers are handled:
+
+* **OnNamespace** (default): Automatically inject `namespace=<resource-namespace>`
+  matchers into silences, ensuring tenant isolation
+* **None**: No automatic namespace injection, allowing cross-namespace silence management
+
+**Namespace-based Tenancy**: When `alertmanagerConfigMatcherStrategy` is set to
+`OnNamespace`, the controller automatically injects a `namespace=<silence-namespace>`
+matcher into all silences created from Silence custom resources, ensuring tenant
+isolation in multi-tenant environments.
+
+Example configuration:
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  name: main-alertmanager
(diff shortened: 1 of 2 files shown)


slashpai:
Lets add @mcbenjemaa as co-author signature should be in https://github.com/prometheus-operator/prometheus-operator/pull/5485

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
+# Silence CRD Implementation
+
+* **Owners:**
+  * [mcbenjemaa](https://github.com/mcbenjemaa)
+  * [@danielmellado](https://github.com/danielmellado)
+  * [@mariofer](https://github.com/mariofer)
+* **Status:**
+  * `Proposed`
```
simonpasquier: (nit) it can be change to `Accepted`.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+  name: maintenance-silence
+  namespace: monitoring
+spec:
+  comment: "Scheduled maintenance window"
+  expiresAt: "2024-01-15T10:00:00Z"
+  matchers:
+  - name: "alertname"
+    value: "HighCPUUsage"
+    matchType: "="
+  alertmanagerRef:
```
simonpasquier: This is different from the approach taken by other configuration resources (AlertmanagerConfig for instance) where the workload resource (Alertmanager) defines how to select the configuration resources using label and namespace selectors. What is the motivation here?

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+* Unit tests for controller logic and API conversion
+* Integration tests with real Alertmanager instances
+* End-to-end scenarios covering multi-instance deployments
+* Migration testing with existing silence configurations
+
+### Migration Strategy
+
+Existing silences can be migrated using discovery and conversion utilities
+that extract current silences via Alertmanager API and transform them to
+Silence CRDs with validation.
```
simonpasquier: Is it covered by the proposal or more a general remark? IMHO we don't need to deal with migration. I would even say that the solution should be compatible with silences being provisioned by external means (e.g. CLI, API, UI).

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Alternatives
+
+### Standalone Controller
```
simonpasquier: Is it about a separate operator or a new controller within the existing Prometheus operator? If the latter, I wouldn't be so definitive against it.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+**Why not**: Creating a separate silence controller would duplicate
+infrastructure (informers, RBAC, configuration) and diverge from the
+established pattern of extending existing controllers for related resources.
+
+### Configuration File Approach
+
+**Why not**: Managing silences through Alertmanager configuration files would
+require configuration reloads, increase complexity, and prevent real-time
+updates. The API-based approach provides immediate feedback and better error
+handling.
```
simonpasquier: there's no native way to configure silences from files.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+
+### External Operator
+
+**Why not**: Using an external silence operator (like silence-operator) would
```
simonpasquier: (nit) can we add a link to the silence-operator repository?

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Status Management**: The Silence status tracks synchronization state across
+multiple Alertmanager instances with Kubernetes conditions.
+
+### Testing and Verification
+
+* Unit tests for controller logic and API conversion
+* Integration tests with real Alertmanager instances
+* End-to-end scenarios covering multi-instance deployments
+* Migration testing with existing silence configurations
```
simonpasquier: should we remove this last item?

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+**Discovery Mechanism**: Alertmanager instances discover relevant silences
+through label selectors (`spec.silenceSelector`), namespace selectors
+(`spec.silenceNamespaceSelector`), or explicit references
+(`spec.alertmanagerRef`).
+
+**API Interaction**: The controller interacts with Alertmanager via REST API
+v2 for CRUD operations, using existing authentication patterns with
+Kubernetes secrets.
+
+**Status Management**: The Silence status tracks synchronization state across
```
simonpasquier: I'd expect more details on how the status subresource will look like.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Implementation Strategy
+
+**Controller Integration**: The Alertmanager controller will be extended with
+silence informers, event handlers, and selection logic.
+
+**Discovery Mechanism**: Alertmanager instances discover relevant silences
+through label selectors (`spec.silenceSelector`), namespace selectors
+(`spec.silenceNamespaceSelector`), or explicit references
+(`spec.alertmanagerRef`).
```
simonpasquier: see my comment above.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+
+```yaml
+apiVersion: monitoring.coreos.com/v1alpha1
+kind: Silence
+metadata:
+  name: maintenance-silence
+  namespace: monitoring
+spec:
+  comment: "Scheduled maintenance window"
+  expiresAt: "2024-01-15T10:00:00Z"
```
simonpasquier: can we provide some details on the supported format? can it be empty?

simonpasquier (commented):
A couple more questions on top of my head:
* How does the operator know when a silences should be created, updated or deleted? How will it match a silence resource to the silences retrieved from the API?
* How do we ensure that a deleted Silence resource is effectively deleted from the Alertmanager API? Finalizers maybe?
* When Alertmanager runs in cluster, should the operator ensure that the silence exists on all instances? E.g. what happens if the replication between Alertmanagers is broken/degraded?
* We would need more details on the communication between the operator and Alertmanager API. Especially how do we provide API credentials to the operator? What are the supported authentication mechanisms?
* How does it integrate with the namespace-based tenancy model? E.g. if user creates a silence in namespace `foo`, should the operator inject a `namespace=foo` matcher into the silence definition?

Documentation/proposals/202508-silence-crd-implementation.md
simonpasquier: Thinking more about it, I see a dedicated controller (within the operator) more appealing: silences are decoupled from the statefulset and config secret provisioning. There's probably little we would reuse from the existing Alertmanager controller anyway.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,160 @@
...
+* Maintain backward compatibility with existing Alertmanager functionality
+
+## Non-Goals
+
+* Automatic cleanup of expired Silence CRs (users manage lifecycle)
+* Cross-cluster silence management (out of scope)
+* Advanced templating features (deferred to future iterations)
+* Real-time sync guarantees (minor delays acceptable)
+
+## How
```
simonpasquier, line 65: Can we document that we'll add a [feature gate](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/operator/feature_gates.go) for the time being?

Also the operator should be smart enough to disable silence management if the CRD doesn't exist or if it doesn't have RBAC permissions on the custom resources (e.g. to avoid disrupting existing users). Example: https://github.com/prometheus-operator/prometheus-operator/blob/f045cfda889010282f635abdffc4f61de8c80f66/cmd/operator/main.go#L333-L346

danielmellado: Yeah, deff, we can add --enable-silence-management=false by default initially, check for CRD existence and RBAC permissions at startup, and  disable the silence controller by default. I'll add this to the proposal.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Will modify that on the next iteration ;)

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: I was keeping alertmanagerRef approach as to provide compatibility with the original proposal. We should probably follow the established AlertmanagerConfig pattern where the Alertmanager resource defines how to select Silence resources using silenceSelector and silenceNamespaceSelector.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Totally agree - let's skip the migration stuff. For now it's just adding complexity we don't need. Much better to focus on playing nicely with existing silences that people create manually.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Agree, a dedicated controller within prometheus-operator makes way more sense. I'd avoid try to to make the current Alertmanager controller way too complex if we can't really reuse much from it.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: let's remove this completely then.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Sure thing! I'll add the link to https://github.com/giantswarm/silence-operator in the next iteration.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Yes, as commented before, let's assume migration is out of the scope

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Good point - we should expand this section to show the actual status fields like silenceID for correlation, observedGeneration for tracking changes, standard conditions (Ready/Synced/Failed), although I saw a comment about that from your side in the API PR. I'll put it there and we can sync thorugh, thanks!

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: same, thanks!

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: As I was thinking it, the expiresAt field should be required and follow RFC3339 format (e.g., "2024-01-15T10:00:00Z"). We shouldn't allow it to be empty since indefinite silences are generally a bad practice, wdyt?

danielmellado:
I've also included some of the feedback and comments in https://github.com/prometheus-operator/prometheus-operator/pull/7806. I'll hold into generating the CRD and the API itself until we agree on a format, tho :D

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,205 @@
...
+    matchType: "="
+  - name: "service"
+    value: "api-.*"
+    matchType: "=~"
+  - name: "severity"
+    value: "warning"
+    matchType: "!="
+status:
+  observedGeneration: 1
+  conditions:
```
simonpasquier: I'd expect the conditions to be related to a single Alertmanager resource. E.g.

````
status:
  bindings: # to re-use the same naming than for ServiceMonitor
  - name: ...
    namespace: ...
    conditions:
    - ...
````

File: Documentation/proposals/202508-silence-crd-implementation.md [hunk 13]
```diff
@@ -0,0 +1,205 @@
...
+## Non-Goals
+
+* Automatic cleanup of expired Silence CRs (users manage lifecycle)
+* Cross-cluster silence management (out of scope)
+* Advanced templating features (deferred to future iterations)
+* Real-time sync guarantees (minor delays acceptable)
+
+## How
+
+*Note: This feature will be protected by a feature gate `--enable-silence-management=false` by default initially, with graceful degradation when CRD or RBAC prerequisites are not met.*
```
simonpasquier: (nit) feature gates are provided via the `--enable-feature` arg and follow the same naming conventions as Kubernetes feature gates (e.g. CamelCase).
```suggestion
*Note: This feature will be protected by a feature gate `EnableSilenceCRD`, with graceful degradation when CRD or RBAC prerequisites are not met.*
```

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,205 @@
...
+the same Silence.
+
+**Cluster Synchronization**: Create silences on one healthy Alertmanager instance
+and rely on gossip protocol for replication, while monitoring sync status across
+all cluster members.
+
+**Lifecycle Management**: Use finalizers (`monitoring.coreos.com/silence-cleanup`)
+to ensure proper cleanup from Alertmanager API when Silence resources are deleted.
+
+**Coexistence**: Only manage silences created by the operator (using correlation
```
simonpasquier: Good point! More specifically how do we correlate silences in the Alertmanager API and Silence custom resources? I was thinking that the "createdBy" field could contain `<namespace>/<name>` so it's easy for the controller to associate a silence from the API with the custom resource..

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,205 @@
...
+matchers) and leave manually-created silences untouched.
+
+### Testing and Verification
+
+* Unit tests for controller logic and API conversion
+* Integration tests with real Alertmanager instances
+* End-to-end scenarios covering multi-instance deployments
```
simonpasquier: We only have unit and e2e tests. I assume that integration tests are equivalent to e2e tests.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,205 @@
...
+**API Interaction**: The controller interacts with Alertmanager via REST API
+v2 for CRUD operations, using existing authentication patterns (basic auth,
+bearer tokens, TLS client certificates, mTLS) with Kubernetes secrets.
+
+**Status Management**: The Silence status tracks the observed generation,
+a single Ready condition, and per-Alertmanager status including silence IDs
+and sync state to handle cases where multiple Alertmanager resources select
+the same Silence.
+
+**Cluster Synchronization**: Create silences on one healthy Alertmanager instance
```
simonpasquier: Can we detail a bit how the controller will detect configuration drift? E.g. what if a user modifies a silence managed by the operator? How will it be "reconciled"?

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,205 @@
...
+* Platform teams managing monitoring infrastructure
+* DevOps engineers implementing GitOps workflows
+* SRE teams needing programmatic silence management
+* Organizations requiring compliance and audit trails
+
+## Goals
+
+* Enable GitOps workflows for silence management
+* Provide RBAC integration for access control
+* Follow existing prometheus-operator patterns and conventions
```
simonpasquier: (nit) I wouldn't list this one as a goal, it's more a general guideline.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,205 @@
...
+* Support multi-Alertmanager deployments
+* Maintain backward compatibility with existing Alertmanager functionality
+
+## Non-Goals
+
+* Automatic cleanup of expired Silence CRs (users manage lifecycle)
```
simonpasquier: It triggers another point. Alertmanager will delete expired silences after some retention period (default: 120h). The option is available at `.spec.retention`. It's something that the controller might use to know if the silence has reached the retention limit.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,205 @@
...
+* Cross-cluster silence management (out of scope)
+* Advanced templating features (deferred to future iterations)
```
simonpasquier: Not sure about this one. What does it cover exactly?

Documentation/proposals/202508-silence-crd-implementation.md
simonpasquier: Yes it matches my expectation. The Alertmanager API will require a valid value.

simonpasquier:
This is looking great!

Posting again 2 questions from my initial review.

> * We would need more details on the communication between the operator and Alertmanager API. Especially how do we provide API credentials to the operator? What are the supported authentication mechanisms?

It kinda relates to https://github.com/prometheus-operator/prometheus-operator/issues/5836 and https://github.com/prometheus-operator/prometheus-operator/pull/7004. Right now the operator doesn't support Prometheus/Alertmanager configured with authentication (basic auth or token).

We need also to support a setup where the Alertmanager API is exposed via a reverse proxy (kube-rbac-proxy for instance).
 
> * How does it integrate with the namespace-based tenancy model? E.g. if user creates a silence in namespace `foo`, should the operator inject a `namespace=foo` matcher into the silence definition?

This would need to be documented too.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: I cherry picked this from the original but I have elaborated this a bit more, I'll upload a new version shortly ;)

simonpasquier:
> I'll hold into generating the CRD and the API itself until we agree on a format, tho :D

We can continue the discussion in the API PR in parallel since it's easier to comment against the actual code.

danielmellado:
> > I'll hold into generating the CRD and the API itself until we agree on a format, tho :D
> 
> We can continue the discussion in the API PR in parallel since it's easier to comment against the actual code.

let me regenerate it then XD

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+  # ... other Alertmanager configuration
+```
+
+### Implementation Strategy
+
+**Controller Integration**: A dedicated Silence controller will be implemented
+within prometheus-operator with its own informers, event handlers, and
+reconciliation logic. The controller will:
+
+1. **Watch Silence CRDs**: Detect creation, updates, and deletions
```
simonpasquier: (nit)
```suggestion
1. **Watch Silence custom resources**: Detect creation, updates, and deletions
```

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+2. **Watch Alertmanager CRDs**: React to changes in silence selectors
```
simonpasquier: (nit)
```suggestion
2. **Watch Alertmanager custom resources**: React to changes in silence selectors
```

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+- **Silence CRD Events**: Create/update/delete operations trigger API calls
+- **Generation Changes**: Spec modifications are detected via `observedGeneration`
+- **Orphaned Silences**: Periodic cleanup of silences no longer matching any CRD
+- **Alertmanager Changes**: Selector modifications trigger re-evaluation of Silence bindings
+
+**API Interaction**: The controller interacts with Alertmanager via REST API
+v2 for CRUD operations. Authentication support will follow the patterns
+established for Prometheus/Alertmanager authentication in issues
+[#5836](https://github.com/prometheus-operator/prometheus-operator/issues/5836)
+and [#7004](https://github.com/prometheus-operator/prometheus-operator/pull/7004). Until those are merged, we'll just use kube-rbac proxy for authentication.
```
simonpasquier: (nit) kube-rbac-proxy is an implementation detail. What's important is that despite the operator not supporting (yet) native Alertmanager configuration for authn/authz, we can't assume that Alertmanager API doesn't require credentials (kube-rbac-proxy being 1 example but users may use other reverse proxies).

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  name: secure-alertmanager
+  namespace: monitoring
+spec:
+  replicas: 3
+  # Basic Authentication
+  basicAuth:
```
simonpasquier: I wouldn't put these fields at the `spec` level. Maybe we need to add a dedicated field for this?

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  name: rbac-protected-alertmanager
+  namespace: monitoring
+spec:
+  replicas: 3
+  # Access via kube-rbac-proxy service
+  externalURL: "https://alertmanager-proxy.monitoring.svc.cluster.local:8443"
```
simonpasquier: I don't think that we should use externalURL (instead we should use the internal service name).

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+resources is achieved using the `createdBy` field in the Alertmanager API, which contains `<namespace>/<name>`
+(e.g., `monitoring/maintenance-silence`) to uniquely identify the source custom resource.
+
+**CRUD Operations**:
+- **Create**: When a new Silence CRD is created, the controller calls the Alertmanager API to create the silence and stores the returned silence ID in the status
+- **Update**: When a Silence CRD is modified, the controller uses the stored silence ID to update the existing silence via the Alertmanager API
+- **Delete**: When a Silence CRD is deleted, finalizers ensure the controller calls the Alertmanager API to delete the silence before removing the CRD
+- **Drift Detection**: The controller periodically compares Silence CRDs with actual silences in Alertmanager (matching by `createdBy` field) and reconciles any differences
+
+**Namespace-based Tenancy**: To support multi-tenant environments, the Alertmanager
```
simonpasquier: we already have `.spec.alertmanagerConfigMatcherStrategy`. I think that it should deserve its own section.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+- **Invalid Silence Data**: Validates before API call, rejects with condition message
+- **Network Timeouts**: Configurable timeout, retry logic, eventual failure reporting
+
+**Resource Conflicts**:
+- **Duplicate Silence IDs**: Use `createdBy` field to identify ownership, ignore non-operator silences
+- **Selector Changes**: Cleanup old bindings, establish new ones based on current selectors
+- **Namespace Deletions**: Handle gracefully via owner references and finalizers
+
+**Controller Failures**:
+- **Leader Election**: Only active controller processes silences to avoid conflicts
```
simonpasquier: we have no leader election.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+and require additional deployment complexity. Integration into prometheus-operator
+provides a unified experience and reuses existing infrastructure.
+
+### External Operator
+
+**Why not**: Using an external silence operator (like [silence-operator](https://github.com/giantswarm/silence-operator)) would
+fragment the ecosystem and require additional deployment complexity.
+Integration into prometheus-operator provides unified experience.
+
+## Action Plan
```
simonpasquier, line 363: (nit) we need to add the feature gate too.

Documentation/proposals/202508-silence-crd-implementation.md
```diff
@@ -0,0 +1,386 @@
...
+Supported authentication mechanisms:
+- **No Authentication**: Direct HTTP access to Alertmanager API
+- **Reverse Proxy Authentication**: Support for proxy layers like kube-rbac-proxy
+  that handle authentication transparently
+
+The operator will read authentication configuration from the Alertmanager CRD
+specification, reusing existing authentication fields where possible to maintain
+consistency with Prometheus authentication patterns.
+
+### Authentication and Reverse Proxy Support
```
simonpasquier, line 182: I'm still a bit unclear on how we should inject credentials into the Alertmanager CR. In particular we want to avoid duplication with related efforts such as #5836.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Agree, I'll remove the specific references to kube-rbac-proxy

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: I removed this from the spec and addressed the comment by adding a paragraph.

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: ack, yeah, that makes sense. If later at the implementation we need this we can just add it back, but I'll defer to the CRD PR for this, thanks Simon!

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: added a specific section about that

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: removed!

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: there's a Feature Gate in the action plan a bit below there, let me clarify it in any case, thanks!

Documentation/proposals/202508-silence-crd-implementation.md
danielmellado: Should we just defer al this configurations to #5836 ? I modified how is this explained to directly do that and align.

PR #7825, opened 2025-08-21
chore: remove po-lint from .gitignore

therealdwright:
This tool has been deprecated so the corresponding gitignore entry
can now be removed.


simonpasquier (approved):
Thanks!

Issue #7792, opened 2025-08-07
Update service monitor status binding only if a change is detected


simonpasquier:
### Component(s)

ServiceMonitor

### What is missing? Please describe.

Every reconciliation of a workload resource will trigger status updates of the selected service monitors even if there's no change to the status of the configuration resource. Eventually the operator should avoid these unnecessary updates.

## Environment
Kubernetes Version:
Prometheus-Operator Version:

PR #7827, opened 2025-08-21
Update service monitor status binding only if a change is detected

yp969803:
fixes: #7792  

## Description

Update service monitor status binding only if a change is detected

_Closes: #ISSUE-NUMBER_

## Type of change

## Verification

## Changelog entry

```release-note

```

### DIFF
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -278,6 +278,12 @@ func UpdateServiceMonitorStatus(
 		if binding.Namespace == c.workload.GetNamespace() &&
 			binding.Name == c.workload.GetName() &&
 			binding.Resource == c.gvr.Resource {
+			if binding.Conditions[0].ObservedGeneration == conditions[0].ObservedGeneration &&
+				binding.Conditions[0].Status == conditions[0].Status &&
+				binding.Conditions[0].Reason == conditions[0].Reason &&
+				binding.Conditions[0].Message == conditions[0].Message {
+				return nil
+			}
 			binding.Conditions = conditions
 			found = true
 			break
--- a/pkg/prometheus/operator.go
+++ b/pkg/prometheus/operator.go
@@ -278,14 +278,8 @@ func UpdateServiceMonitorStatus(
 		if binding.Namespace == c.workload.GetNamespace() &&
 			binding.Name == c.workload.GetName() &&
 			binding.Resource == c.gvr.Resource {
-
-			for _, cond := range binding.Conditions {
-				if cond.ObservedGeneration == conditions[0].ObservedGeneration &&
-					cond.Status == conditions[0].Status &&
-					cond.Reason == conditions[0].Reason &&
-					cond.Message == conditions[0].Message {
-					return nil
-				}
+			if configResStatusConditionsEqual(binding.Conditions, conditions) {
+				return nil
 			}
 			binding.Conditions = conditions
 			found = true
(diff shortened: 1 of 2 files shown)


yp969803:
@simonpasquier PTAL!

pkg/prometheus/operator.go
```diff
@@ -278,6 +278,12 @@ func UpdateServiceMonitorStatus(
 		if binding.Namespace == c.workload.GetNamespace() &&
 			binding.Name == c.workload.GetName() &&
 			binding.Resource == c.gvr.Resource {
+			if binding.Conditions[0].ObservedGeneration == conditions[0].ObservedGeneration &&
```
simonpasquier: we need to compare all conditions to be future-proof.

yp969803: done

yp969803:
@simonpasquier merge request

pkg/prometheus/operator.go
```diff
@@ -278,6 +278,15 @@ func UpdateServiceMonitorStatus(
 		if binding.Namespace == c.workload.GetNamespace() &&
 			binding.Name == c.workload.GetName() &&
 			binding.Resource == c.gvr.Resource {
+
+			for _, cond := range binding.Conditions {
```
simonpasquier: Let's always assume that we can have multiple conditions. We need to iterate over the `conditions` list too and compute the diff between `conditions` and `binding.Conditions`. If the diff is empty then we can  return early. Otherwise we apply the diff.

yp969803: @simonpasquier done

pkg/prometheus/operator.go
```diff
@@ -299,3 +302,20 @@ func UpdateServiceMonitorStatus(
...
 	return nil
 }
+
+func configResStatusConditionsEqual(a, b []monitoringv1.ConfigResourceCondition) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for i := range a {
+		if a[i].Type != b[i].Type ||
```
simonpasquier: It implies that a and b have identical sort order.

yp969803: i assumed that both the condiotions arrays are in correct order

simonpasquier: for safety and reliability, it's better to implement in a way that we don't depend on same order.

yp969803: ok sorted arrays  based on timestamp

pkg/prometheus/operator.go
yp969803: dome @simonpasquier

pkg/prometheus/operator.go
```diff
@@ -299,3 +303,30 @@ func UpdateServiceMonitorStatus(
...
 	return nil
 }
+
+func configResStatusConditionsEqual(a, b []monitoringv1.ConfigResourceCondition) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	ac := append([]monitoringv1.ConfigResourceCondition(nil), a...)
+	bc := append([]monitoringv1.ConfigResourceCondition(nil), b...)
```
simonpasquier, line 313: we don't need to copy the slices because the original slices are left unchanged.

pkg/prometheus/operator.go
```diff
@@ -299,3 +303,30 @@ func UpdateServiceMonitorStatus(
...
+
+	sort.Slice(ac, func(i, j int) bool {
```
simonpasquier, line 315: We need to sort by Type because the API validation ensures that duplicate Type values aren't possible.

pkg/prometheus/operator.go
yp969803: ok

pkg/prometheus/operator.go
yp969803: done

pkg/prometheus/operator.go
simonpasquier: I stand corrected: @yp969803 rightfully pointed out that the slices will be updated but it's ok.

simonpasquier (approved):
Thanks!

PR #7862, opened 2025-09-02
enhancement: `s/staticConfig_targets/staticConfig_hosts`

### LINKED ISSUE #3480: "targets" field in staticConfig in Probe custom resource is wrongly called "static"

rexagod:
## Description

Deprecate Probe `targets` in favor of Probe `hosts` to address the inconsistency between the currently exported `Targets` field and its corresponding JSON representation, "static".

Resumes: #6047
Fixes: #3480 

## Type of change

## Changelog entry

```release-note
- Introduce a `hosts` field in `staticConfig`, that will replace `targets` within the same schema section, in an effort to make the exported field match the JSON representation (earlier: `Targets,static", now: "Hosts,hosts").

```

### DIFF
--- a/Documentation/api-reference/api.md
+++ b/Documentation/api-reference/api.md
@@ -13031,6 +13031,18 @@ More info: <a href="https://prometheus.io/docs/prometheus/latest/configuration/c
 </em>
 </td>
 <td>
+<p>The list of hosts to probe.
+Deprecated: Use <code>hosts</code> instead.</p>
+</td>
+</tr>
+<tr>
+<td>
+<code>hosts</code><br/>
+<em>
+[]string
+</em>
+</td>
+<td>
 <p>The list of hosts to probe.</p>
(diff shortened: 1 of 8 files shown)


simonpasquier:
Thanks! Can we add CEL rules to prevent both fields being defined?

simonpasquier:
Can we also add unit tests to assert that `hosts` takes prioritiy over `targets`?

